PR #1852, opened 2021-01-23
ci: check that Go dependencies are tidied

dzhu:
## Description

In order to keep everybody's dependencies in sync, avoid changing
people's local files after builds, and clean up old dependencies as soon
as possible, it is helpful to ensure that the committed versions of
`go.mod` and `go.sum` files are not changed by `go mod tidy`.

## Test Plan

- [x] [run](https://app.circleci.com/pipelines/github/determined-ai/determined?branch=go-mod-tidy) the same steps with out-of-date files

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -22,6 +22,8 @@ build:
 
 .PHONY: check
 check:
+	go mod tidy
+	git diff --quiet go.mod go.sum
 	golangci-lint --build-tags integration run
 
 .PHONY: fmt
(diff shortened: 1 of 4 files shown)


stoksc (approved):
lgtm. should we add this to `make check` instead though (instead of just failing CI)?

dzhu:
I considered it, but that felt weird to me because (1) fixing the issue requires making a commit and (2) the check itself can alter files in the tree, neither of which is true for other checks/tests. Maybe add `go mod tidy` to `make fmt`? Or we can just do the check in `make check`. I'm not really opposed, just trying to guess at developer ergonomics here.

stoksc:
> Or we can just do the check in make check.

This is sort of what i was thinking? I just think it would be nicer to catch pre-push (I have a hook I use, if it was in check it would save me time)

PR #1853, opened 2021-01-25
fix: fix CLI log tailing with elastic and deep pagination [DET-4883]

stoksc:
## Description
This change stops the CLI from making requests with limit+offset that cause deep pagination in Elasticsearch.

## Test Plan
- [x] `pip install -e .` and test `--head`, `--tail` and everything else with a cluster backed by Elasticsearch.

## Checklist


dzhu (approved):
Oh yeah, the nested function made _way_ more sense when this wasn't a streaming API and we had to loop to do following.

PR #1859, opened 2021-01-26
chore: let CLI verify the master using combined system/custom certs [DET-4666]

dzhu:
## Description

Previously, when the CLI found a stored custom TLS certificate for the
master, it would verify all requests to the master using exactly that
certificate. That worked well enough for a single master, but led to a
thorny development experience when switching between different masters:
with a custom certificate around, the CLI could no longer connect to a
master using a certificate signed by a well-known CA without an
additional prompt (and often not at all, since the verification
available in the CLI needs a full chain to the root and root certs are
usually left out of chains in practice).

This change avoids that issue by instead constructing and using a
temporary bundle that contains both the custom bundle and all well-known
CA certs (as defined by Certifi, which Requests defaults to
normally). For now, we refrain from handling multiple custom
certificates.

This may be a bit dirty. An alternative technique would be to dive into
Requests/urllib3 and do some jiggery-pokery with custom transport
adapters and SSL contexts or something, but I looked into that and
decided it would be much too hairy to be worth it.

## Test Plan

- [x] manually test switching the CLI between servers with CA-signed cert and non-CA-signed cert
- [x] add some unit tests for TLS verification with different settings


common/tests/test_tls.py
```diff
@@ -0,0 +1,26 @@
...
+    for bundle, raises in [
+        (False, False),
+        (True, True),
+        (UNTRUSTED_CERT_FILE, False),
+        (None, True),
+    ]:
+        request.set_master_cert_bundle(bundle)  # type: ignore
+
+        request.get(TRUSTED_DOMAIN, "", authenticated=False)
+        ctx = pytest.raises(requests.exceptions.SSLError) if raises else contextlib.ExitStack()
```
rb-determined-ai, line 24: mabye use `contextlib.nullcontext` instead of `ExitStack`?  Or use `contextlib.ExitStack` always and optionally call `stack.enter_context()` on `pytest.raises`?

common/determined_common/api/request.py
```diff
@@ -12,18 +16,40 @@
...
-_master_cert_bundle = None
+# False to disable cert verification entirely. If set to a path, it should always be a temporary
+# file that we own and can delete.
+_master_cert_bundle = None  # type: Optional[Union[str, bool]]
 
 # The name we use to verify the master.
 _master_cert_name = None
 
 
+@atexit.register
```
rb-determined-ai, line 27: Just a thought, might not be worth it, but can we `@atexit.register` a closure bound to the `path` string inside `_set_master_bundle()`?  Maybe give us an iota more of protection in multithreading situations.

rb-determined-ai (approved):
I am not a fan of anything `@atexit`-related, but I can't really think of a better strategy in this case.

Implementation looks good.

common/tests/test_tls.py
dzhu: `nullcontext` is new in 3.7, so no good here. But I like the idea of actually using `ExitStack` for what it's for.

common/determined_common/api/request.py
dzhu: Okay, I think I like that. It means that there will be multiple files sitting around and only removed on exit if `set_master_cert_bundle` is called multiple times, but the logic is cleaner and multiple calls should be rare.

PR #1864, opened 2021-01-26
fix: CI failures caused by resource pool merge

armandmcqueen:
## Description

Update the CircleCI config to use the newly renamed arguments to det-deploy.

Changed test that was designed around the assumption that the experiment and the tensorboard will run on the same instance from E2E GPU (where the experiment will run in the GPU resource pool and the tensorboard will run in the CPU resource pool) to E2E CPU.

Fix bug in the code where Tensorboards were incorrectly being routed to the default GPU resource pool when Tensorboards are always zero slot tasks. This was caused by all commands defaulting to slots=1, the default resource pool being chosen based on whether slots=0, and the Tensorboard manager not setting slots=0 until AFTER the default resource pool was set.

## Checklist

### DIFF
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -53,7 +53,6 @@ entrypoint: model_def:NoOpTrial
 """
 
 
-@pytest.mark.skip()  # type: ignore
 @pytest.mark.slow  # type: ignore
 @pytest.mark.e2e_cpu  # type: ignore
 def test_start_tensorboard_for_shared_fs_experiment(tmp_path: Path) -> None:
@@ -102,9 +101,9 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str, 
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
 
 
-@pytest.mark.skip()  # type: ignore
+# @pytest.mark.skip()  # type: ignore
 @pytest.mark.slow  # type: ignore
-@pytest.mark.e2e_gpu  # type: ignore
+@pytest.mark.e2e_cpu  # type: ignore
 def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
     """
     Start 3 random experiments configured with the s3 and shared_fs backends,
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -53,7 +53,7 @@ entrypoint: model_def:NoOpTrial
 """
 
 
-@pytest.mark.skip() # type: ignore
+@pytest.mark.skip()  # type: ignore
 @pytest.mark.slow  # type: ignore
 @pytest.mark.e2e_cpu  # type: ignore
 def test_start_tensorboard_for_shared_fs_experiment(tmp_path: Path) -> None:
@@ -102,7 +102,7 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str, 
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
 
 
-@pytest.mark.skip() # type: ignore
+@pytest.mark.skip()  # type: ignore
 @pytest.mark.slow  # type: ignore
 @pytest.mark.e2e_gpu  # type: ignore
 def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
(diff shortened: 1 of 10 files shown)


e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -53,6 +53,7 @@ def s3_config(num_trials: int, secrets: Dict[str, str]) -> str:
 """
 
 
+@pytest.mark.skip()  # type: ignore
```
armandmcqueen, line 56: This are skipped because, as far as I understand, we are pretending the /tmp folder is a shared file system, but now that the Tensorboard is running on a different instance than the experiment, that doesn't hold true so the tests can't pass.

armandmcqueen:
The e2e tests have all passed, they are just failing to upload the CloudWatch logs to S3 due to:

```
Got an error possibly due to concurrent uploads: An error occurred (LimitExceededException) when calling the CreateExportTask operation: Resource limit exceeded..
```

mackrorysd (approved):

e2e_tests/tests/command/test_tensorboard.py
mackrorysd: test_pytorch_load assumes this as well - maybe we should have the test masters spin up a little NFS server to replace our use of /tmp or something.

shiyuann: blocking: it shouldn't matter if shared_fs actually uses shared fs. The TensorBoard will access the /tmp folder, where stores the checkpoints. I don't really understand the reasoning 

> "but now that the Tensorboard is running on a different instance than the experiment, that doesn't hold true so the tests can't pass." 

TensorBoard has always been running on a different instance. What has been changed?

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -101,6 +102,7 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str,
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
 
 
+@pytest.mark.skip()  # type: ignore
```
shiyuann, line 105: blocking: why do you want to skip this?

shiyuann (changes requested):

e2e_tests/tests/command/test_tensorboard.py
shiyuann: If using a different resource pool, TensorBoard will be running on a different agent than the experiment. But this is an e2e cpu test, all the tasks should run on the same agent

shiyuann: And it shouldn't matter if TensorBoard actually uses a shared fs or just a folder.

armandmcqueen: This is not an e2e CPU test , it is run as part of the e2e GPU test - https://app.circleci.com/pipelines/github/determined-ai/determined/9229/workflows/a07e5abd-b93e-4cf0-9849-f5860f519ead/jobs/255468

LMK if I have root caused this incorrectly, but based on the fact that it works if we run the tensorboard in the GPU pool, I believe this is the cause of the 'metrics not found'

e2e_tests/tests/command/test_tensorboard.py
armandmcqueen: Because as far as I understand, the metrics are being generated by the GPU instance and saved to `/tmp` so when the CPU instance launches a tensorboard and looks for metrics in `/tmp/`, it can't find them.

e2e_tests/tests/command/test_tensorboard.py
shiyuann: The link you posted is `test_start_tensorboard_for_multi_experiment`. We should change it to be `e2e_cpu` instead.

e2e_tests/tests/command/test_tensorboard.py
shiyuann: we should change this test to use `e2e_cpu`

e2e_tests/tests/command/test_tensorboard.py
armandmcqueen: Oh you're right , I'll remove this skip

shiyuann (approved):

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -101,8 +101,9 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str,
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
 
 
+# @pytest.mark.skip()  # type: ignore
```
shiyuann, line 104: can you just delete this line?

shiyuann:
@armandmcqueen `Semantic Pull Request` is stuck and you also need to approve the GPUs tests on the CircleCI

PR #1869, opened 2021-01-27
fix: allow larger gRPC response bodies

dzhu:
## Description

For large experiments, some API endpoints can result in response bodies
larger than 4 MiB, which is the default maximum size that the gRPC-HTTP
proxy allows. This bumps the limit to 64 MiB to push the problem farther
away.

## Test Plan

- [x] make a request that fails with `grpc: received message larger than max...`, apply change, check that request now succeeds

## Commentary (optional)

Pagination for our clients is on the way, which is the real solution.


stoksc (approved):
Lgtm

hamidzr (approved):
lgtm https://github.com/grpc/grpc-go/blob/master/rpc_util.go#L300

PR #1871, opened 2021-01-27
build: fix for `make check-schemas` on non-GNU build machines

davetroiano:
Checking schemas had assumed GNU version of the `find` command,
which assumes `.` as the starting point if none is provided. This
could cause build errors on Mac, whose built-in `find` utility
requires a starting point ("find: illegal option -- n").

## Checklist


rb-determined-ai (approved):
Sorry about that, thank you

PR #1875, opened 2021-01-27
feat: HP constraints harness exception handling [DET-4867]

naren-determined:
## Description

Enable the harness to catch `det.InvalidHP` exceptions raised in user code, and propagate that information to the searcher. Note: this is a user-facing API change, thought not a breaking change.

## Test Plan

In any `model_def.py` raise a `det.InvalidHP("msg")` in either `__init__`, `train_batch`, or `evaluate_batch` for PyTorch models. For Keras and Estimator models, raise in either `__init__` or in a callback with validation metrics. Ensure the trial stops training, and that an InvalidHP is raised (should be apparent in the trial logs).

## Checklist

### DIFF
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -16,12 +16,13 @@ class InvalidHP(Exception):
     Attributes:
         message -- explanation of the error
     """
+
     def __init__(self, message="Invalid hyperparameter exception encountered"):
         self.message = message
         super().__init__(self.message)
 
     def __str__(self):
-        return(repr(self.message))
+        return repr(self.message)
 
 
 def _get_gpus(limit_gpus: Optional[int]) -> Tuple[bool, List[str], List[int]]:
@@ -46,7 +47,7 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: workload.stream) -> Any:
+def _catch_invalid_hp(workloads: iter) -> Any:
     try:
         yield
     except InvalidHP as e:
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -10,6 +10,20 @@ from determined import constants, gpu, horovod, workload
 from determined_common import api
 
 
+class InvalidHP(Exception):
+    """Raised by user when InvalidHP exception is encountered
+
+    Attributes:
+        message -- explanation of the error
+    """
+    def __init__(self, message="Invalid hyperparameter exception encountered"):
+        self.message = message
+        super().__init__(self.message)
+
+    def __str__(self):
+        return(repr(self.message))
+
+
 def _get_gpus(limit_gpus: Optional[int]) -> Tuple[bool, List[str], List[int]]:
     gpu_ids, gpu_uuids = gpu.get_gpu_ids_and_uuids()
     if limit_gpus is not None:
@@ -29,6 +43,18 @@ def _catch_sys_exit() -> Any:
             "This might be raised by directly calling or using a library calling sys.exit(). "
             "Please remove any calls to sys.exit() from your model code."
         ) from e
+
+
+@contextlib.contextmanager
+def _catch_invalid_hp(workloads: workload.stream) -> Any:
+    try:
+        yield
+    except InvalidHP as e:
+        logging.info("Invalid hyperparameter exception encountered: e\n", e)
+        wkld, args, response_func = next(workloads)
+        wkld.CompletedMessage.ExitedReason = workload.InvalidHP
+        response_func(wkld.CompletedMessage)
+        sys.exit()
 
 
 def _make_local_execution_exp_config(input_config: Optional[Dict[str, Any]]) -> Dict[str, Any]:
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -13,8 +13,10 @@ from determined_common import api
 class InvalidHP(Exception):
     def __init__(self, msg: str = "...") -> None:
         if not isinstance(msg, str):
-            raise TypeError("InvalidHP exceptions can be initialized with a custom message, "
-                            f"but it must be a string, not {type(msg).__name__}")
+            raise TypeError(
+                "InvalidHP exceptions can be initialized with a custom message, "
+                f"but it must be a string, not {type(msg).__name__}"
+            )
         self.msg = msg
 
 
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -17,11 +17,11 @@ class InvalidHP(Exception):
         message -- explanation of the error
     """
 
-    def __init__(self, message="Invalid hyperparameter exception encountered"):
+    def __init__(self, message="Invalid hyperparameter exception encountered") -> None:
         self.message = message
         super().__init__(self.message)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return repr(self.message)
 
 
@@ -47,13 +47,13 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: iter) -> Any:
+def _catch_invalid_hp(workloads: iter(workload)) -> Any:
     try:
         yield
     except InvalidHP as e:
         logging.info("Invalid hyperparameter exception encountered: e\n", e)
         wkld, args, response_func = next(workloads)
-        wkld.CompletedMessage.ExitedReason = workload.InvalidHP
+        wkld.CompletedMessage.ExitedReason = workload.CompletedMessage.InvalidHP
         response_func(wkld.CompletedMessage)
         sys.exit()
 
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -47,7 +47,7 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: Iterator[workload.Workload]) -> Any:
+def _catch_invalid_hp(workloads: Iterator[Any]) -> Any:
     try:
         yield
     except InvalidHP as e:
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -3,7 +3,7 @@ import logging
 import os
 import pathlib
 import sys
-from typing import Any, Dict, Iterator, List, Optional, Tuple
+from typing import Any, Dict, Iterable, Iterator, List, Optional, Tuple
 
 import determined as det
 from determined import constants, gpu, horovod, workload
@@ -47,7 +47,7 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: iter[workload]) -> Any:
+def _catch_invalid_hp(workloads: Iterable[workload.Workload]) -> Any:
     try:
         yield
     except InvalidHP as e:
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -6,7 +6,7 @@ import sys
 from typing import Any, Dict, Iterator, List, Optional, Tuple
 
 import determined as det
-from determined import constants, gpu, horovod, workload
+from determined import constants, gpu, horovod, util, workload
 from determined_common import api
 
 
@@ -47,15 +47,16 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: Iterator[Any]) -> Any:
+def _catch_init_invalid_hp(workloads: Iterator[Any]) -> Any:
     try:
         yield
     except InvalidHP as e:
-        logging.info("Invalid hyperparameter exception encountered: e\n", e)
+        logging.info("Invalid hyperparameter exception in trial __init__: {}".format(e))
         wkld, args, response_func = next(workloads)
-        wkld.CompletedMessage.ExitedReason = "INVALID_HP"
-        response_func(wkld.CompletedMessage)
-        sys.exit()
+        response_func(
+            util.wrap_metrics({}, False, True)
+        )
+        raise
 
 
 def _make_local_execution_exp_config(input_config: Optional[Dict[str, Any]]) -> Dict[str, Any]:
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -17,7 +17,7 @@ class InvalidHP(Exception):
         message -- explanation of the error
     """
 
-    def __init__(self, message="Invalid hyperparameter exception encountered") -> None:
+    def __init__(self, message: str = "Invalid hyperparameter exception encountered") -> None:
         self.message = message
         super().__init__(self.message)
 
@@ -47,13 +47,13 @@ def _catch_sys_exit() -> Any:
 
 
 @contextlib.contextmanager
-def _catch_invalid_hp(workloads: iter(workload)) -> Any:
+def _catch_invalid_hp(workloads: iter[workload]) -> Any:
     try:
         yield
     except InvalidHP as e:
         logging.info("Invalid hyperparameter exception encountered: e\n", e)
         wkld, args, response_func = next(workloads)
-        wkld.CompletedMessage.ExitedReason = workload.CompletedMessage.InvalidHP
+        wkld.CompletedMessage.ExitedReason = "INVALID_HP"
         response_func(wkld.CompletedMessage)
         sys.exit()
 
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -53,9 +53,7 @@ def _catch_init_invalid_hp(workloads: Iterator[Any]) -> Any:
     except InvalidHP as e:
         logging.info("Invalid hyperparameter exception in trial __init__: {}".format(e))
         wkld, args, response_func = next(workloads)
-        response_func(
-            util.wrap_metrics({}, False, True)
-        )
+        response_func(util.wrap_metrics({}, False, True))
         raise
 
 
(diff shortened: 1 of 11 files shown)


harness/determined/layers/_workload_manager.py
```diff
@@ -132,7 +132,7 @@ def yield_train_for_step(
             )
 
         def _respond(in_response: workload.Response) -> None:
-
+            # raise ValueError(in_response)
```
shiyuann, line 135: nit: remove this line

harness/determined/_train_context.py
```diff
@@ -144,6 +145,26 @@ def set_stop_requested(self, stop_requested: bool) -> None:
         )
         self._stop_requested = stop_requested
 
+    def get_invalid_hp(self) -> bool:
```
shiyuann, line 148: Is this intended to be used by users?

harness/determined/_train_context.py
```diff
@@ -144,6 +145,26 @@ def set_stop_requested(self, stop_requested: bool) -> None:
...
+        """
+        Return whether a trial invalid hp has been requested.
+        """
+        return self._invalid_hp
+
+    def set_invalid_hp(self, invalid_hp: bool) -> None:
```
shiyuann, line 154: Is this supposed to be used by users or should they just raise exception?

harness/determined/exec/harness.py
```diff
@@ -129,21 +130,22 @@ def build_and_run_training_pipeline(env: det.EnvContext) -> None:
...
 
                 with det._catch_sys_exit():
-                    controller = load.prepare_controller(
-                        env,
-                        iter(workload_mgr),
-                        load_path,
-                        socket_mgr.get_rendezvous_info(),
-                        hvd_config,
-                    )
+                    with det._catch_init_invalid_hp(workloads):
```
shiyuann, line 141: If you catch the exception here, you will only be able to catch when executing users' trial init function.

harness/determined/exec/harness.py
```diff
@@ -224,4 +226,8 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    try:
```
shiyuann, line 229: blocking: you should actually catch the exception on running `prepare_controller()` and `controller.run()` not at the top level.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -143,18 +143,51 @@ def _set_data_loaders(self) -> None:
...
-                )
+                except det.InvalidHP as e:
+                    logging.info(
+                        "Invalid hyperparameter exception in trial validation step: {}".format(e)
+                    )
+                    response_func(
+                        util.wrap_metrics(
+                            self._compute_validation_metrics(),
+                            self.context.get_stop_requested(),
+                            self.context.set_invalid_hp(True),
```
shiyuann, line 187: I think you actually intended to just put a `True` here rather than `self.context.set_invalid_hp(True),`

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -143,18 +143,51 @@ def _set_data_loaders(self) -> None:
...
+                        )
+                    )
+                    raise
             elif w.kind == workload.Workload.Kind.COMPUTE_VALIDATION_METRICS:
-                response_func(
-                    util.wrap_metrics(
-                        self._compute_validation_metrics(), self.context.get_stop_requested()
+                try:
```
shiyuann, line 171: I think you actually want to put only `self._compute_validation_metrics()` in the "try" rather than the call to `response_func`.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -143,18 +143,51 @@ def _set_data_loaders(self) -> None:
     def run(self) -> None:
         for w, args, response_func in self.workloads:
             if w.kind == workload.Workload.Kind.RUN_STEP:
-                response_func(
-                    util.wrap_metrics(
-                        self._train_for_step(w.step_id, w.num_batches, w.total_batches_processed),
-                        self.context.get_stop_requested(),
+                try:
+                    response_func(
```
shiyuann, line 147: same

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -773,19 +773,41 @@ def _control_loop(self) -> None:
...
+                        "Invalid hyperparameter exception in trial train step: {}".format(e)
+                    )
+                    response_func(util.wrap_metrics({}, False, True))
+                    raise
             elif wkld.kind == workload.Workload.Kind.COMPUTE_VALIDATION_METRICS:
-                response_func(
-                    det.util.wrap_metrics(
-                        self._compute_validation_metrics(), self.context.get_stop_requested()
+                try:
+                    response_func(
```
shiyuann, line 792: same

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -773,19 +773,41 @@ def _control_loop(self) -> None:
...
             logging.debug(f"Received wkld {wkld.kind} with args {args}.")
             if wkld.kind == workload.Workload.Kind.RUN_STEP:
-                # Configure the state for a training step.
-                self.train_response_func = response_func
-                self.train_workload_batches = 0
-                self.train_workload_metrics = []
-                self.train_workload_len = wkld.num_batches
-                self.multiplexer.set_batches_requested(wkld.num_batches)
-                break
+                try:
```
shiyuann, line 776: There is no need to capture an exception here. You should instead capture the exception in `TFKerasTrialController.run`.

shiyuann (commented):
@rb-determined-ai probably has an opinion on how this should be written in the harness. One question I have is: have we already merged the change in the master? If not, this might break the master if you haven't confirmed.

harness/determined/exec/harness.py
naren-determined: This is the intended functionality for this context manager. We will handle invalid hp exceptions raised in training and validations workloads in the individual trial controllers.

harness/determined/_execution.py
```diff
@@ -6,10 +6,25 @@
...
+    """
+
+    def __init__(self, message: str = "Invalid hyperparameter exception encountered") -> None:
+        self.message = message
+        super().__init__(self.message)
+
+    def __str__(self) -> str:
+        return repr(self.message)
+
+
```
rb-determined-ai, line 27: I actually think that this would probably suffice:
```python
class InvalidHP(Exception):
    def __init__(self, msg: str = "...") -> None:
        if not isinstance(msg, str):
            raise TypeError("InvalidHP exceptions can be initialized with a custom message, "
                f"but it must be a string, not {type(msg).__name__}")
        self.msg = msg
```
I don't think the `__str__` is needed or the super().__init__.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: I think since the `except` clause only catches InvalidHP, and since InvalidHP is only thrown by user code, and since the response_func call is quite different in the two cases, I think this is actually a pretty reasonable strategy.

master/pkg/searcher/asha.go
```diff
@@ -272,6 +272,8 @@ func (s *asyncHalvingSearch) trialExitedEarly(
 		s.InvalidTrials++
 		// Remove metrics associated with InvalidHP trial across all rungs
 		highestRungIndex := s.TrialRungs[requestID]
+		rung := s.Rungs[highestRungIndex]
```
rb-determined-ai, line 275: Why is this in this PR?  I don't understand how it is related.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -650,6 +650,16 @@ def run(self) -> None:
...
+        except det.InvalidHP as e:
+            for _, _, response_func in self.workloads:
+                logging.info("Invalid hyperparameter exception in trial: {}".format(e))
+                response_func(
+                    det.util.wrap_metrics(
+                        {},
+                        self.context.get_stop_requested(),
+                        True,
+                    )
+                )
```
rb-determined-ai, line 662: Hm... this is maybe not going to do quite what you are thinking.  this hack of reading from the workloads is only valid before any workloads have been read (like when you initialize the trial), but I don't see how you can guarantee that if you are catching InvalidHP exceptions from the entire fit loop.

harness/determined/exec/harness.py
rb-determined-ai: It might be worth mentioning why it is designed this way (I helped with this design haha).  This is ultimately a way for us to hack around the layers abstraction of the harness, which in some cases needs to say alive long enough for the master to ask for a checkpoint (right?).

But if there's a failure at init time, we don't need to do that, and we can just exit out of the harness.

(if we could push checkpoints, then we could always decide if we want a checkpoint at the moment we catch the error, then push a checkpoint, then exit, and things would be a bit simpler)

harness/determined/layers/_workload_manager.py
```diff
@@ -144,32 +143,42 @@ def _respond(in_response: workload.Response) -> None:
...
+                out_response = {
+                    "type": "WORKLOAD_COMPLETED",
+                    "workload": wkld,
+                    "start_time": start_time,
+                    "end_time": _current_timestamp(),
+                    "metrics": metrics,
+                }
+                out_response["exited_reason"] = "INVALID_HP"
+                respond(out_response)
+            else:
```
rb-determined-ai, line 156: Is it valid to just `return` here, rather than add an extra layer of indentation to the whole rest of the function?

harness/determined/exec/harness.py
rb-determined-ai: Actually, I think this is the right choice.  When we catch an exception during the `trial.__init__()`, we need to send a response to the master and just exit.  We could call "sys.exit()", but we already catch sys.exit calls and assume the user messed up.  So instead the choice was to let the InvalidHP exception continue to bubble up.

But then you don't want to show an InvalidHP stacktrace necessarily, so we drop that exception here.

master/pkg/searcher/asha.go
naren-determined: This was a bug in the searcher code where some trials weren't being closed out properly, and since it's only a two line change figured I would merge it in here.

naren-determined: Since we are bypassing the `promoteAsync` function due to an InvalidHP, we need to decrement the oustanding trials in a rung normally done by that function, so I do that here.

harness/determined/layers/_workload_manager.py
naren-determined: Yep, good call here.

harness/determined/keras/_tf_keras_trial.py
naren-determined: @rb-determined-ai Then I should probably take a similar approach to `_pytorch_trial.py` and catch exceptions in this area of the code: https://github.com/determined-ai/determined/blob/master/harness/determined/keras/_tf_keras_trial.py#L778 for both keras and estimators?

rb-determined-ai: Hm, yes I think so.   It will be a bit trickier though, because the places where user code is called are a bit more numerous, due the the Keras Callbacks support.

harness/determined/exec/harness.py
shiyuann: I have added a checkpoint API in [a recent PR](https://github.com/determined-ai/determined/pull/1968). Do you think it would help this PR to get less complicated? @rb-determined-ai

harness/determined/exec/harness.py
shiyuann: Although it works, it looks complicated and hacky. Do you know how would this be simplified in a push-arch world?

harness/determined/pytorch/_pytorch_trial.py
shiyuann: > since the response_func call is quite different in the two cases, I think this is actually a pretty reasonable strategy.

I don't understand this. We only run user code in `self._compute_validation_metrics()`, which should be the only place to throw InvalidHP. How would response_func interact with InvalidHP?

harness/determined/exec/harness.py
naren-determined: I think I changed this actually, in a recent commit so if you take another look at the harness.py file I don't think this is an issue anymore.

harness/determined/pytorch/_pytorch_trial.py
naren-determined: If you look at `_workload_manager.py`, I'm handling InvalidHP as a special case when handling the response_func, and that is how information is being propagated to the searcher to handle InvalidHP

harness/determined/util.py
```diff
@@ -114,15 +114,19 @@ def make_metrics(num_inputs: Optional[int], batch_metrics: List[Dict[str, Any]])
...
+    Skipped if not chief.
+    """
     if isinstance(metrics, det.workload.Skipped):
         return metrics
     else:
-        return {
-            "metrics": metrics,
-            "stop_requested": stop_requested,
-        }
+        return {"metrics": metrics, "stop_requested": stop_requested, "invalid_hp": invalid_hp}
```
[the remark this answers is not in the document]
rb-determined-ai: yes, for now at least.  Maybe it will change when we have a full push architecture.

harness/determined/exec/harness.py
rb-determined-ai: absolutely.

harness/determined/estimator/_estimator_trial.py
```diff
@@ -270,19 +271,44 @@ def control_loop(self) -> None:
...
-                # Break out of the control loop so that the train process
-                # re-enters the train_and_evaluate() loop.
-                break
+                try:
+                    # Store values for the training loop.
+                    self.num_batches = wkld.num_batches
+                    self.train_response_func = response_func
+                    # Break out of the control loop so that the train process
+                    # re-enters the train_and_evaluate() loop.
+                    break
```
rb-determined-ai, line 280: Um, I don't think it's possible for these three lines of code to raise a `det.InvalidHP` exception.

This is the tricky part of our plugin-wrapping-a-plugin architecture; in this callback, we interrupt the main training loop to do non-training things, then our callback releases control back to the main loop, and some time later the main loop will re-enter this callback's control_loop.

That means to capture exceptions that occur during training, you'll have to do them somewhere above the scope of the estimator training loop; not inside the estimator training loop.

I actually think what @shiyuann suggested of using the push architecture for pushing checkpoints would massively simplify this.  And handing exceptions anywhere else in a way that lets you keep the harness in-tact enough to receive another message from the master asking for a checkpoint seems so hard that I wouldn't bother... I'd just say "if you raise this exception outside of evaluation you should not expect a checkpoint", and then after shiyuan's push api lands you can clean that up.

shiyuann: @rb-determined-ai What I have right now is the checkpoint and validation APIs, to make them work with the searcher and trial/experiment, there is still a bunch of work to do. For now, I think it's fine to just raise an exception.

rb-determined-ai: cool.  In that case we should probably let invalidHP exceptions during training in estimators *not* result in a checkpoint.  I think that would be too much work, and in the near future there will be an easy way to do it.

harness/determined/_train_context.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext):
         self.hvd_config = hvd_config  # type: horovod.HorovodContext
         self.distributed = DistributedContext(env, hvd_config)
         self._stop_requested = False
+        self._invalid_hp = False
```
rb-determined-ai, line 20: is this still used anywhere?

harness/determined/estimator/_estimator_trial.py
```diff
@@ -277,12 +278,25 @@ def control_loop(self) -> None:
...
-                response_func(
-                    det.util.wrap_metrics(
-                        self._compute_validation_metrics(),
-                        self.estimator_trial_controller.context.get_stop_requested(),
+                try:
+                    response_func(
+                        det.util.wrap_metrics(
+                            self._compute_validation_metrics(),
+                            self.estimator_trial_controller.context.get_stop_requested(),
+                            False,
```
rb-determined-ai, line 286: Could you use keyword argument format `metrics=False` for these?  Otherwise it is not as clear what `False` means to the reader of the code.

rb-determined-ai (commented):
Last comment:

We talked not supporting checkpoint-after-InvaidHP-during-training, which is fine, but I think the way this is now written is such that we won't even tell the master there was an invalid hp at all if you raise it during training.  Is that accurate?

harness/determined/_train_context.py
naren-determined: Nope, good catch.

naren-determined:
> Last comment:
> 
> We talked not supporting checkpoint-after-InvaidHP-during-training, which is fine, but I think the way this is now written is such that we won't even tell the master there was an invalid hp at all if you raise it during training. Is that accurate?

Well in `layers/_workload_manager.py` at the sections starting in Lines 146 and 203 we specifically handle how training and validation workloads respond to the master with an invalid_hp so it can then inform the searcher, right?

rb-determined-ai:
But those process responses inside `respond_func`, and if an exception is raised in training but never caught, how will any `response_func` ever get called?

naren-determined:
> But those process responses inside `respond_func`, and if an exception is raised in training but never caught, how will any `response_func` ever get called?

Oh sorry, yes. For PyTorch it will be caught during training, but for Estimators and Keras it will not be the way it's written at the moment. Eventually, once push architecture comes in this will be updated.

rb-determined-ai (approved):
ok.  I think that the way it's written now is not a good user experience, but if we are not announcing this feature until after push api lands, and we fix this issue before we do announce it, I'm fine with the current limitations.

PR #1876, opened 2021-01-27
fix: add non-scalar metric expectation to protobufs [DET-4893] [DET-4911]

hamidzr:
## Test Plan

- run the provided noop experiment and check taht the api does not fail
- add master integ test. watch it fail, add a fix, watch it pass

## Commentary (optional)

- for a more end-to-end test we need to seed the database for testing the api, mock the api response for testing the UI, or submit the test experiment and wait for it to finish before proceeding to have both. what do you suggest? 

## Checklist

- [x] remove the test commit
- [x] remove test scope limit


webui/tests/test-cluster/master.yaml
```diff
@@ -11,7 +11,7 @@ checkpoint_storage:
 
 db:
   user: postgres
-  host: localhost
```
mackrorysd, line 14: This is more curiosity than anything, but why does this have to change?

mackrorysd:
The protobuf changes themselves look good to me. I don't fully understand why the incompatibility errors manifest the way they do. I'm guessing they operate on generated code, so there are 3 distinct symptoms of changing the type?

Still wrapping my head around the test changes.

webui/tests/test-cluster/master.yaml
hamidzr: I think a short version would be less chance for DNS (or hosts entry?) to mess up and delay responding with the address.

https://determined-ai.slack.com/archives/CSKCTUKEE/p1611877804019800?thread_ts=1611864602.016200&cid=CSKCTUKEE

hamidzr:
> The protobuf changes themselves look good to me. I don't fully understand why the incompatibility errors manifest the way they do. I'm guessing they operate on generated code, so there are 3 distinct symptoms of changing the type?
> 
> Still wrapping my head around the test changes.

don't worry about the test changes they're mostly just setup for testing the change through WebUI e2e tests but I'm now limiting the test scope and just writing integration tests in Go.

hamidzr:
I took out irrelevant changes.

webui/react/src/services/decoder.ts
```diff
@@ -262,10 +263,21 @@ export const decodeExperimentList = (data: Sdk.V1Experiment[]): types.Experiment
   return data.map(decodeV1ExperimentToExperimentItem);
 };
 
+const filterNonScalarMetrics = (metrics: types.RawJson): types.RawJson | undefined => {
+  if (typeof metrics !== 'object') return undefined;
```
hkang1, line 267: **suggestion:** there is an `isObject` in the same place as `isNumber` (utils/data) that does it a bit more reliably. This is because `typeof null` will also report `object`.

hkang1 (approved):
The decoder changes look good with one minor improvement.

master/test/testutils/mocks.go
```diff
@@ -87,3 +87,35 @@ func TrialModel(eID int, opts ...TrialModelOption) *model.Trial {
 	}
 	return t
 }
+
+// StepModelOption is an option that can be applied to a step.
+type StepModelOption interface {
+	apply(*model.Step)
+}
+
+// // StepModelOptionFunc is a type that implements StepModelOption.
```
hamidzr, line 96: remove

mackrorysd (approved):

PR #1877, opened 2021-01-27
fix: trial hangs when it fails to write to the DB [DET-4905]

shiyuann:
## Description

It has been noticed that the trial doesn't deallocate resources when it fails at writing to the DB. The killing logic starts when failing at writing to the DB but then the trial waits for the container terminated message, which will never come. This PR fails the trial directly rather than handling restarting when the failure is caused by the DB.

## Test Plan

N/A

## Commentary (optional)

N/A

## Checklist


stoksc (approved):
lgtm

PR #1881, opened 2021-01-28
fix: allow NULL trials.request_id for backwards compat

stoksc:
## Description
This change makes request ID in `model.Trial` nullable, since it may not exist for older trials and is only used where it is sure to exist with new trials.

## Test Plan
- [x] hit `/api/v1/experiments/{experiment_id}/metrics-stream/trials-sample` and make sure it returns.

## Checklist


mackrorysd (approved):

stoksc:
admin merged since semantic PRs are hung

PR #1882, opened 2021-01-28
fix: don't lose logs of short-lived commands [DET-4907]

dzhu:
## Description

Very short-lived command containers can exit and be autoremoved by
Docker before the agent starts tracking their logs, causing the logs to
be lost. This changes containers that don't have their logs shipped
to Fluent Bit to be explicitly removed by the agent rather than
autoremoved so that we can be sure to get the logs.

This used to lead to a visible error but has lately been completely
silent due to a typo in error checking (also fixed here).

There's now a risk of a container being left lying around after exiting
if the agent is hard-killed. An alternative would be to add a brief
sleep to every command to mostly ensure that the agent has a chance to
start reading logs; that's maybe cleaner in a sense, but I decided it
was sillier overall.

It would be great if Docker allowed you to start tailing container logs
between the create step and the start step, but the tail just ends if
you try that. It would also be great if Docker let you add the
autoremove flag to a container after it started, but that doesn't seem
to be possible.

## Test Plan

- [x] check that the typo fix by itself causes the error to be surfaced
- [x] run short-lived commands, check that logs are not lost and containers are still removed

## Commentary (optional)

We intend to be moving everything over to Fluent Bit at some point as a more comprehensive update, but anyway this is a clear bug.

### DIFF
--- a/agent/internal/docker.go
+++ b/agent/internal/docker.go
@@ -166,6 +166,10 @@ func (d *dockerActor) pullImage(ctx *actor.Context, msg pullImage) {
 }
 
 func (d *dockerActor) runContainer(ctx *actor.Context, msg container.RunSpec) {
+	if !d.spec.RunSpec.UseFluentLogging {
+		msg.HostConfig.AutoRemove = false
+	}
+
 	response, err := d.ContainerCreate(
 		context.Background(), &msg.ContainerConfig, &msg.HostConfig, &msg.NetworkingConfig, "")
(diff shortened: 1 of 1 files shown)


stoksc (approved):
lgtm. interesting problem.

stoksc:
@dzhu it would be nice to plum this through the trial logs infra sooner than later. wonder if that could be an infra ag topic.

PR #1885, opened 2021-01-28
chore: fix type errors in IPC code

mackrorysd:
## Description

This wasn't caught in the PR, but I was able to reproduce it locally after resetting my Conda environment.

## Test Plan

`make -C harness check` passes now, whereas it doesn't in a clean Conda environment (with only the top-level requirements.txt installed) before.

### DIFF
--- a/harness/determined/ipc.py
+++ b/harness/determined/ipc.py
@@ -95,7 +95,7 @@ class ZMQBroadcastServer:
     ) -> None:
         self._num_connections = num_connections
 
-        context = zmq.Context()
+        context = zmq.Context()  # type: ignore
 
         self._pub_socket = context.socket(zmq.PUB)
         self._pull_socket = context.socket(zmq.PULL)
@@ -179,7 +179,7 @@ class ZMQBroadcastServer:
 
 class ZMQBroadcastClient:
     def __init__(self, srv_pub_url: str, srv_pull_url: str) -> None:
-        context = zmq.Context()
+        context = zmq.Context()  # type: ignore
 
         self._sub_socket = context.socket(zmq.SUB)
         # Subscriber always listens to ALL messages.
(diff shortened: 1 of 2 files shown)


mackrorysd:
Hmm, so the linter appears to be unhappy with this in pre-commits, whereas in post-commits it's unhappy with the current state: https://app.circleci.com/pipelines/github/determined-ai/determined/9321/workflows/3bbdca53-1600-4af7-959c-8454e19ac10d/jobs/258493.

rb-determined-ai:
Is it maybe a matter of which version of `zmq` is installed?

mackrorysd:
> Is it maybe a matter of which version of `zmq` is installed?

Could be. I couldn't reproduce this locally until I reset my conda environment (which had already been reset in the last week) and pip installed the top-level requirements.txt. Then I reproduced it. If I go back to before my commit and do the same, I'm on pyzmq 22.0.0, which is the same version I had while reproducing it. I wonder if there's some cache involved for pre-commits that doesn't get used later?

dzhu:
I'm beginning to suspect this is another new-resolver-in-pip-20.3 issue.

dzhu:
Oh, this probably all just happened because pyzmq _just_ released 22.0.0. https://pypi.org/project/pyzmq/#history :zany_face:

harness/setup.py
```diff
@@ -23,7 +23,7 @@
         "packaging",
         "numpy>=1.16.2",
         "psutil",
-        "pyzmq>=18.1.0",
+        "pyzmq>=22.0.0",
```
rb-determined-ai, line 26: We don't need our users to pin this so strictly?  Can we put the requirement somewhere else, that will affect CI but not users?

dzhu: Ah, makes sense. Sean had it there and I suggested the move to `setup.py`, sorry.

rb-determined-ai (approved):

PR #1887, opened 2021-01-28
feat: Documentation and CI of support for NVIDIA A100's and Google A2 instances

mackrorysd:
## Description

Adds some CI coverage for the CUDA 11 images that were recently published in connection with the upgrades to PyTorch and TensorFlow.

## Test Plan

Ran the relevant tests here: https://app.circleci.com/pipelines/github/determined-ai/determined/9339/workflows/5afe5b24-3061-4508-88f1-f31037cf1c24.

## Commentary (optional)

A couple of things to note:

I consider it very hacky, but as a cost-effective way to maximize test coverage here, I'm setting a TF2 image as the TF1 default image, since the TF1 image is THE default, and is specified in many places. There's only 1 test that this causes a problem for (test_horovod_optimizer when tf2=false).

Also I'm having to use the "rapid" channel of GKE versions to use the A100 instances, and this means the version string may be a little fragile moving forward. It's always an easy fix at least...

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1430,7 +1430,7 @@ jobs:
         default: 1
       gke-version:
         type: string
-        default: "1.18.12-gke.1200"
+        default: "1.18.12-gke.1205"
       machine-type:
         type: string
         default: "a2-highgpu-1g"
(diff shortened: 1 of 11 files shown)


mackrorysd:
Meant to submit this PR for the master branch

PR #1888, opened 2021-01-28
feat: Documentation and CI of support for NVIDIA A100's and Google A2 instances

mackrorysd:
## Description

Adds some CI coverage for the CUDA 11 images that were recently published in connection with the upgrades to PyTorch and TensorFlow.

## Test Plan

Ran the relevant tests here: https://app.circleci.com/pipelines/github/determined-ai/determined/9339/workflows/5afe5b24-3061-4508-88f1-f31037cf1c24.

## Commentary (optional)

A couple of things to note:

I consider it very hacky, but as a cost-effective way to maximize test coverage here, I'm setting a TF2 image as the TF1 default image, since the TF1 image is THE default, and is specified in many places. There's only 1 test that this causes a problem for (test_horovod_optimizer when tf2=false).

Also I'm having to use the "rapid" channel of GKE versions to use the A100 instances, and this means the version string may be a little fragile moving forward. It's always an easy fix at least...

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1430,7 +1430,7 @@ jobs:
         default: 1
       gke-version:
         type: string
-        default: "1.18.12-gke.1200"
+        default: "1.18.12-gke.1205"
       machine-type:
         type: string
         default: "a2-highgpu-1g"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,9 +4,10 @@ version: 2.1
 orbs:
   win: circleci/windows@2.3.0
   slack: circleci/slack@3.4.2
-  gke: circleci/gcp-gke@1.0.4
+  gke: circleci/gcp-gke@1.1.0
   kubernetes: circleci/kubernetes@0.11.0
   helm: circleci/helm@1.1.2
+  gcloud: circleci/gcp-cli@2.1.0
 
 executors:
   python-35:
@@ -478,6 +479,20 @@ commands:
           command: |
             gcloud container clusters delete <<parameters.cluster-id>> --quiet --region=<<parameters.region>>
 
+  terminate-gke-cluster-cuda-11:
+    parameters:
+      cluster-id:
+        type: string
+      zone:
+        type: string
+    steps:
+      # Use a run instead of `gke/delete-cluster` because circle CI orbs do not support `when`.
+      - run:
+          name: Terminate GKE Cluster
+          when: always
+          command: |
+            gcloud container clusters delete <<parameters.cluster-id>> --quiet --zone=<<parameters.zone>>
+
   setup-gke-cluster:
     parameters:
       cluster-id:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -478,6 +479,20 @@ commands:
...
             gcloud container clusters delete <<parameters.cluster-id>> --quiet --region=<<parameters.region>>
 
+  terminate-gke-cluster-cuda-11:
+    parameters:
+      cluster-id:
+        type: string
+      zone:
+        type: string
+    steps:
+      # Use a run instead of `gke/delete-cluster` because circle CI orbs do not support `when`.
```
rb-determined-ai, line 489: I'm confused by this comment... isn't there a `when` right there?

.circleci/config.yml
```diff
@@ -1329,6 +1419,86 @@ jobs:
...
+          machine-type: <<parameters.machine-type>>
+          num-machines: <<parameters.num-machines>>
+          gpu-type: <<parameters.gpu-type>>
+          gpus-per-machine: <<parameters.gpus-per-machine>>
+          node-locations: <<parameters.node-locations>>
+      - set-google-application-credentials
+      - run-e2e-tests:
+          mark: <<parameters.mark>>
+          master-host: ${MASTER_HOST}
+      - terminate-gke-cluster-cuda-11:
```
rb-determined-ai, line 1492: Is it clearer to just inline this, so that the "when: always" is right here and it's more obvious that the cluster is always terminated?

.circleci/config.yml
```diff
@@ -1329,6 +1419,86 @@ jobs:
...
+        type: string
+        default: ""
+    docker:
+      - image: determinedai/cimg-base:stable
+    parallelism: <<parameters.parallelism>>
+    environment:
+      # Using the TF2-only image here is a hack to get more test coverage since TF1 is
+      # still the default system-wide. TF1 tests that fail in this configuration should
+      # be skipped when the "CUDA" environment variable is set to 11
+      TF1_CPU_IMAGE: determinedai/environments:cuda-11.0-pytorch-1.7-tf-2.4-gpu-0.9.0
```
rb-determined-ai, line 1465: Why do we have to set the cpu images to gpu?  That seems like it should be unnecessary.

e2e_tests/tests/command/test_run.py
```diff
@@ -522,6 +523,8 @@ def test_horovod_optimizer(
     image = conf.TF1_GPU_IMAGE
     if tf2:
         image = conf.TF2_GPU_IMAGE
+    if not tf2 and "CUDA" in os.environ and os.environ["CUDA"] == "11":
```
rb-determined-ai, line 526: I thought we hardcoded the tf1 images to use the cuda 11 image anyway?

So... if `CUDA == 11` I think that the tf2 bool would not make any difference to the tests, right?  Actually it seems like a bunch of duplicate tests so I'm not opposed to skipping them, but IIUC the message this is emitting is not really why we should skip this test.

rb-determined-ai (commented):
Only the comment on the `pytest.skip()` is blocking.  I don't care too much about the others.

.circleci/config.yml
mackrorysd: IIRC I'm referring to the delete-cluster command in the gke orb. I don't think I could use that and tell it to always run, even after failure, but I could with just running the CLI command directly. Would be happy if I was wrong - I originally wrote this months ago when I was even newer to CircleCI than I am right now...

.circleci/config.yml
mackrorysd: Good point... I thought a lot about making the TF1_GPU image to the cuda-11 image, but I guess for CPU it adds no additional value that we're not already testing.

e2e_tests/tests/command/test_run.py
mackrorysd: Yeah the idea behind pointing the TF1 image variable to the TF2-only CUDA 11 image was that there's a bunch of tests that just use the default, which ends up being the TF1, so it's an easy way to get a lot more tests running on that image. For most tests, it turns out they don't care. Some are parameterized and would try run against both, which is a little wasteful (and given a lot more time I'd love to go over every test and make sure we run only what we need), but this was the only case where it turned out to matter. If it tried using TF1, it would fail. But all the TF2 runs pass.

rb-determined-ai: Right, but how would *this* test fail?  It seems like where pytest.skip() is called, we'd be secretly using TF2 anyway.

.circleci/config.yml
mackrorysd: Maybe clearer, but IMO more error prone if someone re-uses one of the terminate- commands and forgets to add it. Having it as part of the command ensures that the commands never get skipped (as long as they get re-used).

e2e_tests/tests/command/test_run.py
mackrorysd: Something was trying to do set_session: https://gist.github.com/mackrorysd/a87eda55c65dda42f810051c2636cf36

I didn't dig much deeper than that. The TF2 variant of the test was obviously running against TF2 successfully, which was my main concern.

rb-determined-ai (approved):

PR #1890, opened 2021-01-29
refactor: replace step IDs with total batch in the DB

shiyuann:
## Description

Replace the use of step id with the total batch in the DB. Any checkpoint, training metric, validation can be located with a trial id and a total batch number. They used to depend on the concept of steps. In order to hand over the control on checkpointing and validating to the user code, all the internal use of steps should be abandoned.

In the meantime, remove the constraint that all checkpoints and validations need to be associated with a step. We will change the restful APIs later to return the checkpoints and validations of all total_batches rather than just those of steps.

## Test Plan

1. Use the CI for testing.
2. Manually test running experiments of constant hp params and adaptive hp search on 
training metrics, validations, checkpoints, and fault tolerance.

## Commentary (optional)

N/A

## Checklist

### DIFF
--- a/master/internal/db/postgres.go
+++ b/master/internal/db/postgres.go
@@ -222,7 +222,7 @@ FROM (
                        (SELECT v.metrics
                         FROM validations v
                         WHERE v.trial_id = t.id AND v.state = 'COMPLETED'
-                        ORDER BY v.step_id DESC
+                        ORDER BY v.total_batches DESC
                         LIMIT 1
                        ) AS latest_validation_metrics,
                        (SELECT count(*)
@@ -239,12 +239,12 @@ FROM (
                        ) AS best_validation_metric,
                        (SELECT row_to_json(bc)
                         FROM (
-                            SELECT c.id, c.uuid, c.trial_id, c.step_id, c.state,
+                            SELECT c.id, c.uuid, c.trial_id, c.total_batches, c.state,
                                    c.start_time, c.end_time, c.resources, c.metadata,
                                    (v.metrics->'validation_metrics'
                                              ->>const.metric_name)::float8 AS validation_metric
                             FROM checkpoints c LEFT JOIN validations v
-                            ON c.trial_id = v.trial_id AND c.step_id = v.step_id, const
+                            ON c.trial_id = v.trial_id AND c.total_batches = v.total_batches, const
                             WHERE c.trial_id = t.id
                               AND c.state = 'COMPLETED'
                               AND v.state = 'COMPLETED'
@@ -363,16 +363,16 @@ FROM (
                      (SELECT row_to_json(c)
                       FROM (
                           SELECT c.end_time, c.id, c.metadata, c.resources, c.start_time, c.state,
-                                 c.step_id, c.trial_id, c.uuid
+                                 c.total_batches, c.trial_id, c.uuid
                           FROM checkpoints c
-                          WHERE c.trial_id = t.id AND c.step_id = s.id
+                          WHERE c.trial_id = t.id AND c.total_batches = s.total_batches
                       ) c) AS checkpoint,
                      (SELECT row_to_json(v)
                       FROM (
-                          SELECT v.end_time, v.id, v.metrics, v.start_time, v.state, v.step_id,
-                                 v.trial_id
+                          SELECT v.end_time, v.id, v.metrics, v.start_time, v.state, 
+                                 v.total_batches, v.trial_id
                           FROM validations v
-                          WHERE v.trial_id = t.id AND v.step_id = s.id
+                          WHERE v.trial_id = t.id AND v.total_batches = s.total_batches
                       ) v) AS validation
                      FROM steps s
                      WHERE s.trial_id = t.id
@@ -437,8 +437,8 @@ FROM (
                 -- We can't use a computed column in a WHERE clause for the same query, which we
                 -- would like to do here with the "steps" column, so do this subquery.
                 SELECT * FROM (
-                    SELECT c.id, c.trial_id, c.step_id, c.state, c.start_time, c.end_time, c.uuid,
-                           c.resources, c.metadata,
+                    SELECT c.id, c.trial_id, c.total_batches, c.state, c.start_time, c.end_time,
+                           c.uuid, c.resources, c.metadata,
                            (SELECT row_to_json(s)
                             FROM (
                                 SELECT s.end_time, s.id, s.start_time, s.state, s.trial_id,
@@ -446,13 +446,14 @@ FROM (
                                     (SELECT row_to_json(v)
                                     FROM (
                                         SELECT v.end_time, v.id, v.metrics, v.start_time,
-                                            v.state, v.step_id, v.trial_id
+                                            v.state, v.total_batches, v.trial_id
                                         FROM validations v
-                                        WHERE v.trial_id = s.trial_id AND v.step_id = s.id
+                                        WHERE v.trial_id = s.trial_id
+                                              AND v.total_batches = s.total_batches
                                     ) v
                                     ) AS validation
                                 FROM steps s
-                                WHERE s.id = c.step_id AND s.trial_id = c.trial_id
+                                WHERE s.total_batches = c.total_batches AND s.trial_id = c.trial_id
                             ) s
                            ) AS step
                     FROM checkpoints c, trials t, const
@@ -536,20 +537,20 @@ FROM (
                 (SELECT coalesce(jsonb_agg(s ORDER BY id ASC), '[]'::jsonb)
                  FROM (
                      SELECT s.end_time, s.id, s.start_time, s.state, s.trial_id, s.num_batches,
-                     s.prior_batches_processed,
+                     s.prior_batches_processed, s.total_batches,
                      (SELECT row_to_json(c)
                       FROM (
                           SELECT c.end_time, c.id, c.metadata, c.resources, c.start_time, c.state,
-                                 c.step_id, c.trial_id, c.uuid
+                                 c.total_batches, c.trial_id, c.uuid
                           FROM checkpoints c
-                          WHERE c.trial_id = t.id AND c.step_id = s.id
+                          WHERE c.trial_id = t.id AND c.total_batches = s.total_batches
                       ) c) AS checkpoint,
                      (SELECT row_to_json(v)
                       FROM (
-                          SELECT v.end_time, v.id, v.metrics, v.start_time, v.state, v.step_id,
-                                 v.trial_id
+                          SELECT v.end_time, v.id, v.metrics, v.start_time, v.state, 
+                                 v.total_batches, v.trial_id
                           FROM validations v
-                          WHERE v.trial_id = t.id AND v.step_id = s.id
+                          WHERE v.trial_id = t.id AND v.total_batches = s.total_batches
                       ) v) AS validation
                      FROM steps s
                      WHERE s.trial_id = t.id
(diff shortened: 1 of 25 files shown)


master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/google/uuid"
+	_ "github.com/jackc/pgx/v4/stdlib" // Import Postgres driver.
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// AddCheckpoint adds the checkpoint to the database and sets its ID.
+func (db *PgDB) AddCheckpoint(checkpoint *model.Checkpoint) error {
+	if !checkpoint.IsNew() {
```
stoksc, line 18: We lost some correctness constraints provided by the data model here. Should we add them back in until we're ready to part with them? (e.g. the check that a step at least exists for this and should we wait until the API is exposed? even when it is, should creating a validation without a step attached to it create a metric-less step at that many batches so that the UI still remains useful exactly as it is?)

master/internal/db/postgres_validations.go
```diff
@@ -0,0 +1,116 @@
...
+	_ "github.com/golang-migrate/migrate/source/file" // Load migrations from files.
+	_ "github.com/jackc/pgx/v4/stdlib"                // Import Postgres driver.
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// AddValidation adds the validation to the database and sets its ID.
+func (db *PgDB) AddValidation(validation *model.Validation) error {
+	if !validation.IsNew() {
```
stoksc, line 17: Same comments here as for checkpoints. Removing the constraint that there isn't a step associated with the validation fits with the rest of this PR and may now be an OK state for the backend but I'm not sure how the webui will handle. Have you looked into this?

master/internal/trial.go
```diff
@@ -1030,8 +1030,13 @@ func classifyStatus(state terminatedContainerWithState) aproto.ContainerStopped
 }
 
 func (t *trial) reset() error {
-	step := t.sequencer.RollBackSequencer()
-	if err := t.db.RollBackTrial(t.id, step); err != nil {
+	t.sequencer.RollBackSequencer()
+	w, err := t.sequencer.Workload()
```
stoksc, line 1034: maybe just open up a CurrentBatch() on the sequencer?

master/pkg/model/experiment.go
```diff
@@ -378,17 +380,56 @@ type Checkpoint struct {
...
+		TotalBatch:        totalBatch,
 		State:             ActiveState,
 		StartTime:         time.Now().UTC(),
 		Metadata:          JSONObj{},
 		DeterminedVersion: version.Version,
 	}
 }
 
+// CheckpointFromProto returns a checkpoint db model made from the protobuf model.
+func CheckpointFromProto(proto *checkpointv1.Checkpoint) *Checkpoint {
```
stoksc, line 395: Where is this used?

stoksc (commented):
Mostly lgtm, just have meta-questions:
1. I commented a couple of times on this but we removed the constraint that checkpoints and validations are aligned with training steps. Are we ready to remove this yet? I think the APIs will be ok (think it's on the new api and uses the api backed by `master/static/srv/proto_get_trial_workloads.sql` so it looks like it may be ok but I'd check with @hamidzr. I know this constraint isn't broken yet, but it opens the door for it to be accidentally broken. 
2. Will push API include pushing training steps too?
3. We should definitely time that migration, and when you land this land it in the middle of the night -  my revert today was because I landed a costly migration that got clobbered by another merge that happened right after it (it made it get killed halfway through, then the follow up failed)

hamidzr:
the WebUI isn't using step_id anymore instead it is using batch numbers on each workload to line up the checkpoint and validation workloads with the corresponding training workload but this constraint can go away with a redesign in the future.  We might need to also look at CLI to see if it would affect anything there.

master/internal/db/postgres_proto.go
```diff
@@ -13,7 +13,10 @@ import (
 // QueryProto returns the result of the query. Any placeholder parameters are replaced
 // with supplied args. Enum values must be the full name of the enum.
 func (db *PgDB) QueryProto(queryName string, v interface{}, args ...interface{}) error {
-	return db.queryRowsWithParser(db.queries.getOrLoad(queryName), protoParser, v, args...)
+	return errors.Wrapf(
```
stoksc, line 16: nice

master/internal/trial.go
```diff
@@ -1030,8 +1030,13 @@ func classifyStatus(state terminatedContainerWithState) aproto.ContainerStopped
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get workload from sequencer in reset")
+	}
+	err = t.db.RollBackTrial(t.id, w.TotalBatches())
```
stoksc, line 1038:
```suggestion
	w, err := t.sequencer.Workload()
	if err != nil {
		return errors.Wrap(err, "failed to get workload from sequencer in reset")
	}
	err = t.db.RollBackTrial(t.id, w.TotalBatchesProcessed)
```
This used to take the last step we completed, now it takes the current step we're on (just in batches instead of steps) which I think introduces a master restart bug. This means we're just to miss rolling back the current step we're on since `t.db.RollBackTrial` purges everything _after_ `w.TotalBatches()`, not including it, and the trial is going to start and run the batch at `w.TotalBatches`.

Should able to confirm this is broken/fixed by this change by running the adaptive chaos noop and looking at the logs.

master/pkg/workload/workload.go
```diff
@@ -30,6 +30,11 @@ type Workload struct {
 	TotalBatchesProcessed int  `json:"total_batches_processed"`
 }
 
+// TotalBatches returns the total batch number after the workload is processed.
```
stoksc, line 33: I think we just just remove this or call it `TotalBatchesProcessed` and rename `total_batches_processed` to `prior_batches_processed` like it should've been (it makes more sense at total in the harness which is where it originated but master side it's just confusing).

If that's too much work, maybe at least comment `Workload.TotalBatchesProcessed` that it's really prior batches processed

master/static/migrations/20210204000000_add-total-batch.down.sql
```diff
@@ -0,0 +1,32 @@
+-- Replace total_batches with step_id in the checkpoints table.
```
stoksc, line 1: Since the release party cluster is up right now, can you time this up and down on it? For context, I've noticed from us to our largest users usually results in a 5-20x increase in duration on migrations like this (relative to the size out latest master/release party).

stoksc (commented):
mostly lgtm. nearly ready to approve but i think i see a little trial reset bug.

master/internal/db/postgres_checkpoints.go
shiyuann: > Should we add them back in until we're ready to part with them?

Don't think it is necessary to keep this constraint in the DB since we only record checkpoints on specific steps. So the WebUI will work fine.

We should remove them in the push architecture. I've created a [follow-up ticket](https://determinedai.atlassian.net/browse/DET-4980) to return all the checkpoints/validations for each trial.

master/internal/db/postgres_validations.go
shiyuann: Same.

master/pkg/model/experiment.go
shiyuann: Removed this.

master/internal/db/postgres.go
```diff
@@ -1196,30 +1197,30 @@ WHERE id = :id`, setClause(toUpdate)), trial)
 }
 
 // RollBackTrial deletes from the database all steps, checkpoints, and validations for the trial
-// that correspond to steps past lastStep.
-func (db *PgDB) RollBackTrial(id int, lastStep int) error {
+// that correspond to totalBatch.
```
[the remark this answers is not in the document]
shiyuann: probably use `total_batches` for consistency.

master/pkg/workload/workload.go
shiyuann: > rename total_batches_processed to prior_batches_processed

This will break many places. I'd do that later since I will need to remove this field later.

master/static/migrations/20210204000000_add-total-batch.down.sql
shiyuann: Do you mean time this and put it in the release note?

master/internal/trial.go
shiyuann: This used to take the last step we completed, now it takes the current step we're on (just in batches instead of steps) which I think introduces a master restart bug.

> Good catch!

shiyuann:
> I commented a couple of times on this but we removed the constraint that checkpoints and validations are aligned with training steps. Are we ready to remove this yet? I think the APIs will be ok (think it's on the new api and uses the api backed by master/static/srv/proto_get_trial_workloads.sql so it looks like it may be ok but I'd check with @hamidzr. I know this constraint isn't broken yet, but it opens the door for it to be accidentally broken.

I think we are ready to remove the constraint since the only places that write checkpoints and validations to the DB are for those checkpoints/validations that are associated with steps. We roll back the trial every time we start a trial so I think it's fine. 

> Will push API include pushing training steps too?

yes. It will include pushing training metrics that are not associated with step IDs but total_batches.

> We should definitely time that migration, and when you land this land it in the middle of the night - my revert today was because I landed a costly migration that got clobbered by another merge that happened right after it (it made it get killed halfway through, then the follow up failed)

I see. So it's only for internal migration.

master/static/migrations/20210204000000_add-total-batch.down.sql
stoksc: Ah you could put it in the release note, but I just meant time it so we'd know to warn people we help upgrade.

master/internal/trial_workload_sequencer.go
```diff
@@ -366,10 +366,14 @@ func (s *trialWorkloadSequencer) snapshotState() {
...
 // RollBackSequencer rolls back the sequencer to the latest checkpoint and sets the latest
 // checkpoint back to the one we just rolled back to.
-func (s *trialWorkloadSequencer) RollBackSequencer() int {
+func (s *trialWorkloadSequencer) RollBackSequencer() (int, error) {
 	s.trialWorkloadSequencerState = *s.LatestSnapshot
 	s.LatestSnapshot = s.trialWorkloadSequencerState.deepCopy()
-	return s.CurStepID
+	wkld, err := s.Workload()
+	if err != nil {
+		return 0, err
```
stoksc, line 374: can you wrap this so if we see it we know it's here immediately.

master/internal/trial.go
```diff
@@ -1030,8 +1030,12 @@ func classifyStatus(state terminatedContainerWithState) aproto.ContainerStopped
 }
 
 func (t *trial) reset() error {
-	step := t.sequencer.RollBackSequencer()
-	if err := t.db.RollBackTrial(t.id, step); err != nil {
+	totalBatches, err := t.sequencer.RollBackSequencer()
+	if err != nil {
+		return errors.Wrap(err, "failed to get workload from sequencer in reset")
```
stoksc, line 1035: oh I see this now.. may as well wrap the other though (`RollBackSequencer`) and have this not say what the function it called was doing.

stoksc (approved):
lgtm. thanks for doing this.

master/internal/trial.go
stoksc: I guess when I was thinking about this, I was trying to make sure that called `t.sequencer.Workload()` is always valid when we'll be resetting.. I think yeah, for sure. We should never reset then have nothing to do.

PR #1892, opened 2021-01-29
refactor: paginate experiment trials [DET-4900, DET-4921, DET-4922]

hkang1:
## Description

- Upgrade experiment details page to use pagination for experiment trials endpoint.
- Remove using experiment trials data to figure out whether we allow or disallow viewing in tensorboard. (since we will not have the full list of trials to be able to determine this)

## Test Plan

Check to make sure experiment detail page trials load properly, can paginate through and sort the various columns.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
@@ -6,7 +6,7 @@ import {
   activateExperiment, archiveExperiment, cancelExperiment, killExperiment,
   openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
 } from 'services/api';
-import { ExperimentBase, RunState, TBSourceType, TrialItem } from 'types';
+import { ExperimentBase, RunState, TBSourceType } from 'types';
 import { cancellableRunStates, killableRunStates, terminalRunStates } from 'utils/types';
 import { openCommand } from 'wait';
 
@@ -27,24 +27,11 @@ interface Props {
     [key in Action]?: () => void;
   };
   onSettled: () => void; // A callback to trigger after an action is done.
-  trials: TrialItem[];
 }
 
 type ButtonLoadingStates = Record<Action, boolean>;
 
-/*
-  * We use `numSteps` or `totalBatchesProcessed` as a
-  * proxy to trials that definietly have some metric.
-  */
-const experimentWillNeverHaveData = (experiment: ExperimentBase, trials: TrialItem[]): boolean => {
-  const isTerminal = terminalRunStates.has(experiment.state);
-  const trialsWithSomeMetric = trials.filter(trial => {
-    return trial.totalBatchesProcessed > 0;
-  });
-  return isTerminal && trialsWithSomeMetric.length === 0;
-};
-
-const ExperimentActions: React.FC<Props> = ({ experiment, onClick, trials, onSettled }: Props) => {
+const ExperimentActions: React.FC<Props> = ({ experiment, onClick, onSettled }: Props) => {
   const [ btnLoadingStates, setBtnLoadingStates ] = useState<ButtonLoadingStates>({
     Activate: false,
     Archive: false,
@@ -177,7 +164,7 @@ const ExperimentActions: React.FC<Props> = ({ experiment, onClick, trials, onSet
         key="tensorboard"
         loading={btnLoadingStates.Tensorboard}
         onClick={handleCreateTensorboard}>View in TensorBoard</Button>,
-      showIf: (exp): boolean => !experimentWillNeverHaveData(exp, trials),
+      showIf: (exp): boolean => terminalRunStates.has(exp.state),
     },
     {
       button: <Button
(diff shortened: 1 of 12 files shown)


webui/react/src/types.ts
```diff
@@ -322,6 +324,11 @@ export interface ValidationMetrics {
   validationMetrics: Record<string, number>;
 }
 
+export interface TrialPagination {
```
hamidzr, line 327: suggestion: I think pagination like startEndTimes is gonna repeat in all our endpoints that return a list exactly the same. wdy about sth like this


~another thought if we define it this way maybe we don't define a new  Trials or TrialPagination type. anywhere we return a value that has pagination we define the return type as `T & Pagination`~

webui/react/src/services/utils.ts
```diff
@@ -169,5 +170,19 @@ export const consumeStream = async <T = unknown>(
   }
 };
 
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+export const validateDetApiEnum = (enumObject: unknown, value?: unknown): any => {
```
hamidzr, line 174: nit: could you please add a comment about the usage?

webui/react/src/services/types.ts
```diff
@@ -27,7 +27,7 @@ interface ApiBase {
 // Designed for use with Swagger generated api bindings.
 export interface DetApi<Input, DetOutput, Output> extends ApiBase {
   postProcess: (response: DetOutput) => Output;
-  request: (params: Input) => Promise<DetOutput>;
+  request: (params: Input, options?: unknown) => Promise<DetOutput>;
```
hamidzr, line 30: what options would go here instead of `params`

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
```diff
@@ -177,7 +164,7 @@ const ExperimentActions: React.FC<Props> = ({ experiment, onClick, trials, onSet
         key="tensorboard"
         loading={btnLoadingStates.Tensorboard}
         onClick={handleCreateTensorboard}>View in TensorBoard</Button>,
-      showIf: (exp): boolean => !experimentWillNeverHaveData(exp, trials),
+      showIf: (exp): boolean => terminalRunStates.has(exp.state),
```
hamidzr, line 167: blocking: I think we can allow the button to show even when experiment is not in a terminal state.

hamidzr: (that is the existing behavior as well)

webui/react/src/services/types.ts
hamidzr: we might want to include in as part of the params `Input & {options?: unknown}` just to make it easier to add future arguments this will make it sort of a keyword argument instead of a positional one

hamidzr: same thing  as we did with fetch option

webui/react/src/types.ts
hkang1: great idea, I just did another one (active experiments) and thought the same thing

webui/react/src/services/types.ts
hkang1: My thinking here is that there are some other http options that we can potentially pass, even though we are only using `signal` at the moment. So this is to avoid any potential namespace pollution between params and options. Also to me, it seems a bit cleaner to call out the http options separately from params, so it's obvious what are API specific vs HTTP options.

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
hkang1: ah yes, I totally did not think this part through. At this point TBs should just be allowed regardless of state.

webui/react/src/services/types.ts
hamidzr: would we be using this for signal? It looks like you added this a while back. https://github.com/determined-ai/determined/blob/204d363985a0024c651785f0d49101d9f1f4e336/webui/react/src/services/utils.ts#L105

hamidzr (approved):
LGTM having pagination here should help a lot. IMO having both the fetchOptionParams and the newly added `options` to generateDetApi can get confusing later on so it'd be good to consolidate those too here or in another PR.

hamidzr (approved):
LGTM. 

this applies to all our pagination tables: when we switch pages since the action is async we should probably put the table in loading state until the new page is loaded so that the user gets immediate feedback. probably best to leave for a separate ticket

hamidzr:
created this to track. https://determinedai.atlassian.net/browse/DET-4962

do we also want a release note?

webui/react/src/pages/ExperimentDetails/ExperimentOverview.tsx
```diff
@@ -157,6 +157,7 @@ const ExperimentOverview: React.FC<Props> = ({
         },
         { signal: canceler.signal },
       );
+      console.log('trials', experimentTrials);
```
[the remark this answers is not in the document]
hkang1: 🤦

hkang1:
> created this to track. https://determinedai.atlassian.net/browse/DET-4962
> 
Thanks!

> do we also want a release note?
>
👍

PR #1894, opened 2021-01-29
chore: support task tokens in harness for authentication [DET-4897]

shiyuann:
## Description

All the harness internal APIs will access the restful APIs with authentication. It uses a task token that is passed into the container. All task tokens' lifecycles are tied with the tasks' lifecycles.

Specifically, all the requests can have an HTTP header key-value pair with the key `Grpc-Metadata-x-task-token`. The gRPC server on the master will truncate the prefix `Grpc-` in the HTTP header and look it up in the task session list that exists in the database.

## Test Plan

Use it only internally. So tested it manually by the following command:

```
det command run --config 'resources.slots=0' 'curl -H "Grpc-Metadata-x-task-token: Bearer $DET_TASK_TOKEN" GET "$DET_MASTER/api/v1/experiments"'
```

## Commentary (optional)

N/A

## Checklist

### DIFF
--- a/common/determined_common/api/request.py
+++ b/common/determined_common/api/request.py
@@ -107,8 +107,10 @@ def maybe_upgrade_ws_scheme(master_address: str) -> str:
 
 
 def add_token_to_headers(headers: Dict[str, str]) -> Dict[str, str]:
+    task_token = authentication.Authentication.instance().get_task_token()
+    if task_token:
+        return {**headers, "Grpc-Metadata-x-task-token": "Bearer {}".format(task_token)}
     token = authentication.Authentication.instance().get_session_token()
-
     return {**headers, "Authorization": "Bearer {}".format(token)}
 
 
--- a/master/internal/checkpoint_gc.go
+++ b/master/internal/checkpoint_gc.go
@@ -3,6 +3,8 @@ package internal
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	"github.com/determined-ai/determined/master/internal/db"
 	"github.com/determined-ai/determined/master/internal/sproto"
 	"github.com/determined-ai/determined/master/pkg/actor"
@@ -19,6 +21,7 @@ type checkpointGCTask struct {
 	agentUserGroup *model.AgentUserGroup
 	taskSpec       *tasks.TaskSpec
 
+	task *sproto.AllocateRequest
 	// TODO (DET-789): Set up proper log handling for checkpoint GC.
 	logs []sproto.ContainerLog
 }
@@ -26,16 +29,23 @@ type checkpointGCTask struct {
 func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 	switch msg := ctx.Message().(type) {
 	case actor.PreStart:
-		ctx.Tell(t.rm, sproto.AllocateRequest{
+		t.task = &sproto.AllocateRequest{
+			ID:   sproto.NewTaskID(),
 			Name: fmt.Sprintf("Checkpoint GC (Experiment %d)", t.experiment.ID),
 			FittingRequirements: sproto.FittingRequirements{
 				SingleAgent: true,
 			},
 			TaskActor:      ctx.Self(),
 			NonPreemptible: true,
-		})
+		}
+		ctx.Tell(t.rm, *t.task)
 
 	case sproto.ResourcesAllocated:
+		taskToken, err := t.db.StartTaskSession(string(msg.ID))
+		if err != nil {
+			return errors.Wrap(err, "cannot start a new task session for a GC task")
+		}
+
 		config := t.experiment.Config.CheckpointStorage
 
 		checkpoints, err := t.db.ExperimentCheckpointsToGCRaw(t.experiment.ID,
@@ -49,6 +59,7 @@ func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 		for _, a := range msg.Allocations {
 			taskSpec := *t.taskSpec
 			taskSpec.AgentUserGroup = t.agentUserGroup
+			taskSpec.TaskToken = taskToken
 			taskSpec.SetInner(&tasks.GCCheckpoints{
 				ExperimentID:     t.experiment.ID,
 				ExperimentConfig: t.experiment.Config,
@@ -79,6 +90,11 @@ func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 		t.logs = append(t.logs, msg)
 
 	case actor.PostStop:
+		if t.task != nil {
+			if err := t.db.DeleteTaskSessionByTaskID(string(t.task.ID)); err != nil {
+				ctx.Log().WithError(err).Error("cannot delete task session for a GC task")
+			}
+		}
 
 	default:
 		return actor.ErrUnexpectedMessage(ctx)
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -1,0 +1,59 @@
+package db
+
+import (
+	"github.com/o1egl/paseto"
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// initTaskSessions creates a row in the task_sessions table.
+func (db *PgDB) initTaskSessions() error {
+	_, err := db.sql.Exec("DELETE FROM task_sessions")
+	return err
+}
+
+// StartTaskSession creates a row in the task_sessions table.
+func (db *PgDB) StartTaskSession(taskID string) (string, error) {
+	taskSession := &model.TaskSession{
+		TaskID: taskID,
+	}
+
+	query := "INSERT INTO task_sessions (task_id) VALUES (:task_id) RETURNING id"
+	if err := db.namedGet(&taskSession.ID, query, *taskSession); err != nil {
+		return "", err
+	}
+
+	v2 := paseto.NewV2()
+	token, err := v2.Sign(db.tokenKeys.PrivateKey, taskSession, nil)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate task authentication token")
+	}
+	return token, nil
+}
+
+// TaskSessionByToken returns a task session given an authentication token.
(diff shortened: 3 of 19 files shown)


master/pkg/tasks/task.go
```diff
@@ -228,6 +228,7 @@ func TrialEnvVars(t TaskSpec, rendezvousPorts []string, tPortOffset int) map[str
 
 	envVars := defaultEnvVars()
 	envVars["DET_EXPERIMENT_ID"] = fmt.Sprintf("%d", exp.InitialWorkload.ExperimentID)
+	envVars["DET_TASK_ID"] = t.TaskID
```
stoksc, line 231: It doesn't seem like it is a _ton_ of work to generate a token and user it here instead. Task ID is exposed in:
* The `/tasks` endpoint.
* The names of k8s pods and in the labels of pods specs and config maps created by the master. 
* The fitting function in the logs it when it can't fit a task.
I think this is probably just as bad as not having auth.

stoksc: Logs and k8s are the ones you potentially don't need Determined auth to even see.

stoksc: We could basically duplicate all the user session logic and have `TaskSessions`:
```
// UserSession corresponds to a row in the "user_sessions" DB table.
type TaskSession struct {
	ID     SessionID `db:"id" json:"id"`
	TaskId string    `db:"task_id" json:"task_id"`
	Expiry time.Time `db:"expiry" json:"expiry"`
}
```

dzhu: Hmm, yeah, using the literal task ID may be a bit too exposed. Shiyuan, what do you think?

shiyuann: Changed the PR to use task tokens that are assigned before starting task containers.

master/internal/grpc/auth.go
```diff
@@ -85,13 +105,15 @@ func streamAuthInterceptor(db *db.PgDB) grpc.StreamServerInterceptor {
...
 
-func unaryAuthInterceptor(db *db.PgDB) grpc.UnaryServerInterceptor {
+func unaryAuthInterceptor(db *db.PgDB, rm *actor.Ref) grpc.UnaryServerInterceptor {
 	return func(
 		ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler,
 	) (resp interface{}, err error) {
 		if !unauthenticatedMethods[info.FullMethod] {
-			if _, _, err := GetUser(ctx, db); err != nil {
-				return nil, err
+			if ok := ValidateTaskID(ctx, rm); !ok {
```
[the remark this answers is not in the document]
shiyuann: Changed the implementation so this is outdated

master/internal/grpc/auth.go
```diff
@@ -41,6 +44,23 @@ var (
...
+		return false
+	}
+	tokens := md[taskIDHeader]
+	if len(tokens) == 0 {
+		return false
+	}
+
+	taskID := sproto.TaskID(tokens[0])
+	taskSummary := rm.System().Ask(
+		rm, sproto.GetTaskSummary{ID: &taskID}).GetOrElse(nil)
```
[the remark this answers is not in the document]
shiyuann: Changed the implementation so this is outdated

common/determined_common/api/request.py
```diff
@@ -85,9 +85,11 @@ def maybe_upgrade_ws_scheme(master_address: str) -> str:
 
 
 def add_token_to_headers(headers: Dict[str, str]) -> Dict[str, str]:
-    token = authentication.Authentication.instance().get_session_token()
-
-    return {**headers, "Authorization": "Bearer {}".format(token)}
+    task_id = authentication.Authentication.instance().get_task_id()
+    if task_id == "":
```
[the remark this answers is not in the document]
shiyuann: Changed the implementation so this is outdated

shiyuann:
I'll add a test once you feel the approach is right. @dzhu @stoksc

common/determined_common/api/request.py
```diff
@@ -107,9 +107,11 @@ def maybe_upgrade_ws_scheme(master_address: str) -> str:
 
 
 def add_token_to_headers(headers: Dict[str, str]) -> Dict[str, str]:
-    token = authentication.Authentication.instance().get_session_token()
-
-    return {**headers, "Authorization": "Bearer {}".format(token)}
+    task_token = authentication.Authentication.instance().get_task_token()
+    if not task_token:
```
dzhu, line 111: I feel like inverting the condition would be more intuitive.

master/internal/checkpoint_gc.go
```diff
@@ -49,6 +59,7 @@ func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 		for _, a := range msg.Allocations {
 			taskSpec := *t.taskSpec
 			taskSpec.AgentUserGroup = t.agentUserGroup
+			taskSpec.TaskToken = taskToken
```
dzhu, line 62: Could this assignment be done by the resource manager (in `Start`) rather than by every task type? That feels a little less error-prone to me, though I'm not sure.

master/static/migrations/20210211000000_add-task-session.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SEQUENCE public.task_sessions_id_seq
+    AS integer
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1;
+
+CREATE TABLE public.task_sessions (
```
dzhu, line 9: Not super important, but we may want to have some way to clean up old sessions in case anything gets left around after a hard crash of the master.

master/static/migrations/20210211000000_add-task-session.up.sql
```diff
@@ -0,0 +1,12 @@
...
+      id integer NOT NULL PRIMARY KEY DEFAULT nextval('public.task_sessions_id_seq'::regclass) ,
```
dzhu, line 10:
```suggestion
      id integer NOT NULL PRIMARY KEY DEFAULT nextval('public.task_sessions_id_seq'::regclass),
```

dzhu (commented):
Whoops, got started on this before the reassignment.

master/internal/checkpoint_gc.go
shiyuann: I'd rather put shared code in a different logic component (db) rather than the shared component (resource manager). And let the business logic (tasks) use the logic components respectively. I know there are shared code between tasks but I think it can be simplified by defining a task interface rather than putting shared code in the wrong logic component.

We used to put proxy, container management, resource management, task state, and etc, into the resource manager, which is so terrible because debugging individual logic components became impossible.

master/static/migrations/20210211000000_add-task-session.up.sql
shiyuann: We did. There is a function to clean up all the task sessions when the master restarts.

stoksc:
@dzhu i'll let you and @shiyuann stick to it. the reassignment was just because i'm on a pr blocked by this one.

master/internal/db/postgres_tasks.go
```diff
@@ -0,0 +1,59 @@
+package db
+
+import (
+	"github.com/o1egl/paseto"
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// initTaskSessions creates a row in the task_sessions table.
```
stoksc, line 10: this comment needs to be changed

master/internal/db/postgres_tasks.go
```diff
@@ -0,0 +1,59 @@
...
+func (db *PgDB) initTaskSessions() error {
+	_, err := db.sql.Exec("DELETE FROM task_sessions")
```
stoksc, line 12: Since this is just run on startup, I think we're fine to use `TRUNCATE TABLE task_sessions` and it may be a bit quicker. https://www.postgresql.org/docs/9.1/sql-truncate.html

master/internal/db/postgres_tasks.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// TaskSessionByToken returns a task session given an authentication token.
+func (db *PgDB) TaskSessionByToken(token string) (*model.TaskSession, error) {
+	v2 := paseto.NewV2()
+
+	var session model.TaskSession
+	err := v2.Verify(token, db.tokenKeys.PublicKey, &session, nil)
+	if err != nil {
+		return nil, ErrNotFound
```
stoksc, line 42: maybe 403 instead?

dzhu (approved):

master/internal/grpcutil/auth.go
```diff
@@ -41,6 +42,33 @@ var (
...
+	}
+	token = strings.TrimPrefix(token, "Bearer ")
+
+	switch session, err := d.TaskSessionByToken(token); err {
+	case nil:
+		return session, nil
+	case db.ErrNotFound:
+		return nil, ErrInvalidCredentials
+	default:
+		return nil, err
```
stoksc, line 68: maybe 403 here for security reasons, and log the actual error

shiyuann:
ignore the webui failure since it has nothing to do with the WebUI

master/internal/checkpoint_gc.go
stoksc: I sort of see both sides here. A clean interface for tasks is nice and individual tasks being allowed to cleanly configure what they need will help us long term, but it is a little silly/error-prone to have the logic duplicated in every task. What if instead we had a `taskSpec.WithTaskToken` bool, as the interface, instead of tasks needing to know how to construct a task token? The only downside I see is the task spec needs a db to construct itself then.

stoksc (commented):
mostly lgtm, few q's

stoksc:
lol @shiyuann i missed the bell

master/internal/checkpoint_gc.go
stoksc: Either way, I think this is easy to change later if it becomes clear one is better than the other.

PR #1625, opened 2020-11-20
Detailed debug

rb-determined-ai:
## Description

(this is a reopening of #1403 which I apparently cannot actually reopen since I force-pushed to the branch)

Presently, setting debug: true will result in log vomit. We need a way to enable specific subsystems of debug output.

This will save us lots of time in debugging customers, because we can productize some of the one-off patches we've been writing a lot of.

Additionally, this PR changes how we do logging throughout the harness. As a library, we should never be using logging.info() because that will use the root logger, and we should be using loggers specific to the subsystem which is generating the information.

PR #1895, opened 2021-01-30
docs: add model debug doc

rb-determined-ai:
## Description

It is looking like the observability project is going to solve what I was trying to solve in https://github.com/determined-ai/determined/pull/1625, only much better.  As a result, I'm pulling out just the debug documentation as a separate PR to land now.

### DIFF
--- a/docs/how-to/model-debug.txt
+++ b/docs/how-to/model-debug.txt
@@ -7,37 +7,50 @@
 This document aims to provide useful guidelines for debugging models
 with Determined. Hopefully, it will help you become a power user of
 Determined, but if you get stuck please don't hesitate to contact us on
-TODO:ref-support-channels!
+`Slack <https://determined-community.slack.com>`_! (`signup link
+<https://join.slack.com/t/determined-community/shared_invite/zt-cnj7802v-KcVbaUrIzQOwmkmY7gP0Ew>`_)
 
 This document focuses on model debugging, not cluster debugging, so it
 is assumed that you have already successfully :ref:`installed Determined
 <install-cluster>`.
 
+Successfully running code on a Determined cluster differs from normal
+training scripts in the following ways:
+
+   -  Your code will be organized into a Trial API.
+   -  The code will run in a docker container on another machine.
+   -  Your model may be ran many times in a hyperparameter search,
+      and/or it may be distributed across multiple GPUs or machines.
+
+This guide will introduce each change incrementally as we work towards
+achieving a fully-function model in Determined.
+
 The basic steps for debugging are:
 
-   -  Model-related issues:
+**Model-related issues:**
 
-        1. Does the original code run locally?
-        2. Does each method of your Trial class work locally?
-        3. Does local test mode work?
+   1. Does the original code run locally?
+   2. Does each method of your Trial class work locally?
+   3. Does local test mode work?
 
-   -  Docker- or Cluster-related issues
+**Docker- or Cluster-related issues**
 
-        4. Does the original code run in a Notebook or Shell?
-        5. Does each method of your Trial class work in a Notebook or Shell?
-        6. Does local test mode work in a Notebook or Shell?
+   4. Does the original code run in a Notebook or Shell?
+   5. Does each method of your Trial class work in a Notebook or Shell?
+   6. Does local test mode work in a Notebook or Shell?
 
-   -  Higher-level issues
+**Higher-level issues**
 
-        7. Does cluster test mode work with ``slots_per_trial=1``?
-        8. Does a single-GPU experiment work?
-        9. Does a multi-GPU experiment work?
+   7. Does cluster test mode work with ``slots_per_trial=1``?
+   8. Does a single-GPU experiment work?
+   9. Does a multi-GPU experiment work?
 
-Model-related issues
-====================
+**********************
+ Model-related issues
+**********************
 
 1. Does the original code run locally?
---------------------------------------
+======================================
 
 This step assumes you have have ported your model from some code outside
 of Determined. If your Trial code is not based on any such code, skip to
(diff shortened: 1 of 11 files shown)


docs/how-to/model-debug.txt
```diff
@@ -0,0 +1,337 @@
...
+Step 2.
+
+Probably you already know your code works, but this is a reminder to
+double-check before proceeding.
+
+2. Does each method of your Trial class work locally?
+-----------------------------------------------------
+
+This step assumes you have a working local environment for training. If
+you normally run your code in a docker environment of some instead, skip
```
neilconway, line 53: Grammar ("some instead"?)

docs/how-to/model-debug.txt
```diff
@@ -0,0 +1,337 @@
...
+you do not, skip to Step 4.
+
+**How to test:** Simply create an experiment, passing ``--local --test``
+on the command line:
+
+.. code:: bash
+
+   det experiment create myconfig.yaml my_model_dir --local --test
+
+**How to diagnose failures:** Local test mode does very few things; it
```
neilconway, line 83: Non-blocking: I might move the explanation of what local test mode does up a bit higher, to the rationale for why you should do this debugging step.

I also feel like this discussion might be improved by a concrete example -- e.g., what kind of problem might be caught by local test mode that wouldn't be caught by the earlier debugging steps?

docs/how-to/model-debug.txt
```diff
@@ -0,0 +1,337 @@
...
+      experiment config.
+
+   -  Any ``pip``-install or ``apt``-install commands needed in the
+      interactive environment must either be built into a custom
+      docker image or they can be written into file called
+      ``startup-hook.sh`` in the root of the model definition
+      directory.  The commands in ``startup-hook.sh`` will be executed
+      automatically before training begins.
+
+      The startup-hook.sh option is quiker to try but will be slower at
```
neilconway, line 224: Typo.

docs/how-to/model-debug.txt
```diff
@@ -0,0 +1,337 @@
...
+   det experiment create myconfig.yaml my_model_dir -f
+
+**How to diagnose failures:**  If you are using the ``determined``
+library APIs correctly, then theoretically distributed training should
+"just work".  However, you should be aware of some common pitfalls:
+
+   - Determined normally controls the details of distributed training
+     for you.  If you are also trying to control those details, such as
+     calling ``tf.config.set_visible_devices()`` in a KerasTrial or
+     EstimatorTrial will very likely cause issues.
```
neilconway, line 328: Grammar ("If you are trying to " -> "Attempting to", for example).

docs/how-to/model-debug.txt
```diff
@@ -0,0 +1,337 @@
...
+
+Then create your experiment:
+
+.. code:: bash
+
+   det experiment create myconfig.yaml my_model_dir -f
+
+**How to diagnose failures:**  If you are using the ``determined``
+library APIs correctly, then theoretically distributed training should
+"just work".  However, you should be aware of some common pitfalls:
```
neilconway, line 323: Talk about networking issues? And maybe scheduling problems / hangs?

neilconway (commented):
Overall, I think this is great! I added a few minor comments.

I think the discussion might benefit from a bit more elaboration on the thought process behind these debugging steps. For example, you could start by saying "code running in Determined is different from code running in your local environment in several ways: (a) runs as part of the Trial API (b) runs in a container on a remote machine (c) might run as part of distributed training. When debugging, it helps to isolate whether any of those environmental differences are the root cause of any observed issues." And then add some details to each section, "now that we have confirmed that the code works locally, the next step is to ..."

docs/how-to/model-debug.txt
rb-determined-ai: Added a detail about not being scheduled.

Are there common networking issue that affect distributed training and not normal training?   Almost all of the networking issues I've helped debug were issues that affected even a single notebook.

re hangs: I think hangs in our harness are nearly 100% our responsibility.  I can't think of any hangs that were due to the user doing something wrong.

neilconway: Yeah -- I mostly meant hangs due to dtrain trials not being scheduled.

Re: networking, I was thinking of issues like: (a) matching network interface names (b) additional ports we use for NCCL, ssh, etc. traffic we don't use for single-GPU training (c) issues with istio or other proxies that seem to be more problematic for dtrain trials (e.g., https://determined-community.slack.com/archives/CV3MTNZ6U/p1607112511338900)

docs/how-to/model-debug.txt
rb-determined-ai: Rewrote steps 1 and 2, let me know what you think.

docs/how-to/model-debug.txt
rb-determined-ai: I think things like port configurations belong in a separate "debug/validate your determined cluster", where we would have a series of tests that new cluster admins could run to ensure that their users' code would work.  I think this document should target the ml end user

rb-determined-ai:
Danny is going to update the `rstfmt` command to support non-auto-enumerated lists so I can pass `make -C doc check`, otherwise I got rid of all TODOs in the doc and I think it's ready for another round of review.

docs/how-to/model-debug.txt
neilconway: sg.

neilconway (approved):
lgtm. Can we find a few places to add links to this doc elsewhere in the docs? e.g., maybe add it as a suggestion in the Keras/PyTorch tutorials, or add an FAQ, etc.

rb-determined-ai:
I added it in several places:
  - FAQ
  - Keras/Pytorch/Estimator API docs
  - Model definition topic guides

PR #1898, opened 2021-01-30
chore: reword resource pool ui presentation [DET-4925]

hamidzr:
## Checklist


armandmcqueen:
I'm not sure if the `/` in `Max CPU containers / agent` is going to be clear to users, otherwise I think it looks good.

hamidzr:
> I'm not sure if the `/` in `Max CPU containers / agent` is going to be clear to users, otherwise I think it looks good.

I went with `per` instead and it was a bit longer.. I can switch it back

armandmcqueen (approved):

PR #1899, opened 2021-02-01
feat: support more fields to sortBy in /api/v1/experiments/trials [DET-4219, DET-4920]

stoksc:
## Description
This change adds support to sort by validation end time, validation metric, duration and total batches processed to `/api/v1/experiments/$1/trials`. It also fixes a bug; it seems none of the sort by's worked previously (this can be verified on latest master). Also, it adds the SQL function `public.page_info` that replicates the behavior of `api.Pagination()` in Go code and fixes a regression in `/api/v1/experiments` where using offset and limit as provided didn't work as expected (e.g. GET `/api/v1/experiments` with no query params returned `[]` on a full cluster).

## Test Plan
- [x] hit all the endpoints a lot with swagger
- [ ] maybe add integrations if i get a second


master/internal/db/postgres.go
```diff
@@ -1571,6 +1571,38 @@ WHERE id = :id`, setClause(toUpdate)), validation)
 		return errors.Wrapf(err, "error updating (%v) in validation (%v, %v)",
 			strings.Join(toUpdate, ", "), trialID, stepID)
 	}
+
+	if validation.State == model.CompletedState {
```
stoksc, line 1575: This is the only part I'm not really in love with, those sort by best validation is a very common action and scanning the trials table and the validations table for each trial to order them all is probably ridiculously expensive, so somewhere like this is where we're at.

master/static/srv/proto_get_trial_ids_for_experiment.sql
```diff
@@ -0,0 +1,66 @@
...
+        t.end_time,
+        coalesce(t.end_time, now()) - t.start_time AS duration,
+        (
+            SELECT s.prior_batches_processed + s.num_batches
+            FROM steps s
+            WHERE s.trial_id = t.id
+                AND s.state = 'COMPLETED'
+            ORDER BY s.id DESC
+            LIMIT 1
+        ) AS total_batches_processed,
```
shiyuann, line 30: :) wondering if we should also make this a separate column.

master/static/srv/proto_get_trial_ids_for_experiment.sql
```diff
@@ -0,0 +1,66 @@
...
+    WHERE t.experiment_id = $1
+      AND ($2 = '' OR t.state IN (SELECT unnest(string_to_array($2, ','))::trial_state))
+), page_info AS (
+    SELECT public.page_info((SELECT COUNT(*) AS count FROM filtered_experiment_trials), $3, $4) AS page_info
+)
+SELECT
+    (SELECT coalesce(json_agg(paginated_experiment_trials), '[]'::json) FROM (
+        SELECT id FROM filtered_experiment_trials
+        ORDER BY %s
+        OFFSET (SELECT p.page_info->>'start_index' FROM page_info p)::bigint
```
shiyuann, line 63: nonblocking: Why not just use `$3` or `$4` for OFFSET and LIMIT?

master/static/srv/proto_get_trial_ids_for_experiment.sql
```diff
@@ -0,0 +1,66 @@
...
+                LIMIT 1)
+            ELSE
+                -- For trials before `public.trials.best_validation_id` was added.
+                (SELECT searcher_info.sign * (v.metrics->'validation_metrics'->>searcher_info.metric_name)::float8
+                 FROM validations v
+                 WHERE v.trial_id = t.id
+                 ORDER BY searcher_info.sign * (v.metrics->'validation_metrics'->>searcher_info.metric_name)::float8 ASC
+                 LIMIT 1)
+            END
+        ) as best_signed_search_metric,
```
shiyuann, line 45: Is it used somewhere?

master/static/srv/proto_get_trial_ids_for_experiment.sql
```diff
@@ -0,0 +1,66 @@
...
+        (
+           SELECT v.end_time
+           FROM validations v
+           WHERE v.end_time IS NOT NULL
+           ORDER BY v.id DESC
+           LIMIT 1
+        ) as validation_end_time
```
shiyuann, line 52: same question?

master/internal/api_experiment.go
```diff
@@ -121,60 +120,19 @@ func (a *apiServer) GetExperiments(
...
 		[]interface{}{orderExpr},
-		&res,
+		resp,
 		stateFilterExpr,
 		archivedExpr,
 		userFilterExpr,
 		labelFilterExpr,
 		req.Description,
 		req.Offset,
 		req.Limit,
```
shiyuann, line 134: BTW, do you mind setting the limit to be 100 here? That fix for [DET-4917](https://determinedai.atlassian.net/browse/DET-4917) also needs to get into the next release.

shiyuann (commented):
I saw you have put a file `proto/buf.image.bin` in the PR. Isn't it a generated binary file?

master/static/migrations/20210129225517_add-validation-columns.up.sql
```diff
@@ -0,0 +1,26 @@
+ALTER TABLE public.trials
+    ADD COLUMN best_validation_id int NULL;
```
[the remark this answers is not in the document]
stoksc: I'd rather leave it as the ID; there are other places that use this and need the whole (see `proto_get_trials_plus` which gets the trial's best validation and returns it).

master/static/srv/proto_get_trial_ids_for_experiment.sql
stoksc: heh. yeah, sgtm!

master/static/srv/proto_get_trial_ids_for_experiment.sql
stoksc: this and above are used in the order by

master/static/srv/proto_get_trial_ids_for_experiment.sql
stoksc: If you just use `$4` for limit then if it's zero we dont default it. latest master `GET /api/v1/experiments` with no params is returning `[]` which is a bug. Offset has some similar problems. Also this is partially to keep the same pagination behavior as before.

master/internal/api_experiment.go
stoksc: It's fixed already, but in the `public.page_info` function. It defaults to all like it used to.

stoksc:
> I saw you have put a file proto/buf.image.bin in the PR. Isn't it a generated binary file?

Yeah, initially I had a breaking change as part of this, but I reverted it (I removed sort by state originally). I'll revert that.

master/static/srv/proto_get_trial_ids_for_experiment.sql
stoksc: Really, you could default limit outside and just use them - the other reason I was avoid that was it was nice this way to return the full `apiv1.Pagination` json so we can just `QueryProtof` into the resp object.

master/internal/db/postgres.go
```diff
@@ -72,7 +72,7 @@ func (db *PgDB) Migrate(migrationURL string) error {
 	}
 
 	if err = m.Up(); err != migrate.ErrNoChange {
-		return errors.Wrap(err, "error applying migrations")
+		return errors.Wrap(err, "d")
```
shiyuann, line 75:
```suggestion
		return errors.Wrap(err, "error applying migrations")
```

stoksc: oops.. good catch.

shiyuann (approved):
LGTM!

PR #1900, opened 2021-02-01
chore: migrate webui tasks api endpoint [DET-4018]

apizzini:
## Test Plan

This has been tested by manually triggering experiments and the 4 kind of tasks and observing the 2 pages which displays tasks (dashboard and tasks).

## Checklist

### DIFF
--- a/proto/src/determined/command/v1/command.proto
+++ b/proto/src/determined/command/v1/command.proto
@@ -4,11 +4,23 @@ package determined.command.v1;
 option go_package = "github.com/determined-ai/determined/proto/pkg/commandv1";
 
 import "google/protobuf/timestamp.proto";
+import "protoc-gen-swagger/options/annotations.proto";
 
 import "determined/container/v1/container.proto";
 
 // Command is a single container running the configured command.
 message Command {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      required: [
+        "id",
+        "description",
+        "start_time",
+        "username",
+        "resource_pool"
+      ]
+    }
+  };
   // The id of the command.
   string id = 1;
   // The description of the command.
--- a/webui/react/src/contexts/AppContexts.tsx
+++ b/webui/react/src/contexts/AppContexts.tsx
@@ -3,40 +3,31 @@ import React, { useCallback, useEffect, useState } from 'react';
 import ActiveExperiments from 'contexts/ActiveExperiments';
 import Agents from 'contexts/Agents';
 import ClusterOverview from 'contexts/ClusterOverview';
-import { Commands, Notebooks, Shells, Tensorboards } from 'contexts/Commands';
+import {
+  useFetchCommands, useFetchNotebooks, useFetchShells, useFetchTensorboards,
+} from 'contexts/Commands';
 import Users from 'contexts/Users';
 import usePolling from 'hooks/usePolling';
 import useRestApi from 'hooks/useRestApi';
-import {
-  getAgents, getCommands, getExperiments, getNotebooks,
-  getShells, getTensorboards, getUsers,
-} from 'services/api';
+import { getAgents, getExperiments, getUsers } from 'services/api';
 import { EmptyParams, GetExperimentsParams } from 'services/types';
-import { Command, DetailedUser, ExperimentPagination } from 'types';
+import { DetailedUser, ExperimentPagination } from 'types';
 import { activeRunStates } from 'utils/types';
 
 const AppContexts: React.FC = () => {
   const [ canceler ] = useState(new AbortController());
   const setUsers = Users.useActionContext();
   const setAgents = Agents.useActionContext();
-  const setCommands = Commands.useActionContext();
   const setActiveExperiments = ActiveExperiments.useActionContext();
-  const setNotebooks = Notebooks.useActionContext();
-  const setShells = Shells.useActionContext();
-  const setTensorboards = Tensorboards.useActionContext();
   const setOverview = ClusterOverview.useActionContext();
   const [ usersResponse, triggerUsersRequest ] =
     useRestApi<EmptyParams, DetailedUser[]>(getUsers, {});
-  const [ commandsResponse, triggerCommandsRequest ] =
-    useRestApi<EmptyParams, Command[]>(getCommands, {});
-  const [ notebooksResponse, triggerNotebooksRequest ] =
-    useRestApi<EmptyParams, Command[]>(getNotebooks, {});
-  const [ shellsResponse, triggerShellsRequest ] =
-    useRestApi<EmptyParams, Command[]>(getShells, {});
-  const [ tensorboardsResponse, triggerTensorboardsRequest ] =
-    useRestApi<EmptyParams, Command[]>(getTensorboards, {});
   const [ activeExperimentsResponse, triggerActiveExperimentsRequest ] =
     useRestApi<GetExperimentsParams, ExperimentPagination>(getExperiments, {});
+
+  const fetchActiveExperiments = useCallback((): void => {
+    triggerActiveExperimentsRequest({ states: activeRunStates });
+  }, [ triggerActiveExperimentsRequest ]);
 
   const fetchAgents = useCallback(async (): Promise<void> => {
     try {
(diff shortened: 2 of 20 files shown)


proto/src/determined/command/v1/command.proto
```diff
@@ -4,11 +4,24 @@ package determined.command.v1;
...
 
 // Command is a single container running the configured command.
 message Command {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      required: [
+        "id",
+        "description",
+        "start_time",
+        "container",
```
hamidzr, line 19: container can be missing.

hamidzr: same for other 'container' attributes in the other commands

webui/react/src/services/decoder.ts
```diff
@@ -117,37 +117,81 @@ export const jsonToGenericCommand = (data: unknown, type: types.CommandType): ty
   return command as types.Command;
 };
 
-const jsonToGenericCommands = (data: unknown, type: types.CommandType): types.Command[] => {
-  const io = ioTypes.decode<ioTypes.ioTypeGenericCommands>(ioTypes.ioGenericCommands, data);
-  return Object.keys(io).map(genericCommandId => {
-    return jsonToGenericCommand(io[genericCommandId], type);
-  });
-};
+const mapV1ContainerStateToCommandState =
```
hamidzr, line 120: should this use `ResourceState` type?

webui/react/src/services/apiConfig.ts
```diff
@@ -4,21 +4,31 @@ import queryString from 'query-string';
...
-  jsonToAgents, jsonToCommands, jsonToDeterminedInfo,
-  jsonToLogin, jsonToLogs, jsonToNotebook, jsonToNotebooks, jsonToShells, jsonToTaskLogs,
-  jsonToTensorboard, jsonToTensorboards, jsonToUsers,
-} from 'services/decoder';
 import * as decoder from 'services/decoder';
 import {
-  CommandIdParams, CreateExperimentParams, CreateNotebookParams, CreateTensorboardParams, DetApi,
-  EmptyParams, ExperimentDetailsParams, ExperimentIdParams, GetExperimentsParams, GetTrialsParams,
-  HttpApi, LoginResponse, LogsParams, PatchExperimentParams, SingleEntityParams, TaskLogsParams,
+  CommandIdParams,
```
hamidzr, line 9: I like the properties of importing like this but it ends up being too many lines. IMO what I think works better would be cleaner (but less efficient probably) to import everytype as `TP` or something and use that. 

suggestion here we probably should stick to the format we use everywhere else to be consistent

webui/react/src/services/api.ts
```diff
@@ -2,9 +2,23 @@ import resourcePools from 'fixtures/responses/cluster/resource-pools.json';
 import * as Api from 'services/api-ts-sdk';
 import * as Config from 'services/apiConfig';
 import {
-  ApiSorter, CommandIdParams, CreateExperimentParams, CreateNotebookParams, CreateTensorboardParams,
-  EmptyParams, ExperimentDetailsParams, ExperimentIdParams, GetExperimentsParams, GetTrialsParams,
-  LoginResponse, LogsParams, PatchExperimentParams, SingleEntityParams, TaskLogsParams,
+  ApiSorter,
```
hamidzr, line 5: ditto

webui/react/src/services/api.ts
```diff
@@ -186,10 +211,10 @@ export const openOrCreateTensorboard = async (
   params: CreateTensorboardParams,
 ): Promise<Command> => {
   const tensorboards = await getTensorboards({});
-  const match = tensorboards.find(tensorboard =>
-    !terminalCommandStates.has(tensorboard.state)
-    && tsbMatchesSource(tensorboard, params));
-  if (match) return match;
+  // const match = tensorboards.find(tensorboard =>
```
hamidzr, line 214: todo

webui/react/src/contexts/Commands.tsx
```diff
@@ -1,30 +1,104 @@
...
+  return useCallback(async (): Promise<void> => {
+    const commandsResponse = await getCommands({ signal: canceler.signal });
+    setCommands({
+      type: Commands.ActionType.Set,
+      value: {
+        data: commandsResponse,
+        errorCount: 0,
+        hasLoaded: true,
+        isLoading: false,
+      },
```
hamidzr, line 33: question: are we using all the state this is providing anywhere? if not I'd be for simplifying it. or maybe not make this change if we don't need to? 

same for other command types below.

proto/src/determined/command/v1/command.proto
apizzini: That means status might be unavailable since it's a child of container.

Is it ok to default the state to PENDING if unavailable?

webui/react/src/services/decoder.ts
apizzini: CommandTask state is defined as a CommandState. Should I change that?

webui/react/src/contexts/Commands.tsx
apizzini: Yes, this data structure is still used in:
- webui/react/src/pages/TaskList.tsx
- webui/react/src/pages/Dashboard.tsx

webui/react/src/services/api.ts
apizzini: should be good now.

hamidzr:
@apizzini could you please add your test plan, eg how you checked the changes manually, to the pr description?

webui/react/src/services/decoder.ts
hamidzr: you're right. CommandState works well here. https://github.com/determined-ai/determined/blob/master/master/internal/command/summary.go

ResourceState is container state + 'Pending" which is a special case that only happens when launching new commands. I think we should do below is that if command.container is missing set the state to `CommandState.PENDING` . would this cover number 1 here https://github.com/determined-ai/determined/pull/1900#issuecomment-771903472 ?

apizzini:
> @apizzini could you please add your test plan, eg how you checked the changes manually, to the pr description?

Sure. 2 questions, thanks in advance:
1. I need to fix the state issue (protobuf) because I'm getting an errors sporadically when creating a tensorboard (due to state). Waiting on your feedback on this.
2. I'm unsure about how to create shells and commands tasks, probably via terminal?

hamidzr:
https://github.com/determined-ai/determined/pull/1900#issuecomment-771903472 👍🏼  working on wrapping up the review that should cover 1.

about 2. yes through cli : `det shell start` and `det command run 'my command'`

proto/src/determined/command/v1/command.proto
hamidzr: yep. that's that it should be. I added more details in a comment below

webui/react/src/contexts/AppContexts.tsx
```diff
@@ -54,49 +45,32 @@ const AppContexts: React.FC = () => {
...
+  const fetchTensorboards = useFetchTensorboards(canceler);
 
   const fetchUsers = useCallback((): void => {
     triggerUsersRequest({ url: '/users' });
   }, [ triggerUsersRequest ]);
 
+  usePolling(fetchActiveExperiments);
   usePolling(fetchAgents);
-  usePolling(fetchAll);
+  usePolling(fetchCommands);
```
hamidzr, line 59: wdy think about keeping these under a single fetchTasks, or similar call so there is no need for extra usePolling calls

webui/react/src/contexts/Commands.tsx
hamidzr: yeah, let's not make the pr bigger if we don't need to.
I might be missing something: where is `isLoading` getting set to true?

apizzini: `isLoading` and `errorCount` are (now) never used. What's used is hasLoaded.

webui/react/src/services/decoder.ts
apizzini: Yes, pushing a fix soon.

proto/src/determined/command/v1/command.proto
apizzini: Thanks, can you please check the changes to:
- proto/src/determined/api/v1/tensorboard.proto
- proto/src/determined/container/v1/container.proto

webui/react/src/contexts/AppContexts.tsx
apizzini: I think that smaller/dedicated functions are better for re-usability in the future.

hamidzr: there might be a miscommunication here 🤔  agree having separate fetch functions is good I'm not saying we shouldn't keep those. I'm thinking about passing them onto usePolling once (like how it was). IIRC each react hook we added onto a component added an additional re-render to the page. @hkang1 would know better than me here.

hamidzr (approved):
LGTM

webui/react/src/contexts/AppContexts.tsx
hkang1: agree with @hamidzr here. Though the individual fetches are not related to each other directly, I think there are a couple of benefits for keeping them together. The re-render is reduced with fewer `usePolling` hooks. IMO, it's generally a good thing to just have less `setTimeouts` in general, as it can complicate the residual cleanup step. Speaking of that, the `usePolling` returns a `stopPolling` function, which can be called when the component / page unmounts. Something like this:
```
  const stopPolling = usePolling(fetchX);

  useEffect(() => {
    return () => {
      stopPolling();
      canceler.abort();
    };
  }, [ canceler, stopPolling ]);
```

webui/react/src/contexts/Commands.tsx
apizzini: Since this has to wait to land I can get this right now?

Can simply change TS interface from `RestApiState` to `ApiState` and this will cause those object to have just `isLoading` (to use instead of `hasLoaded`) and `data`.

What do you think?

webui/react/src/contexts/AppContexts.tsx
apizzini: Are you sure `usePolling` is directly causing re-renders? I might be wrong but I think re-renders are caused by (eg) `setCommands`, which updates the state in the context, and causes a re-render where the state is listened.

Added the fix btw, because surely having less timeouts is better.

webui/react/src/contexts/Commands.tsx
hamidzr: hmm IIRC we were talking about cleaning up the `ApiState` and `RestApiState` types and unifying them. which would also get easier once all apis are migrated to the new API. I'd suggest just leaving as is and keeping the changes smaller here so we also get a chance to talk about it with the team

hamidzr:
great work 🚀

PR #1905, opened 2021-02-02
ci: limit concurrent deploys from master

stoksc:
## Commentary (optional)
Maybe we can turn on 3rd party orbs for this - it's the most recent recommended solution to this https://discuss.circleci.com/t/sequentially-running-workflows/23288/2.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1259,8 +1259,8 @@ jobs:
     steps:
       - queue/until_front_of_line:
           only-on-branch: master
-          # Basically wait.. forever - we would prefer not to fail
-          # deploys and we'll likely never be this backend up.
+          # Basically wait forever -- we would prefer not to fail deploys and
+          # we'll likely never be this backed up.
           time: '10000'
       - checkout
       - set-slack-user-id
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1256,6 +1257,11 @@ jobs:
     docker:
       - image: determinedai/cimg-base:stable
     steps:
+      - queue/until_front_of_line:
+          only-on-branch: master
+          # Basically wait.. forever - we would prefer not to fail
+          # deploys and we'll likely never be this backend up.
```
dzhu, line 1263:
```suggestion
          # Basically wait forever -- we would prefer not to fail deploys and
          # we'll likely never be this backed up.
```

dzhu (approved):
Cool! I was pondering how we might do this ourselves; good thing a solution exists already. Maybe wait until after the release to land in case something goes awry.

Also, I think you may need to do something with setting up an API key before this will work, according to the [website](https://eddiewebb.github.io/circleci-queue/)?

stoksc:
> Also, I think you may need to do something with setting up an API key before this will work, according to the website

Yeah I think we may need to do _something_.

stoksc:
OK @dzhu, I added the API key with the default name to the Determined project so I think this should "just work".

stoksc:
I'll land once release it cut and done.

PR #1907, opened 2021-02-02
feat: support more types of CPU instances on AWS [DET-4939]

armandmcqueen:
## Description

Increase the list of acceptable agent types to include most general purpose instance types - t2, t3, c4, c5, m4, m5 and many variants. AMD and ARM variants are not included because we should do testing before being confident that we support them fully. 

## Checklist


master/internal/provisioner/aws_config.go
```diff
@@ -215,3 +195,106 @@ func (t ec2InstanceType) Validate() []error {
...
+	"p2.16xlarge":   16,
+	"p3.2xlarge":    1,
+	"p3.8xlarge":    4,
+	"p3.16xlarge":   8,
+	"p3dn.24xlarge": 8,
+	"t2.medium":     0,
+	"t2.large":      0,
+	"t2.xlarge":     0,
+	"t2.2xlarge":    0,
+	"t3.nano":       0,
```
shiyuann, line 222: wondering if this instance type actually is able to run docker.

shiyuann (approved):

master/internal/provisioner/aws_config.go
armandmcqueen: It is

PR #1914, opened 2021-02-03
feat: Support PyTorch-native Automatic Mixed Precision [DET-4753]

mackrorysd:
## Description

Adds tests, documentation, and a couple of new APIs (all optional - no changes required if you're not using AMP) to support native PyTorch's native AMP API.

There's an "automatic" option (though personally I'm not committed to that term - the 'A' in 'AMP' already stands for automatic) where the scaler is wrapped (currently the wrap is a no-op except for setting some properties), the forward pass is wrapped in the autocast context, and everything else is done for the user.

There's also a manual option where the only difference from the vanilla PyTorch API (e.g. https://pytorch.org/tutorials/recipes/recipes/amp_recipe.html) is that the scaler is passed to our step_optimizer function, whereas normally you would call scaler.step(optimizer) directly. Strictly speaking, the wrap is not required for this option, but I'm encouraging that to be future-proof in case things change.

## Test Plan

Also manually ran versions of fastercrnn_coco_pytorch and bert_squad_pytorch with the changes applied. Checked for any errors, and also sanity-checked performance. Running the fasterrcnn_coco_pytorch example with a batch size of 8 on NVIDIA A100's:

Times to train 100 batches:
No AMP: 1:15.25, 1:14.42
Manual mode: 1:07.09, 1:06.08
Auto mode: 1:06.15, 1:05.10

Time to compute validation metrics:
No AMP: 0:01.85
Manual mode: 0:01.56
Auto mode: 0:01.58

### DIFF
--- a/e2e_tests/tests/config.py
+++ b/e2e_tests/tests/config.py
@@ -105,7 +105,7 @@ def set_aggregation_frequency(config: Dict[Any, Any], aggregation_frequency: int
     return config
 
 
-def set_amp_level(config: Dict[Any, Any], amp_level: str) -> Dict[Any, Any]:
+def set_apex_level(config: Dict[Any, Any], amp_level: str) -> Dict[Any, Any]:
     config = config.copy()
     config.setdefault("optimizations", {})
     config["optimizations"]["mixed_precision"] = amp_level
--- a/e2e_tests/tests/config.py
+++ b/e2e_tests/tests/config.py
@@ -105,13 +105,6 @@ def set_aggregation_frequency(config: Dict[Any, Any], aggregation_frequency: int
     return config
 
 
-def set_apex_level(config: Dict[Any, Any], amp_level: str) -> Dict[Any, Any]:
-    config = config.copy()
-    config.setdefault("optimizations", {})
-    config["optimizations"]["mixed_precision"] = amp_level
-    return config
-
-
 def set_tensor_auto_tuning(config: Dict[Any, Any], auto_tune: bool) -> Dict[Any, Any]:
     config = config.copy()
     config.setdefault("optimizations", {})
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -9,7 +9,7 @@ The methods `train_batch` and `evaluate_batch` are modified to use an autocast
 context during the forward pass.
 """
 
-from typing import Any, Dict, Sequence, Tuple, Union, cast
+from typing import Dict, Sequence, Union
 
 import torch
 from model_def import MNistTrial
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -9,11 +9,10 @@ The methods `train_batch` and `evaluate_batch` are modified to use an autocast
 context during the forward pass.
 """
 
-from model_def import MNistTrial
-
 from typing import Any, Dict, Sequence, Tuple, Union, cast
 
 import torch
+from model_def import MNistTrial
 
 from determined.pytorch import PyTorchTrialContext
 
@@ -22,5 +21,5 @@ TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
 
 class MNistAutoAMPTrial(MNistTrial):
     def __init__(self, context: PyTorchTrialContext) -> None:
-        self.scaler = context.use_amp()
+        self.scaler = context.experimental.use_amp()
         super().__init__(context)
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -22,33 +22,5 @@ TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
 
 class MNistAutoAMPTrial(MNistTrial):
     def __init__(self, context: PyTorchTrialContext) -> None:
+        self.scaler = context.use_amp()
         super().__init__(context)
-        self.scaler = self.context.use_amp()
-
-    def train_batch(
-        self, batch: TorchData, epoch_idx: int, batch_idx: int
-    ) -> Dict[str, torch.Tensor]:
-        batch = cast(Tuple[torch.Tensor, torch.Tensor], batch)
-        data, labels = batch
-
-        with autocast():
-            output = self.model(data)
-            loss = torch.nn.functional.nll_loss(output, labels)
-
-        self.context.backward(loss)
-        self.context.step_optimizer(self.optimizer)
-
-        return {"loss": loss}
-
-    def evaluate_batch(self, batch: TorchData) -> Dict[str, Any]:
-        batch = cast(Tuple[torch.Tensor, torch.Tensor], batch)
-        data, labels = batch
-
-        with autocast():
-            output = self.model(data)
-            validation_loss = torch.nn.functional.nll_loss(output, labels).item()
-
-        pred = output.argmax(dim=1, keepdim=True)
-        accuracy = pred.eq(labels.view_as(pred)).sum().item() / len(data)
-
-        return {"validation_loss": validation_loss, "accuracy": accuracy}
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -9,12 +9,11 @@ The methods `train_batch` and `evaluate_batch` are modified to use an autocast
 context during the forward pass.
 """
 
-from model_def import MNistTrial
-
 from typing import Any, Dict, Sequence, Tuple, Union, cast
 
 import torch
-from torch.cuda.amp import autocast, GradScaler
+from model_def import MNistTrial
+from torch.cuda.amp import autocast
 
 from determined.pytorch import PyTorchTrialContext
 
@@ -24,7 +23,7 @@ TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
 class MNistAutoAMPTrial(MNistTrial):
     def __init__(self, context: PyTorchTrialContext) -> None:
         super().__init__(context)
-        self.scaler = self.context.wrap_scaler(GradScaler(), automatic=True)
+        self.scaler = self.context.use_amp()
 
     def train_batch(
         self, batch: TorchData, epoch_idx: int, batch_idx: int
@@ -33,8 +32,8 @@ class MNistAutoAMPTrial(MNistTrial):
         data, labels = batch
 
         with autocast():
-          output = self.model(data)
-          loss = torch.nn.functional.nll_loss(output, labels)
+            output = self.model(data)
+            loss = torch.nn.functional.nll_loss(output, labels)
 
         self.context.backward(loss)
         self.context.step_optimizer(self.optimizer)
@@ -46,8 +45,8 @@ class MNistAutoAMPTrial(MNistTrial):
         data, labels = batch
 
         with autocast():
-          output = self.model(data)
-          validation_loss = torch.nn.functional.nll_loss(output, labels).item()
+            output = self.model(data)
+            validation_loss = torch.nn.functional.nll_loss(output, labels).item()
 
         pred = output.argmax(dim=1, keepdim=True)
         accuracy = pred.eq(labels.view_as(pred)).sum().item() / len(data)
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -9,11 +9,11 @@ The methods `train_batch` and `evaluate_batch` are modified to use an autocast
 context during the forward pass.
 """
 
+from model_def import MNistTrial
+
 from typing import Any, Dict, Sequence, Tuple, Union, cast
 
 import torch
-from model_def import MNistTrial
-from torch.cuda.amp import autocast
 
 from determined.pytorch import PyTorchTrialContext
 
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -1,0 +1,55 @@
+"""
+This example demonstrates how to modify a model to use PyTorch's native AMP
+(automatic mixed precision) support in Determined.
+
+In the `__init__` method, amp_init() is called (and this accepts parameters to
+tune the GradScaler).
+
+The methods `train_batch` and `evaluate_batch` are modified to use an autocast
+context during the forward pass.
+"""
+
+from model_def import MNistTrial
+
+from typing import Any, Dict, Sequence, Tuple, Union, cast
+
+import torch
+from torch.cuda.amp import autocast, GradScaler
+
+from determined.pytorch import PyTorchTrialContext
+
+TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
+
+
+class MNistAutoAMPTrial(MNistTrial):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        super().__init__(context)
+        self.scaler = self.context.wrap_scaler(GradScaler(), automatic=True)
+
+    def train_batch(
+        self, batch: TorchData, epoch_idx: int, batch_idx: int
+    ) -> Dict[str, torch.Tensor]:
+        batch = cast(Tuple[torch.Tensor, torch.Tensor], batch)
+        data, labels = batch
+
+        with autocast():
+          output = self.model(data)
+          loss = torch.nn.functional.nll_loss(output, labels)
+
+        self.context.backward(loss)
+        self.context.step_optimizer(self.optimizer)
+
+        return {"loss": loss}
+
+    def evaluate_batch(self, batch: TorchData) -> Dict[str, Any]:
+        batch = cast(Tuple[torch.Tensor, torch.Tensor], batch)
+        data, labels = batch
+
+        with autocast():
+          output = self.model(data)
+          validation_loss = torch.nn.functional.nll_loss(output, labels).item()
+
+        pred = output.argmax(dim=1, keepdim=True)
+        accuracy = pred.eq(labels.view_as(pred)).sum().item() / len(data)
+
+        return {"validation_loss": validation_loss, "accuracy": accuracy}
--- a/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
@@ -21,5 +21,5 @@ TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
 
 class MNistAutoAMPTrial(MNistTrial):
     def __init__(self, context: PyTorchTrialContext) -> None:
-        self.scaler = context.experimental.use_amp()
+        context.experimental.use_amp()
         super().__init__(context)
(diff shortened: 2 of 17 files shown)


e2e_tests/tests/config.py
```diff
@@ -105,7 +105,7 @@ def set_aggregation_frequency(config: Dict[Any, Any], aggregation_frequency: int
     return config
 
 
-def set_amp_level(config: Dict[Any, Any], amp_level: str) -> Dict[Any, Any]:
+def set_apex_level(config: Dict[Any, Any], amp_level: str) -> Dict[Any, Any]:
```
rb-determined-ai, line 108: Can you just remove this function?  Since we removed the pre-pytorch-flexible-primitives API, this config setting is now totally unused

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -492,13 +543,31 @@ def clip_grads(params):
...
+                if self.hvd_config.use:
+                    # TODO can we do a type assertion for horovod.torch._DistributedOptimizer
+                    with optimizer.skip_synchronize():  # type: ignore
+                        scaler.step(optimizer)
+                else:
+                    scaler.step(optimizer)
             else:
-                optimizer.step()
+                if self.hvd_config.use:
+                    # TODO can we do a type assertion for horovod.torch._DistributedOptimizer
```
rb-determined-ai, line 563: Answer: definitely not because we do not have a hard dependency on horovod.

Also, it is our responsibility to wrap the optimizer in the first place, so we should be capable of tracking that information.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -440,6 +487,10 @@ def step_optimizer(
         optimizer: torch.optim.Optimizer,
         clip_grads: Optional[Callable[[Iterator], None]] = None,
         auto_zero_grads: bool = True,
+        scaler: Optional[Any] = None,
```
rb-determined-ai, line 490: this should be added to the docstring

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -440,6 +487,10 @@ def step_optimizer(
...
+        # Should be torch.cuda.amp.GradScaler, but:
+        #   * other implementations might be possible
+        #   * requiring this type forces upgrades to PyTorch 1.6+
```
rb-determined-ai, line 493: Other implementations would probably subclass GradScaler, I assume?  But I think leaving it as Any for pytorch versioning reasons is fine.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -492,13 +543,31 @@ def clip_grads(params):
...
-                optimizer.step()
+                if self.hvd_config.use:
+                    # TODO can we do a type assertion for horovod.torch._DistributedOptimizer
+                    with optimizer.skip_synchronize():  # type: ignore
+                        optimizer.step()
+                else:
+                    optimizer.step()
+
+            if self._scaler and self._auto_amp:
+                self._scaler.update()
```
rb-determined-ai, line 570: Question: what happens if a user has only one scaler but many optimizers and losses?

I think we are doing it wrong to update the scaler every optimizer step often.  See https://pytorch.org/docs/master/notes/amp_examples.html#working-with-multiple-models-losses-and-optimizers

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -477,12 +528,12 @@ def clip_grads(params):
         if self._should_communicate_and_update():
             # Communication needs to be synchronized so that is completed
             # before we apply gradient clipping and `step()`.
-            if self.hvd_config.use and not self._use_amp:
+            if self.hvd_config.use and not self._use_apex and not self._scaler:
                 optimizer.synchronize()  # type: ignore
```
rb-determined-ai, line 532: wait, does optimizer.synchronize ever get called when self._scaler is called()?

rb-determined-ai (commented):
I'm not going to lie, every time I look at our pytorch dtrain + amp code, I get confused.  I'm pretty sure there are some bugs in here.

What if, as part of this PR, we added some developer documentation (In a comment probably, nothing fancy) that explained the allowable ordering of calls against the optimizer, and why they have to happen in that order?

Aaron used to be the human version of that document.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Yeah there's the versioning issue, but also I'm not aware of any other implementations of GradScaler at all - for all I know the first one would simply implement the same interface, and I don't see anything in PyTorch that would force it to be subclassed.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: I had forgotten about this. Makes sense (avoiding the hard dependency) - but I had to add a lot of `# type: ignore`s recently and I didn't want to use them where being more specific was possible.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Since we won't really know which one is the last one, IMO if you have multiple optimizers or losses then you need to use the manual API. I can add a comment to make that explicit. Thanks for the link, though - I had looked for something along those lines when I started this and hadn't seen that spelled out.

mackrorysd: And actually I can make sure we throw an error if you try to use the automatic API and multiple optimizers get wrapped.

rb-determined-ai: Wouldn't we just call update on the scaler once after train_batch finished to support that case?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -385,7 +429,7 @@ def backward(
                 be constructed, allowing to compute higher order derivative
                 products. Defaults to ``False``.
         """
-        if self._use_amp:
+        if self._use_apex:
```
shiyuann, line 432: q: Are we still supporting `apex` even with the native mixed precision? If we got rid of that, we might not even need `step_optimizer` and `backwards` any more.

mackrorysd: Let's discuss. I don't think we can git rid of apex until we fully support AMP, and I think we should have a transition period to more fully validate the new approach before dropping the old one. But if we can really drop those functions, that would be great. I'll check, but I thought they included a bunch of Horovod logic for DTrain too.

e2e_tests/tests/config.py
mackrorysd: I'm pretty sure at least 1 customer still has it in use. I think we should generally deprecate APEX anyway, but let's leave a window of compatibility until this has been used successfully in the wild and then deprecate / remove all of APEX.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Hmm... I actually don't remember now why optimizer.synchronize() is not called in the case of APEX and if that applies to AMP. We use skip_synchronize() down below when actually stepping the Horovod optimizer but that is done regardless of AMP.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: So one factor here is that in a straightforward case, users can enable AMP on existing models by simply calling `wrap_caller(automatic=True)`, and wrap their forward pass with `with autocast():`. If we don't have our own backward function anymore, then they also need to scale their own losses, and little by little the "automatic" option is not really any simpler.

That's not a deal breaker, but if we would like to get rid of automatic, IMO we should get rid of the automatic option right now.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: So it looks like we don't call it for APEX, because if APEX is enabled we actually make the call in backward(). This is apparently to unsure that we synchronize before unscaling.

However, the only reason I know of to unscale is right before you clip gradients, and we optionally do that for the user before calling their clip_grads function, or we document that they must call unscale at the beginning of that function. Both cases occur after this optimizer synchronize().

If the user would (for some reason unknown to me) need to unscale for some reason before calling step_optimizer, we would still need to do the same thing, actually. There's a comment that migrating to native PyTorch AMP addresses this but it doesn't explain further.

For now, I'm removing this change so that optimizer.synchronize() is called at least once for PyTorch. But I need to look into if there's another reason to call unscale() earlier.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: ... yeah that would be the smart way to do it, huh. Done.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -218,6 +222,43 @@ def to_device(self, data: pytorch._Data) -> pytorch.TorchData:
...
+        """
+        Prepares to use automatic mixed precision through PyTorch’s native AMP API.
+
+        .. warning::
+            When enabling the "automatic" option, only calling backwards() once is currently
+            supported.
+
+        Arguments:
+            models (``torch.cuda.amp.GradScaler``):  Scaler to wrap and track.
+            automatic (``bool``, optional, default=False):  If True, handles all operations for the
```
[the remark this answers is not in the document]
mackrorysd: That's a good point - one more thing we can hide for the simple case.

examples/tutorials/mnist_pytorch/model_def.py
```diff
@@ -11,15 +11,13 @@
 
 from typing import Any, Dict, Sequence, Tuple, Union, cast
 
+import data
```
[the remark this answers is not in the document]
mackrorysd: Don't have a reason to have changed that manually - I'll revert it unless it's `make fmt` that's doing this.

e2e_tests/tests/config.py
mackrorysd: Ooh okay - someone told me that they really liked that Apex was enabled with config only and no code changes, so I assumed that's what they were actually doing.

examples/tutorials/mnist_pytorch/model_def.py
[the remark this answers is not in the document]
mackrorysd: Huh - `make fmt` seems to be doing this, but only sometimes. It's very weird... I may back out most of this commit and just use Circle CI to see what it complains about.

e2e_tests/tests/config.py
mackrorysd: Oh I didn't realize this was in tests, too. So the tests have just not been exercising AMP. I'll actually add an APEX test alongside my auto / manual tests for now, with the expectation that we deprecate it as soon as native AMP is definitely ready for prime time.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -85,6 +89,16 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
         model_id = len(self.models)
         self._main_model.__setattr__(f"model_{model_id}", model)
 
+        if self._auto_amp:
+            # Ensure the forward pass is wrapped in an autocast context:
+            model.forward = torch.cuda.amp.autocast()(model.forward)
+
+            # Eliminate any need for the loss functions to be in that context:
+            def hook(module, input, output):
+                return output.float32() if output.dtype == torch.float16 else output
```
mackrorysd, line 98: Actually this check if it's a Tensor, and only then type and change the dtype. BERT returns what is essentially a dictionary - not sure if we then need to worry about ITS internal data types?

e2e_tests/tests/fixtures/pytorch_amp/auto_amp_model_def.py
```diff
@@ -0,0 +1,26 @@
...
+import torch
+
+from determined.pytorch import PyTorchTrialContext
+
+TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
+
+
+class MNistAutoAMPTrial(MNistTrial):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        self.scaler = context.use_amp()
```
rb-determined-ai, line 25: context.experimental.amp()?  I thought that was the plan for auto mode?

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: Oh snap.  Probably just make this recursive to handle the basic nesting types (dict, tuple list) like we do with `pytorch._data.to_device()`.  That'll cover nearly everyone.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -467,41 +532,63 @@ def clip_grads(params):
...
+                parameters=parameters, divisor=self.hvd_config.aggregation_frequency
             )
 
-            if self.hvd_config.average_aggregated_gradients:
-                self._average_gradients(
-                    parameters=parameters, divisor=self.hvd_config.aggregation_frequency
-                )
+        if clip_grads is not None:
+            if self._scaler and self._auto_amp:
+                self._scaler.unscale()
```
rb-determined-ai, line 569: Shouldn't this be `self._scaler.unscale_(optimizer)`?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -218,6 +232,47 @@ def to_device(self, data: pytorch._Data) -> pytorch.TorchData:
         """
         return pytorch.to_device(data, self.device, self._to_device_warned_types)
 
+    def use_amp(self) -> None:
+        self.wrap_scaler(torch.cuda.amp.GradScaler(), automatic=True)
+
+    def wrap_scaler(self, scaler: Any, automatic: bool = False) -> Any:
```
rb-determined-ai, line 238: Can we avoid having two user-facing ways to set automatic mode?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -467,41 +532,63 @@ def clip_grads(params):
                 stepping the optimizer. If false, you need to call ``optimizer.zero_grad()``
                 manually. Note that if :ref:`optimizations.aggregation_frequency
                 <config-aggregation-frequency>` is greater than 1, ``auto_zero_grads`` must be true.
+            scaler(``torch.cuda.amp.GradScaler``, optional): The scale to use for stepping the
+                optimizer. This should be unset if not using AMP, and is unnecessary if
+                ``PyTorchContext.wrap_scaler`` was called with ``automatic = True``.
```
rb-determined-ai, line 537: remember to update this docstring

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -467,41 +532,63 @@ def clip_grads(params):
...
-            # Communication needs to be synchronized so that is completed
-            # before we apply gradient clipping and `step()`.
-            if self.hvd_config.use and not self._use_amp:
-                optimizer.synchronize()
-
-            parameters = (
-                [p for group in optimizer.param_groups for p in group.get("params", [])]
-                if not self._use_amp
-                else apex.amp.master_params(optimizer)
+
```
rb-determined-ai, line 545: wow, this function looks like a million times better than before.

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Yeah :) I've fixed this twice now and gotten it mixed up in my testing - copying this file back and forth between experiments :D

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Yeah when I started doing my big test with this (BERT) I saw this need to handle things other than Tensors. BERT sends back a dictionary, but yeah I guess I need to check the types inside that too...

harness/determined/pytorch/_pytorch_context.py
mackrorysd: Just to close the loop on this: seems to me we would still want step_optimizer because it includes distributed training logic. It can be simplified in a post-APEX world when we get there.

Seems to me backward() is not actually required today unless you're using APEX. It's also required if you're using the experimental easy-AMP mode. But other than that, if a user just used PyTorch's backward() natively, I don't think anything would be affected. But we don't say that in the docs: very likely in the future something will come up that'll make us wish we could easily inject stuff in backward(). So I'm going to stick with the status quo here. I think it's best to remove APEX support once we know native AMP is viable and successfully in use with customers. But we should keep backward() as-is.

shiyuann: sounds good to me. do we have a plan to know how we decide "native AMP is viable and successfully in use with customers"?

mackrorysd: We'll just ask :D There's already been some customer input that went into the design choices here.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -477,12 +531,12 @@ def clip_grads(params):
         if self._should_communicate_and_update():
             # Communication needs to be synchronized so that is completed
             # before we apply gradient clipping and `step()`.
```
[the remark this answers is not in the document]
rb-determined-ai: We do not support multiple losses with AMP.

It seems to me that we were very indecisive on how much we supported apex amp because we have e.g. the `num_losses` parameter but we also have checks against calling `context.backward()` multiple times in a single `train_batch`.  We also wrap the `optimizer.synchronize()` call in a `self._should_communicate_and_update()` check but we have checks to guarantee that amp is not used when aggregation frequency >1.

I am not sure why it was written that way anymore (did we think we could eventually solve those problems?) but this code could stand to get a lot clearer by just explaining those assumptions in the places where we depend on them.

I'm pretty happy with the comment as you have it right now.

rb-determined-ai (commented):
Everything but the after-train_batch-hook looks correct to me.

A few cleanup things remain though.

Great work on this!

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -85,6 +94,48 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
         model_id = len(self.models)
         self._main_model.__setattr__(f"model_{model_id}", model)
 
+        if self.experimental._auto_amp:
+            # Ensure the forward pass is wrapped in an autocast context:
+            saved_model = model
```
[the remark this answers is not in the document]
mackrorysd: So something wacky was going on with the scopes and bindings - without the saved_model thing there was infinite recursion because model == self as far as the class is concerned. A function would also clean this up nicely though.

examples/tutorials/mnist_pytorch/model_def.py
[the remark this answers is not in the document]
mackrorysd: Ignored the files that are symlinked to /examples/. It's a little messy - I would've preferred it if isort had an option to ignore symlinks, but... if you have a better idea I'm all ears :) Reverted the commit that persisted this, too.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -258,6 +258,16 @@ def _auto_step_lr_scheduler_per_batch(
...
                 lr_scheduler.step()
 
+    def _should_update_scaler(self) -> bool:
+        if not self.context._scaler or not self.context.experimental._auto_amp:
+            return False
+        if self.hvd_config.use:
+            if self.context._should_communicate_and_update():
+                return True
+            else:
+                return False
```
rb-determined-ai, line 268: nit: this could just be `return self.context._should_communicate_and_update()`

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -73,6 +76,62 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        def end_fp16(
+            module: torch.nn.Module,
+            input: Any,
+            output: Any,
+            warned_types: Optional[Set[Type]] = None,
+        ) -> Any:
+
+            # Never print errors recursively.
+            if warned_types is None:
+                warned_types = set()
```
rb-determined-ai, line 110: This will prevent multiple calls per forward pass, but not multiple calls across forward passes.

I think if you put warned_types outside of `end_fp16` instead of having as a parameter this should work fine.

The reason `to_device()` does it differently is that `to_device` is defined at the library level but the warnings are tracked at the context level.  In this case, `end_fp` is defined once per model and that's probably a fine granularity to just use a single bound variable.

rb-determined-ai (approved):
Great work!

PR #1921, opened 2021-02-03
fix: add default query limit and add missing sort by state [DET-4919]

stoksc:
This change updates the how Postgres function `public.page_info` behaves with different `limit` parameters:
* limit -2 will return an empty list with the page info.
* limit -1 will be unlimited
* limit 0 will default to 100
* limit > 0 will have the usual behavior.
It also adds a missing entry for order by state.
## Test Plan
- [x] run a few large adaptive noops and click through swagger to see all limits and sort bys working as intended.


master/static/migrations/20210203150406_change-limit-behavior.up.sql
```diff
@@ -0,0 +1,24 @@
...
+    RETURNS json
+    LANGUAGE sql IMMUTABLE
+AS $$
+WITH start_index AS (
+    SELECT (CASE WHEN "offset" < 0
+                 THEN total + "offset"
+                 ELSE "offset" END) AS start_index
+), end_index AS (
+    SELECT (CASE WHEN (SELECT start_index FROM start_index) + "limit" > total THEN total
+                 WHEN "limit" = 0 THEN least(100::bigint, total - (SELECT start_index FROM start_index))
```
shiyuann, line 11: shouldn't this be:
```
WHEN "limit" = 0 THEN least(100::bigint + (SELECT start_index FROM start_index), total)
```
?

stoksc: yeah sorry. i had fixed that and another issue why my pre push hook failed so it didn't push.. stupid gocritic linter.

master/static/migrations/20210203150406_change-limit-behavior.up.sql
```diff
@@ -0,0 +1,24 @@
+CREATE OR REPLACE FUNCTION public.page_info(total bigint, "offset" int, "limit" int)
+    RETURNS json
+    LANGUAGE sql IMMUTABLE
+AS $$
+WITH start_index AS (
+    SELECT (CASE WHEN "offset" < 0
+                 THEN total + "offset"
+                 ELSE "offset" END) AS start_index
+), end_index AS (
+    SELECT (CASE WHEN (SELECT start_index FROM start_index) + "limit" > total THEN total
```
shiyuann, line 10: I am thinking of this case:
```
offset = a number > total
start_index = offset
end_index = start_index + limit < total
```
We also should fail limit < -2...

stoksc: Ok. -2 already fails and that case (and offset < -total) now just default to returning no rows (which makes plenty of sense to me).

webui/react/src/contexts/AppContexts.tsx
```diff
@@ -59,7 +59,7 @@ const AppContexts: React.FC = () => {
     triggerNotebooksRequest({});
     triggerShellsRequest({});
     triggerTensorboardsRequest({});
-    triggerActiveExperimentsRequest({ limit: 0, states: activeRunStates });
+    triggerActiveExperimentsRequest({ limit: -2, states: activeRunStates });
```
hkang1, line 62: 🎉

shiyuann (approved):
Looks good!

PR #1922, opened 2021-02-03
fix: correct cancel confirm button label to confirm [DET-4966]

hkang1:
## Description

Before:

After:
Screen Shot 2021-02-03 at 1 48 48 PM

Note: The table batch action will remain as `Cancel` but the confirmation button in the modal will be labeled as `Confirm`.

## Checklist


hamidzr (approved):
LGTM IMO it still could be a  bit confusing. I think ideally we'd have something other than cancel for aborting.

PR #1924, opened 2021-02-03
fix: correct the comparison function when numbers are fractions [DET-4969]

hkang1:
## Description

number comparison function is failing when the numbers being compared are fractions

## Test Plan

- [ ] check to make sure the metric numbers are sorted properly under trial details

## Checklist

### DIFF
--- a/webui/react/src/utils/data.ts
+++ b/webui/react/src/utils/data.ts
@@ -126,7 +126,7 @@ export const numericSorter = (a?: number, b?: number, reverseOrder = false): num
     return 0;
   }
   if (a != null && b == null) return reverseOrder ? -1 : 1;
-  if (b == null && b != null) return reverseOrder ? 1 : -1;
+  if (a == null && b != null) return reverseOrder ? 1 : -1;
   return 0;
 };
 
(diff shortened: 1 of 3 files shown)


webui/react/src/utils/data.ts
```diff
@@ -119,7 +119,12 @@ export const metricNameSorter = (a: MetricName, b: MetricName): number => {
 };
 
 export const numericSorter = (a?: number, b?: number, reverseOrder = false): number => {
-  if (a != null && b != null) return reverseOrder ? b - a : a - b;
+  if (a != null && b != null) {
+    const diff = reverseOrder ? b - a : a - b;
+    if (diff < 0) return -1;
+    if (diff > 0) return 1;
+    return 0;
+  }
```
hkang1, line 127: there's probably a more elegant way of doing this but don't have the brain power to spend here :D

hamidzr: this lg and is very readable

hamidzr: this would be a great candidate and opportunity for a unit test.

hamidzr (approved):

PR #1930, opened 2021-02-04
fix: always deallocate resources if a task actor fail [DET-4973]

shiyuann:
## Description

Previously, the agent resource pool relies on the container state change message to deallocate resources. This PR changes that to always deallocate resources when a task actor fails.

## Test Plan

1. Start the cluster
2. Alter the table using the following SQL to fail the trial actor.

```sql
alter table trials drop column seed;
```

3. submit an adaptive experiment.
4. see if the resources get deallocated.

## Commentary (optional)

This PR doesn't fix that the experiment state should error out if a trial fails to write to the DB.

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/resource_pool.go
+++ b/master/internal/resourcemanagers/resource_pool.go
@@ -146,6 +146,12 @@ func (rp *ResourcePool) releaseResource(ctx *actor.Context, handler *actor.Ref) 
 
 func (rp *ResourcePool) resourcesReleased(ctx *actor.Context, handler *actor.Ref) {
 	ctx.Log().Infof("resources are released for %s", handler.Address())
+	if allocated := rp.taskList.GetAllocations(handler); allocated != nil {
+		for _, allocation := range allocated.Allocations {
+			typed := allocation.(*containerAllocation)
+			typed.agent.deallocateContainer(typed.container.id)
+		}
+	}
 	rp.taskList.RemoveTaskByHandler(handler)
 }
 
@@ -219,7 +225,6 @@ func (rp *ResourcePool) Receive(ctx *actor.Context) error {
 	case
 		sproto.AddAgent,
 		sproto.AddDevice,
-		sproto.FreeDevice,
 		sproto.RemoveDevice,
 		sproto.RemoveAgent:
 		return rp.receiveAgentMsg(ctx)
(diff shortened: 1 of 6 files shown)


stoksc (commented):
I worry a little there are edge cases where trials/commands (really trials) don't send `sproto.ResourcesReleased` properly and `sproto.FreeDevice` may be handling properly freeing the slots unknown to us. This change looks like moves that all to trigger on `sproto.ResourcesReleased` now. Didn't we just have an issue in `trial.go` where that wasn't getting sent right?

shiyuann:
@stoksc What edge cases are thinking in your mind? There is no need to worry that trials/commands don't always send `sproto.ResourcesReleased` to the resource manager because when a task is added the resource manager will register a special type of notification actor to track the life of the task actor. If a task actor ends, the resource manager will get the notification.

This PR fixes `sproto.FreeDevice` not being sent to the resource manager if a task actor fail between being allocated and starting the allocated containers.

> This change looks like moves that all to trigger on `sproto.ResourcesReleased` now

Yes. That's the purpose. The container state will be totally managed by the task actor. However, we don't have the logic to clean up containers if a task actor fails.

> Didn't we just have an issue in trial.go where that wasn't getting sent right?

No, the issue was `FreeDevice` didn't get sent. The resource manager used to depend on `FreeDevice` message to know which devices are freed. The issue was the trial fail before starting a container and after being allocated. So the device is not technically occupied so no `FreeDevice` will be ever sent to the resource manager.

stoksc:
@shiyuann This just seems like one of those changes that could introduce subtle bugs, which is why I'm trying to make sure I understand it.

> There is no need to worry that trials/commands don't always send sproto.ResourcesReleased to the resource manager because when a task is added the resource manager will register a special type of notification actor to track the life of the task actor. If a task actor ends, the resource manager will get the notification.

Now that I've thought about it a bit, I have an example. We've had bugs before where a task actor (trials) were corrupted in restart or during a run and left hung in an active state. Now this would also cause the trial to not release its resources too, right?

I guess is I worry this will uncover or make worse other bugs with task termination logic, so I wanted to ask what you thought. I can take some time to review really thoroughly later for these kinds of bugs.

stoksc:
> Now that I've thought about it a bit, I have an example. We've had bugs before where a task actor (trials) were corrupted in restart or during a run and left hung in an active state. Now this would also cause the trial to not release its resources too, right?

We talked offline and agree the resource pool really shouldn't be in charge of this, and if this happens we should fix the task actors.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -146,6 +146,13 @@ func (rp *ResourcePool) releaseResource(ctx *actor.Context, handler *actor.Ref)
 
 func (rp *ResourcePool) resourcesReleased(ctx *actor.Context, handler *actor.Ref) {
 	ctx.Log().Infof("resources are released for %s", handler.Address())
+	allocated := rp.taskList.GetAllocations(handler)
+	if allocated != nil && len(allocated.Allocations) > 0 {
```
stoksc, line 150:
```suggestion
	if allocated != nil {
```

stoksc (approved):
lgtm

shiyuann:
ignore the webui e2e test since it's unrelated.

PR #1932, opened 2021-02-04
fix: show zero value metrics [DET-4686]

hkang1:
## Description

- show zero value metrics (currently treated as empty)
- fix handling of negative numbers (treated as decimals always currently)
- avoid showing integers as decimal or scientific notations

Before:
Screen Shot 2021-02-04 at 10 46 15 AM

After:
Screen Shot 2021-02-04 at 10 45 56 AM

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails.tsx
+++ b/webui/react/src/pages/TrialDetails.tsx
@@ -179,7 +179,7 @@ const TrialDetailsComp: React.FC = () => {
     const metricRenderer = (metricName: MetricName) => {
       const metricCol = (_: string, record: Step) => {
         const value = extractMetricValue(record, metricName);
-        return value ? <HumanReadableFloat num={value} /> : undefined;
+        return value != null ? <HumanReadableFloat num={value} /> : undefined;
       };
       return metricCol;
     };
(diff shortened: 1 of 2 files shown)


webui/react/src/pages/TrialDetails.tsx
```diff
@@ -179,7 +179,7 @@ const TrialDetailsComp: React.FC = () => {
     const metricRenderer = (metricName: MetricName) => {
       const metricCol = (_: string, record: Step) => {
         const value = extractMetricValue(record, metricName);
-        return value ? <HumanReadableFloat num={value} /> : undefined;
+        return value != null ? <HumanReadableFloat num={value} /> : undefined;
```
apizzini, line 182: `!=` instead of `!==` on purpose?

hkang1: Ah yes, so `!==` will check against only `null`. `value != null` is the same as `value !== null && value !== undefined`. `value == null` is equivalent to `value === null || value === undefined`.

Basically it protects against both `null` and `undefined`.

hkang1: https://stackoverflow.com/questions/16607761/why-null-undefined-is-true-in-javascript

apizzini: Thanks :)

apizzini (approved):
LGTM

PR #1934, opened 2021-02-04
chore: single batch evaluation for local test mode [DET-2931]

hamidzr:
## Description

- supporting `PyTorch`, `TFEstimator` and `TFKeras` trials

## Checklist

### DIFF
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -62,10 +62,10 @@ def _make_local_execution_exp_config(input_config: Optional[Dict[str, Any]]) -> 
 
 def _make_local_execution_env(
     managed_training: bool,
+    test_mode: bool,
     config: Optional[Dict[str, Any]],
     hparams: Optional[Dict[str, Any]] = None,
     limit_gpus: Optional[int] = None,
-    test_mode: bool = False,
 ) -> Tuple[det.EnvContext, det.RendezvousInfo, horovod.HorovodContext]:
     config = det.ExperimentConfig(_make_local_execution_exp_config(config))
     hparams = hparams or api.generate_random_hparam_values(config.get("hyperparameters", {}))
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -65,6 +65,7 @@ def _make_local_execution_env(
     config: Optional[Dict[str, Any]],
     hparams: Optional[Dict[str, Any]] = None,
     limit_gpus: Optional[int] = None,
+    test_mode: bool = False
 ) -> Tuple[det.EnvContext, det.RendezvousInfo, horovod.HorovodContext]:
     config = det.ExperimentConfig(_make_local_execution_exp_config(config))
     hparams = hparams or api.generate_random_hparam_values(config.get("hyperparameters", {}))
@@ -97,6 +98,7 @@ def _make_local_execution_env(
         det_cluster_id="",
         trial_seed=config.experiment_seed(),
         managed_training=managed_training,
+        test_mode=test_mode,
     )
     rendezvous_ports = env.rendezvous_ports()
     rendezvous_info = det.RendezvousInfo(
--- a/harness/determined/_execution.py
+++ b/harness/determined/_execution.py
@@ -65,7 +65,7 @@ def _make_local_execution_env(
     config: Optional[Dict[str, Any]],
     hparams: Optional[Dict[str, Any]] = None,
     limit_gpus: Optional[int] = None,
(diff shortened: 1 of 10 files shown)


harness/determined/_execution.py
```diff
@@ -65,6 +65,7 @@ def _make_local_execution_env(
     config: Optional[Dict[str, Any]],
     hparams: Optional[Dict[str, Any]] = None,
     limit_gpus: Optional[int] = None,
+    test_mode: bool = False,
```
rb-determined-ai, line 68: Can we make this a required parameter?  It's not really optional; we always need to know if the environment is in test mode, the same as we always need to know if it's a managed_training or not.

harness/determined/experimental/_native.py
```diff
@@ -97,7 +97,7 @@ def _make_test_workloads(
     check.eq(len(batch_metrics), config.scheduling_unit())
     logging.debug(f"Finished training, metrics: {batch_metrics}")
```
rb-determined-ai, line 98: Can you also make this logging.debug into an info while you are here?  Also below with the validation metrics.

harness/determined/experimental/_native.py
```diff
@@ -192,13 +192,16 @@ def test_one_batch(
     trial_class: Optional[Type[det.Trial]] = None,
     config: Optional[Dict[str, Any]] = None,
 ) -> Any:
+
```
rb-determined-ai, line 195: this extra line seems unnecessary

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -334,8 +335,10 @@ def _compute_validation_metrics(self) -> workload.Response:
             batch_metrics = []
 
             self.validation_loader = cast(torch.utils.data.DataLoader, self.validation_loader)
-            check.gt(len(self.validation_loader), 0)
-            for batch in self.validation_loader:
+            val_loader_len = len(self.validation_loader)
+            check.gt(val_loader_len, 0)
+            batch_stop_idx = val_loader_len if not self.context.env.test_mode else 1
+            for batch in islice(self.validation_loader, 0, batch_stop_idx):
```
rb-determined-ai, line 341: non-blocking:  What if you just put an `if self.env.test_mode: break` at the end of the loop?  I had never heard of `islice()` before, and while it is definitely not wrong to use advanced features of a language, in this case the simple thing seems so simple it's at least worth suggesting.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -747,6 +747,9 @@ def _launch_evaluate(self) -> Any:
         use_model_metrics = version.parse(tf.__version__) < version.parse("2.2.0")
         evaluate_kwargs = {} if use_model_metrics else {"return_dict": True}
 
+        if self.context.env.test_mode:
```
rb-determined-ai, line 750: throughout: All these TrialController objects have a `self.env` that you should use directly.

harness/determined/experimental/_native.py
```diff
@@ -192,13 +192,16 @@ def test_one_batch(
...
     # Override the scheduling_unit value to 1.
     # TODO(DET-2931): Make the validation step a single batch as well.
     config = {**(config or {}), "scheduling_unit": 1}
 
     logging.info("Running a minimal test experiment locally")
     checkpoint_dir = tempfile.TemporaryDirectory()
-    env, rendezvous_info, hvd_config = det._make_local_execution_env(True, config, limit_gpus=1)
+    env, rendezvous_info, hvd_config = det._make_local_execution_env(
+        True, config, limit_gpus=1, test_mode=True
```
rb-determined-ai, line 203: while you are here, can you switch this to be all keyword arguments?  That initial bare `True` is a little opaque.

rb-determined-ai (commented):
This is looking great.

harness/determined/_execution.py
hamidzr: hmm by providing default we'd make the usage simpler but it'll be less explicit. either way works

harness/determined/pytorch/_pytorch_trial.py
hamidzr: originally I was designing it to allow arbitrary batch length but early breaking is simpler 👍🏼

rb-determined-ai (approved):
Looks fantastic.  Very nice work.

hamidzr:
thanks @rb-determined-ai

PR #1937, opened 2021-02-05
feat: expose primitives for pytorch dataloaders

rb-determined-ai:
## Description

My first idea for handling iterable datasets was  #1928, but I think that PR offered tooling to the user at the wrong layer; lower-layer than what the happy-path users wanted, and higher-layer than what the advanced users needed.

This PR is another idea, which offers high-level tooling for the happy-path users, and clearly documented primitives for building data loaders that are "determined-compliant" in other cases.

Here are some meeting notes from the #ml-ag meeting on 2 Nov 2020 (abriged):

> Aaron: Accept arbitrary Pytorch DataLoaders to reduce initial porting overhead. Could also upstream some of this.
> - AH: We wanted to do it for at least IterableDatasets, but we should do this for all data loaders (while explaining caveats). Doesn’t require YogaDL necessarily (but may have caveats to behaviors). 1. It’s up to users to make sure that each GPU gets a unique data sample. 2. Pause/restart for HP search.
> - KP: question: they’ll need to do some work to get the GPU sampling to work? What does this look like from a user perspective?
>   - RB: maybe the context should offer a special sampler… (editor's note: #1928 was a failed attempt at this)
>   - AH: the other problem is the start/restart logic. We shouldn’t even try to fix that if they’re using a PT data loader
>   - RB: you’d fix that in the sampler anyway
>   - AH: our easy 1-line config change is for a DDL. if you don’t want to use it, there are these disclaimers, will take more on your side. Kind of what we do for TF dataset right now, restart logic is just not correct unless you use YogaDL but we don’t force that
> - VM: devil’s in the details, what do we do if we get a custom data loader that doesn’t work w/ Determined?
>   - AH: no, we don’t fail for TF datasets right now etc.
>   - RB: the answer is some people want it to fail, some people don’t. Like compiler warnings. We kind of need a feature like this. Document correct errors, which conditions, and allow users to request strictness depending on their use case
>   - AH: i think this is useful, but i also don’t think we should block on this
>   - VM: we need to let people know what they’re buying into; 1. If a user hasn’t read the docs and say BYODL, there’s no warning that there’s a science-compromising issue

As may be clear from the above discourse, the issue here has less to do with difficulty of implementation and more to do with correctness and being able to give assurance to the user about correctness.  That gets harder as we move away from plugin-api paradigms and into sprinkle api paradigms.  However, no users have ever complained about the pytorch-flexible-primitives API, where we made a similar transition from rigid, guaranteed-correct APIs to a "remember to make these calls" style of API.

**The conclusion that this PR takes is: "make the happy-path easy and clearly document how to make the advanced path correct"**

This "clear documentation" is at the top of `harness/determined/pytorch/samplers.py` right now. 

## Example usage:

This example uses the sampler primitives directly, as an academic exercise.

```python
class MNistTrial(PyTorchTrial):
    def build_training_data_loader(self) -> torch.utils.data.DataLoader:
        train_data = data.get_dataset(self.download_directory, train=True)

        mode = "I like pain"

        if mode == "I like easy":
            return det.pytorch.DataLoader(train_data)

        if mode == "I like pain":

            from determined.pytorch import samplers

            seed = self.context.get_trial_seed()
            rank = self.context.distributed.get_rank()
            size = self.context.distributed.get_size()
            skip = self.context.get_initial_batch()
            batch_size = self.context.get_per_slot_batch_size()

            # Shuffle-repeat-shard-batch-skip, just like Determined told me to.
            sampler = torch.utils.data.SequentialSampler(train_data)
            sampler = samplers.ReproducibleShuffleSampler(sampler, seed)
            sampler = samplers.RepeatSampler(sampler)
            sampler = samplers.DistributedSampler(sampler, num_workers=size, rank=rank)
            batch_sampler = torch.utils.data.BatchSampler(sampler, batch_size, drop_last=False)
            batch_sampler = samplers.SkipBatchSampler(batch_sampler, skip)

            return torch.utils.data.DataLoader(train_data, batch_sampler=batch_sampler)
```

## A more realistic advanced use case:

This is how @katport could have ported the detectron2 model dataloaders, with their fancy IterableDataset-based speedups.

See the original function [here](https://github.com/katport/determined/blob/DET-3158/examples/experimental/trial/detectron2/detectron2_files/data.py#L407)

```python
def build_detection_train_loader(cfg, batch_size, mapper=None):
    """
    A data loader is created by the following steps:
    1. Use the dataset names in config to query :class:`DatasetCatalog`, and obtain a list of dicts.
    2. Coordinate a random shuffle order shared among all processes (all GPUs)
    3. Each process spawn another few workers to process the dicts. Each worker will:
       * Map each metadata dict into another format to be consumed by the model.
       * Batch them by simply putting dicts into a list.
    The batched ``list[mapped_dict]`` is what this dataloader will yield.
    Args:
        cfg (CfgNode): the config
        mapper (callable): a callable which takes a sample (dict) from dataset and
            returns the format to be consumed by the model.
            By default it will be `DatasetMapper(cfg, True)`.
    Returns:
        an infinite iterator of training data
    """
    dataset_dicts = get_detection_dataset_dicts(
        cfg.DATASETS.TRAIN,
        filter_empty=cfg.DATALOADER.FILTER_EMPTY_ANNOTATIONS,
        min_keypoints=cfg.MODEL.ROI_KEYPOINT_HEAD.MIN_KEYPOINTS_PER_IMAGE
        if cfg.MODEL.KEYPOINT_ON
        else 0,
        proposal_files=cfg.DATASETS.PROPOSAL_FILES_TRAIN if cfg.MODEL.LOAD_PROPOSALS else None,
    )
    dataset = DatasetFromList(dataset_dicts, copy=False)

    if mapper is None:
        mapper = DatasetMapper(cfg, True)
    dataset = MapDataset(dataset, mapper)

    sampler_name = cfg.DATALOADER.SAMPLER_TRAIN
    print ('sampler_name: ', sampler_name)
    logger = logging.getLogger(__name__)
    logger.info("Using training sampler {}".format(sampler_name))
    # TODO avoid if-else?
    if sampler_name == "TrainingSampler":
        sampler = TrainingSampler(len(dataset))
    elif sampler_name == "RepeatFactorTrainingSampler":
        repeat_factors = RepeatFactorTrainingSampler.repeat_factors_from_category_frequency(
            dataset_dicts, cfg.DATALOADER.REPEAT_THRESHOLD
        )
        sampler = RepeatFactorTrainingSampler(repeat_factors)
    else:
        raise ValueError("Unknown training sampler: {}".format(sampler_name))

################################
# Use the new sampler primitives

    # We know:
    #  - the shuffle is handled by TrainingSampler
    #  - the repeat is handled by TrainingSampler
    #  - we still need sharding (detectron2 only shards for torch-native dtrain)
    #  - we still need skipping to support for continuing training
    #
    # We assume:
    #  - reproducibility is not required; it is enough to train on fresh data
    #    when we continue from a checkpoint.  It would be nearly impossible
    #    to repopulate the buckets of the AspectRatioGroupedDataset in a
    #    reproducible way.
    from determined.pytorch import samplers

    context = ... # You need to get this object here somehow

    # Shard before skip:
    rank = context.distributed.get_rank()
    size = context.distributed.get_size()
    sampler = samplers.DistributedSampler(sampler, num_workers=size, rank=rank)

    # Skip at the record level, since all batches are the same size in training
    batch_size = context.get_per_slot_batch_size()
    initial_batch = context.get_initial_batch()
    skip_records = batch_size * initial_batch
    sampler = det.pytorch.samplers.SkipSampler(sampler, skip=skip_records)

##############################

    return build_batch_data_loader(
        dataset,
        sampler,
        cfg.SOLVER.IMS_PER_BATCH,
        aspect_ratio_grouping=cfg.DATALOADER.ASPECT_RATIO_GROUPING,
        num_workers=cfg.DATALOADER.NUM_WORKERS,
        batch_size=batch_size
    )
```

### DIFF
--- a/harness/determined/_train_context.py
+++ b/harness/determined/_train_context.py
@@ -176,6 +176,9 @@ class _TrainContext(metaclass=abc.ABCMeta):
         )
         self._stop_requested = stop_requested
 
+    def get_initial_batch(self) -> int:
+        return self.env.initial_workload.total_batches_processed
+
 
 class TrialContext(_TrainContext):
     """
--- a/harness/determined/pytorch/samplers.py
+++ b/harness/determined/pytorch/samplers.py
@@ -1,0 +1,368 @@
+"""
+Guidelines for Reproducible Datasets
+====================================
+
+.. note::
+
+   Normally, using ``det.pytorch.DataLoader`` is required and handles all of the below details
+   without any special effort on your part (see :ref:`pytorch-data-loading`).  When
+   ``det.pytorch.DataLoader()`` is not suitable (especially in the case of ``IterableDatasets``),
+   you may disable this requirement by calling
+   :meth:`context.experimental.disable_dataset_reproducibility_checks()
+   <determined.pytorch.PyTorchExperimentalContext.disable_dataset_reproducibility_checks>`
+   in your Trial's ``__init__()`` method.  Then you may choose to follow the below guidelines for
+   ensuring dataset reproducibility on your own.
+
+Achieving a reproducible dataset that is able to pause and continue (sometimes called "incremental
+training") is easy if you follow a few rules.
+
+- Even if you are going to ultimately return an IterableDataset, it is best to use PyTorch's Sampler
+  class as the basis for choosing the order of records.  Operations on Samplers are quick and cheap,
+  while operations on data afterwards are expensive.  For more details, see the discussion of random
+  vs sequential access `here <https://yogadl.readthedocs.io>`_.  If you don't have a custom sampler,
+  start with a simple one:
+
+  .. code::python
+
+     sampler = torch.utils.data.SequentialSampler(my_dataset)
+
+- **Shuffle first**: Always use a reproducible shuffle when you shuffle.  Determined provides two
+  shuffling samplers for this purpose; the ``ReproducibleShuffleSampler`` for operating on records
+  and the ``ReproducibleShuffleBatchSampler`` for operating on batches.  You should prefer to
+  shuffle on records (use the ``ReproducibleShuffleSampler``) whenever possible, to achieve the
+  highest-quality shuffle.
+
+- **Repeat when training**: In Determined, you always repeat your training dataset and you
+  never repeat your validation datasets.  Determined provides a RepeatSampler and a
+  RepeatBatchSampler to wrap your sampler or batch_sampler.  For your training dataset, make sure
+  that you always repeat AFTER you shuffle, otherwise your shuffle will hang.
+
+- **Always shard, and not before a repeat**: Use Determined's DistributedSampler or
+  DistributedBatchSampler to provide a unique shard of data to each worker based on your sampler or
+  batch_sampler.  It is best to always shard your data, and even when you are not doing distributed
+  training, because in non-distributed-training settings, the sharding is nearly zero-cost, and it
+  makes distributed training seamless if you ever want to use it in the future.
+
+  It is generally important to shard after you repeat, unless you can guarantee that each shard
+  of the dataset will have the same length.  Otherwise, differences between the epoch boundaries for
+  each worker can grow over time, especially on small datasets.  If you shard after you repeat, you
+  can change the number of workers arbitrarily without issue.
+
+- **Skip when training, and always last**: In Determined, training datasets should always be able to
+  start from an arbitrary point in the dataset.  This allows for advanced hyperparameter searches
+  and responsive preemption for training on spot instances in the cloud.  The easiest way to do
+  this, which is also very efficient, is to apply a skip to the sampler.
+
+  Determined provides a SkipBatchSampler that you can apply to your batch_sampler for this purpose.
+  There is also a SkipSampler that you can apply to your sampler, but you should prefer to skip on
+  batches unless you are confident that your dataset always yields identical size batches, where the
+  number of records to skip can be reliably calculatd from the number of batches already trained.
+
+  Always skip AFTER your repeat, so that the skip only happens once, and not on every epoch.
+
+  Always skip AFTER your shuffle, to preserve the reproducibility of the shuffle.
+
+Here is some example code that follows each of these rules that you can use as a starting point if
+you find that the built-in context.DataLoader() does not support your use case.
+
+  .. code::python
+
+    def make_batch_sampler(
+        sampler_or_dataset,
+        mode,  # mode="training" or mode="validation"
+        shuffle_seed,
+        num_workers,
+        rank,
+        batch_size,
+        skip,
+    ):
+        if isinstance(sampler_or_dataset, torch.utils.data.Sampler):
+            sampler = sampler_or_dataset
+        else:
+            # Create a SequentialSampler if we started with a Dataset.
+            sampler = torch.utils.data.SequentialSampler(sampler_or_dataset)
+
+        if mode == "training":
+            # Shuffle first.
+            sampler = samplers.ReproducibleShuffleSampler(sampler, shuffle_seed)
+
+            # Repeat when training.
+            sampler = samplers.RepeatSampler(sampler)
+
+        # Always shard, and not before a repeat.
+        sampler = samplers.DistributedSampler(sampler, num_workers=num_workers, rank=rank)
+
+        # Batch before skip, because Determined counts batches, not records.
+        batch_sampler = torch.utils.data.BatchSampler(sampler, batch_size, drop_last=False)
+
+        if mode == "training":
+            # Skip when training, and always last.
+            batch_sampler = samplers.SkipBatchSampler(batch_sampler, skip)
+
+        return batch_sampler
+
+
+    class MyPyTorchTrial(det.pytorch.PyTorchTrial):
+        def __init__(self, context):
+            context.experimental.disable_dataset_reproducibility_checks()
+
+        def build_training_data_loader(self):
+            my_dataset = ...
+
+            batch_sampler = make_batch_sampler(
+                dataset=my_dataset,
+                mode="training",
+                seed=self.context.get_trial_seed(),
+                num_workers=self.context.distributed.get_size(),
+                rank=self.distributed.get_rank(),
+                batch_size=self.context.get_per_slot_batch_size(),
+                skip=self.context.get_initial_batch(),
+            )
+
+            return torch.utils.data.DataLoader(my_dataset, batch_sampler=batch_sampler)
+"""
+from typing import Iterator
+
+import numpy as np
+import torch
+
+from determined.common import check
+
+
+class RepeatSampler(torch.utils.data.Sampler):
+    """
+    RepeatSampler yields infinite batches indices by repeatedly iterating
+    through the batches of another Sampler. __len__ is just the length of
+    the underlying Sampler.
+    """
+
+    def __init__(self, sampler: torch.utils.data.Sampler) -> None:
+        self._sampler = sampler
+
+    def __len__(self) -> int:
+        return len(self._sampler)  # type: ignore
+
+    def __iter__(self) -> Iterator:
+        while True:
+            yield from self._sampler
+
+
+class RepeatBatchSampler(torch.utils.data.BatchSampler):
+    """
+    RepeatBatchSampler yields infinite batches indices by repeatedly iterating
+    through the batches of another BatchSampler. __len__ is just the length of
+    the underlying BatchSampler.
+    """
+
+    def __init__(self, batch_sampler: torch.utils.data.BatchSampler) -> None:
+        self.batch_sampler = batch_sampler
+
+    def __len__(self) -> int:
+        return len(self.batch_sampler)
+
+    def __iter__(self) -> Iterator:
+        while True:
+            yield from self.batch_sampler
+
+
+class DistributedSampler(torch.utils.data.Sampler):
+    """
+    DistributedSampler will iterate through an underlying sampler and return samples which
+    belong to this shard.
+
+    DistributedSampler is different than the PyTorch built-in torch.utils.data.DistributedSampler
+    because theirs is meant to be a standalone sampler.  Theirs does shuffling and assumes a
+    constant size dataset as an input.  Ours is meant to be used a building block in a chain of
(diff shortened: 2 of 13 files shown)


examples/tutorials/mnist_pytorch/model_def.py
```diff
@@ -61,9 +61,34 @@ def build_training_data_loader(self) -> DataLoader:
             self.data_downloaded = True
 
         train_data = data.get_dataset(self.download_directory, train=True)
-        return DataLoader(train_data, batch_size=self.context.get_per_slot_batch_size())
 
-    def build_validation_data_loader(self) -> DataLoader:
+        mode = "I like pain"
```
katport, line 65: I want to call out that this helps porting in some cases, but this does not solve the case where a user has a get_data function that returns a dataloader and just passing that into build_data_loader. Talking to @rb-determined-ai it doesn't seem possible.

rb-determined-ai: Right, with this change, they can do exactly that, but if it's wrong, it's wrong.  There's no way for us to modify their existing data loader to make it correct, we can only give the user to tools to take care of it themselves.

If they have a map dataset, they can still return a `det.pytorch.DataLoader` as before.

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
+"""
+Guidelines for Reproducible Datasets.
+
+.. note::
+
+   If you are using a MapDataset and not an IterableDataset, make your life easy and use the
+   context.TrainingDataLoader() and context.ValidationDataLoader(), which are a drop-in replacement
+   for pytorch's built-in DataLoader, and Determined will handle all of these details.
```
katport, line 8: add link to trainingdataloader docs

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
...
+
+Achieving a reproducible dataset that is able to pause and continue (sometimes called "incremental
+training") is easy if you follow a few rules.
+
+- Even if you are going to ultimate return an IterableDataset, it is best to use PyTorch's Sampler
+  class as the basis for choosing the order of which records to read.  Operations on Samplers are
+  quick and cheap, while operations on data afterwards are expensive.  For more details, see the
+  discussion at TODO:ref_yogadl_readme.  If you don't have a custom sampler, start with a simple
+  one:
```
katport, line 17: "choosing the order of records"

katport: "Operations on Samplers are
  quick and cheap, while operations on data afterwards are expensive.  For more details, see the
  discussion at TODO:ref_yogadl_readme. " is a bit confusing. Can we add a one line sentence that elaborate a bit more. Maybe though the TODO will help though

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
...
+
+  .. code::python
+
+     sampler = torch.utils.data.SequentialSampler(my_dataset)
+
+- **Shuffle first**: Always use a reproducible shuffle when you shuffle. Determined provides a
+  ReproducibleShuffleSampler for this purpose.  Determined also provides a
+  ReproducibleShuffleBatchSampler class, but you should prefer to shuffle at the record level
+  whenever possible, to achieve the highest-quality shuffle.
```
katport, line 26: "but you should prefer to shuffle at the record level" - wait shouldnt the determined shuffler do that? Would it be better to say "if you write a custom sampler, you should shuffle at the record level ...."

harness/determined/pytorch/samplers.py
katport: also can we add a link of why you would use this approach: i.e. why do i need to use iterable dataset here and not the easy dropin

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
...
+
+  Determined provides a SkipBatchSampler that you can apply to your batch_sampler for this purpose.
+  There is also a SkipSampler that you can apply to your sampler, but you should prefer to skip on
+  batches unless you are confident that your dataset always yields identical size batches, where the
+  number of records to skip can be reliably calculatd from the number of batches already trained.
+
+  Always skip AFTER your repeat, so that the skip only happens once, and not on every dataset.
+
+  Always skip AFTER your shuffle, to preserve the reproducibility of the shuffle.
+
```
katport, line 57: can we add a quick bullet point that reiterates above of the order i need. This way when using as a reference its like a recipe.

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
...
+        skip,
+    ):
+        if isinstance(sampler_or_dataset, torch.utils.data.Sampler):
+            sampler = sampler_or_dataset
+        else:
+            # Create a SequentialSampler if we started with a Dataset.
+            sampler = torch.utils.data.SequentialSampler(sampler_or_dataset)
+
+        # Shuffle first.
+        if shuffle:
```
katport, line 80: its a bit confusing in this example with all the else ifs. Didn't we say above that you really should use all these steps?

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,361 @@
...
+            sampler = samplers.ReproducibleShuffleSampler(sampler, seed)
+
+        if repeat:
+            sampler = samplers.RepeatSampler(sampler)
+
+        # Always shard, and not before a repeat.
+        sampler = samplers.DistributedSampler(sampler, num_workers=num_workers, rank=rank)
+
+        # Batch before skip, because Determined counts batches, not records.
+        batch_sampler = torch.utils.data.BatchSampler(sampler, batch_size, drop_last=False)
```
katport, line 90: so even if i'm using sampler I should still put it in as a batch sampler for best use with determined?

harness/determined/pytorch/samplers.py
rb-determined-ai: I actually think I dislike the `context.TrainingDataLoader` and `context.ValidationDataLoader()`.  I think I would feel more comfortable adding *only* the low-level API in this PR, and leave the high-level API as just the `det.pytorch.DataLoader()` (which is what we currently have)

rb-determined-ai: > also can we add a link of why you would use this approach: i.e. why do i need to use iterable dataset here and not the easy dropin

Hm?  How do I know why you need to use an iterable dataset?  I feel like that's the user's concern

harness/determined/pytorch/samplers.py
rb-determined-ai: good point, I'll clear this code example up with a `mode` corresponding to `"training"` or `"validation"`

harness/determined/pytorch/samplers.py
rb-determined-ai: yes, which is explained under the `skip` bullet point

harness/determined/pytorch/samplers.py
rb-determined-ai: This is talking about the difference between the ReproducibleShuffleSampler and the ReproducibleShuffleBatchSampler.  I'll try to make that more obvious in the text.

harness/determined/pytorch/samplers.py
rb-determined-ai: it will point to yogadl.readthedocs.io, do you think that would be helpful?  (that's the README you helped edit)

harness/determined/pytorch/samplers.py
rb-determined-ai: Um... I'm a bit hesitant to do that because I feel like if I say:

> Remember: Shuffle-repeat-shard-batch-skip and you're good

then that simplifies things to the point of being confusing, like "why would I have to shuffle and repeat my validation dataset?"

I tried really hard to let the bolded parts of the bullets be a simple checklist:

* **Shuffle first**: ...
* **Repeat when training**: ...
* **Always shard, and not before a repeat**: ...
* **Skip when training, and always last**: ...

I'm not sure that it's very safe to simplify it any simpler than that...

examples/tutorials/mnist_pytorch/const.yaml
```diff
@@ -15,3 +15,4 @@ searcher:
       batches: 937 #60,000 training images with batch size 64
   smaller_is_better: true
 entrypoint: model_def:MNistTrial
+max_restarts: 0
```
katport, line 18: remove

harness/determined/pytorch/samplers.py
```diff
@@ -0,0 +1,357 @@
...
+                skip=self.context.get_initial_batch(),
+            )
+
+            return torch.utils.data.DataLoader(my_dataset, batch_sampler=batch_sampler)
+"""
+from typing import Iterator
+
+import numpy as np
+import torch
+from determined_common import check
```
katport, line 119: determined.common

harness/determined/_train_context.py
```diff
@@ -176,6 +176,9 @@ def set_stop_requested(self, stop_requested: bool) -> None:
         )
         self._stop_requested = stop_requested
 
+    def get_initial_batch(self) -> int:
```
rb-determined-ai, line 179: This function's existence needs more consideration.

We can only provide this information reliably when we have a lot of control over the checkpoint state (ie. not in the generic API).

Maybe it's ok to leave in the Trial context, since we do have that level of control, it almost seems like we could be a lot less magical about this by having the (very few) users who need this value do this:

```python
initial_batch = 0
last_checkpoint = generic_context.checkpointing.get_restart_checkpoint()  # this API doesn't exist yet
if last_checkpoint is not None:
    inital_batch = last_checkpoint.get_metadata("last_batch")  # this would be checkpoint metadata
                                                               # that is always saved by our
                                                               # TrialController
```

I say this is "less magical" because it makes the TrialContext a little less different than the generic api context.

harness/determined/pytorch/samplers.py
katport: Okay. Lets leave it then as is.

katport (approved):
LGTM

ioga (approved):

PR #1939, opened 2021-02-05
chore: Use Ubuntu 20.04 images in CI [DET-4961]

mackrorysd:
## Description

The main motivation here is that new circle CI machine images include a fix so that the lzma module in Python works. It's also good to be moving off of Ubuntu 16.04 as it will be unsupported soon.

This requires unpinning a couple of dependencies but I am unable to find a good reason to pin them. h5py was pinned because of TensorFlow requirements, but that dependency can be expressed directly by Tensorflow for those users who need it. The harness itself doesn't need it. That version of Cloudpickle is known to not work on Python 3.8.


.circleci/config.yml
```diff
@@ -199,7 +199,9 @@ commands:
       - run:
           name: Setup venv
           command: |
-            python3 -m venv /tmp/venv
+            sudo apt-get update
+            sudo apt-get install -y python3-venv
+            /usr/bin/python3 -m venv /tmp/venv
```
mackrorysd, line 204: For context: I don't know why it's not pre-installed in the new images, but `python3 -m venv` will fail and end up creating a Python 2 environment if you don't installed python3-venv first. Also note that this switches us to the system Python (3.8.5 on Ubuntu 20.04) instead of using pyenv. The new images have pyenv pre-install 3.9.x, which causes a lot of problems for our dependencies at this point.

mackrorysd:
Oh builds of the older PyTorch frameworks aren't available on Python 3.8 yet, which is why this broke - we reverted the framework version updates right before I submitted this. Will withdraw until we can move to the new images and stop reinstalling Python every time.

PR #1941, opened 2021-02-08
ci: deploy NGC images as part of release [DET-4910]

shiyuann:
## Description

Push docker images to the NGC repo and also push the helm charts.

1. Build the helm charts for all the branches and all the PRs.
2. Push the release docker images to `nvcr.io/isv-ngc-partner/determined`.
3. Push the release and release candidates helm charts to `helm/ngc.nvidia.com/isv-ngc-partner/determined`.

## Test Plan

1. manually run a test release candidate in the CI.
2. check if the NGC repo has the docker image.
3. check if the NGC helm repo has the charts.
4. install the helm chart from the NGC repo into a k8s cluster.

The steps are:
1. Create a k8s cluster and node pool.
2. Create a secret using [kubectl create secret](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
3. Modify the ``imageRegistry`` field to ``nvcr.io/isv-ngc-partner/determined`` in ``values.yaml``.
4. Modify the ``imagePullSecretName`` field to your secret name in ``values.yaml``.
5. Install the helm chart using ``helm install <name> helm/charts/determined``.

## Commentary (optional)

N/A

## Checklist

### DIFF
--- a/helm/charts/determined/Chart.yaml
+++ b/helm/charts/determined/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: determined
 description: A Helm chart for Determined
-version: 0.4.0
+version: "0.4.0"
 icon: https://github.com/determined-ai/determined/blob/master/determined-logo.png?raw=true
 home: https://github.com/determined-ai/determined.git
 
@@ -9,4 +9,4 @@ home: https://github.com/determined-ai/determined.git
 # a non-release version (e.g., X.Y.Z.dev0) you will have to specify an
 # existing official release version (e.g., X.Y.Z) or specify a commit has
 # that has been publicly published (all commits from master).
-appVersion: 0.14.2.dev0
+appVersion: "0.14.2.dev0"
--- a/helm/charts/determined/templates/master-deployment.yaml
+++ b/helm/charts/determined/templates/master-deployment.yaml
@@ -23,16 +23,16 @@ spec:
       serviceAccount: determined-master-{{ .Release.Name }}
       containers:
       - name: determined-master-{{ .Release.Name }}
-        {{ $image := "determinedai/determined-master" }}
+        {{ $image := "determined-master" }}
         {{- if .Values.enterpriseEdition -}}
-          {{ $image = "determinedai/determined-ee-master" }}
+          {{- $image = "determined-ee-master" -}}
         {{- end -}}
         {{ $tag := (required "A valid Chart.AppVersion entry required!" .Chart.AppVersion) }}
         {{- /* detVersion is used for CI to override the appVersion. */ -}}
         {{- if .Values.detVersion -}}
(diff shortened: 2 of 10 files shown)


stoksc (commented):
would this end up tagging and pushing these images to nvcr.io even on dev/rc branches?

shiyuann:
@stoksc No. it will just push release images.

stoksc (approved):
ah, i see. i had never read those makefiles closely (I see publish is its own thing and release runs goreleaser). cool lgtm.

stoksc:
I think this PR was done when I reviewed and helm would be done separately. :disappear: I'll like @hoanghphan or @dzhu review instead.

shiyuann:
@eecsliu I have changed the helm chart to include the change we talked about.

helm/charts/determined/templates/master-deployment.yaml
```diff
@@ -51,10 +51,10 @@ spec:
             {{- if .Values.masterMemRequest }}
             memory: {{ .Values.masterMemRequest  | quote }}
             {{- end}}
-      {{- if .Values.enterpriseEdition }}
+      {{- if .Values.imagePullSecretName}}
```
eecsliu, line 54: Blocking: This looks ok to me, but I'm concerned about not being able to warn the user if they don't use the imagePullSecret in EE. Is this not a problem, or do we think it's sufficiently easy to diagnose a failed image pull without the warning? Should we add another line of logic to cover this?

eecsliu (commented):
Looks mostly good to me. Just a couple of questions.

helm/charts/determined/Chart.yaml
```diff
@@ -1,12 +1,12 @@
...
+version: "0.4.0"
 icon: https://github.com/determined-ai/determined/blob/master/determined-logo.png?raw=true
 home: https://github.com/determined-ai/determined.git
 
 # appVersion controls the version Determined that is deployed. If using a
 # a non-release version (e.g., X.Y.Z.dev0) you will have to specify an
 # existing official release version (e.g., X.Y.Z) or specify a commit has
 # that has been publicly published (all commits from master).
-appVersion: 0.14.0.dev0
+appVersion: "0.14.0.dev0"
```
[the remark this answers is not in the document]
shiyuann: It is suggested by the charts documentation to use quote since it will make yaml parser not see this as a float number if we only specify `0.14`

helm/charts/determined/templates/master-deployment.yaml
shiyuann: It's very easy to diagnose a failed image pull without a warning. Even if you add the previous check, we can't validate the secret so the check didn't really help a lot.

shiyuann:
This has already passed the tests. I rebased so will just skip the e2e tests.

PR #1942, opened 2021-02-08
fix: add backwards compatability for logs before 0.13.8

stoksc:
## Description
This change tweaks the query used to retrieve trial logs to select the timestamp from the old field `public.trial_logs.message` where it doesn't exist for old logs.

## Test Plan
- [x] checkout 0.13.5, run an experiment and peek that `public.trial_logs` is missing the new fields
- [x] checkout this branch, load those logs in the correct order, scroll up/down work as expected
- [x] explain analyze the query, make sure we don't take a huge hit.
- [x] run `select to_timestamp(substring(convert_from(message, 'UTF-8') from '\[([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z)\]'), 'YYYY-MM-DD hh24:mi:ss') from trial_logs where timestamp is NULL order by trial_id desc;` to test this for every log on latest master.

## Commentary (optional)

### DIFF
--- a/master/internal/db/postgres_trial_logs.go
+++ b/master/internal/db/postgres_trial_logs.go
@@ -36,7 +36,7 @@ SELECT
     END AS message,
     l.agent_id,
     l.container_id,
-	CASE
+    CASE
       WHEN l.timestamp is NOT NULL THEN l.timestamp
       ELSE to_timestamp(
         substring(convert_from(message, 'UTF-8') from '\[([0-9]{4}-[0-9]{2}-[0-9]{2}T.*)\]'),
--- a/master/internal/db/postgres_trial_logs.go
+++ b/master/internal/db/postgres_trial_logs.go
@@ -36,14 +36,20 @@ SELECT
     END AS message,
     l.agent_id,
     l.container_id,
-    l.timestamp,
+	CASE
+      WHEN l.timestamp is NOT NULL THEN l.timestamp
+      ELSE to_timestamp(
+        substring(convert_from(message, 'UTF-8') from '\[([0-9]{4}-[0-9]{2}-[0-9]{2}T.*)\]'),
+        'YYYY-MM-DD hh24:mi:ss'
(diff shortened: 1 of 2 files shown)


master/internal/db/postgres_trial_logs.go
```diff
@@ -36,14 +36,20 @@ SELECT
     END AS message,
     l.agent_id,
     l.container_id,
-    l.timestamp,
+    CASE
+      WHEN l.timestamp is NOT NULL THEN l.timestamp
+      ELSE to_timestamp(
+        substring(convert_from(message, 'UTF-8') from '\[([0-9]{4}-[0-9]{2}-[0-9]{2}T.*)\]'),
```
stoksc, line 42: one sec.. pushed the wrong version of this

determined-dsw:
@stoksc Did you test the version with the new field in it to make sure everything still works?  I'm not sure if that last step in your test plan is doing that.

stoksc:
@determined-dsw we have tests that cover that 🎊 but i'll hit it anyway

determined-dsw:
@stoksc Can you add a release note?

master/internal/db/postgres_trial_logs.go
```diff
@@ -36,14 +36,21 @@ SELECT
     END AS message,
     l.agent_id,
     l.container_id,
-    l.timestamp,
+    CASE
```
dzhu, line 39: non-blocking: This could use `coalesce`, which [I believe](https://stackoverflow.com/questions/505971/coalesce-guaranteed-to-short-circuit) should evaluate the same way (i.e., not evaluate the second argument any more eagerly).

dzhu (approved):
Could you also check that a log that doesn't match the pattern yields NULL rather than causing a query error? I'm _pretty_ sure it should, but it would be good to check.

PR #1944, opened 2021-02-09
fix: make trial log timestamp filters backwards compatible 

stoksc:
## Description
This change makes the query to gather trial logs with timestamp filters backwards compatible with logs created before 0.13.8.

## Test Plan
- [x] built 0.13.6 from scratch, ran an experiment with a ton of logs, checked out my branch and built it from scratch and checked and all the logs, start at the bottom and scroll up, start at top and scroll down, everything works.

## Commentary (optional)
Bit of a hack, but it's necessary


stoksc:
need to rebase this off master.. one sec

dzhu (approved):

PR #1947, opened 2021-02-09
chore: fix error on empty metric in metric-snapshot endpoint [DET-4923]

mackrorysd:
## Description

This sometimes is an issue if a DB row exists but not a metric value at the moment the endpoint does a query.

## Test Plan

Re-ran all existing tests. Reproducing original error is hard (root cause is non-deterministic), but this same fix was verified on a similar instance in a different call.


stoksc (approved):
lgtm

PR #1948, opened 2021-02-10
feat: allow retaining CloudWatch log groups of det-deploy clusters

dzhu:
## Description

`det-deploy aws up` now supports two new options: `--retain-log-group`
and `--log-group-prefix`. The former sets the CloudWatch log group that
is created for the stack to not be deleted when the rest of the stack
is, which may be useful in some situations. The latter sets the prefix
that is applied to the stack name to generate the log group name
(default "determined", leading to the same behavior as before).

Because CloudFormation's templating is terrible, the implementation
looks a bit awkward. The `DeletionPolicy` field is required to be a
literal, so it is not possible to use an expression to set it
conditionally. Instead, we have to define two separate resources, one
retained and one not, each of which is created conditionally. As far as
I can tell, those conditions can't be written inline but instead have to
be defined in a separate section and referred to by name, which also
means we need one for each case. Having two separate resources in turn
means that we can't refer to the log group by its logical ID later on
when we need the log group name; instead, we have to insert the
expression used to compute the log group name everywhere we need
it (since it also seems to be impossible to define derived variables in
the template without doing something crazy like going through AWS
Lambda or storing config snippets in S3).

## Test Plan

- [x] start stacks with and without `--retain-log-group`, check that log group is deleted or not
- [x] run with `--log-group-prefix`

## Commentary (optional)

CloudFormation makes this so terrible. I'm honestly not 100% sure this is worth doing, given the hoops that we need to jump through; accordingly, I've only updated the simple template so far so we can evaluate first.

The prefix option doesn't strictly have to go along with the retention one, but I figure it'll be useful to allow a bit more control over where logs go if there are likely to be more of them (for CI-generated logs in particular).

### DIFF
--- a/deploy/determined_deploy/aws/templates/simple.yaml
+++ b/deploy/determined_deploy/aws/templates/simple.yaml
@@ -141,7 +141,7 @@ Parameters:
     Type: String
     Description: Whether preemption is enabled (only supported for priority scheduler).
     Default: false
-  
+
   CpuEnvImage:
     Type: String
     Description: Docker image for CPU tasks
@@ -151,6 +151,20 @@ Parameters:
     Type: String
     Description: Docker image for GPU tasks
     Default: ""
+
+  LogGroupPrefix:
+    Type: String
+    Description: Prefix for output CloudWatch log group (the full log group will be "/<prefix>/<stack>")
+    Default: determined
+
+  RetainLogGroup:
+    Type: String
+    Description: Whether to retain CloudWatch log group after the stack is deleted
+    Default: false
+
+Conditions:
+  RetainLogGroup: !Equals [!Ref RetainLogGroup, true]
+  DeleteLogGroup: !Not [!Equals [!Ref RetainLogGroup, true]]
 
 Resources:
   CheckpointBucket:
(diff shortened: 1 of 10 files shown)


deploy/determined_deploy/aws/templates/simple.yaml
```diff
@@ -320,10 +334,19 @@ Resources:
         MinCapacity: 2
         MaxCapacity: 8
 
-  LogGroup:
+  RetainedLogGroup:
```
stoksc, line 337: this is a lot of effort for this.. wow.

stoksc (approved):
lgtm, but if it's going to be in the `--help` text , should we drop this into the other templates too 🤷 . Probably, no, doesn't matter much.

hoanghphan (approved):

dzhu:
> lgtm, but if it's going to be in the `--help` text , should we drop this into the other templates too :shrug:. Probably, no, doesn't matter much.

Well, I was planning to do that and just wanted to get this up to get opinions on whether the skulduggery required was worth it at all, but if you both like it as-is, I suppose I can land it and defer the rest until someone complains. :zany_face: Though I think I will extend the prefix option to the others and add a note to the help text.

stoksc:
> Well, I was planning to do that and just wanted to get this up to get opinions on whether the skulduggery required was worth it at all, but if you both like it as-is, I suppose I can land it and defer the rest until someone complains. 🤪 Though I think I will extend the prefix option to the others and add a note to the help text.

sgtm.

PR #1949, opened 2021-02-10
chore: check for closed connections before sending [DET-4829]

mackrorysd:
## Description

As suggested by Bradley in a previous review, the pattern of periodically checking for a dead context and then returning nil allows us to gracefully terminate the loop when the client connection is closed. This was previously done after sleeping before beginning another iteration, but doing it right before actually sending the response drops the probability of a mess in the logs for no good reason about as low as we can get it.

Also refactoring more of this into it's own function, since it now appears 8 times in this file.

I don't like the name I chose for duplicate import: "grpcroot". Trying to think of a better alternative.

## Test Plan

test_metrics.py has decent coverage of these functions.

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -10,7 +10,6 @@ import (
 	"strings"
 	"time"
 
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -18,7 +17,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/determined-ai/determined/master/internal/db"
-	internalGrpc "github.com/determined-ai/determined/master/internal/grpc"
+	"github.com/determined-ai/determined/master/internal/grpcutil"
 	"github.com/determined-ai/determined/master/internal/hpimportance"
 	"github.com/determined-ai/determined/master/internal/lttb"
 	"github.com/determined-ai/determined/master/pkg/actor"
@@ -515,7 +514,7 @@ func (a *apiServer) CreateExperiment(
 		return &apiv1.CreateExperimentResponse{}, nil
 	}
 
-	user, _, err := internalGrpc.GetUser(ctx, a.m.db)
+	user, _, err := grpcutil.GetUser(ctx, a.m.db)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to get the user: %s", err)
 	}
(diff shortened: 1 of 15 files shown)


master/internal/api_experiment.go
```diff
@@ -537,6 +538,11 @@ func (a *apiServer) CreateExperiment(
 
 var defaultMetricsStreamPeriod = 30 * time.Second
 
+func connectionIsClosed(resp grpc.ServerStream) bool {
```
stoksc, line 541: if you dont mind, change my stuff in `api_trials.go` to use this too?

master/internal/api_experiment.go
```diff
@@ -10,14 +10,15 @@ import (
...
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/pkg/errors"
 
 	"github.com/determined-ai/determined/master/internal/db"
-	"github.com/determined-ai/determined/master/internal/grpc"
+	internalGrpc "github.com/determined-ai/determined/master/internal/grpc"
```
stoksc, line 21: i'd be ok with renaming this package `grpcutil` or something

stoksc (approved):
lgtm

master/internal/api_experiment.go
stoksc: it's just a lot semantically clearer

mackrorysd: For sure

master/internal/api_experiment.go
mackrorysd: Yeah I think that's good, especially if we're gonna import the google.golang.org/grpc into api_trials now too. I see even our grpc package imports the top-level grpc package from upstream, so that conflict is all over.

master/internal/api_experiment.go
mackrorysd: So I added a similar implementation to logs.go - there ended up being a cyclic dependency if I reused code from (what is now called) grpcutil and this didn't seem worth starting a new generic "util" package. That code also just has the context, not the original GRPC stream, so in the original case we can make the code just a little cleaner too.

Sound good?

master/internal/api_experiment.go
mackrorysd: Since I moved the original version of connectionIsClosed to grpcutil, there's no longer the immediate need to rename one of the grpc pacakges, but I went ahead with the rename: there's a bunch of places where we import internal/grpc AND child packages of google.golang.org/grpc. It's confusion waiting to happen...

mackrorysd:
Overriding Linter. See https://github.com/determined-ai/determined/pull/1961 and https://github.com/determined-ai/determined/pull/1957

PR #1956, opened 2021-02-11
chore: add json-schema-based struct definitons in Go and Python

rb-determined-ai:
## Commentary

All code is still dead code right now.

But the full test suite passes in both Go and Python.

This PR won't pass tests yet because the results of `make -C master gen` have not yet been committed... "I didn't want the PR to be too big."

But seriously, I can break this up into each of the 4 individual commits if you would like, or leave it as-is.  Just depends on whether you like seeing the big picture more or small PRs more.


rb-determined-ai:
Closing this PR in favor of smaller PRs.  First of those is #1957.

PR #1958, opened 2021-02-11
fix: add state to task endpoints [DET-4987]

hamidzr:
## Description

- update tasks api to return a top-level state
- share the logic with the old endpoint

## Checklist

- [x] add the remaining commands, shells, and tensorboards

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -352,7 +352,8 @@ func (c *command) readinessChecksPass(ctx *actor.Context, log sproto.ContainerLo
 	return len(c.readinessChecks) == 0
 }
 
-// State returns the command's state. This mirros the container state if available.
+// State returns the command's state. This mirros the associated container's state
+// if available.
 func (c *command) State() State {
 	state := Pending
 	switch {
@@ -396,6 +397,7 @@ func (c *command) toNotebook(ctx *actor.Context) (*notebookv1.Notebook, error) {
 func (c *command) toCommand(ctx *actor.Context) *commandv1.Command {
 	return &commandv1.Command{
 		Id:           ctx.Self().Address().Local(),
+		State:        c.State().Proto(),
 		Description:  c.config.Description,
 		Container:    c.container.Proto(),
 		StartTime:    protoutils.ToTimestamp(ctx.Self().RegisteredTime()),
@@ -407,6 +409,7 @@ func (c *command) toCommand(ctx *actor.Context) *commandv1.Command {
 func (c *command) toShell(ctx *actor.Context) *shellv1.Shell {
 	return &shellv1.Shell{
 		Id:           ctx.Self().Address().Local(),
+		State:        c.State().Proto(),
 		Description:  c.config.Description,
 		StartTime:    protoutils.ToTimestamp(ctx.Self().RegisteredTime()),
 		Container:    c.container.Proto(),
@@ -428,6 +431,7 @@ func (c *command) toTensorboard(ctx *actor.Context) *tensorboardv1.Tensorboard {
 	}
 	return &tensorboardv1.Tensorboard{
 		Id:             ctx.Self().Address().Local(),
+		State:          c.State().Proto(),
 		Description:    c.config.Description,
 		StartTime:      protoutils.ToTimestamp(ctx.Self().RegisteredTime()),
 		Container:      c.container.Proto(),
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -352,13 +352,37 @@ func (c *command) readinessChecksPass(ctx *actor.Context, log sproto.ContainerLo
 	return len(c.readinessChecks) == 0
 }
 
+func (c *command) getState() State {
+	state := Pending
+	switch {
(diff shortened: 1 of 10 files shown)


proto/src/determined/api/v1/task.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
```
stoksc, line 1: Do we need this?

master/internal/command/command.go
```diff
@@ -352,13 +352,37 @@ func (c *command) readinessChecksPass(ctx *actor.Context, log sproto.ContainerLo
...
+
 func (c *command) toNotebook(ctx *actor.Context) (*notebookv1.Notebook, error) {
 	serviceAddress, err := generateServiceAddress(string(c.taskID))
 	if err != nil {
 		return nil, errors.Wrapf(err, "generating service address for %s", c.taskID)
 	}
+
 	return &notebookv1.Notebook{
 		Id:             ctx.Self().Address().Local(),
+		State:          c.getState().Proto(),
```
stoksc, line 385: Why are we not adding this too for other commands (regular commands, shells, etc..)?

stoksc (commented):
lgtm, just had some questions

master/internal/command/command.go
```diff
@@ -352,13 +352,37 @@ func (c *command) readinessChecksPass(ctx *actor.Context, log sproto.ContainerLo
 	return len(c.readinessChecks) == 0
 }
 
+func (c *command) getState() State {
```
stoksc, line 355: I'd just make this `State()` and doc it. Meant to leave this the first time around oops.

master/internal/command/command.go
hamidzr: I will (in this pr). The rest are gonna be very similar and I wanted to confirm one set before applying it to the rest.

proto/src/determined/api/v1/task.proto
hamidzr: I think so

> The first line of the file specifies that you're using proto3 syntax: if you don't do this the protocol buffer compiler will assume you are using proto2. This must be the first non-empty, non-comment line of the file.

master/internal/command/command.go
hamidzr: although I now realize the changes were so small I should probably added it https://github.com/determined-ai/determined/pull/1958/commits/e0d4b20d7706eaa1a66d3a3806486b5ceebbe887

proto/src/determined/api/v1/task.proto
stoksc: I meant the file at all

proto/src/determined/task/v1/task.proto
```diff
@@ -0,0 +1,24 @@
...
+  // The task state is unknown.
+  STATE_UNSPECIFIED = 0;
+  // The task is pending assignment.
+  STATE_PENDING = 1;
+  // The task has been assigned to an agent but has not started yet.
+  STATE_ASSIGNED = 2;
+  // The task's base image is being pulled from the Docker registry.
+  STATE_PULLING = 3;
+  // The image has been built and the task is being started, but the
+  // service in the task is not ready yet.
```
stoksc, line 17:
```suggestion
  // The image has been pulled and the task is being started, but the task is not ready yet.
```

proto/src/determined/task/v1/task.proto
```diff
@@ -0,0 +1,24 @@
...
+  STATE_STARTING = 4;
+  // The service in the task is able to accept requests.
```
stoksc, line 19:
```suggestion
  // The service in the task is running.
```
able to accept requests is api-y, most of our stuff isn't accepting requests

proto/src/determined/task/v1/task.proto
```diff
@@ -0,0 +1,24 @@
...
+  STATE_RUNNING = 5;
+  // The task has completely exited or the task has been aborted prior
+  // to getting assigned.
```
stoksc, line 22:
```suggestion
  // The task has exited or has been aborted.
```

master/internal/command/state.go
```diff
@@ -0,0 +1,50 @@
...
+	Pending State = "PENDING"
+	// Assigned state denotes that the command has been assigned to an agent but has not started
+	// yet.
+	Assigned State = "ASSIGNED"
+	// Pulling state denotes that the command's base image is being pulled from the Docker registry.
+	Pulling State = "PULLING"
+	// Starting state denotes that the image has been built and the command is being started, but the
+	// service in the command is not ready yet.
+	Starting State = "STARTING"
+	// Running state denotes that the service in the command is able to accept requests.
```
stoksc, line 25: same comments as the proto

stoksc (approved):
lgtm

proto/src/determined/api/v1/task.proto
hamidzr: aah I like the idea of having it separate otherwise we'd put in one of the sub tasks and importing from those is a bit more awkward. do you have a suggestion for where to put it?

PR #1959, opened 2021-02-11
feat: pytorch lightning adapter support

hamidzr:
## Description

- fix issue with hvd aggregation_frequency check in step_optimizer 05560c5
- add bactch_idx arg to evaluate_batch in pytrochTrial
- add PyTorch Lightning Adapter for Pytroch Trial
- add 2 usage examples

## Commentary (optional)

TODOs left
- [x] rename ptl to pl (pytorch_lightning's choice)
- [x] serialization issue when returning nested dict of metrics: this is inherited from PytorchTrial and expected
- [x] set the env image for examples once they're published on docker-hub
- [ ] remove dev tools from example

## Checklist


harness/determined/pytorch/_pytorch_context.py
```diff
@@ -440,6 +440,7 @@ def step_optimizer(
         optimizer: torch.optim.Optimizer,  # type: ignore
         clip_grads: Optional[Callable[[Iterator], None]] = None,
         auto_zero_grads: bool = True,
+        on_before_zero_grad: Optional[Callable[[torch.optim.Optimizer], None]] = None,
```
hamidzr, line 443: this wouldn't be a breaking change would it?

examples/gan/_gan_mnist_pl/pl.py
```diff
@@ -0,0 +1,262 @@
...
+            # put on GPU because we created this tensor inside training_loop
+            valid = torch.ones(imgs.size(0), 1)
+            valid = valid.type_as(imgs)
+
+            # adversarial loss is binary cross-entropy
+            g_loss = self.adversarial_loss(self.discriminator(self(z)), valid)
+            tqdm_dict = {'g_loss': g_loss}
+            output = OrderedDict({
+                'loss': g_loss,
+                # 'progress_bar': tqdm_dict, #  FIXME unserializable object torch.Tensor
```
hamidzr, line 200:
```
 File "/run/determined/pythonuserbase/lib/python3.6/site-packages/determined/util.py", line 150, in json_serializer
<[]>
raise TypeError("Unserializable object {} of type {}".format(obj, type(obj)))
```

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+
+        optimizers = [self.context.wrap_optimizer(opt) for opt in optimizers]
+        lr_schedulers = [lightning_scheduler_dict_to_det(lrs) for lrs in lr_scheduler_dicts]
+        return optimizers, lr_schedulers
+
+    def train_batch(
+        self, batch: TorchData, epoch_idx: int, batch_idx: int
+    ) -> Union[torch.Tensor, Dict[str, Any]]:
+        self.lm.on_train_batch_start(batch, batch_idx, dataloader_idx=0)
+        type(self.lm).global_step = mock.PropertyMock(return_value=batch_idx)
```
shiyuann, line 162: I tried `type(self.lm).global_step=batch_idx`. It worked.
 maybe we don't need this mock package?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+
+        opt_metrics = []
+
+        _optimizers = self.optimizers
+
+        def get_optimizers(self, *args, **kwargs) -> List[Optimizer]:
+            return _optimizers
+
+        for opt_idx, opt in enumerate(self.optimizers):
+            with monkey_patch.monkey_patch(self.lm, "optimizers", get_optimizers):
```
shiyuann, line 172: You can directly use lambda here.

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+                self.lm.toggle_optimizer(opt, opt_idx)
```
shiyuann, line 173: I checked the source code of `toggle_optimizer`. It doesn't use `.optimizers` at all. Why do you need to monkey-patch here?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+            metrics = self.lm.training_step(batch, batch_idx, optimizer_idx=opt_idx)
```
shiyuann, line 174: Is it intentional to only make `optimizer_idx` keyword argument?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+            elif not isinstance(metrics, dict):
+                metrics = {"loss": metrics}
+
+            self.context.backward(metrics["loss"])
+            self.lm.on_after_backward()
+            self.context.step_optimizer(opt, on_before_zero_grad=self.lm.on_before_zero_grad)
+
+            opt_metrics.append(metrics)
+            with monkey_patch.monkey_patch(self.lm, "optimizers", get_optimizers):
+                self.lm.untoggle_optimizer(opt_idx)
```
shiyuann, line 189: I didn't see `LightningModule` had such member function `untoggle_optimizer`. Which version are you on?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+
+        agg_metrics = {}
+        for opt_idx, rv in enumerate(opt_metrics):
+            for k, v in rv.items():
+                agg_metrics[f"opt{opt_idx}_{k}"] = v
```
shiyuann, line 194: I think it's wrong to add prefixes here since users are the ones to control the names. If they specify the same names for different optimizers, it should be wrong unless PytorchLightning itself implements in this way

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+        return agg_metrics
+
+    def evaluate_batch(self, batch: TorchData) -> Dict[str, Any]:
+        batch_idx = self.context.current_batch()
+        type(self.lm).global_step = mock.PropertyMock(return_value=batch_idx)
```
shiyuann, line 199: Why do you need to set gloabl_step in validation? Is it a behavior of Pytorch Lightning?

examples/gan/_gan_mnist_pl/model_def.py
```diff
@@ -0,0 +1,30 @@
...
+        self.dm = pl.MNISTDataModule(batch_size=context.get_global_batch_size())
+        channels, width, height = self.dm.size()
+        lm = pl.GAN(channels, width, height,
+                    batch_size=context.get_global_batch_size(),
+                    lr=context.get_hparam('lr'),
+                    b1=context.get_hparam('b1'),
+                    b2=context.get_hparam('b2'),
+        )
+
+        context.wrap_model(lm.generator.model)
```
shiyuann, line 16: shouldn't this be called within `PTLAdapter.__init__`?

harness/determined/errors.py
```diff
@@ -72,3 +72,12 @@ class SkipWorkloadException(Exception):
     """
 
     pass
+
+
+class UnsupportedException(Exception):
```
shiyuann, line 77: `UnsupportedException` is a pretty vague name. I'd use 
```
class InvalidModelException(InvalidExperimentException):
```

harness/determined/pytorch/_lightning/README.md
```diff
@@ -0,0 +1,168 @@
+# PTL Adapter
```
shiyuann, line 1: Is this `README.md` ready to be merged? It looks more like a development note rather than being useful to users.

shiyuann (commented):
Is this PR in a ready-to-be-merged state? I saw two examples, which I assume will only leave the GAN in the codebase. I also saw a `README.md` that seems not to be for users. Otherwise, the PR looks fine overall. There are places that are not consistent with our coding style in other harness code.

harness/determined/pytorch/_lightning/adapter.py
shiyuann: Actually, I am wrong. You need to add prefixes.

harness/determined/pytorch/_lightning/README.md
hamidzr: no, that's right. it's here mostly for reviewers benefit I'll take it out

examples/gan/_gan_mnist_pl/model_def.py
hamidzr: I was thinking we leave this out for the user to call for multi model/optimizer algorithms like GANs

shiyuann: You could actually just call wrap_model on the LightningModule

harness/determined/pytorch/_lightning/adapter.py
hamidzr: I think we can probably keep the user's names unless they collide with each other. always adding prefixes was an easier first approach.

examples/computer_vision/_mnist_ptl/.detignore
```diff
@@ -0,0 +1 @@
+lightning_logs/
```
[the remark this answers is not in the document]
hamidzr: if we don't ignore it they'll get uploaded to determined with the experiment and show up as dirty files in git (at least for how the experiment is set up right now) do you see any downsides for having it?

examples/gan/_gan_mnist_pl/model_def.py
```diff
@@ -0,0 +1,30 @@
+from determined.pytorch import PyTorchTrial, PyTorchTrialContext, DataLoader as DetDataLoader
```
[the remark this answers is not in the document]
hamidzr: I found just `DataLoader`  to be a bit confusing for myself but that's subjective. I'll update it to be consistent.

examples/gan/_gan_mnist_pl/model_def.py
```diff
@@ -0,0 +1,30 @@
...
+from determined.pytorch._lightning import PTLAdapter
+import pl
+
+class GanTrial(PTLAdapter):
```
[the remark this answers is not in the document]
hamidzr: I might be missing something here wdy mean? it is `GanTrial` already

harness/determined/pytorch/_lightning/__init__.py
```diff
@@ -0,0 +1,4 @@
+from determined.pytorch._lightning.adapter import (
+    HyperparamsProvider,
```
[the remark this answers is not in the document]
hamidzr: I tried two different ways of passing the context down in the two examples the gan one doesn't use this but the other one does. IMO it's helpful for usecase 1 but the GAN example showcases PL's use of `.hparams` and `save_hyperparameters` which if the user opts to use they wouldn't need this type definition.

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    Checks function for presence of a positional or kw argument.
+    """
+    # QUESTION what's the most robust way of detecting if their fn expectes a param or not?
+    # they could have it defined as a postional or kwarg. eg dataloader_idx in validation_step
+    kw_present = name in inspect.getfullargspec(fn)[0]
+    return kw_present
+
+
+def check_compat(lm: ptl.LightningModule):
```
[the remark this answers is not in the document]
hamidzr: yes to check the compatiblity of LM with PTLAdapter. any name suggestions?

harness/determined/pytorch/_lightning/adapter.py
hamidzr: you're right we don't need to update it
`ag 'global_step.* = ' `

hamidzr: I was getting the batch index here wrong. it should be batch index scoped to the epoch. I might need to add batch_idx as an arg to `evaluate_batch` to pass it through

harness/determined/pytorch/_lightning/adapter.py
hamidzr: 1.1.8. it looks like they added it in 1.1.6

hamidzr: what version did we use in prd?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+        super().__init__(context)
+        check_compat(lightning_module)
+        self.lm = lightning_module
+        self.context = context
+
+        optimizers, lr_schedulers = self.setup_optimizers_schedulers(context)
+        self.optimizers = optimizers
+
+        # user should merge these with their own callbacks
+        self.build_callbacks = lambda: build_callbacks(lightning_module, self.context)
```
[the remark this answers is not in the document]
hamidzr: good point. I updated it to be how pytorch trial sets it up. would it be reasonable to expect the users to call `super().build_callbacks` and merge the output with their callbacks. or should I try to do that for them maybe during `__init__`?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
+HyperparamsProvider = Callable[[str], Any]
+
+
+def bail(msg: str = ""):
+    msg = f"Not supported in PTL adapter: {msg}"
+    raise UnsupportedException(msg)
+
+
+def expects_param(fn: Callable[..., Any], pos: Optional[int], name: str) -> bool:
```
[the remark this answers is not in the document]
hamidzr: 👍🏼 I updated it to has_param. do you have any thoughts suggestions on the question I left here in the code?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+
+class PTLUtils(TrainerOptimizersMixin):
+    pass
+
+
+ptl_utils = PTLUtils()
+
+
+class PTLAdapter(PyTorchTrial):
+    context: PyTorchTrialContext
```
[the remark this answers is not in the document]
hamidzr: I thought this is a hint for the typesystem but I might be wrong.

harness/determined/pytorch/_callback.py
```diff
@@ -72,3 +72,33 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         Load the state of this using the deserialized ``state_dict``.
         """
         pass
+
+    def on_before_zero_grad(self) -> None:
```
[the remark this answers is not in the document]
hamidzr: leftover. I removed it

harness/determined/pytorch/_pytorch_context.py
hamidzr: hmm why do we have the auto_zero_grads param then? 

also separately isn't something wrong with this check? 
```
        check.true(
            auto_zero_grads or self.hvd_config.aggregation_frequency > 1,
            "if optimizations.aggregation_frequency is larger than 1, "
            "you can only set auto_zero_grads to be true. ",
        )
```

examples/computer_vision/_mnist_ptl/.gitignore
```diff
@@ -0,0 +1 @@
+lightning_logs/
```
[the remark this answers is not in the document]
hamidzr: let's chat on the other comment. https://github.com/determined-ai/determined/pull/1959#discussion_r579528864

examples/gan/_gan_mnist_pl/.detignore
```diff
@@ -0,0 +1 @@
+lightning_logs/
```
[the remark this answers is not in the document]
hamidzr: let's chat on the other comment. #1959 (comment)

examples/computer_vision/_mnist_ptl/ptl.py
```diff
@@ -0,0 +1,105 @@
+import torch
```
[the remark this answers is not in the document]
hamidzr: hmm why not both? this one is much simpler could it be useful?

harness/determined/pytorch/_lightning/__init__.py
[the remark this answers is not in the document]
hamidzr: I'll just keep the more native pl usage which also means removing the `HyperparametersProvider`

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -20,6 +20,16 @@
     pass
 
 
+def iterate_batches(dataset_len: int, start: int, end: int) -> Iterable[Tuple[int, range]]:
+    """
+    A helper to iterate over a dataset using start and end indices.
+    """
+    for epoch_idx in range(start // dataset_len, end // dataset_len + 1):
```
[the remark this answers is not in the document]
hamidzr: we can't really use `self` considering this being defined as a utility function we could maybe add this to the trial controller but instead, IMO it might be better to name that division to make it more explicit? I update it.

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+    lm: ptl.LightningModule
```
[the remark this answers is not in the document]
hamidzr: same as https://github.com/determined-ai/determined/pull/1959#discussion_r579552501 let's chat there

examples/gan/_gan_mnist_pl/model_def.py
hamidzr: wouldn't wrap model need to be called once per nn module?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,209 @@
...
+                'scheduler': None,
+                'name': None,  # no custom name
+                'interval': 'epoch',  # after epoch is over
+                'frequency': 1,  # every epoch/batch
+                'reduce_on_plateau': False,  # most often not ReduceLROnPlateau scheduler
+                'monitor': monitor,  # value to monitor for ReduceLROnPlateau
+                'strict': True,  # enforce that the monitor exists for ReduceLROnPlateau
+            }
+            """
+            # TODO support custom frequencies with the manual step.
```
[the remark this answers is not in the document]
hamidzr: DET-5021

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -242,42 +252,51 @@ def _train_for_step(
...
+                for name, metric in tr_metrics.items():
+                    # Convert PyTorch metric values to NumPy, so that
+                    # `det.util.encode_json` handles them properly without
+                    # needing a dependency on PyTorch.
+                    if isinstance(metric, torch.Tensor):
+                        metric = metric.cpu().detach().numpy()
+                    tr_metrics[name] = metric
+
+                epoch_per_batch_metrics.append(tr_metrics)
+            for callback in self.callbacks.values():
```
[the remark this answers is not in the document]
hamidzr: good catch. I assume you mean `on_train_epoch_end`?

examples/gan/_gan_mnist_pl/pl.py
hamidzr: I found in pytorch trial we don't support returning nested dictionaries in metrics yet.

harness/determined/pytorch/_lightning/adapter.py
hamidzr: I didn't actually give it much thought. is there a best practice around this that I should follow?

harness/determined/pytorch/_lightning/adapter.py
hamidzr: 1.1.8. it looks like they added it in 1.1.6

harness/determined/pytorch/_lightning/adapter.py
hamidzr: probably a version mismatch? it uses it here `0b7f5a88a pytorch_lightning/core/lightning.py          1183 (ananthsub               2021-02-04 14:50:57 -0800 1183)         for opt in self.optimizers(use_pl_optimizer=False):`

harness/determined/pytorch/_lightning/adapter.py
hamidzr: I added this argument and a patch for it for backward compat. can you take a look, please?

examples/gan/_gan_mnist_pl/model_def.py
shiyuann: yeah all the submodules will be called on automatically. LightningModule itself is a nn.Module.

hamidzr: I moved this into `__PLAdapter.__init__`  we wrap the passed in lightning module for the user and relying on the pytorch's behavior on sub model for nn.Module. If the user has separate models defined that are not part of the lM they'll have to call wrap_model on it themselves in `model_def`. does that sound right?

hamidzr:
thanks for the review. I addressed the comments. For some of them I left more comments that'd need your attention.

I'm working on addressing the mypy lint failures. some of them are a bit unclear for example:
`determined/pytorch/_pytorch_context.py:443: error: Name 'torch.optim.Optimizer' is not defined` where it has no problem finding it for two lines above it.

examples/computer_vision/_mnist_ptl/.detignore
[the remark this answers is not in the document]
shiyuann: The downside is that it introduces an unnecessary file to the examples. You should just set the logs to use `/tmp`.

examples/gan/_gan_mnist_pl/model_def.py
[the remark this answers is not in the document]
shiyuann: Sorry. I mistyped. It should be `GANTrial`. `GAN` should use all capital characters.

harness/determined/pytorch/_pytorch_trial.py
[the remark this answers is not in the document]
shiyuann: There are two issues with this function. One issue is that this function is introducing the dependency of our training loop on the concept of the epoch, which is dependent on the length of the training loader. If the training loader returns a wrong number, the current training loop still works except for the epoch-related callbacks. The other issue is that this function is duplicate logic from `PyTorchTrialController.get_epoch_idx`. I meant to just use `self.context. get_epoch_idx(idx)` directly in the loop if you really need to use it. Otherwise, I don't see any benefits to introduce this complexity of writing two-level loops.

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
shiyuann: Okay. It looks fine.

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
shiyuann: `has_param` looks like it should be in `determined/util.py` together with `is_overridden`. What question? I didn't see it.

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
shiyuann: Maybe `check_compatability`?

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
shiyuann: Yes. I think it's fine to let users call `super().build_callbacks` as long as we are clear of it in the docstring.

harness/determined/pytorch/_lightning/adapter.py
shiyuann: I mean: why not just make all the arguments keyword arguments?

harness/determined/pytorch/_lightning/adapter.py
shiyuann: >  I might need to add batch_idx as an arg to evaluate_batch to pass it through

You should instead add a function in the context object `get_cur_batch_idx` rather than change the arguments. We are moving towards using the context object as many times as possible.

examples/gan/_gan_mnist_pl/model_def.py
shiyuann: > If the user has separate models defined that are not part of the lM they'll have to call wrap_model on it themselves in model_def. does that sound right?

Yes if they don't define the models to be a member of the lightning module.

shiyuann:
@hamidzr It seems using `# type: ignore` is fine.

examples/gan/_gan_mnist_pl/model_def.py
```diff
@@ -0,0 +1,27 @@
+from determined.pytorch import PyTorchTrial, PyTorchTrialContext, DataLoader
+from determined.pytorch._lightning import PLAdapter
+import pl
```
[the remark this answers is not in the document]
shiyuann: Can you rename the file `pl.py` to be something else? maybe `gan.py`. It's a bit confusing.

examples/gan/_gan_mnist_pl/pl.py
```diff
@@ -0,0 +1,253 @@
+# -*- coding: utf-8 -*-
```
shiyuann, line 1: @rb-determined-ai If we include the original file from Pytorch Lightning, do we need to include their copyright info here?

harness/determined/errors.py
shiyuann: I didn't see this file updated.

harness/determined/monkey_patch.py
```diff
@@ -30,3 +31,15 @@ def inner(*args, **kwargs):  # type: ignore
         setattr(obj, attr, inner)
 
     return wrap
+
+
+def has_param(fn: Callable[..., Any], name: str, pos: Optional[int] = None) -> bool:
```
shiyuann, line 36: It seems that it should exist in `determined/util.py`

harness/determined/pytorch/_callback.py
```diff
@@ -72,3 +72,27 @@ def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
...
+        """
+        pass
+
+    def on_train_epoch_end(self, output: List[Any]) -> None:
+        """
+        Run after a new training epoch has finished
+        """
+        pass
+
+    def on_validation_epoch_start(self) -> None:
```
shiyuann, line 88: Use `on_validate_epoch_start` to be consistent with the other names?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+from determined.errors import UnsupportedException
+from determined.monkey_patch import has_param, monkey_patch
+from determined.pytorch import LRScheduler, PyTorchCallback, PyTorchTrial, PyTorchTrialContext
+from determined_common import check
+
+TorchData = Union[Dict[str, torch.Tensor], Sequence[torch.Tensor], torch.Tensor]
+
+
+def check_compat(lm: pl.LightningModule) -> None:
+    prefix = "Unsupported usage in PLAdapter: "
```
shiyuann, line 20:
```suggestion
    prefix = "Unsupported usage in PTLAdapter: "
```

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+
+        def on_train_epoch_start() -> None:
+            if context._current_batch_idx is not None:
+                type(lm).current_epoch = context.current_train_epoch()
+            lm.on_train_epoch_start()
+
+        def on_validation_epoch_start() -> None:
+            lm.on_validation_epoch_start()
+
+        cb.on_validation_epoch_start = on_validation_epoch_start
```
shiyuann, line 91: Dynamically changing a member field of a Python object is generally a bad practice, which makes code hard to debug. Why don't you just define an inner class:

```
class PTLAdapterCallback(PyTorchCallback):
   def ...():
      ...
```

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+                'scheduler': None,
+                'name': None,  # no custom name
+                'interval': 'epoch',  # after epoch is over
+                'frequency': 1,  # every epoch/batch
+                'reduce_on_plateau': False,  # most often not ReduceLROnPlateau scheduler
+                'monitor': monitor,  # value to monitor for ReduceLROnPlateau
+                'strict': True,  # enforce that the monitor exists for ReduceLROnPlateau
+            }
+            """
+            # TODO(DET-5021) support custom frequencies with the manual step.
```
shiyuann, line 126: What are custom frequncies?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+
+        optimizers = [self.context.wrap_optimizer(opt) for opt in optimizers]
+        lr_schedulers = [lightning_scheduler_dict_to_det(lrs) for lrs in lr_scheduler_dicts]
+        return optimizers, lr_schedulers
+
+    def train_batch(
+        self, batch: TorchData, epoch_idx: int, batch_idx: int
+    ) -> Union[torch.Tensor, Dict[str, Any]]:
+        self.lm.on_train_batch_start(batch, batch_idx, dataloader_idx=0)
+        type(self.lm).global_step = batch_idx
```
shiyuann, line 142: Is it set after calling `on_train_batch_start` in Lightning?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+
+        opt_metrics = []
+
+        for opt_idx, opt in enumerate(self.optimizers):
+            with monkey_patch(self.lm, "optimizers", lambda *args, **kwargs: self.optimizers):
+                self.lm.toggle_optimizer(opt, opt_idx)
+            metrics = self.lm.training_step(batch, batch_idx, optimizer_idx=opt_idx)
+
+            self.lm.on_train_batch_end(metrics, batch, batch_idx, dataloader_idx=0)
```
shiyuann, line 151: Should this be called after the backward pass?

harness/determined/pytorch/_pytorch_context.py
shiyuann: blocking: Actually, I take the word back. We shouldn't add this callback. The Pytorch interface is meant to let users call anything they want directly in `train_batch` rather than using callbacks. You should remove this callback.

`auto_zero_grads` is just for users who want to call zero grading themselves.

>  also separately isn't something wrong with this check?

Do you mean the check itself is wrong or the message?

harness/determined/errors.py
hamidzr: I must have missed it. updated.

examples/computer_vision/_mnist_pl/model_def.py
```diff
@@ -0,0 +1,31 @@
...
+In the `__init__` method, the model and optimizer are wrapped with `wrap_model`
+and `wrap_optimizer`. This model is single-input and single-output.
+
+The methods `train_batch` and `evaluate_batch` define the forward pass
+for training and evaluation respectively.
+"""
+
+from determined.pytorch import PyTorchTrial, PyTorchTrialContext, DataLoader
+from determined.pytorch._lightning import PLAdapter
+import pl
```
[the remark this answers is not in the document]
hamidzr: this one is just a relative import. maybe I should rename the file

harness/determined/pytorch/_lightning/adapter.py
hamidzr: interval != 1

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
rb-determined-ai: Wait, PyTorchTrial is for the user to define, why are we putting annotations on user objects?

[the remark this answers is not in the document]
shiyuann: `PTLAdapter` is a middle layer between user-defined and training-internal "layers". I see this class as something that can be user-defined.

examples/computer_vision/_mnist_pl/model_def.py
```diff
@@ -0,0 +1,31 @@
+"""
```
rb-determined-ai, line 1: This example will not be part of the PR that lands, right?  I'm guessing that's why `_mnist_py` starts with a `_`

harness/determined/pytorch/_lightning/adapter.py
rb-determined-ai: Isn't it unsafe to assume the user can accept `optimizer_idx`?  That's what I understand from the docs

rb-determined-ai (commented):
I'm mostly reviewing the API here.

Overall, this is quite a bit harder of a challeng than I was thinking, and I think you have a great start with a ton of details already covered.

I think @shiyuann's suggesting of subclasses here could result in a cleaner API.  The attributes that we want to overwrite on the lightning module could be set more cleanly.

Also, even though I normally really am opposed to mixing user code and Determined code, and even though I'm normally opposed to APIs where `super().__init__()` is part of the API, I agree with your design; those are the right choices in this case.  But we can make things a bit better by namspacing the variables we use into a plain-old-data class.

I'm thinking something like:
```python
# namespaced variables
class _LightningAdapterState:
    def __init__(self, context, lm, ...):
        self.context = context
        self.lm = lm
        ...

# lightweight extension of PyTorchTrialContext
class PyTorchLightningContext(PyTorchTrialContext):
    def __init__(self, ...):
        super().__init__(...)
        self._lm = None

    # user-facing call
    def wrap_lightning_module(self, lm):
        assert self._lm is None, "don't call this twice"

        # dynamically-created class, bound around variable `lm`
        class _WrapperClass(type(lm)):
            def __init__(self):
                self.ddp = False
                self.dp = False
                ...

            def __getattr__(wrapper, name):
                return getattr(lm, name)

            def __setattr__(wrapper, name, value):
                return setattr(lm, name, value)

            def __delattr__(wrapper, name):
                return delattr(lm, name)

            def optimizers(wrapper, use_pl_optimizer: bool = False):
                assert use_pl_optimizer is False, "use_pl_optimizer not supported"
                # do some determined magic?
                ...
                return lm.optimizers()

        return _WrapperClass()

class PyTorchLightningTrial(PyTorchTrial):
    trial_context_class = PyTorchLightningContext

    def __init__(self, context):
        if context._lm is None:
            raise ValueError("wrap before super().__init__ please")

        # low-impact on user namespace
        self._adapter_state = _LightningAdapterState(
            context = context,
            lm = context._lm,
            ...
        )

    def build_callbacks(self):
        # based on self._adapter_state
        ...

## user code

class MyTrial(PyTorchLightningTrial):
    def __init__(self, context):
        self.context = context
        self.my_lm = ...
        self.my_lm = context.wrap_lightning_module(my_lm)
        super().__init__(context)
```

shiyuann:
@rb-determined-ai I am not opposed to separate out user code and Determined code. But having a class that has only abstract methods and no member variables is an extreme, IMO. Generally, the design in Keras and Pytorch uses inheritance, which allows users to put everything they define in a separate class but have good interaction with the library code. I have also seen `super().__init__` being part of the API design many times. That being said, I think it's unnecessary to have a dynamically generated class just to provide access to the context object. Also, I think generally using a dynamically generated class is bad, which really abuses Python and makes code hard to read and debug. (I think the Keras interface needs to be changed.)

Besides, I think having a separate namespace is contradictory to the goal of the adapter approach. `PTLAdapter` aims at providing a middle layer between `PytorchTrial` and user code. `PyTorchLightningTrial` and `PyTorchLightningContext` sounds more like providing a separate training loop context by hiding the details with inheritance, which also will let users mix with the extension support.

rb-determined-ai:
> Besides, I think having a separate namespace is contradictory to the goal of the adapter approach

How so?  Users who use `PyTorchLightningTrial` do not want to learn the innards of `PyTorchLightningTrial`, and they won't want name conflicts between us and them.  Those sorts issues are very hard to debug for users because from the user's perspective, there is absolutely no reason why their code is wrong; it just breaks after the `super().__init__()` call because they also set `self.lm` or something.

(I feel pretty strongly this namespacing of our variables; I don't care nearly as much about the dynamic class vs monkeypatching thing)

> But having a class that has only abstract methods and no member variables is an extreme

What do you mean?  Neither class I sketched out is like that.

> it's unnecessary to have a dynamically generated class just to provide access to the context object. Also, generally using a dynamically generated class abuses Python and makes code hard to read and debug

What do you mean?  The dynamically generated class is for overriding attributes and methods without monkeypatching.  I think the monkeypatching strategy is bad because it looks to me like we are going to be making assumptions about *when* the user is going to call e.g. `optimizers`.

> Also, generally using a dynamically generated class abuses Python and makes code hard to read and debug

I mostly agree with you, I just think it's less bad than monkeypatching.  But what is the real "right" answer?  I think the real right answer is to *stop trying to control the training loop*.  We should be allowing users to just invoke the real lightning trainer with a Determined callback.  After having looked closer at the PTL training loop while reviewing this PR, I have real doubts that we can imitate the PTL trainer well enough to offer a good experience, and I would personally prefer that we focus on push architecture and sprinkle API first.

Side question: are there times that users have complained about readability or debuggability of TFKerasTrial due to the wrapped class?  I have consistently seen users complain about TFKerasTrial having different behavior than tf.keras, but I have not ever heard somebody complain about the dynamic class.

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
hamidzr: 👍🏼 . nvm I removed the question

harness/determined/pytorch/_lightning/adapter.py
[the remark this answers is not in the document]
hamidzr: I updated the docstring

examples/computer_vision/_mnist_pl/model_def.py
```diff
@@ -0,0 +1,31 @@
...
+This example shows how to interact with the Determined PyTorch interface to
```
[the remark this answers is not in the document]
hamidzr: how about 

"""
This example shows how to interact with the Determined PyTorch Lightning Adapter
interface to build a basic MNIST network. The PLAdapter utilizes the provided
LightningModule with Determined's trainer.
"""

examples/gan/_gan_mnist_pl/model_def.py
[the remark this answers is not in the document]
hamidzr: did the same for the other example as well

harness/determined/pytorch/_lightning/adapter.py
hamidzr: hmm  I changed these when we talked about using PL instead of PTL everywhere

harness/determined/pytorch/_callback.py
hamidzr: hmm what are you suggesting to be consistent with? our other callbacks start with `on_validation`
```
hrz ➜ determined git:(ptl-2) ag on_valida* harness/determined/pytorch/_callback.py                                                                                                   
17:        (except for :meth:`on_validation_end`, :meth:`on_validation_step_end` and
22:    def on_validation_start(self) -> None:
28:    def on_validation_end(self, metrics: Dict[str, Any]) -> None:
37:    def on_validation_step_start(self) -> None:
44:    def on_validation_step_end(self, metrics: Dict[str, Any]) -> None:
88:    def on_validation_epoch_start(self) -> None:
94:    def on_validation_epoch_end(self, outputs: List[Any]) -> None:
```

harness/determined/pytorch/_lightning/adapter.py
hamidzr: it should be set before all the hooks.

hamidzr: it starts at 0 and then increments https://github.com/PyTorchLightning/pytorch-lightning/blob/6cc1a060785448eacd016f6689d592ac822adaaf/pytorch_lightning/trainer/training_loop.py#L552

harness/determined/pytorch/_lightning/adapter.py
hamidzr: yeah you're right https://github.com/PyTorchLightning/pytorch-lightning/blob/6cc1a060785448eacd016f6689d592ac822adaaf/pytorch_lightning/trainer/training_loop.py#L505

harness/determined/pytorch/_pytorch_context.py
hamidzr: I removed it in the last review, there are no `auto_zero_grads` callbacks.


> Do you mean the check itself is wrong or the message?

￼the message? I don't know enough about hvd_config.aggregation_frequency to know for sure.

harness/determined/pytorch/_lightning/adapter.py
hamidzr: I used what the PL trainer does. https://github.com/determined-ai/determined/pull/1959/commits/11f1a4a659678a59d37c8c08a484ca98aff6feca

harness/determined/pytorch/_pytorch_trial.py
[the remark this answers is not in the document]
hamidzr: the duplicate logic issue for `get_epoch_idx` we can solve by another utility fn but it's small enough that I didn't. 

hmm one benefit that breaking up the loop will provide would be that it'll let us do operations we need to do on epoch boundaries without having to call context.is_epoch_end every iteration. for example we need it to split per_batch_metrics by epoch for on_train_epoch_end callback. that overhead seems small.

do we support dataloaders where we can't know the length of or with variable lengths?

harness/determined/pytorch/_lightning/adapter.py
hamidzr: FTR we decided to keep the batch_idx argument.

examples/computer_vision/_mnist_pl/model_def.py
hamidzr: it is why? depending on when we land it it might initially be with the _

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,184 @@
...
+        self.lm.use_ddp2 = False
+        self.lm.use_dp = False
+        self.lm.use_tpu = False
+        type(self.lm).local_rank = self.context.distributed.get_local_rank()
+        type(self.lm).global_rank = self.context.distributed.get_rank()
+        self.lm.use_amp = self.context._use_amp
+        self.lm.to(self.context.device)
+
+    def build_callbacks(self) -> Dict[str, PyTorchCallback]:
+        context = self.context
```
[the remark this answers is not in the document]
hamidzr: less verbosity. it's copying only a reference right?

shiyuann:
@rb-determined-ai 
There are a couple of questions here:

1. adapter support vs. extension support

> After having looked closer at the PTL training loop while reviewing this PR, I have real doubts that we can imitate the PTL trainer well enough to offer a good experience, and I would personally prefer that we focus on push architecture and sprinkle API first.

This is already been discussed at the beginning of the project. Although we couldn't avoid imitating Trainer, the places we do so is where we set the epoch index, batch index, and etc. The pro is building this adapter support can get immediate benefits to users who want to use LightningModule and don't care about Trainer a lot. The decision made was to use a couple of weeks to build this and make users happy.


> I think the real right answer is to stop trying to control the training loop. We should be allowing users to just invoke the real lightning trainer with a Determined callback.

yeah. What you described is the extension support.

2. monkey patching vs dynamic subclassing

Essentially, both of these two methods are imitating the trainer. They are both bad in dynamically changing the original methods or variables. But the reason I dislike dynamic subclassing is that it creates an unnecessary type.

3. Namespacing (PyTorchLightningTrial) vs. no separate namespace

`LightningTrial` is already booked for the extension support. If we provide two trials, it will be confusing to understand two different types of trials. However, if we provide `PTLAdapter(PytorchTrial)` and `LightningTrial`, it would be very clear that `PTLAdapter` is based on our Pytorch training loop and users shouldn't expect it behaves the same as Lightning itself without Determined.

4. getting rid of super().init() in the API

This is actually very common practice, at least being used widely in Pytorch Lightning. So I think it's totally okay to have such design for Lightning users.

5. Side question: are there times that users have complained about readability or debuggability of TFKerasTrial due to the wrapped class? I have consistently seen users complain about TFKerasTrial having different behavior than tf.keras, but I have not ever heard somebody complain about the dynamic class.

We don't. I think generally we should avoid both monkey patching and dynamic subclassing. The submodeling in tf.keras was introduced by the Native API, which tries to be accurate and flexible while fit into the `TFKerasTrialController`. It is a bit unnecessary if we remove the Native API. I am more in favor of the following design

```
class TFKerasTrial:
def __init__():
    self.model = self.context.Model(...)
    self.context.compile(...)

def train():
    self.model.fit()

def validate():
    self.model.evaluate()
```

shiyuann:
@rb-determined-ai feel free to join our weekly sync if you think we need to change anything.

harness/determined/pytorch/_pytorch_context.py
hamidzr: TODO remove the callback from the argument list

examples/computer_vision/_mnist_pl/model_def.py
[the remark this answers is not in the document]
hamidzr: Determined's trainer. => determined's pytorch control loop

harness/determined/pytorch/_callback.py
hamidzr: or on.train to on.training

harness/determined/pytorch/_pytorch_context.py
hamidzr: I included the fix for the aggregation frequency here.

harness/determined/pytorch/_pytorch_trial.py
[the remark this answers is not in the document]
hamidzr: I reverted back to the old loop

harness/determined/pytorch/_callback.py
hamidzr: I updated ontrain to be consistent with the rest. let's continue in slack with @ahj-determined

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class PLAdapter(PyTorchTrial):
+    def __init__(self, context: PyTorchTrialContext, lightning_module: pl.LightningModule):
+        super().__init__(context)
+        check_compatibility(lightning_module)
+        context.wrap_model(lightning_module)
+        optimizers, lr_schedulers = self.setup_optimizers_schedulers(context, lightning_module)
+        pls = _PLAdapterState(context, lightning_module, optimizers)
+        self._pls = pls
```
hamidzr, line 62: short name suggestions with low collision possibility?
pls: pl state

examples/computer_vision/_mnist_pl/adaptive.yaml
```diff
@@ -0,0 +1,17 @@
+description: mnist_pytorch_lightning_adaptive
+max_restarts: 1
```
shiyuann, line 2: remove this when merging

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -308,8 +313,14 @@ def _train_for_step(
...
 
-            per_batch_metrics.append(tr_metrics)
+            epoch_per_batch_metrics.append(tr_metrics)
+            if self.context.is_epoch_end():
+                for callback in self.callbacks.values():
+                    callback.on_training_epoch_end(epoch_per_batch_metrics)
+                per_batch_metrics += epoch_per_batch_metrics
+                epoch_per_batch_metrics = []
 
+        per_batch_metrics += epoch_per_batch_metrics
```
shiyuann, line 323: This isn't right. If a step has batches in two sequential epochs, `per_batch_metrics` contains the metrics of all the batches but `epoch_per_batch_metrics` would clear itself once it hits an epoch end.

examples/computer_vision/_mnist_pl/model_def.py
```diff
@@ -0,0 +1,26 @@
...
+
+class MNISTTrial(PLAdapter):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        lm = mnist.LightningMNISTClassifier(lr=context.get_hparam('learning_rate'))
+        self.dm = mnist.MNISTDataModule()
+
+        super().__init__(context, lightning_module=lm)
+
+    def build_training_data_loader(self) -> DataLoader:
+        self.dm.setup()
```
shiyuann, line 19: This doesn't seem to be right since this function downloads data to the same directory for all the GPUs, which might cause errors

examples/gan/_gan_mnist_pl/model_def.py
```diff
@@ -0,0 +1,33 @@
...
+                    lr=context.get_hparam('lr'),
+                    b1=context.get_hparam('b1'),
+                    b2=context.get_hparam('b2'),
+        )
+
+        super().__init__(context, lightning_module=lm)
+        self.dm.prepare_data()
+
+    def build_training_data_loader(self) -> DataLoader:
+        self.dm.setup()
```
shiyuann, line 26: same

harness/determined/pytorch/_lightning/adapter.py
shiyuann: just `_pl`? I am fine with any names.

examples/computer_vision/_mnist_pl/mnist.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import Optional
+import os
+
+
+class LightningMNISTClassifier(pl.LightningModule):
+
+    def __init__(self, *args, lr: float, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.save_hyperparameters()
```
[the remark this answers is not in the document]
hamidzr: So that the inputs get saved under `self.hparams` if I'm not reading this wrong https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#save-hyperparameters

harness/determined/util.py
```diff
@@ -37,6 +38,18 @@ def is_overridden(full_method: Any, parent_class: Any) -> bool:
     return False
 
 
+def has_param(fn: Callable[..., Any], name: str, pos: Optional[int] = None) -> bool:
```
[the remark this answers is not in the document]
hamidzr: that was my question too. Would param fit better if we go by this? https://stackoverflow.com/a/1788926

harness/determined/pytorch/_pytorch_trial.py
hamidzr: 👍🏼  we removed support for on_train_epoch_end since we can't guarantee epochwide metrics for this callback with mid-epoch preemption

examples/computer_vision/_mnist_pl/model_def.py
hamidzr: I moved the download section to dm.prepare() as PL suggests and used local_rank to point it to avoid collisions

hamidzr:
I squashed the history so that the reviews are easier to read.

hamidzr:
I did some testing with slots_per = 2 on a single machine/agent. it might be good to also test in a multi-agent scenario

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,275 @@
+import inspect
```
hamidzr, line 1: should this file be named _adapter?

harness/determined/pytorch/_lightning/adapter.py
```diff
@@ -0,0 +1,275 @@
...
+            lightning_module,
+        )
+        for freq in opt_frequencies:
+            check.eq(freq, 1, "custom optimizer frequencies are not supported")
+        optimizers = cast(List[Optimizer], optimizers)
+        lr_scheduler_dicts = cast(List[dict], lr_scheduler_dicts)
+
+        def lightning_scheduler_dict_to_det(lrs: dict) -> _LRScheduler:
+            """
+            input_dict = {
```
shiyuann, line 172: blocking: Need to add checks for the fields that we don't support. We need to support frequency. If `monitor` and `strict` are set, we should add checks and fail.

hamidzr: > We need to support frequency.

let's leave this for a follow-up pr. I'll add the checks

examples/computer_vision/_mnist_pl/model_def.py
shiyuann: @rb-determined-ai The example will land as part of the PR. And let's remove the underscore in the file names.

examples/gan/_gan_mnist_pl/distributed.yaml
```diff
@@ -0,0 +1,29 @@
+description: gan_mnist_pytorch_lightning_adaptive
```
shiyuann, line 1:
```suggestion
description: gan_mnist_pytorch_lightning_distributed
```

shiyuann (approved):
good work!

examples/computer_vision/_mnist_pl/pl.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def configure_optimizers(self):
+        optimizer = torch.optim.Adam(self.parameters(),
+                                     lr=self.hparams.lr)
+        return optimizer
+
+
+class MNISTDataModule(pl.LightningDataModule):
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
hamidzr: Just following the pattern that lightning suggests https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#minimal-example

harness/determined/pytorch/_lightning/adapter.py
hamidzr: https://determinedai.atlassian.net/browse/DET-5021

hamidzr:
thanks for the many reviews!

PR #1964, opened 2021-02-12
chore: commit generated python schemas

rb-determined-ai:
## Description

It was a mistake to think these did not need to be generated.


stoksc (approved):
lgtm

PR #1965, opened 2021-02-12
feat: HP importance implementation [DET-4465]

mackrorysd:
## Description

This is a combination of the original code submitted with this draft pull request, all of Katie's work from https://github.com/determined-ai/determined/pull/2005, and a bunch of iteration from both since then including most of Bradley's original code review feedback.

## Test Plan

The nightly integration test was run locally.


agent/go.sum
```diff
@@ -744,6 +744,7 @@ github.com/rs/xid v1.2.1/go.mod h1:+uKXf+4Djp6Md1KODXJxgGQPKngRmWyn10oCKFzNHOQ=
 github.com/rs/zerolog v1.13.0/go.mod h1:YbFCdg8HfsridGWAh22vktObvhZbQsZXe4/zB0OKkWU=
 github.com/rs/zerolog v1.15.0/go.mod h1:xYTKnLHcpfU2225ny5qZjxnj9NvkumZYjJHlAThCjNc=
 github.com/russross/blackfriday/v2 v2.0.1/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
+github.com/ryanbressler/CloudForest v0.0.0-20161201194407-d014dc32840a/go.mod h1:295JVys1fl5dBchZafEH0L0FfJONcP/6lnXweezSn5w=
```
mackrorysd, line 747: Not quite sure what to do about this. The `go install` of growforest in the Makefile seems to be what triggers this. That seems... wrong. `go mod tidy` gets rid of it, but this is going to keep showing up every time someone else builds if I can't suppress it.

master/internal/core.go
```diff
@@ -580,25 +580,28 @@ func (m *Master) Run(ctx context.Context) error {
 	template.RegisterAPIHandler(m.echo, m.db, authFuncs...)
 
 	if m.config.Telemetry.Enabled && m.config.Telemetry.SegmentMasterKey != "" {
-		if telemetry, err := telemetry.NewActor(
+		if telemetryActor, tErr := telemetry.NewActor(
```
stoksc, line 583: what's the point of telemetry -> telemetryActor? ask because it's kinda a go-ism to not name things their [type](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_dont_name_your_variables_for_their_types)

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+func createDataFile(data map[int][]model.HPImportanceTrialData,
+	experimentConfig *model.ExperimentConfig, workingDir string) (int, error) {
+	f, err := os.Create(path.Join(workingDir, arffFile))
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		err = f.Close()
+		if err != nil {
+			fmt.Printf("failed to close arff file: %s", err.Error())
```
stoksc, line 47: Use logrus instead here. You don't need to pass it from the actor ctx, you can just import logrus as log and use it.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+		if batchID < maxNumBatches/maxDiffCompBatch {
+			break
+		}
+		if totalNumTrials > minNumberTrials && batchID <= maxNumBatches/idealBatchDiff {
+			break
+		}
+		batchIDStr := fmt.Sprintf("%v", batchID)
+		for _, trial := range data[batchID] {
+			var st string
+			st += fmt.Sprintf("%v", trial.Metric)
```
stoksc, line 111: Could use a `strings.Builder` [instead](https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go/47798475#47798475). Alternatively, you could wrap the file as a `bufio.NewWriter` if you were concerned about each call to `WriteString` causing a context switch and copy from user space into kernel buffers.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+		}
 	}
-	return output
+	return totalNumTrials, nil
+}
+
+// Read the data from the output importance file and return as a json.
+func parseImportanceOutput(filename string) (map[string]float64, error) {
+	// Ignore security warning because none of this is user-provided input
+	// #nosec G304
```
stoksc, line 131:
```suggestion
	// #nosec G304 // Ignore security warning because none of this is user-provided input
```
usually the explanation goes like this

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open HP importance file: %s", err.Error())
```
stoksc, line 134: I commented this somewhere else, but make sure to use `logrus` everywhere or else these logs are going to stick out like sore thumbs.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+			return nil, fmt.Errorf("failed to read HP importance file: %s", err.Error())
+		}
+
+		if record[1] == "metric" || record[1] == "numBatches" {
+			continue
+		}
+		hpi[record[1]], err = strconv.ParseFloat(record[2], 64)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse HP importance value: %s", err.Error())
+		}
```
stoksc, line 161: can you drop a comment to describe this csv file? maybe just say the headers.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+	r := csv.NewReader(file)
+	r.Comma = '\t'
+	for {
+		record, err := r.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to read HP importance file: %s", err.Error())
+		}
```
stoksc, line 153: should use a switch here. see `agent/internal/detect.go:93`

agent/go.sum
mackrorysd: Ooohhhh it's the GO111MODULE thing. The go tooling will detect it's running inside a module, so that's how it's ending up in there even though the code never references it (except through the process invocation - but I tested - that has nothing to do with it). Just turning that off when doing the `go install` fixes it.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+		// number of replications/permutations
+		"-ace", strNumTrees,
+		// output file
+		"-importance", path.Join(workingDir, importanceFile),
+		// file name to output predictor forest in sf format.
+		"-rfpred", path.Join(workingDir, "rface.sf"),
+	).CombinedOutput()
+	if err != nil {
+		fmt.Printf("growforest failed:\n" + string(output))
+		return nil, fmt.Errorf("random forest failed: %s", err.Error())
```
stoksc, line 222: I'd just use wrap here, same above.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+
+	arffFile       = "data.arff"
+	importanceFile = "importance.txt"
+)
+
+// The data needs to be put into an arff format.
+// Arff files require variable declaration at the top,
+// where order does matter. We need to keep define the Hps and
+// keep track of the order so the data columns will match.
+func createDataFile(data map[int][]model.HPImportanceTrialData,
```
stoksc, line 38: as of now, how are these file getting cleaned up?

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+	experimentConfig *model.ExperimentConfig, workingDir string) (int, error) {
+	f, err := os.Create(path.Join(workingDir, arffFile))
```
stoksc, line 40: maybe just pass `path.Join(workingDir, arffFile)` in, since it gets constructed here and outside to pass to the `growforest` cmd, and changing one without the other would be a bug.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,227 @@
...
+		"-importance", path.Join(workingDir, importanceFile),
+		// file name to output predictor forest in sf format.
+		"-rfpred", path.Join(workingDir, "rface.sf"),
+	).CombinedOutput()
+	if err != nil {
+		fmt.Printf("growforest failed:\n" + string(output))
+		return nil, fmt.Errorf("random forest failed: %s", err.Error())
+	}
+
+	hpi, err := parseImportanceOutput(path.Join(workingDir, importanceFile))
```
stoksc, line 225: same comment for this `path.Join`. construct it above and use it in both places maybe.

master/internal/hpimportance/manager.go
```diff
@@ -59,23 +71,39 @@ type stateRecord struct {
...
 // NewManager initializes the master actor (of which there should only be one instance running).
-func NewManager(db *db.PgDB, system *actor.System, config HPImportanceConfig) actor.Actor {
+func NewManager(db *db.PgDB, system *actor.System, config HPImportanceConfig, masterRoot string,
+) (actor.Actor, error) {
+	// growforest should either be installed in PATH (when running from source) or package with the
+	// master (when running from binary packages).
+	growforest := path.Join(masterRoot, growforestBin)
+	_, err := os.Stat(growforest)
+	if os.IsNotExist(err) {
+		growforest = growforestBin
```
stoksc, line 89: maybe check the path here and error if it's not there. and print an error that says how to install?

master/packaging/fetch-cloud-forest.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Install native version to ensure it's on PATH
+GO111MODULE=off go install github.com/ryanbressler/CloudForest/growforest
+
+# Ensure the Linux binary is installed regardless of OS
+GO111MODULE=off GOOS=linux go install github.com/ryanbressler/CloudForest/growforest
```
stoksc, line 7: I said this elsewhere, but my concern with this is with variations in GOARCH not being accounted for. Also we talked offline but I dont think `GO111MODULE=off` is the way to go for this.

stoksc (commented):
looks pretty good. this is just a first pass. I'll take another before the eod.

master/internal/hpimportance/hpimportance.go
stoksc: i saw this later on, lgtm.

master/internal/core.go
mackrorysd: This came up at the same time as when there started being errors about shadowing (err -> tErr). It also seemed bad practice to have a variable reuse a name used by a package. I'll switch this back and see if a specific failure comes back.

stoksc: ah, yeah, it's a package name. mostly that doesn't matter but then you can't use the package later in the same scope. 🤷 idc, you can leave it this way. i was just curious why.

master/internal/hpimportance/manager.go
```diff
@@ -59,23 +71,39 @@ type stateRecord struct {
...
+	}
+
+	err = os.MkdirAll(workingDir, 0700)
```
stoksc, line 92: talked offline, on restart (so here) we should wipe everything under this dir, to clean of stuff that got left from a previous run.

master/packaging/fetch-cloud-forest.sh
mackrorysd: Switched to tools.go

master/internal/core.go
mackrorysd: Whatever error motivated it hasn't come back - perhaps a misunderstanding

master/internal/hpimportance/hpimportance.go
mackrorysd: This isn't printing it to logs - this one should get plumbed back and stored with the HP importance state in the DB. Eventually it'll be available to users when they try to see the results.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,242 @@
...
+		}
+		if totalNumTrials > minNumberTrials && batchID <= maxNumBatches/idealBatchDiff {
+			break
+		}
+		batchIDStr := fmt.Sprintf("%v", batchID)
+		for _, trial := range data[batchID] {
+			var st string
+			st += fmt.Sprintf("%v", trial.Metric)
+			hparamsVals := trial.Hparams
+			for _, hp := range hpsOrder {
```
stoksc, line 118: `trial.Hparams` is a `hparamSample`, which will have samples for const hyperparameters, but we don't write an `@attribute` out for const above. If there is a const hyperparameter, would this break because the arff file would be malformed? just trying to make sure i understand.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,242 @@
...
+		"-importance", growforestOutput,
+		// file name to output predictor forest in sf format.
+		"-rfpred", path.Join(workingDir, "rface.sf"),
+	).CombinedOutput()
+	if err != nil {
+		log.Error("growforest failed:\n " + string(output))
+		return nil, fmt.Errorf("random forest failed: %s", err.Error())
+	}
+
+	hpi, err := parseImportanceOutput(path.Join(workingDir, importanceFile))
```
stoksc, line 240:
```suggestion
	hpi, err := parseImportanceOutput(growforestOutput)
```

stoksc (commented):
had a few last q's/comments.

master/internal/hpimportance/hpimportance.go
mackrorysd: I've switched to a string builder everywhere we were doing file IO before. There's still a few Sprintf's where the functionality of the format string adds value, or to minimize the amount of error checking we ought to do. I realize errors from the StringBuilder are unlikely, but as the API is exactly the same as file IO, I replaced the file I/O calls with StringBuilder calls and nothing else.

master/Makefile
```diff
@@ -74,6 +76,7 @@ pre-package:
 	cp ../common/dist/*.whl build/wheels/
 	cp ../cli/dist/*.whl build/wheels/
 	cp ../harness/dist/*.whl build/wheels/
+	cp $(shell packaging/fetch-cloud-forest.sh) build/
```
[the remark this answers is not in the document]
mackrorysd: Fair enough - I don't disagree, but I need to dig a bunch more into goreleaser and see if we could do this in the Docker phase.

master/internal/hpimportance/hpimportance.go
mackrorysd: So when we're building the data file we first look at all the parameter types and we filter out the ones that are const. Then as we're going through the data, we only use the ones that weren't filtered out.

If we eventually refactor this into shipping the data over the network somewhere, then this would be a good thing to improve: filter out the data from the const hyperparameters earlier. But no, this won't break. In fact the integration test explicitly covers this: there are const parameters, and we check that they're excluded from HP importance results but that it doesn't otherwise break.

stoksc: Oh I see, I missed `hpsOrder` was the filtered list somehow. I understand now.

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,241 @@
 package hpimportance
 
+/**
```
stoksc, line 3: these comments usually go above the package declaration in go https://github.com/golang/go/blob/master/src/crypto/cipher/cbc.go#L1-L12

stoksc (approved):
lgtm

master/internal/hpimportance/hpimportance.go
stoksc: sgtm. it shouldn't be so much data that building the string in memory is an issue, right?

mackrorysd: Discussed offline: the sprintfs are small. That total data may be larger, but shouldn't exceed other memory usage associated with this feature. But all things considered, bufio is a good idea.

PR #1967, opened 2021-02-16
chore: stop hardcoding python3.6 [DET-4929] [DET-4539]

rb-determined-ai:
Support setting the DET_PYTHON_EXECUTABLE as an escape hatch for very
strange custom environments, but since it seems unlikely many users
ever need it, document it through error messages rather than in docs.

Also support startup-hook.sh in notebooks and shells while we're at it.


master/static/srv/notebook-entrypoint.sh
```diff
@@ -15,6 +25,7 @@ if [ "$HOME" = "/" ] ; then
     export HOME
 fi
 
-python3.6 -m pip install -q --user /opt/determined/wheels/determined*.whl
+"$DET_PYTHON_EXECUTABLE" -m pip install -q --user /opt/determined/wheels/determined*.whl
 
-jupyter lab --config /run/determined/workdir/jupyter-conf.py --port=${NOTEBOOK_PORT}
+(cd ${WORKING_DIR} && test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}")
```
dzhu, line 30: With this in a subshell, some things (like exporting environment variables) won't work.

master/static/srv/shell-entrypoint.sh
```diff
@@ -2,13 +2,25 @@
...
+fi
 
 # Unlike trial and notebook entrypoints, the HOME directory does not need to be
 # modified in this entrypoint because the HOME in the user's ssh session is set
 # by sshd at a later time.
 
-python3.6 -m pip install -q --user /opt/determined/wheels/determined*.whl
+"$DET_PYTHON_EXECUTABLE" -m pip install -q --user /opt/determined/wheels/determined*.whl
+
+(cd ${WORKING_DIR} && test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}")
```
dzhu, line 23: Also subshell.

docs/release-notes/1967-python3-and-startup-hooks.txt
```diff
@@ -0,0 +1,18 @@
...
+
+-  Support ``startup-hook.sh`` for notebooks and shells. This is the
+   same mechanism supported by experiments. See
+   :ref:`notebook-configuration` for more details.
+
+**Improvements**
+
+-  Invoke ``python`` as ``python3`` rather than as ``python3.6``. This
+   makes it possible to use custom images with Determined with higher
+   versions of python (3.6 is still the minimum required version).
```
dzhu, line 13: Maybe this?
```suggestion
   makes it possible to use custom images containing higher versions of
   Python with Determined (3.6 is still the minimum required version).
```

dzhu (commented):
I haven't actually tried running tasks with an alternate Python, but I tested that the value of the environment variable is picked up and executed.

What were those old Dockerfiles doing? Were they just unused?

docs/how-to/notebooks.txt
```diff
@@ -117,14 +117,14 @@ Notebooks may be supplied an optional *notebook configuration* to
 control aspects of the notebook's environment. For example, to launch a
 notebook that uses two GPUs:
 
-.. code::
+.. code:: bash
```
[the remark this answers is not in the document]
rb-determined-ai: ah, makes sense.

master/static/srv/notebook-entrypoint.sh
rb-determined-ai: good point; I'll use pushd/popd

docs/release-notes/1967-python3-and-startup-hooks.txt
rb-determined-ai: yeah I like that

rb-determined-ai:
Yes, the old dockerfiles were unused.  They used to be used to test the old build-at-runtime features I think, and then to test running custom images, but somewhere that was removed.  I'm fine with that, honestly; I don't see a real risk of us accidentally not working with custom images anymore.

dzhu (approved):

PR #1968, opened 2021-02-16
refactor: add internal restful APIs for checkpoints and metrics [DET-4896]

shiyuann:
## Description

Add internal restful APIs in the master. Those APIs are used for registering saved checkpoints and both training and validation metrics.

## Test Plan

Tested it manually by launch new experiments, pausing them, checking if the WebUI and CLI display the experiments correctly.

## Commentary (optional)

N/A

## Checklist

### DIFF
--- a/harness/determined/__init__.py
+++ b/harness/determined/__init__.py
@@ -1,4 +1,5 @@
 from determined.__version__ import __version__
+from determined import errors, workload
 from determined._env_context import EnvContext
 from determined._rendezvous_info import RendezvousInfo
 from determined._execution import (
@@ -17,5 +18,4 @@ from determined._trial_controller import (
     LoopTrialController,
     TrialController,
 )
-from determined import errors
 from determined import util
--- a/harness/determined/_env_context.py
+++ b/harness/determined/_env_context.py
@@ -33,6 +33,7 @@ class EnvContext:
         trial_seed: int,
         managed_training: bool = True,
         test_mode: bool = False,
+        local_mode: bool = False,
     ):
         self.master_addr = master_addr
         self.master_port = master_port
@@ -58,6 +59,7 @@ class EnvContext:
         self.trial_seed = trial_seed
         self.managed_training = managed_training
         self.test_mode = test_mode
+        self.local_mode = local_mode
 
         self._per_slot_batch_size, self._global_batch_size = self._calculate_batch_sizes()
 
@@ -117,3 +119,58 @@ class EnvContext:
     @property
     def global_batch_size(self) -> int:
         return self._global_batch_size
+
+    def get_experiment_config(self) -> Dict[str, Any]:
+        """
+        Return the experiment configuration.
+        """
+        return self.experiment_config
+
+    def get_data_config(self) -> Dict[str, Any]:
+        """
+        Return the data configuration.
+        """
+        return self.get_experiment_config().get("data", {})
+
+    def get_experiment_id(self) -> int:
+        """
+        Return the experiment ID of the current trial.
+        """
+        return int(self.det_experiment_id)
+
+    def get_trial_id(self) -> int:
+        """
+        Return the trial ID of the current trial.
+        """
+        return int(self.det_trial_id)
+
+    def get_trial_seed(self) -> int:
+        return self.trial_seed
+
+    def get_hparams(self) -> Dict[str, Any]:
+        """
+        Return a dictionary of hyperparameter names to values.
+        """
+        return self.hparams
+
+    def get_hparam(self, name: str) -> Any:
+        """
+        Return the current value of the hyperparameter with the given name.
+        """
+        if name not in self.hparams:
+            raise ValueError(
+                "Could not find name '{}' in experiment "
+                "hyperparameters. Please check your experiment "
+                "configuration 'hyperparameters' section.".format(name)
+            )
+        if name == "global_batch_size":
+            logging.warning(
+                "Please use `context.get_per_slot_batch_size()` and "
+                "`context.get_global_batch_size()` instead of accessing "
+                "`global_batch_size` directly."
+            )
+        return self.hparams[name]
+
+    def get_unary_host(self) -> str:
+        scheme = "https" if self.use_tls else "http"
+        return f"{scheme}://{self.master_addr}:{self.master_port}"
--- a/harness/determined/layers/_workload_manager.py
+++ b/harness/determined/layers/_workload_manager.py
@@ -5,8 +5,9 @@ from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, cast
 
 import determined as det
-from determined import tensorboard, workload
+from determined import tensorboard, util, workload
 from determined_common import storage
+from determined_common.api import request as req
 from determined_common.check import (
     check_eq,
     check_len,
@@ -14,6 +15,8 @@ from determined_common.check import (
     check_not_isinstance,
     check_not_none,
 )
+
+TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 
 def _current_timestamp() -> datetime:
@@ -122,15 +125,6 @@ class _TrialWorkloadManager(WorkloadManager):
     def yield_train_for_step(
         self, wkld: workload.Workload, respond: workload.ResponseFunc
     ) -> workload.Stream:
-        start_time = _current_timestamp()
-
-        for callback in self.callbacks:
-            if wkld.step_id == 1:
-                callback.on_trial_begin()
-            callback.on_train_step_begin(
-                wkld.step_id, wkld.num_batches, wkld.total_batches_processed
-            )
-
         def _respond(in_response: workload.Response) -> None:
             # Only the chief container should actually respond to TRAIN_FOR_STEP.
             if self.rendezvous_info.get_rank() != 0:
@@ -143,12 +137,14 @@ class _TrialWorkloadManager(WorkloadManager):
             metrics = in_response["metrics"]
             metrics = cast(workload.Metrics, metrics)
 
+            end_time = _current_timestamp()
+
             if in_response.get("invalid_hp", False):
                 out_response = {
                     "type": "WORKLOAD_COMPLETED",
                     "workload": wkld,
                     "start_time": start_time,
-                    "end_time": _current_timestamp(),
+                    "end_time": end_time,
                     "metrics": metrics,
                 }
                 out_response["exited_reason"] = "INVALID_HP"
@@ -157,7 +153,7 @@ class _TrialWorkloadManager(WorkloadManager):
 
             batch_metrics = metrics["batch_metrics"]
             # Sanity-check training metrics.
-            det.util.validate_batch_metrics(batch_metrics)
+            util.validate_batch_metrics(batch_metrics)
             check_len(batch_metrics, wkld.num_batches)
 
             for callback in self.callbacks:
@@ -171,23 +167,63 @@ class _TrialWorkloadManager(WorkloadManager):
                 "type": "WORKLOAD_COMPLETED",
                 "workload": wkld,
                 "start_time": start_time,
-                "end_time": _current_timestamp(),
+                "end_time": end_time,
                 "metrics": metrics,
             }
 
             if in_response.get("stop_requested", False):
                 out_response["exited_reason"] = "USER_CANCELED"
 
-            # Send the response up.
+            if not self.env.local_mode:
+                training_metrics = {
+                    "trial_id": self.env.get_trial_id(),
+                    "step_id": wkld.step_id,
+                    "start_batch": wkld.total_batches_processed,
+                    "end_batch": wkld.target_batch(),
+                    "start_time": start_time.strftime(TIMESTAMP_FORMAT),
+                    "end_time": end_time.strftime(TIMESTAMP_FORMAT),
+                    "state": "STATE_COMPLETED",
+                    "metrics": util.make_serializable(metrics),
+                }
+                req.put(
+                    self.env.get_unary_host(),
+                    "/api/v1/training_metrics",
+                    body={"training_metrics": training_metrics},
+                )
+
+            # Send the response up to the socket manager.
             respond(out_response)
 
+        start_time = _current_timestamp()
+
+        for callback in self.callbacks:
+            if wkld.step_id == 1:
+                callback.on_trial_begin()
+            callback.on_train_step_begin(
+                wkld.step_id, wkld.num_batches, wkld.total_batches_processed
+            )
+
+        if not self.env.local_mode:
+            training_metrics = {
+                "trial_id": self.env.get_trial_id(),
+                "step_id": wkld.step_id,
+                "start_batch": wkld.total_batches_processed,
+                "end_batch": wkld.target_batch(),
+                "start_time": start_time.strftime(TIMESTAMP_FORMAT),
+                "state": "STATE_ACTIVE",
+            }
+            req.post(
+                self.env.get_unary_host(),
+                "/api/v1/training_metrics",
+                body={"training_metrics": training_metrics},
+            )
+
+        # Yield the workload to the framework layer.
         yield wkld, [], _respond
 
     def yield_compute_validation_metrics(
         self, wkld: workload.Workload, respond: workload.ResponseFunc
     ) -> workload.Stream:
-        start_time = _current_timestamp()
-
         def _respond(in_response: workload.Response) -> None:
 
             # Only the chief container should actually respond to COMPUTE_VALIDATION_METRICS.
@@ -274,47 +310,77 @@ class _TrialWorkloadManager(WorkloadManager):
                 for metric_name in non_serializable_metrics:
                     del v_metrics[metric_name]
 
+            end_time = _current_timestamp()
+
             out_response = {
                 "type": "WORKLOAD_COMPLETED",
                 "workload": wkld,
                 "start_time": start_time,
-                "end_time": _current_timestamp(),
+                "end_time": end_time,
                 "metrics": metrics,
             }
 
             if in_response.get("stop_requested", False):
                 out_response["exited_reason"] = "USER_CANCELED"
 
+            if not self.env.local_mode:
+                validation_metrics = {
+                    "trial_id": self.env.get_trial_id(),
+                    "total_batches": wkld.target_batch(),
+                    "start_time": start_time.strftime(TIMESTAMP_FORMAT),
+                    "end_time": end_time.strftime(TIMESTAMP_FORMAT),
+                    "state": "STATE_COMPLETED",
+                    "metrics": util.make_serializable(metrics),
+                }
+                req.put(
+                    self.env.get_unary_host(),
+                    "/api/v1/validation_metrics",
+                    body={"validation_metrics": validation_metrics},
+                )
+
+            # Send the response up to the socket manager.
             respond(out_response)
+
+        start_time = _current_timestamp()
+        if not self.env.local_mode:
+            validation_metrics = {
+                "trial_id": self.env.get_trial_id(),
+                "total_batches": wkld.target_batch(),
+                "start_time": start_time.strftime(TIMESTAMP_FORMAT),
+                "state": "STATE_ACTIVE",
+            }
+            req.post(
+                self.env.get_unary_host(),
+                "/api/v1/validation_metrics",
+                body={"validation_metrics": validation_metrics},
+            )
 
         for callback in self.callbacks:
             callback.on_validation_step_begin(wkld.step_id, wkld.total_batches_processed)
 
+        # Yield the workload to the framework layer.
         yield wkld, [], _respond
 
     def yield_checkpoint_model(
         self, wkld: workload.Workload, respond: workload.ResponseFunc
     ) -> workload.Stream:
-        start_time = _current_timestamp()
-
-        # Only the chief container should checkpoint.
-        if self.rendezvous_info.get_rank() != 0:
-            respond(workload.Skipped())
-            return
 
         # Save the workload completed message for after checkpoint upload completes.
         message = None  # type: Optional[workload.Response]
 
         def _respond(checkpoint_info: workload.Response) -> None:
             checkpoint_info = cast(Dict[str, Any], checkpoint_info)
+            resources = storage.StorageManager._list_directory(path)
+            framework = checkpoint_info.get("framework", "")
+            format = checkpoint_info.get("format", "")
+
             metadata = storage.StorageMetadata(
-                storage_id,
-                storage.StorageManager._list_directory(path),
-                checkpoint_info.get("framework", ""),
-                checkpoint_info.get("format", ""),
+                uuid,
+                resources,
+                framework,
+                format,
             )
 
-            logging.info("Saved trial to checkpoint {}".format(metadata.storage_id))
             self.tensorboard_mgr.sync()
 
             nonlocal message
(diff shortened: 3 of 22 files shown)


stoksc:
Added @rb-determined-ai for the harness side of stuff. I'll give it a review too but a second opinion there would be good.

master/internal/api_checkpoint.go
```diff
@@ -28,6 +29,42 @@ func (a *apiServer) GetCheckpoint(
...
+	modelC, err := model.CheckpointFromProto(req.Checkpoint)
+	if err != nil {
+		return nil, errors.Wrapf(
+			err, "error adding checkpoint %s (trial %d, batch %d) in database",
+			req.Checkpoint.Uuid, req.Checkpoint.TrialId, req.Checkpoint.BatchNumber)
+	}
+	err = a.m.db.AddCheckpoint(modelC)
+	return &apiv1.CreateCheckpointResponse{Checkpoint: req.Checkpoint},
+		errors.Wrapf(err, "error adding checkpoint %s (trial %d, batch %d) in database",
+			req.Checkpoint.Uuid, req.Checkpoint.TrialId, req.Checkpoint.BatchNumber)
```
stoksc, line 46:
```suggestion
err = a.m.db.AddCheckpoint(modelC)
if err != nil {
	return nil, errors.Wrapf(err, "error adding checkpoint %s (trial %d, batch %d) in database",
		req.Checkpoint.Uuid, req.Checkpoint.TrialId, req.Checkpoint.BatchNumber)
}
return &apiv1.CreateCheckpointResponse{Checkpoint: req.Checkpoint}, nil
```
I have a preference for this so that resp is nil when there are errors, not just non-sense/corrupted data.

master/internal/api_checkpoint.go
```diff
@@ -28,6 +29,42 @@ func (a *apiServer) GetCheckpoint(
 	}
 }
 
+func (a *apiServer) CreateCheckpoint(
+	_ context.Context, req *apiv1.CreateCheckpointRequest,
+) (*apiv1.CreateCheckpointResponse, error) {
+	log.Infof("adding checkpoint %s (trial %d, batch %d)",
```
stoksc, line 35: I would change these to `Debugf` before landing or make a note to do it after a release or two if you want them for debugging

master/internal/db/postgres.go
```diff
@@ -1325,21 +1325,26 @@ func (db *PgDB) AddStep(step *model.Step) error {
 	if !step.IsNew() {
 		return errors.Errorf("unexpected state for new step: %v", step)
 	}
-	trial, err := db.TrialByID(step.TrialID)
+	var count int
+	err := db.namedGet(&count, `
```
stoksc, line 1329: We should just add a uniqueness constraint to this column set in the database instead of trying to manage the integrity manually (in hindsight, we should've added this with your last migration). I'm not saying we _have_ to do that, but that if you see a bug with two steps at the same count, this would be it. But I think we should since if the trial has a bug where it sends the same step out back to back or something, we actually may see this.

master/internal/experiment_utils.go
```diff
@@ -57,50 +56,3 @@ func checkpointFromCheckpointMetrics(metrics workload.CheckpointMetrics) model.C
 		Format:    metrics.Format,
 	}
 }
-
-func saveWorkload(db *db.PgDB, w workload.Workload) error {
-	switch w.Kind {
```
stoksc, line 62: Jeez.. deleting these makes me so happy, just because I've seen these error messages so much while debugging issues. I'll try to think about the class of bugs they surfaced though and make sure we're covered. But I think it's mostly covered by the new messages that complain about inserting/updating duplicate/existing training and validation workloads, and those getting bubbled up from http calls now.

master/internal/trial.go
```diff
@@ -739,9 +728,6 @@ func (t *trial) sendNextWorkload(ctx *actor.Context) error {
 		t.TerminationSent = true
 		actors.NotifyAfter(ctx, terminateTimeoutPeriod, terminateTimeout{runID: t.RunID})
 	} else {
-		if err := saveWorkload(t.db, w); err != nil {
```
stoksc, line 742: seems like a kinda weird in-between state to be in where workloads are getting pushed to the rest api to be persisted into the steps table but also pushed back through the websocket for the workload sequencer to coordinate next steps with the trial. do you think we need a feature branch for the second half of this (moving the workload sequencer) and this to be lumped together into a single change? or maybe wait until you know you can hit them both in a single release?

stoksc (commented):
go side looks pretty good to me. i need to let this one sink in a bit though to try to figure out what i may be missing that could go sideways.

master/internal/trial.go
stoksc: Maybe we can talk about this offline.

harness/determined/_model_metric_manager.py
```diff
@@ -0,0 +1,231 @@
...
+        self.state = "STATE_ERRORED"
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):  # type: ignore
+        if self.state == "STATE_COMPLETED":
+            return
+
+        self.failed()
+
+
+class ValidationMetricManager:
```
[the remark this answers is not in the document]
shiyuann: It's not code duplication. Checkpoints, steps, validations are independent data models that don't have a common ancestor.

harness/determined/_checkpoint_manager.py
```diff
@@ -0,0 +1,122 @@
...
+            "end_time": end_time,
+            "state": "STATE_ERRORED",
+            "experiment_config": self.experiment_config,
+            "experiment_id": self.experiment_id,
+            "hparams": self.hparams,
+            "determined_version": det.__version__,
+            "format": format,
+        }
+
+        r = req.put(self.master_url, "/api/v1/checkpoints", body={"checkpoint": ckpt})
```
stoksc, line 112: Should we maybe be consuming auto generated bindings for this API since they could exist?

harness/determined/util.py
```diff
@@ -129,6 +129,37 @@ def wrap_metrics(
         return {"metrics": metrics, "stop_requested": stop_requested, "invalid_hp": invalid_hp}
 
 
+def make_serializable(obj: Any) -> Any:
```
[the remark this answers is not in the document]
shiyuann: does it just support numpy?

harness/determined/_model_metric_manager.py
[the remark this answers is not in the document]
stoksc: Yeah but there are 3 context managers all with the same interface, __enter__, __exit__, complete, failed. What I'm saying is that there duplication there in the interface is indicative of the fact they're not independent - they're all workloads, and these 3 interfaces all handle/act on them like they're workloads.

harness/determined/util.py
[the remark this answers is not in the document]
shiyuann: It looks like it doesn't.

harness/determined/_model_metric_manager.py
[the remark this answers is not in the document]
stoksc: You could have a `WorkloadManager` that knows how workloads are started and then either errored or completed and just let it know how to construct that particular workload type.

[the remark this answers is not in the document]
shiyuann: We will be removing checkpointing/steps workloads. Also, the future workload won't be associated with just one checkpoint, validation, or training/step metric. So it doesn't make sense to have `BaseWorkloadManager`, which looks like the current data model.

harness/determined/util.py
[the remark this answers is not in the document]
stoksc: ah ok. i'm not super familiar with the code and I was wondering why we need new code for this --  why can't we use `json_encode` below then?

master/internal/api_checkpoint.go
shiyuann: it is not for debugging. We will be removing checkpointing workload in favor of logging for checkpointing itself.

harness/determined/util.py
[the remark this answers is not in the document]
shiyuann: `json_encode` doesn't "just" support numpy

harness/determined/_model_metric_manager.py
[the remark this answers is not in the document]
stoksc: > We will be removing checkpointing/steps workloads.

I mean, we removed searcher emitted checkpoints but we won't be removing the concept of checkpoint workloads in general right?

> Also, the future workload won't be associated with just one checkpoint, validation, or training/step metric.

How could a workload be associated with more than one? I'm confused.

harness/determined/util.py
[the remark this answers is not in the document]
stoksc: I don't understand, it has the same exact code for supporting numpy as the code you added. It is what is already used to encode the metrics as sent through the harness websocket.

master/internal/api_checkpoint.go
stoksc: let's talk about this, I was unaware we were removing checkpoint workloads.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
```diff
@@ -0,0 +1,85 @@
...
+  int32 start_batch = 6;
+  // End batch number of training metrics, equivalent to total batches
+  // for checkpoints and validations.
+  int32 end_batch = 7;
+  // Timestamp when this began being saved to persistent storage.
+  google.protobuf.Timestamp start_time = 8;
+  // Timestamp when this completed being saved to persistent storage.
+  google.protobuf.Timestamp end_time = 9;
+  // Metrics calculated on the training set.
+  google.protobuf.Struct metrics = 10;
```
rb-determined-ai, line 59: Looking at a normal step from `det t describe --metrics --json 7 | jq ".steps[0]"` on my cluster, I see:
```json
{
  "id": 1,
  "state": "COMPLETED",
  "metrics": {
    "num_inputs": 8,
    "avg_metrics": {
      "lr": 9.999999974752427e-07,
      "loss": 2668.724609375,
      "accuracy": 0
    },
    "batch_metrics": [
      {
        "lr": 9.999999974752427e-07,
        "loss": 5476,
        "accuracy": 0
      },
      {
        "lr": 9.999999974752427e-07,
        "loss": 4986.25732421875,
        "accuracy": 0
      },
      {
        "lr": 9.999999974752427e-07,
        "loss": 2601.38232421875,
        "accuracy": 0
      },
      {
        "lr": 9.999999974752427e-07,
        "loss": 2668.724609375,
        "accuracy": 0
      }
    ]
  },
  "end_time": "2021-02-20T00:37:20.441492+00:00",
  "trial_id": 7,
  "checkpoint": null,
  "start_time": "2021-02-20T00:37:14.777461+00:00",
  "validation": null,
  "num_batches": 4,
  "total_batches": 4,
  "prior_batches_processed": 0
}
```

In particular, the `metrics` object at the step level consists of the subfields:
* `num_inputs`
* `avg_metrics`
* `batch_metrics`

I don't think that e.g. our webui or the hpviz project would behave very nicely if the metrics were not packed that way.

Does it make sense to split these things out so we can ensure that we pack them that way in the database?  For example, the rest API could take `num_inputs`, `avg_metrics`, and `batch_metrics` separately, with e.g. `batch_metrics` being optional (I don't think we use those anywhere in the webui or backend, but @mackrorysd or @katport could probably correct me if I'm wrong on that).

harness/determined/__init__.py
```diff
@@ -17,5 +20,4 @@
     LoopTrialController,
     TrialController,
 )
-from determined import errors
```
rb-determined-ai, line 20: What is the goal of removing this line?

rb-determined-ai (commented):
I would like to know some more context about the goals of this change:

* are the managers (`CheckpointManager`, etc) meant to be public APIs at some point?  If not, what actually is their purpose?
* This PR adds a new codepath for putting metrics in the database.  That's great!  But it seems like the old codepath must also be followed for the purposes of the searcher and the overall trial state.  That seems ok for now, but when are we thinking about cleaning that up?

proto/src/determined/modelmetrics/v1/modelmetrics.proto
mackrorysd: batch_metrics being the individual batch datapoints from which avg_metrics is aggregated for each step, right? If so, no we're not using those for HP Viz. We use avg_metrics from training and validation metrics as-they-are.

harness/determined/_checkpoint_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        resources: Optional[Dict[str, Any]] = None,
+        metadata: Optional[Dict[str, Any]] = None,
+        framework: Optional[str] = None,
+        format: Optional[str] = None,
+    ) -> None:
+        logging.info(
+            f"Finishing saving checkpoint {uuid} for "
+            f"trial {self.trial_id} batch {self.batch_number}."
+        )
+        start_time = self.start_time.astimezone(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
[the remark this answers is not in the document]
shiyuann: good catch. I was intended to use the current UTC time but wrote it wrong.

harness/determined/_checkpoint_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        experiment_config: Optional[Dict[str, Any]] = None,
+        experiment_id: int = 0,
+        hparams: Optional[Dict[str, Any]] = None,
+    ):
+        self.master_url = master_url
+        self.trial_id = trial_id
+        self.batch_number = batch_number
+        self.experiment_config = experiment_config
+        self.experiment_id = experiment_id
+        self.hparams = hparams
```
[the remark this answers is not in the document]
shiyuann: removed them.

harness/determined/_checkpoint_manager.py
```diff
@@ -0,0 +1,122 @@
...
+            "format": format,
+        }
+
+        r = req.put(self.master_url, "/api/v1/checkpoints", body={"checkpoint": ckpt})
+        if not hasattr(r, "headers"):
+            raise Exception(r)
+
+        self.state = "STATE_COMPLETED"
+
+    def failed(self) -> None:
```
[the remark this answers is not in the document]
shiyuann: I don't understand why the master need to mark this as erroed. This PR is the first step to remove the checkpoint workload. Why does it have anything to do with workloads?

harness/determined/_checkpoint_manager.py
shiyuann: yes, we should. However, we don't have generated bindings now and this PR doesn't aim at changing anything about it.

harness/determined/_env_context.py
```diff
@@ -117,3 +117,58 @@ def per_slot_batch_size(self) -> int:
     @property
     def global_batch_size(self) -> int:
         return self._global_batch_size
+
```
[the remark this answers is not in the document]
shiyuann: I am just adding getters. It has nothing to do with making it public. I just don't want to duplicate code that transforms the raw string to trial id, experiment id, and etc.

harness/determined/_env_context.py
```diff
@@ -117,3 +117,58 @@ def per_slot_batch_size(self) -> int:
...
+            )
+        if name == "global_batch_size":
+            logging.warning(
+                "Please use `context.get_per_slot_batch_size()` and "
+                "`context.get_global_batch_size()` instead of accessing "
+                "`global_batch_size` directly."
+            )
+        return self.hparams[name]
+
+    def get_unary_host(self) -> str:
```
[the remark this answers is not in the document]
shiyuann: It returns the http/https master url rather than the websocket urls. So it's named "unary", which reuses the names in the master gRPC server.

harness/determined/_model_metric_manager.py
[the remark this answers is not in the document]
shiyuann: > I mean, we removed searcher emitted checkpoints but we won't be removing the concept of checkpoint workloads in general right?

We will.

> How could a workload be associated with more than one? I'm confused.

Workloads are just an abstraction for things that the master tells the harness to do. In the ideal world, the harness registers checkpoints, validations, and training metrics itself rather than creates a workload, executes it, and return an artifact for each workload. We will still keep the workloads of validation, add a training workload, and remove the step and checkpoint workloads.

harness/determined/util.py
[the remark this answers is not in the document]
shiyuann: Sorry, I gave a wrong explanation in my last comment. The requesting function `request.put()` requires the body to be a dictionary rather than a json string. I don't need the json encoding part `json_encode`.

master/internal/api_checkpoint.go
shiyuann: fixed.

master/internal/db/postgres.go
shiyuann: removed this since we already have the uniqueness constraint in the DB

master/internal/trial.go
shiyuann: >do you think we need a feature branch for the second half of this (moving the workload sequencer) and this to be lumped together into a single change? or maybe wait until you know you can hit them both in a single release?

We don't need a feature branch since it won't break anything. It won't be weird after removing the trial workload sequencer entirely from the master. I don't want to do all the changes at once.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
shiyuann: I think `num_inputs` is totally unnecessary since the data itself contains the starting batch number and end batch number. And `avg_metrics` should be renamed to `reduced_metrics`. 

> Does it make sense to split these things out so we can ensure that we pack them that way in the database?

In other parts of the system, we put these three things in one object. It's better to be consistent. Also, I think the current schema is more flexible.

shiyuann (commented):
@rb-determined-ai 

> I would like to know some more context about the goals of this change:
> are the managers (CheckpointManager, etc) meant to be public APIs at some point? If not, what actually is their purpose?

This goal has nothing to do with adding new APIs. The added APIs are tagged with "Internal" if you noticed that in `api.proto`. If we want to add new APIs or modify the APIs for exposing them to users, it won't break users.

> This PR adds a new code path for putting metrics in the database. That's great! But it seems like the old code path must also be followed for the purposes of the searcher and the overall trial state. That seems ok for now, but when are we thinking about cleaning that up?

Right. I will have a follow-up PR to remove the trial workload sequencer from the master, which should clean up the old code path.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
rb-determined-ai: I think `num_inputs` is important because it's really about records, not batches, and you can't count that from the master side accurately.

master/internal/db/postgres.go
stoksc: Forgot it already existed.

master/internal/trial.go
stoksc: I understand not doing them all at once. I think the approach for that is good. Just want to make sure no in between parts fall through the cracks.

harness/determined/_checkpoint_manager.py
stoksc: Yeah, I know we're starting to consume the API all over the place.. we should get that figured out soon. I'll put it on th infra-ag agenda.

harness/determined/util.py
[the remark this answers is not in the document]
stoksc: I'd rather have a `put_json` than write this code twice. Or at least pull out a common func?

proto/src/determined/modelmetrics/v1/modelmetrics.proto
shiyuann: @rb-determined-ai  I still don't understand why `num_inputs` is needed on the master. Can you explain that?

harness/determined/util.py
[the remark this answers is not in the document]
shiyuann: fixed

shiyuann:
@rb-determined-ai @stoksc I updated the PR on these changes:

1. move the error handling logic back to the trial actor.
2. remove the use of context manager on the checkpointing and metrics APIs.
3. fix the issues mentioned in the commens

harness/determined/_checkpoint_manager.py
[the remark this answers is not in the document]
shiyuann: Fix it to let the master handle errors.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
shiyuann: @rb-determined-ai I keep what the current state as the master is. Maybe it makes more sense if we expose the APIs to the users in the future.

rb-determined-ai: It's just useful information for visualizations and such.  It's not needed for the experiment/trial/searcher actors or anything.

harness/determined/_env_context.py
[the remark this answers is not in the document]
rb-determined-ai: :+1:

harness/determined/layers/_workload_manager.py
```diff
@@ -274,59 +308,130 @@ def _respond(in_response: workload.Response) -> None:
...
+                    "uuid": uuid,
+                    "resources": resources,
+                    "metadata": metadata.__dict__,
+                    "format": format,
+                    "framework": framework,
+                    "determined_version": det.__version__,
+                }
+                req.put(
+                    self.env.get_unary_host(), "/api/v1/checkpoints", body={"checkpoint": checkpoint}
+                )
```
rb-determined-ai, line 400: This is a bug: the checkpoint is being reported before it is being actually uploaded.  This `put` must occur after the `with self.storage_mgr.store_path() as (uuid, path):` context manager exits.

(I only recognize this bug because I made the exact same mistake before : )

harness/determined/workload.py
```diff
@@ -58,6 +58,9 @@ def from_json(dict: Dict[str, Any]) -> "Workload":
             dict["total_batches_processed"],
         )
 
+    def total_batch(self) -> int:
```
rb-determined-ai, line 61: Can we rename this to `target_batch()` maybe?  I don't think `total_batch()` communicates the right thing.

harness/tests/experiment/utils.py
```diff
@@ -145,6 +145,7 @@ def make_default_env_context(
         det_experiment_id="1",
         det_cluster_id="uuid-123",
         trial_seed=trial_seed,
+        test_mode=True,
```
rb-determined-ai, line 148: This seems like a very aggressive change... won't this prevent us from training or validating more than a single batch in all of our unit tests?

harness/determined/layers/_workload_manager.py
```diff
@@ -143,12 +135,14 @@ def _respond(in_response: workload.Response) -> None:
             metrics = in_response["metrics"]
             metrics = cast(workload.Metrics, metrics)
 
+            end_time = _current_timestamp()
+
             if in_response.get("invalid_hp", False):
                 out_response = {
                     "type": "WORKLOAD_COMPLETED",
                     "workload": wkld,
                     "start_time": start_time,
```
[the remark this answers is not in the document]
shiyuann: : ) Python is dynamic. There is no compiling check on if a variable is defined in the code. It just goes to the memory for the variable ``end_time``. It's all decided in the runtime.

harness/determined/layers/_workload_manager.py
shiyuann: : ) thanks for pointing it out! This makes me hate the response function more.

proto/src/determined/api/v1/api.proto
```diff
@@ -726,6 +728,54 @@ service Determined {
...
+      post: "/api/v1/training_metrics"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "Internal"
+    };
+  }
+
+  // Update training metrics.
+  rpc UpdateTrainingMetrics(UpdateTrainingMetricsRequest)
```
stoksc, line 744: I would call these `Patch`s since they only update the parts provided. Also, fwiw, with that scheme as opposed to using field masks / an actual patch, we can't differentiate update to default from unprovided. Not sure I see a huge issue with that, but it's something to think about.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
stoksc: I agree with ryan, without `num_inputs` we will try to infer it from batches and we will be wrong.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
```diff
@@ -0,0 +1,85 @@
...
+package determined.modelmetrics.v1;
+option go_package = "github.com/determined-ai/determined/proto/pkg/modelmetricsv1";
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+import "protoc-gen-swagger/options/annotations.proto";
+
+// The current state of the metrics.
+enum State {
+  // The state of the metrics is unknown.
```
stoksc, line 12: Semantically, this is all kinda funny, I agree with your earlier points. It's weird for metrics to have state. Maybe these comments should be worded like `The state of the workload associated with some batch of metrics...`.

stoksc (commented):
mostly lgtm, i'll let @rb-determined-ai and you finish open discussions before a last pass.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
shiyuann: Doesn't make any difference if you just change a comment here. : ) And we are removing workloads. So it doesn't make sense to mention workloads either.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
shiyuann: @rb-determined-ai @stoksc 

`num_inputs` is actually inferred across the codebase based on a wrong assumption that all batches have the same size.

To me, I don't think having `num_inputs` is useful or used at all. Users might want to pack any arbitrary fields as they want if we support initializing metrics in the harness. The packing schema that doesn't make the visualizations work well sound like bugs on the WebUI and CLI side.

This PR doesn't aim at providing any user-facing APIs, fixing any wrong assumptions, or adding any new constraints to packing metrics. Before we have a concrete consensus on what should be done, I won't change any behavior in this PR.

harness/determined/layers/_workload_manager.py
```diff
@@ -171,23 +165,63 @@ def _respond(in_response: workload.Response) -> None:
...
                 out_response["exited_reason"] = "USER_CANCELED"
 
-            # Send the response up.
+            if not self.env.test_mode:
+                training_metrics = {
+                    "trial_id": self.env.get_trial_id(),
+                    "step_id": wkld.step_id,
+                    "start_batch": wkld.total_batches_processed,
+                    "end_batch": wkld.total_batch(),
+                    "start_time": start_time.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
```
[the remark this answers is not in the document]
stoksc: > Z:           A suffix which, when applied to a time, denotes a UTC
                  offset of 00:00; often spoken "Zulu" from the ICAO
                  phonetic alphabet representation of the letter "Z".

From the RFC for this format, so it looks weird but it's normal. Also, from what I see it doesn't:

```
➜  bulk-insert-gin python
Python 3.6.12 (default, Jan 11 2021, 14:46:48) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.59)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.utcnow().isoformat() + "Z"
'2021-03-05T15:24:36.193631Z'
```

Either way 🤷 no biggie, just leave it as it. It was just a suggestion in reply to Ryan's comment.

proto/src/determined/modelmetrics/v1/modelmetrics.proto
stoksc: > `num_inputs` is actually inferred across the codebase based on a wrong assumption that all batches have the same size.

Yeah, it is master side but it isn't harness side where this comes from right?

PR #1969, opened 2021-02-16
fix: rolling back trials when restarting a trial [DET-4994]

shiyuann:
## Description

Fix rolling back trials when restarting a trial. The bug was introduced in [this commit](https://github.com/determined-ai/determined/commit/07be51016588165535e7c1dab58a68746a0ececb), which should use the current active step totalBatches rather than the last completed step.

## Test Plan

Manually tested it

## Checklist


stoksc:
@shiyuann I opened #1970 for this.

PR #1970, opened 2021-02-16
fix: restore trials with operations intact to avoid unintended behavior [DET-4994]

stoksc:
This commit changes how trial operations snapshotted in the experiment state are given to trials during a restore. Previous, the experiment would sent all these operations to trials after starting, now they are added to the trial before it is restored.

## Test Plan
- [x] run adaptive chaos and restore it,

## Commentary (optional)
The only part really necessary for this fix is the bit that changes how we were calling `s.Workload()` in `t.reset()`. The other bits are because, the way restart was written, we were starting a trial in an invalid state (without its operations) which I realized later (now) is totally wrong.

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -187,7 +187,11 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			Handler:  ctx.Self(),
 		})
 
-		e.processOperations(ctx, e.searcher.TrialOperations, nil)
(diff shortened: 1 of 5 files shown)


rb-determined-ai (approved):
I like this a lot.  This is more like what I was asking for in the master restart PR when I asked for trials to be created with their restorable state, right in newTrial().

PR #1972, opened 2021-02-16
chore: remove searcher emitted checkpoints [DET-4996]

stoksc:
## Description
This change removes searcher emitted checkpoints and instead changes PBT (the only searcher that relies on them) to just be sure a checkpoint will be there anyways (since it will).

## Test Plan
- [x] PBT ain't broke (run PBT on prior version, pause, upgrade, unpause, PBT continues)

## Commentary (optional)
Yay, I've wanted to do this forever.


stoksc:
This is blocked on #1970.

master/static/migrations/20210216185819_shim-pbt-snapshots.up.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE experiment_snapshots
+SET content = content #- '{searcher_state,search_method_state,waiting_checkpoints}';
```
stoksc, line 2: Making a breaking change to snapshots is as easy as this.

master/pkg/searcher/operations.go
```diff
@@ -144,14 +135,14 @@ func NewCreate(
 	}
 }
 
-// NewCreateFromCheckpoint initializes a new Create operation with a new request ID and the given
+// NewCreateFromParent initializes a new Create operation with a new request ID and the given
 // hyperparameters and checkpoint to initially load from.
-func NewCreateFromCheckpoint(
-	rand *nprand.State, s hparamSample, checkpoint Checkpoint,
+func NewCreateFromParent(
```
shiyuann, line 140: q: does parent mean parent trial?

master/internal/restore.go
```diff
@@ -119,6 +119,7 @@ func (e *experiment) restoreTrial(
 	if trialID != nil {
 		t.processID(*trialID)
 	}
+	t.processOperations(ops)
```
shiyuann, line 122: q: why moving this function here?

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -35,9 +35,6 @@ enum RunnableType {
   RUNNABLE_TYPE_TRAIN = 1;
   // Signals to a trial runner it should compute validation metrics.
   RUNNABLE_TYPE_VALIDATE = 2;
-  // Signals to the trial runner that the current model state should be
-  // checkpointed.
-  RUNNABLE_TYPE_CHECKPOINT = 3;
```
shiyuann, line 40: q: will this be breaking?

shiyuann (commented):
Praise: didn't expect this to be so clean after removing cachedcheckpoint. Nice work!

master/pkg/searcher/operations.go
stoksc: ah, originally i changed the create to have a parent request ID instead of a checkpoint, but then i realized it caused issues (would need shim all the saved creates for PBT) so I decided not to do it.

master/internal/restore.go
stoksc: ah, this was just so the ops would be there in the restore call, just incase it needed them. it doesn't matter though since it doesnt use it.

proto/src/determined/experiment/v1/searcher.proto
stoksc: yeah, but only breaking for `det preview-search` (which reminds me there is CLI code I can remove too)

stoksc: breaking for `det preview-search` consumed via the new rest api (which I think we're _probably_ safe on)

master/internal/api_experiment.go
```diff
@@ -236,10 +236,6 @@ func (a *apiServer) PreviewHPSearch(
 			return experimentv1.RunnableOperation{
 				Type: experimentv1.RunnableType_RUNNABLE_TYPE_VALIDATE,
 			}, nil
-		case searcher.Checkpoint:
```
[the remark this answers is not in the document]
stoksc: done

master/pkg/searcher/operations.go
stoksc: I mean, I'm still going to use the parent ID, I just decided to leave the checkpoint wrapper around it to save on breakage. I can change it and shim it also if you think we should.

master/pkg/searcher/operations.go
```diff
@@ -28,11 +28,10 @@ type (
 
 // All the operation types that support serialization.
 const (
-	CreateOperation OperationType = iota
-	TrainOperation
-	ValidateOperation
-	CheckpointOperation
-	CloseOperation
+	CreateOperation   OperationType = 0
```
stoksc, line 31: iotas are bad and lead to bugs, i'm totally convinced now (in some specific cases, or if you're an idiot like me)

master/static/migrations/20210216185819_shim-pbt-snapshots.down.sql
```diff
@@ -0,0 +1 @@
+-- Down is impossible, data was removed.
```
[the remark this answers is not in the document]
stoksc: removed this in favor of shimming in go

master/internal/restore.go
```diff
@@ -180,3 +189,93 @@ func (e *experiment) snapshotAndSave(ctx *actor.Context, ts trialSnapshot) {
...
+type snapshotShimFunc func([]byte) ([]byte, error)
+
+// identidyShim is a shim that does nothing.
+func identidyShim(snapshot []byte) ([]byte, error) {
+	return snapshot, nil
+}
+
+// Version 0 => 1 shims
+
+func shimExperimentSnapshotV0(snapshot []byte) ([]byte, error) {
```
rb-determined-ai, line 244: Probably worth including some context in a comment about when/why we wrote this shim.  Details like:
* we used to have searcher checkpoints
* now checkpoints are always created before validations, so they don't need to be explicit
* only pbt cared about searcher checkpoints completing anyway
* this is moving us in the direction of a world where checkpoints are pushed to the searcher, rather than asked for by the searcher

Normally, it's good practice to avoid comments which are true at a specific point in time (anything that starts with "it used to be like X but now ..."), but since this shim totally corresponds to a specific point in time, I think this would be pretty helpful.

rb-determined-ai (approved):

PR #1973, opened 2021-02-17
chore: tune parcoords [DET-4991]

hkang1:
## Description

Tasks
- add filtered table to PC plot
- show all trials and then filter based off window
- make column order reset to default order (unable to lock or prevent the columns from being re-ordered)
- fiddle with the labels to make them stand out but without taking even more space (angle them to 45)
- truncate the hp label to 20 characters for long labels
- limit the hp to 10
- add tips on how to filter on parallel coordinates

Thought process on color scale gradients
- avoid red/green due to it being the most common color blindness
- use blue/red color scale for searcher metric, for when we do know if a metric is good or bad within a range. The color gradient is reversed when `smallerIsBetter` is true.
- use blue/orange color scale for non-searcher metric to avoid using red, which generally suggests something is bad. orange for neutral color that sharply contrasts against blue being on the opposite side of the color wheel from blue

chart with searcher metric selected
Screen Shot 2021-02-16 at 4 46 34 PM

chart with non-searcher metric selected
Screen Shot 2021-02-16 at 5 16 15 PM

filtered applied
Screen Shot 2021-02-16 at 5 22 45 PM

## Checklist

### DIFF
--- a/webui/react/src/utils/color.test.ts
+++ b/webui/react/src/utils/color.test.ts
@@ -1,0 +1,129 @@
+import { GLASBEY } from 'constants/colors';
+
+import * as util from './color';
+
+describe('color utility', () => {
+  const colors = [
+    {
+      hex: '#000000',
+      hsl: { h: 0, l: 0, s: 0 },
+      hslStr: 'hsl(0, 0%, 0%)',
+      rgb: { b: 0, g: 0, r: 0 },
+      rgbStr: 'rgb(0, 0, 0)',
+    },
+    {
+      hex: '#999999',
+      hsl: { h: 0, l: 60, s: 0 },
+      hslStr: 'hsl(0, 0%, 60%)',
+      rgb: { b: 153, g: 153, r: 153 },
+      rgbStr: 'rgb(153, 153, 153)',
+    },
+    {
+      hex: '#ffffff',
+      hsl: { h: 0, l: 100, s: 0 },
+      hslStr: 'hsl(0, 0%, 100%)',
+      rgb: { b: 255, g: 255, r: 255 },
+      rgbStr: 'rgb(255, 255, 255)',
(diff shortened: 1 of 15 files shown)


webui/react/src/utils/color.test.ts
```diff
@@ -0,0 +1,129 @@
+import { GLASBEY } from 'constants/colors';
```
hamidzr, line 1: 👍🏼

webui/react/src/utils/color.ts
```diff
@@ -1,35 +1,61 @@
...
   s: number;
 }
 
+/*
+ * r - red between 0 and 255
+ * g - green between 0 and 255
+ * b - blue between 0 and 255
+ * a - alpha between 0.0 and 1.0
+ */
+interface RgbaColor {
```
hamidzr, line 20: are there libraries we could use that do these color conversions for us?

hamidzr (commented):
other than some issues with setting and moving hp filter ranges this looks good to me. while I edit the range and drag it up and down sometimes there range indicator jumps backward

webui/react/src/utils/color.ts
hkang1: this one looks promising but it adds a decent amount in size (7.6kB) for something so small...

https://www.npmjs.com/package/color
https://bundlephobia.com/result?p=color@3.1.3

hamidzr: hmm since you've added a test for it as well using our own implementation sgtm

hamidzr (approved):
LGTM. We talked about the issue of the hp filter shifting under the cursor goes back to rerendering while the user is moving the filter and it can be helped by eliminating/reducing rerenders (in another pr)

PR #1974, opened 2021-02-17
chore: refactor trial log viewer to improve rendering performance [DET-4866]

apizzini:
This is still a WIP because there's an issue in the follow=true API that I prefer to check before this get's merged.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewerLevel.tsx
+++ b/webui/react/src/components/LogViewerLevel.tsx
@@ -11,19 +11,22 @@ import css from './LogViewer.module.scss';
 export const ICON_WIDTH = 26;
 
 interface Props {
-  logLevel: LogLevel|undefined;
+  logLevel?: LogLevel;
 }
 
 const LogViewerLevel: React.FC<Props> = ({ logLevel }) => {
   const levelStyle = { width: toRem(ICON_WIDTH) };
 
-  const classes = [ css.level ];
-  if (logLevel) classes.push(css[logLevel]);
+  if (!logLevel) {
+    logLevel = LogLevel.Info;
+  }
+
+  const classes = [ css.level, logLevel ];
 
   return (
-    <Tooltip placement="top" title={`Level: ${capitalize(logLevel || '')}`}>
+    <Tooltip placement="top" title={`Level: ${capitalize(logLevel)}`}>
       <div className={classes.join(' ')} style={levelStyle}>
-        <div className={css.levelLabel}>&lt;[{logLevel || ''}]&gt;</div>
+        <div className={css.levelLabel}>&lt;[{logLevel}]&gt;</div>
         <Icon name={logLevel} size="small" />
       </div>
     </Tooltip>
--- a/webui/react/src/components/LogViewerLevel.tsx
+++ b/webui/react/src/components/LogViewerLevel.tsx
@@ -1,0 +1,33 @@
+import { Tooltip } from 'antd';
+import React from 'react';
+
+import { LogLevel } from '../types';
+import { toRem } from '../utils/dom';
+import { capitalize } from '../utils/string';
+
+import Icon from './Icon';
+import css from './LogViewer.module.scss';
+
+export const ICON_WIDTH = 26;
+
+interface Props {
+  logLevel: LogLevel;
+}
+
+const LogViewerLevel: React.FC<Props> = ({ logLevel }) => {
+  const levelStyle = { width: toRem(ICON_WIDTH) };
+
+  const classes = [ css.level ];
+  if (logLevel) classes.push(css[logLevel]);
+
+  return (
+    <Tooltip placement="top" title={`Level: ${capitalize(logLevel || '')}`}>
+      <div className={classes.join(' ')} style={levelStyle}>
+        <div className={css.levelLabel}>&lt;[{logLevel || ''}]&gt;</div>
+        <Icon name={logLevel} size="small" />
+      </div>
+    </Tooltip>
+  );
+};
+
+export default LogViewerLevel;
--- a/webui/react/src/components/LogViewerLevel.tsx
+++ b/webui/react/src/components/LogViewerLevel.tsx
@@ -11,7 +11,7 @@ import css from './LogViewer.module.scss';
 export const ICON_WIDTH = 26;
 
 interface Props {
-  logLevel: LogLevel;
+  logLevel: LogLevel|undefined;
 }
 
 const LogViewerLevel: React.FC<Props> = ({ logLevel }) => {
--- a/webui/react/src/components/LogViewerStore.ts
+++ b/webui/react/src/components/LogViewerStore.ts
@@ -1,0 +1,40 @@
+import { TrialLog } from '../types';
+
+export interface ViewerLog extends TrialLog {
+  formattedTime: string;
+}
+
+export enum LogStoreActionType {
+  Clear,
+  Append,
+  Prepend,
+}
+
+export type LogStoreAction =
+  | { type: LogStoreActionType.Clear; }
+  | { type: LogStoreActionType.Append; value: ViewerLog[] }
+  | { type: LogStoreActionType.Prepend; value: ViewerLog[] };
+
+const logUniqueFilter = (log: ViewerLog, index: number, self: ViewerLog[]) => {
+  return self.map(mapObj => mapObj.id).indexOf(log.id) === index;
+};
+
+export const logStoreReducer = (state: ViewerLog[], action: LogStoreAction): ViewerLog[] => {
+  switch (action.type) {
+    case LogStoreActionType.Clear: {
+      return [];
+    }
+    case LogStoreActionType.Append:
+      return [
+        ...state,
+        ...action.value,
+      ].filter(logUniqueFilter);
+    case LogStoreActionType.Prepend:
+      return [
+        ...action.value,
+        ...state,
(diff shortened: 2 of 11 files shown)


webui/react/src/utils/dom.ts
```diff
@@ -18,6 +18,26 @@ export const toRem = (px?: number | string): string => {
...
+  if (matches?.length === 3) {
+    const type = matches[2];
+    const value = parseFloat(matches[1]);
+    if (type === 'rem') return `${value * 10}px`;
+    if (type === 'px') return `${value}px`;
+  }
+
+  return rem;
+};
+
```
hkang1, line 40: **suggestion:** I've been bad at this but trying to be better about writing tests. Can you write some unit tests for `toPixel` in `dom.test.ts`? 🙏

webui/react/src/utils/dom.ts
```diff
@@ -18,6 +18,26 @@ export const toRem = (px?: number | string): string => {
...
+/*
+ * Calculates Pixel from Rem based on base font size of 62.5%.
+ * This causes the ratio between REM to px to be 1 to 10.
+ * So 1rem becomes 10px.
+ */
+export const toPixel = (rem?: number | string): string => {
+  if (rem == null) return 'auto';
+  if (typeof rem === 'number') return `${rem * 10}px`;
+
+  const matches = rem.match(/(\d+\.?\d*)\s*(px|rem)/i);
```
hkang1, line 30: **suggestion:** make it also support something like `.5rem` where there isn't a digit in front of a decimal. Something like this (untested):
```
/(\d+\.?\d*|\.\d+)\s*(px|rem)/i
```

webui/react/src/pages/TrialLogs.tsx
```diff
@@ -149,19 +149,18 @@ const TrialLogs: React.FC = () => {
     return <Message title={`Unable to find Trial ${trialId}`} type={MessageType.Warning} />;
   }
 
-  const experimentDetailPath = `/experiments/${trial.data?.experimentId}`;
+  const experimentDetailPath = `/experiments/${experimentId}`;
   const trialDetailPath = `${experimentDetailPath}/trials/${trialId}`;
```
hkang1, line 153: **suggestion:** under `routes/utils.ts` there is an object `paths` with functions to generate paths. Could you add these to that and use that instead? There is one for `trialDetails` already so far, but could use an `experimentDetails` one too.

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -1,27 +1,28 @@
...
+import { LogViewerTimestampFilterComponentProp } from 'pages/TrialLogs/TrialLogFilters';
 import { FetchArgs } from 'services/api-ts-sdk';
 import { consumeStream } from 'services/utils';
 import { LogLevel, TrialLog } from 'types';
 import { formatDatetime } from 'utils/date';
-import { ansiToHtml, copyToClipboard, toRem } from 'utils/dom';
-import { capitalize } from 'utils/string';
+import { ansiToHtml, copyToClipboard, toPixel, toRem } from 'utils/dom';
+
+import useGetCharMeasureInContainer from '../hooks/useGetCharMeasureInContainer';
```
hkang1, line 21: **nit:** use absolute path of just `/hooks/useGetCharMeasureInContainer`

webui/react/src/components/LogViewerLevel.tsx
```diff
@@ -0,0 +1,33 @@
...
+const LogViewerLevel: React.FC<Props> = ({ logLevel }) => {
+  const levelStyle = { width: toRem(ICON_WIDTH) };
+
+  const classes = [ css.level ];
+  if (logLevel) classes.push(css[logLevel]);
+
+  return (
+    <Tooltip placement="top" title={`Level: ${capitalize(logLevel || '')}`}>
+      <div className={classes.join(' ')} style={levelStyle}>
+        <div className={css.levelLabel}>&lt;[{logLevel || ''}]&gt;</div>
```
hkang1, line 26: **nit:** I know this is code from before, but can `logLevel` ever be `undefined` or `null`? Basically can we just drop the ` || ''` fallback portion?

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -1,27 +1,28 @@
 import { Button, notification, Space, Tooltip } from 'antd';
 import dayjs, { Dayjs } from 'dayjs';
 import React, {
-  useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState,
+  Reducer, RefObject,
+  useCallback, useEffect, useLayoutEffect, useMemo, useReducer, useRef, useState,
 } from 'react';
+import { ListChildComponentProps, ListOnItemsRenderedProps, VariableSizeList } from 'react-window';
```
hkang1, line 7: sweeeetttt

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -109,167 +103,155 @@ const LogViewerTimestamp: React.FC<Props> = ({
...
+    }, [ addLogs, fetchToLogConverter, onFetchLogAfter, onFetchLogBefore ]);
+
+  const getItemHeight = (index: number): number => {
+    const log = logs[index];
+    if (!log) {
+      return charMeasures.height;
+    }
+
+    const columnWidth = listMeasure.width - ICON_WIDTH - dateTimeWidth;
+    const messageWidth = log.message.length * charMeasures.width;
```
hkang1, line 209: **suggestion:** One caveat to this is that it will miss the case of when there is a newline within the message itself. I think I previously calculated it by splitting up the message based on the newlines, calculate how much height each part takes up and adding them up.

webui/react/src/components/LogViewerStore.ts
```diff
@@ -0,0 +1,40 @@
+import { TrialLog } from '../types';
```
hkang1, line 1: **nit:** '/types' without '..'

hkang1: **suggestion:** how about calling this file `LogViewer.store.ts`? To follow the pattern of a the component that is supports (LogViewer), followed by it's function (store). Similar to `X.table.tsx` (although it should be `ts` instead of `tsx` for tables)

hkang1:
Overall this looks really good and the trial log viewer feels really snappy to me! I came across a couple of visual issues:

2021-02-18 at 2 34 PM

1) the icon is showing a star instead of the info icon for a default message type
2) the height seems to be off by one, that's noticeable in smaller / narrower view. I commented more specifics about this near the part of the code where the item height is calculated.

webui/react/src/utils/dom.ts
apizzini: great feedback, thanks!

webui/react/src/components/LogViewerLevel.tsx
apizzini: Reading TS interfaces (in particular `TrialLog` from `webui/react/src/types.ts`) it is set to be optional, so that's possible, but on the other side `LogViewerLevel` `props` are setting `logLevel` as required.

So surely this is needing a fix in one of the 2 places.

But looking how API response is converted (`jsonToTrialLog` from `webui/react/src/services/decoder.ts`), `Sdk.V1TrialLogsResponse` is setting `level` as required, but `V1LogLevel` !== `LogLevel`, in fact `decodeV1LogLevelToLogLevel` could potentially return `undefined`.

So I guess we should update TS to set the `undefined` case as expected and keep the fallback? (I did that).

apizzini: But given the other feedback (star instead of the info icon for a default message type) we could simply default to info if undefined?

webui/react/src/components/LogViewerTimestamp.tsx
apizzini: Whoops, I thought about this and forgot it! Thanks!

apizzini:
Thanks for the deep review @hkang1! There should be just 1 thing left to fix, which is how to handle the logMessage being possibly undefined (see replies to [this conv](https://github.com/determined-ai/determined/pull/1974#discussion_r578764837)).

webui/react/src/components/LogViewerLevel.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { toRem } from '../utils/dom';
+import { capitalize } from '../utils/string';
+
+import Icon from './Icon';
+import css from './LogViewer.module.scss';
+
+export const ICON_WIDTH = 26;
+
+interface Props {
+  logLevel: LogLevel|undefined;
```
hkang1, line 14: **nit:** this can just be
```
  logLevel?: LogLevel;
```

webui/react/src/components/LogViewerLevel.tsx
hkang1: > we could simply default to info if undefined?

I think the info icon as the default works.

webui/react/src/utils/dom.test.ts
```diff
@@ -1,4 +1,4 @@
-import { ansiToHtml, toRem } from './dom';
+import { ansiToHtml, toPixel, toRem } from './dom';
```
hkang1, line 1: excellent thanks!

hkang1 (approved):
LGTM! Nice work. Just one minor nit in the `Props` section.

PR #1978, opened 2021-02-17
chore: check if experiment exists in /api/v1/experiments/:id/trials

stoksc:


master/internal/api_trials.go
```diff
@@ -311,6 +311,14 @@ func (a *apiServer) KillTrial(
 func (a *apiServer) GetExperimentTrials(
 	_ context.Context, req *apiv1.GetExperimentTrialsRequest,
 ) (*apiv1.GetExperimentTrialsResponse, error) {
+	ok, err := a.m.db.CheckExperimentExists(int(req.ExperimentId))
```
shiyuann, line 314: It's weird that line 369 doesn't catch the experiment id. maybe fix that place?

shiyuann (approved):
Looks good!

master/internal/api_trials.go
stoksc: discussed offline, queryproto'ing a slice will never return not found so it's sort of impossible to fix that way without changing that.

PR #1979, opened 2021-02-17
fix: k8s resource pool API response should not have nil field [DET-4989]

armandmcqueen:
## Description

Change the "details" field in the Resource Pool GET API to be an empty struct instead of nil to be consistent with the non-kubernetes behavior. This fixes a bug where the Cluster page in the WebUI would crash because it expects that field to exist in the response.

## Test Plan

Manually tested on a k8s cluster.

## Checklist


hamidzr (approved):
LGTM. 

details is set in boths places now `master/internal/resourcemanagers/agent_resource_manager.go`

PR #1991, opened 2021-02-19
fix: call Sequence.on_epoch_end after validation [DET-5007]

rb-determined-ai:
Imitate keras more exactly.

Tested with user-provided repro example.

### DIFF
--- a/docs/release-notes/1991-keras-validation-end.txt
+++ b/docs/release-notes/1991-keras-validation-end.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Fixes**
+
+-  Ensure that ``keras.utils.Sequence`` objects receive their
+   ``on_epoch_end()`` calls after validation is completed.
--- a/harness/determined/keras/_tf_keras_trial.py
+++ b/harness/determined/keras/_tf_keras_trial.py
@@ -350,6 +350,19 @@ class TFKerasTrialController(det.LoopTrialController):
         self.training_data = train_config.training_data
         self.validation_data = train_config.validation_data
 
+        # Support the deprecated SequenceAdapter API.
+        if isinstance(self.training_data, keras.SequenceAdapter):
+            self.context._configure_fit(
+                workers=self.training_data.workers,
+                use_multiprocessing=self.training_data.use_multiprocessing,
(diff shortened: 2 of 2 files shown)


docs/release-notes/1991-keras-validation-end.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Fixes**
+
+-  Ensure that ``keras.utils.Sequence`` objects recieve their
```
davetroiano, line 5: receive

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -760,6 +758,18 @@ def _launch_evaluate(self) -> Any:
...
+        if isinstance(self.validation_data, tf.keras.utils.Sequence):
+            enqueuer.stop()
+            self.enqueuers.remove(enqueuer)
+
+            # A special side-effect of converting the keras sequence to a generator and passing
+            # steps explicitly is that keras will exit our generator after N steps and the
+            # Sequence.on_epoch_end() that normally runs after the last yield won't run at all
+            # because the fit loop will call next() exactly `steps` times.  So we try to match the
+            # exact keras behavior by manually calling on_epoch_end() here.
+            self.validation_data.on_epoch_end()
```
davetroiano, line 771: Can there ever be a case where calling this here is wrong? e.g., users call configure_fit w/ steps_per_epoch and validation_steps specified, such that partial epoch or multiple epochs are actually validated per validation workload?  (it seems fine if so but it's worth public documentation if there are any corner cases)

rb-determined-ai: I don't think that's possible.

This is the signature of configure_fit:
```python
    def configure_fit(
        self,
        verbose: Optional[bool] = None,
        class_weight: Any = _arg_not_provided,
        workers: Optional[int] = None,
        use_multiprocessing: Optional[bool] = None,
        max_queue_size: Optional[bool] = None,
        shuffle: Optional[bool] = None,
        validation_steps: Any = _arg_not_provided,
    ) -> None:
```
and `validation_steps` only applies to tf.data.Dataset objects (per the keras standard behavior).

davetroiano (approved):

PR #1992, opened 2021-02-19
refactor: remove latest checkpoint internal state

shiyuann:
## Description

Remove the latest checkpoint field in the trial state since the trial can always get that from the DB.

## Test Plan

N/A

## Commentary (optional)

N/A

## Checklist

### DIFF
--- a/master/internal/trial_workload_sequencer.go
+++ b/master/internal/trial_workload_sequencer.go
@@ -29,8 +29,7 @@ type (
 		CurOpIdx  int `json:"cur_op_idx"`
 		CurStepID int `json:"cur_step_id"`
 
-		LatestCheckpoint *model.Checkpoint            `json:"latest_checkpoint"`
-		LatestSnapshot   *trialWorkloadSequencerState `json:"latest_snapshot"`
+		LatestSnapshot *trialWorkloadSequencerState `json:"latest_snapshot"`
 	}
 
 	// trialWorkloadSequencer manages transforming the work requested by the searcher into Workloads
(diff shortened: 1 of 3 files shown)


master/internal/trial_workload_sequencer.go
```diff
@@ -29,8 +29,7 @@ type (
 		CurOpIdx  int `json:"cur_op_idx"`
 		CurStepID int `json:"cur_step_id"`
 
-		LatestCheckpoint *model.Checkpoint            `json:"latest_checkpoint"`
-		LatestSnapshot   *trialWorkloadSequencerState `json:"latest_snapshot"`
+		LatestSnapshot *trialWorkloadSequencerState `json:"latest_snapshot"`
```
stoksc, line 32: You're going to need to write a shim for this state change.

stoksc (commented):
lgtm but I think we need to shim the change to `trialWorkloadSequencerState` or else on upgrade trials will panic.

master/internal/trial_workload_sequencer.go
shiyuann: Ohhh, I saw you didn't include the shimming code for `cached_checkpoint` so I thought it not necessary to do so. Am I misunderstanding your code?

stoksc: Ah I think I should've removed that too, oops. Do you mind adding it to my shim? You can piggy back this change into my shim soon since they're going to go in a release together.

shiyuann: Okay. I'll include that too in this PR.

PR #1999, opened 2021-02-22
chore: make experiment list page url sharable [DET-4745]

apizzini:
## Checklist

### DIFF
--- a/webui/react/src/components/Navigation.tsx
+++ b/webui/react/src/components/Navigation.tsx
@@ -37,7 +37,7 @@ const NavigationItem: React.FC<ItemProps> = ({ path, status, ...props }: ItemPro
   useEffect(() => setIsActive(location.pathname === path), [ location.pathname, path ]);
 
   const link = (
-    <Link className={classes.join(' ')} path={path} {...props}>
+    <Link className={classes.join(' ')} disabled={isActive} path={path} {...props}>
       <Icon name={props.icon} size="large" />
       <div className={css.label}>{props.label}</div>
       {status && <div className={css.status}>{status}</div>}
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -26,6 +26,7 @@ import handleError, { ErrorLevel, ErrorType } from 'ErrorHandler';
 import useExperimentTags from 'hooks/useExperimentTags';
 import usePolling from 'hooks/usePolling';
 import useStorage from 'hooks/useStorage';
+import { parseUrl } from 'routes/utils';
 import {
   activateExperiment, archiveExperiment, cancelExperiment, getExperiments,
   killExperiment, openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
@@ -55,6 +56,8 @@ enum Action {
   Unarchive = 'Unarchive',
 }
 
+const URL_ALL = '';
+
 const STORAGE_PATH = 'experiment-list';
 const STORAGE_FILTERS_KEY = 'filters';
 const STORAGE_LIMIT_KEY = 'limit';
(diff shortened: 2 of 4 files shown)


webui/react/src/pages/ExperimentList.tsx
```diff
@@ -81,13 +81,119 @@ const ExperimentList: React.FC = () => {
...
-  const [ sorter, setSorter ] = useState(initSorter);
+  const [ isUrlParsed, setIsUrlParsed ] = useState<boolean>(false);
+  const [ pagination, setPagination ] = useState<Pagination>({ limit: initLimit, offset: 0 });
   const [ search, setSearch ] = useState('');
-  const [ experiments, setExperiments ] = useState<ExperimentItem[]>();
   const [ selectedRowKeys, setSelectedRowKeys ] = useState<string[]>([]);
+  const [ sorter, setSorter ] = useState(initSorter);
+  const [ total, setTotal ] = useState(0);
+
+  /*
```
apizzini, line 93: It might be a good idea to split those 2 effects out from the main component to increase code readability.

Something line `useParseAndUpdatePageURL(filters, pagination, search, sorter)`.

If you like the sound of that, what should be the file name? `webui/react/src/pages/ExperimentList.utils.ts`? Or `webui/react/src/pages/ExperimentList/utils.ts`?

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -81,13 +81,119 @@ const ExperimentList: React.FC = () => {
...
+      searchParams.sortKey = sorter.key;
+    }
+    if (filters.states) {
+      searchParams.states = filters.states.join(',');
+    }
+    if (filters.username) {
+      searchParams.username = filters.username;
+    }
+
+    const searchParamsString = new URLSearchParams(searchParams).toString();
```
apizzini, line 121: I don't like the fact that `toString` is forcing the conversion of URL special chars (eg: `...&labels=x,y,z&...` becomes  `...&labels=x%2Cy%2Cz&...`.

Not sure if you have any suggestion about this, because the only alternative is to manually "merge" that URL params object instead of relying on `toString`.

webui/react/src/pages/ExperimentList.tsx
hkang1: how about making it more generic so it can take in a config also to provide info on what fields and how to parse each field. Then this hook can just sit under `src/hooks`?

webui/react/src/pages/ExperimentList.tsx
hkang1: I think param values are always encoded to avoid injections.

If we manually "merge" or override this behavior, then it would be on us to parse and validate the values and escape and/or ignore certain characters before we feed it to the API.

Another option can be to use the special characters that are left alone. Maybe the period can be an option, where it looks close enough to a comma.
 
This [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#encodeuri_vs_encodeuricomponent) shows that the reserved special characters that ARE encoded:
`;,/?:@&=+$#`

where as these are NOT escaped:
`-_.!~*'()`

hkang1 (approved):
Everything worked beautifully except for the labels due to something going on with the endpoint for that particular query param. I think that will get fixed soon (I've recently reported it).

hkang1:
Oh there is one weird issue I just encountered. If I have `All` set for `labels` filter and then reload the page, the labels filter will show `1 selected` but when viewing the options in the dropdown, none are selected. It's possible that there is a residue of a user preference set to a label that no longer exists and doesn't have a way of clearing out. One minor fix (assuming the previously described case was the problem) could be to delete all labels user preferences (under local storage) when the user selects on `All`.

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -80,13 +81,123 @@ const ExperimentList: React.FC = () => {
...
+   * When filters changes update the page URL.
+   */
+  useEffect(() => {
+    if (!isUrlParsed) return;
+
+    const searchParams: Record<string, string> = {
+      archived: filters.showArchived ? '1' : '0',
+      limit: pagination.limit.toString(),
+      offset: pagination.offset.toString(),
+      sortDesc: sorter.descend ? '1' : '0',
```
hkang1, line 103: should this only be set if `sortKey` is set?

webui/react/src/pages/ExperimentList.tsx
apizzini: Another thing to consider is that labels can contain all of those special characters... 
Schermata 2021-02-24 alle 09 54 23

so having a query string with &labels=x.y will not work.

webui/react/src/pages/ExperimentList.tsx
apizzini: We discussed via IM that all params has to be set in the URL bc otherwise an unset param will cause the page to be rendered with the user preference (that might be different from the global default value).

webui/react/src/pages/ExperimentList.tsx
apizzini: We discussed via IM that is best to avoid "overengineering" this, we'll split the re-usable code when that will be needed (next page that will need a sharable URL).

webui/react/src/pages/ExperimentList.tsx
apizzini: Updated to build/parse url like this `...&label=x&label=y&label=z...`

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -80,13 +81,126 @@ const ExperimentList: React.FC = () => {
...
+   */
+  useEffect(() => {
+    if (!isUrlParsed) return;
+
+    const searchParams: string[][] = [
+      [ 'archived', filters.showArchived ? '1' : '0' ],
+      [ 'limit', pagination.limit.toString() ],
+      [ 'offset', pagination.offset.toString() ],
+      [ 'sortDesc', sorter.descend ? '1' : '0' ],
+    ];
```
hkang1, line 104: **blocker:** I believe all of the params need to be set with a default of at least an empty string `''`. For example if `username` is only set when `filters.username` is available, then in the URL query string, `&username=` will not be available. And when that link is copied or shared, if Jane pasting the URL has a user preference of `username` set to `jane`, the URL which should show All users, will default to only showing `jane`.

Then on the parsing end when you do a `params.get('username')` you should get an empty string instead of a `null` to know that it's set to All as opposed to using the user preference default. Since we are taking the URL params as priority #1 followed by user preferences, then default settings and values.

hkang1: so instead of
```
    if (filters.username) {
      searchParams.push([ 'username', filters.username ]);
    }
```
It would be something more like:
```
    const searchParams: string[][] = [
      ...
      [ 'username', filters.username || '' ],
    ];
```

hkang1: I noticed that you had `URLSearchParams` as a record originally before you changed to array of an array. What changed your mind to this format? Just curious.

apizzini: 1. I missed to handle the case for the `username`, was already there for `label`. Pushed a fix for this.
2. TS interface for `URLSearchParams` accept `string[][] | Record<string, string> | string | URLSearchParams`. Since we have the `label` params that needs to accept multiple values `Record` would have no longer worked. That's why I migrate to a `string[][]` array.

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -80,13 +83,136 @@ const ExperimentList: React.FC = () => {
...
+  const [ sorter, setSorter ] = useState(initSorter);
+  const [ total, setTotal ] = useState(0);
+
+  /*
+   * When filters changes update the page URL.
+   */
+  useEffect(() => {
+    if (!isUrlParsed) return;
+
+    const searchParams = new URLSearchParams;
```
hkang1, line 101: I didn't know this was a thing where you can just create a new instance without the parentheses! (`new URLSearchParams` vs `new URLSearchParams()`) You learn something new everyday! :)

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -80,13 +83,136 @@ const ExperimentList: React.FC = () => {
...
-  const [ experiments, setExperiments ] = useState<ExperimentItem[]>();
   const [ selectedRowKeys, setSelectedRowKeys ] = useState<string[]>([]);
+  const [ sorter, setSorter ] = useState(initSorter);
+  const [ total, setTotal ] = useState(0);
+
+  /*
+   * When filters changes update the page URL.
+   */
+  useEffect(() => {
+    if (!isUrlParsed) return;
```
hkang1, line 99: If we remove this line, this should ensure the experiment list page will always have query params set right?

hkang1 (approved):
This is looking sharp! One final detail of always showing the query params in the URL (even when navigating to a plain path of `/experiments` without query params). It should just be a simple removal of one line.

webui/react/src/pages/ExperimentList.tsx
apizzini: That `isUrlParsed` is preventing a race condition between updating and parsing the query params.

1st load of the page we should try to parse it (and block updating), after the 1st render we should instead always update it (and avoid parsing it).

webui/react/src/components/Navigation.tsx
```diff
@@ -35,7 +35,7 @@ const NavigationItem: React.FC<ItemProps> = ({ path, status, ...props }: ItemPro
   useEffect(() => setIsActive(location.pathname === path), [ location.pathname, path ]);
 
   const link = (
-    <Link className={classes.join(' ')} path={path} {...props}>
+    <Link className={classes.join(' ')} disabled={isActive} path={path} {...props}>
```
apizzini, line 38: @hkang1 I've added this like discussed yesterday. Seems to be working fine. What do you think?

hkang1: **nit:** can you also modify `Naviagition.module.scss` to update:
```
  .navItem.active {
    background-color: var(--theme-colors-base-0);
    cursor: default;
  }
```
to disable the pointer cursor when it is "active", that way the user doesn't think it's a broken link.

hkang1 (approved):
Thanks for all the updates this is all looking solid! One small style change and it's good to go!

webui/react/src/components/Navigation.tsx
apizzini: Yes, good idea :)

PR #2002, opened 2021-02-22
fix: fix v0 experiment snapshot shim

stoksc:
## Description
This change fixes the shim to not explode stuff that isn't PBT.

## Test Plan
- [ ] test with asha, also write a shim for `cached_checkpoints` that is missing

## Commentary (optional)
Eh.. I definitely tested with asha, I'm not sure what happened. I'll test it all again.

## Checklist

### DIFF
--- a/master/internal/restore.go
+++ b/master/internal/restore.go
@@ -261,7 +261,7 @@ func shimExperimentSnapshotV0(snapshot []byte) ([]byte, error) {
 		return snapshot, nil
 	}
 
-	// And queue them to be sent to trials immediately (also filter out checkpoints).
+	// And queue them to be sent to trials immediately.
 	for _, ops := range waitingCheckpoints.(map[string]interface{}) {
 		searcherState["trial_operations"] = append(searcherState["trial_operations"].([]interface{}),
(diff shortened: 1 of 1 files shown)


master/internal/restore.go
```diff
@@ -251,33 +251,33 @@ func shimExperimentSnapshotV0(snapshot []byte) ([]byte, error) {
...
-				// Any numeric types parsed from JSON into an interface{} will be a float64.
-				if op["OperationType"].(float64) == 3 {
-					continue
-				}
-				trialOperations = append(trialOperations, op)
-			}
-			searcherState.(map[string]interface{})["trial_operations"] = append(
-				trialOperations, waitingCheckpointOps...)
+
+	// And queue them to be sent to trials immediately (also filter out checkpoints).
```
rb-determined-ai, line 264: is the parenthetical part of the comment correct?  seems like that happens later.

rb-determined-ai (approved):

master/internal/restore.go
stoksc: yeah yep. these lines used to do both.

PR #2003, opened 2021-02-22
fix: correct batches column order on trial detail page [DET-5023]

hkang1:
## Description

Batches column on the trial info table is not sorting purely as numbers but as alpha-numeric order. Correct it to use a numeric sorter.

## Test Plan

- [ ] check on the trial detail page to make sure batches and metrics are sorting properly on the trial info table.

## Checklist

### DIFF
--- a/webui/react/src/utils/sort.ts
+++ b/webui/react/src/utils/sort.ts
@@ -1,0 +1,93 @@
+import { CommandState, MetricName, MetricType, Primitive, RunState, State } from 'types';
+
+const runStateSortValues: Record<RunState, number> = {
+  [RunState.Active]: 0,
+  [RunState.Paused]: 1,
+  [RunState.StoppingError]: 2,
+  [RunState.Errored]: 3,
+  [RunState.StoppingCompleted]: 4,
+  [RunState.Completed]: 5,
+  [RunState.StoppingCanceled]: 6,
+  [RunState.Canceled]: 7,
+  [RunState.Deleted]: 7,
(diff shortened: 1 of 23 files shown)


apizzini (approved):

webui/react/src/utils/sort.ts
```diff
@@ -0,0 +1,93 @@
...
+    return alphanumericSorter(a, b);
+  }
+  return 0;
+};
+
+export const runStateSorter = (a: RunState, b: RunState): number => {
+  return runStateSortValues[a] - runStateSortValues[b];
+};
+
+export const stringTimeSorter = (a: string, b: string): number => {
```
apizzini, line 79: Is that one really needed? since time strings (eg: `2021-02-24T07:53:06.911722Z`) can simply be sorted alphabetically?

apizzini:
I saw the updates and those changes are great!

webui/react/src/utils/sort.ts
hkang1: good question. Not sure what circumstances were, but I remember Hamid's and my respective local clusters reported different strings. Where his had precision to the microseconds where mine only had milliseconds. `2021-02-24T07:53:06.911722Z` vs `2021-02-24T07:53:06.911Z`. I also don't know if we can completely rely on timezones to always be UTC. So in the case where we are comparing different timezones and different precisions, converting them to date objects prior to comparison might be safer.

apizzini: Thanks for the explanation.

PR #2004, opened 2021-02-22
fix: fix shim for stateless searchers (grid, single, etc..)

stoksc:
## Description
Let's try this again..

## Test Plan
- [x] run adaptive, pbt, etc on 0.14.1, upgrade to 0.14.3dev so shim occurs and validate experiments complete as expected.

## Checklist

### DIFF
--- a/master/internal/restore.go
+++ b/master/internal/restore.go
@@ -252,9 +252,12 @@ func shimExperimentSnapshotV0(snapshot []byte) ([]byte, error) {
 		return nil, err
 	}
 
-	// Get the waiting operations from PBT.
 	searcherState := experimentSnapshotV0["searcher_state"].(map[string]interface{})
+	if searcherState["search_method_state"] == nil {
+		return snapshot, nil
+	}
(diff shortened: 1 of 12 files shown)


rb-determined-ai:
Can we include the name of the searcher in the searcher state?  Or at least pass it in from the experiment actor or something?   Seems like this information would eliminate the bugs we have hit.

shiyuann:
Since you are working on this, maybe also include the shim for "cached_checkpoint".

stoksc:
@rb-determined-ai I don't want to add that state on the grounds that to shim it you need the experiment config since single, random and grid are stateless and can't be differentiated from their state alone and so to figure it out you need the experiment config and I don't want to tie their states together like that.. if I could add it without giving shims an experiment config (that _wouldn't_ be the version from which they are shimming) I'd be onboard. I get it would be useful to have.. I'm just not exactly sure how to get it without make things worse lol.

also I think the answer here is to probably make the next eng that writes a shim right a test for it.

rb-determined-ai (approved):
Looks good.  Hopefully this will be much easier next time

PR #2005, opened 2021-02-23
feat: HP importance implementation [DET-4465]

katport:
## Checklist


master/Makefile
```diff
@@ -74,6 +77,7 @@ pre-package:
 	cp ../common/dist/*.whl build/wheels/
 	cp ../cli/dist/*.whl build/wheels/
 	cp ../harness/dist/*.whl build/wheels/
+	cp $(shell which growforest) build/
```
stoksc, line 80: Is this portable? Like if I run this on a mac wouldn't this package the GOOS=darwin version of growforest into a Linux container for the master?

master/internal/core.go
```diff
@@ -580,7 +580,8 @@ func (m *Master) Run(ctx context.Context) error {
 	template.RegisterAPIHandler(m.echo, m.db, authFuncs...)
 
 	if m.config.Telemetry.Enabled && m.config.Telemetry.SegmentMasterKey != "" {
-		if telemetry, err := telemetry.NewActor(
+		var telemetryActor actor.Actor
+		if telemetryActor, err = telemetry.NewActor(
```
stoksc, line 584: if this was a shadowing go lint thing, you could also just call err `tErr`

master/internal/db/postgres_experiments.go
```diff
@@ -497,12 +501,16 @@ FROM trials t
...
 		if err != nil {
 			return nil, errors.Wrap(err,
 				"Failed to process validation metrics for hyperparameter importance")
 		}
-		results = append(results, result)
+		if previous, exists := results[batches]; exists {
+			results[batches] = append(previous, result)
+		} else {
+			results[batches] = []model.HPImportanceTrialData{result}
+		}
```
stoksc, line 513:
```suggestion
		results[batches] = append(results[batches], result)
```
go should handle this just fine e.g. https://play.golang.org/p/6xaMh0qJ8SJ

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,154 @@
...
+	// TODO: use the bar chart from hp viz to improve hp importance
+	numTrees := int64(totalNumTrials / nReplications)
+
+	if numTrees > maxNumTrees {
+		numTrees = maxNumTrees
+	}
+	strNumTrees := strconv.FormatInt(numTrees, 10)
+
+	// Call Random Forest
+	_, err := exec.Command(growforest, "-train", workingDir+"/data.arff", "-target=metric", "-nCores", nCores, "-ace", strNumTrees, "-importance", workingDir+"/importance.txt", "-rfpred", workingDir+"rface.sf").Output()
```
stoksc, line 128:
```suggestion
	_, err := exec.Command(
		growforest,
		// comment explaining this arg
		"-train", workingDir+"/data.arff",
		// comment explaining...
		"-target=metric",
		//
		"-nCores", nCores,
		//
		"-ace", strNumTrees,
		//
		"-importance",
		//
		workingDir+"/importance.txt",
		//
		"-rfpred",
		//
		workingDir+"rface.sf").Output()
```

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,154 @@
...
+			break
+		}
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		if record[1] == "metric" {
+			continue
+		}
+		output[record[1]], _ = strconv.ParseFloat(record[2], 32) // 32 or 64 bit?
```
stoksc, line 151: This all feels really brittle; is there really no good way to consume growforest as a lib? I peeked and the answer looked grim lol. If not can we break this into a func called like "parseGrowforestOutput` with a link to the docs that describe the format being parsed?

master/internal/hpimportance/hpimportance.go
stoksc, line 128: Maybe just set `cmd.Stdout` and `cmd.Stderr` to /dev/null and call `Run()` instead?

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,154 @@
...
+	strNumTrees := strconv.FormatInt(numTrees, 10)
+
+	// Call Random Forest
+	_, err := exec.Command(growforest, "-train", workingDir+"/data.arff", "-target=metric", "-nCores", nCores, "-ace", strNumTrees, "-importance", workingDir+"/importance.txt", "-rfpred", workingDir+"rface.sf").Output()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Read and Return Output
+	file, err := os.Open(workingDir + "/importance.txt")
```
stoksc, line 134: I think you need to defer closing this file, also don't forget to check the error before doing that

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,154 @@
...
+
+	if numTrees > maxNumTrees {
+		numTrees = maxNumTrees
+	}
+	strNumTrees := strconv.FormatInt(numTrees, 10)
+
+	// Call Random Forest
+	_, err := exec.Command(growforest, "-train", workingDir+"/data.arff", "-target=metric", "-nCores", nCores, "-ace", strNumTrees, "-importance", workingDir+"/importance.txt", "-rfpred", workingDir+"rface.sf").Output()
+	if err != nil {
+		log.Fatal(err)
```
stoksc, line 130: I think we should have real error handling, this will just log and panic.. that being said it _is_ being run in an actor that should recover the panic and tell its parent but I didn't review that code with panic's below in mind so I'd feel better just returning errs

master/internal/hpimportance/manager.go
```diff
@@ -66,16 +72,31 @@ type manager struct {
...
+) (actor.Actor, error) {
+	// growforest should either be installed in PATH (when running from source) or package with the
+	// master (when running from binary packages).
+	growforest := path.Join(masterRoot, "growforest")
+	_, err := os.Stat(growforest)
+	if os.IsNotExist(err) {
+		growforest = "growforest"
+	}
+
+	workingDir := "/tmp/determined/growforest"
```
stoksc, line 85: I'd put this and all the other file related constants in a 
```
const (
    	workingDir := "/tmp/determined/growforest"
)
```
block together with really good comments.

master/internal/hpimportance/worker.go
```diff
@@ -99,7 +117,16 @@ func taskHandlerFactory(db *db.PgDB, system *actor.System) func(interface{}) int
...
-		results := computeHPImportance(trials)
+		taskDir := path.Join(workingDir, fmt.Sprint(actorId))
+		err = os.Mkdir(taskDir, 0066)
+		if err != nil {
+			sendWorkFailed(system, work, err.Error())
+		}
+		results := computeHPImportance(trials, experimentConfig, masterConfig, growforest, taskDir)
+		err = os.RemoveAll(taskDir)
+		if err != nil {
+			ctx.Log().Errorf("Failed to clean up temporary directory %s", taskDir)
```
stoksc, line 128: dont forget a with err so we dont lose what actually happened
```suggestion
			ctx.Log().WithErr(err).Errorf("Failed to clean up temporary directory %s", taskDir)
```

master/internal/hpimportance/hpimportance.go
```diff
@@ -1,15 +1,154 @@
...
-	// This is a place-holder for a future Random Forest implementation to actually compute the
-	// importance of each hyperparameter.
+const (
+	maxDiffCompBatch = 4
+	minNumberTrials  = 50
+	idealBatchDiff   = 2
+	nReplications    = 2
+)
+
+func createDataFile(data map[int][]model.HPImportanceTrialData, experimentConfig *model.ExperimentConfig, workingDir string) int {
```
stoksc, line 23: can you leave a comment describing this func and documentation to the growforest page that describes this file format?

stoksc (commented):
i know i reviewed a little early (i'm not assigned), sorry if i commented a bunch of stuff you're going to get to eventually anyway.

master/Makefile
mackrorysd: I'm kinda new to Go dependency management / packaging so I'm very open to suggestions. I wasn't too concerned about it because we don't build packages from people's workstations, do we? I assumed that was done in CI on an OS we choose very carefully for this. (and for context - I'm speaking like I wrote this line because I did - this commit is a mix of stuff I did to get Katie started, and I've actually been working on a fork of this PR doing some of the same things you suggest here :D)

master/internal/db/postgres_experiments.go
mackrorysd: Already done in my fork. I think gofmt -s actually did this for me, which impresses me..

master/internal/hpimportance/hpimportance.go
mackrorysd: Consuming it as a lib would be nice, although I like having it as a separate process from a resource isolation standpoint. There's a bunch of logic we would need to duplicate in their CLI in order to use it as a lib without modifying their lib a bunch, but the big blocker was that the ability to restrict this to a certain number of CPU cores is done with GOMAXTHREADS and that affects the whole process.

master/internal/hpimportance/hpimportance.go
mackrorysd: Yeah in an update this is reported back as an error, and it's stored in the DB to be bubbled up to the user when they try and see their results.

master/internal/hpimportance/hpimportance.go
mackrorysd: I figured the output would be useful for debugging, though I switched this to log CombinedOutput() to capture stderr as well.

stoksc: What we did with fluentbit on the agents was to store it in a buffer and in the event of an error dump the buffer in the master logs (agent logs for fluent), which is the last 50-100 lines of logs. You get the info in the event of failures and not too much bloat.

PR #2007, opened 2021-02-23
feat: scatter plot and heat map [DET-4453, DET-4459, DET-5085]

hkang1:
## Description

Add new HP Visualizations:

HP Scatter Plots
Screen Shot 2021-03-01 at 12 44 13 PM

HP Heat Maps
Screen Shot 2021-03-01 at 12 44 13 PM

## Test Plan

- [] New HP charts can render after an experiment has terminated.
- [] New HP charts can render during an experiment run.
- [] Filters for new HP charts filter appropriately.

## Checklist


hamidzr (commented):
great work as always. the only that jumped to me was the identity plots and whether we want to keep them or not which we talked about.

webui/react/src/utils/number.ts
```diff
@@ -3,3 +3,8 @@ export const percent = (n: number, decimals = 1): number => {
   const factor = Math.pow(10, decimals);
   return Math.round(normalized * 100 * factor) / factor;
 };
+
+export const roundToPrecision = (n: number, precision = 6): number => {
+  const factor = 10 ** precision;
+  return Math.round((n + Number.EPSILON) * factor) / factor;
```
apizzini, line 9: Need to ask: why is important to `+ Number.EPSILON`?

webui/react/src/components/GridListRadioGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+const GridListRadioGroup: React.FC<Props> = ({ onChange, value }: Props) => {
+  const [ view, setView ] = useState(value);
+
+  const handleChange = useCallback((e: RadioChangeEvent) => {
+    if (onChange) onChange(e.target.value as GridListView);
+    setView(e.target.value as GridListView);
+  }, [ onChange ]);
+
+  return (
+    <Radio.Group className={css.base} value={view} onChange={handleChange}>
```
apizzini, line 28: nit visual thing, not sure if wanted/compromised?

Schermata 2021-03-03 alle 09 48 03

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -49,34 +47,38 @@ const ExperimentDetails: React.FC = () => {
...
     try {
-      const experiment = await getExperimentDetails({ id }, { signal: experimentCanceler.signal });
-      const validationHistory = await getExpValidationHistory({ id });
-      if (!isEqual(experiment, experimentDetails.data)) {
-        setExperimentDetails(prev => ({ ...prev, data: experiment, isLoading: false }));
+      const [ experimentData, validationHistory ] = await Promise.all([
+        getExperimentDetails({ id }, { signal: canceler.signal }),
+        getExpValidationHistory({ id }, { signal: canceler.signal }),
+      ]);
+      if (!isEqual(experimentData, experiment)) {
```
apizzini, line 64: I guess you already know that, `isEqual` JSON comparison might fail in many ways, but is acceptable for those cases? (and worst case scenario will trigger an un-needed re-render?)

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpScatterPlots.tsx
```diff
@@ -0,0 +1,239 @@
...
+  onMetricChange,
+  selectedBatch,
+  selectedHParams,
+  selectedMetric,
+}: Props) => {
+  const baseRef = useRef<HTMLDivElement>(null);
+  const [ hasLoaded, setHasLoaded ] = useState(false);
+  const [ chartData, setChartData ] = useState<HpMetricData>();
+  const [ pageError, setPageError ] = useState<Error>();
+  const resize = useResize(baseRef);
```
apizzini, line 66: ultra-nit: in general I tend to put breaklines between interdependent vars, like in that case `resize` depends on `baseRef`. Are rules like this enforced in the codebase as good practices? Or there are others? Or I'm overlooking a super-minor thing?

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpScatterPlots.tsx
```diff
@@ -0,0 +1,239 @@
...
+        setHasLoaded(true);
+      },
+    ).catch(e => setPageError(e));
+
+    return () => canceler.abort();
+  }, [ experiment, hParams, selectedBatch, selectedMetric ]);
+
+  if (pageError) {
+    return <Message title={pageError.message} />;
+  } else if (hasLoaded && !chartData) {
```
apizzini, line 167: no need for that `else` (out of code readability)?

apizzini (approved):
That looks amazing!

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpScatterPlots.tsx
hkang1: I think that's a good rule, it's just typically hard to enforce if it's not automatic. But I'll do that going forward. Stuff like this would be great to discuss in WebUI AG. to get everyone on the same page.

webui/react/src/pages/ExperimentDetails.tsx
hkang1: Yeah we should probably update this function. I'll add a ticket for this.

webui/react/src/utils/number.ts
hkang1: Great question. There are some [significant digit issues without`Number.EPSILON`](https://stackoverflow.com/a/11832950/5402432)

webui/react/src/components/GridListRadioGroup.tsx
hkang1: Great catch! I'll fix this.

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpScatterPlots.tsx
hkang1: yeah exactly, just a little bit more succinct. More so when we have more than two paths.

PR #2009, opened 2021-02-24
fix: webui experiment detail action spinner to persist until state changes [DET-4977]

apizzini:
## Description

1. The 2 `useEffect` will watch `experiment.archived` and `experiment.state` and reset the related button when either has (really) changed. By default all loading state will already be `false`, but triggering one of the 6 action will set the individual state to `true`. It will get back to `false` when the `experiment.archived` and `experiment.state` get changed.
2. The loading state for Tensorboard has been removed. I'm not sure if Tensorboard has a state and if that can/should be tracked. Btw my guess (maybe wrong) is that this is not needed (Tensorboard is opened in another window).

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
@@ -1,5 +1,5 @@
 import { Button, Popconfirm, Space } from 'antd';
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { ConditionalButton } from 'components/types';
 import {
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
```diff
@@ -29,91 +29,68 @@ interface Props {
...
-    Activate: false,
-    Archive: false,
-    Cancel: false,
-    Fork: false,
-    Kill: false,
-    Pause: false,
-    Tensorboard: false,
-    Unarchive: false,
-  });
+  const [ isRunningActivate, setIsRunningActivate ] = useState<boolean>(false);
```
hamidzr, line 33: question: any downsides you saw with the way loadingStates were being tracked before?

hamidzr (commented):
Goo work! Checking the actual target states makes sense. two comments: 

- tensorboard loading state could on the request(s) that were made before it opened in a new tab (as it did before)
- if the requests fail the respective buttons will be stuck in the loading state. Ideally the loading state would go away letting the user trigger it again.

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
apizzini: I think it's a good practice to flatten data structure, because otherwise changing any of the multiple states will require to compose a new object. In that case, practically, the 2 `useEffect` would need to be dependent on that object, increasing the probability of errors and the complexity of code.

apizzini:
> if the requests fail the respective buttons will be stuck in the loading state. Ideally the loading state would go away letting the user trigger it again.

So you think would be a good idea to add a timeout to disable the loading state after 10 seconds?

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
```diff
@@ -29,91 +29,68 @@ interface Props {
...
   const handleCreateTensorboard = useCallback(async () => {
-    setBtnLoadingStates(state => ({ ...state, Tensorboard: true }));
-    try {
-      const tensorboard = await openOrCreateTensorboard({ experimentIds: [ experiment.id ] });
-      openCommand(tensorboard);
-      onSettled();
-    } finally {
-      setBtnLoadingStates(state => ({ ...state, Tensorboard: false }));
-    }
+    const tensorboard = await openOrCreateTensorboard({ experimentIds: [ experiment.id ] });
```
apizzini, line 71: With this:

> tensorboard loading state could on the request(s) that were made before it opened in a new tab (as it did before)

you mean that the code should be more similar to:
```javascript
const handleCreateTensorboard = useCallback(async () => {
  setIsRunningTensorboard(true);
  const tensorboard = await ...
  openCommand(tensorboard);
  onSettled();
  setIsRunningTensorboard(false);
}, [...]);
```

?

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
```diff
@@ -29,91 +29,68 @@ interface Props {
...
+    setIsRunningPause(false);
+  }, [ experiment.state ]);
 
   const handleArchive = useCallback(() => async (): Promise<void> => {
-    setBtnLoadingStates(state => ({ ...state, Archive: true }));
-    try {
-      await archiveExperiment({ experimentId: experiment.id });
-      onSettled();
-    } finally {
-      setBtnLoadingStates(state => ({ ...state, Archive: false }));
```
hamidzr, line 52: like this but maybe just catch instead of finally?

hamidzr:
> > if the requests fail the respective buttons will be stuck in the loading state. Ideally the loading state would go away letting the user trigger it again.
> 
> So you think would be a good idea to add a timeout to disable the loading state after 10 seconds?

hmm timeout could work. Wdy think about tracking whether the promise for the request failed or not (similar to how it was before) this way we wouldn't need to timeout for an arbitrary period (https://github.com/determined-ai/determined/pull/2009/files#r585881024)

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
hamidzr: that makes sense. it'd be easy to test it by introducing an artificial delay to requests using chrome dev tools.

apizzini: I've added this. Probably will need to be updated with `try ... catch ...` like the other ones, will continue the discussion on the other points.

webui/react/src/pages/ExperimentDetails/ExperimentActions.tsx
apizzini: Right, that's better. Should be good now?

hamidzr (approved):
looks great!

PR #2010, opened 2021-02-24
chore: webui uniform path generation [DET-4734]

apizzini:
## Checklist

### DIFF
--- a/webui/react/src/routes/utils.ts
+++ b/webui/react/src/routes/utils.ts
@@ -3,6 +3,7 @@ import { MouseEvent, MouseEventHandler } from 'react';
 
 import { globalStorage } from 'globalStorage';
 import history from 'routes/history';
+import { CommandTask } from 'types';
 import { clone } from 'utils/data';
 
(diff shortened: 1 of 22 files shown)


webui/react/src/routes/utils.ts
```diff
@@ -130,10 +131,62 @@ export const linkPath = (aPath: string, external = false): string => {
...
+  },
+  docReleaseNotes: (): string => {
+    return '/docs/release-notes.html';
+  },
+  docRestApi: (): string => {
+    return '/docs/rest-api/';
+  },
+  docUsers: (): string => {
+    return '/docs/topic-guides/users.html';
+  },
```
hkang1, line 154: **non-blocking:** don't have a good answer, but if the docs list grows any longer, it might just be best to have a general `docs` path where it will just take a suffix from the caller. For example:
```
docs: (suffix?: string): string => {
  return `/docs${suffix}`;
},
```

hkang1 (approved):
So much more organized! 🙏

PR #2011, opened 2021-02-24
build: auto-gen swagger api bindings

hamidzr:
## Commentary (optional)

- [x] update circle ci

## Checklist


bindings/README.md
```diff
@@ -0,0 +1,8 @@
+# Determined API Bindings
+
+To generate the bindings, place the Swagger specifications file
```
hamidzr, line 3: update me

bindings/Makefile
```diff
@@ -0,0 +1,45 @@
+PROJECT_ROOT := ..
+SWAGGER_SPEC := ${PROJECT_ROOT}/proto/build/swagger/determined/api/v1/api.swagger.json
+CODEGEN_BIN := swagger-codegen-cli-2.4.14.jar
+
+.PHONY: all
+all: get-deps build-proto
+	$(MAKE) get-swagger-spec
```
hamidzr, line 7: remove missing targets

bindings/Makefile
```diff
@@ -0,0 +1,45 @@
+PROJECT_ROOT := ..
```
rb-determined-ai, line 1: Seems unnecessary

bindings/Makefile
```diff
@@ -0,0 +1,45 @@
...
+	java -jar deps/$(CODEGEN_BIN) generate -i ${SWAGGER_SPEC} \
+		-l $* -o build/$*
+	touch build/$*
+
+.PHONY: build-default
+build-default: build/typescript-fetch build/python
+
+.PHONY: build
+build: ${SWAGGER_SPEC}
+	$(MAKE)	build-default
```
rb-determined-ai, line 30: why is this recursively calling make rather than just depending on `build-default`?

bindings/examples/python/main.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
```
rb-determined-ai, line 1: Who is the intended audience for the examples?

bindings/Makefile
```diff
@@ -0,0 +1,45 @@
...
+.PHONY: get-deps
+get-deps: deps/${CODEGEN_BIN}
+
+deps/${CODEGEN_BIN}:
+	mkdir -p deps
+	curl https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.4.14/${CODEGEN_BIN} \
+		-o deps/${CODEGEN_BIN}
+
+build/%: ${SWAGGER_SPEC}
+	[ -f ./${SWAGGER_SPEC} ] || exit 1
```
rb-determined-ai, line 19: when would this fire?

bindings/examples/python/main.py
```diff
@@ -0,0 +1,100 @@
...
+import swagger_client
+from swagger_client import models
+from pathlib import Path
+from swagger_client.rest import ApiException
+from pprint import pprint
+from determined_common import context, yaml, constants
+
+
+SERVER_ADDRESS = "http://localhost:8080"
+SAMPLE_EXPERIMENT_DIR = Path.home().joinpath("projects/da/determined/e2e_tests/tests/fixtures/no_op")
```
rb-determined-ai, line 20: hardcoded path?

rb-determined-ai (commented):
I suppose that the python swagger bindings more-or-less are required to be their own package?

That's a bit unfortunate but it does look like that to me.

Otherwise I think this all looks pretty good.

Are the `build/python` and `build/typescript-fetch` targets idempotent?

bindings/Makefile
hamidzr: initially, these two (build-default and ${SWAGGER_SPEC}) were not parallelizable

bindings/examples/python/main.py
hamidzr: this file is code samples I had for each of the languages we think we or external users would use. I'd probably leve the examples directory out of the pr

bindings/Makefile
hamidzr: good point. it didn't used to depend on the same file

bindings/Makefile
rb-determined-ai: Then make a chain of targets?

hamidzr:
> I suppose that the python swagger bindings more-or-less are required to be their own package?

Do you mean a python package?  yeah that's how the default python generator generates them. what downside do you see?

bindings/examples/python/main.py
rb-determined-ai: yeah let's do that.  Until we fix this issue with the rest api where you have to send a jsonized tarball to create an experiment, I don't think we should be thinking of the rest api as consumable by external users.

And I desperately do not want to teach external users how to do that because that would make it harder for us to fix.

rb-determined-ai:
> what downside do you see?

Just slightly more complex from the user's perspective.  We already have users (and internal folks too) who don't really understand the relationship between the 4 packages we already have.

hamidzr:
> Are the `build/python` and `build/typescript-fetch` targets idempotent?

I'm counting on it their only input is the swagger spec file and the jar file. I should add that as a dependency

bindings/Makefile
hamidzr: yeah, it used be used in more places. I'll remove

hamidzr:
@rb-determined-ai I added a bit in the postscript to remove the createExperiment operation. it'll still leave the request and response objects for it which we can use internally if we need to but the main operation is gone and not advertised. this will remove it from our docs too.

bindings/Makefile
```diff
@@ -0,0 +1,37 @@
...
+		-o deps/${CODEGEN_BIN}
+
+build/%: ${SWAGGER_SPEC} deps/${CODEGEN_BIN}
+	mkdir -p build/$*
+	java -jar deps/$(CODEGEN_BIN) generate -i ${SWAGGER_SPEC} \
+		-l $* -o build/$*
+	touch build/$*
+
+.PHONY: build-default
+build-default: build/typescript-fetch build/python
```
rb-determined-ai, line 23: Whats the purpose of this target?

bindings/Makefile
```diff
@@ -0,0 +1,37 @@
...
+
+.PHONY: build
+build: ${SWAGGER_SPEC} build-default
```
rb-determined-ai, line 26: I think if you just depend on `build/python` and `build/typescript-fetch` here, but those depend on `${SWAGGER_SPEC}` you can just drop the build-default target.

rb-determined-ai (approved):
This looks good.

Before we use the python bindings in code, we'll need to add tooling for publishing them automatically, and we'll need to pass a `--config` when we generate so we can name them something better and version them properly.

But I'm fine leaving that for a followup PR.

bindings/Makefile
hamidzr: it was to allow different bindings to be built concurrently with -j but I needed to do this https://github.com/determined-ai/determined/pull/2011/commits/9d999f52620ec6bb3e8a7f1c5e15a76f1bd391d3

bindings/Makefile
hamidzr: https://github.com/determined-ai/determined/pull/2011#discussion_r583291102

PR #2012, opened 2021-02-24
fix: fix query for GET experiments [DET-5039]

stoksc:
## Description
This change updates the GET experiments query to handle null jsonb (not just a NULL column) when filtering by labels. The labels column initially is returned as NULL if no labels key exists in the config jsonb. The expression works fine in this case, but if labels are added to an experiment and the removed it will get rewritten as null instead of removing the column which causes `jsonb_array_elements_text` to barf on the call with the arg `null` jsonb. This PR just checks for `null` jsonb and converts it to NULL.

## Test Plan
- [x] run against latest master, where this is broken

## Commentary (optional)
spent like five minutes wondering why `null` wasn't getting coalesced before I realized it was a null jsonb object in a not null column lol.

## Checklist

### DIFF
--- a/master/static/srv/get_experiments.sql
+++ b/master/static/srv/get_experiments.sql
@@ -19,7 +19,13 @@ WITH filtered_exps AS (
         AND ($3 = '' OR (u.username IN (SELECT unnest(string_to_array($3, ',')))))
         AND (
                 $4 = ''
-                OR string_to_array($4, ',') <@ ARRAY(SELECT jsonb_array_elements_text(e.config->'labels'))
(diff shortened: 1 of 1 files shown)


shiyuann (commented):
Have we checked if we had other places of this?

shiyuann (approved):
Looks good

stoksc:
> Have we checked if we had other places of this?

I did a quick think when I opened the PR, I think the is the only place it would happen (nothing else returns something like labels, a nested field in some jsonb, as a first class thing and allows filtering on it afaik).

PR #2020, opened 2021-03-01
chore: add scaffolding for trial profiler metrics [DET-5059]

stoksc:
## Description
This change adds scaffolding for the trial profiler metrics to be added by the observability project. This includes:
 * GetTrialProfilerMetrics API stub
 * GetTrialProfilerAvailableSeries API stub
 * PostTrialProfilerMetricsBatch API stub
 * trial_profiler_metrics table and trial_profiler_metric_labels on said table.
All of this work is discussed in depth in the ERD for the observability project.

## Test Plan
- [x] build the master, run the migrations, 👁️ the api specs in swagger

## Commentary (optional)
Benchmarked the table and will withstand everything I think we'll throw at it, but the final implementations will not preclude opening a backdoor to ship metrics to and read metrics from a real metrics store. In the average case, for the average user, the proposed solution benchmarks well within limitations and will lead to a much more manageable install experience.

## Checklist


armandmcqueen (approved):

armandmcqueen (changes requested):
Change the rank_id field to be gpu_uuid

stoksc:
@armandmcqueen yeah I'm aware. I gotta add an ID to the metrics too. I don't plan on landing this until we're closer to done with m2, in case things change up some.

master/static/migrations/20210301160616_add_trial_profiler_metrics.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE trial_profiler_metrics (
+    values  float4[]    NOT NULL,
```
armandmcqueen, line 2: I'm curious why it's better to store the metrics in postgres as batches rather than a row per individual metric. Is there a performance implication?

stoksc: Yeah, well if you go a row per metric then you're also storing all the labels per metric too. That was the main concern, but yeah, I expect 10 rows of batches of 100 to do better than 10k rows also (the index scan has to gather a lot less).

stoksc:
closed in favor of #2051

PR #2021, opened 2021-03-01
fix: fix forking failure on k8s [DET-5087]

eecsliu:
## Description
Forking fails on k8s due to the "default" resource manager not existing. This changes the kubernetes resource manager to return "" instead of "default" as the default resource pools so that forked experiments can pass the resource pool check.

## Checklist


armandmcqueen (approved):
LGTM!

PR #2026, opened 2021-03-02
feat: add endpoint for raw resource allocation information [DET-5043]

dzhu:
## Description

The new endpoint groups up all steps, checkpoints, and validations to
give information about slot time used in a given time period.

## Test Plan

- [x] have some workloads in the database, play around with various queries

## Commentary (optional)

The CSV endpoint needs to do something to deal with multiple labels. Right now it separates them with `|` and adds some escaping. Not sure if there's a standard thing to do here. Quoting?

### DIFF
--- a/master/internal/api_master.go
+++ b/master/internal/api_master.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"context"
+	"time"
 
 	"github.com/google/uuid"
 
@@ -86,4 +87,31 @@ func (a *apiServer) MasterLogs(
 		actor.Addr("logStore-"+uuid.New().String()),
 		api.NewLogStoreProcessor(resp.Context(), lReq, fetch, onBatch, nil, nil),
 	).AwaitTermination()
+}
+
+func (a *apiServer) ResourceAllocationRaw(
+	_ context.Context,
+	req *apiv1.ResourceAllocationRawRequest,
+) (*apiv1.ResourceAllocationRawResponse, error) {
+	resp := &apiv1.ResourceAllocationRawResponse{}
+
+	if req.StartDate == nil {
+		return nil, errors.New("no start date provided")
+	}
+	if req.EndDate == nil {
+		return nil, errors.New("no end date provided")
+	}
+	startDate := time.Unix(req.StartDate.Seconds, int64(req.StartDate.Nanos)).UTC()
+	endDate := time.Unix(req.EndDate.Seconds, int64(req.EndDate.Nanos)).UTC()
+	if startDate.After(endDate) {
+		return nil, errors.New("start date cannot be after end date")
+	}
+
+	if err := a.m.db.QueryProto(
+		"allocation_raw", &resp.ResourceEntry, startDate.UTC(), endDate.UTC(),
+	); err != nil {
+		return nil, errors.Wrap(err, "error fetching allocation data")
+	}
+
+	return resp, nil
 }
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"context"
 	"crypto/tls"
+	"encoding/csv"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -14,10 +15,14 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
+	"google.golang.org/protobuf/types/known/timestamppb"
+
 	"github.com/determined-ai/determined/master/internal/elastic"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
 
 	"github.com/google/uuid"
 	"github.com/gorilla/websocket"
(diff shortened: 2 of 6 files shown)


master/internal/core.go
```diff
@@ -146,6 +149,65 @@ func (m *Master) getMasterLogs(c echo.Context) (interface{}, error) {
...
+		if _, err := fmt.Fprintf(
+			c.Response(), "%d,%s,%s,%s,%d,%s,%s,%f\n",
+			entry.ExperimentId, entry.Kind, entry.Username, strings.Join(labels, "|"), entry.Slots,
+			formatTimestamp(entry.StartTime), formatTimestamp(entry.EndTime), entry.Seconds,
+		); err != nil {
+			return err
+		}
+		c.Response().Flush()
+	}
+	return nil
```
stoksc, line 208: any reason for this vs using a `csv.NewWriter(c.Response())`. I haven't checked but I assume `c.Reponse()` is an `io.Writer`.

master/static/srv/allocation_raw.sql
```diff
@@ -0,0 +1,79 @@
+WITH const AS (
+    SELECT
+        $1 :: timestamp AS day_start,
+        $2 :: timestamp AS day_end
+),
+-- Workloads that had any overlap with the target interval, with their start and end times bounded
+-- to the interval.
```
stoksc, line 7: Just front the comment: should just bound the end time to the interval? So we always count full workloads but never miss workloads that spanned interval boundaries?

master/static/srv/allocation_raw.sql
```diff
@@ -0,0 +1,79 @@
...
+                    ELSE all_workloads.end_time
+                END - CASE
+                    WHEN all_workloads.start_time < const.day_start THEN const.day_start
+                    ELSE all_workloads.start_time
+                END
+        ) AS seconds
+    FROM
+        (
+            -- Summarize the common relevant fields from all workload types. We might want this to
+            -- be a CTE, but I think that would cause PostgreSQL <12 to insert an optimization fence
```
stoksc, line 29: I believe this is true, so yeah, agree.

master/static/srv/allocation_raw.sql
```diff
@@ -0,0 +1,79 @@
...
+                'checkpoint' AS kind,
+                trial_id,
+                start_time,
+                end_time
+            FROM
+                checkpoints
+        ) AS all_workloads,
+        const
+    WHERE
+        all_workloads.start_time <= const.day_end
```
stoksc, line 57: Yeah, I think we should do `all_workloads.end_time IS NOT NULL AND all_workloads.end_time > const.day_start AND all_workloads.end_time <= const.day_end` or else don't we lose accounting when workloads span days?

master/internal/core.go
```diff
@@ -146,6 +149,65 @@ func (m *Master) getMasterLogs(c echo.Context) (interface{}, error) {
...
+
+	resp := &apiv1.ResourceAllocationRawResponse{}
+	if err := m.db.QueryProto(
+		"allocation_raw", &resp.ResourceEntry, afterDate.UTC(), beforeDate.UTC(),
+	); err != nil {
+		return errors.Wrap(err, "error fetching allocation data")
+	}
+
+	c.Response().Header().Set("Content-Type", "text/csv")
+	c.Response().WriteHeader(http.StatusOK)
```
stoksc, line 181: should we be writing this header before we successfully return?

stoksc (commented):
lgtm, just had some questions on specifics of what we decided to implement

master/static/srv/allocation_raw.sql
dzhu: I think the condition you're suggesting would exclude workloads that started during or before the period but ended after it, which we don't want.

master/static/srv/allocation_raw.sql
dzhu: Can't quite tell what you mean here -- let's discuss it live. Though I did notice the comment is now a bit out of date; the time column does only count the time during the period, but the stand and and time are the originals. I think that makes sense, though I can see alternatives. Let's go over that as well.

master/static/srv/allocation_raw.sql
stoksc: Yeah, it would but then if you query day by day you won't have any workloads not included in this aggregation (those you mention would get included in the next day). With it as is, if you run it daily, you'll miss information on workloads that don't fit into the interval [const.day_start, const.day_end) repeated over days (so workloads that start and end are in different days). If that is what this endpoint is for, this makes sense, but if someone is going to be calling if with [01-01-2020 00:00:00, 01-02-2020 00:00:00), [01-02-2020 00:00:00, 01-03-2020 00:00:00), etc.. this is going to miss stuff (e.g. a workload that spanned [01-01-2020 23:59:59, 01-02-2020 00:00:01] would never get included I think).

dzhu: No, that workload would be counted in _both_ days, contributing one second to each. The start time is before the ends of both periods and the end time is after the beginning of both periods.

master/static/srv/allocation_raw.sql
```diff
@@ -0,0 +1,79 @@
...
+workloads AS (
+    SELECT
+        all_workloads.trial_id,
+        all_workloads.kind,
+        all_workloads.start_time,
+        all_workloads.end_time,
+        extract(
+            epoch
+            FROM
+                CASE
```
stoksc, line 17: totally missed this code 🤦 . this is perfect lol.

master/static/srv/allocation_raw.sql
stoksc: til i'm an idiot, missed the code where you only count the portion of the interval overlapping the queried.

stoksc (approved):
lgtm

master/internal/api_master.go
```diff
@@ -87,3 +88,30 @@ func (a *apiServer) MasterLogs(
 		api.NewLogStoreProcessor(resp.Context(), lReq, fetch, onBatch, nil, nil),
 	).AwaitTermination()
 }
+
+func (a *apiServer) ResourceAllocationRaw(
+	_ context.Context,
```
stoksc, line 93: little late now, but in the future @dzhu may be nice to have a `QueryProtoContext` that uses `QueryxContext` underneath, especially for this one. Since it's costly, if the request died I imagine that makes sure the query gets terminated too.

PR #2027, opened 2021-03-02
chore: better go tooling

rb-determined-ai:
## Description

Replace FillDefaults -> WithDefaults.

Let WithDefaults() and Merge() return interface{} types. After go
generics are available, these will be typed.

Introduce generated getters for all fields of expconf structs, which are
correctly typed according to whether a defaulted object will have fields
defaulted or not.

## Commentary (optional)

This is the final version of the go tooling, which solves the final problem of go expconf objects, which was "how do we not introduce nil pointer exceptions all over the code in the future?".  The answer is "use generated getters which dereference pointers automatically when it is known that a default value will have been set."

This allows the expconf structs to use pointers for values that might not be set by the user, but all the master code which consumes the expconf structs will interact with typed getters to avoid either extreme of unnecessary nil pointer checks everywhere or blindly dereferencing variables everywhere.


rb-determined-ai:
This does not yet have the newly generated code, to make reviewing easier.

master/pkg/schemas/copy_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"gotest.tools/assert"
+)
+
+// Copy is the non-reflect version of copy, but mostly the reflect version is called from other
+// reflect code, so it's defined here in test code.
+func Copy(src interface{}) interface{} {
+	return cpy(reflect.ValueOf(src)).Interface()
+}
+
+func TestCopy(t *testing.T) {
```
stoksc, line 18: This doesn't seem to test anything

master/pkg/schemas/copy_test.go
```diff
@@ -0,0 +1,57 @@
...
+	assert.NilError(t, err)
+	assert.Assert(t, bytes.Equal(byts, []byte("null")))
+
+	src := []string{}
+	byts, err = json.Marshal(src)
+	assert.NilError(t, err)
+	assert.Assert(t, bytes.Equal(byts, []byte("[]")))
+}
+
+func TestCopyAllocatedSlice(t *testing.T) {
```
stoksc, line 30: may be nice to have some tests that test nested/filled types (and structs)

stoksc (commented):
mostly lgtm, just had minor q's about tests

master/pkg/schemas/copy_test.go
rb-determined-ai: oops, this is a weird vestige of a previous version.

master/pkg/schemas/copy_test.go
rb-determined-ai: done

stoksc (approved):
lgtm

PR #2029, opened 2021-03-03
fix: update experiment trials endpoint when changing filters or sorting

hkang1:
## Description

With the recent change to using a more generic version of Experiment list endpoint, we need to detect additional changes to the dependencies for the api (sorters, pagination, filters, etc) and make another API call instead of waiting for the polling to kick in to make the UI feel more responsive. The experiment list is updated to reflect this, the experiment trials table requires this fix as well.

## Test Plan

- [ ] Check to make sure sorting and pagination works for experiment trials table

## Checklist


hamidzr (approved):
LGTM thanks @hkang1 

IMO it'd be nice if we switch to the loading state while a new sort/filter is being applied in the future. this becomes more important the longer the requests take.

hamidzr (approved):
loading state looks great

PR #2032, opened 2021-03-03
chore: upgrade to Go 1.16

neilconway:
Not 100% clear to me why the output of `go mod tidy` ought to change as part of upgrading Go.


stoksc:
> Not 100% clear to me why the output of go mod tidy ought to change as part of upgrading Go.

@neilconway it's possible things are using `// +build` tags to switch implementations for different go version like https://github.com/gorilla/mux/blob/d07530f46e1eec4e40346e24af34dcc6750ad39f/mux_httpserver_test.go#L1. Maybe that answers it, maybe it doesn't though.

neilconway:
> @neilconway it's possible things are using `// +build` tags to switch implementations for different go version like https://github.com/gorilla/mux/blob/d07530f46e1eec4e40346e24af34dcc6750ad39f/mux_httpserver_test.go#L1. Maybe that answers it, maybe it doesn't though. Maybe something has `// +build go1.16` to use the new `embed` or `fs` package or something.

Got it, makes sense. Could also be transitive consequences of upgrading the mux dependency.

stoksc (approved):
lgtm, once the new lint error is fixed. we probably should wait until 0.14.4.rc0 gets cut though to land.

neilconway:
@stoksc lint error is fixed, I'll wait on branching 0.14.4 before landing.

stoksc:
@neilconway https://groups.google.com/g/golang-nuts/c/zn5xGPHDxoA

PR #2034, opened 2021-03-03
chore: bumpenvs

rb-determined-ai:
## Description

Jumped the gun on aws gov support, so this contains a commit which reverts it.


rb-determined-ai:
we do not want to default to pytorch 1.7 yet.

PR #2038, opened 2021-03-03
fix: correct trial log scrolling behaviors [DET-5096, DET-5097]

hkang1:
## Description

In trial logs, when clicking on scroll-to-end while already in the tailing mode, the log viewer gets cleared but it does not attempt to load the logs again because the direction is already set to `TAILING`. The `direction` value doesn't change, which in turn does not trigger an API call. The same is true for the `OLDEST` behavior.

## Test Plan

- [ ] Clicking on `Scroll-to-end` button multiple times should work.
- [ ] Clicking on `Scroll-to-oldest` button multiple times should work.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewerTimestamp.tsx
+++ b/webui/react/src/components/LogViewerTimestamp.tsx
@@ -239,18 +239,20 @@ const LogViewerTimestamp: React.FC<Props> = ({
   }, [ onDownloadClick ]);
 
   const handleEnableTailing = useCallback(() => {
+    if (direction === DIRECTIONS.TAILING) return;
     clearLogs();
     setDirection(DIRECTIONS.TAILING);
-  }, [ clearLogs ]);
+  }, [ clearLogs, direction ]);
 
   const handleFullScreen = useCallback(() => {
     if (baseRef.current && screenfull.isEnabled) screenfull.toggle();
   }, []);
 
   const handleScrollToTop = useCallback(() => {
+    if (direction === DIRECTIONS.OLDEST) return;
     clearLogs();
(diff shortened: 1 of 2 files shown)


vishnu2kmohan (approved):
Approved, modulo @hamidzr's sign-off.
Thanks for the quick turnaround @hkang1!

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -239,18 +239,20 @@ const LogViewerTimestamp: React.FC<Props> = ({
   }, [ onDownloadClick ]);
 
   const handleEnableTailing = useCallback(() => {
+    if (direction === DIRECTIONS.TAILING) return;
```
hamidzr, line 242: I think this would prevent "jump to bottom" behavior if you've already hit the arrow down button and you have scrolled up. 

or maybe more correctly tailing should stop when you scroll up. this used to be the case IIRC

hamidzr (commented):
thanks for working on this. the logs don't get cleared anymore but directional buttons seem to have lost some functionality.

hamidzr:
@vishnu2kmohan it's too late on east coast but did you get a chance to checkout the preview https://app.circleci.com/pipelines/github/determined-ai/determined/10301/workflows/e41db70f-3257-4820-81e2-bf0349f9e74e/jobs/290213/artifacts ?

vishnu2kmohan:
@hamidzr I didn't test the preview when I (conditionally) approved the PR, largely because the issue manifests when attempting to interact with the logs from a live trial (and I don't how to submit an experiment against the preview backend).

Is there more work to do on this PR then?

PR #2039, opened 2021-03-03
fix: nas dtrain bug

naren-determined:
## Checklist

### DIFF
--- a/harness/determined/exec/worker_process.py
+++ b/harness/determined/exec/worker_process.py
@@ -61,5 +61,5 @@ if __name__ == "__main__":
     try:
         main()
     except det.InvalidHP:
-        logging.info("InvalidHP detected, gracefully exiting trial")
+        logging.info("InvalidHP detected, worker is exiting")
         pass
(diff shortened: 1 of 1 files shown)


harness/determined/exec/worker_process.py
```diff
@@ -58,4 +58,8 @@ def main() -> None:
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except det.InvalidHP:
+        logging.info("InvalidHP detected, gracefully exiting trial")
```
rb-determined-ai, line 64: maybe reword this to not say the whole trial is exiting?

naren-determined: Well the whole trial is exiting, but the experiment is continuing?

rb-determined-ai: Only the worker is exiting when this gets printed.  If you have 64 workers it makes sense to see 64 'worker exiting' messages, but not to see 64 "gracefully exiting trial messages", only to then see a decidedly non-graceful exit occur.

rb-determined-ai (approved):

PR #2040, opened 2021-03-03
chore: tweak resource allocation endpoint

dzhu:
## Description

- Pluralize field name in API.
- Change query parameters to match trial log endpoint.
- Remove mention of steps from the returned values.
- Clarify an API comment.
- Refactor SQL query to use PostgreSQL's time range type.

## Test Plan

- [x] test results with some workloads in the database and different endpoints


master/static/srv/allocation_raw.sql
```diff
@@ -1,61 +1,51 @@
...
             FROM
-                CASE
-                    WHEN all_workloads.end_time IS NULL THEN const.day_end
-                    WHEN all_workloads.end_time > const.day_end THEN const.day_end
-                    ELSE all_workloads.end_time
-                END - CASE
-                    WHEN all_workloads.start_time < const.day_start THEN const.day_start
-                    ELSE all_workloads.start_time
-                END
+                -- `*` computes the intersection of the two ranges.
```
stoksc, line 16: Magical

stoksc (approved):
Lgtm

PR #2041, opened 2021-03-04
style: add link to batches column for experiment trials table [DET-5102]

hkang1:
## Description

Move batches column to be between columns `ID` and `Batches Processed` and add anchor tag to `batches` values.

Screen Shot 2021-03-03 at 5 24 11 PM

## Checklist


hamidzr:
I agree with the repositioning but I'm not sure about linking batches 🤔

vishnu2kmohan:
After thinking about it, I agree with @hamidz, the batch numbers will change depending on experiment progress (esp. during HP Search) so this may not have the effect we all desire, which is a bigger "hit area" for users wishing to look at a particular trial's details.

hkang1:
great points, I'll close out this PR.

PR #2045, opened 2021-03-04
fix: support trial filtering based on categorical hps [DET-5107]

hkang1:
## Description

Fix issue of categorical hp filtering was not working, so the trial table would remain empty whenever a filter is applied for categorical hps.

The core issue is that categorical hp values are converted to a numerical value to space them out nicely in the parallel coordinates. For example, [ 8, 16, 32 ] categorical values will be converted into [ 2, 4, 6 ] with a range of 1 through 7. 1, 3, 5, 7 are used as spacers to keep each categorical value visually separate. When the user applies a filter to this, the filter constraint comes back as something like [ 1.5, 2.5 ] for filtering on categorical value of 8. Simply need to translate it back to the category.

Screen Shot 2021-03-04 at 1 54 29 PM

## Test Plan

- [ ] check to make sure filtering on categorical hps work in HP viz parallel coordinates plot.

## Checklist

### DIFF
--- a/webui/react/src/components/ParallelCoordinates.tsx
+++ b/webui/react/src/components/ParallelCoordinates.tsx
@@ -90,6 +90,7 @@ const ParallelCoordinates: React.FC<Props> = ({
         const label = truncate(key, MAX_LABEL_LENGTH);
         const hpDimension: Record<string, unknown> = {
           label,
+          multiselect: false,
           range: dimension.range,
           values: data[key],
         };
--- a/webui/react/src/components/ParallelCoordinates.tsx
+++ b/webui/react/src/components/ParallelCoordinates.tsx
@@ -174,12 +174,10 @@ const ParallelCoordinates: React.FC<Props> = ({
 
         // Translate constraints back to categorical values.
         if (dim.categories && range && isNumber(range[0]) && isNumber(range[1])) {
+          // Create a list of acceptable categorical values.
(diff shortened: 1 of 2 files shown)


armandmcqueen (approved):

hamidzr (commented):
I run into an issue with filtering string based categorical hparams where the trial list doesn't change. working with @hkang1

hkang1:
> I run into an issue with filtering string based categorical hparams where the trial list doesn't change. working with @hkang1

Great catch on this one. Looks like it's due to string and boolean based categorical hparams. Should be fixed now.

webui/react/src/components/ParallelCoordinates.tsx
```diff
@@ -160,20 +166,33 @@ const ParallelCoordinates: React.FC<Props> = ({
...
+        const constraint: Constraint = { range };
+
+        // Translate constraints back to categorical values.
+        if (dim.categories && range && isNumber(range[0]) && isNumber(range[1])) {
+          const minIndex = Math.round((Math.ceil(range[0]) - 1) / 2);
+          const maxIndex = Math.round((Math.floor(range[1]) - 1) / 2);
+          const values = [];
+
+          // Create a list of acceptable categorical values.
+          for (let i = minIndex; i <= maxIndex; i++) values.push(dim.categories[i]);
```
hamidzr, line 182: nit: array.slice?

hamidzr (approved):
🚢

PR #2048, opened 2021-03-05
fix: detr and deformabledetr e2e tests

liamcli:
## Description
Test for single gpu was broken due to a bug in the custom metric reducer that only affected single gpu eval.  Also changed batch size for distributed deformabledetr test to 2 instead of 1 to allow 1 datapoint per gpu.  

## Test Plan
- [x] Test both single and multi-gpu eval for detr and deformable detr

### DIFF
--- a/examples/computer_vision/deformabledetr_coco_pytorch/model_def.py
+++ b/examples/computer_vision/deformabledetr_coco_pytorch/model_def.py
@@ -4,6 +4,7 @@ from attrdict import AttrDict
 import numpy as np
 import sys
 import os
+import copy
 import time
 
 sys.path.append("./ddetr")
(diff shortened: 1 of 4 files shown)


examples/computer_vision/deformabledetr_coco_pytorch/model_def.py
```diff
@@ -71,9 +72,9 @@ def cross_slot_reduce(self, per_slot_metrics):
             coco_evaluator.eval_imgs[iou_type] = np.concatenate(
                 coco_evaluator.eval_imgs[iou_type], 2
             )
-            create_common_coco_eval(
```
rb-determined-ai, line 74: should you remove the create_common_coco_eval as well?

examples/computer_vision/deformabledetr_coco_pytorch/model_def.py
```diff
@@ -71,9 +72,9 @@ def cross_slot_reduce(self, per_slot_metrics):
...
-                coco_eval, coco_evaluator.img_ids, coco_evaluator.eval_imgs[iou_type]
-            )
+            coco_eval.evalImgs = list(coco_evaluator.eval_imgs[iou_type].flatten())
+            coco_eval.params.imgIds = list(coco_evaluator.img_ids)
+            coco_eval._paramsEval = copy.deepcopy(coco_eval.params)
```
rb-determined-ai, line 77: why is the copy necessary?  Seems like since this is example code, and what's going on seems opaque enough, maybe we should have some comments as to what is going on?

rb-determined-ai (approved):

PR #2049, opened 2021-03-05
feat: add aws govcloud support for det-deploy

hoanghphan:
## Checklist

### DIFF
--- a/deploy/determined_deploy/aws/constants.py
+++ b/deploy/determined_deploy/aws/constants.py
@@ -4,7 +4,8 @@ class deployment_types:
     VPC = "vpc"
     EFS = "efs"
     FSX = "fsx"
-    DEPLOYMENT_TYPES = [SIMPLE, SECURE, VPC, EFS, FSX]
+    GOVCLOUD = "govcloud"
+    DEPLOYMENT_TYPES = [SIMPLE, SECURE, VPC, EFS, FSX, GOVCLOUD]
 
 
 class defaults:
@@ -74,4 +75,6 @@ class misc:
         "us-east-1",
         "us-east-2",
         "us-west-2",
+        "us-gov-east-1",
+        "us-gov-west-1",
     ]
--- a/deploy/determined_deploy/aws/templates/govcloud.yaml
+++ b/deploy/determined_deploy/aws/templates/govcloud.yaml
@@ -1,0 +1,708 @@
+---
+AWSTemplateFormatVersion: 2010-09-09
+Description: Determined Template
+Mappings:
+  RegionMap:
+    us-gov-east-1:
+      Master: ami-af9379de 
+      Agent: ami-0b71e56244b8ac1cb
+    us-gov-west-1:
+      Master: ami-a6e8d2c7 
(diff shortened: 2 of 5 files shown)


deploy/determined_deploy/aws/templates/govcloud.yaml
```diff
@@ -0,0 +1,708 @@
+---
+AWSTemplateFormatVersion: 2010-09-09
+Description: Determined Template
+Mappings:
+  RegionMap:
+    us-gov-east-1:
+      Master: ami-af9379de 
+      Agent: ami-0b71e56244b8ac1cb
+    us-gov-west-1:
+      Master: ami-a6e8d2c7
```
mackrorysd, line 10: Until we have a good answer to this (https://bugs.launchpad.net/cloud-images/+bug/1917080), could you add a blurb to the tools/scripts/README.md about what to manually look for and update when we run refresh-ubuntu-amis.py?

deploy/determined_deploy/aws/templates/govcloud.yaml
```diff
@@ -0,0 +1,708 @@
+---
```
mackrorysd, line 1: Meta-comment: I was gonna say "boy this looks like a lot of code that isn't Gov-specific, but I see we already have that for simple.yaml vs secure.yaml, etc. I guess there's not good solution to modularizing or inheritance beween templates...

deploy/determined_deploy/aws/constants.py
```diff
@@ -74,4 +75,6 @@ class misc:
         "us-east-1",
         "us-east-2",
         "us-west-2",
+        "us-gov-east-1",
```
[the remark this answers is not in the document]
hoanghphan: US Gov East and West have G instances that should have comparable GPUs (although not as many options as the other regions).

deploy/determined_deploy/aws/templates/govcloud.yaml
hoanghphan: Agreed I'm sure there's a more efficient way to do this but thought I'd start by following our existing convention.

deploy/determined_deploy/aws/templates/govcloud.yaml
```diff
@@ -0,0 +1,708 @@
...
+
+  DatabaseSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: Security Group For Database
+      Tags:
+        - Key: user
+          Value: !Ref AWS::StackName
+
+  Database:
```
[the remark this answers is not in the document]
hoanghphan: This one deviates from our current setup because Gov Cloud doesn't offer Serverless as an Engine option so we have to provision the DB ourselves. I think the preferred way is still Serverless if it is available.

deploy/determined_deploy/aws/templates/govcloud.yaml
hoanghphan: Yup just added a blurb there.

mackrorysd (approved):

PR #2050, opened 2021-03-05
chore: update fetch batches endpoint to allow ranges [DET-5089]

mackrorysd:
## Description

To deal with the problem of experiments where trials may report their metrics at unique points in their progress, users may want to get all the experiments that are *near* a certain point in progress.

## Test Plan

Manually verified that:
* If a trial is included twice (rare) it is always in order, so the most recent datapoint in the stream per trial is the one to render.
* Additional results are included


proto/src/determined/api/v1/experiment.proto
```diff
@@ -352,6 +352,8 @@ message TrialsSnapshotRequest {
       }];
   // Seconds to wait when polling for updates.
   int32 period_seconds = 5;
+  // A range either side of batches_processed to include near-misses.
+  int32 batches_margin = 6;
```
hkang1, line 356: **nit:** What do you think about putting `batches_margin` right after `batches_processed`?

stoksc: you cant change the order or it becomes a breaking change to grpc. which is probably fine, but i imagine that's why it is here.

master/internal/db/postgres_experiments.go
```diff
@@ -194,28 +195,31 @@ func snapshotWrapperToTrial(r snapshotWrapper) (*apiv1.TrialsSnapshotResponse_Tr
...
+	metricName string, startTime time.Time) (trials []*apiv1.TrialsSnapshotResponse_Trial,
+	endTime time.Time, err error) {
 	var rows []snapshotWrapper
 	err = db.queryRows(`
 SELECT
   t.id AS trial_id,
   t.hparams AS hparams,
   s.metrics->'avg_metrics'->$1 AS metric,
-  s.end_time AS end_time
+  s.end_time AS end_time,
```
stoksc, line 213: bit of a nit, but now these query can potentially return trial id and hparams multiple times per trial. could get a little inventive with an agg to prevent that but i think then it would logically follow to change the API too, which we probably don't want to do. how large of ranges are we planning on selecting? e.g. if a user runs a 3 day adaptive with a lot of hparams then calls this for min=0 max=inf are they going to oom the master?

stoksc (commented):
lgtm, just had one question about adding some validation onto the allowed ranges.

proto/src/determined/api/v1/experiment.proto
mackrorysd: Yeah typically when adding new fields to a protobuf you always go in increasing order, and changing what an existing number field means is an incompatibility. Of course, having seen our JavaScript bindings and the fact that they rely exclusively on the order of fields when making calls, and with the fact that this API is marked as Internal, I thought it would be best. I'll look into how to force an incompatible change through...

master/internal/db/postgres_experiments.go
mackrorysd: Yeah this has been discussed. If they use the API (which is marked as "Internal", fwiw) they could select a massive range and just return all metrics. On the front-end we're going to have a drop-down to select a few pre-defined ranges OR a heuristic that selects the range for us.

The reason for adding this is specifically to deal with fairly small ranges: you want to select a point of progress in the trial, and we'll show all trials that recorded data at that particular point. But little cases pop up where there's only 1 trial at each point, but they're just a couple of batches offset from each other.

proto/src/determined/api/v1/experiment.proto
mackrorysd: Oh it's just `make -C proto gen-buf-image`

master/internal/db/postgres_experiments.go
stoksc: > On the front-end we're going to have a drop-down to select a few pre-defined ranges OR a heuristic that selects the range for us.

any way to enforce these ranges are well behaved master side?

mackrorysd: Still wanting to confirm with AMLEs, but we discussed a reasonable limit that's above what the front-end will suggest.

stoksc (approved):
lgtm

PR #2051, opened 2021-03-05
feat: add trial profiler metrics APIs [DET-5065, DET-5059]

stoksc:
## Description
This change adds implementations for the trial profiler metrics to be added by the observability project. This includes:
 * GetTrialProfilerMetrics API and tests
 * GetTrialProfilerAvailableSeries API and tests
 * PostTrialProfilerMetricsBatch API and tests
 * trial_profiler_metrics table and trial_profiler_metric_labels index on said table.
All of this work is discussed in depth in the ERD for the observability project. These changes are design to mimic trial logs design, and not preclude an escape hatch to ship to a real TSDB for large installs, though for the most part these features for for small and ephemeral about of profiling data just to add users in porting.

## Test Plan
- [x] add tests, run the test
- [x] write a "trial" in go that spins up N threads and ships in the expected manner, prof master while doing this and tune (plus do some math we're within the ballpark of these being possible still). I did this, everything look totally fine. Really we're not shipping that many metrics now that we scaled the required sampling frequency down so much, so we're totally within the realm of what Postgres will handle just fine.

## Commentary (optional)
There are lots of patterns / code I reused from trial logs here that has the word log in it or uses log related stuff. I choose to leave it as is and follow it up with a refactoring PR, that I'm going to start working on now.

## Checklist


master/pkg/model/experiment.go
```diff
@@ -517,6 +519,24 @@ func (t TrialLogBatch) ForEach(f func(interface{}) error) error {
 	return nil
 }
 
+// TrialProfilerMetricsBatchBatch represents a batch of trialv1.TrialProfilerMetricsBatch.
+type TrialProfilerMetricsBatchBatch []*trialv1.TrialProfilerMetricsBatch
```
stoksc, line 523: generics, gah! this whole batch this is just because i need to len(T[]) a few places. In 1.18 i'll make this beautiful.

master/test/integration/api/api_trials_test.go
```diff
@@ -134,6 +162,135 @@ func trialDetailAPITests(
...
+		_, err = cl.PostTrialProfilerMetricsBatch(creds, &apiv1.PostTrialProfilerMetricsBatchRequest{
+			Batch: notMatch,
+		})
+		assert.NilError(t, err, "failed to insert mocked unmatched trial profiler metrics")
+
+		recvMetricsBatch, err := tlCl.Recv()
+		assert.NilError(t, err, "failed to stream metrics")
+
+		// Just nil the values since the floats and timestamps lose a little precision getting thrown
+		// around so much.
```
stoksc, line 204: I could probably do a better job at this, but somewhere from proto->go->postgres->go->proto timestamps are losing some precision. i can change it before merge to round to nearest 1/1000th or something.

master/internal/api_trials.go
```diff
@@ -419,3 +409,141 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+		return b.ForEach(func(r interface{}) error {
+			return resp.Send(&apiv1.GetTrialProfilerMetricsResponse{
+				Batch: r.(*trialv1.TrialProfilerMetricsBatch),
+			})
+		})
+	}
+
+	return a.m.system.MustActorOf(
+		actor.Addr("trialMetricsStream-"+uuid.New().String()),
+		api.NewLogStoreProcessor(
```
stoksc, line 454: There is a follow up where i just use goland to rename and move a bunch of this "log" crap. Just putting it in the same PR makes it unreviewable.

master/internal/api_trials.go
```diff
@@ -29,12 +33,16 @@ import (
 )
 
 const (
-	batchSize = 1000
+	batchSize                     = 1000
```
armandmcqueen, line 36: If we have two different batch sizes, should be change the name of this one to be more specific? Or is that part of the larger refactor follow-on?

master/internal/api_trials.go
```diff
@@ -419,3 +409,141 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+	req *apiv1.PostTrialProfilerMetricsBatchRequest,
+) (*apiv1.PostTrialProfilerMetricsBatchResponse, error) {
+	switch exists, err := a.m.db.CheckTrialExists(int(req.Batch.Labels.TrialId)); {
+	case err != nil:
+		return nil, err
+	case !exists:
+		return nil, status.Error(codes.NotFound, "trial not found")
+	}
+
+	if len(req.Batch.Values) != len(req.Batch.Batches) &&
```
armandmcqueen, line 514: If we are trying to detect if the arrays differ in size at all, should this be an OR rather than an AND?

master/test/integration/api/api_trials_test.go
```diff
@@ -134,6 +162,135 @@ func trialDetailAPITests(
...
+		Labels: &trialv1.TrialProfilerMetricLabels{
+			TrialId:    int32(trial.ID),
+			Name:       "gpu_util",
+			AgentId:    "brad's agent",
+			MetricType: trialv1.TrialProfilerMetricLabels_PROFILER_METRIC_TYPE_SYSTEM,
+		},
+	})
+	assert.NilError(t, err, "failed to initiate trial profiler metrics stream")
+
+	for i := 0; i < 10; i++ {
```
armandmcqueen, line 187: This testing logic is substantial enough that I would suggest adding comments describing what we are testing for

master/test/integration/api/api_trials_test.go
```diff
@@ -134,6 +162,135 @@ func trialDetailAPITests(
...
+		expected = append(expected, tb.Labels.Name)
+		_, err := cl.PostTrialProfilerMetricsBatch(creds, &apiv1.PostTrialProfilerMetricsBatchRequest{
+			Batch: tb,
+		})
+		assert.NilError(t, err, "failed to insert mocked trial profiler metrics")
+
+		// This may need 2 or more attempts; gRPC streaming does not provide any backpressure mechanism, if the client
+		// is not ready to receive a message when it is sent to the stream server side, the server will buffer it and
+		// any calls to Send will return, allowing the code to chug along merrily and Send more and more data while the
+		// client doesn't consume it. Because of this, we give the test a change to read out stale entries before
```
armandmcqueen, line 265:
```suggestion
		// client doesn't consume it. Because of this, we give the test a chance to read out stale entries before
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -103,3 +103,50 @@ message Trial {
...
+  // The ID of the trial.
+  int32 trial_id = 1;
+  // The name of the metric.
+  string name = 2;
+  // The agent ID associated with the metric.
+  string agent_id = 3;
+  // The GPU UUID associated with the metric.
+  string gpu_uuid = 4;
+  // The type of the metric.
+  ProfilerMetricType metric_type = 5;
```
armandmcqueen, line 132: What do you think about using a bool instead of an enum here? Protobuf enums kind of suck from a UX perspective

proto/src/determined/trial/v1/trial.proto
```diff
@@ -103,3 +103,50 @@ message Trial {
...
+// TrialProfilerMetricsBatch is a batch of trial profiler metrics. A batch will
+// contain metrics pertaining to a single series. The fields values, batches and
+// timestamps will be equal length arrays with each index corresponding to a
+// reading.
+message TrialProfilerMetricsBatch {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [ "values", "batches", "timestamps", "labels" ] }
+  };
+  // The measurement for a reading, repeated for the batch of metrics.
+  repeated float values = 1;
```
armandmcqueen, line 144: Why did we choose this format of three independent lists instead of a list of Metric objects where each Metric object holds `(value, batch, timestamp)`?

stoksc: nothing against that, i guess this is just exactly how it's stored. @hkang1 thoughts?

hkang1: WebUI can take advantage of a list of objects to only traverse all of the data once instead of 4x (once for each field). 🙏

proto/src/determined/trial/v1/trial.proto
stoksc: Not really a fan of loading meaning other than true/false into a bool.

proto/src/determined/trial/v1/trial.proto
stoksc: actually.. i really don't want to do this. if we wrap these, they're all going to need to be in an object, so instead of
```
{ batches: [ ... ], values: [ ... ], timestamps: [ ... ] }
```
We end up with 
```
{ metrics: [ { batch: X, value: X, timestamp: X }, { ... }] }
```
which is a lot more overhead when we're using the grpc-gateway APIs without any compression.

I know I've been pretty picky about the kinda stuff, I just don't want to get pushed too far away from what has been benchmarked again.

hkang1: the original approach works for the WebUI too. We just need to know that `batches`, `values`, and `timestamps` will all return the same sized lists.

armandmcqueen (approved):
LGTM!

PR #2056, opened 2021-03-08
feat: add DELETE /api/v1/experiments/:id API [DET-4022]

stoksc:
## Description
This change adds an admin-only API to delete experiments: DELETE /api/v1/experiments/:id . This endpoint deletes everything from Postgres, starts and awaits a checkpoint GC task, and deletes trial logs from the configured backend.

## Test Plan
- [x] run and test with elastic and postgres, and loads of checkpoints

## Commentary (optional)
The experiment gc code is begging to be cleaned up, but I'm really not trying to save the world here. This endpoint blocks until the experiment is _actually_ deleted. If we think this is a problem, rework experiment gc a bit to make it less in the way. As written this endpoint should be ok to be interrupted and restarted and come to the same result, in the event that happens.

## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -84,11 +84,11 @@ func (a *apiServer) GetExperiment(
 func (a *apiServer) DeleteExperiment(
 	_ context.Context, req *apiv1.DeleteExperimentRequest,
 ) (*apiv1.DeleteExperimentResponse, error) {
-	id := int(req.ExperimentId)
-	exp, err := a.m.db.ExperimentByID(id)
+	expID := int(req.ExperimentId)
+	exp, err := a.m.db.ExperimentByID(expID)
 	switch {
 	case errors.Cause(err) == db.ErrNotFound:
-		return nil, status.Errorf(codes.NotFound, "experiment %d not found", id)
+		return nil, status.Errorf(codes.NotFound, "experiment %d not found", expID)
 	case err != nil:
 		return nil, fmt.Errorf("failed to retrieve experiment: %w", err)
 	}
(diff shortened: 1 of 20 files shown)


master/internal/db/postgres.go
```diff
@@ -988,6 +1004,25 @@ WHERE trials.experiment_id = $1
 	return numTrials, nil
 }
 
+// ExperimentTrialIDs returns the trial IDs for the experiment.
+func (db *PgDB) ExperimentTrialIDs(expID int) ([]int, error) {
+	var trialIDRows []struct {
```
stoksc, line 1002: apparently lib/pq has a convenient for this to `Scan(pq.Array(...))` but I didn't see an equivalent in pgx (but I didn't look too hard tbh)

master/internal/grpcutil/auth.go
```diff
@@ -33,6 +33,10 @@ var unauthenticatedMethods = map[string]bool{
 	"/determined.api.v1.Determined/GetTelemetry": true,
 }
 
+var adminMethods = map[string]bool{
+	"/determined.api.v1.Determined/DeleteExperiment": true,
```
dzhu, line 37: non-blocking: Would it be feasible to do something like use the API annotations for this sort of thing (along the lines of the current `tags: "Experiments"`) rather than this separate thing? This'll fail insecure if anything changes or we add new APIs that should be protected. Probably not a big deal yet, but I twitch a bit at seeing this.

master/internal/api_trials.go
```diff
@@ -46,6 +46,7 @@ type TrialLogBackend interface {
 	AddTrialLogs([]*model.TrialLog) error
 	TrialLogCount(trialID int, filters []api.Filter) (int, error)
 	TrialLogFields(trialID int) (*apiv1.TrialLogsFieldsResponse, error)
+	DeleteTrialLogs(trialIDs []int) error
```
dzhu, line 49: random thought: Since the type is `TrialLogBackend`, the method names could probably all bear to have `Trial` removed from them. :stuck_out_tongue:

master/internal/api_experiment.go
```diff
@@ -78,6 +81,72 @@ func (a *apiServer) GetExperiment(
...
+	}
+	addr := actor.Addr(fmt.Sprintf("delete-checkpoint-gc-%s", uuid.New().String()))
+	if gcErr := a.m.system.MustActorOf(addr, &checkpointGCTask{
+		agentUserGroup: agentUserGroup,
+		taskSpec:       a.m.taskSpec,
+		rm:             a.m.rm,
+		db:             a.m.db,
+		experiment:     exp,
+	}).AwaitTermination(); gcErr != nil {
+		return nil, errors.Wrapf(gcErr, "failed to gc checkpoints for experiment: %d", exp.ID)
```
dzhu, line 129: Eh, not sure the errors need to have the experiment ID embedded, since this is in response to a request for a specific experiment anyway.

master/internal/elastic/elastic_trial_logs.go
```diff
@@ -182,6 +182,50 @@ func (e *Elastic) TrialLogs(
...
+			},
+		},
+	}
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(query); err != nil {
+		return errors.Wrap(err, "failed to encoding query")
+	}
+
+	// TODO(brad): Here and elsewhere, we really should just hit indices that could possibly have
```
dzhu, line 215: How big a performance land mine is this in its current state?

dzhu (approved):

master/internal/grpcutil/auth.go
stoksc: I mean, this is just an access control list lol, just a crummy one. Yeah, having at the API layer would be neat, but I was moving with the assumption something like that would get wrapped into an RBAC project.

PR #2057, opened 2021-03-09
feat: add batch margins [DET-5073]

hkang1:
## Checklist


webui/react/src/components/Grid.tsx
```diff
@@ -12,6 +12,7 @@ export enum GridMode {
 interface Props {
   border?: boolean;
   gap?: ShirtSize;
+  length?: number;
```
apizzini, line 15: Looking to how this is used, couldn't just update `mode?: GridMode|number`?

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
```diff
@@ -236,86 +183,40 @@ const HpHeatMaps: React.FC<Props> = ({
...
-                </ConditionalWrapper>
-              ))}</>
-            </ConditionalWrapper>
+            <Grid
+              border={true}
+              length={!isListView ? selectedHParams.length : undefined}
+              minItemWidth={resize.width > 320 ? 35 : 27}
+              mode={GridMode.AutoFill}>
+              {selectedHParams.map(hParam1 => (
+                <>{selectedHParams.map(hParam2 => {
```
apizzini, line 192: I think the fragment too needs the key, maybe can just avoid it?

`{selectedHParams.map(hParam1` => (
  selectedHParams.map(hParam2 => {`

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/ExperimentVisualizationFilters.tsx
```diff
@@ -0,0 +1,160 @@
+import { Select } from 'antd';
```
apizzini, line 1: A nice to have would be to debounce on changes, eg. when you select/deselect N HP quickly, it will trigger 3 rerenders of charts which is pretty heavy.

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/ExperimentVisualizationFilters.tsx
```diff
@@ -0,0 +1,160 @@
...
+            value={selectedBatchMargin}
+            onChange={handleBatchMarginChange}>
+            {BATCH_MARGIN_OPTIONS.map(option => (
+              <Option key={option} value={option}>{option}</Option>
+            ))}
+          </SelectFilter>
+        </>
+      )}
+      {showMetrics && (
+        <MetricSelectFilter
```
apizzini, line 138: Not really related to this but most of the times I change this value I get a sequence of errors 
Schermata 2021-03-10 alle 15 19 21

apizzini: and after that the page tend to crash.

webui/react/src/components/Grid.tsx
hkang1: yep, that's probably what a smart person would've done! :D

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
hkang1: doh, forgot to remove the fragment

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/ExperimentVisualizationFilters.tsx
hkang1: There was an uncaught error around HP parcoords data processing (unexpected undefined) that cause the crash. That should be fixed now.

hkang1: The `gd.emit` error occurs when plotly cleanup `Purge` runs against the div container that doesn't have any charts rendered. Plotly doesn't expose if a chart has rendered or not publicly, so I don't currently have a way to run the cleanup appropriately. I'll take another look at this in a separate PR for the sake of release but this warning shouldn't cause any issues.

webui/react/src/components/ParallelCoordinates.tsx
```diff
@@ -199,15 +204,15 @@ const ParallelCoordinates: React.FC<Props> = ({
     });
 
     return () => {
-      if (ref) Plotly.purge(ref);
+      Plotly.purge(ref);
     };
     /* eslint-disable-next-line react-hooks/exhaustive-deps */
```
apizzini, line 209: Since that's an anti-pattern we should (?) avoid, what do you think about adding a small comment to document why that's needed? (there are other occurrences in <ScatterPlot>)

apizzini: I tried to understand the reason for skipping exaustive deps, is that mainly related to:
- uplot "watching" the chartData (no need to trigger a refresh)
- not wanting to re-generate the chart but needing to pass the data on the initialization?

I think that can be solved in 2 ways:
1. generating chartData in a useEffect and storing into a useRef (changing the value will not trigger the useeffect).
2. providing no data to uplot on init and (i believe that's possible) call a method on the uplot instance to update the chart data?

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/ExperimentVisualizationFilters.tsx
apizzini: sounds good, and yes i see a couple of uplot errors here and there but having no side effects now.

apizzini (approved):
lgtm. i would just try to avoid tweaking useEffect dependencies but the current code looks to be working fine.

webui/react/src/components/ParallelCoordinates.tsx
hkang1: > I tried to understand the reason for skipping exaustive deps, is that mainly related to:
> 
> * uplot "watching" the chartData (no need to trigger a refresh)
> * not wanting to re-generate the chart but needing to pass the data on the initialization?
> 
> I think that can be solved in 2 ways:
> 
> 1. generating chartData in a useEffect and storing into a useRef (changing the value will not trigger the useeffect).
> 2. providing no data to uplot on init and (i believe that's possible) call a method on the uplot instance to update the chart data?

Very good points, and would love to chat about this in the WebUI AG. Curious if you have been able to avoid skipping the exhaustive dependencies linting completely.

hkang1:
> lgtm. i would just try to avoid tweaking useEffect dependencies but the current code looks to be working fine.

Cleaned up the dependencies with useRef. Thanks for pointing them out!

PR #2061, opened 2021-03-09
feat: add stopping-based variant of adaptive search [DET-3669]

liamcli:
## Description
Add a stopping-based variant of adaptive search that does not require fault-tolerance for early-stopping.  This variant is based on  the implementation of ASHA in Ray Tune: https://docs.ray.io/en/master/tune/api_docs/schedulers.html#tune-scheduler-hyperband.  

## Test Plan
- [x] Add simulated tests
- [x] Run local HP search experiment

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label. **I'm not sure if this counts as user-facing api change.**

### DIFF
--- a/master/pkg/searcher/asha_stopping.go
+++ b/master/pkg/searcher/asha_stopping.go
@@ -2,7 +2,6 @@ package searcher
 
 import (
 	"encoding/json"
-	"fmt"
 	"math"
 	"sort"
 
@@ -27,9 +26,6 @@ type asyncHalvingStoppingSearch struct {
 
 func newAsyncHalvingStoppingSearch(config model.AsyncHalvingConfig) SearchMethod {
 	rungs := make([]*rung, 0, config.NumRungs)
-	if !config.StopOnce {
-		panic(fmt.Sprintf("stop_once is %t, should not be running stopping-based ASHA", config.StopOnce))
-	}
 	for id := 0; id < config.NumRungs; id++ {
 		// We divide the MaxLength by downsampling rate to get the target units
 		// for a rung.
@@ -174,28 +170,28 @@ func (s *asyncHalvingStoppingSearch) promoteAsync(
 	} else {
 		// This is not the top rung, so do promotions to the next rung.
 		nextRung := s.Rungs[rungIndex+1]
+		// We need to run continueTraining even if the trial was terminated early so that we
+		// can add the metric to the rung.
 		promoteTrial := rung.continueTraining(
 			requestID,
 			metric,
 			s.Divisor,
 		)
-		if promoteTrial {
-			s.TrialRungs[requestID] = rungIndex + 1
-			nextRung.OutstandingTrials++
-			if !s.EarlyExitTrials[requestID] {
+		// In contrast to promotion-based ASHA, we will not let early-exited trials add
+		// -/+inf metrics to higher rungs even if portion of terminated trials in bottom rung
+		// is greater than 1 - 1 / divisor.
+		if !s.EarlyExitTrials[requestID] {
+			if promoteTrial {
+				s.TrialRungs[requestID] = rungIndex + 1
+				nextRung.OutstandingTrials++
 				unitsNeeded := max(nextRung.UnitsNeeded.Units-rung.UnitsNeeded.Units, 1)
 				ops = append(ops, NewTrain(requestID, model.NewLength(s.Unit(), unitsNeeded)))
 				ops = append(ops, NewValidate(requestID))
 				addedTrainWorkload = true
 			} else {
-				// We make a recursive call that will behave the same
-				// as if we'd actually run the promoted job and received
-				// the worse possible result in return.
-				return s.promoteAsync(ctx, requestID, ashaExitedMetricValue)
+				ops = append(ops, NewClose(requestID))
+				s.ClosedTrials[requestID] = true
 			}
-		} else {
-			ops = append(ops, NewClose(requestID))
-			s.ClosedTrials[requestID] = true
 		}
 	}
 
@@ -240,8 +236,8 @@ func (s *asyncHalvingStoppingSearch) trialExitedEarly(
 			for i, trialMetric := range rung.Metrics {
 				if trialMetric.RequestID == requestID {
 					rung.Metrics = append(rung.Metrics[:i], rung.Metrics[i+1:]...)
+					break
 				}
-				break
 			}
 		}
 		// Add new trial to searcher queue
--- a/master/pkg/searcher/asha_stopping.go
+++ b/master/pkg/searcher/asha_stopping.go
@@ -1,0 +1,259 @@
+package searcher
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"sort"
+
+	"github.com/determined-ai/determined/master/pkg/workload"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// AsyncHalvingStoppingSearch implements a modified version of the asynchronous successive
+// halving algorithm (ASHA) that does not require fault tolerance to perform early-stopping.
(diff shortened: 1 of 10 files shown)


rb-determined-ai:
Does this algorithm work if you have only one slot?  And if so, how?

master/pkg/searcher/asha_stopping.go
```diff
@@ -0,0 +1,259 @@
...
+type asyncHalvingStoppingSearch struct {
+	defaultSearchMethod
+	model.AsyncHalvingConfig
+	asyncHalvingSearchState
+}
+
+func newAsyncHalvingStoppingSearch(config model.AsyncHalvingConfig) SearchMethod {
+	rungs := make([]*rung, 0, config.NumRungs)
+	if !config.StopOnce {
+		panic(fmt.Sprintf("stop_once is %t, should not be running stopping-based ASHA", config.StopOnce))
```
stoksc, line 31: this seems unnecessary but i'm fine with it if you think it will prevent misuse one day.

master/pkg/searcher/asha_stopping.go
```diff
@@ -0,0 +1,259 @@
...
+		if promoteTrial {
+			s.TrialRungs[requestID] = rungIndex + 1
+			nextRung.OutstandingTrials++
+			if !s.EarlyExitTrials[requestID] {
+				unitsNeeded := max(nextRung.UnitsNeeded.Units-rung.UnitsNeeded.Units, 1)
+				ops = append(ops, NewTrain(requestID, model.NewLength(s.Unit(), unitsNeeded)))
+				ops = append(ops, NewValidate(requestID))
+				addedTrainWorkload = true
+			} else {
+				// We make a recursive call that will behave the same
```
stoksc, line 191: thinking through this, I'm not sure if this code path is even possible anymore, if a trial exited early, it would've completed no operations and `trialExitedEarly` would've been called for it, if it is here and has completed a validation.. well it wouldn't have gotten here if it exited early.

master/pkg/searcher/asha_stopping.go
```diff
@@ -0,0 +1,259 @@
...
+		// Remove metrics associated with InvalidHP trial across all rungs
+		highestRungIndex := s.TrialRungs[requestID]
+		for rungIndex := 0; rungIndex <= highestRungIndex; rungIndex++ {
+			rung := s.Rungs[rungIndex]
+			for i, trialMetric := range rung.Metrics {
+				if trialMetric.RequestID == requestID {
+					rung.Metrics = append(rung.Metrics[:i], rung.Metrics[i+1:]...)
+				}
+				break
+			}
```
stoksc, line 245: wouldn't this only remove the metric from the rung if it was the first one, otherwise it would break? it seems like the break is in the wrong place.

stoksc (commented):
mostly lgtm, just had a few questions. this is a cool variant of asha.

stoksc:
> Does this algorithm work if you have only one slot? And if so, how?

I think it would work fine, the first trial would run all the way i think (maybe the first few even for some configs?) but then as it progress they would need to be in the top 1/downsampling rate of the rung they completed among the trials that have already completed to get promoted. It seems to me you could have runs where the bottom rungs aren't 100% full though, if the later trials are all poor performers. the key is https://github.com/determined-ai/determined/pull/2061/files#diff-06f786f0d211d0667ae8bdb7d3221330e818545a2dc1cccda85f6d0e8dd3ce46R69

is that even half correct @liamcli ?

liamcli:
> > Does this algorithm work if you have only one slot? And if so, how?
> 
> I think it would work fine, the first trial would run all the way i think (maybe the first few even for some configs?) but then as it progress they would need to be in the top 1/downsampling rate of the rung they completed among the trials that have already completed to get promoted. It seems to me you could have runs where the bottom rungs aren't 100% full though, if the later trials are all poor performers. the key is https://github.com/determined-ai/determined/pull/2061/files#diff-06f786f0d211d0667ae8bdb7d3221330e818545a2dc1cccda85f6d0e8dd3ce46R69
> 
> is that even half correct @liamcli ?

yep, the main difference between promotion-based asha and stopping-based asha is that stopping defaults to continuing training unless there is evidence that a trial is unlikely to be in top 1/divisor of trials in a rung (this is determined using empirical distribution of metrics for trials in a rung), while promotions errs on the side of waiting until there is enough evidence to promote.  
hence stopping will always train the first trial to the maximum number of steps since there isn't enough evidence to stop.  subsequent trials will be compared to existing metrics to determine stopping.  one drawback of stopping is that it's possible for a trial to be stopped even if we would've made the opposite decision had we gathered more evidence; furthermore, we will never resume this trial and will simply have to wait for more trials to finish with better metrics to fill in promotions.

master/pkg/searcher/asha_stopping.go
liamcli: took a look and agree that given how we create the searchers, this is unnecessary.

master/pkg/searcher/asha_stopping.go
liamcli: good catch.  looks like this bug is in asha.go as well.  do you think i should fix asha.go in this PR or have a separate PR for that?

master/pkg/searcher/asha_stopping.go
liamcli: good catch.  i've modified how we handle early exit trials for stopping to never promote.  this is okay for stopping since we are guaranteed to have a trial train for the max_length as long as there is a trial that is not exited early.  this is not okay for promotion-based asha since we need to fill in enough trials in the rungs to promote one to the max_length.

master/pkg/searcher/asha_stopping.go
```diff
@@ -0,0 +1,255 @@
...
+	s.TrialRungs[requestID] = 0
+	return nil, nil
+}
+
+func (s *asyncHalvingStoppingSearch) trialClosed(
+	ctx context, requestID model.RequestID) ([]Operation, error) {
+	s.TrialsCompleted++
+	s.ClosedTrials[requestID] = true
+	return nil, nil
+}
```
stoksc, line 129: thinking to the future a bit, does asha (or any searcher) really need to be aware of a trial _actually_ starting or stopping, or could this happen with newcreate/newclose. trying to think of ways to minimize the searcher interface as we move towards sprinkle api?

stoksc (approved):
sick, lgtm.

master/pkg/searcher/asha_stopping.go
stoksc: on the same note, how do you feel about removing train operations (or at the the callback to expect them back, i dont think anything meaningfully hooks into training completed except maybe progress or something)

stoksc: i guess as a more directed question, are there any hp search algorithms that need to know training metrics, when the scheduler started trials, or when the trial responded to a close and terminated?

liamcli: agree that training metrics aren't really relevant for HP search.  

there are definitely hp algos that need to track how many hps have been evaluated/are being evaluated and what each step the validation metrics correspond to.  in the case of asha for example, the algo presumably has a max_trials so it needs to know how many trials total have been created in the bottom rung, including ones that are in progress and do not have a validation metric associated with it.  similarly, for PBT, the algorithm needs to know how many trials have been launched for a particular evolution round.  

in a sprinkle world, hp workers would go to our hp searcher and say give me a workload and once it finishes, it returns the validation metric associated to the searcher and requests the next workload.  the searcher will be in charge of telling a worker there are no more workloads remaining and to exit.  being able to detect the stopping condition would likely require knowing when a trial is created.  in this scheme, the workers will also need to tell the searchers which trials to exclude from further consideration (e.g. to handle invalidhp or early exit trials).

PR #2071, opened 2021-03-12
fix: force line-break to prevent overlapping in trial logs [DET-5121]

apizzini:
not an easy find, but noticed that the issue should be related to separators like quotation marks not properly breaked (see red rect here, the quotation should be there, not going to newline).
Schermata 2021-03-12 alle 11 32 28

Found this https://github.com/w3c/csswg-drafts/issues/1171#issuecomment-494164252
That lead me to this https://developer.mozilla.org/en-US/docs/Web/CSS/line-break

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer.module.scss
+++ b/webui/react/src/components/LogViewer.module.scss
@@ -64,6 +64,7 @@
   .line > .time::after { content: '\00a0'; }
   .line > .message {
     flex-grow: 1;
+    line-break: anywhere;
     white-space: pre-wrap;
     word-break: break-all;
   }


hkang1 (commented):
Nice find! I reproduced it once initially by resizing the window on the fly as you can see below, but I don't if that's some weird cached page on my end because I couldn't reproduce it again...
Screen Shot 2021-03-12 at 12 05 35 PM
@naren-determined would you mind trying this PR out to see if you can reproduce this bug? (via circle CI react-preview, lemme know if you don't know how to do this)

naren-determined (approved):
Thanks for the super quick turnaround!

hkang1 (approved):
Talked with @naren-determined this looks to resolve his issue, thanks @apizzini !

PR #2072, opened 2021-03-12
docs: minor improvements

neilconway:
## Checklist

### DIFF
--- a/docs/how-to/installation/aws.txt
+++ b/docs/how-to/installation/aws.txt
@@ -99,12 +99,12 @@ using the ``--deployment-type`` argument (e.g., ``det-deploy aws
 -  ``efs``: The EFS deployment creates an `EFS
    <https://aws.amazon.com/efs/>`_ drive and a Determined cluster into a
    separate VPC. The EFS drive attaches to agent instances at
-   `/mnt/efs`.
+   ``/mnt/efs``.
 
 -  ``fsx``: The FSX deployment creates a `Lustre FSx
    <https://aws.amazon.com/fsx/lustre/>`_ drive and a Determined cluster
    into a separate VPC. The FSx drive attaches to agent instances at
-   `/mnt/fsx`.
+   ``/mnt/fsx``.
 
 -  ``secure``: The secure deployment creates resources to lock down the
    Determined cluster. These resources are:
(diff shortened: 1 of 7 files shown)


docs/how-to/installation/aws.txt
```diff
@@ -205,20 +205,23 @@ Spinning up the Cluster
...
-         spot market exceeds this value, Determined will not create new
-         instances. Optional
+      -  The maximum price to use when launching spot instances. If the
+         price on the spot market exceeds this value, Determined will
+         not create new instances. If no price is configured, the
+         maximum price will be the on-demand price for the configured
+         instance type and region.
 
-      -  The on-demand price for the instance type.
+      -  Not set
```
dzhu, line 224: I actually rather like having the default column say "the on-demand price", though maybe that's too long or something with "for the configured..."?

docs/topic-guides/model-definitions/index.txt
```diff
@@ -9,12 +9,10 @@ application framework (e.g., Keras, TensorFlow), in terms of loading
...
 underlying iterative optimization training algorithms.
 
-There interface you can use to write model definition is the Trial API.
-With the :ref:`model-definitions_trial-api`, developers expose the
-components of a model definition to Determined’s training loop
-implementation.
-
-The command line interface and Python SDK support the Trial API.
+Model definitions are written using the Trial API, which is an interface
+provided by Determined. With the :ref:`model-definitions_trial-api`,
```
dzhu, line 13: As a general rule, let's stick to
```suggestion
provided by Determined. With the :ref:`Trial API <model-definitions_trial-api>`,
```
or something like
```suggestion
provided by Determined. See :ref:`model-definitions_trial-api` to learn more...
```
i.e., either treat the title of the page as a location or provide link text explicitly to be robust against page title changes.

Also, while we're here, the first mention should probably be the link instead, and maybe we don't need another link in the list below (or only link there; either way).

docs/topic-guides/model-definitions/index.txt
```diff
@@ -9,12 +9,10 @@ application framework (e.g., Keras, TensorFlow), in terms of loading
...
+developers expose the components of a model definition to Determined’s
```
dzhu, line 14: Use ASCII quote.

dzhu (approved):
Thanks for keeping an eye on things!

docs/topic-guides/model-definitions/index.txt
neilconway: Sounds reasonable. I used the first form. I kept both references linked just because it seems like more links is safer than less :)

docs/how-to/installation/aws.txt
neilconway: I couldn't really find a way to phrase this that I was happy with, so I just kept the current text for now.

PR #2073, opened 2021-03-12
[WIP] Harness collects system metrics

armandmcqueen:
## Description

This PR adds the ProfilerAgent. The ProfilerAgent will not be hooked up to training yet (i.e. it is never actually used) - there is enough code that it's worth reviewing the implementation of the ProfilerAgent before hooking it into the rest of the harness.

The ProfilerAgent has defined a public API for adding Timings, but the internals of that have not been implemented yet.

The code needs some cleaning up before it is production ready - adding type annotations, linting, etc. I also still need to incorporate the feedback around handling threads and conditional variables. I added new fields to the EnvContext, but haven't updated all places that use EnvContext.

There are many TODOs still in this code, some of which are quite major. 

## Checklist


examples/tutorials/mnist_pytorch/hotpatches/harness_hotpatch.py
```diff
@@ -0,0 +1,688 @@
...
+                    DISK_THRU_READ_METRIC: [],
+                    DISK_THRU_WRITE_METRIC: [],
+                    FREE_MEM_METRIC: [],
+                    SIMPLE_CPU_UTIL_METRIC: []
+                }
+                batch_start_time = time.time()
+
+
+
+    def update_current_batch(self, new_current_batch):
```
stoksc: where/how is this going to be called?

stoksc (commented):
the design basically looks like what I was expecting. just added some q's, you may have addressed them from under me though 😄

examples/tutorials/mnist_pytorch/hotpatches/harness_hotpatch.py
```diff
@@ -0,0 +1,688 @@
...
+        metadata = storage.StorageMetadata.from_json(checkpoint)
+        logging.info("Restoring trial from checkpoint {}".format(metadata.storage_id))
+
+        with storage_mgr.restore_path(metadata) as path:
+            yield pathlib.Path(path)
+
+
+def build_and_run_training_pipeline(env: det.EnvContext) -> None:
+    import gc
+    gc.disable()
```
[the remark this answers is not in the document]
armandmcqueen: It was for perf testing - making sure that gc wasn't contributing to the slowdown. It will be removed when this is ready for merging

examples/tutorials/mnist_pytorch/hotpatches/harness_hotpatch.py
```diff
@@ -0,0 +1,688 @@
...
+        self.is_active = True
+        self.current_batch = 1
+
+        self.dispatch_queue = queue.Queue()
+        self.sending_thread = SystemMetricsSendingThread(self.dispatch_queue)
+        self.sending_thread.start()
+
+        self.current_metric_batch = {
+            GPU_UTIL_METRIC: [],
+            # "GPU_MEM": [],
```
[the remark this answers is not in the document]
armandmcqueen: It measures the percent of time that the memory allocator was active, not the GBs of memory used

examples/tutorials/mnist_pytorch/hotpatches/harness_hotpatch.py
```diff
@@ -0,0 +1,688 @@
...
+                last_measurement_time = time.time()
+                batch_start_time = time.time()
+                network_throughput_collector.reset()
+                disk_collector.reset()
+
+
+
+
+            # Check if it is time to take a new measurement
+            if time.time() - last_measurement_time > self.MEASUREMENT_INTERVAL:
```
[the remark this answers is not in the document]
armandmcqueen: I added a sleep

examples/tutorials/mnist_pytorch/hotpatches/harness_hotpatch.py
armandmcqueen: It will be called by the controller. The code has been updated to do that for the PyTorch controller

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+        network_throughput_collector = NetThroughputCollector()
+        free_memory_collector = FreeMemoryCollector()
+        disk_collector = DiskReadWriteRateCollector()
+
+        while True:
+            if self.quitting:
+                break
+
+            if not self.is_active:
+                time.sleep(1)
```
rb-determined-ai, line 196: Please use conditional variables instead of sleeping

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+                last_measurement_time = time.time()
+
+
+            # Check if it is time to flush the batch and start a new batch
+            if time.time() - batch_start_time > self.FLUSH_INTERVAL:
+                self.enqueue_for_async_send(self.current_metrics.convert_to_post_format())
+                self.current_metrics.reset()
+                batch_start_time = time.time()
+
+            time.sleep(0.02)
```
rb-determined-ai, line 241: what is this delay for?

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+        timer = QuickTimer("SimpleCpuUtilCollector")
+        cpu_util = psutil.cpu_percent()
+        timestamp = datetime.datetime.utcnow()
+        timer.stop()
+        return Measurement(timestamp, batch_idx, cpu_util)
+
+
+
+class FreeMemoryCollector:
+    # We choose to report free memory instead of available memory because it is useful to
```
rb-determined-ai, line 328: I would like us not to show memory usage for cached files because I think that's almost always unhelpful.  We've had customers whose machines would look like they were teetering on OOM when in reality the kernel was just caching a lot of files, and cached files never cause OOM (see https://www.linuxatemyram.com/).

Personally I found psutil.memory_percent() to be pretty useful; not quite sure how that's derived from system values though.

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+        # I don't think it will ever happen (https://docs.python.org/3/library/queue.html#queue.Queue.put)
+        # self.log("Enqueuing metric batch", metric_batch)
+        self.dispatch_queue.put_nowait(metric_batch)
+
+    def __enter__(self) -> "SystemMetricsThread":
+        self.start()
+        return self
+
+    def __exit__(self, *arg: Any) -> None:
+        self.quitting = True
```
rb-determined-ai, line 257: while you did show me that it's not required in python to join threads that you have started, can we join it anyway?  I feel like this breaks the basic principle of "cleanup after yourself".

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+    def run(self) -> None:
+        while True:
+            if self.quitting:
+                break
+
+            try:
+                batch_to_send = self.inbound_queue.get_nowait()
+            except queue.Empty:
+                time.sleep(self.POLL_INTERVAL_SECS)
+                continue
```
rb-determined-ai, line 289: Just use queue.get(); and skip the spurious wake-ups.

harness/determined/metrics.py
```diff
@@ -0,0 +1,454 @@
...
+            api.post_trial_profiler_metrics(
+                self.master_url,
+                post_body["values"],
+                post_body["batches"],
+                post_body["timestamps"],
+                post_body["labels"],
+            )
+
+    def quit(self):
+        self.quitting = True
```
rb-determined-ai, line 309: ditto on joining.

harness/determined/metrics.py
armandmcqueen: Okay, I'll switch it to available. Cached file usage can be helpful for analyzing performance, but only if you can break it out as a single number.

examples/nlp/albert_squad_pytorch/startup-hook.sh
```diff
@@ -2,3 +2,4 @@
 pip install sentencepiece==0.1.91
 pip install transformers==3.1.0
 pip install -e git+git://github.com/LiyuanLucasLiu/RAdam.git@baf4f65445c00d686d4098841b3ca1f62a886326#egg=radam
+pip install pynvml
```
armandmcqueen, line 5: This is a reminder that we now rely on this library and need to add this to the environments.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        write_throughput = Measurement(timestamp, batch_idx, write_throughput_bytes_per_second)
+        iops = Measurement(timestamp, batch_idx, read_throughput_count_per_second + write_throughput_count_per_second)
+
+        return read_throughput, write_throughput, iops
+
+
+
+class GpuUtilCollector:
+    def __init__(self):
+        pynvml.nvmlInit()
```
stoksc, line 498: I'm adding this to environments, we can do something like
```
try:
    import pynvml
    pynvml.nvmlInit()
except ModuleNotFoundError:
    pass
except pynvml.NVMLError_LibraryNotFound:
    pass
```
here to check whether we should enable this collector or not.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+            if self.quitting:
+                break
+
+            try:
+                batch_to_send = self.inbound_queue.get_nowait()
+            except queue.Empty:
+                time.sleep(ProfilerSenderThread.POLL_INTERVAL_SECS)
+                continue
+
+            self.send_batch(batch_to_send)
```
stoksc, line 361: It seems like something like I listed below would work just fine for this.
```suggestion
            self.send_batch(self.inbound_queue.get(timeout=SomethingReasonableSoWeDontBlockForever))
```

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+                last_measurement_time = time.time()
+
+
+            # Check if it is time to flush the batch and start a new batch
+            if time.time() - batch_start_time > self.FLUSH_INTERVAL:
+                self.enqueue_for_async_send(self.current_batch.convert_to_post_format())
+                self.current_batch.clear()
+                batch_start_time = time.time()
+
+            time.sleep(0.02)
```
stoksc, line 330: I think it's probably just fine to sleep the (measurement_interval - time_taken) at the end of every run, that may even be more accurate since this could be off up to 0.02 + flush time if the condition on 302 is false and 325 is true on a pass through.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+                        post_body["values"],
+                        post_body["batches"],
+                        post_body["timestamps"],
+                        post_body["labels"],
+                    )
+                    break
+                except Exception as e:
+                    # TODO: We could handle specific API errors differently as some are non-recoverable
+                    # TODO: Log info about error
+                    time.sleep(i**2)  # exponential backoff
```
stoksc, line 384: We might end up getting pretty backed up here (this could wait 31 seconds per batch total while they're piling in if the master blips). It's probably ok to try one time in a constant interval and bail if we can't. It's just a single batch.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+
+        sent_throughput_gigabits_per_second = sent_throughput_bytes_per_second * 8 * GIGA
+        recv_throughput_gigabits_per_second = recv_throughput_bytes_per_second * 8 * GIGA
+
+        timestamp = datetime.datetime.fromtimestamp(end_time)
+        return Measurement(timestamp, batch_idx, sent_throughput_gigabits_per_second), \
+               Measurement(timestamp, batch_idx, recv_throughput_gigabits_per_second)
+
+
+class DiskReadWriteRateCollector:
```
stoksc, line 449: Collector per metric got pretty verbose it seems. Didn't you look at what yoav did forever ago here https://github.com/determined-ai/determined/blob/master/harness/determined/layers/_harness_profiler.py#L113-L133? Could save a bit of space but not really a show stopper.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        self.batch[metric_type][gpu_uuid].append(measurement)
+
+    def convert_to_timestamp_str(self, timestamp: datetime.datetime):
+        assert isinstance(timestamp, datetime.datetime), \
+            f"Input to conversion function must be a datetime object. Instead got {type(timestamp)}"
+        return timestamp.isoformat() + "Z"
+
+    def convert_to_post_format(self):
+        post_formatted = []
+        for metric_type in self.batch.keys():
```
stoksc, line 184: This seems a little involved, I think it's a symptom of the `add_nongpu_measurement`/`add_gpu_measurement` dichotomy. It may make more sense/save a lot of effort to have the `Measurement` type have a `gpu_uuid` that is filled in some cases and not in others and then just let the rest of the code behave the same.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        self.end_after_batch = end_after_batch
+
+        # Track duration to stop collecting after 5 minutes. start_time also serves as
+        # the indicator that collection has begun.
+        self.start_time = None
+        # TODO: Currently we only check for shutdown in update_batch_idx(). This is a problem when there
+        #       is a long period of time between batch_idx being updated (such as during validation)
+        #       because it won't shut down until validation completes. We probably need to create a
+        #       new thread to enforce the timeout correctly.
+        self.max_collection_seconds = 300
```
stoksc, line 63: Could we just enforce this with a check in the metrics loop and stop shipping after this interval is exceeded (similar to the flush interval check, but instead of flushing just exit).

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+        self.sender_thread.start()
+
+        # Launch the system metric collecting thread, but not the actual collection of metrics
+        self.sys_metric_collector_thread = SysMetricCollectorThread(trial_id, agent_id, self.send_queue)
+        self.sys_metric_collector_thread.start()
+
+        # TODO: Add data structure to batch timings and then send to SenderThread
+        #       Does this need to be its own thread to flush correctly?
+        # self.timings_batcher = TimingsBatcher()
```
stoksc, line 77: as far as I see, this could nearly be the same code as the `SysMetricBatcher`, if it either it was smart enough to know which names go to which type or `add_nongpu_measurement` took a type too. All the rest of the code looks like it would work fine.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+
+
+SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
+
+
+
+
+class ProfilerAgent:
+    """
+    Agent that collects metrics and sends them to the master. It has:
```
stoksc, line 17: can we drop a comment about all the threads there are and which starts/joins which just for readers?

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+    profiler_agent = ProfilerAgent(self, trial_id, agent_id, master_url, start_on_batch, end_after_batch)
+
+    for batch_idx, batch in enumerate(batches):
+        profiler_agent.update_batch_idx(batch_idx)
+
+        # NOTE: Timing API has not been fully developed yet
+        forward_pass_timing = Timing("forward_pass")
+        forward_pass_timing.start()
+        # Do forward pass
+        forward_pass_timing.end()
```
stoksc, line 44: could be something nice like
```
with profiler_agent.timing("forward_pass"):
    # forward pass
```

stoksc (commented):
had some notes on the design overall, just a first pass to get it rolling.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        disk_collector = DiskReadWriteRateCollector()
+
+        while True:
+            if self.quitting:
+                # We drop any partial batches
+                break
+
+            if not self.is_active:
+                time.sleep(1)
+                continue
```
rb-determined-ai, line 291: This is going to result in very low granularity of control on startup times.  If you wanted profile, for example, just one batch, this polling strategy won't let you do that at all; you'll likely be done with the batch before the collection thread even responds to `_begin_collection()`.

A more responsive design would be something like:

```python
class ActivateMessage:
    pass


class QuitMessage:
    pass


class ProfilerAgent:
    # ...
    def _begin_collection(self):
        self.sys_metric_collector_thread.msg_queue.put(ActivateMessage())

    def _end_collection(self):
        self.sys_metric_collector_thread.msg_queue.put(QuitMessage())
        # ...
        

class SysMetricCollectorThread(threading.Thread):
    # ...

    def run(self):
        # ...
        msg = self.msg_queue.get()
        if isinstance(msg, QuitMessage):
             # None means "quit"
             return
        
        # avoid drift in period by explicitly calculating timing
        # rather than relying on sleep(period) 
        next_collection = time.time()

        while True:
            now = time.time()
            if now < next_collection:
                # check for quit message while we wait
                sleep_time = next_collection - now
                try:
                    msg = self.msg_queue.get(timeout=sleep_time)
                    if isinstance(msg, QuitMessage):
                        return
                except queue.Empty:
                   pass

            # avoid drift
            next_collection += period

            # ... do collection and batching and such
```

The above pattern will have maximal responsiveness to the control thread.  It uses timeouts on the `Queue.get()` to implement the periodicity, and the only time it's not waiting on the `.get()` is during the time we are actually collecting metrics.  It also has the benefit that it does not have any shared memory at all; you could use multiprocessing Processes just as easily as threading Threads.

It doesn't full solve the `accurately-profile-one-batch` solution though; that would require a synchronization message in the reverse direction.  But it's a good start.

harness/determined/profiler.py
```diff
@@ -0,0 +1,524 @@
...
+        self.quitting = False
+        super().__init__()
+
+    def run(self) -> None:
+        while True:
+            if self.quitting:
+                break
+
+            try:
+                batch_to_send = self.inbound_queue.get_nowait()
```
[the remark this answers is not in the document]
armandmcqueen: Changed

harness/determined/profiler.py
armandmcqueen: Changed

harness/determined/profiler.py
armandmcqueen: Maybe. It seems a little convoluted to have the subsidiary thread doubling as a timer and we would need to have a way for the metrics thread to then propagate that timer back up the to the enclosing object. I'm thinking it will be better to spawn a new timer thread separately.

armandmcqueen:
Moved to #2198

PR #2074, opened 2021-03-13
chore: update to new images including TensorFlow, PyTorch, Python and CUDA upgrades

mackrorysd:
## Description

- Update the images, including updates to CUDA drivers and toolkits, Python, PyTorch and TensorFlow
- A few tweaks for updating GovCloud images
- Updating documentation to refer to the older, still support models
- Update the framework versions used to drive tests, and some corresponding test tweaks
- Update type annotations where needed due to PyTorch upgrade

## Test Plan

Testing of new Docker images was described here: https://github.com/determined-ai/environments/pull/75

I also tested that you can run PyTorch and Tensorflow models using the old images with the modified harness.

I plan to add a test similar to our CUDA 11 to regularly exercise the previous default images until we remove support for them.


e2e_tests/tests/requirements.txt
```diff
@@ -3,12 +3,9 @@ appdirs
 pytest>=6.0.1
 pytest-timeout
 pexpect
-torch==1.4
-torchvision==0.5.0
-tensorflow==1.15.4
-# TF 1.15.4 has requirement numpy<1.19.0,>=1.16.0
```
rb-determined-ai, line 9: thank you for paying enough attention to delete this : )

harness/Makefile
```diff
@@ -30,15 +30,15 @@ check:
 
 .PHONY: test
 test:
-	pytest -v --runslow --durations=0 tests
+	pytest -v -s --runslow --durations=0 tests
```
rb-determined-ai, line 33: why is `-s` needed now?

harness/setup.py
```diff
@@ -14,7 +14,7 @@
     python_requires=">=3.6",
     package_data={"determined": ["py.typed"]},
     install_requires=[
-        "cloudpickle==0.5.3",
+        "cloudpickle",
```
rb-determined-ai, line 17: is cloudpickle forward-compatible?  like can you safely load a checkpoint saved with old cloudpickle using new cloudpickle?

harness/setup.py
```diff
@@ -27,10 +27,14 @@
...
-        "pytorch-14-cuda100": ["torch==1.4.0+cu100", "torchvision==0.5.0+cu100"],
-        "pytorch-14-cpu": ["torch==1.4.0", "torchvision==0.5.0"],
+        "tf-115-cuda102": ["tensorflow-gpu==1.15.5"],
+        "tf-115-cpu": ["tensorflow==1.15.5"],
+        "tf-240-cuda102": ["tensorflow-gpu==2.4.1"],
+        "tf-240-cpu": ["tensorflow==2.4.1"],
+        "tf-241-cuda110": ["tensorflow-gpu==2.4.1"],
+        "pytorch-18-cuda102": ["torch==1.8.0", "torchvision==0.9.0"],
+        "pytorch-18-cuda110": ["torch==1.8.0", "torchvision==0.9.0"],
+        "pytorch-18-cpu": ["torch==1.7.0", "torchvision==0.8.2"],
```
rb-determined-ai, line 37: 1.7 and 0.8.2?

harness/Makefile
mackrorysd: Last time we considered making this change test_to_device_warnings had problems and the logger didn't work properly unless we added -s. This time, that appears to be insufficient so I'm actually trying a couple of different things. Although you said the test wasn't that valuable, so next step is to just remove it,.

harness/setup.py
mackrorysd: Oh that's wrong. I'll fix that. I might have been trying to work around a bug that I thought was due to the fact that you can't embed CUDA versions in the package name for 1.8.0 (they're going to fix that in 1.8.1). But it wasn't.

.circleci/config.yml
```diff
@@ -797,7 +799,7 @@ jobs:
           determined-common: true
           determined-cli: true
           determined: true
-          extras-requires: "tensorflow==1.14 torch==1.4"
+          extras-requires: "tensorflow==2.4 torch==1.8"
```
hamidzr, line 802: wdy think about moving torch1.8 into the requirements.txt file for this module?

harness/Makefile
mackrorysd: I've removed this change, because even with it PyTest was still hanging on that test. I actually just changed the test to not use a subprocess. There was a comment about how we didn't want to mess up the existing processes logging configuration, but I think simply adding a queue logger is negligible. Happy to revisit...

harness/setup.py
mackrorysd: That's a very good question. As a sanity check, I did create a TF / Keras checkpoint in the old image and ensured I could continue training in the new one. I'll see if I can get a definitive answer though.

For context, the pressure to move off of 0.5.3 is that it will not work with Python 3.8.

rb-determined-ai: ok, but I'm pretty sure only pytorch used cloudpickle anyway, so I'm a bit more interested in that case.  Like basically, can I continue training from my 0.14.3 checkpoint in a 0.14.4 cluster?  That would be a pretty big breakage if it was not true.

webui/react/src/fixtures/responses/experiment-details/non-scalar-metrics-4078.json
```diff
@@ -31,8 +31,8 @@
     "description": "Fork of Fork of mnist_tp_to_estimator_const",
```
hamidzr, line 31: nit: we probably don't need to update anything under webui/react/src/fixtures.

mackrorysd: Yeah this is just part of the bumpenv.sh process. Unless it's worth adding this to an ignore-list in bumpenvs.sh, it's just automated to update this every time.

.circleci/config.yml
mackrorysd: Discussed offline: let's handle that separately and just keep the delta here as small as possible.

harness/setup.py
mackrorysd: I've tested a bunch of PyTorch models as well - CIFAR and the NLP models. No problems. You can't go backwards, of course, because of the PyTorch archive version issues that we've sometimes hit in tests when upgrading the taskenv but not the test driver.

There unfortunately doesn't seem to be much of a cloudpickle community - I could find a mailing list or anything, so I filed an issue a few days ago to clarify the *intended* support but haven't heard anything.

rb-determined-ai (approved):
you have my blessing

PR #2077, opened 2021-03-15
fix: only upload model code as model code [DET-3865]

rb-determined-ai:
Before we always copied the working directory as model code.  Users
often saw apparent hangs waiting for checkpoint upload, which in fact
were caused by things like us silently trying to upload their entire
dataset to S3.

It might not be a good idea to upload auto-discovered model definitions
to checkpoint storage.  But rather than fundamentally alter the
assumptions of our checkpointing system, this PR just adds a clean copy
of the model which was submitted to the master into the training
container, and during managed training, that clean copy gets uploaded
rather than the /run/determined/workdir.

### DIFF
--- a/docs/release-notes/2077-model-upload.txt
+++ b/docs/release-notes/2077-model-upload.txt
@@ -1,0 +1,9 @@
+:orphan:
+
+**Fixes**
+
+-  Model code uploaded to checkpoints will now match the model code
+   provided during experiment creation. Previously, it may have included
+   additional files that had been bind-mounted in with a
+   ``container_path`` that was either relative or was a subdirectory of
+   ``/run/determined/workdir``.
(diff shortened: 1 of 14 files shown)


shiyuann (approved):
Looks good to me! I didn’t test. But it should be fine if it passes the CI.

rb-determined-ai:
Oops, it turns out `managed_training` is not synonymous with "being on-cluster", so I just added a new flag to EnvContext.

I also made the three "extra" flags to EnvContext into mandatory arguments like all the others.

Please re-review.

rb-determined-ai:
Talked to Shiyuan, he thinks it's fine.

shiyuann (approved):
looks good!

PR #2083, opened 2021-03-16
feat: delete Tensorboards with delete API [DET-5143]

stoksc:
## Description
This change adds code to GC Tensorboards when DELETE experiment is run. It is piggybacked as a new flag onto the checkpoint GC script.

## Test Plan
- [x] test with shared fs
- [x] test with s3
- [ ] test with hdfs
- [x] test with gcs bucket

## Commentary (optional)
Looking for a bit of feedback on whether the interface is OK.. I usually don't touch this harness stuff.

### DIFF
--- a/harness/determined/exec/gc_checkpoints.py
+++ b/harness/determined/exec/gc_checkpoints.py
@@ -37,11 +37,11 @@ def delete_tensorboards(manager: tensorboard.TensorboardManager, dry_run: bool =
     Delete all Tensorboards associated with a single experiment.
     """
     if dry_run:
-        logging.info("Dry run: deleting Tensorboards for {}".format(manager.experiment_sync_path))
+        logging.info("Dry run: deleting Tensorboards for {}".format(manager.sync_path))
         return
 
     manager.delete()
-    logging.info("Finished deleting Tensorboards for {}".format(manager.experiment_sync_path))
+    logging.info("Finished deleting Tensorboards for {}".format(manager.sync_path))
 
 
 def json_file_arg(val: str) -> Any:
@@ -105,15 +105,10 @@ def main(argv: List[str]) -> None:
     delete_checkpoints(manager, args.delete["checkpoints"], dry_run=args.dry_run)
 
     if args.delete_tensorboards:
-        # HACK: Trial ID "1" is passed as a dummy since we never call sync and it is never used.
-        # I can get around this by:
-        #    1. Separating deletes from the TensorboardManager.
-        #    2. Making TensorboardManagers be "unsyncable" or not attached to a particular trial and
-        #       only calls to delete succeed.
         tb_manager = tensorboard.build(
             os.environ["DET_CLUSTER_ID"],
             args.experiment_id,
-            "1",
+            None,
             storage_config,
             container_path=constants.SHARED_FS_CONTAINER_PATH,
         )
--- a/harness/determined/exec/gc_checkpoints.py
+++ b/harness/determined/exec/gc_checkpoints.py
@@ -79,7 +79,7 @@ def main(argv: List[str]) -> None:
     parser.add_argument(
         "--delete-tensorboards",
         action="store_true",
-        default=False,
+        default=os.getenv("DET_DELETE_TENSORBOARDS", False),
         help="Delete Tensorboards from storage",
     )
     parser.add_argument(
--- a/harness/determined/exec/gc_checkpoints.py
+++ b/harness/determined/exec/gc_checkpoints.py
@@ -9,6 +9,7 @@ import sys
 from typing import Any, Dict, List
 
 import determined as det
+from determined import tensorboard
 from determined_common import constants, storage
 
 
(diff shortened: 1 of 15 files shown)


stoksc:
blocked on #2056 but if you want to 👁️ the last commit @rb-determined-ai, feel free.

shiyuann:
@stoksc is this PR for deleting experiments or just TensorBoards?

stoksc:
@shiyuann yeah it is, but it's blocked on another.

harness/determined/tensorboard/s3.py
```diff
@@ -40,3 +46,13 @@ def sync(self) -> None:
 
             self.client.upload_file(str(path), self.bucket, key_name)
             self._synced_event_sizes[path] = path.stat().st_size
+
+    def delete(self) -> None:
+        self.resource.Bucket(self.bucket).objects.filter(Prefix=str(self.sync_path)).delete()
+
+        # If we're attached to using the low level APIs:
```
stoksc, line 53: preference @rb-determined-ai ?

rb-determined-ai: The StorageManager code is some of the oldest code we have.  I don't think we have any attachment to the low-level APIs, that's just how it's always been.

rb-determined-ai: (so I prefer the easy way)

harness/determined/exec/gc_checkpoints.py
```diff
@@ -46,6 +57,7 @@ def main(argv: List[str]) -> None:
         action="version",
         version="Determined checkpoint GC, version {}".format(det.__version__),
     )
+    parser.add_argument("--experiment-id", help="The experiment ID to run the GC job for")
```
rb-determined-ai, line 60: It occurs to me that some day when we want to version the harness/master protocol, we'll have to either pack the gc script in at runtime or we will have to edit it several versions before we start using a new feature in it.

Just a thought, not related to this PR in particular.

harness/determined/exec/gc_checkpoints.py
```diff
@@ -64,6 +76,12 @@ def main(argv: List[str]) -> None:
         default=os.getenv("DET_DELETE", []),
         help="Checkpoints to delete (JSON-formatted file)",
     )
+    parser.add_argument(
+        "--delete-tensorboards",
+        action="store_true",
+        default=False,
```
rb-determined-ai, line 82: maybe accept the environment variable to be consistent?  Or delete all the env variables that we haven't used in like forever...

Either way.

harness/determined/tensorboard/base.py
```diff
@@ -14,7 +14,11 @@ class TensorboardManager:
     implements the sync method.
     """
 
-    def __init__(self, base_path: pathlib.Path, sync_path: pathlib.Path):
+    def __init__(
+        self,
+        base_path: pathlib.Path,
+        sync_path: pathlib.Path,
+    ):
```
rb-determined-ai, line 21: wait... this file isn't being linted, or this `__init__` without a `-> None` would throw an error... right?

harness/determined/tensorboard/base.py
```diff
@@ -53,3 +57,9 @@ def sync(self) -> None:
         Save the object to the backing persistent storage.
         """
         pass
+
+    def delete(self) -> None:
+        """
+        Delete all objects from the backing persistent storage.
+        """
+        pass
```
rb-determined-ai, line 65: If this method isn't optional (seems like it isn't), we should either raise a NotImplementedError or (better) we should do `class TensorboardManager(metaclass=abc.ABCmeta` and decorate `delete` with `@abc.abstractmethod`

harness/determined/tensorboard/s3.py
stoksc: I wasn't sure if it was some weird minio compat thing I didn't know about, but minio looked like it should work with both APIs from what I saw (I'm unsure if it's actually a different API or it's just a nice client side wrapper around the low levels). I'll just use the nice one.

rb-determined-ai (commented):
Please look into the linting thing, but otherwise this looks great!

harness/determined/tensorboard/s3.py
rb-determined-ai: nah, our s3 code was written way before we ran it against minio

harness/determined/tensorboard/base.py
stoksc: i'll peek at this.

harness/determined/tensorboard/base.py
stoksc: I'll make it an abstract class

harness/determined/tensorboard/base.py
stoksc: OK, the file is definitely being linted, e.g. if I remove the ` -> None` from `sync`, it flips out, but if I remove it from any `__init__` call it doesn't seem to mind 🙃 ?

harness/determined/exec/gc_checkpoints.py
stoksc: yeah.. that's a good point. interesting.

rb-determined-ai (approved):

PR #2085, opened 2021-03-17
feat: add aggregated resource allocation endpoint and job [DET-5044]

dzhu:
## Description

We now support returning cluster allocation metrics aggregated by
day. It seems difficult, after some experimentation, to do aggregation
on the fly sufficiently quickly in the face of very large histories, so
the aggregation is done at the end of each day and the endpoint then
does a fast query over those results. There's a new aggregation actor
that stays alive in the background and updates the aggregated table at
the end of each day or on master startup.

## Test Plan

- [x] generate test database with a bunch of steps and check that results look right
- [x] test with times spanning the DST transition, which caught a bug due to a missing `.UTC()`


master/internal/resource_allocation.go
```diff
@@ -0,0 +1,45 @@
...
+type allocationAggregator struct {
+	db *db.PgDB
+}
+
+func (a *allocationAggregator) schedule(ctx *actor.Context) {
+	now := time.Now().UTC()
+	then := time.Date(now.Year(), now.Month(), now.Day(), 0, 1, 0, 0, time.UTC)
+	if then.Before(now) {
+		then = then.AddDate(0, 0, 1)
+	}
```
stoksc, line 22: maybe pull this out into something like `nearestDayStart` so it's immediately clear to the reader the intent?

master/internal/db/postgres.go
```diff
@@ -1919,3 +1919,46 @@ WHERE t.id = $1;
 `, id)
 	return errors.Wrapf(err, "error updating best validation for trial %d", id)
 }
+
+// RunResourceAllocationAggregation updates the aggregated resource allocation table.
+func (db *PgDB) RunResourceAllocationAggregation() error {
```
stoksc, line 1924:
```suggestion
func (db *PgDB) UpdateResourceAllocationAggregations() error {
```
Maybe to make it more clear this is expensive and may touch multiple rows?

proto/src/determined/master/v1/master.proto
```diff
@@ -26,3 +26,17 @@ message ResourceAllocationRawEntry {
   // The number of slots used by the allocation.
   int32 slots = 8;
 }
+
+// One instance of slots in the cluster being allocated to a task during a day
+// (aggregated).
+message ResourceAllocationAggregatedEntry {
+  // The type of aggregation performed for this entry ("total").
+  string aggregation_type = 1;
```
stoksc, line 34: should we make aggregation type an enum?

master/internal/db/postgres.go
```diff
@@ -1919,3 +1919,46 @@ WHERE t.id = $1;
...
+		if err != nil {
+			return errors.Wrap(err, "failed to find first step")
+		}
+
+		periodStart = firstDatePtr.UTC()
+	} else {
+		periodStart = lastDatePtr.UTC().AddDate(0, 0, 1)
+	}
+	targetDate := time.Now().UTC().AddDate(0, 0, -1)
+	for ; periodStart.Before(targetDate); periodStart = periodStart.AddDate(0, 0, 1) {
```
stoksc, line 1948: with this would we end up not aggregating yesterday until tomorrow always? like they'd always have 2 days delay? i'm sorry if i'm reading this wrong, just looks like period start would need to be before yesterday?

stoksc (approved):
looks great.

master/internal/core.go
```diff
@@ -213,6 +213,62 @@ func (m *Master) getResourceAllocationRaw(c echo.Context) error {
...
+		"get_aggregated_allocation", &resp.ResourceEntries, start.UTC(), end.UTC(),
+	); err != nil {
+		return errors.Wrap(err, "error fetching allocation data")
+	}
+
+	c.Response().Header().Set("Content-Type", "text/csv")
+
+	csvWriter := csv.NewWriter(c.Response())
+	formatTimestamp := func(ts *timestamppb.Timestamp) string {
+		t := time.Unix(ts.Seconds, int64(ts.Nanos)).UTC()
```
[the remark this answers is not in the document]
dzhu: Ah yes, turns out to be `ptypes.Timestamp`. Why isn't that a method on `Timestamp`? It's a mystery!

proto/src/determined/master/v1/master.proto
stoksc: Yeah I sort of agree, but I also think it _is_ an enum.. it really is grpc gateway that translates them in such an overly-verbose way. and the verbosity is the only really downside.

master/internal/core.go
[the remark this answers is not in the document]
dzhu: Oh, they actually did add an `AsTime` method to the generated types very recently. I'll go ahead and upgrade the protobuf package to get that.

proto/src/determined/master/v1/master.proto
dzhu: Okay, I made it an enum.

PR #2092, opened 2021-03-18
feat: provide user hint on aws, gcp auth in det-deploy [DET-4846]

ioga:
## Description

- credentials check for aws / cloudformation
- credentials check for gcp / terraform
- have local cluster creation fail on docker-compose failures

## Checklist

### DIFF
--- a/deploy/determined_deploy/gcp/gcp.py
+++ b/deploy/determined_deploy/gcp/gcp.py
@@ -7,22 +7,30 @@ import time
 from typing import Any, Dict, List, Optional
 
 import googleapiclient.discovery
+from google.auth.exceptions import DefaultCredentialsError
+from termcolor import colored
 
 TF_VARS_FILE = "terraform.tfvars.json"
 
 
 def deploy(configs: Dict, env: Dict, variables_to_exclude: List) -> None:
     terraform_init(configs, env)
+    validate_gcp_credentials(configs)
     terraform_apply(configs, env, variables_to_exclude)
 
 
 def dry_run(configs: Dict, env: Dict, variables_to_exclude: List) -> None:
     terraform_init(configs, env)
+    validate_gcp_credentials(configs)
     terraform_plan(configs, env, variables_to_exclude)
 
 
 def terraform_dir(configs: Dict) -> str:
     return os.path.join(configs["local_state_path"], "terraform")
+
+
+def get_terraform_vars_file_path(configs: Dict) -> str:
+    return os.path.join(configs["local_state_path"], TF_VARS_FILE)
 
 
 def terraform_read_variables(vars_file_path: str) -> Dict:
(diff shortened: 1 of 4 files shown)


deploy/determined_deploy/gcp/gcp.py
```diff
@@ -261,3 +267,30 @@ def delete(configs: Dict, env: Dict) -> None:
...
     subprocess.check_call(command, env=env, shell=True, stdout=sys.stdout)
+
+
+def validate_gcp_credentials(configs: Dict) -> None:
+    vars_file_path = get_terraform_vars_file_path(configs)
+    tf_vars = terraform_read_variables(vars_file_path)
+    set_gcp_credentials_env(tf_vars)
+
+    try:
+        googleapiclient.discovery.build("compute", "v1")
```
stoksc, line 278: if we wanted to get even more fancy one day, it looks like we can test the permissions associated with an IAM account to make sure they match what we need https://cloud.google.com/iam/docs/testing-permissions#iam-testing-permissions-python

deploy/determined_deploy/local/cluster_utils.py
```diff
@@ -90,7 +90,7 @@ def docker_compose(
         for extra_file in extra_files:
             base_command += ["-f", extra_file]
     args = base_command + args
-    subprocess.run(args, env=process_env)
+    subprocess.check_call(args, env=process_env)
```
stoksc, line 93: doesn't check call just raise a CalledProcessError with the return code? do we need the stdout/stderr from the call for the issue to be clear?

stoksc (commented):
looks great, just a few q's

deploy/determined_deploy/gcp/gcp.py
```diff
@@ -7,24 +7,32 @@
...
 import googleapiclient.discovery
+from google.auth.exceptions import DefaultCredentialsError
+from termcolor import colored
 
 TF_VARS_FILE = "terraform.tfvars.json"
 
 
 def deploy(configs: Dict, env: Dict, variables_to_exclude: List) -> None:
     terraform_init(configs, env)
+    validate_gcp_credentials(configs)
```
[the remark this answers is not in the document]
ioga: done.

deploy/determined_deploy/gcp/gcp.py
ioga: yeah - I suppose it'll need to include both this (what we do from python code) and whatever resources terraform requires.

deploy/determined_deploy/local/cluster_utils.py
ioga: both `run` and `check_call` will print stdout/stderr, so this won't change what's printed, you'd still see the underlying docker-compose logs.

so what I've encountered is that docker-compose would fail e.g. when docker daemon is down, or it can't find an image. with `run`, code just continues to roll and fails later on in an odd place. we should fail as soon as docker-compose fails instead, and that's what this `check_call` does.

stoksc (approved):
lgtm

PR #2100, opened 2021-03-18
chore: testing gang-scheduling [DET-5134]

mackrorysd:
## Description

Adding automated tests of the coscheduler / gang scheduling feature.

## Test Plan

This *is* the test plan! The test fails (times out) about 50% of the time when the coscheduler isn't enabled, but reliably passes when it is. This test also has the benefit of exercising the coscheduler more, since our other test clusters usually have auto-scaling, with which the coscheduler isn't supported.

I'm concerned this may be a little flaky because the coscheduler is known to sometimes have timeouts when an experiment requires the entire cluster, which is the case with this test. Since a multi-node job must consume 100% of the GPUs on a certain number of nodes, one alternative would be for the tests to use a 3-node-cluster and 2 (or more) 2-node tests. But let's see if that necessary before making that change.

Ran this on an upstream branch: https://app.circleci.com/pipelines/github/determined-ai/determined/10669/workflows/368ab9ad-e7e8-41ac-bfa1-ebfc21da41f0/jobs/302416

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -565,7 +565,8 @@ commands:
             detVersion=<<parameters.det-version>>,\
             maxSlotsPerPod=<<parameters.gpus-per-machine>>,\
             checkpointStorage.type=gcs,\
-            checkpointStorage.bucket=det-ci
+            checkpointStorage.bucket=det-ci,\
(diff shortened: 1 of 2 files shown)


eecsliu (approved):
LGTM!

PR #2104, opened 2021-03-19
fix: master and agent release target

shiyuann:
## Checklist


justin-determined-ai (approved):
Remove the `config.yml` changes as we discussed and I think this looks good!

justin-determined-ai:
The diff I saw has both `agent/` and `master/` changes to `go.sum`, but your diff only (now) has changes to `agent/`, is that intended?

PR #2107, opened 2021-03-19
feat: expose resource allocation endpoints in CLI [DET-5045]

dzhu:
## Description

Exactly what it says on the tin.

## Test Plan

- [x] run the CLI and check that the output accurately reflects the API response and that reasonable error messages show up on bad inputs


stoksc (approved):
lgtm

PR #2111, opened 2021-03-22
chore: add pytorch-lightning to docs requirements

neilconway:
PTL is now required to build the docs.

## Checklist

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,9 +1,12 @@
 # pin docutils at 0.15.2 to avoid transitive dependency conflict with botocore (requires < 0.16)
 docutils==0.15.2
-sphinx==2.4.4
-sphinx-argparse>=0.2.5
-git+https://github.com/determined-ai/determined_sphinx_theme.git@9edfa7c6ce6926def9fc69b8ddd7666f3419a907
-sphinx-sitemap>=2.2.0
+
 pillow
+pytorch-lightning==1.2.0
 rstfmt==0.0.7
+sphinx-argparse>=0.2.5
 sphinx-copybutton
+sphinx-sitemap>=2.2.0
+sphinx==2.4.4
+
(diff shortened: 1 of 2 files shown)


mackrorysd:
This may be dependent on https://github.com/determined-ai/determined/pull/2109 to fully fix things: apparently this can trigger a circular dependency because of Tensorboard. That said, I'm not sure what workflow actually breaks and `make all` is currently broken, so I'll +1 this to unblock and ping Hamid.

mackrorysd (approved):

docs/requirements.txt
```diff
@@ -1,9 +1,12 @@
 # pin docutils at 0.15.2 to avoid transitive dependency conflict with botocore (requires < 0.16)
 docutils==0.15.2
-sphinx==2.4.4
-sphinx-argparse>=0.2.5
-git+https://github.com/determined-ai/determined_sphinx_theme.git@9edfa7c6ce6926def9fc69b8ddd7666f3419a907
-sphinx-sitemap>=2.2.0
+
 pillow
+pytorch-lightning==1.2.4
```
hamidzr, line 5: our lightning support is based off of `1.2.0` can we have the same version here? should help with the API when developing harness.

ioga:
works for me on macos 👍

hamidzr (approved):
LGTM

> This may be dependent on #2109 to fully fix things: apparently this can trigger a circular dependency because of Tensorboard. That said, I'm not sure what workflow actually breaks and make all is currently broken, so I'll +1 this to unblock and ping Hamid.

I was suspecting this run into the issue @mackrorysd  was describing in a local build but this + #2074 's move to tf2 look like to be all we need.

Thanks for fixing this.

docs/requirements.txt
neilconway: 1.2.1 through 1.2.4 are just "standard weekly patch releases" and mostly include bug fixes (https://github.com/PyTorchLightning/pytorch-lightning/releases). Happy to use `1.2.0` for now but IMO hardcoding a single specific minor version of PTL to support is probably not going to be tenable.

PR #2112, opened 2021-03-22
refactor: cleanup react hook dependencies [DET-5158, DET-5159, DET-5160]

hkang1:
## Description

Refactor useEffect dependencies.
- Rewrite components where we can avoid having to skip the linting rule of `react-hooks/exhaustive-deps`. Hooks are the exception.

Refactor `usePolling`.
- Expose `startPolling`, `stopPolling` and `isPolling` state.
- Use references appropriately to prevent re-rendering where `usePolling` is used.
- Update all instances of `usePolling` to follow new pattern.

Refactor `CreateExperimentModal`
- Move api call out of `CreateExperimentModal` component to make it a "dumb" component.
- Move user-friend config form into `CreateExperimentModal` component from `TrialDetails` to share the logic with experiment Fork.
- Fix the experiment detail polling issue where it will either over-poll or under-poll upon an experiment fork.
- Fix re-routing flow upon a successful experiment fork or continue trial.

These three areas all stem from cleaning up `usePolling` and exhaustive `useEffect` hook dependencies.

## Test Plan

- [ ] Check that polling works on navigation (cluster agent info)
- [ ] Check that polling works on pages (dashboard, experiment list, task list, experiment detail, trial detail, master logs)
- [ ] Check that forking experiment flow works.
- [ ] Check that continue trial flow works.

## Checklist


webui/react/src/hooks/usePolling.ts
```diff
@@ -21,42 +32,44 @@ const usePolling = (pollingFn: PollingFn, { delay }: PollingOptions = {}): StopF
...
     savedPollingFn.current = pollingFn;
-    if (!active.current) startPolling();
-  }, [ pollingFn, startPolling ]);
+  }, [ pollingFn ]);
 
   // Start polling when mounted and stop polling when umounted.
   useEffect(() => {
     startPolling();
     return () => stopPolling();
+    /*
```
apizzini, line 65: Maybe I'm wrong but that should still happen by adding:
- `stopPolling`: depending only on `clearTimer` which never changes
- `startPolling`: depending on the params passed to `usePolling` (never changing) and `clearTimer` which never changes?

apizzini (approved):
great cleanup!

webui/react/src/hooks/usePolling.ts
hkang1: yep that's very true. I did make a minor tweak to make `startPolling` only reliant on references. The polling option is not stored in a `useRef` reference so it should not cause the `startPolling` useCallback to be redefined which in turn should not cause the `useEffect` to re-run also.

PR #2114, opened 2021-03-23
chore: remove cluster v1 page [DET-5163]

hamidzr:
## Description

We've had the new cluster page out for a few releases now, it's time to say goodbye to the old one.

## Checklist


webui/react/src/pages/ClusterV1.tsx
```diff
@@ -1,58 +0,0 @@
-import React, { useEffect, useMemo, useState } from 'react';
-
-import Grid from 'components/Grid';
-import Message from 'components/Message';
-import Page from 'components/Page';
-import ResourceChart from 'components/ResourceChart';
```
apizzini, line 6: what about deleting `ResourceChart` code?

webui/react/src/pages/ClusterV1.tsx
```diff
@@ -1,58 +0,0 @@
...
-import Spinner from 'components/Spinner';
-import Agents from 'contexts/Agents';
-import { Resource, ResourceType } from 'types';
-import { categorize } from 'utils/data';
```
apizzini, line 10: same for `categorize`, was used only here.

apizzini (approved):
lgtm, just made a couple of suggestions about other code that potentially can be removed since used only in the old cluster page.

webui/react/src/pages/ClusterV1.tsx
hamidzr: 🤔  this one I think we'll use and it shouldn't add to our bundle size as long as we don't import and use it anywhere. If you prefer we don't keep it I'll remove it

apizzini: totally fine to keep!

PR #2120, opened 2021-03-24
ci: put GPUs in a separate GKE node pool from master

mackrorysd:
## Description

There are a few changes we had to make for CUDA 11 that we now also need to make for CUDA 10.2 images to have sufficiently updated drivers. GKE seems to silently fail at a lot of GPU-related things when your default node pool is GPUs, and it prefers to have a second node pool for accelerators.

Also, we no longer need to be on the "rapid" release channel for CUDA 11 support or use the beta CLI, so I've also started harmonizing between the CUDA-11 specific entries in CI and the generic GKE ones. The GKE orb is no longer usable for any tasks so it's gone and simply replaced with direct use of the gcloud CLI.

## Test Plan

Running CI here: https://app.circleci.com/pipelines/github/determined-ai/determined?branch=mackrory. There's still a PyTorch 1.8-related bug we need to address causing failures, but this fixes the GKE provisioning / driver version problems.


eecsliu (approved):
🚢 🛳 ⛴

mackrorysd:
The e2e-test is blocked on some unrelated GPU test failures I'm currently working on (one of which was an A100 capacity issue and that one's still waiting to timeout). This does get the tests to progress further, so merging anyway.

PR #2123, opened 2021-03-24
fix: tweak aggregated resource allocation history endpoint

dzhu:
## Description

- Add missing query SQL file, without which it didn't work at all.
- Format dates as YYYY-MM-DD in gRPC responses.
- Support monthly aggregation.

## Test Plan

- [x] carefully examine the list of unstaged files in my working tree
- [x] run the query on some test data and check that the dates come out right


master/internal/db/postgres_proto.go
```diff
@@ -38,18 +39,32 @@ func protoParser(rows *sqlx.Rows, val interface{}) error {
...
 		switch parsed := value.(type) {
 		case []byte:
 			var marshaled interface{}
-			if err := json.Unmarshal(parsed, &marshaled); err != nil {
+			if err = json.Unmarshal(parsed, &marshaled); err != nil {
 				return errors.Wrapf(err, "error parsing field: %s", key)
 			}
 			dest[key] = marshaled
+		case time.Time:
+			if colTypes[key] == "DATE" {
```
stoksc, line 65: if the column type for this key isn't `DATE`, it seems like we should error or something, not just continue without scanning a value.

master/internal/db/postgres_proto.go
```diff
@@ -38,18 +39,32 @@ func protoParser(rows *sqlx.Rows, val interface{}) error {
...
 	for key, value := range dest {
 		switch parsed := value.(type) {
 		case []byte:
 			var marshaled interface{}
-			if err := json.Unmarshal(parsed, &marshaled); err != nil {
+			if err = json.Unmarshal(parsed, &marshaled); err != nil {
 				return errors.Wrapf(err, "error parsing field: %s", key)
 			}
 			dest[key] = marshaled
+		case time.Time:
```
stoksc, line 64: could you just to_char the date in SQL to get around needing this? I think this is what has been done historically to make protoParser happy.  I don't really like that.. at all. Just curious what you think about protoParser in general.

master/internal/core.go
```diff
@@ -218,45 +218,62 @@ func (m *Master) getRawResourceAllocation(c echo.Context) error {
...
 	resp := &apiv1.ResourceAllocationAggregatedResponse{}
-	if err := m.db.QueryProto(
-		"get_aggregated_allocation", &resp.ResourceEntries, start.UTC(), end.UTC(),
-	); err != nil {
-		return errors.Wrap(err, "error fetching allocation data")
-	}
 
-	c.Response().Header().Set("Content-Type", "text/csv")
+	switch args.Period {
+	case "RESOURCE_ALLOCATION_AGGREGATION_PERIOD_DAILY":
```
stoksc, line 232: Could you convert args.Period to a `masterv1.ResourceAllocationAggregationPeriod` and then de-dup this code with the code in `api_master.go`?

stoksc (commented):
looks good, just some q's

master/internal/core.go
dzhu: Great idea, done!

master/static/srv/get_monthly_aggregated_allocation.sql
```diff
@@ -0,0 +1,25 @@
...
+        ELSE 'RESOURCE_ALLOCATION_AGGREGATION_TYPE_UNSPECIFIED'
+    END AS aggregation_type,
+    resource_aggregates.aggregation_key,
+    sum(resource_aggregates.seconds) AS seconds
+FROM
+    resource_aggregates,
+    const
+WHERE
+    -- `@>` determines whether the range contains the time.
+    const.period @> resource_aggregates.date
```
stoksc, line 21: did we consider a btree index on resource_aggregates.date? I think i remember the conversation somewhere.. if there are N users, M labels, this is what, N + M rows per date? If you query for a month maybe ~10^3 rows, after a year maybe ~10^4 rows in the table. By dumb math, it makes me think an index could speed up by a factor of 10 after this exists a while?

stoksc (approved):
looks great.

master/static/srv/get_monthly_aggregated_allocation.sql
dzhu: Sounds good, I'll add that and land.

dzhu: Oh, there's already a unique constraint, which [automatically adds an index](https://www.postgresql.org/docs/10/indexes-unique.html).

PR #2129, opened 2021-03-26
chore: change output format of JSON aggregated resource data 

dzhu:
Contains #2123; look at the last commit of this PR only.

## Description

Update the structure of the JSON returned for the aggregated endpoint to
match the ERD. Rather than a separate entry for each aggregation key, we
have one entry per period with maps containing the individual
subentries.

The CSV output remains the same, since we don't really want to express
structured data like that in CSV.

## Test Plan

- [x] run `det res agg` with all combinations of `--monthly` and `--json`, eyeball results


master/static/srv/get_monthly_aggregated_allocation.sql
```diff
@@ -1,25 +1,63 @@
 WITH const AS (
     SELECT
         daterange($1 :: date, $2 :: date, '[]') AS period
+),
+months AS (
```
stoksc, line 5: after reading a few times, with some parameters, seems this query can produce the other. i don't think it's a big enough deal to change though.

proto/src/determined/master/v1/master.proto
```diff
@@ -55,16 +55,17 @@ enum ResourceAllocationAggregationType {
   RESOURCE_ALLOCATION_AGGREGATION_TYPE_LABEL = 3;
 }
 
-// One instance of slots in the cluster being allocated to a task during a day
-// (aggregated).
+// One instance of slots in the cluster being allocated to a task during a
+// period (aggregated).
 message ResourceAllocationAggregatedEntry {
```
stoksc, line 60: Took me a second to grok this API, probably a lot better for a user I agree, but dang is it a lot funkier on the backend. Immediately think of https://en.wikipedia.org/wiki/Object%E2%80%93relational_impedance_mismatch.

stoksc (approved):
lgtm

proto/src/determined/master/v1/master.proto
dzhu: Heh, yeah; I hadn't heard that term, but that's basically exactly how I felt. I tell myself it's our job to do the funky stuff so everything comes out nice.

PR #2130, opened 2021-03-26
feat: add per-resource-pool task_container_defaults [DET-5177]

rb-determined-ai:
## Description

This is a great example of why we should finish the experiment config work.

Parsing user config first and merging defaults in afterwards would let us flexibly merge slightly different defaults in based on the contents of the user config, which is basically what the product ask is.


rb-determined-ai:
closing in favor of a feature branch (#2149)

PR #2132, opened 2021-03-26
chore: Moving back to Python 3.7 and PyTorch 1.7.1

mackrorysd:
## Description

Discovered some problems involving Python 3.8 (cloudpickle compatibility concerns that need to be addressed) and PyTorch 1.8 (needs some fixes from upstream Horovod, which has not fully tested it yet).

This includes the usual bumpenvs process, but also manually updating a few docs and the PyTorch lightning examples that referenced other out of date images.

## Test Plan

I will run post-commits here before merging (that said, I thought I did this last time and obviously missed some glaring problems): https://app.circleci.com/pipelines/github/determined-ai/determined?branch=mackrory.

## Commentary (optional)

Our docs are also lacking a section about the TF2 images, but I will add that separately.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,11 +97,11 @@ commands:
       - when:
           condition: <<parameters.tf1>>
           steps:
-            - run: docker pull determinedai/environments:py-3.7-pytorch-1.8-tf-1.15-cpu-145c0d8
+            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-1c4ea10
       - when:
           condition: <<parameters.tf2>>
           steps:
-            - run: docker pull determinedai/environments:py-3.7-pytorch-1.8-lightning-1.2-tf-2.4-cpu-145c0d8
+            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-lightning-1.2-tf-2.4-cpu-1c4ea10
 
   login-docker:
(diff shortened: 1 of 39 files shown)


mackrorysd:
test-webui failures are flakes caused by DET-5157. I've notified the web UI team that this is happening more often than they think - it's happening every time for now.

mackrorysd:
For the purpose of code-review, here are the changes I made in addition to what bumpenvs did. Apologies for squashing the commits in this particular branch before submitting the pull request.

rb-determined-ai:
Out of curiosity, how did these images get out-of-sync?  Do we know?  I know my expconf ones were because the time between when I started that work and when I landed it was so long.  But it looks like an alarming number of out-of-date images were laying around.

rb-determined-ai (approved):

mackrorysd:
> Out of curiosity, how did these images get out-of-sync? Do we know? I know my expconf ones were because the time between when I started that work and when I landed it was so long. But it looks like an alarming number of out-of-date images were laying around.

I would guess that could be the case for some of the others: patches which we were written before but committed after a change. I have a suspicion that might have been the case for PyTorch Lightning too. I might add a section to the bumpenv's README about doing a search for determinedai/environments each time to check for out of sync ones. mmdetection tests pull a pre-built image, so there would be some CI work to get that updated / tested regularly.

PR #2137, opened 2021-03-30
perf: support max_concurrent_trials for random and grid search [DET-3664]

liamcli:
## Description
This PR allows users to limit the maximum number of concurrent trials training at any given time for grid and random search.  While there is some overlap in capability with `max_slots`, `max_concurrent_trials` works for dynamic agents, which is not true for `max_slots`.  We expect to deprecate `max_slots` in a future release.   

## Test Plan
Existing tests for searchers pass and verified that the `max_concurrent_trials` constraint is not violated.

## Checklist

### DIFF
--- a/master/pkg/searcher/grid.go
+++ b/master/pkg/searcher/grid.go
@@ -10,9 +10,14 @@ import (
 )
 
 type (
-	// gridSearchState stores the state for grid. Though grid is stateless, it is useful
-	// on restart to know the type of searcher during restore, in the event it must be shimmed.
+	// gridSearchState stores the state for grid. The state will track the remaining hp settings
+	// that have yet to be created for evaluation.  PendingTrials tracks how many trials have
+	// active workloads and is used to check max_concurrent_trials for the searcher is respected.
+	// Tracking searcher type on restart gives us the ability to differentiate grid searches
+	// in a shim if needed.
 	gridSearchState struct {
+		PendingTrials    int              `json:"pending_trials"`
+		RemainingTrials  []hparamSample   `json:"remaining_trials"`
 		SearchMethodType SearchMethodType `json:"search_method_type"`
 	}
 	// gridSearch corresponds to a grid search method. A grid of hyperparameter configs is built. Then,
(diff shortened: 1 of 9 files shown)


master/pkg/searcher/grid.go
```diff
@@ -13,6 +13,8 @@ type (
 	// gridSearchState stores the state for grid. Though grid is stateless, it is useful
 	// on restart to know the type of searcher during restore, in the event it must be shimmed.
```
eecsliu, line 14: blocking: wouldn't adding `PendingTrials` and `RemainingTrials` make this stateful? Is the reasoning the same as what the comment describes for `SearchMethodType`? Also, update the comment if so.

master/pkg/searcher/random.go
```diff
@@ -8,10 +8,12 @@ import (
 )
 
 type (
-	// randomSearchState stores the state for grid. Though random is stateless, it is useful
+	// randomSearchState stores the state for random. Though random is stateless, it is useful
 	// on restart to know the type of searcher during restore, in the event it must be shimmed.
 	// This gives us the ability to differentiate single and random in a shim.
```
eecsliu, line 13: blocking: same thing as for grid - does this change make it stateful/modify comment?

eecsliu (commented):
Looks good - just three questions here

master/pkg/searcher/random.go
liamcli: thanks for catching this

master/pkg/searcher/util_test.go
```diff
@@ -367,7 +368,7 @@ func simulateOperationComplete(
 	return ops, nil
 }
 
-func saveAndReload(method SearchMethod) error {
+func saveAndReload(ctx context, method SearchMethod) error {
```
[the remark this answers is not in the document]
liamcli: not needed, was used in a previous iteration

eecsliu (approved):
👩‍🍳😗👌

stoksc:
@liamcli i'm not entirely sure what the behavior of this is going to be since there is no shim for the state changes you made. if someone pauses across an upgrade, their random/grid searcher may be in a funky state.

PR #2138, opened 2021-03-30
chore: move unets_tf_keras back to previous images

mackrorysd:
## Description

We know (and document) that there are some models that have fluctuations in performance on different versions of CUDA. This appears to be a significant case of that, causing nightlies to fail because of a timeout. The model does converge to high accuracy regardless. Simply reverting to the previous images for now - we need some tests exercising the old images anyway as they are still supported.


naren-determined (approved):
Looks good!

PR #2139, opened 2021-03-30
ci: adding coscheduler to static k8s test clusters, and test

mackrorysd:
## Description

This is a second attempt at this patch. The last time the tests passed by coincidence (they succeed 50% of the time even without the fix they are validating). The root cause of this is that Helm seems to have some flawed indentation logic for the values overridden via the CLI, and coscheduler was specific with additional newlines and whitespace. If we change the order of the values, it's the GCS bucket that ends up being indented - though it seems to cause no problem. Simply trimming whitespace from the coscheduler value.

## Test Plan

Verified on an upstream branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -543,7 +543,8 @@ commands:
             detVersion=<<parameters.det-version>>,\
             maxSlotsPerPod=<<parameters.gpus-per-machine>>,\
             checkpointStorage.type=gcs,\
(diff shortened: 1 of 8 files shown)


eecsliu (approved):
👨‍🍳 😗 👌

mackrorysd:
Merging - the CI failure is an unrelated and pre-existing test-webui flaked that is being worked on.

PR #2141, opened 2021-03-30
feat: add allocation aggregation by agent label and resource pool

dzhu:
## Description

Just add some new things parallel to the existing ones, pretty much.

A few existing field names are edited to be clearer, by request. That
will break behavior on previously computed aggregations, and also the
new aggregations won't be filled in for any databases that have already
been run with the previous version, but I consider all that acceptable
because we haven't had a release since that landed. If we ever need to
add more aggregations in the future, we'll figure it out then.

Also disable a Go lint that was raising an error on some of the new code
but turns out to be obsolete (and replaced by other lints that are still
enabled and not complaining) according to its documentation.

## Test Plan

- [x] populate database with test data and run queries


master/internal/core.go
```diff
@@ -306,17 +306,27 @@ func (m *Master) getAggregatedResourceAllocation(c echo.Context) error {
 	}
 
 	for _, entry := range resp.ResourceEntries {
-		for label, seconds := range entry.ByLabel {
-			if err = write("label", label, entry.PeriodStart, seconds); err != nil {
-				return err
+		do := func(agg string, vals map[string]float32) error {
```
armandmcqueen, line 309: Is there a more descriptive name we could use than `do`?

armandmcqueen (approved):

master/internal/core.go
dzhu: Perhaps I took the names-in-smaller-scopes-should-be-shorter principle too far. :slightly_smiling_face:  Renaming to `writeAggType`.

PR #2144, opened 2021-03-30
fix: avoid showing no-data message for a split second in hp-viz [DET-5099]

hkang1:
## Description

There is a split second where hp-viz shows `No data to plot` message when loading the data to populate the chart. The logic just needs a slight tweaking to avoid this.

## Test Plan

- [ ] when loading hp-viz, `No data to plot` should not show up before loading the learning curve chart.

## Checklist


apizzini (approved):
lgtm!

PR #2149, opened 2021-03-31
feat: infiniband support

rb-determined-ai:
Features in this PR:

* ResourcesConfig: Add a field for Devices
* EnvironmentConfig: Add fields for AddCapabilities, DropCapabilities
* TaskContainerDefaults: Add fields for AddCapabilities, DropCapabilities, and Devices
* ResourcePoolConfig: Add a field for TaskContainerDefaults

### DIFF
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -352,10 +352,79 @@ type CreateExperimentParams struct {
 	ValidateOnly  bool            `json:"validate_only"`
 }
 
+// getResourcePoolTaskContainerDefaults is a helper function that partially parses an experiment
+// config, and returns a TaskContainerDefaultsConfig that can be fed to
+// model.DefaultExperimentConfig().  This is of course way too convoluted to be "nice", but all of
+// this will go away when we can use schemas.Merge() rather than layers of json.Unmarshal().
+func getResourcePoolTaskContainerDefaults(
+	configBytes []byte,
+	masterConfig Config,
+) model.TaskContainerDefaultsConfig {
+	var out model.TaskContainerDefaultsConfig
+
+	// Lowest-priority settings are the main TaskContainerDefaults section.
+	if b, err := yaml.Marshal(masterConfig.TaskContainerDefaults); err != nil {
+		return out
+	} else if err = yaml.Unmarshal(b, &out); err != nil {
+		return out
+	}
+
+	// Now find out what ResourcePool the experiment wants.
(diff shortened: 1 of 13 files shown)


mackrorysd (approved):
+1 LGTM. Was able to test this successfully via experiment config and resource pool config.

Set capabilities to "IPC_LOCK" (in 1-length list form) and mapped /dev/infiniband to /dev/infiniband.

master/internal/core_experiment.go
```diff
@@ -352,10 +352,79 @@ type CreateExperimentParams struct {
 	ValidateOnly  bool            `json:"validate_only"`
 }
 
+// getResourcePoolTaskContainerDefaults is a helper function that partially parses an experiment
+// config, and returns a TaskContainerDefaultsConfig that can be fed to
+// model.DefaultExperimentConfig().  This is of course way too convoluted to be "nice", but all of
+// this will go away when we can use schemas.Merge() rather than layers of json.Unmarshal().
+func getResourcePoolTaskContainerDefaults(
```
armandmcqueen, line 359: This behavior is inconsistent with how per-resource-pool settings are done elsewhere (e.g. scheduler). In that case we use the global defaults if there are no resource-pool specific settings otherwise we use only the per-resource-pool settings (no partial inheritance). https://docs.determined.ai/latest/topic-guides/system-concepts/resource-pool.html#migrating-to-resource-pools

I think having different inheritance rules depending on which field we are talking about will be confusing.

stoksc:
@rb-determined-ai just curious, do we have a long term approach here, as opposed to users manually "scheduling" their own ib resources? also, what would happen in multi dtrain single node world if both users tried to add all in ib devices for two jobs on the same node? that case is semi-nonsense, but i'm just curious wyt.

master/internal/core_experiment.go
rb-determined-ai: After thinking about it, I agree that the merging behavior is the wrong behavior.

Also I think it goes without saying that we should be matching the other resource pool behaviors (which I was unaware of, so nice catch).

rb-determined-ai:
@stoksc 

> do we have a long term approach here, as opposed to users manually "scheduling" their own ib resources?

I don't have one; I'm open to ideas though.  Right now I think the users who want this feature are also the users who want to use integer numbers of entire compute nodes for each experiment.

rb-determined-ai:
Closing this PR since each commit was pushed to `upstream/infiniband` independent of this PR.

PR #2151, opened 2021-03-31
fix: avoid loop of effect in hp-viz when experiment is not supported [DET-5189]

hkang1:
## Description

When the experiment does not support hp-viz (single trial experiment or PBT experiment) there is an `useEffect` that causes an infinite loop of constantly updating the `filters` state.

## Test Plan

- [ ] opening hp-viz on a single trial or PBT experiment should not kill/bog down the page.

## Checklist


webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
```diff
@@ -244,6 +252,7 @@ const ExperimentVisualization: React.FC<Props> = ({
 
   // Update default filter hParams if not previously set.
   useEffect(() => {
+    if (!isSupported) return;
```
hkang1, line 255: This is the main line where filters were constantly being set.

hamidzr (approved):
code lgtm

PR #2154, opened 2021-03-31
fix: fix an issue with wrapping lr_scheduler

hamidzr:
## Checklist

### DIFF
--- a/harness/determined/pytorch/lightning/_adapter.py
+++ b/harness/determined/pytorch/lightning/_adapter.py
@@ -293,10 +293,12 @@ class LightningAdapter(PyTorchTrial):
                     "- Tuple of dictionaries as described, with an optional ‘frequency’ key.\n"
                 )
 
-            check.check_isinstance(lrs["scheduler"].optimizer, _LRScheduler,
-                                   "A returned LRScheduler from `configure_optimizers` is "
-                                   "missing the optimizer attribute."
-                                  )
+            check.check_isinstance(
+                lrs["scheduler"].optimizer,
+                Optimizer,
+                "A returned LRScheduler from `configure_optimizers` is "
+                "missing the optimizer attribute.",
+            )
 
             # switch the user's unwrapped optimizer with the wrapped version.
             lrs["scheduler"].optimizer = wrapped_opt
--- a/harness/determined/pytorch/lightning/_adapter.py
+++ b/harness/determined/pytorch/lightning/_adapter.py
@@ -6,6 +6,7 @@ import pytorch_lightning as pl
 import torch
 from pytorch_lightning.trainer.optimizers import TrainerOptimizersMixin
 from pytorch_lightning.utilities.model_helpers import is_overridden
+from torch.optim.lr_scheduler import _LRScheduler
 from torch.optim.optimizer import Optimizer
 from typing_extensions import Literal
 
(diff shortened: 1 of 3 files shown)


harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -266,18 +277,20 @@ def lightning_scheduler_dict_to_det(lrs: dict) -> LRScheduler:
...
-                    opt,
-                    self._pls.optimizers,
-                    "Must use an optimizer that is returned by wrap_optimizer()",
-                )
-            return LRScheduler(lrs["scheduler"], step_mode, frequency=lrs["frequency"])
 
-        optimizers = [self._pls.context.wrap_optimizer(opt) for opt in optimizers]
+            wrapped_opt = optimizers_dict[getattr(lrs["scheduler"], "optimizer", None)]
+            if wrapped_opt is None:
+                raise InvalidModelException("missing optimizer for lr_scheduler")
```
rb-determined-ai, line 283: As it is written, users who hit this message not know how to fix the issue or why they are encountering it.  Please make this message clearer as to what we are doing, why we are doing it, and how the user can work around the issue.

See `wrap_lr_scheduler()` for an example of what I mean:

```python
    raise det.errors.InvalidExperimentException(
        "detected that context.wrap_lr_scheduler() was called with an instance of "
        "torch.optim.lr_scheduer.ReduceLROnPlateau as the lr_scheduler.  This lr "
        "scheduler class does not have the usual step() parameters, and so it can "
        "only be used with step_mode=MANUAL_STEP.\n"
        "\n"
        "For example, if you wanted to step it on every validation step, you might "
        "wrap your lr_scheduler and pass it to a callback like this:\n"
        "\n"
        "class MyLRStepper(PyTorchCallback):\n"
        "    def __init__(self, wrapped_lr_scheduler):\n"
        "        self.wrapped_lr_scheduler = wrapped_lr_scheduler\n"
        "\n"
        "    def on_validation_end(self, metrics):\n"
        '        self.wrapped_lr_scheduler.step(metrics["validation_error"])\n'
    )
```

rb-determined-ai: Users will definitely break this.  I was helping a user who was passing a custom LRScheduler without a `step` method just a couple days ago.

harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -266,18 +277,31 @@ def lightning_scheduler_dict_to_det(lrs: dict) -> LRScheduler:
...
+                    "LRSchedulers (or lr_dict).\n"
+                    "- Dictionary, with an ‘optimizer’ key, and (optionally) a ‘lr_scheduler’ key"
+                    "whose value is a single LR scheduler or lr_dict.\n"
+                    "- Tuple of dictionaries as described, with an optional ‘frequency’ key.\n"
                 )
-            return LRScheduler(lrs["scheduler"], step_mode, frequency=lrs["frequency"])
 
-        optimizers = [self._pls.context.wrap_optimizer(opt) for opt in optimizers]
+            # switch the user's unwrapped optimizer with the wrapped version.
+            lrs["scheduler"].optimizer = wrapped_opt
```
rb-determined-ai, line 297: Can you also emit a check here that the scheduler object actually has an `optimizer` attribute?

Sorry, I didn't understand the check above, I thought that is what I was asking for before.

But the error message above looks fantastic : )

rb-determined-ai (approved):

shiyuann (approved):

harness/determined/pytorch/lightning/_adapter.py
hamidzr: : ) . it seems like the base lrscheduler has a [similar check ](https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L28) but doesn't hurt to have it here as well.

rb-determined-ai: Yes, but the issue is that people pass LRSchedulers which are not actually API compatible with the real _LRScheduler at all

PR #2167, opened 2021-04-05
chore: show trial metrics on webui [DET-5060]

apizzini:
## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/Profiles/utils.ts
+++ b/webui/react/src/pages/TrialDetails/Profiles/utils.ts
@@ -1,5 +1,6 @@
 import { useEffect, useState } from 'react';
 import { debounce } from 'throttle-debounce';
+import { AlignedData } from 'uplot';
 
 import {
   V1GetTrialProfilerAvailableSeriesResponse, V1GetTrialProfilerMetricsResponse,
@@ -27,7 +28,7 @@ export type MetricsAggregateInterface = {
 };
 
 export const convertMetricsToUplotData =
-  (data: Record<number, Record<string, number>>): Array<number[]> => {
+  (data: Record<number, Record<string, number>>): AlignedData => {
     const series: Array<number[]> = [];
     const timeSerie: number[] = [];
 
--- a/webui/react/src/pages/TrialDetails/Profiles/utils.ts
+++ b/webui/react/src/pages/TrialDetails/Profiles/utils.ts
@@ -1,0 +1,182 @@
+import { useEffect, useState } from 'react';
+import { debounce } from 'throttle-debounce';
+
+import {
+  V1GetTrialProfilerAvailableSeriesResponse, V1GetTrialProfilerMetricsResponse,
+} from 'services/api-ts-sdk';
+import { detApi } from 'services/apiConfig';
+import { consumeStream } from 'services/utils';
+
+export enum MetricType {
+  System = 'PROFILER_METRIC_TYPE_SYSTEM',
+  Timing = 'PROFILER_METRIC_TYPE_TIMING',
+}
+
+// {[metric_type]: {[name]: {[agent]: [gpu, ..], ..}, ..}, ..}
+export type AvailableSeries = Record<string, Record<string, Record<string, string[]>>>;
+
+export type MetricsAggregateInterface = {
+  // group information by {[batch]: {[name]: value, ..}, ..}
+  dataByBatch: Record<number, Record<string, number>>,
+  // group information by {[time]: {[name]: value, ..}, ..}
+  dataByUnixTime: Record<number, Record<string, number>>,
+  // set to false when the 1st event is received
+  isLoading: boolean,
+  // names to ease building the chart later
+  names: string[],
+};
+
+export const convertMetricsToUplotData =
+  (data: Record<number, Record<string, number>>): Array<number[]> => {
+    const series: Array<number[]> = [];
+    const timeSerie: number[] = [];
+
+    Object.entries(data).forEach(([ timeString, nameList ]) => {
+      timeSerie.push(parseInt(timeString));
+
+      Object.keys(nameList).forEach((name, nameIndex) => {
+        if (!series[nameIndex]) {
+          series[nameIndex] = [];
+        }
+
+        series[nameIndex].push(nameList[name]);
+      });
+    });
+
+    return [ timeSerie, ...series ];
+  };
+
+export const getUnitForMetricName = (metricName: string): string => {
+  if (metricName === 'cpu_util_simple') return '%';
+  if (metricName === 'disk_iops') return '';
+  if (metricName === 'disk_throughput_read') return 'bytes/second';
+  if (metricName === 'disk_throughput_write') return 'bytes/second';
+  if (metricName === 'free_memory') return 'Gigabytes';
+  if (metricName === 'gpu_util') return '%';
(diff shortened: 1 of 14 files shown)


webui/react/src/utils/number.ts
```diff
@@ -25,3 +25,12 @@ export const roundToPrecision = (n: number, precision = 6): number => {
   const factor = 10 ** precision;
   return Math.round((n + Number.EPSILON) * factor) / factor;
 };
+
+export const findFactorOfNumber = (n: number): number[] => {
```
hkang1, line 29: **nit:** move this above `isPercent` to keep it alphabetical?

webui/react/src/utils/number.ts
```diff
@@ -25,3 +25,12 @@ export const roundToPrecision = (n: number, precision = 6): number => {
...
+  const abs = Math.abs(n);
+  const factors = [];
+  for (let i = 1; i <= abs; i++) {
+    if (abs % i === 0) factors.push(i);
+  }
+  return factors;
```
hkang1, line 35: **nit suggestion:** You can cut the loop down to half by figuring out what the other multiplier is when if you find one factor. So for example if you are evaluating 18 and you come across 3 as a factor, you also know 18 / 3 = 6 is the other factor pair right then instead of having to reach 18 % 6 === 0 portion of the loop.

webui/react/src/utils/number.test.ts
```diff
@@ -65,4 +65,14 @@ describe('number utility', () => {
       expect(roundToPrecision(0.523843984)).toBe(0.523844);
     });
   });
+
+  describe('findFactorOfNumber', () => {
```
hkang1, line 69: good man :)

webui/react/src/services/decoder.ts
```diff
@@ -220,6 +220,7 @@ export const ioToExperimentConfig =
     description: io.description,
     hyperparameters: ioToExperimentHyperparameters(io.hyperparameters),
     labels: io.labels || undefined,
+    profiling: { enabled: io?.profiling?.enabled === true },
```
hkang1, line 223: **nit:** haven't tested this but this should work?
```
profiling: { enabled: !!io.profiling?.enabled },
```
- `?` after `io` shouldn't be needed
- double bang to evaluate to a boolean (down side is if `enabled` is anything other than undefined, null, 0, empty string it will be true, but should be perfectly fine here)

hkang1: just saw in `ioTypes` that it has to be boolean or undefined so double bang will for sure work

webui/react/src/utils/number.ts
apizzini: Do you think what I just pushed works?

hkang1:
2021-04-08 at 11 45 AM

webui/react/src/utils/number.ts
```diff
@@ -2,6 +2,21 @@ import { isString } from './data';
...
 
+export const findFactorOfNumber = (n: number): number[] => {
+  const abs = Math.abs(n);
+  const factors = [];
+
+  for (let i = 1; i <= Math.floor(Math.sqrt(abs)); i++) {
+    if (abs % i !== 0) continue;
+    factors.push(i);
+
+    if (abs/i === i) continue;
```
hkang1, line 13: shouldn't need this check because if `i` is divisible `abs/i` should also be.

apizzini: Are you 100% sure? I'm not and not really sure how to verify.

webui/react/src/utils/number.ts
```diff
@@ -2,6 +2,21 @@ import { isString } from './data';
...
+    factors.push(abs/i);
+  }
+
+  return factors.sort((a, b) => a - b);
```
hkang1, line 17: one way to avoid sorting is to just have two sets of factors, one for all the `i`s (aFactors) and one for all the `abs/i`s (bFactors). insert the `aFactors` with push and unshift for `bFactors` then return [ ...aFactors, ...bFactors ]. This is untested but the idea is that the cost of joining two already sorted arrays should be cheaper than iterating through an array and sorting them (best performance being `n log n`)

webui/react/src/utils/number.ts
```diff
@@ -2,6 +2,21 @@ import { isString } from './data';
 
 const PERCENT_REGEX = /^(\d+\.?\d*|\.\d+)%$/;
 
+export const findFactorOfNumber = (n: number): number[] => {
+  const abs = Math.abs(n);
+  const factors = [];
+
+  for (let i = 1; i <= Math.floor(Math.sqrt(abs)); i++) {
```
hkang1, line 9: nice! I didn't think to use `sqrt` (was thinking `abs/2` instead), but this is better.

webui/react/src/pages/TrialDetails/Profiles/utils.ts
```diff
@@ -0,0 +1,182 @@
+import { useEffect, useState } from 'react';
```
hkang1, line 1: **question:** we can cross that bridge if it does happen, but would any of this code be used anywhere else besides trial profiling?

hkang1:
changing of the filters has a few seconds pause before it updates the chart. Can we show a spinner in the meantime for the charts to indicate something is happening?

webui/react/src/pages/TrialDetails/Profiles/utils.ts
apizzini: I'm not sure, in case we can just move on a lower level? But seems premature because this code is very specific now and if we're gonna consider chunking the app we'll have to move back.

apizzini:
> changing of the filters has a few seconds pause before it updates the chart. Can we show a spinner in the meantime for the charts to indicate something is happening?

it's the default behaviour, don't rely on testing with the odd file i sent you.

hkang1 (commented):
The charts look snazzy!

hkang1 (approved):
Discussed all the outstanding issues. Once they are resolved, this should be good to go. Thanks for the great work @apizzini !

webui/react/src/utils/number.test.ts
```diff
@@ -65,4 +65,14 @@ describe('number utility', () => {
...
+    it('should find factors of number', () => {
+      expect(findFactorOfNumber(-12)).toStrictEqual([ 1, 2, 3, 4, 6, 12 ]);
+      expect(findFactorOfNumber(-1)).toStrictEqual([ 1 ]);
+      expect(findFactorOfNumber(0)).toStrictEqual([ ]);
+      expect(findFactorOfNumber(1)).toStrictEqual([ 1 ]);
+      expect(findFactorOfNumber(12)).toStrictEqual([ 1, 2, 3, 4, 6, 12 ]);
```
hkang1, line 75: can we had one more test case of:
```
1092 = 2 × 2 × 3 × 7 × 13
```

apizzini: 1092 has more factors than those, it's divisor are: 1,2,3,4,6,7,12,13,14,21,26,28,39,42,52,78,84,91,156,182,273,364,546,1092

Added more tests for a larger number with less divisors (to don't have a test resulting too long to respect esling max line length) and a test with a large prime number.

webui/react/src/utils/number.ts
apizzini: pushed.

PR #2170, opened 2021-04-05
feat: local clusters to store checkpoint data in home [DET-5154]

ioga:
## Description

- Default storage location for checkpoint data in local clusters deployed via `det deploy local` has been changed to OS-specific user data dir (e.g. `$XDG_DATA_HOME/determined` or `~/.local/share/determined` on Linux, and `~/Library/Application Support/determined` on macOS). Previously, `/tmp` was used.
- This location can be changed using `--storage-host-path` command line flag of `det deploy local`.
- If users provide their own custom `master.yaml` via `--master-config-path`, `checkpoint_storage` configuration in yaml will continue to take precedence.

## Test Plan
- Check that the old experiment data in /tmp is not lost after upgrade.
- Check the new default directory works ok, --storage-host-path, and --master-config-path.
 
## Checklist

### DIFF
--- a/master/cmd/determined-master/init.go
+++ b/master/cmd/determined-master/init.go
@@ -7,7 +7,6 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/determined-ai/determined/master/internal"
-	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/version"
 )
 
@@ -45,6 +44,10 @@ func (c configKey) AccessPath() string {
 
 func (c configKey) FlagName() string {
 	return strings.Join(c, "-")
+}
+
+func registerEnv(name configKey) {
+	v.BindEnv(name.AccessPath(), name.EnvName())
 }
 
 func registerString(flags *pflag.FlagSet, name configKey, value string, usage string) {
@@ -131,8 +134,8 @@ func registerConfig() {
 	registerString(flags, name("telemetry", "segment-webui-key"),
 		defaults.Telemetry.SegmentWebUIKey, "the Segment write key for the WebUI")
 
-	registerString(flags, name("checkpoint-storage", "type"),
-		model.DefaultCheckpointStorageType, "checkpoint storage type")
-	registerString(flags, name("checkpoint-storage", "host-path"),
-		model.DefaultSharedFSHostPath, "checkpoint storage host path")
+	// These env vars are used by `det deploy` to override host_path.
+	// We don't register pflags for these to avoid setting a default.
+	registerEnv(name("checkpoint-storage", "type"))
+	registerEnv(name("checkpoint-storage", "host-path"))
 }
--- a/master/cmd/determined-master/init.go
+++ b/master/cmd/determined-master/init.go
@@ -7,6 +7,7 @@ import (
 	"github.com/spf13/viper"
 
 	"github.com/determined-ai/determined/master/internal"
+	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/version"
 )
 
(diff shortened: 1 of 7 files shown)


master/cmd/determined-master/init.go
```diff
@@ -129,4 +130,9 @@ func registerConfig() {
 		defaults.Telemetry.SegmentMasterKey, "the Segment write key for the master")
 	registerString(flags, name("telemetry", "segment-webui-key"),
 		defaults.Telemetry.SegmentWebUIKey, "the Segment write key for the WebUI")
+
+	registerString(flags, name("checkpoint-storage", "type"),
```
stoksc, line 134: this should work fine afaik, but, you know, test it, lol. i'm not 100% sure how this `registerString` will work with our union types.

master/pkg/model/defaults.go
```diff
@@ -29,6 +29,13 @@ const (
 	defaultGPUImage = "determinedai/environments:cuda-10.2-pytorch-1.7-tf-1.15-gpu-1c4ea10"
 )
 
+const (
+	// DefaultCheckpointStorageType is SharedFS.
+	DefaultCheckpointStorageType = "shared_fs"
+	// DefaultSharedFSHostPath is the default path on hosts for SharedFS storage mounts.
+	DefaultSharedFSHostPath = "/tmp"
```
stoksc, line 36: I think adding this default will mean users can now not supply `checkpoint_storage` anywhere and have a valid exp conf, and this is new afaik, it always had to be set somewhere previously. Now that there is a valid default i'd make sure the defaulting logic with the master config's checkpoint config and the template checkpoint configs doesn't get busted. I'm not sure right off what would happen.

stoksc: If the flag is going to apply to the master config, and det deploy just needs to edit the master config defaults, do we need this code messing with the exp conf at all?

stoksc (commented):
looks good, just some q's

master/cmd/determined-master/init.go
ioga: I've had to register this type (and not just host_path) because the union type validator would throw up on host_path without type=shared_fs.

master/pkg/model/defaults.go
ioga: 1. I assume previously it'd come from the pre-packaged master.yaml, which explicitly specifies /tmp. I suppose the change is that if you pass a master.yaml which doesn't have checkpoint_storage, we'd now store data in /tmp instead of exiting. That probably won't break existing users.
2. There's a unit test `TestUnmarshalMasterConfigurationViaViper`, which effectively checks that the config values are the same as in `internal.DefaultConfig`. All of this helps to make that true.

stoksc: Yeah, but this change updates the default experiment config (`model.DefaultExperimentConfig`), not master config (`internal.DefaultConfig`).

ioga: master's checkpoint storage config comes from experiment config: https://github.com/ioga/determined/blob/master/master/internal/config.go#L30
do you think default checkpoint storage config should be separate and different for master and experiment configs?

stoksc: yeah i think that makes sense. I think experiments should have no default but master config should have a default (or have no default and have det-deploy supply the default for local cases).

ioga: went down the latter path of not changing the default, please see the updated code.

It doesn't use `registerString` machinery and doesn't expose the same config as a flag, because `pflags` methods require you to provide the default, so using `BindPFlag` (e.g. in `registerString`) ends up setting the default. And it can't be empty string either, because the parser throws up because it's not a valid config.

stoksc: sweet, thanks for updating.

stoksc (approved):
lgtm

PR #2171, opened 2021-04-05
docs: Update custom environment example

katport:
## Checklist

### DIFF
--- a/docs/how-to/custom-env.txt
+++ b/docs/how-to/custom-env.txt
@@ -159,9 +159,7 @@ those extra dependencies are quick to install, you may want to consider
 using a :ref:`startup hook <startup-hooks>`. For situations where
 installing dependencies via ``startup-hook.sh`` would take too long, we
 suggest building your own Docker image and publishing to a Docker
-registry like `Docker Hub <https://hub.docker.com/>`__. We recommend
-that custom images use one of the official Determined images as a base
-image (using the ``FROM`` instruction).
+registry like `Docker Hub <https://hub.docker.com/>`__.
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


docs/how-to/custom-env.txt
```diff
@@ -171,17 +171,19 @@ image (using the ``FROM`` instruction).
...
-.. code::
+.. code:: bash
 
    FROM determinedai/environments:cuda-10.2-pytorch-1.7-tf-1.15-gpu-0.10.0
-   RUN apt-get update && apt-get install -y unzip python-opencv graphviz
-   COPY environment.yml /tmp/environment.yml
-   COPY pip_requirements.txt /tmp/pip_requirements.txt
-   RUN conda env update --name base --file /tmp/environment.yml && \
-       conda clean --all --force-pkgs-dirs --yes
+   RUN apt-get update
```
mackrorysd, line 177: Nit: I think because of the way Dockerfiles skip lines that were previously successful, it's best-practice to keep apt-get update and apt-get install on the same line with the &&'s.

docs/how-to/custom-env.txt
```diff
@@ -171,17 +171,19 @@ image (using the ``FROM`` instruction).
 Here is an example of a ``Dockerfile`` that installs both ``conda``- and
```
mackrorysd, line 171: I'm not a big fan of the `# optional` comments because literally everything in the example is optional except maybe the FROM. What if we reworded the preamble here to be: "To simply add some additional dependencies so that they do not have to be installed at runtime, simply use a standard image as the base and add your dependencies. Here is an example of a ``Dockerfile`` that installs ``conda``-, ``pip``- and ``apt``-based dependencies.

docs/how-to/custom-env.txt
```diff
@@ -171,17 +171,19 @@ image (using the ``FROM`` instruction).
...
+   RUN apt-get install -y unzip python-opencv graphviz
+   COPY environment.yml /tmp/environment.yml # Optional
+   COPY pip_requirements.txt /tmp/pip_requirements.txt # Optional
+   RUN conda env update --name base --file /tmp/environment.yaml # Optional
+   RUN conda clean --all --force-pkgs-dirs --yes
    RUN eval "$(conda shell.bash hook)" && \
-       conda activate base && \
-       pip install --requirement /tmp/pip_requirements.txt
+       conda activate base
+   RUN pip install --requirement /tmp/pip_requirements.txt # Optional
```
mackrorysd, line 186: I think breaking this into a separate RUN would actually break things: https://pythonspeed.com/articles/activate-conda-dockerfile/

mackrorysd (changes requested):

docs/how-to/custom-env.txt
katport: updated

docs/how-to/custom-env.txt
katport: changed to separate the sections

docs/how-to/custom-env.txt
katport: Okay, updated but the line is now longer though because the tz data has to occur after the update but before graphviz.

mackrorysd (approved):
Looks good to me. Thank you!

PR #2177, opened 2021-04-05
chore: `make` default goal should be `build`.

ioga:
## Description
Rearrange simple Makefiles and set `DEFAULT_GOAL` in complex one so that the default make goal is to build.

## Checklist


dzhu (approved):
Heh, thanks for fixing this! I'd be mildly inclined to reorder the targets in all of the files, but it doesn't matter much.

Issue #2175, opened 2021-04-05
Not possible to turn off telemetry via values.yaml


PeterSulcs:
In the latest helm chart (0.5.0) it is not possible to disable telemetry via a `values.yaml` file:

https://github.com/determined-ai/determined/blob/master/helm/charts/determined/templates/master-config.yaml#L93

```
    {{- if .Values.telemetry }}
    {{- if .Values.telemetry.enabled }}
    telemetry:
      enabled: {{ .Values.telemetry.enabled }}
    {{- end }}
    {{- end }}
```

if `.Values.telemetry.enabled` is `false` this section is not added to the config yaml at all. Removing the inner-if statement results in the desired behavior.

vishnu2kmohan:
Thanks for the report @PeterSulcs, we'll look into this...

vishnu2kmohan:
Thanks to @dzhu for the [quick turnaround](https://github.com/determined-ai/determined/pull/2178), the fix should make our next official release.

PR #2178, opened 2021-04-05
fix: allow telemetry to be disabled under Helm [DET-5224]

dzhu:
## Description

The telemetry setting defaults to true, so it can't use the same pattern
as things that default to false.

Closes: #2175

## Test Plan

- [x] run the master locally on Kubernetes with `values.yaml` specifying telemetry on or off and check that the master responds


eecsliu (approved):
👩‍🍳😗👌

PR #2179, opened 2021-04-06
refactor: improve experiment terminal state [DET-5202]

hkang1:
## Description

While the experiment is active, the experiment detail page polls to keep the page updated with fresh data. When the experiment reaches a terminal state, the detail page stops polling. This can cause the rest of the page to be out of sync.

Solution is to allow the secondary polling to run one final time before it is terminated. A soft polling stoppage enables this behavior. The polling mechanism will mark the polling to be stopped but the timer for polling is allowed to still run its course to resolve the polling function one final time.

## Test Plan

- [ ] run a multi-trial experiment (`/e2e-tests/tests/fixtures/no_op/adaptive.yaml` is a good one) and view the experiment detail page. Change the trial table `# / page` option to `100 / page`. Leave the page open and when the experiment completes, all of the trials state should show `completed`.

## Checklist

### DIFF
--- a/webui/react/src/hooks/usePolling.ts
+++ b/webui/react/src/hooks/usePolling.ts
@@ -13,6 +13,12 @@ interface PollingOptions {
   runImmediately?: boolean;
 }
 
+/*
+ * When calling `stopPolling` with `terminateGracefully` set to true,
+ * the polling will be marked as stopped but we avoid killing the timer.
+ * This means that the polling function will allowed to run one last time
+ * before terminating.
+ */
 interface StopOptions {
   terminateGracefully?: boolean;
 }
--- a/webui/react/src/hooks/usePolling.ts
+++ b/webui/react/src/hooks/usePolling.ts
@@ -5,12 +5,16 @@ type PollingFn = (() => Promise<void>) | (() => void);
 interface PollingHooks {
   isPolling: boolean;
   startPolling: () => void;
-  stopPolling: () => void;
+  stopPolling: (options?: StopOptions) => void;
 }
 
 interface PollingOptions {
   interval?: number;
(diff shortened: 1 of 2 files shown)


webui/react/src/hooks/usePolling.ts
```diff
@@ -5,14 +5,18 @@ type PollingFn = (() => Promise<void>) | (() => void);
...
+  stopPolling: (options?: StopOptions) => void;
 }
 
 interface PollingOptions {
   interval?: number;
   runImmediately?: boolean;
 }
 
+interface StopOptions {
+  terminateGracefully?: boolean;
```
hamidzr, line 17: nit: could you add a comment as to what this option would/should do so it's externally more easily understandable?

webui/react/src/hooks/usePolling.ts
```diff
@@ -47,9 +51,9 @@ const usePolling = (pollingFn: PollingFn, options: PollingOptions = {}): Polling
     poll();
   }, [ poll ]);
 
-  const stopPolling = useCallback(() => {
+  const stopPolling = useCallback((options: StopOptions = {}) => {
```
hamidzr, line 54: unrelated to this pr: could we just   return `stopPolling` instead of ` return () => stopPolling();` line 67

hamidzr (approved):
works great. thanks, @hkang1

webui/react/src/hooks/usePolling.ts
hkang1: I used to return the function itself (if it's the only thing that needs to run during cleanup) but I ran into issues of the cleanup not running when it's returned in that format. 🤷

PR #2180, opened 2021-04-06
chore: idempotent searcher progress API [DET-5211]

stoksc:
## Description
This change alters the way in which trials notify the experiments of their progress to be idempotent, which helps to make the experiment's handling of progress stateless (eventually on a crash, progress will recover without being restored), makes the API easier to use if/when it is exposed publicly, and simplifies progress logic a bit.

## Test Plan
- [x] run tests, check progress for all searchers is reasonable.

## Checklist

### DIFF
--- a/master/internal/restore.go
+++ b/master/internal/restore.go
@@ -16,7 +16,7 @@ import (
 
 // The current experiment snapshot version. Once this is incremented, older versions should be
 // shimmed. Experiment and trial snapshots share a version currently.
-const experimentSnapshotVersion = 1
+const experimentSnapshotVersion = 2
 
 // Restore works by restoring from distributed consistent snapshots taken through the course
 // of an experiment. Snapshots within the system flow from the bottom up, starting with the
@@ -193,11 +193,13 @@ func (e *experiment) snapshotAndSave(ctx *actor.Context, ts trialSnapshot) {
 // experimentSnapshotShims maps a version to the shim that bumps that version.
 var experimentSnapshotShims = map[int]snapshotShimFunc{
 	0: shimExperimentSnapshotV0,
+	1: shimExperimentSnapshotV1,
 }
 
 // trialSnapshotShims maps a version to the shim that bumps that version.
 var trialSnapshotShims = map[int]snapshotShimFunc{
 	0: shimTrialSnapshotV0,
+	1: noopShim,
 }
 
 // shimExperimentSnapshot shims a trial snapshot to the version required by the master,
@@ -211,6 +213,10 @@ func shimExperimentSnapshot(snapshot []byte, version int) ([]byte, error) {
 // returning an error in the same cases as shimExperimentSnapshot.
 func shimTrialSnapshot(snapshot []byte, version int) ([]byte, error) {
 	return shimSnapshot(trialSnapshotShims, snapshot, version)
+}
+
+func noopShim(snapshot []byte) ([]byte, error) {
+	return snapshot, nil
 }
 
 func shimSnapshot(shims map[int]snapshotShimFunc, snapshot []byte, version int) ([]byte, error) {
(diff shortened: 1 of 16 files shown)


master/internal/restore.go
```diff
@@ -298,3 +304,22 @@ func shimTrialSnapshotV0(snapshot []byte) ([]byte, error) {
 
 	return json.Marshal(trialSnapshotV0)
 }
+
+// Version 1 => 2 shims
+
+// shimExperimentSnapshotV1 shims a v1 snapshot to a v2 snapshot. From v1 to v2,
+//
```
rb-determined-ai, line 311: finish typing this?

master/pkg/model/length.go
```diff
@@ -105,14 +105,14 @@ func NewLengthInEpochs(epochs int) Length {
...
 	switch ctx.defaultUnit {
 	case Records:
-		return float64(batches * ctx.globalBatchSize)
+		return Length{Unit: Records, Units: batches * ctx.globalBatchSize}
 	case Batches:
-		return float64(batches)
+		return Length{Unit: Batches, Units: batches}
 	case Epochs:
-		return float64(batches*ctx.globalBatchSize) / float64(ctx.recordsPerEpoch)
+		return Length{Unit: Epochs, Units: (batches * ctx.globalBatchSize) / ctx.recordsPerEpoch}
```
rb-determined-ai, line 115: Will this result in far less granular progress updates?

Also the comment says "rounded up" but this looks like rounding down to me.

Also rounding up seems like the wrong thing to do anyway.

rb-determined-ai (commented):
Nice work.  Looks like a nice cleanup even if we didn't need it for push arch.

I actually think that we should leave trial progress as an unitless float though... it doesn't need to be a Length type because the unit is already known, and the magnitude shouldn't have to be an integer.  The progress calculation shouldn't have to be math on Length objects, you could just pull the total expected length of trainings out, convert that to a float, and then divide by the sum of all the floats for each trial progress.

master/pkg/model/length.go
stoksc: yeah, i realized this while dozing off last night and forgot until now, thanks for reminding me. back to a unit-less float it is.

master/pkg/searcher/search_method.go
```diff
@@ -122,3 +121,11 @@ func (defaultSearchMethod) trialExitedEarly( //nolint: unused
 	context, model.RequestID, workload.ExitedReason) ([]Operation, error) {
 	return []Operation{Shutdown{Failure: true}}, nil
 }
+
+func sumTrialLengths(us map[model.RequestID]model.PartialUnits) model.PartialUnits {
```
rb-determined-ai, line 125: does this take into account invalidHP-exited trials?

stoksc: yeah they're deleted from the list elsewhere.

master/pkg/searcher/searcher.go
```diff
@@ -95,8 +94,7 @@ func (s *Searcher) TrialExitedEarly(
 
 	switch exitedReason {
 	case workload.InvalidHP:
-		unitsCancelled := s.UnitsCompletedByTrial[requestID]
-		s.TotalUnitsCompleted -= unitsCancelled
+		delete(s.TrialProgress, requestID)
```
rb-determined-ai, line 97: Oh, this answers my previous question.

Is there any risk of a race condition with this?  If the trial fails before the searcher state is saved?  Or is that all atomic?

stoksc: All atomic. The trial snapshot that indicates invalid hp will be saved in a transaction with the message that push us into this code path once the message is handled entirely.

master/internal/restore.go
```diff
@@ -298,3 +304,23 @@ func shimTrialSnapshotV0(snapshot []byte) ([]byte, error) {
...
+	var experimentSnapshotV1 map[string]interface{}
+	if err := json.Unmarshal(snapshot, &experimentSnapshotV1); err != nil {
+		return nil, err
+	}
+
+	searcherState := experimentSnapshotV1["searcher_state"].(map[string]interface{})
+
+	delete(searcherState, "total_units_completed")
+	delete(searcherState, "units_completed_by_trial")
+	searcherState["trial_progress"] = map[model.RequestID]model.Length{}
```
rb-determined-ai, line 323: Will this cause experiments after master-restart to end at < 100% progress if some of those trials are finished at the time that the experiment is paused?

stoksc: yes, this would happen but only on the upgrade from 0.14.6 to the next version. I felt this was fine, but I supposed I could just use `units_completed_by_trial` to properly shim this.

rb-determined-ai: I'd like to avoid the "this experiment is stuck at 50% even though its completed" bug reports

stoksc: Ok i changed it.

rb-determined-ai (commented):
Am I approving this or the followup version that removes SHA?

Seems like they should be reordered or combined.  I don't really care which.

stoksc:
@rb-determined-ai I think landing in this order is fine. that comment is outdated (oops), sha's progress is perfectly fine and that one has some work left i think.

rb-determined-ai (approved):

PR #2182, opened 2021-04-06
fix: bug in random search leading to incorrect total trials

liamcli:
## Description
Fixing bug where we incorrectly incremented `CreatedTrials` for an invalid HP trial.


eecsliu (approved):
lgtm, good fix!

PR #2185, opened 2021-04-07
chore: remove self from codeowners

stoksc:
## Checklist


justin-determined-ai (approved):
As long as lines don't break when there's no owner attached at all LGTM!

PR #2188, opened 2021-04-07
feat: add key tracker

hkang1:
## Description

Provide the ability for the WebUI to capture key events at the App level. Implement a publish/subscribe design so any component at any level can listening in on the key events to provide keyboard interactivity.

Implement one example: Listen for left and right arrow key to navigate through the modal gallery.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useKeyTracker.ts
+++ b/webui/react/src/hooks/useKeyTracker.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events';
 
-import { useCallback, useEffect, useRef } from 'react';
+import { useCallback, useEffect } from 'react';
 
 export enum KeyEvent {
   KeyUp = 'KeyUp'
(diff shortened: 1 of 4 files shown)


hkang1:
@cla-bot[bot] check

webui/react/src/hooks/useKeyTracker.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (e.target) {
+      const element = e.target as Element;
+      if ([ 'input', 'textarea' ].includes(element.tagName.toLowerCase())) return;
+      if (element.getAttribute('contenteditable')) return;
+    }
+    keyEmitter.emit(KeyEvent.KeyUp, e);
+  }, []);
+
+  useEffect(() => {
+    if (externalListeners.length !== 0) return;
```
apizzini, line 26: This is to make sure that if `useKeyTracker()` is called more than 1 time in the whole app, only 1 key tracker event listener/broadcaster runs, right?

apizzini (approved):
LGTM

webui/react/src/hooks/useKeyTracker.ts
hkang1: yeah exactly, I was trying to think of a singleton pattern to use for hooks but it seems like that is considered an anti-pattern in React. I did a slight update to just keep track of count because I am not actually using the array of listeners for removal as I originally intended, just to keep the tracker size smaller.

PR #2189, opened 2021-04-08
chore: remove validation operations [DET-5213]

stoksc:
## Description
This change alters the searcher to only emit a single runnable operation, "ValidateAfter". It also cleans up the trial <-> experiment interface to match and use what the coming sprinkle API's searcher API will look like.

## Test Plan
- [x] fix tests/code, existing tests should cover this all plenty
- [x] run the shim and make sure experiments complete as expected (took a snapshot and iterated until it was perfect)

## Commentary (optional)
There is some more simplification that could fall out of this, but I think this PR is plenty big/complex as is.

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -27,41 +27,35 @@ type (
 		create searcher.Create
 		trialSnapshot
 	}
-	trialCompletedOperation struct {
-		op      searcher.Runnable
-		metrics interface{}
+	trialReportValidation struct {
+		metrics workload.ValidationMetrics
 		trialSnapshot
 	}
-	trialValidation struct {
-		validationMetrics *workload.ValidationMetrics
-	}
-	trialCompletedWorkload struct {
-		completedMessage workload.CompletedMessage
-		completedOps     []trialCompletedOperation
-		earlyExit        *trialExitedEarly
+	trialReportEarlyExit struct {
+		reason workload.ExitedReason
 		trialSnapshot
+	}
+	trialReportProgress struct {
+		requestID model.RequestID
+		progress  model.PartialUnits
+	}
+	trialQueryIsBestValidation struct {
+		validationMetrics workload.ValidationMetrics
 	}
 	// trialClosed is used to replay closes missed when the master dies between when a trial closing in
 	// its actor.PostStop and when the experiment snapshots the trial closed.
 	trialClosed struct {
 		requestID model.RequestID
 	}
-	trialExitedEarly struct {
-		trialID      int
-		exitedReason workload.ExitedReason
-	}
-	trialSnapshot struct {
-		requestID model.RequestID
-		trialID   int
-		snapshot  []byte
-	}
-	trialReportProgress struct {
-		requestID model.RequestID
-		progress  model.PartialUnits
-	}
 	getTrial       struct{ trialID int }
 	killExperiment struct{}
 )
+
+type trialSnapshot struct {
+	requestID model.RequestID
+	trialID   int
+	snapshot  []byte
+}
 
 type trialSnapshotCarrier interface {
 	getSnapshot() trialSnapshot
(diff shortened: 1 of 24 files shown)


master/internal/trial_workload_sequencer.go
```diff
@@ -110,78 +109,82 @@ func (s *trialWorkloadSequencer) SetTrialID(trialID int) {
 }
 
 // WorkloadCompleted receives the searcher.CompletedMessage and updates the internal state of the
-// trialWorkloadSequencer accordingly. It is the only method that should update the state.
+// trialWorkloadSequencer accordingly. It and WorkloadFailed are the only methods that should update the state.
 // It snapshots this state whenever we've just completed a checkpoint, and should be the only method
 // to alter the snapshot, too.
 func (s *trialWorkloadSequencer) WorkloadCompleted(
```
stoksc, line 115: There's a lot i would cleanup in here now if it were necessary, but it's totally not, since this is moving to python.

stoksc:
@rb-determined-ai I put this up like we discussed to review the approach before I fix all the tests, cause fixing all the tests is a ton of work.

master/pkg/searcher/operations.go
```diff
@@ -168,7 +162,8 @@ func (c Checkpoint) String() string {
 	return fmt.Sprintf("{Checkpoint %s}", c.RequestID)
 }
 
-// Train is an operation emitted by search methods to signal the trial train for a specified length.
+// Train is an operation emitted by search methods to signal the trial train until
+// its total batches trained equals the specified length.
 type Train struct {
```
rb-determined-ai, line 167: Might be worth renaming the "Train" operation which ends with a "validaitonCompleted".  Naming-wise it looks more like we took a more full-featured system and deleted things in a way that resulted in these naming convetions, rather than that we chose the best names for each object.  And if we don't rename it now I doubt we ever do...

rb-determined-ai (commented):
The overall structure looks good.

master/internal/experiment.go
```diff
@@ -27,42 +27,36 @@ type (
 		create searcher.Create
 		trialSnapshot
```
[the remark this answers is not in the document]
stoksc: >  Isn't this PR sufficient to change the trial snapshotting to be just "start from wherever you last checkpointed"?

I don't think so. I don't think that can happen entirely until the trial workload sequencer moves - or we would need to make a change to reconstruct the trial workload sequencer state from where the last checkpoint was.

master/pkg/searcher/operations.go
stoksc: Yeah ok, hmm. Could call it `TrainUntilAndValidate`? I mean, that's exactly what it is.

master/internal/experiment.go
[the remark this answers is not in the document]
stoksc: Like, for that to be the case, the trial/trial_workload_sequencer would need to be able to recover their state from just the operations they had and their total_batches_processed.. I can take a hack if you'd like, but I would rather avoid overhauling the trial workload sequencer as it's getting rewritten. Up to you though, if it's helpful I could.

master/pkg/searcher/operations.go
rb-determined-ai: Maybe `ValidateAt` or `ValidateAfter`?

master/internal/experiment.go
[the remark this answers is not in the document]
rb-determined-ai: makes sense, I'm fine with this.

master/internal/experiment.go
```diff
@@ -27,42 +27,36 @@ type (
...
 	}
-	trialExitedEarly struct {
-		trialID      int
-		exitedReason workload.ExitedReason
-	}
-	trialSnapshot struct {
-		requestID model.RequestID
-		trialID   int
-		snapshot  []byte
+	trialQueryIsBestValidation struct {
```
shiyuann, line 42: Would those messages be replaced by the message types that are generated by prototype?

master/internal/experiment.go
```diff
@@ -199,22 +193,14 @@ func (e *experiment) Receive(ctx *actor.Context) error {
...
-			e.processOperations(ctx, ops, err)
-		}
-	case trialValidation:
-		if msg.validationMetrics != nil {
-			ctx.Respond(e.isBestValidation(*msg.validationMetrics))
-		}
+	case trialReportEarlyExit:
+		ops, err := e.searcher.TrialExitedEarly(msg.trialID, msg.reason)
+		e.processOperations(ctx, ops, err)
+	case trialQueryIsBestValidation:
```
shiyuann, line 202: can this be a field in `trialReportValidation`? we could handle the logic of saving the best there.

master/internal/trial_workload_sequencer.go
```diff
@@ -281,6 +254,14 @@ func (s trialWorkloadSequencer) Workload() (workload.Workload, error) {
 	}
 }
 
+func (s trialWorkloadSequencer) hasSearcherValidation() bool {
```
shiyuann, line 257: Is the plan to implement the searcher API and then move this logic to there?

master/internal/experiment.go
stoksc: protobuf? yeah maybe/probably. just gonna save that for a big PR to make all this into API.

master/internal/trial_workload_sequencer.go
stoksc: This logic will be in the python workload sequencer once it is implemented.

master/internal/experiment.go
stoksc: probably, with out it's written it's sort of a chicken/egg situation, `sequencer.WorkloadCompleted` needs to know if it was the best validation and it tells us whether to report the validation to the searcher, so the call ordering has to be ask trialQueryIsBestValidation -> sequencer.WorkloadCompleted -> maybe tell trialReportValidation. I could untie this but since @rb-determined-ai is rewriting the sequencer in python, I figured I may as well wait rather than refactoring code that'll be removed shortly.

rb-determined-ai (approved):
Sorry, didn't realize this was blocked on me.  I don't have any more comments, looks good.

PR #2190, opened 2021-04-08
chore: make the first glasbey color our brand color

hkang1:
## Description

The very first glasbey color currently is purple. Update the first color to be our brand color so for charts that only has one series, it will display using our active brand color.

## Checklist


webui/react/src/constants/colors.ts
```diff
@@ -21,7 +22,6 @@ export const GLASBEY = [
   [ 197, 89, 14 ],
   [ 67, 135, 23 ],
   [ 0, 211, 145 ],
-  [ 0, 121, 206 ],
```
hkang1, line 24: Removing this color because it is fairly close to the newly added color and it's only 15 or so down the list.

apizzini (approved):
great :)

PR #2196, opened 2021-04-09
fix: commands honor per-resource-pool settings

rb-determined-ai:
## Description

Fix the known bug in the per-resource-pool configurations with commands/notebooks/shells. 

## Test Plan

Tested manually and will test on customer hardware.

## Commentary (optional)

This refactors how the master's TaskSpec is handled throughout the codebase.


rb-determined-ai:
closing this PR against a feature branch as the actual feature is about to land on master.

PR #2197, opened 2021-04-09
docs: clarify k8s default pod spec behavior

eecsliu:
## Description
Added a short blurb clarifying the behavior of custom pod specs.

## Checklist


vishnu2kmohan (approved):

neilconway (commented):
I'm still a bit confused here -- what is the set of "necessary" fields, and what values will Determined use for them?

eecsliu:
> I'm still a bit confused here -- what is the set of "necessary" fields, and what values will Determined use for them?

There are a bunch of fields that Determined, most of them can be described as environment variables and pod metadata.

neilconway (approved):

eecsliu:
@neilconway I thought about the wording a little more - what do you think about something like this instead: 

"Fields that are not specified will remain at their default Determined values."

Do you think maybe this an easier way of breaking it to the user?

neilconway:
@eecsliu sgtm!

PR #2198, opened 2021-04-09
feat: harness collects profiler metrics [DET-5061]

armandmcqueen:
## Description

Moving from PR #2073 so that branch isn't in personal repo and can be used as a feature branch

## Checklist


armandmcqueen:
Note from old PR on how to disable GPU metrics if pynvml isn't available or there are no GPUs:
```
try:
    import pynvml
    pynvml.nvmlInit()
except ModuleNotFoundError:
    pass
except pynvml.NVMLError_LibraryNotFound:
    pass
```

harness/determined/profiler.py
```diff
@@ -0,0 +1,517 @@
...
+        #       is a long period of time between batch_idx being updated (such as during validation)
+        #       because it won't shut down until validation completes. We probably need to create a
+        #       new thread to enforce the timeout correctly.
+        self.max_collection_seconds = 300
+        self.has_finished = False
+
+        # Set up the thread responsible for making API calls
+        self.send_queue = queue.Queue()
+        self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+        self.sender_thread.start()
```
rb-determined-ai, line 78: Normally, python threds have a distinct "start/stop" lifetime which is a subset of their "init/gc" lifetime.  Or in other words, normally threads aren't started in `__init__()`, only created there.

Examples:
```python
# threading library
t = threading.Thread(target=my_func)
t.start()
t.join()
# gc happens later
```

Can we have a `ProfilingAgent.start()` method that starts both threads?  And a matching `ProfilingAgent.close()` method that (idempotently) stops/joins both threads?

Then the step after that is to write an `__enter__()` method that does `self.start(): return self` and an `__exit__()` that does `self.close` and you'll have a real nice API that will always shut down the threads it started, even in the case of uncaught exceptions.

harness/determined/profiler.py
```diff
@@ -0,0 +1,517 @@
...
+        self.start_on_batch = start_on_batch
+        self.end_after_batch = end_after_batch
+
+        # Track duration to stop collecting after 5 minutes. start_time also serves as
+        # the indicator that collection has begun.
+        self.start_time = None
+        # TODO: Currently we only check for shutdown in update_batch_idx(). This is a problem when there
+        #       is a long period of time between batch_idx being updated (such as during validation)
+        #       because it won't shut down until validation completes. We probably need to create a
+        #       new thread to enforce the timeout correctly.
```
rb-determined-ai, line 71: Why not just explicity shut down the collection when we transition from training -> validation?  Or hardcode the timeout into the collection thread rather than ProfilerAgent?

harness/determined/profiler.py
```diff
@@ -0,0 +1,517 @@
...
+        self.trial_id = trial_id
+        self.master_url = master_url
+        self.start_on_batch = start_on_batch
+        self.end_after_batch = end_after_batch
+
+        # Set up timer to stop collecting after 5 minutes
+        self.has_started = False
+        self.has_finished = False
+        self.max_collection_seconds = 300
+        self.shutdown_timer = threading.Timer(self.max_collection_seconds, self._end_collection)
```
rb-determined-ai, line 68: Kicking off a thread just for this seems like overkill when you could just start the SysMetricCollectorThread with a parameter of 300, so that it just dies if it collects that far.

That would have the extra benefit that you wouldn't the shutdown_lock since `_end_collection` would only get called from the main thread.

harness/determined/profiler.py
```diff
@@ -0,0 +1,517 @@
...
+                return
+
+            # Shut down in reverse creation order
+            self.sys_metric_collector_thread.kill()
+            self.sys_metric_collector_thread.join()
+
+            self.sender_thread.kill()
+            self.sender_thread.join()
+
+            self.has_finished = True
```
rb-determined-ai, line 124: Do you actually need the shutdown lock and the has_finished check?  I think the kill() calls and the join() are basically already idempotent (other than extra dummy messages getting put into queues... pretty harmless)

harness/determined/profiler.py
armandmcqueen: So if you don't mind, let's defer this problem. There is a PR that follows up on this one that hooks it into the TrialControllers. The current proposal is to use a context manager so we can move the start into that. If we decide not to use a context manager, then we can add a start() function

harness/determined/profiler.py
armandmcqueen: Collecting Timings and batching them will probably require a second thread to run in parallel to the SysMetricsCollector thread that should also shut down after 5 minutes. In that case it is cleaner to have a single shutdown timer. If not, the cost of having the shutdown timer really isn't that high.

harness/determined/profiler.py
armandmcqueen: I think using a single lock is cleaner, easier to reason about and it comes at no meaningful cost. `kill` doesn't need to be idempotent, that's just a benefit of how it's currently implemented, which could easily change in the future. I'd rather write code that handles the concurrency explicitly in the function that needs to be threadsafe rather than relying on implementation details of a function that has no reason to remain idempotent.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # TODO: Add data structure to batch timings and then send to SenderThread
+        #       Does this need to be its own thread to flush correctly?
+        # self.timings_batcher = TimingsBatcher()
+
+    @property
+    def is_enabled(self) -> bool:
+        return self.has_started and not self.has_finished
+
+    def update_batch_idx(self, new_batch_idx: int) -> None:
```
shiyuann, line 108: q: should this be called for each batch? If so, probably add a check here.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+                last_measurement_time = time.time()
+
+            # Check if it is time to flush the batch and start a new batch
+            if time.time() - batch_start_time > self.FLUSH_INTERVAL:
+                self.send_queue.put(self.current_batch.convert_to_post_format())
+                self.current_batch.clear()
+                batch_start_time = time.time()
+
+
+# This is a thread that exists solely so that we can make API calls without blocking
```
shiyuann, line 406: nit: can you move this to the docstring of this class.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+                sleep_time = next_collection - now
+                try:
+                    msg = self.control_queue.get(timeout=sleep_time)
+                    if isinstance(msg, ShutdownMessage):
+                        # Drop any partial batches
+                        return
+                except queue.Empty:
+                    pass
+
+            next_collection += self.MEASUREMENT_INTERVAL
```
shiyuann, line 348: q: should this be
```suggestion
            next_collection = time.now() + self.MEASUREMENT_INTERVAL
```
in case `MEASUREMENT_INTERVAL` is very small?

 In practice, `next_collection - now` could be negative.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+    - NetworkSentThroughput = Measured in Gigabit/s
+    - NetworkRecvThroughput = Measured in Gigabit/s
+    - DiskIops
+    - DiskReadThroughput = Measured in bytes/second
+    - DiskWriteThroughput = Measured in bytes/second
+    - GpuUtilization = Measured in percent
+    """
+
+    FLUSH_INTERVAL = 10  # How often to make API calls
+    MEASUREMENT_INTERVAL = 0.1
```
shiyuann, line 297: q: should this be configurable?

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+                "trialId": self.trial_id,
+                "name": name,
+                "agentId": self.agent_id,
+                "gpuUuid": gpu_uuid_label,
+                "metricType": metric_type,
+            }
+
+        trial_profiler_metrics_batches = []
+        for metric_name in self.batch.keys():
+            # TODO: Don't forget to include GPU Memory
```
shiyuann, line 245: style: generally, we need to include a ticket number or this would be resolved in the same PR otherwise it wouldn't be tracked.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+            if metric_name != SysMetricType.GPU_UTIL_METRIC and len(self.batch[metric_name]) > 0:
+                trial_profiler_metrics_batches.append(
+                    to_post_format(
+                        self.batch[metric_name],
+                        make_labels(metric_name, SYSTEM_METRIC_TYPE_ENUM),
+                    )
+                )
+
+            # GPU Metrics need to be grouped by GPU UUID
+            # TODO: Don't forget to include GPU Memory
```
shiyuann, line 255: same

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+        self.control_queue: "queue.Queue[Union['StartMessage', 'ShutdownMessage']]" = queue.Queue()
+        self.current_batch = SysMetricBatcher(trial_id, agent_id)
+        self.current_batch.clear()
+
+        super().__init__(daemon=True)
+
+    def activate(self) -> None:
+        self.control_queue.put(StartMessage())
+
+    def kill(self) -> None:
```
shiyuann, line 312: This is actually "stop" rather than "kill".

```suggestion
    def stop(self) -> None:
```

shiyuann (commented):
looks good overall. So I left some minor comments.

harness/determined/profiler.py
rb-determined-ai: This was my idea actually.  This keeps the interval as close to regular as possible for normal intervals.  If the interval is too small, it exhibits the behavior of collecting as fast as technically possible, which seems like the right thing to do if the interval is too small.

harness/determined/profiler.py
rb-determined-ai: Sure, I'm ok deferring this.  Although I think we should use a context manager and the context manager should call start() in it's `__enter__()` function.  That way the primary usage pattern is super easy, but you still give the flexibility to call it in whatever strange calling pattern you want (such as we needed to integrate it into a Keras Callback).

harness/determined/profiler.py
rb-determined-ai: Honestly I think any differences between cleanliness and ease of reasoning between either the lock solution or the queue solution are pretty tiny.

I guess that's not really what I'm concerned about.  I'll respond in more detail on the other comment.

harness/determined/profiler.py
rb-determined-ai: Sure, but why would you not just have the collector thread always send a Shutdown message to the sender thread?  That way you really do only have one timer in the system (I'm counting the collector thread as a form of a timer during its normal operation).  You also would get that totally for free.

I think what I am more concerned about here is that there's no way to close down this Timer thread.  It's not that it's *that* expensive (though it is creating an entire thread where a simple `if` statement would suffice); it's not that it's *that* risky to let it just sit around for 5 minutes in the background.  But to me it's definitely an "ease of reasoning about" issue.

If there is one pattern in concurrency that makes concurrency easy to reason about, it's [structured concurrency](http://libdill.org/structured-concurrency.html), which is basically the rule that anything that spins off subroutines will always clean up the subroutines before exiting.  It's easy to reason about because when the parent object is closed, you know that all child objects (threads, subprocesses, whatever) are also closed.  Also, it is easy to test because if you ever run two similar tests in a row you don't have leftovers from one test stepping on the next one.

We didn't used to practice structured concurrency in the harness, but we would run into weird bugs where e.g. unit tests would OOM a GPU if you tried to run the whole unit test suite all at once (https://determinedai.atlassian.net/browse/DET-2026) or there was indeterministic shutdown order between when ZMQ sockets got closed vs when the interpreter shut down, which left cryptic error messages everywhere.

harness/determined/profiler.py
shiyuann: If sleep time can be negative, it seems fine to me.

harness/determined/profiler.py
armandmcqueen: You have the Profiler wrapper, then you have two child threads the SysMetricCollector and the TimingsBatcher, both of which depend on a third child thread, SenderThread. Putting the SysMetricCollector in charge of controlling timeouts and the SenderThread doesn't make any sense to me - then you need to synchronize across two threads to shut down correctly when the SysMetricCollector and TimingsBatcher threads otherwise wouldn't need to communicate at all. 

If you insist, I will implement a custom Timer class that can be manually cleaned up, but I feel like that is spending time solving a problem that doesn't really occur. Is the specific case you are concerned about is when you write a test that run the harnesses multiple times inside the same process so the timer persists across harness executions?

harness/determined/exec/worker_process.py
```diff
@@ -27,6 +28,9 @@ def main() -> None:
 
     config_logging(worker_process_env)
 
+    # API code expects credential to be available as an environment variable
```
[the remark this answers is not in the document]
armandmcqueen: This is just how authentication works in our `common/api` code AFAIU. While it's not ideal, I think that refactoring the python API code to not rely on environment variables is out of scope of this PR. Non-DTrain already relies on this envvar-based pattern IIUC, this code is just fixing the bug where the credentials aren't passed around correctly in DTrain.

harness/determined/profiler.py
```diff
@@ -0,0 +1,559 @@
...
+SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
+
+LOG_NAMESPACE = "determined-profiler"
+
+
+class ProfilerAgent:
+    """
+    Agent that collects metrics and sends them to the master. It has:
+    - a thread to send data to the Master API (sender_thread)
+    - a thread to collect System Metrics and periodically flush to
```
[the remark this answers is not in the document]
armandmcqueen: It's using capitalization to emphasize that these are concepts in Determined and not just descriptions.

harness/determined/profiler.py
armandmcqueen: It should be configurable by us in the future, but it doesn't not need to be configurable by users

harness/determined/profiler.py
armandmcqueen: I just missed these TODOs. They need to be fixed before this PR is merged and I'll fix that now. Nice catch!

harness/determined/profiler.py
armandmcqueen: Why is this "stop" rather than "kill"? It doesn't complete active work before shutting down, it just drops any in-progress batches, which seems more like "kill" to me.

harness/determined/profiler.py
armandmcqueen: I'll write a custom Timer class. That seems like the solution that will make us both happy

harness/determined/profiler.py
armandmcqueen: That's an interesting point. It should be called for every batch, but I'm not certain if it's incorrect if it isn't, particularly when we think about it being used outside of our harness/with the generic APIs. We'll talk about this at standup.

harness/determined/_execution.py
```diff
@@ -126,7 +126,9 @@ def _make_local_execution_env(
         det_trial_unique_port_offset=0,
         det_trial_runner_network_interface=constants.AUTO_DETECT_TRIAL_RUNNER_NETWORK_INTERFACE,
         det_trial_id="",
+        det_agent_id="local",
```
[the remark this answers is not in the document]
armandmcqueen: I'll make it empty. Out of curiosity, why does it matter?

harness/determined/profiler.py
armandmcqueen: If the timeout is negative, the code will raise an exception. Good catch. I changed it so that the sleep time is never less than 0.

harness/determined/profiler.py
armandmcqueen: Fixed

harness/determined/_execution.py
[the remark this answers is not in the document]
armandmcqueen: Fixed

harness/determined/profiler.py
armandmcqueen: Fixed

harness/determined/profiler.py
armandmcqueen: Fixed

armandmcqueen:
Updated the PR a good bit since the last reviews:

- Addressed most comments
- We now check if we pynvml is functioning so that we don't get failures if pynmvl isn't available/there aren't GPUs. I don't love the implementation, but a full-blown wrapper around pynvml like PolyHorovod seemed like overkill
- Added more inputs to the ProfilingAgent such as global rank and local rank. 
- Separated the concept of 'enabled' from 'active'. There is more detail on this in the design doc
- Added `start` and `end` functions as well as a context manager for the ProfilingAgent
- Added a custom timer that can be cleaned up if we need to exit before it completes

I also added a design doc to the codebase, which I feel like people might object to. My thinking is that in 6 months or a year when I don't remember the details of this code or when someone else needs to work on this code, it's really useful to have some docs to remember how it works and why it was written this way.  When I inline substantial documentation, the problem is that the implementation and the documentation drift apart over time and the documentation becomes misleading. One pattern I've seen is to include immutable design docs directly in the codebase with a date attached. This way it doesn't need to be updated, but it is still useful as a reference. I'm curious what people think about the idea

harness/determined/profiler.py
armandmcqueen: We talked in person. Sticking with the Timer thread for now (custom so it can follows structured concurrency)  so we don't need to introduce communication between the SysMetricCollector and TimingsBatcher threads.

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+
+            self.has_finished = True
+
+    def record_timing(self, timing: float) -> None:
+        if not self.is_active:
+            return
+        # TODO: Add new timing to TimingBatcher
+
+
+class CustomTimer(threading.Thread):
```
rb-determined-ai, line 231: PreemptibleTimer?

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+    timer.kill()  # This is idempotent and will work fine if the timer has gone off
+    ```
+    """
+
+    def __init__(
+        self, duration: int, callback: Callable, callback_args: Optional[List[Any]] = None
+    ):
+        self.duration = duration
+        self.callback = callback
+        self.callback_args = callback_args if callback_args is not None else []
```
rb-determined-ai, line 251: callback_args seems unnecessary; I think args should be bound into a closure.  `labmda: my_fn(my_args)` is pretty common in python.

harness/determined/profiler_design_april_2021.md
```diff
@@ -0,0 +1,114 @@
+# Profiler Design Notes (April 2021)
```
rb-determined-ai, line 1: Is the plan to commit this into the codebase?

harness/setup.py
```diff
@@ -55,6 +55,7 @@
         "docker-compose>=1.13.0",
         "tqdm",
         "appdirs",
+        "backoff",
```
rb-determined-ai, line 58: I'm flagging this as requiring changes to task environments so we are not surprised on release-party day.

harness/determined/profiler_design_april_2021.md
rb-determined-ai: I see your comment now.  I'm not opposed to inline developer docs on principle (I've written [some](https://github.com/determined-ai/determined/blob/master/harness/determined/pytorch/README.md) myself).

But in this case I don't think the profiling system here is hard to understand, or convoluted, or presents a complex API to justify committing the docs to the repository.  Honestly I think the docstring that is already in ProfilingAgent's sums the system up plenty well.

harness/determined/exec/worker_process.py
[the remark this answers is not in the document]
shiyuann: @rb-determined-ai did quite a lot of work getting rid of using environment variables to pass around information. If he is fine with this, let's create a ticket to refactor this later.

harness/determined/profiler.py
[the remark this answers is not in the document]
shiyuann: I see. I couldn't find them in the docs or the codebase. Can you define them in this file or somewhere else?

harness/determined/profiler.py
shiyuann: It looks like `new_batch_idx` should be at least larger than or equal to `current_batch_idx`.

shiyuann (commented):
The current state looks good. The only things are the TODOs are not attached with ticket numbers so I'll review them after you fill them up.

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+            not self.is_active
+            and not self.has_finished
+            and self.current_batch_idx >= self.start_on_batch
+        ):
+            self._begin_collection()
+
+        # Check if we should stop collecting metrics due to batch idx being exceeded
+        if (
+            self.is_active
+            and self.end_after_batch is not None
```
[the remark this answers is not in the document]
armandmcqueen: I think we have to keep the second line so I'm going to leave this as is

harness/determined/profiler.py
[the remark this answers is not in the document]
armandmcqueen: We'll add them in the docs. I updated the docs ticket to make sure we do that.

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+import psutil
+
+from determined.common import api
+from determined.common.api import TrialProfilerMetricsBatch
+
+SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
+
+LOG_NAMESPACE = "determined-profiler"
+
+SHOULD_PROFILE_GPUS = None
```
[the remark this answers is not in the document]
armandmcqueen: Fixed

harness/determined/profiler.py
armandmcqueen: Fixed

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+    """
+
+    def __init__(
+        self, duration: int, callback: Callable, callback_args: Optional[List[Any]] = None
+    ):
+        self.duration = duration
+        self.callback = callback
+        self.callback_args = callback_args if callback_args is not None else []
+        self._timer_has_begun = False
+        self.control_queue: "queue.Queue[Union['StartMessage', 'ShutdownMessage']]" = queue.Queue()
```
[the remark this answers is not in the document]
armandmcqueen: Fixed

harness/determined/profiler.py
armandmcqueen: Fixed - I just removed the concept of args altogether since it's just dead code now

harness/determined/profiler_design_april_2021.md
armandmcqueen: Removed it.

harness/setup.py
armandmcqueen: I think this means that after this PR is merged and prior to the next release, I need to make sure that the task envs is rebuilt by adding any commit. Is that correct?

harness/determined/profiler.py
```diff
@@ -0,0 +1,666 @@
...
+        if not self.is_enabled:
+            return
+
+        self.current_batch_idx = new_batch_idx
+        self.sys_metric_collector_thread.update_batch_idx(self.current_batch_idx)
+
+        # Check if we should start collecting metrics
+        if (
+            not self.is_active
+            and not self.has_finished
```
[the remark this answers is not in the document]
armandmcqueen: Fixed

harness/determined/profiler.py
armandmcqueen: Fixed

armandmcqueen:
All comments have been addressed except for the ones that I left unresolved.

harness/setup.py
rb-determined-ai: Yes.

harness/determined/exec/worker_process.py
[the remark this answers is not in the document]
rb-determined-ai: I'm fine with this for now.  But yeah, let's make a ticket.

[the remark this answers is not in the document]
rb-determined-ai: @armandmcqueen can you make the ticket?

rb-determined-ai (approved):
nice work!

PR #2199, opened 2021-04-09
chore: hook up agent and timings [DET-5062, DET-5204]

stoksc:
## Description
This change hooks the profiler agent up to the PyTorch, Estimators and Keras trial controllers and adds timing collection code.

## Test Plan
- [x] gonna try to write a short e2e test

## Commentary (optional)
WIP, I'm still getting it working but the general ideas are here.

## Checklist


harness/determined/profiler.py
```diff
@@ -8,17 +14,61 @@
...
+    @abc.abstractmethod
+    def update_batch_idx(self, new_batch_idx: int):
+        pass
+
+    @abc.abstractmethod
+    def record_timing(self, metric_name: str):
+        pass
+
+
+class NoopProfilerAgent(ProfilerAgent):
```
armandmcqueen, line 53: I don't think we need a secondary class. The ProfilerAgent has a property called `is_enabled`, and all of the operations should be no-ops if that property is false. We will need that logic in order to keep the API consistent after the profiling is done, so I think we should reuse it instead of adding NoopProfilerAgent.

harness/determined/profiler.py
```diff
@@ -91,6 +140,7 @@ def is_enabled(self):
     def update_batch_idx(self, new_batch_idx: int):
         self.current_batch_idx = new_batch_idx
         self.sys_metric_collector_thread.update_batch_idx(self.current_batch_idx)
+        self.send_queue.put(self.timings_batcher.convert_to_post_format())
```
armandmcqueen, line 143: Flushing every batch is too many API calls I think. We should be able to handle training at hundreds of batches per second.

harness/determined/profiler.py
stoksc: I have an abstract class to keep the interface in line. I just thought in real code, it was better to pass then accidentally do anything in real training code.

harness/determined/profiler.py
stoksc: Yeah, i think i can easily change this to everything N batches, or timing. This was just a rough sketch of the idea and I was kinda avoiding fixing this code much cause I foresaw conflicts.

stoksc: You can prelim review but a lot may change in the next few minutes,  I'm cleaning it up now after rebasing, sorry I'll unassign.

harness/determined/profiler.py
```diff
@@ -105,29 +155,45 @@ def update_batch_idx(self, new_batch_idx: int):
...
-    def record_timing(self):
+    @contextmanager
+    def record_timing(self, metric_name: str) -> None:
+        # TODO(brad): Check within batch range.
         if not self.is_enabled:
             return
-        # TODO: Add new timing to TimingBatcher
-
+        start = default_timer()
+        yield
```
armandmcqueen, line 185: I don't think we should make the main public facing API a context manager. It's not very flexible for non-PyTorch frameworks.

harness/determined/profiler.py
```diff
@@ -225,6 +291,53 @@ def make_labels(name: str, metric_type: str, gpu_uuid_label: str = "") -> Dict[s
...
+                "timestamps":timestamps,
+                "labels": {
+                    "trialId": self.trial_id,
+                    "name": metric_type,
+                    "agentId": self.agent_id,
+                    "metricType": TIMING_METRIC_TYPE_ENUM
+                }
+            }
+            post_formatted.append(single_metric_batch)
+        self.batch = {}
```
armandmcqueen, line 337: We should remove this side-effect from `convert_to_post_format` or rename the function

harness/determined/profiler.py
```diff
@@ -532,3 +645,7 @@ def measure(self, batch_idx):
             except pynvml.NVMLError as e:
                 logging.info(f"{LOG_NAMESPACE}: failed to sample GPU memory for GPU {i}: {e}")
         return measurements
+
+
+def get_rank_if_horovod_process_else_return_zero() -> Optional[int]:
+    return int(os.getenv("HOROVOD_RANK", 0))
```
armandmcqueen, line 651: Won't this mean that only one agent collects metrics?

harness/determined/profiler.py
```diff
@@ -78,9 +129,7 @@ def __init__(self, trial_id, agent_id, master_url, start_on_batch, end_after_bat
         self.sys_metric_collector_thread = SysMetricCollectorThread(trial_id, agent_id, self.send_queue)
         self.sys_metric_collector_thread.start()
 
-        # TODO: Add data structure to batch timings and then send to SenderThread
-        #       Does this need to be its own thread to flush correctly?
-        # self.timings_batcher = TimingsBatcher()
+        self.timings_batcher = TimingMetricBatcher(trial_id, agent_id)
 
     @property
     def is_enabled(self):
```
armandmcqueen, line 135: Since the experiment config field is called "profiling enabled", we should probably rename this variable so it isn't confusing. Maybe `is_active`?

harness/determined/profiler.py
stoksc: yeah true good catch, need local rank.

stoksc: or maybe it's a feature to not crash the master 🤣

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -285,20 +287,28 @@ def _train_for_step(
...
 
         for batch_idx in range(start, end):
-            batch = next(self.training_iterator)
+            self.profiler.update_batch_idx(batch_idx)
+            with self.profiler.record_timing("dataloader_next"):
+                batch = next(self.training_iterator)
             num_inputs += pytorch.data_length(batch)
-            batch = self.context.to_device(batch)
+
+            with self.profiler.record_timing("to_device"):
```
armandmcqueen, line 295: Maybe "dataloader_to_device"? We can keep the current names for now, but once we look at the webui, we will probably want to think more about what we should be naming the pieces.

harness/determined/_trial_controller.py
```diff
@@ -64,6 +64,7 @@ def pre_execute_hook(env: det.EnvContext, hvd_config: horovod.HorovodContext) ->
     @abc.abstractmethod
     def from_trial(
         trial_inst: "det.Trial",
+        prof: profiler.ProfilerAgent,
```
armandmcqueen, line 67: So I'll leave it up to you since it doesn't really matter, but I would prefer we call the variable `profiler`

stoksc: I just don't like shadowing package names

armandmcqueen: That's fair

harness/determined/profiler.py
armandmcqueen: No worries, it's good to take a look earlier. 

The main concern I had is that in order to do timing-based flushing, I think we probably need to do it in a separate thread of control - otherwise we have the same issue as the 5 minute timeout. It's not hard, but the more I thought about it, the more I think that TimingBatcher needs its own thread.

harness/determined/pytorch/_pytorch_trial.py
stoksc: I'll leave it for now, but totally open to changing all of this.

harness/determined/profiler.py
stoksc: this line is just because I don't have a `clear()`. calling `convert_to_post_format` consumes the batch. I can make that more explicit. I gotta change this to use all the API code now that I rebased it in so I was waiting on that

stoksc:
to be upstreamed outside of this pr

PR #2200, opened 2021-04-09
harness metric collection improvements

armandmcqueen:
## Description

Mostly renames and applying feedback from previous PR. 

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -348,11 +348,10 @@ class ProfilerSenderThread(threading.Thread):
     def __init__(self, inbound_queue: queue.Queue, master_url: str) -> None:
         self.master_url = master_url
         self.inbound_queue = inbound_queue
-        self.control_queue = queue.Queue()
         super().__init__(daemon=True)
 
     def kill(self):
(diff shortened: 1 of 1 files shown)


stoksc (approved):
looks a solid improvement over the feature branch. I can try to wire the timing stuff into all this and get it all running.

harness/determined/profiler.py
```diff
@@ -297,81 +310,64 @@ def run(self) -> None:
...
 # This is a thread that exists solely so that we can make API calls without blocking
 # It has a Queue through which work is sent to the thread
 class ProfilerSenderThread(threading.Thread):
-    POLL_INTERVAL_SECS = 0.5
     def __init__(self, inbound_queue: queue.Queue, master_url: str) -> None:
         self.master_url = master_url
         self.inbound_queue = inbound_queue
-        self.quitting = False
-        super().__init__()
+        self.control_queue = queue.Queue()
```
stoksc, line 351: where are we reading off this control queue, am I missing it?

armandmcqueen: good catch, the inbound queue also serves as the control so we don't need that extra queue.

PR #2203, opened 2021-04-09
chore: avoid unnecessary re-rendering of experiment list

hkang1:
## Description

Experiment list re-renders every time we poll even if no data has changed. Do a quick simple dirty check to avoid this behavior.

## Checklist


apizzini (approved):
great 👍

PR #2204, opened 2021-04-09
chore: stop using cloudpickle to write PyTorch checkpoints [DET-5175]

mackrorysd:
## Description

cloudpickle explicitly says not to use it for long-term storage and we know it has problems across Python versions (like 3.6/7->3.8). It was initially added in connection with MLFlow stuff and is not believed to be needed now (as opposed to standard pickle).

## Test Plan

I already ran our GPU post-commit tests against this: https://app.circleci.com/pipelines/github/determined-ai/determined?branch=mackrory

I also ran the cifar10 model in Pytorch, then modified the Docker image to remove cloudpickle, patched the experiment to apply this change to harness, used the previous trial as the searcher.source_trial_id, and confirmed that training was able to pick up where it left off.

### DIFF
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -4,7 +4,6 @@ import random
 from abc import abstractmethod
 from typing import Any, Dict, List, Optional, Tuple, Union, cast
 
-import cloudpickle
 import numpy as np
 import torch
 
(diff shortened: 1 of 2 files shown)


mackrorysd:
The test-intg failure is not believed to be related to this change. I've reverted it locally and reproduced it from a clean environment. It is believed to be a grpc streaming failure. A fix is in the works.

rb-determined-ai (approved):

PR #2208, opened 2021-04-11
chore: Update to Ubuntu 20.04 for agent, master and bastion images [DET-5238]

mackrorysd:
## Description

Ubuntu 16.04 is about to be EOL so we need to upgrade now. Moving all the way to Ubuntu 20.04 to support Ubuntu 20.04 Docker images safely and because there's no compelling reason to not update further than 18.04...

## Test Plan

Ran all GPU and nightly tests. Only failure was unrelated and exists on master (test_hp_importance has no output for 10 minutes - which is expected and needs to be worked around). Doing `det deploy` to AWS with the secure deployment type fails - there's still a need to see why, likely the Bastion host.

### DIFF
--- a/harness/determined/deploy/aws/templates/secure.yaml
+++ b/harness/determined/deploy/aws/templates/secure.yaml
@@ -3,46 +3,46 @@ Description:  This template deploys a VPC, with a public and private subnet. It 
 Mappings:
   RegionMap:
     ap-northeast-1:
-      Master: ami-005021e678ab162ea
-      Agent: ami-039e09c6582ec61c5
-      Bastion: ami-04fe60822d08387bb
+      Master: ami-09ff2b6ef00accc2e
+      Agent: ami-0aa0009aef18b0bf6
+      Bastion: ami-09ff2b6ef00accc2e
     # TODO(DET-4258) Uncomment these when we fully support all P3 regions.
     # ap-northeast-2:
-    #   Master: ami-090b3f1f06a12d28d
-    #   Agent: ami-063ee28bcaee0c2b1
-    #   Bastion: ami-05f375b54be4ab849
+    #   Master: ami-0b329fb1f17558744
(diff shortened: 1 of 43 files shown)


stoksc (approved):
lgtm, what are we going to do about the secure deployment? seems like we should make a ticket and try to fix before release but get this in to give it some time to bake.

harness/determined/deploy/aws/templates/secure.yaml
```diff
@@ -3,46 +3,46 @@ Description:  This template deploys a VPC, with a public and private subnet. It
 Mappings:
   RegionMap:
     ap-northeast-1:
-      Master: ami-005021e678ab162ea
-      Agent: ami-039e09c6582ec61c5
-      Bastion: ami-04fe60822d08387bb
+      Master: ami-09ff2b6ef00accc2e
+      Agent: ami-0aa0009aef18b0bf6
+      Bastion: ami-09ff2b6ef00accc2e
```
stoksc, line 8: when i ran `./refresh-ubuntu-amis.py bumpenvs.yaml` the bastion ami's didn't refresh. not sure what your run did differently.

mackrorysd:
> lgtm, what are we going to do about the secure deployment? seems like we should make a ticket and try to fix before release but get this in to give it some time to bake.

I think I'm good with that. It's not isolated to the Bastion host, though - the VPC deployment also fails. I'm digging into it now.

harness/determined/deploy/aws/templates/secure.yaml
stoksc: > I updated the release line of Ubuntu

Where is this (since I did run bump envs with your changes)?

stoksc: oh i see in tools/scripts

PR #2211, opened 2021-04-12
fix: TFKerasTrial on tf2 with tf.compat.v1.disable_v2_behavior.

ioga:
## Checklist

### DIFF
--- a/harness/determined/keras/_tf_keras_trial.py
+++ b/harness/determined/keras/_tf_keras_trial.py
@@ -567,7 +567,7 @@ class TFKerasTrialController(det.LoopTrialController):
 
         # Save RNG state.
         rng_state = {"np_rng_state": np.random.get_state(), "random_rng_state": random.getstate()}
-        if version.parse(tf.__version__) < version.parse("2.0.0"):
+        if not (version.parse(tf.__version__) >= version.parse("2.0.0") and tf._tf2.enabled()):
             rng_state["tf_rng_global_seed"] = tf.compat.v1.random.get_seed(0)[0]
         else:
             generator = tf.random.get_global_generator()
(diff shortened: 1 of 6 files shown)


harness/determined/keras/_tf_keras_trial.py
```diff
@@ -567,7 +567,7 @@ def _save_checkpoint(self, path: pathlib.Path) -> workload.Response:
 
         # Save RNG state.
         rng_state = {"np_rng_state": np.random.get_state(), "random_rng_state": random.getstate()}
-        if version.parse(tf.__version__) < version.parse("2.0.0"):
+        if not (version.parse(tf.__version__) >= version.parse("2.0.0") and tf._tf2.enabled()):
```
rb-determined-ai, line 570: Why convert the  `<` into a `not >=`, out of curiosity?

rb-determined-ai (commented):
What about the other cases where `version.parse()` is used in out `keras` module?

harness/determined/keras/_tf_keras_trial.py
ioga: it may be matter of taste, but in my opinion, `NOT(version >= 2 AND version2-specific-condition)` is more readable, because inside of NOT you have a standard pattern of checking for something specific for the new version. 
This makes use of short circuiting (otherwise `tf._tf2` will error out), and short circuiting based on OR's may be harder to read fast.

rb-determined-ai:
What about the other cases where version.parse() is used in out keras module?

ioga:
I didn't change other cases of version.parse() cause it's not clear they need this particular check based only on the code and in the absence particular bugs. 
But I've discovered that the RNG state management in keras and estimator was done inconsistently, so I moved it out for reuse, started to check for eager execution because it matters, and added a test for this bug.

rb-determined-ai (approved):
Looks great.  Tried to break it but couldn't.

PR #2214, opened 2021-04-13
feat: per-resource-pool configs [DET-5173]

rb-determined-ai:
## Description

Add new configuration options in the experiment config and master
config.

Experiment Config:

* Add resources.devices, a list of --device settings for docker run
  (ignored by kubernetes resource manager).
* Add environment.add_capabilities, a list of --cap-add settings for
  docker run (ignored by kubernetes resource manager).
* Add environment.drop_capabilities, a list of --cap-drop settings for
  docker run (ignored by kubernetes resource manager).

Master Config:

* Add task_container_defaults.add_capabilities
* Add task_container_defaults.drop_capabilities
* Add task_container_defaults.devices
* Support a new task_container_defaults field each of the entries of
  resource_pools.  The task_container_defaults applied to a
  experiment/notebook/shell/command is matching resource_pool's
  task_container_defaults, or the top-level task_container_defaults if
  the per-resource_pool setting is not defined.

## Test Plan

Tested manually.

## Commentary

This PR includes a refactor of how the `master.taskSpec` is handled throughout the master.  Previously it was passed around to different API endpoint handlers, which would derive their own TaskSpec from the master's TaskSpec.  Now, it is necessary to construct the TaskSpec on the fly (due to having multiple task_container_defaults) and so the master passes around a function which takes as input a resource pool name and a number of slots and returns an appropriate TaskSpec.

### DIFF
--- a/docs/reference/cluster-config.txt
+++ b/docs/reference/cluster-config.txt
@@ -296,8 +296,8 @@ The master supports the following configuration settings:
       -  ``serveraddress`` (optional)
       -  ``email`` (optional)
 
-   -  ``add_capabilities``: The default list of capability strings to
-      pass to the Docker daemon. Ignored by resource managers of type
+   -  ``add_capabilities``: The default list of Linux capabilities to
+      grant to task containers. Ignored by resource managers of type
       ``kubernetes``. See :ref:`environment.add_capabilities
       <exp-environment-add-capapbilities>` for more details.
 
--- a/docs/reference/cluster-config.txt
+++ b/docs/reference/cluster-config.txt
@@ -407,8 +407,6 @@ The master supports the following configuration settings:
       -  ``master_service_name``: The service account Determined uses to
          interact with the Kubernetes API.
 
-.. _resource-pool-config:
-
 -  ``resource_pools``: A list of resource pools. A resource pool is a
    collection of identical computational resources. Users can specify
    which resource pool a job should be assigned to when the job is
--- a/docs/reference/cluster-config.txt
+++ b/docs/reference/cluster-config.txt
@@ -213,6 +213,8 @@ The master supports the following configuration settings:
 -  ``port``: The TCP port on which the master accepts all incoming
    connections. Defaults to ``8080``.
 
+.. _master-task-container-defaults:
+
 -  ``task_container_defaults``: Specifies Docker defaults for all task
    containers. A task represents a single schedulable unit, such as a
    trial, command, or tensorboard.
@@ -293,6 +295,18 @@ The master supports the following configuration settings:
       -  ``password`` (required)
       -  ``serveraddress`` (optional)
       -  ``email`` (optional)
+
+   -  ``add_capabilities``: The default list of capability strings to
+      pass to the Docker daemon. Ignored by resource managers of type
+      ``kubernetes``. See :ref:`environment.add_capabilities
+      <exp-environment-add-capapbilities>` for more details.
+
+   -  ``drop_capabilities``: Just like ``add_capabilities`` but for
+      dropping capabilities.
+
+   -  ``devices``: The default list of devices to pass to the Docker
+      daemon. Ignored by resource managers of type ``kubernetes``. See
+      :ref:`resources.devices <exp-resources-devices>` for more details.
 
 -  ``root``: Specifies the root directory of the state files. Defaults
    to ``/usr/share/determined/master``.
@@ -393,6 +407,8 @@ The master supports the following configuration settings:
       -  ``master_service_name``: The service account Determined uses to
          interact with the Kubernetes API.
 
+.. _resource-pool-config:
+
 -  ``resource_pools``: A list of resource pools. A resource pool is a
    collection of identical computational resources. Users can specify
    which resource pool a job should be assigned to when the job is
@@ -406,6 +422,13 @@ The master supports the following configuration settings:
 
    -  ``max_cpu_containers_per_agent``: The maximum number of CPU-only
       containers that can be scheduled on each agent in this pool.
+
+   -  ``task_container_defaults``: Each resource pool may specify a
+      ``task_container_defaults`` that overrides the :ref:`top-level
+      setting <master-task-container-defaults>` for all tasks launched
+      in that resource pool. There is no merging behavior; when a
+      resource pool's ``task_container_defaults`` is set, tasks launched
+      in that pool will completely ignore the top-level setting.
 
    -  ``scheduler``: Specifies how Determined schedules tasks to agents.
       The scheduler configuration on each resource pool will override
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -100,16 +100,24 @@ type (
 // Create a new experiment object from the given model experiment object, along with its searcher
 // and log. If the input object has no ID set, also create a new experiment in the database and set
 // the returned object's ID appropriately.
-func newExperiment(master *Master, expModel *model.Experiment) (*experiment, error) {
+func newExperiment(master *Master, expModel *model.Experiment, taskSpec *tasks.TaskSpec) (
+	*experiment, error,
+) {
 	conf := expModel.Config
 
+	// Validate the ResourcePool setting.  The reason to do it now and not in postExperiment like
+	// all the other validations is that the resource pool should be revalidated every time the
+	// master restarts.
 	if err := sproto.ValidateRP(master.system, conf.Resources.ResourcePool); err != nil {
 		return nil, err
 	}
-
 	// If the resource pool isn't set, fill in the default.
 	if expModel.Config.Resources.ResourcePool == "" {
-		expModel.Config.Resources.ResourcePool = sproto.GetDefaultGPUResourcePool(master.system)
+		if expModel.Config.Resources.SlotsPerTrial == 0 {
+			expModel.Config.Resources.ResourcePool = sproto.GetDefaultCPUResourcePool(master.system)
+		} else {
+			expModel.Config.Resources.ResourcePool = sproto.GetDefaultGPUResourcePool(master.system)
+		}
 	}
 
 	method := searcher.NewSearchMethod(conf.Searcher)
@@ -162,7 +170,7 @@ func newExperiment(master *Master, expModel *model.Experiment) (*experiment, err
 		warmStartCheckpoint: checkpoint,
 
 		agentUserGroup: agentUserGroup,
-		taskSpec:       master.taskSpec,
+		taskSpec:       taskSpec,
 
 		faultToleranceEnabled: true,
 	}, nil
--- a/master/internal/resourcemanagers/config.go
+++ b/master/internal/resourcemanagers/config.go
@@ -26,7 +26,7 @@ func (r *ResourceConfig) ResolveResource() error {
 		r.ResourceManager.AgentRM = &AgentResourceManagerConfig{}
 	}
 	if r.ResourceManager.AgentRM != nil && r.ResourcePools == nil {
-		defaultPool := *defaultRPConfig()
+		defaultPool := defaultRPConfig()
 		defaultPool.PoolName = defaultResourcePoolName
 		r.ResourcePools = []ResourcePoolConfig{defaultPool}
 	}
--- a/master/internal/resourcemanagers/config.go
+++ b/master/internal/resourcemanagers/config.go
@@ -26,12 +26,9 @@ func (r *ResourceConfig) ResolveResource() error {
 		r.ResourceManager.AgentRM = &AgentResourceManagerConfig{}
 	}
 	if r.ResourceManager.AgentRM != nil && r.ResourcePools == nil {
-		r.ResourcePools = []ResourcePoolConfig{
-			{
-				PoolName:                 defaultResourcePoolName,
-				MaxCPUContainersPerAgent: 100,
-			},
-		}
+		defaultPool := *defaultRPConfig()
+		defaultPool.PoolName = defaultResourcePoolName
+		r.ResourcePools = []ResourcePoolConfig{defaultPool}
 	}
 	return nil
 }
--- a/master/pkg/model/environment_config.go
+++ b/master/pkg/model/environment_config.go
@@ -31,6 +31,10 @@ type Environment struct {
 	RegistryAuth   *types.AuthConfig `json:"registry_auth,omitempty"`
 	ForcePullImage bool              `json:"force_pull_image"`
 	PodSpec        *k8sV1.Pod        `json:"pod_spec"`
+
+	// omitempty since they are not officially announced features yet
+	AddCapabilities  *[]string `json:"add_capabilities,omitempty"`
+	DropCapabilities *[]string `json:"drop_capabilities,omitempty"`
 }
 
 // RuntimeItem configures the runtime image.
--- a/master/pkg/model/environment_config.go
+++ b/master/pkg/model/environment_config.go
@@ -32,9 +32,8 @@ type Environment struct {
 	ForcePullImage bool              `json:"force_pull_image"`
 	PodSpec        *k8sV1.Pod        `json:"pod_spec"`
 
-	// omitempty since they are not officially announced features yet
-	AddCapabilities  *[]string `json:"add_capabilities,omitempty"`
-	DropCapabilities *[]string `json:"drop_capabilities,omitempty"`
+	AddCapabilities  []string `json:"add_capabilities"`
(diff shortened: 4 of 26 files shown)


master/internal/resourcemanagers/config.go
```diff
@@ -26,12 +26,9 @@ func (r *ResourceConfig) ResolveResource() error {
...
 	if r.ResourceManager.AgentRM != nil && r.ResourcePools == nil {
-		r.ResourcePools = []ResourcePoolConfig{
-			{
-				PoolName:                 defaultResourcePoolName,
-				MaxCPUContainersPerAgent: 100,
-			},
-		}
+		defaultPool := *defaultRPConfig()
+		defaultPool.PoolName = defaultResourcePoolName
+		r.ResourcePools = []ResourcePoolConfig{defaultPool}
```
stoksc, line 31: nit but seems either `defaultRPConfig` should not return a pointer or `r.ResourcePools` should be a list pointer.

master/pkg/model/task_container_defaults.go
```diff
@@ -26,6 +27,11 @@ type TaskContainerDefaultsConfig struct {
 	Image                  *RuntimeItem          `json:"image,omitempty"`
 	RegistryAuth           *types.AuthConfig     `json:"registry_auth,omitempty"`
 	ForcePullImage         bool                  `json:"force_pull_image,omitempty"`
+
+	// omitempty since they are not officially announced features yet
+	AddCapabilities  *[]string      `json:"add_capabilities,omitempty"`
+	DropCapabilities *[]string      `json:"drop_capabilities,omitempty"`
```
stoksc, line 33: Plain old nil slices will get omitted so you don't need `*[]string`.

master/pkg/tasks/task.go
```diff
@@ -83,6 +83,28 @@ func ToContainerSpec(t TaskSpec) container.Spec {
...
+
+	var capDrop []string
+	if env.DropCapabilities != nil {
+		capDrop = *env.DropCapabilities
+	}
+
+	resources := t.ResourcesConfig()
+	var devices []docker.DeviceMapping
+	if resources.Devices != nil {
+		for _, device := range *resources.Devices {
```
stoksc, line 99: i think just this will work fine unless i'm missing something
```suggestion
	for _, device := range resources.Devices {
```

master/pkg/tasks/task.go
```diff
@@ -83,6 +83,28 @@ func ToContainerSpec(t TaskSpec) container.Spec {
 		shmSize = t.TaskContainerDefaults.ShmSizeBytes
 	}
 
+	var capAdd []string
+	if env.AddCapabilities != nil {
+		capAdd = *env.AddCapabilities
+	}
```
stoksc, line 89: Said this earlier but I think `env.AddCapabilities` can just be a `[]string` and directly placed here (in both cases `capAdd` would end up being a nil string slice).

master/internal/experiment.go
```diff
@@ -100,16 +100,24 @@ type (
...
 	}
-
 	// If the resource pool isn't set, fill in the default.
 	if expModel.Config.Resources.ResourcePool == "" {
-		expModel.Config.Resources.ResourcePool = sproto.GetDefaultGPUResourcePool(master.system)
+		if expModel.Config.Resources.SlotsPerTrial == 0 {
+			expModel.Config.Resources.ResourcePool = sproto.GetDefaultCPUResourcePool(master.system)
+		} else {
+			expModel.Config.Resources.ResourcePool = sproto.GetDefaultGPUResourcePool(master.system)
+		}
```
stoksc, line 120: This change seems outside of the scope of this PR, but probably worth discussing? I imagine @armandmcqueen would know why it's not this way already (I've wondered this myself)?

master/pkg/model/environment_config.go
```diff
@@ -31,6 +31,10 @@ type Environment struct {
 	RegistryAuth   *types.AuthConfig `json:"registry_auth,omitempty"`
 	ForcePullImage bool              `json:"force_pull_image"`
 	PodSpec        *k8sV1.Pod        `json:"pod_spec"`
+
+	// omitempty since they are not officially announced features yet
+	AddCapabilities  *[]string `json:"add_capabilities,omitempty"`
```
stoksc, line 36: Same here with pointers to []string being unnecessary.

master/pkg/model/experiment_config.go
```diff
@@ -184,6 +186,33 @@ func (l *Labels) UnmarshalJSON(data []byte) error {
 	return err
 }
 
+// DevicesConfig is the configuration for bind mounts.
```
stoksc, line 189: Update comment?

master/pkg/model/experiment_config.go
```diff
@@ -184,6 +186,33 @@ func (l *Labels) UnmarshalJSON(data []byte) error {
...
+// DeviceConfig configures trial runner filesystem bind mounts.
+type DeviceConfig struct {
+	HostPath      string `json:"host_path"`
+	ContainerPath string `json:"container_path"`
+	Mode          string `json:"mode"`
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (d *DeviceConfig) UnmarshalJSON(data []byte) error {
+	d.Mode = "mrw"
```
stoksc, line 211: obviously with all of this, it would be nice if there was a separate defaulting step but I think we can cross that bridge when we get there. this works just fine as far as i'm concerned for now.

master/internal/experiment.go
stoksc: Oh wait, wouldn't `expModel.Config.Resources.SlotsPerTrial` be defaulted to 1?

master/pkg/model/experiment_config.go
```diff
@@ -197,6 +226,32 @@ type ResourcesConfig struct {
 	AgentLabel     string  `json:"agent_label"`
 	ResourcePool   string  `json:"resource_pool"`
 	Priority       *int    `json:"priority,omitempty"`
+
+	// omitempty since is not an officially announced feature yet
+	Devices *DevicesConfig `json:"devices,omitempty"`
```
stoksc, line 231: ditto on the empty array thing.

master/pkg/model/experiment_config.go
```diff
@@ -197,6 +226,32 @@ type ResourcesConfig struct {
...
+}
+
+// ParseJustResources is a helper function for breaking the circular dependency where we need the
+// TaskContainerDefaults to unmarshal an ExperimentConfig, but we need the Resources.ResourcePool
+// setting to know which TaskContainerDefaults to use.  It does not throw errors; if unmarshalling
```
stoksc, line 236: I'd rather it throw errors and the caller discard as necessary.

stoksc (commented):
just had mostly comments about go stuff, the structure makes sense and i like that you just passed a fn around instead of some crazy actor nonsense. still don't love how the defaulting is happening, want to take some time to think on it.

master/pkg/model/task_container_defaults.go
```diff
@@ -60,8 +66,28 @@ func validatePortRange(portRange string) []error {
...
 }
 
+// UnmarshalJSON implements the json.Unmarshaler interface.
+// Setting defaults here is necessary over our usual "Define a default struct and unmarshal into it"
+// strategy because there are places (resource pool configs) where we need to know if the task
+// container defaults were set at all or if they were not; if they were set then that resource
+// pool's task container defaults are used instead of the toplevel master config's settings.  To
+// know if the user set them at the resource pool level, the resource pool has to have a nullable
+// pointer, which is not compatible with our usual strategy for defaults.
+func (c *TaskContainerDefaultsConfig) UnmarshalJSON(data []byte) error {
```
stoksc, line 76: Damn it.. I left a really long comment on this but I didn't "add it" and when I submitted the review it went away but I'll try again..

So to be clear: this is defaulted here because you need to distinguish between not provided resource pool task container defaults and provided ones, but you need to default the provided ones that are missing these fields still? Took me a bit to grok that a resource pool config with nil task containter default json would never unmarshal into it's task container defaults and thus stay nil.. am I even getting it now lol? i want to try to think about a better way to do this..

master/internal/experiment.go
rb-determined-ai: Seemed like a pretty clear bug but I'm happy to hear if there was a reason why it was that way.

It does default to 1 but  I'm not sure why that's relevant, since the user-provided value has already been parsed by here.  The apparent bug was that the resource pool setting that is configured for the experiment (but not the one saved in the database) seems to not be correct.  Maybe... is slots_per_trial not allowed to be zero for an experiment?  I seem to recall that being the case...

master/pkg/model/experiment_config.go
rb-determined-ai: I don't think there's another way to do the defaulting for entire lists of objects with our current system.  There will be with expconf though; but this is simply a limitation of the json-unmarshal-into-defaults strategy.

master/pkg/model/task_container_defaults.go
rb-determined-ai: > you need to distinguish between not provided resource pool task container defaults and provided ones, but you need to default the provided ones that are missing these fields still

Yes, exactly.  So the "set the defaults and unmarshal into them" strategy does not work at all.

> Took me a bit to grok that a resource pool config with nil task containter default json would never unmarshal into it's task container defaults and thus stay nil (and not have null pointer panics here)

Yes; the json unmarshalling code is what allocates fields before calling UnmarshalJSON whenever it sees keys in the json blob that match the tagname from `json:"TAGNAME"`.

> i want to try to think about a better way to do this

This thinking and work has already been done; it's in the experiment config rework.  That machinery has the following order:

1. Validate user input, exactly as it is provided
2. For valid inputs, unmarshal into structs.  Nil pointers may be present for fields the user did not provide.
3. Merge any partial configs you'd like to fill gaps into the user-provided config.
4. Call WithDefaults() on the object to fill nil pointers that have defaults
5. When reading data from the config object, use a generated getter whose return type automatically is a pointer only in the case that a field is optional but has no default (for stronger compile-time checks, basically).

So in this case, the master config should be parsed exactly as it was provided.  Defaults on all provided objects are set by WithDefaults().  In this case, the top-level task_container_defaults will always be filled, even if the user does not provide it, but the per-resource-pool task_container_defaults will stay as nil pointers.  Only if the user provided them will WithDefaults() set defaults on them.

This has the additional benefit that this whole mess of trying to pre-unmarshal the experiment config so you can figure out the right task_container_defaults in order to actually unmarshal the experiment config goes away, because in the new system: steps 1 and 2 happen, then you read the resource_pool setting, then you pick which task_container_default you want, then you continue with steps 3-5.

master/pkg/model/experiment_config.go
stoksc: yeah I totally agree. we should do this like expconf one day, but today isn't the day.

master/pkg/model/task_container_defaults.go
stoksc: yep, totally on the same page. I guess what I meant was I don't see a good solution that isn't a hell of a lotta work 😄.

master/internal/resourcemanagers/config.go
rb-determined-ai: I vote no pointer.  I think that our config handling is *very* loose about what things should be pointers vs what things are pointers because something similar to that was a pointer somewhere else.

master/pkg/tasks/task.go
rb-determined-ai: It will, but only after removing the pointer from resources.Devices, I believe.

master/pkg/model/experiment_config.go
rb-determined-ai: done

master/pkg/model/experiment_config.go
rb-determined-ai: Wait... this is still going to be weird because the docstring is basically now, "it always returns a valid ResourcesConfig, even in error cases", which seems equally weird to me.  I actually think it's fine as-is, especially since it's going to go away in a couple of week.

docs/reference/cluster-config.txt
```diff
@@ -294,6 +296,18 @@ The master supports the following configuration settings:
       -  ``serveraddress`` (optional)
       -  ``email`` (optional)
 
+   -  ``add_capabilities``: The default list of capability strings to
+      pass to the Docker daemon. Ignored by resource managers of type
+      ``kubernetes``. See :ref:`environment.add_capabilities
```
stoksc, line 301: curious, how would this work for k8s? devices and caps have to be added to the containers (as far as i see) so I don't think it would be configurable in the current system, unless we propagated these configurations to the pod spec for the user.

docs/reference/cluster-config.txt
```diff
@@ -294,6 +296,18 @@ The master supports the following configuration settings:
       -  ``serveraddress`` (optional)
       -  ``email`` (optional)
 
+   -  ``add_capabilities``: The default list of capability strings to
+      pass to the Docker daemon. Ignored by resource managers of type
```
stoksc, line 300: Maybe similar rewording elsewhere
```suggestion
   -  ``add_capabilities``: The default list of Linux capabilities to
      grant to task containers. Ignored by resource managers of type
```

master/pkg/tasks/task.go
```diff
@@ -83,6 +83,19 @@ func ToContainerSpec(t TaskSpec) container.Spec {
 		shmSize = t.TaskContainerDefaults.ShmSizeBytes
 	}
 
+	capAdd := env.AddCapabilities
+	capDrop := env.DropCapabilities
```
stoksc, line 87: nit, i dont think this is necessary now (just assign on 118)

master/pkg/model/task_container_defaults.go
```diff
@@ -60,8 +65,28 @@ func validatePortRange(portRange string) []error {
...
+	if err := json.Unmarshal(data, DefaultParser(c)); err != nil {
+		return errors.Wrap(err, "failed to parse task container defaults")
+	}
+	return nil
+}
+
 // Validate implements the check.Validatable interface.
-func (c TaskContainerDefaultsConfig) Validate() []error {
+func (c *TaskContainerDefaultsConfig) Validate() []error {
+	if c == nil {
```
stoksc, line 87: this is kinda funky, i imagine you agree. it's ok if it's nil in some places but in others it means it hasn't been defaulted right? but i guess it those places it couldn't be nil in this call, it would be a `&TaskContainerDefaultsConfig{}`, so it works, kinda feels like it barely correctly handles it in a complex way though. but it's ok for now.

stoksc (approved):
lgtm, there are some rough edges, but I think the UX is fine and it should be workable. most of the issues I have I think we share and are tech debt with master config that can't really be addressed here quickly.

docs/reference/cluster-config.txt
rb-determined-ai: Basically there's [add/drop caps](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container) that are pretty straight-forward.  I believe k8s users could specify those things in the pod spec right now on tip-of-master.  I chose to not propagate these settings into the pod spec because a) I don't like multiple ways to configure the same thing, and b) it's easier to add it later if this is the wrong call than to remove it later.

But I'm open to reversing that decision, if anybody feels strongly.

However, [supporting arbitrary devices](https://github.com/kubernetes/kubernetes/issues/7890#issuecomment-766088805) is like holy-shit-hard.  Way beyond the scope of this work.

master/pkg/model/task_container_defaults.go
rb-determined-ai: Indeed I do agree.

docs/reference/cluster-config.txt
stoksc: > Basically there's add/drop caps that are pretty straight-forward.

Yeah these are specified on the container, so I didn't know how to specify them in the pod spec, wasn't sure how the merging behavior worked if a container spec is in the pod spec.

docs/reference/cluster-config.txt
rb-determined-ai: done; affected experiment config, cluster config, and command/notebook config docs.

master/internal/experiment.go
stoksc: Yeah, slots_per_trial is defaulted to 1 and validated as > 1 for an experiment so it seems this change is just a no-op.

armandmcqueen (approved):
🚀

PR #2217, opened 2021-04-13
ci: allow CI build with linting errors

hkang1:
## Description

With `react-scripts 4.x` the build process will no longer allow linting errors to exist. Introduce a small environment variable to allow it (`ESLINT_NO_DEV_ERRORS=true`).
[stackoverflow details](https://stackoverflow.com/a/64630234/5402432)

## Checklist


hkang1:
The better approach this to set an env `DISABLE_ESLINT_PLUGIN` to be `true` for local builds only. Specifically:
```
DISABLE_ESLINT_PLUGIN=true npm run build
```

PR #2218, opened 2021-04-13
chore: allow dev lint errors

hkang1:
## Description

With CRA `react-scripts` updated to `4.x`, any linting errors encountered in the dev environment will halt the app. Set environment var `ESLINT_NO_DEV_ERRORS=true` to allow dev env to be more forgiving with linting errors.

## Checklist

### DIFF
--- a/webui/react/.gitignore
+++ b/webui/react/.gitignore
@@ -21,3 +21,6 @@ build-storybook/
 node_modules/
 
 report.*.json
+
+# .env with ESLINT_NO_DEV_ERRORS=true allows CRA to build with linting errors
+!.env
--- a/webui/react/.gitignore
+++ b/webui/react/.gitignore
@@ -21,6 +21,3 @@ build-storybook/
 node_modules/
 
 report.*.json
-
-# .env with ESLINT_NO_DEV_ERRORS=true allows CRA to build with linting errors
-!.env
--- a/webui/react/package.json
+++ b/webui/react/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "postshrinkwrap": "npx npm-force-resolutions && npm install --ignore-scripts",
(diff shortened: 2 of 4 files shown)


webui/react/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "postshrinkwrap": "npx npm-force-resolutions && npm install --ignore-scripts",
-    "start": "react-app-rewired start",
+    "start": "ESLINT_NO_DEV_ERRORS=true react-app-rewired start",
```
hamidzr, line 7: could we set this in makefile in a way that is overridable?

hamidzr: probably having it the same way as how `PUBLIC_URL` is set might also let us not need the `.env` file

webui/react/.gitignore
```diff
@@ -21,3 +21,6 @@ build-storybook/
 node_modules/
 
 report.*.json
+
+# .env with ESLINT_NO_DEV_ERRORS=true allows CRA to build with linting errors
+!.env
```
hamidzr, line 26: what do you think about ignoring it or using a non react specific way to set the env vari?

hamidzr:
thanks for addressing this quickly. I added this var to my env and my dev flow is back to how it was before.

webui/react/package.json
hkang1: I was split on that versus baking it in. Chose the latter just because I couldn't think of a case where I would want my dev environment to not run if there are errors, but can't hurt to have it as an option to keep this area consistent.

hamidzr (approved):
thanks @hkang1

webui/react/package.json
hamidzr: > I would want my dev environment to not run if there are errors,

you're right I don't see a use for that either it's just that setting it the old way does not respect the user's wish to set that var. minor difference anyway :)

PR #2222, opened 2021-04-14
fix: nightly mnist_pytorch slots failed to free [DET-5239]

naren-determined:
## Description

The `mnist_pytorch` example has been failing on nightly tests due to the checkpoint GC CPU-only tasks taking longer to terminate than the default time period. To resolve this issue, I bumped the `max_secs_to_free_slots` to a 15 min time period. If the slots free before that time period is up, the loop will break so this should not affect the overall time spent running nightly examples. I want to ensure this fixes the problem in nightly examples, after which I can always reduce the time period to free slots if needed.

## Checklist

### DIFF
--- a/e2e_tests/tests/experiment/experiment.py
+++ b/e2e_tests/tests/experiment/experiment.py
@@ -460,7 +460,7 @@ def verify_completed_experiment_metadata(
     # When the experiment completes, all slots should now be free. This
     # requires terminating the experiment's last container, which might
     # take some time.
-    max_secs_to_free_slots = 30
+    max_secs_to_free_slots = 900
     for _ in range(max_secs_to_free_slots):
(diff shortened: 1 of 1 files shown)


e2e_tests/tests/experiment/experiment.py
```diff
@@ -460,7 +460,7 @@ def verify_completed_experiment_metadata(
     # When the experiment completes, all slots should now be free. This
     # requires terminating the experiment's last container, which might
     # take some time.
-    max_secs_to_free_slots = 30
+    max_secs_to_free_slots = 900
```
katport, line 463: Q: 30s -> 15min is pretty big gap. How was this time decided on?

naren-determined:
Bug in GC Checkpoints solved this issue.

PR #2223, opened 2021-04-14
chore: validate grid list enum value from local storage [DET-5187]

hkang1:
## Description

Fix the issue of cluster page not having a Grid/List view selected (neither are selected).

The core issue is that the enum value has changed from numeric to string, and if the user has a numeric value saved previously, neither will get selected.

## Checklist


hamidzr (approved):
LGTM. this is a common issue that we could frequently trigger when serializing and persisting values but then deprecating the path to deserialize them. let's add a sync topic to see what we can do about these at a higher level.

PR #2224, opened 2021-04-14
fix: gallery modal [DET-5257]

hkang1:
## Description

Problem: Navigation (previous chart / next chart) for gallery modal view on HP Scatter Plots and HP Heatmap do not work on the very first load.
Technical issue: The problem is that the very initial size of the chart is 0 in width and height and the re-render ignores when the width and height of the resize event are 0. It's actually not 0 so change to dynamically getting the width and height to determine if the size of the chart is actually 0.

## Checklist


hamidzr (approved):
🚀

PR #2226, opened 2021-04-14
fix: correct crash when changing filter on multi-selected rows [DET-5258]

hkang1:
## Description

Problem: WebUI crashes when user selects a rows in multi-selectable tables and then changing filters on the table.
Solution: Reset the selected rows when filter changes.

## Test Plan

- [ ] on experiment list page, select multiple rows then change any filter (user, label, archived, etc)
- [ ] on task list page, select multiple rows then change any filter (user, label, archived, etc)

## Checklist

### DIFF
--- a/docs/release-notes/2226-fix-user-filter-crash.txt
+++ b/docs/release-notes/2226-fix-user-filter-crash.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  WebUI: Fix the issue of the WebUI crashing when user selects a row in
+   experiment list page then changes the user filter.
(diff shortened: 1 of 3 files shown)


webui/react/.gitignore
```diff
@@ -21,3 +21,6 @@ build-storybook/
 node_modules/
 
 report.*.json
+
+# .env with ESLINT_NO_DEV_ERRORS=true allows CRA to build with linting errors
+!.env
```
hamidzr, line 26: is this intentional? these are already being worked on another pr of yours.

hamidzr (approved):
LGTM. minus the react/.env changes comment

webui/react/.gitignore
hkang1: oops need to get rid of this

PR #2225, opened 2021-04-14
docs: Release notes for 0.15.0.

justin-determined-ai:
## Commentary (optional)

Release notes for 0.14.6 were never merged back into master, so this PR includes both. Will be fixed post-release.

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234

PR #2227, opened 2021-04-14
docs: add release note for removing cloudpickle

mackrorysd:
## Description

Adding release note about removing cloudpickle from checkpoints. Confident that it should have no impact because I tested loading an old checkpoint and continuing to train. Also, we weren't using cloudpickle to load checkpoints anyway, and there is no logic in PyTorch to dynamically decide cloudpickle should be used for loading.

## Test Plan

Pre-commits should be sufficient

### DIFF
--- a/docs/release-notes/2204-stop-using-cloudpickle.txt
+++ b/docs/release-notes/2204-stop-using-cloudpickle.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**Improvements**
+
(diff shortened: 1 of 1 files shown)


vishnu2kmohan (approved):
Thanks @mackrorysd!

mackrorysd:
Pulled into https://github.com/determined-ai/determined/pull/2225

PR #2228, opened 2021-04-14
fix: fetch agents only when authenticated [DET-5259]

hkang1:
## Description

Problem: Agent API fetch is happening in the navigation component even when the user is signed out.
Solution: Check to make sure the user is authenticated before making the agent fetch.

## Checklist


justin-determined-ai (approved):

webui/react/src/pages/Cluster/ClusterOverview.tsx
```diff
@@ -107,8 +112,10 @@ const ClusterOverview: React.FC = () => {
   }, []);
 
   useEffect(() => {
+    fetchAgents();
```
hkang1, line 115: agent fetch is already being polled at the top level of Navigation, but it makes sense for it to get called at lease during a mount for cluster overview, otherwise the info can be pretty out of sync right off the bat.

PR #2229, opened 2021-04-14
fix: force checkpoint GC to use the master's default environment [DET-5260]

stoksc:
## Description
This change updates the checkpoint GC task to use either the default `environment.image` or the defaults from task container defaults if they were defined.

## Test Plan
- [x] submit exp with custom image, make sure it doesn't use it.

## Commentary (optional)
This uses the task spec from the experiment, which is the task spec from master, which contains the master's task container defaults, not the task container defaults for the resource pool this got scheduled on. Maybe that is a shortcoming, but it probably will work fine.

## Checklist

### DIFF
--- a/master/pkg/tasks/task_spec.go
+++ b/master/pkg/tasks/task_spec.go
@@ -237,8 +237,10 @@ func (g GCCheckpoints) Entrypoint() []string {
 }
 
 // Environment implements InnerSpec.
-func (g GCCheckpoints) Environment(TaskSpec) model.Environment {
-	return g.ExperimentConfig.Environment
+func (g GCCheckpoints) Environment(t TaskSpec) model.Environment {
+	env := model.DefaultExperimentConfig(&t.TaskContainerDefaults).Environment
+	env.EnvironmentVariables = g.ExperimentConfig.Environment.EnvironmentVariables
+	return env
 }
 
 // EnvVars implements InnerSpec.


master/pkg/tasks/task_spec.go
```diff
@@ -238,7 +239,9 @@ func (g GCCheckpoints) Entrypoint() []string {
 
 // Environment implements InnerSpec.
 func (g GCCheckpoints) Environment(TaskSpec) model.Environment {
-	return g.ExperimentConfig.Environment
+	env := model.DefaultExperimentConfig(&g.TaskContainerDefaults).Environment
+	env.EnvironmentVariables = g.ExperimentConfig.Environment.EnvironmentVariables
```
stoksc, line 243: Copied the environment vars of the experiment in the off chance someone is relying on `GOOGLE_APPLICATION_CREDENTIALS` being in there or something.

master/pkg/tasks/task_spec.go
```diff
@@ -238,7 +239,9 @@ func (g GCCheckpoints) Entrypoint() []string {
 
 // Environment implements InnerSpec.
 func (g GCCheckpoints) Environment(TaskSpec) model.Environment {
-	return g.ExperimentConfig.Environment
+	env := model.DefaultExperimentConfig(&g.TaskContainerDefaults).Environment
```
stoksc, line 242: Uses `TaskContainerDefaults.Image` if it's there, otherwise our default images, in the event of an airgapped environment with task container defaults set.

dzhu (approved):
I have a vague fear there's going to be some weird edge case, but this is probably better than before, anyway.

PR #2230, opened 2021-04-14
chore: fix panics in hp importance actor [DET-5263]

stoksc:
## Description
Only send `hpimportance.ExperimentCompleted` messages when the experiment is completed.

## Test Plan
- [ ] kill an experiment real fast and make sure there is no panic.

## Commentary (optional)
Could also check the progress on the receiving side is greater than some threshold or 100% but shouldn't rely on that IMO/I'm not sure exactly what that would be.

## Checklist

### DIFF
--- a/master/internal/hpimportance/hpimportance.go
+++ b/master/internal/hpimportance/hpimportance.go
@@ -19,6 +19,8 @@ import (
 	"sort"
 	"strconv"
 
+	"github.com/pkg/errors"
+
 	log "github.com/sirupsen/logrus"
 
 	"github.com/determined-ai/determined/master/pkg/model"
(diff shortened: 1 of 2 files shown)


master/internal/hpimportance/hpimportance.go
```diff
@@ -183,6 +184,10 @@ func parseImportanceOutput(filename string) (map[string]float64, error) {
 func computeHPImportance(data map[int][]model.HPImportanceTrialData,
 	experimentConfig *model.ExperimentConfig, masterConfig HPImportanceConfig,
 	growforest string, workingDir string) (map[string]float64, error) {
+	if len(data) == 0 {
+		return nil, errors.New("not enough data to compute HP importance")
```
stoksc, line 188: may as well fix both sides 🤷

mackrorysd (approved):
Thanks for finding and fixing this, Bradley!

PR #2233, opened 2021-04-15
chore: fix panic from dependency creation race [DET-5271]

stoksc:
## Description
This change fixes a bug where restored experiments would attempt to send an `ExperimentCreated` message to the hp importance actor prior to its creation, resulting in a panic from a send on a nil actor.

## Test Plan
- [x] restore experiment and see no panic

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -191,8 +191,8 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			e.restoreTrialsFromPriorOperations(ctx, e.searcher.TrialOperations)
 		} else {
 			e.processOperations(ctx, e.searcher.TrialOperations, nil)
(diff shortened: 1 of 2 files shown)


master/internal/experiment.go
```diff
@@ -191,8 +191,8 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			e.restoreTrialsFromPriorOperations(ctx, e.searcher.TrialOperations)
 		} else {
 			e.processOperations(ctx, e.searcher.TrialOperations, nil)
+			ctx.Tell(e.hpImportance, hpimportance.ExperimentCreated{ID: e.ID})
```
stoksc, line 194: might as well fix it in both places. only send on creation not restore and create the manager sooner.

mackrorysd (approved):

PR #2234, opened 2021-04-15
chore: replace metric chart with uplot [DET-4303]

apizzini:
## Checklist


webui/react/src/pages/ExperimentDetails.module.scss
```diff
@@ -1,5 +0,0 @@
-.base {
-  .topRow {
-    height: 40rem;
```
apizzini, line 3: @hkang1 I removed those because were preventing the tab content to be scrolled on mobile.

apizzini:
@cla-bot check

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
```diff
@@ -0,0 +1,43 @@
+.tooltip {
```
hkang1, line 1: **nit:** Is it possible to move `Trial X` to be above `Time Elapsed (sec): X`?

Screen Shot 2021-04-19 at 4 18 53 PM

webui/react/src/components/UPlotChart/tooltipsPlugin.ts
```diff
@@ -0,0 +1,131 @@
+import uPlot, { Plugin } from 'uplot';
```
hkang1, line 1: **suggestion:** Would it make sense to filter out data where in a given `x` value is all series values are `null`? This more noticeable when there's only one series such as the screenshot below:
Screen Shot 2021-04-19 at 5 39 25 PM

So if the data input was something like this:
```
[
  [ 0, 100,  200,  300  ], # x-data
  [ 1, null, null, 1    ], # series 1
  [ 2, null, 2,    2    ], # series 2
  [ 3. null, 3,    null ], # series 3
]
```
where the first row is the X data and 2nd ~ 4th rows are series data, it would drop the 2nd column of each of the rows because the data for when X is 100 is all `null`. Basically making it:
```
[
  [ 0, 200,  300  ],
  [ 1, null, 1    ],
  [ 2, 2,    2    ],
  [ 3, 3,    null ],
]
```

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
hkang1: **nit:** This is a weird one but the cursor dashed lines and the vertical cross line on the highlighted datapoint doesn't span all the way down and to the left of the charting area? The top and the right don't have the same problem.

2021-04-19 at 10 56 PM

apizzini: **> nit: Is it possible to move Trial X to be above Time Elapsed (sec): X?**

We should keep in mind that there's the possibility of multiple series with multiple labels to add (one for each value).
What do you think it's the best thing to do to accomodate both this requirements?

apizzini: (i'm gonna reply to the other question on another code line to start another thread).

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
```diff
@@ -0,0 +1,43 @@
...
+  background: rgba(0, 0, 0, 0.65);
+  color: #fff;
```
apizzini, line 3: **> nit: This is a weird one but the cursor dashed lines and the vertical cross line on the highlighted datapoint doesn't span all the way down and to the left of the charting area? The top and the right don't have the same problem. Hopefully it's a simple fix 🤞**

I'm not sure about this. The issue seems to be that uplot doesn't draw X and Y "borders" and I don't see an option to do that. But if that's needed we can simply add a border-bottom and border-left on a div that uplot draws.

As an example and to show a demo of a possible result see the black lines here:
Schermata 2021-04-20 alle 14 45 50

webui/react/src/components/UPlotChart/tooltipsPlugin.ts
apizzini: Consider that we're hiding series, so if you have a data array like this and the 1st serie is hidden, that would not work.

```
[
  [ 0, 100,  200,  300  ], # x-data
  [ 1, 1, null, 1    ], # series 1 (hidden)
  [ 2, null, 2,    2    ], # series 2
  [ 3. null, 3,    null ], # series 3
]
```

Possible alternative is to avoid displaying the toolltip and line if all the values are null?

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
leeoniya: > But if that's needed we can simply add a border-bottom and border-left on a div that uplot draws.

use a hard `box-shadow` instead, since that does not affect the position and will avoid slightly offseting the cursor coordinate calculations.

webui/react/src/components/UPlotChart/tooltipsPlugin.ts
hkang1: > Possible alternative is to avoid displaying the toolltip and line if all the values are null?

Thanks, this is a much better solution LOL, yes please do this if doable. 🙏

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
hkang1: thanks @leeoniya!

webui/react/src/components/UPlotChart/tooltipsPlugin.module.scss
hkang1: > What do you think it's the best thing to do to accomodate both this requirements?

Ah I see. Thoughts on supporting both cases (single series vs multi-series) a little differently?
Single Series:
```
[series label]
[x label: x data point]
(color) [y label: y data point]
```
Multi-Series:
```
[x-label: x data point]
(color1) [y1 series label]
              [y1 label: y1 data point]
(color2) [y2 series label]
              [y2 label: y2 data point]
(color3) [y3 series label]
              [y3 label: y3 data point]
...
```

hkang1 (approved):
Awesome work @apizzini ! The charts look great.

One final non-blocking chore before merging, could you update the colors and styles to use css variables from the theme? Rough match would be good enough 🙏

PR #2238, opened 2021-04-15
chore: no pointers to maps or slices in expconf

rb-determined-ai:
## Description

When I first designed the golang framework for expconf, I didn't actually realize that maps and arrays could be `nil`, so I was using redundant pointers to those objects.  This PR removes that.

## Test Plan

Added relevant objects to the test structs and tests still pass.

## Commentary

The only part of this change that I think is a bit opaque is the hardcoded whitelist of type aliases in gen.py.  To help improve the discoverability of the whitelist, I wrote very clear error messages when the dev might need to add an element to it.

### DIFF
--- a/harness/determined/common/schemas/expconf/_gen.py
+++ b/harness/determined/common/schemas/expconf/_gen.py
@@ -2675,6 +2675,26 @@ schemas = {
                 "null"
             ],
             "default": null
+        },
+        "defaulted_array": {
+            "type": [
+                "array",
+                "null"
+            ],
+            "default": [],
+            "items": {
+                "type": "string"
+            }
+        },
+        "nodefault_array": {
+            "type": [
+                "array",
+                "null"
+            ],
+            "default": null,
+            "items": {
+                "type": "string"
+            }
         }
     }
 }
--- a/master/pkg/schemas/defaults.go
+++ b/master/pkg/schemas/defaults.go
@@ -45,52 +45,55 @@ func getDefaultSource(obj reflect.Value) interface{} {
 	return nil
 }
 
+func allocateWithDefaultBytes(typ reflect.Type, defaultBytes []byte) reflect.Value {
+	// json.Unmarshal requires a pointer to work.
+	ptr := reflect.New(typ)
+	err := json.Unmarshal(defaultBytes, ptr.Interface())
+	if err != nil {
+		panic(
+			fmt.Sprintf(
+				"failed to unmarshal defaultBytes into %T: %q: %v",
+				ptr.Interface(),
+				string(defaultBytes),
+				err.Error(),
+			),
+		)
+	}
+	return ptr.Elem()
+}
+
 // withDefaults is the recursive layer under WithDefaults.  withDefaults will return a clean copy
 // of the original value, with defaults set.
 func withDefaults(obj reflect.Value, defaultBytes []byte, name string) reflect.Value {
 	// fmt.Printf("withDefaults on %v (%T)\n", name, obj.Interface())
 
-	// Handle pointers first.
-	if obj.Kind() == reflect.Ptr {
+	// Handle nil values and defaultBytes all in one place.
+	if obj.Kind() == reflect.Interface ||
+		obj.Kind() == reflect.Ptr ||
+		obj.Kind() == reflect.Slice ||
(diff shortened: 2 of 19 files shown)


master/pkg/schemas/defaults.go
```diff
@@ -45,52 +45,55 @@ func getDefaultSource(obj reflect.Value) interface{} {
 	return nil
 }
 
+func allocateWithDefaultBytes(typ reflect.Type, defaultBytes []byte) reflect.Value {
```
stoksc, line 48: I'm beginning to think this should all return errors not panic. Yes, a panic means bad code, but that means bad code in a bad spot (not in an actor or http handler, which is basically none of our code but still) is unrecoverable which kinda freaks me out.

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -117,6 +121,9 @@ func (l LabelsV0) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON implements the json.Unmarshaler interface.
 func (l *LabelsV0) UnmarshalJSON(data []byte) error {
```
stoksc, line 123: remind me, why did/do we support a list and map here?

harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -2675,6 +2675,26 @@
                 "null"
             ],
             "default": null
+        },
```
stoksc, line 2678: why do we ever need to differentiate these? sorry coming back to this after not looking at it for a while.

stoksc (commented):
lgtm, just some q's first to help me understand

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -107,6 +108,9 @@ type LabelsV0 map[string]bool
 
 // MarshalJSON implements the json.Marshaler interface.
 func (l LabelsV0) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
rb-determined-ai: No; this is converting a set (`map[string]bool`) into a list.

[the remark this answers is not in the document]
stoksc: oh yeh, i see now.

master/pkg/schemas/defaults.go
rb-determined-ai: I mean, you've been saying that for a while.

I still don't think so though.  These errors can only occur with invalid default values, which are hardcoded json values and should be tested beforehand to be valid.

A better language would execute them at compile time as some sort of const expression.  In this case we can compensate for go with a bunch of tests.

But overall, something that conceptually means "take this struct and fill nil pointers with constant values" should not throw an error for anything other than a malloc failure (in which case Go panics anyway).

I could write a test that checks every default to make sure it is valid json if you'd like?

master/pkg/schemas/expconf/experiment_config.go
rb-determined-ai: I have no idea.  I know it happened on commit `b8c5cdb72b` but I can't figure out why that was what we did.

harness/determined/common/schemas/expconf/_gen.py
rb-determined-ai: Honoring `nil` values is equivalent to knowing if the user provided a value or not.  Trampling them would mean that you couldn't tell if you were looking at a defaulted value of if the user provided e.g. an empty list explicitly.

master/pkg/schemas/defaults.go
stoksc: Nah that's way too much work. We should just be really careful to write tests for every expconf thing?

harness/determined/common/schemas/expconf/_gen.py
stoksc: Makes sense.

master/pkg/schemas/expconf/experiment_config.go
stoksc: Weird. I wonder why it wasn't _always_ a list.

stoksc (approved):

PR #2239, opened 2021-04-15
fix: update label picker when labels change [DET-5254]

hkang1:
## Description

Problem: Label picker becomes out of sync when user edits existing experiment labels on the experiment list page.
Solution: Make another call to fetch updated labels when labels change on the experiment list page.

## Checklist


apizzini (approved):
lgtm 👍 

@hkang1 just out of curiosity, why did choose to fetch labels into pulling instead of labels change? Just to handle the case where labels are changed in other tabs or other users?

hkang1:
> lgtm 👍
> 
> @hkang1 just out of curiosity, why did choose to fetch labels into pulling instead of labels change? Just to handle the case where labels are changed in other tabs or other users?

I thought about avoiding the network call all together and update the existing list of labels. What ultimately veered me away from that approach was that the scope of the label changes are restricted to the one particular experiment. So in the case of where the user deleted a label for one experiment, there's no guarantee that I have all of the info to know that, that particular deleted label no longer exists in the list (if another experiment not shown in the paginated area also has the same label for instance). So another pull was the easiest approach to not have to worry about these specific edge cases.

PR #2241, opened 2021-04-16
ci: fix windows test with lmdb 1.2

rb-determined-ai:
## Description

LMDB 1.2 includes a new build-time dependency, only with Windows.  Add that dependency.

## Test Plan

Tests should pass again


hkang1 (approved):
LGTM! Should be good to go when `test-cli-win/default` job runs successfully! Thanks for the quick turn-around @rb-determined-ai !

PR #2246, opened 2021-04-16
fix: `TFKerasTrial` check for tf2 behavior on 2.2.0 [DET-5277]

ioga:
## Description

- Checking if `tensorflow.compat.v1.disable_tf2_behavior` was called was different in tf 2.2. Use the "current" way by default and fallback to the old one otherwise.

## Checklist


naren-determined (approved):
🚢

rb-determined-ai (approved):

PR #2248, opened 2021-04-16
chore: don't check searcher shutdown to trigger work hp importance work

stoksc:
## Description
Exactly what it says on the tin.

## Commentary (optional)
I really don't think this should make a difference, but I don't have a lot of time to look and we can expend no effort to see if it works.

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -297,7 +297,7 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			experiment:     e.Experiment,
 		})
 
-		if e.searcher.Shutdown {
+		if e.State == model.CompletedState {
 			ctx.Tell(e.hpImportance, hpimportance.ExperimentCompleted{ID: e.ID})
 		}
 
--- a/master/pkg/searcher/searcher.go
+++ b/master/pkg/searcher/searcher.go
@@ -150,11 +150,11 @@ func (s *Searcher) TrialClosed(requestID model.RequestID) ([]Operation, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "error while handling a trial closed event: %s", requestID)
 	}
-	s.Record(operations)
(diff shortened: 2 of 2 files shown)


master/internal/experiment.go
```diff
@@ -297,10 +299,6 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			experiment:     e.Experiment,
 		})
 
-		if e.searcher.Shutdown {
```
stoksc, line 300: didn't move this up for any particular reason.

master/internal/experiment.go
```diff
@@ -284,6 +284,8 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 		}
 		telemetry.ReportExperimentStateChanged(ctx.Self().System(), e.db, *e.Experiment)
 
+		ctx.Tell(e.hpImportance, hpimportance.ExperimentCompleted{ID: e.ID})
+
```
mackrorysd, line 288: Didn't you hit a panic before, that wouldn't be prevent by this re-ordering?

mackrorysd: I just tested changing the e.Searcher.shutdown conditional to `if e.State == model.CompletedState {` and it seems to work well, but I don't know exactly what you do to hit all the panics you did.

stoksc: I think it should be handled by https://github.com/determined-ai/determined/pull/2230/files#diff-f8f79514e8276c806fcfa81745d24ada457e637648894ebf1d006ac5d4a29e46R188 now.

>  just tested changing the e.Searcher.shutdown conditional to if e.State == model.CompletedState { and it seems to work well, but I don't know exactly what you do to hit all the panics you did.

That is super strange..

mackrorysd (approved):

mackrorysd (approved):

PR #2250, opened 2021-04-19
chore: add webui support for deleting experiments [DET-5244]

apizzini:
## Checklist

### DIFF
--- a/webui/react/src/components/TaskActionDropdown.tsx
+++ b/webui/react/src/components/TaskActionDropdown.tsx
@@ -8,8 +8,8 @@ import Icon from 'components/Icon';
 import handleError, { ErrorLevel, ErrorType } from 'ErrorHandler';
 import { paths } from 'routes/utils';
 import {
-  activateExperiment, archiveExperiment, cancelExperiment, killExperiment, killTask,
-  openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
+  activateExperiment, archiveExperiment, cancelExperiment, deleteExperiment, killExperiment,
+  killTask, openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
 } from 'services/api';
 import { AnyTask, CommandTask, ExperimentTask, RunState } from 'types';
 import { capitalize } from 'utils/string';
@@ -24,6 +24,7 @@ export enum Action {
   Activate = 'activate',
   Archive = 'archive',
   Cancel = 'cancel',
+  Delete = 'delete',
   Kill = 'kill',
   Pause = 'pause',
   Tensorboard = 'tensorboard',
@@ -43,6 +44,7 @@ const TaskActionDropdown: React.FC<Props> = ({ task, onComplete }: Props) => {
   const isExperimentTerminal = terminalRunStates.has(task.state as RunState);
   const isArchivable = isExperiment && isExperimentTerminal && !(task as ExperimentTask).archived;
   const isUnarchivable = isExperiment && isExperimentTerminal && (task as ExperimentTask).archived;
+  const isDeletable = isExperiment && isExperimentTerminal;
   const isKillable = isTaskKillable(task);
   const isPausable = isExperiment
     && task.state === RunState.Active;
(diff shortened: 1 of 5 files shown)


webui/react/src/components/TaskActionDropdown.tsx
```diff
@@ -8,8 +8,8 @@ import Icon from 'components/Icon';
 import handleError, { ErrorLevel, ErrorType } from 'ErrorHandler';
 import { paths } from 'routes/utils';
 import {
-  activateExperiment, archiveExperiment, cancelExperiment, killExperiment, killTask,
-  openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
+  activateExperiment, archiveExperiment, cancelExperiment, deleteExperiment, killExperiment,
```
hkang1, line 11: I think the code overall looks good. Other than the experiment detail delete flow, the ones triggered from the dashboard and experiment list makes the UI seem unresponsive. Started a discussion with Infra to see if we can make this better (most notably some way for the UI to be able to distinguish something that is in the process of getting deleted)

apizzini: Do you think remembering that a delete is in progress and show a "loading/spinner" state on those elements would work?

hkang1: I think that's a big improvement but it wouldn't solve the case of if the user reloads or moves away and back to the page, because it'll just reset to the previous state of nothing ever happened. How about this, for the sake of landing this and offering it as a feature earlier on without backend dependencies, let's show the spinner state whenever possible on all the pages that support delete. We can't do much about the pages being off sync if they get reloaded, but I don't think it's the end of the world for the user to keep re-triggering the delete when they see it not going away for a few minutes. When the backend is able to provide a `DELETING` state in some form, then we can tie the deleting spinner to that particular state and disable a repeat of triggering the `Delete` action.

apizzini: 1. Is the DELETE request possibly gonna take so long that this issue is so relevant?
2. Can ask the INFRA team if it's possible to make sure that the request takes super short time to return, maybe flag the experiment as deleted in the DB and trigger a full delete async?
3. Any way when the API will be updated to have a "DELETING" state, we'll need to have spinners in those 2 places, so probably doesn't need to wait after all?

hkang1: > 1. Is the DELETE request possibly gonna take so long that this issue is so relevant?

Some of the deletes taking 2-3 minutes from what I've observed on the latest-master. Actually I haven't seen any experiment get deleted under a minute.

> 2. Can ask the INFRA team if it's possible to make sure that the request takes super short time to return, maybe flag the experiment as deleted in the DB and trigger a full delete async?

This crossed my mind but the delete can potentially fail. If we show a spinner or deleting status, block the UI from allowing a repeat delete, and unblock the user from doing other things.

> 3. Any way when the API will be updated to have a "DELETING" state, we'll need to have spinners in those 2 places, so probably doesn't need to wait after all?

Yeah this is true. Here's what I am thinking for the delete flow (assuming with deleting state support but we can fake it on the UI side for now).

Deleting from task cards on the dashboard or from the experiment row on the experiment list page:
1. User clicks overflow
2. clicks `Delete`
3. confirmation modal pops up
4. user confirms
5. task card or experiment row shows `DELETING` state and overflow no longer shows `Delete` option.

User deletes multiple rows of experiments from the experiment list page:
1. User selects multiple experiments with some that are deletable
2. confirmation modal pops up
3. user confirms
4. applicable experiment rows shows `DELETING` state and overflow no longer shows `Delete` option.

User deletes experiment from the experiment action bar from the experiment detail page:
1. User clicks delete,
2. confirmation modal pops up
3. user confirms
4. user immediately gets taken to the experiment list page

hkang1:
reached out to Infra AG again and hopefully we can get this unblocked soon.

vishnu2kmohan:
@hkang1 I imagine it's safe to close this PR now that we landed support for experiment deletion in https://github.com/determined-ai/determined/pull/2977?

PR #2251, opened 2021-04-19
feat: support configurable port and container name for Fluent Bit [DET-5272, DET-5273]

dzhu:
## Description

The `--fluent-port` option was present and affected log messages, but it
didn't actually change the port. That is fixed.

Also support changing the name of the Fluent Bit container (mostly for
testing purposes).

## Test Plan

- [x] run the agent with both options set, check that they have the right effect on the container and that trial logs still come through


rb-determined-ai (approved):
Works great!  Thank you!

hamidzr:
should this be tagging https://determinedai.atlassian.net/browse/DET-5272?

PR #2252, opened 2021-04-19
refactor: update analytics to be more targeted [DET-4857]

hkang1:
## Description

Problem: Current method collecting analytics from the WebUI is cluster centric. This is too general and causes all user activities to be lumped together under the cluster umbrella. Google analytics is unable to report user-specific analysis.
Solution: Change the identity to be user-centric instead of cluster centric.

## Test Plan

This is going to be tough because we don't record anything on local or test deployments (latest-master included), but this should work given that the mechanism of recording has not changed, and that it's just the data being supplied has.

## Checklist

### DIFF
--- a/webui/react/src/Analytics.ts
+++ b/webui/react/src/Analytics.ts
@@ -48,7 +48,36 @@ const getReadyAnalytics = (): SegmentAnalytics.AnalyticsJS | undefined => {
   return undefined;
 };
 
-export const setupAnalytics = async (auth: Auth, info: DeterminedInfo): Promise<void> => {
+export const loadAnalytics = async (info: DeterminedInfo): Promise<void> => {
+  if (!data.analytics) data.analytics = getReadyAnalytics();
+  if (!data.analytics || data.isEnabled) return;
+
+  /*
+   * Segment key should be 32 characters composed of upper case letters,
+   * lower case letters and numbers 0-9.
+   */
+  if (info.isTelemetryEnabled) {
+    try {
+      const telemetry = await getTelemetry({});
+      const isProperKey = telemetry.segmentKey && /^[a-z0-9]{32}$/i.test(telemetry.segmentKey);
+      if (isProperKey) {
+        data.analytics.load(telemetry.segmentKey || '');
+        data.isEnabled = true;
+      }
+    } catch (e) {
+      handleError({
+        error: e,
+        level: ErrorLevel.Error,
+        message: e.message,
+        publicMessage: 'Failed to load telemetry.',
+        silent: true,
+        type: ErrorType.Server,
+      });
+    }
+  }
+};
+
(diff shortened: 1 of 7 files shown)


webui/react/src/Analytics.ts
```diff
@@ -48,21 +48,27 @@ const getReadyAnalytics = (): SegmentAnalytics.AnalyticsJS | undefined => {
...
    */
-  if (info.isTelemetryEnabled) {
+  if (auth.user && info.isTelemetryEnabled) {
     try {
       const telemetry = await getTelemetry({});
       const isProperKey = telemetry.segmentKey && /^[a-z0-9]{32}$/i.test(telemetry.segmentKey);
       if (isProperKey) {
         data.analytics.load(telemetry.segmentKey || '');
-        data.analytics.identify(info.clusterId);
+        data.analytics.identify(auth.user.username, {
```
hamidzr, line 65: the first argument to the identify call seems to be used as userId. If we want to to be unique  by itself maybe we should merge it with clusterId? eg `username@clusterid`. Otherwise we'll probably get a lot of similar usernames; most likely `determined`

hamidzr (commented):
👍🏼 I left a comment about the change of user id. Also, this might make it harder to link usage this point forward to the previous usage (should still be doable since the clusterid is passed in).

Was segment  also generating its own userid where it would stick around whether the user is logged in or not and be unique to users without having to record a specific username?

webui/react/src/Analytics.ts
```diff
@@ -48,21 +48,27 @@ const getReadyAnalytics = (): SegmentAnalytics.AnalyticsJS | undefined => {
...
+export const setupAnalytics = async (auth: Auth, info: DeterminedInfo): Promise<void> => {
   if (!data.analytics) data.analytics = getReadyAnalytics();
   if (!data.analytics || data.isEnabled) return;
 
   /*
    * Segment key should be 32 characters composed of upper case letters,
    * lower case letters and numbers 0-9.
    */
-  if (info.isTelemetryEnabled) {
+  if (auth.user && info.isTelemetryEnabled) {
```
hamidzr, line 59: question: do we want to collect visits to unauthenticated pages? so far only the login page.

hkang1: separated out the identify from load so we can still get analytics going for unauthenticated flow

webui/react/src/Analytics.ts
hkang1: modified this portion to use `user id` as the identify target (instead of username or cluster id)

hamidzr (approved):
code lgtm. Let's also confirm that we're good to store userId with the team

hkang1:
will revisit this in a month or two, pending decisions. will close for now.

PR #2253, opened 2021-04-20
chore: add option to login through the new api w/ pre-hashed pwd [DET-5270]

hamidzr:
## Description

- add a param to /api/v1/auth/login to allow passing presalted and hashed password
- if the username is not provided respond with an invalid argument instead of invalid credentials.

## Checklist

### DIFF
--- a/proto/src/determined/api/v1/auth.proto
+++ b/proto/src/determined/api/v1/auth.proto
@@ -15,6 +15,8 @@ message LoginRequest {
   string username = 1;
   // The password of the user.
   string password = 2;
+  // Indicate whether the password is pre-salted & hashed or not.
+  bool is_hashed = 3;
 }
 // Response to LoginRequest.
 message LoginResponse {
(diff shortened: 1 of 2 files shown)


proto/src/determined/api/v1/auth.proto
```diff
@@ -15,6 +15,8 @@ message LoginRequest {
   string username = 1;
   // The password of the user.
   string password = 2;
+  // Indicate whether the password is pre-salted & hashed or not.
```
stoksc, line 18:
```suggestion
  // Indicate whether the provided password is pre-salted & hashed or not.
```
A little less ambiguous.

stoksc (approved):

hamidzr:
that was fast thanks!

proto/src/determined/api/v1/auth.proto
hamidzr: I shoulda committed through the ui 🤦🏼

stoksc: @hamidzr changed this to make sure people know we don't just store the prehashed password ha.

PR #2254, opened 2021-04-20
chore: migrate to use the new login api [DET-5287]

hamidzr:
## Description

followup to #2253  

## Test Plan

- check cookies still get set and websocket connections are authenticated fine.

## Commentary (optional)

image

## Checklist

### DIFF
--- a/webui/react/src/components/DeterminedAuth.tsx
+++ b/webui/react/src/components/DeterminedAuth.tsx
@@ -35,11 +35,13 @@ const DeterminedAuth: React.FC<Props> = ({ canceler }: Props) => {
     setCanSubmit(false);
     try {
       const options = { signal: canceler.signal };
-      const loginRequest: V1LoginRequest = {
-        password: creds.password || '',
-        username: creds.username || '',
-      };
-      const { token } = await login({ ...loginRequest, ...options });
+      const { token } = await login(
+        {
+          password: creds.password || '',
+          username: creds.username || '',
+        }
+        , options,
+      );
       updateDetApi({ apiKey: `Bearer ${token}` });
       const user = await getCurrentUser(options);
       storeDispatch({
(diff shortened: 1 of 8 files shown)


webui/react/src/components/DeterminedAuth.tsx
```diff
@@ -37,7 +35,11 @@ const DeterminedAuth: React.FC<Props> = ({ canceler, source }: Props) => {
     setCanSubmit(false);
     try {
       const options = { signal: canceler.signal };
-      const { token } = await login({ ...creds as Credentials, cancelToken: source.token });
+      const loginRequest: V1LoginRequest = {
+        password: creds.password || '',
```
hamidzr, line 39: I thought maybe this might be better be moved under the `services/apiConfig` but then we'd need to define a new type other that v1LoginRequest

webui/react/src/types.ts
```diff
@@ -32,11 +32,6 @@ export interface Auth {
   user?: DetailedUser;
 }
 
-export interface Credentials {
```
hamidzr, line 35: maybe we could leave this in and use this instead

apizzini: I think it's better to rely on generated TS interfaces instead of replicating them, having a "middle" interface looks close to code-replication.

Of course there are case where the argument could be different (composition, ...).

webui/react/src/components/DeterminedAuth.tsx
apizzini: Maybe I'm wrong but I think passing the `signal` as a `login` prop doesn't work?

Probably should be fixed like this:
Schermata 2021-04-21 alle 16 07 01

hamidzr:
new change: https://github.com/determined-ai/determined/pull/2254/commits/3b15a33838b19076a2c877034eea6a64606ba157

removed the call to get the current user as the new login endpoint returns that

apizzini (approved):
looks great! good catch on deleting the user request after login 👍

hamidzr:
thanks for the quick reviews

PR #2255, opened 2021-04-20
fix: reduce test flakes around task cards [DET-5192]

hkang1:
## Description

When getting a list of elements based on a selector, each element is tracked using the text content of the element (taiko `text()`). The issue arises when the content of the element changes when asynchronously fetching the element text. For example, in task cards, the recent time info can change within each second, which causes taiko to count the card twice because the previous time might be `2s ago` followed by `3s ago`.

Update the `getElements` routine to take in another optional selector that allows the test to pinpoint exactly what it wants to use for the hash id. If nothing is provided, it defaults to the current behavior of using all of the text found within the element.

## Test Plan

- [ ] rerun the `webui-e2e-tests` multiple times on circle CI and check to see if the task card flake occurs.

## Checklist

### DIFF
--- a/webui/tests/tests/StepImplementation.ts
+++ b/webui/tests/tests/StepImplementation.ts
@@ -62,21 +62,23 @@ const goto = async (url: string) => {
 /*
  * Taiko `elements()` often return duplicate elements after upgrading to 1.2.3.
  * Use the text values of the element as a hash to dedupe the elements.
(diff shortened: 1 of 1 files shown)


apizzini (approved):
I think this look good, would just suggest to add a comment about why and how the `hashSelector` param should be used so that it might be easier to understand it without understanding the code of the method :)

PR #2256, opened 2021-04-20
chore: clarify API for expconf objects

rb-determined-ai:
Make the API clear that the getters and setters are the right way to
work with expconf objects.  Rename all fields to be RawDescription,
RawHyperparameters, etc.

The RawField might be a pointer, but its getter, Field(), will only be a
pointer if no default exists for schemas.WithDefaults() to fill in.

For operations which need to manipulate the expconf object before
schemas.WithDefaults() is called, the RawField may be read or set
directly.

The RawField must be a public field, since the reflect library can't
work with private fields at all(!).

Make all versioned types in pkg/schemas/expconf into private types
and only expose the latest versions via type aliasing.


schemas/gen.py
```diff
@@ -299,22 +301,48 @@ def go_getters(struct: str, schema: Schema, spec: List[FieldSpec]) -> List[str]:
...
         if defaulted_type == type:
-            lines.append(f"func ({x} {struct}) Get{field}() {type} {{")
+            # Getter for nonpointer field.
+            lines.append(f"func ({x} {struct}) {getter}() {type} {{")
             lines.append(f"\treturn {x}.{field}")
             lines.append("}")
             lines.append("")
+
+            # Setter for nonpointer field.
+            lines.append(f"func ({x} {struct}) Set{getter}(val {type}) {{")
```
stoksc, line 323: Should this generate a setter with a pointer receiver? Seems like most of these setters won't work.
```suggestion
            lines.append(f"func ({x} *{struct}) Set{getter}(val {type}) {{")
```

schemas/gen.py
```diff
@@ -299,22 +301,48 @@ def go_getters(struct: str, schema: Schema, spec: List[FieldSpec]) -> List[str]:
...
+            lines.append(f"\t{x}.{field} = val")
+            lines.append("}")
+            lines.append("")
+
         else:
-            lines.append(f"func ({x} {struct}) Get{field}() {defaulted_type} {{")
+            # Getter for pointer field.
+            lines.append(f"func ({x} {struct}) {getter}() {defaulted_type} {{")
```
stoksc, line 330: +1 generating more idiomatic go

schemas/gen.py
```diff
@@ -299,22 +301,48 @@ def go_getters(struct: str, schema: Schema, spec: List[FieldSpec]) -> List[str]:
...
-                f'\t\tpanic("You must call WithDefaults on {struct} before .Get{field}")'
+                f'\t\tpanic("You must call WithDefaults on {struct} before .{field}")'
             )
             lines.append("\t}")
             lines.append(f"\treturn *{x}.{field}")
             lines.append("}")
             lines.append("")
 
+            # Setter for pointer field.
+            lines.append(f"func ({x} {struct}) Set{getter}(val {defaulted_type}) {{")
```
stoksc, line 341: same comment here about pointer receivers. potentially couldn't hurt to make generated code lintable, if this is an issue I think it would've caught it.

schemas/gen.py
```diff
@@ -382,10 +410,12 @@ def go_unions(
     # Define getters for each of the common members of the union.
     common_members = get_union_common_members(file, package, union_types)
     for common_field, type in common_members:
-        lines.append(f"func ({x} {struct}) Get{common_field}() {type} {{")
+        # Getter.
+        getter = common_field[len("Raw") :]
+        lines.append(f"func ({x} {struct}) {getter}() {type} {{")
         for field, _ in union_spec:
             lines.append(f"\tif {x}.{field} != nil {{")
```
stoksc, line 417: use a switch 🤣... jk..

stoksc (commented):
looks good, just a few q's

schemas/gen.py
```diff
@@ -17,7 +17,7 @@ def camel_to_snake(name: str) -> str:
     out = name[0].lower()
     for c0, c1, c2 in zip(name[:-2], name[1:-1], name[2:]):
         # Catch lower->upper transitions.
-        if c0.islower() and c1.isupper():
+        if not c0.isupper() and c1.isupper():
```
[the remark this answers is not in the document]
rb-determined-ai: corner case was around `S3Config` -> `s3config` rather than `s3_config`.

schemas/gen.py
rb-determined-ai: Ah, yes!  Great catch.

schemas/gen.py
rb-determined-ai: I can look into this.  I don't have any interest in e.g. teaching gen.py how to reflow long lines, but maybe I disable only certain linters.

schemas/gen.py
rb-determined-ai: You laugh but I expect this will happen if even partial linting is enabled.

master/pkg/schemas/expconf/length.go
```diff
@@ -21,9 +21,9 @@ const (
 
 // UnitContext contains all the context for switching the Unit of a Length freely.
 type UnitContext struct {
-	defaultUnit     Unit
```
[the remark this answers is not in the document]
rb-determined-ai: This was actually an accident.  I can revert this part.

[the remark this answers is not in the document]
rb-determined-ai: (many versions of this PR were writen, and each version but this one hit some restriction in the reflect library, so this change an accidental byproduct of some large, sweeping rage-fix after one of those)

schemas/gen.py
stoksc: Eh, it's fine for now. I'm not sure anything would've caught it actually.

rb-determined-ai:
I also fixed an issue where the generated getters needed to return non-pointers when a struct had a null default but it was required of the user (like checkpoint storage).

Also, I noticed that the golang linters already ignore code that has a "generated by..." header.

stoksc (approved):
lgtm

PR #2259, opened 2021-04-21
fix: gracefully handle Docker binding published ports to ipv4 and ipv6 [DET-5295]

stoksc:
## Description
This change updates how we determine the Addresses for a container to gracefully handle when Docker publishes ports for ipv4 and ipv6 (e.g. looks like `0.0.0.0:8080->8080/tcp, :::8080->8080/tcp`) by coalescing either to `agent-ip:host-port:...` and de-duplicating the Addresses returned to consumers when ipv4 and ipv6 are coalesced to the same address.

## Test Plan
- [ ] run this on plain ubuntu 20.04 w/o nvidia docker or host mode and make sure it works.

## Commentary (optional)
There are several ways to "fix" this - this feels the most correct. `Addresses()` still returns all addresses associated with the container in the event we need to rely on others in the future (and making this return exactly what rendezvous wants feels weird, it's in the wrong place) and it works when the container only has ipv6 bound too. The way we determine rendezvous ports - just assuming it's all the addresses associated with a container - still feels a little brittle, but it's fine for now. Alternative, if we ever expose ports other than rendezvous for a trial would just be knowing the container ports used for rendezvous and then finding how those got exposed on container created by grabbing the port binding's corresponding container port's host ports.

## Checklist

### DIFF
--- a/master/pkg/agent/master_message.go
+++ b/master/pkg/agent/master_message.go
@@ -83,23 +83,31 @@ func (c ContainerStarted) Addresses() []container.Address {
 		for _, network := range networks {
 			ipAddresses = append(ipAddresses, network.IPAddress)
 		}
+		// uniqueBindings is used to filter out bindings to 0.0.0.0 and ::
+		// that are coalesced to the same container.Address.
+		uniqueBindings := map[string]bool{}
 		for port, bindings := range info.NetworkSettings.Ports {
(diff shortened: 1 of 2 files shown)


shiyuann (approved):
LGTM

stoksc:
related https://github.com/moby/moby/issues/42313

PR #2260, opened 2021-04-21
chore: add internal preemption API [DET-5216]

stoksc:
## Description
🦗 

## Test Plan
- [x] don't break stuff.

## Commentary (optional)
This is really all there is to it.. between `func (t *trial) terminate(ctx *actor.Context, kill bool)`/`func (t *trial) terminated(ctx *actor.Context)`, there is some refactoring I would _love_ to do, but I think it'll fall naturally out of moving the workload sequencer to the harness and doing it early is just a risk since I know we'll do it later.

## Checklist


master/internal/trial.go
```diff
@@ -1169,6 +1195,17 @@ func (t *trial) terminated(ctx *actor.Context) {
...
 }
 
+func (t *trial) MarkPendingGracefulTermination(ctx *actor.Context) {
+	t.PendingGracefulTermination = true
+	for w := range t.preemptionWatchers {
+		select {
+		case w <- struct{}{}:
+		default:
+			ctx.Log().Info("skipping sending preemption signal because channel was full")
+		}
```
rb-determined-ai, line 1205: This seems like not the right behavior; why is not telling the trial to shutdown at all the right call?

master/internal/api_trials.go
```diff
@@ -531,6 +534,41 @@ func (a *apiServer) PostTrialProfilerMetricsBatch(
...
+	signal := make(chan struct{}, 1)
+	a.m.system.TellAt(trialAddr, trialWatchPreemption{signal: signal})
+	unwatch := func() { a.m.system.TellAt(trialAddr, trialUnwatchPreemption{signal: signal}) }
+
+	for {
+		select {
+		case _, ok := <-signal:
+			if !ok {
+				// Do not call unwatch() when the channel is closed.
+				return nil
```
rb-determined-ai, line 559: So if the trial actor panics or shuts down for some reason, shouldn't we start returning True from the preemption API?

And if that is the case, is there a good reason to distinguish between a close message and just closing the channel?

master/internal/trial.go
```diff
@@ -296,6 +311,14 @@ func (t *trial) Receive(ctx *actor.Context) error {
 		// the code below this switch statement to handle releasing resources in
 		// the scheduler. This should be refactored into the terminating logic.
 
+	case trialWatchPreemption:
+		t.preemptionWatchers[msg.signal] = true
+		if t.PendingGracefulTermination {
+			msg.signal <- struct{}{}
+		}
+	case trialUnwatchPreemption:
+		delete(t.preemptionWatchers, msg.signal)
```
rb-determined-ai, line 320: Do you not have to close channels in go?  I'm actually unsure.

master/internal/api_trials.go
```diff
@@ -531,6 +534,41 @@ func (a *apiServer) PostTrialProfilerMetricsBatch(
...
+
+	signal := make(chan struct{}, 1)
+	a.m.system.TellAt(trialAddr, trialWatchPreemption{signal: signal})
+	unwatch := func() { a.m.system.TellAt(trialAddr, trialUnwatchPreemption{signal: signal}) }
+
+	for {
+		select {
+		case _, ok := <-signal:
+			if !ok {
+				// Do not call unwatch() when the channel is closed.
```
rb-determined-ai, line 558: I think the cleanup process could be simplified by putting `unwatch` in a `defer`.  There's some design philosophy here, since what I'm asking for is technically unnecessary.

Any time you have objects with non-concentric lifetimes you have to deal with how things shut each other down.  The best discussion about this that I've ever seen is in the [GObject memory model](https://developer.gnome.org/gobject/stable/gobject-memory.html), especially the section on reference cycles.

My practical takeaway from that article is that shutdown events between objects that reference each other are inherently racy and it's much easier to reason about them if they are idempotent and always get called.

Or in other words, after studying the code as written, I can see that the only time the trial would close the signal is a case where unwatch would be pointless.  But an extra `unwatch` call in that case is a noop, so if you put `unwatch` in a defer statement, you can read `TrialPreemptSignal` and understand that it is correct without having to think through the asynchronous interactions between two different parts of the system.

master/internal/trial.go
stoksc: No, channels are gc'd when all references too them are gone

master/internal/trial.go
stoksc: this wouldn't happen in actor because they have technically unbounded inboxes. since this channel is bounded, it is possible we try to send on it and it is full, but since it is buffered and size of 1 then send should never block unless we recently sent another preemption signal that was not yet received and sent on by grpc (which is also non-blocking operation). so basically we could only ever see this if there were back to back preemption signals before the grpc sender thread got to run.

master/internal/api_trials.go
stoksc: > My practical takeaway from that article is that shutdown events between objects that reference each other are inherently racy and it's much easier to reason about them if they are idempotent and always get called.

yeah this makes sense.

> Or in other words, after studying the code as written, I can see that the only time the trial would close the signal is a case where unwatch would be pointless. But an extra unwatch call in that case is a noop, so if you put unwatch in a defer statement, you can read TrialPreemptSignal and understand that it is correct without having to think through the asynchronous interactions between two different parts of the system.

yeah i can call in defer, for some reason i didn't realize a tell to a nil actor was a noop

master/internal/api_trials.go
```diff
@@ -531,6 +534,41 @@ func (a *apiServer) PostTrialProfilerMetricsBatch(
...
+	switch {
+	case err == db.ErrNotFound:
+		return trialNotFound
+	case err != nil:
+		return err
+	}
+	trialAddr := actor.Addr("experiments", eID, rID)
+
+	signal := make(chan struct{}, 1)
+	a.m.system.TellAt(trialAddr, trialWatchPreemption{signal: signal})
```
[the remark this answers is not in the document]
stoksc: dig it

master/internal/api_trials.go
stoksc: trial panicing is a good thought.. i'll think about it.

stoksc: if the trial closes i'm going to return to preempt.. anything still listening should definitely preempt

master/internal/trial.go
rb-determined-ai: Seems like that would be fixed by sending the message and closing the channel, which makes more sense to me anyway.

master/internal/api_trials.go
rb-determined-ai: Oh, I guess I don't know if Telling a `nil` actor is safe, but it should never be `nil` if we get to the `defer` statement.  It might be closed, but shouldn't be nil.

master/internal/trial.go
stoksc: Sending on the channel may block though.

stoksc: This is to may sure we don’t block on this send

stoksc: oh do you mean sending the first right away (preempt or noop) then closing for the second (always preempt)? that works but I do hate encoding a protocol implicity in the msg and close like that. I guess i could leave a really good comment.

rb-determined-ai: Oh, no I was forgetting about the first message.

I mean this: make a 2-message buffered queue.  Send the first message.  Later, when you decide to preempt, send the second message, close the channel, and remove it from the list of watchers.

Right now, `MarkPendingGracefulTermination` leaves the channels open but that seems strange to me.

stoksc: Ah makes sense. To me it's not much different than this, since this really should never happen, but i'll swap it to that.

master/internal/trial.go
```diff
@@ -99,6 +101,14 @@ type (
 		ContainerID cproto.ID
 		socket      *websocket.Conn
 	}
+
+	// trialWatchPreemption begins watching if the trial has been preempted,
+	// either by the scheduler, as is the priority scheduler, or by the user, when paused.
```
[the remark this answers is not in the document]
stoksc: yeah, this wasn't meant to be an exhaustive list of all possible preemption, but just examples. i'll add it though (and qualify that priority scheduler only has preemption when it (preemption) is enabled.

master/internal/api_trials.go
```diff
@@ -531,6 +536,80 @@ func (a *apiServer) PostTrialProfilerMetricsBatch(
...
+func (a *apiServer) TrialPreemptionSignal(
+	req *apiv1.TrialPreemptionSignalRequest,
+	resp apiv1.Determined_TrialPreemptionSignalServer,
+) error {
+	trial, err := a.trialActorFromID(int(req.TrialId))
+	if err != nil {
+		return err
+	}
+
+	id := uuid.New()
```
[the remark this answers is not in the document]
stoksc: it's because once i type the trial's ref as `chan<- bool` and the listener's ref as `<-chan bool`, afaik there is no way to recover the `chan<- bool` from the `<-chan bool` (those types only convert in one direction, and using the recv chan in the send chan map is a compile error)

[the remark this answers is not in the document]
stoksc: i could probably use reflect to do it 😄

master/internal/trial.go
rb-determined-ai: it occured to me afterwards that it might be cleaner and clearer to pass the initial response in the Ask response, and only pass one message through the channel, ever.  That way the watcher responds with the initial message outside of the for loop, which would make the code structure reflect the behavior a teeny bit more.

But I don't think it makes a whole lot of difference.

master/internal/api_trials.go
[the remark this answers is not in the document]
rb-determined-ai: ah, I didn't know this.  uuid is fine.

master/internal/trial.go
stoksc: Ok, I did this but just kept the channel as a generic indicator of "one or more bools" to be received, rather that a bool in some cases and chan bool in others.

PR #2264, opened 2021-04-22
ci: enable taiko get elements logs

hkang1:
## Description

Turn on `getElements` log to see why flakes occur around this. Unable to replicate this flake both locally or on CI via PR. Take the next step enabling logging so we'll be able to see what's happening when the flake occurs.

Primarily for DET-5192

## Checklist


hamidzr (approved):
LGTM. maybe log though a logger? could be something as simple as 
```
const createLogger = (level, namespace) => ((msg) => console[level](namespace + : + msg));
```
similar idea to https://www.npmjs.com/package/debug

PR #2265, opened 2021-04-22
feat: expose user id as part of the user object [DET-4856]

hamidzr:
## Description
update v1User to include user ids

related https://github.com/determined-ai/determined/pull/2252

## Commentary (optional)

we could probably only expose each user id to only that user itself but I'm not sure if that'd get us anything.
Do you see any security concerns here? 

## Checklist


proto/src/determined/user/v1/user.proto
```diff
@@ -10,8 +10,8 @@ message User {
   option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
     json_schema: { required: [ "username", "admin", "active" ] }
   };
-  reserved 1;
-  reserved "id";
+  // The user ID.
+  int32 id = 1;
```
hamidzr, line 14: todo mark as required

stoksc (approved):
lgtm

PR #2270, opened 2021-04-23
chore: fix common fields on union types

rb-determined-ai:
## Description

chore: fix common fields on union types

It is way too complex to allow unions to contain common fields, which
can be specified by themselves (without any union member) and merged
into other fully-specified structures.  This was, in fact, exactly how
the old union marshaling and unmarshaling worked.

This complexity is why the master yaml has a checkpoint storage config
which is just a slice of bytes; it is possible to specify partial values
which cannot be represented as a CheckpointStorage object at all; they
can only be unmarshaled into other objects.

The json-schema definitions originally tried to eliminate this
complexity by doing away with common fields altogether and putting
duplicate fields on the union member structs.  That way the typing is
fairly simple to understand; in python terms:

    CheckpointStorage = Union[
        SharedFSConfig, S3Config, GCSConfig, HDFSConfig
    ]

However, that does not support the common-field-merging behavior or the
partial-config-specification behavior that we have explicitly offered
for CheckpointStorage in the master config, and also for templates.

This change redefines the v0 json schema for checkpoint storage,
searcher, and data layer to support to more accurately reflect the
current behavior, for hopefully a seamless user experience.

Interestingly, hyperparameters in the old system did not support partial
definitions like the other union types, due to the behavior of json
unmarshaling into maps.

Future versions of the expconf schema will have simpler mechanics.

## Test Plan

The correct behavior was manually tested with templates, and tests have been added to ensure correct behavior moving forward.


master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -23,6 +23,12 @@ type HyperparameterV0 struct {
 	RawCategoricalHyperparameter *CategoricalHyperparameterV0 `union:"type,categorical" json:"-"`
 }
 
+// Merge prevents recursive merging of hyperparameters; they are considered atomic units.
```
stoksc, line 26: is atomic the right word? maybe const after initialized?

master/pkg/schemas/expconf/searcher_config.go
```diff
@@ -56,12 +61,7 @@ func (s SearcherConfigV0) Unit() Unit {
 //go:generate ../gen.sh
 // SingleConfigV0 configures a single trial.
 type SingleConfigV0 struct {
-	RawMetric               string  `json:"metric"`
```
stoksc, line 59: praise.. goodness, i like this change

schemas/gen.py
```diff
@@ -144,9 +144,14 @@ def next_struct_name(file: str, start: int) -> str:
...
             if lineno < start:
                 continue
+            # Detect struct types.
             match = re.match("type ([\\S]+) struct", line)
             if match is not None:
                 return match[1]
+            # Detect non-struct types.
+            match = re.match("type ([\\S]+)", line)
+            if match is not None:
+                return match[1]
```
stoksc, line 154: I'm confused why `next_struct_name` detects non structs. Also, this wouldn't match some structs valid according to the go spec (e.g. multiple declared in a block with ()), so maybe some docs/warning there would be good.

schemas/gen.py
```diff
@@ -271,9 +279,10 @@ def get_defaulted_type(schema: Schema, tag: str, type: str) -> Tuple[str, str, b
             pass
         elif type.startswith("**"):
             raise AssertionError(f"{tag} type ({type}) must not be a double pointer")
-        elif default is not None:
+        elif not required:
             raise AssertionError(
-                f"ERROR: {tag} type ({type}) must be a pointer since it can be defaulted!\n"
+                f"ERROR: {schema.golang_title}.{tag} type ({type}) must be a pointer since is "
+                "not required!\n"
```
stoksc, line 285: This block is confusing to me (maybe it's just the error message).

stoksc (commented):
lgtm just a few q's.

schemas/gen.py
rb-determined-ai: Yeah, I'll rename it accordingly.

schemas/gen.py
rb-determined-ai: I'll see if I can't make the logic and the message clearer.

master/pkg/schemas/expconf/hyperparameters_config.go
rb-determined-ai: atomic is probably not the right word.  I'll just omit the second half.

schemas/gen.py
rb-determined-ai: How is this for a simulated message (If I remove `HyperparametersV0` from the whitelist above):

```
ERROR: ExperimentConfigV0.hyperparameters type (HyperparametersV0) must be nil-able, 
since it is not required.  This is because `nil` is how we represent values which were not provided 
by the user.  Nilable types are pointers, slices, or maps.

Note: if HyperparametersV0 is a type alias for a map or slice type, like:

    type BindMountsConfigV0 []BindMounts

then you can safely just add "HyperparametersV0" to KNOWN_MAP_OR_SLICE_ALIAS_TYPES
in schemas/gen.py to avoid this error.
```

schemas/gen.py
```diff
@@ -134,19 +134,30 @@ def gen_go_schemas_package(schemas: List[Schema]) -> List[str]:
...
             if lineno < start:
                 continue
+            # Detect struct types.
             match = re.match("type ([\\S]+) struct", line)
             if match is not None:
                 return match[1]
+            # Detect non-struct types.
+            match = re.match("type ([\\S]+)", line)
+            if match is not None:
+                return match[1]
```
stoksc, line 160: wouldn't 157-160 match everything 153-156 matches, plus non-structs? can 153-156 be removed?

stoksc (approved):
lgtm

PR #2271, opened 2021-04-24
fix: lr_scheduler step when using gradient_aggregation [DET-5289]

liamcli:
## Description
Fix a bug in how we step the `lr_scheduler` in pytorch when gradient aggregation frequency > 1.

## Test Plan
Tested locally.

### DIFF
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -251,13 +251,24 @@ class PyTorchTrialController(det.LoopTrialController):
         """
         This function aims at automatically step a LR scheduler. It should be called per batch.
         """
+
+        # Never step lr when we do not step optimizer.
+        if not self.context._should_communicate_and_update():
+            return
+
         if lr_scheduler._step_mode == pytorch.LRScheduler.StepMode.STEP_EVERY_BATCH:
-            if (batch_idx + 1) % lr_scheduler._frequency == 0:
-                lr_scheduler.step()
+            start_idx = batch_idx - self.hvd_config.aggregation_frequency + 1
+            for i in range(start_idx, batch_idx + 1):
+                if (i + 1) % lr_scheduler._frequency == 0:
+                    lr_scheduler.step()
         elif lr_scheduler._step_mode == pytorch.LRScheduler.StepMode.STEP_EVERY_EPOCH:
-            mod = (batch_idx + 1) % (len(self.training_loader) * lr_scheduler._frequency)
-            if mod == 0 or mod < self.hvd_config.aggregation_frequency:
-                lr_scheduler.step()
+            # We will step if the next optimizer step will land in the next epoch.
+            epoch_idx = self.get_epoch_idx(batch_idx)
+            next_steppable_batch = batch_idx + self.hvd_config.aggregation_frequency
+            next_batch_epoch_idx = self.get_epoch_idx(next_steppable_batch)
+            for e in range(epoch_idx, next_batch_epoch_idx):
+                if (e + 1) % lr_scheduler._frequency == 0:
+                    lr_scheduler.step()
 
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
--- a/docs/release-notes/2271-fix-lr-scheduler.txt
+++ b/docs/release-notes/2271-fix-lr-scheduler.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  PyTorchTrial: Fix learning rate scheduler behavior when used with
+   gradient aggregation.


harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -252,12 +252,32 @@ def _auto_step_lr_scheduler_per_batch(
...
+                        % self.hvd_config.aggregation_frequency
+                    )
+                    - 1
+                )
+
+                # If aggregation_frequency==1, we step at the end of an epoch.
+                # If aggregation_frequency>1, we step before the first optimizer step in next_epoch.
+                mod = (batch_idx + 1) % len(self.training_loader)
+                if mod == first_batch_idx_with_optimizer_step:
+                    lr_scheduler.step()
```
rb-determined-ai, line 280: My concern about this `elif` clause is that after over 10 minutes of specifically trying to grok what it was doing, I was unable to grok all of it.

I assume that what you have written is correct, but that's not actually enough... when code reaches a certain level of opacity, we have hard-to-detect bugs like the one you're fixing right now.

So in that vein, what do you think of writing the `elif` clause like this:

```python
# never step lr when we aren't stepping the optimizer
if not self.context._should_communicate_and_update():
    return
if lr_scheduler._step_mode == pytorch.LRScheduler.StepMode.STEP_EVERY_BATCH:
    ...
elif lr_scheduler._step_mode == pytorch.LRScheduler.StepMode.STEP_EVERY_EPOCH:
    # Step now if we know the next steppable batch would land in the next epoch.
    epoch_idx = self.get_epoch_idx(batch_idx)
    next_steppable_batch = batch_idx + self.hvd_config.aggegation_frequency
    next_batch_epoch_idx = self.get_epoch_idx(next_steppable_batch)
    # Take into account lr_scheduler frequency.
    if epoch_idx // lr_scheduler._frequency  == next_batch_epoch_idx // lr_scheduler._frequency:
        lr_scheduler.step()
```

Is that correct, and is it easy to understand?

liamcli: thanks for the suggestions.  agree that it's much clearer than my code and i will adopt your approach.

it does bring up a question of when we should step when mode is `STEP_EVERY_EPOCH`.  should we step before the first optimizer step with ANY data from a new epoch (yours) or step just before the first optimizer step with all batches from the new epoch?  

does it apply anything about how we do batch sampling when gradient-aggregation is used?  should we treat the effective batch size to be global_batch_size * aggregation_frequency and drop batches accordingly?

rb-determined-ai: >  should we step before the first optimizer step with ANY data from a new epoch (yours) or step just before the first optimizer step with all batches from the new epoch?

Ok, I think this will be easier to discuss with a little diagram.  Let's suppose aggregation_frequency of 4 and a dataset length of 10 batches.  IIUC the behaviors between what is in the PR right now ("as-written") vs the proposed behavior:

| `batch_idx` | `epoch_idx` | optimizer | lr_sched (as-written) | lr_sched (proposed) |
| ---         | ---         | ---       | ---                   | ---                 |
| 0           | 0           | step      | -                     | -                   |
| 1           | 0           | -         | -                     | -                   |
| 2           | 0           | -         | -                     | -                   |
| 3           | 0           | -         | -                     | -                   |
| 4           | 0           | step      | -                     | -                   |
| 5           | 0           | -         | -                     | -                   |
| 6           | 0           | -         | -                     | -                   |
| 7           | 0           | -         | -                     | -                   |
| 8           | 0           | step      | step                  | -                   |
| 9           | 0           | -         | -                     | -                   |
| 10          | 1           | -         | -                     | -                   |
| 11          | 1           | -         | -                     | step                |
| 12          | 1           | step      | -                     | -                   |
| 13          | 1           | -         | -                     | -                   |
| 14          | 1           | -         | -                     | -                   |
| 15          | 1           | -         | -                     | -                   |
| 16          | 1           | step      | -                     | -                   |

My conclusions are:
* the as-written mechanism could produce false-positive warnings about stepping the LR scheduler before the optimizer after pausing/restarting.  The proposed mechanism could not
* the proposed mechanism is easier to explain to users, and also easier to implement for ourselves
* normally the learning rate doesn't affect anything until the optimizer is stepped, so the only difference between the two behaviors should be side-effects of stepping the LR scheduler.  And I'm happy to tell users with such special needs that they need to step their scheduler manually.

> does it apply anything about how we do batch sampling when gradient-aggregation is used?

Gradient aggregation does not affect our batch sampling whatsoever.

>  should we treat the effective batch size to be global_batch_size * aggregation_frequency...

Oh please no.  We've taught users for so long that effective batch size == global_batch_size, I really don't want to reverse that for this.

> ... and drop batches accordingly?

Are you getting at, "should we be synchronizing end-of-epochs and stepping the LR scheduler at the end of epochs?"  My personal take is that training should either take the "continuous stream with approximate epoch boundaries" form (what we currently have) or the "global synchronization every epoch" approach.  I don't see much value in mixing them; the resulting explosion of configuration options (truncate partial batches? truncate partial frequency steps? truncate partial shards?) seems uselessly large. 

I do think that some time after sprinkle API lands, we should rework the epoch-ends in pytorchtrial to support three modes:
 * stream-based training
   * exactly what we have now
 * epoch-based training, no truncation
   * train all shards to completion, even with mismatched shard lengths (I'm actually not sure this is possible with a horovod backend)
   * train all batches in each shard, even if not divisible by aggregation frequency
   * always step optimizer and lr_scheduler at the end of the (also not possible with horovod backend, I'm pretty sure)
 * epoch-based training, truncate everything
   * truncate shards so every worker has the same number of batches
   * truncate again to ensure frequencies
   * always train in increments of global_batch_size * aggregation_frequency
   * all the things are synchronized after every epoch

Ok, now that I've written that out, I think actually only options 1 and 3 are possible unless we support torch.distributed in pytorchtrial.

liamcli: Thanks for the detailed response.  

batch_idx | epoch_idx | optimizer | lr_sched (as-written) | lr_sched (proposed)
-- | -- | -- | -- | --
0 | 0 | - | - | -
1 | 0 | - | - | -
2 | 0 | - | - | -
3 | 0 | step | - | -
4 | 0 | - | - | -
5 | 0 | - | - | -
6 | 0 | - | - | -
7 | 0 | step | step | -
8 | 0 | - | - | -
9 | 0 | - | - | -
10 | 1 | - | - | -
11 | 1 | step | - | -
12 | 1 | - | - | -
13 | 1 | - | - | -
14 | 1 | - | - | -
15 | 1 | step | - | step
16 | 1 | - | - | -
17 | 1 | - | - | -
18 | 1 | - | - | -
19 | 1 | step | step | step
20 | 2 | - | - | -
21 | 2 | - | - | -
22 | 2 | - | - | -
23 | 2 | step | - | -
24 | 2 | - | - | -
25 | 2 | - | - | -
26 | 2 | - | - | -
27 | 2 | step | step | -
28 | 2 | - | - | -
29 | 2 | - | - | -
30 | 3 | - | - | -
31 | 3 | step | - | -
32 | 3 | - | - | -
33 | 3 | - | - | -
34 | 3 | - | - | -
35 | 3 | step | - | step
36 | 3 | - | - | -
37 | 3 | - | - | -
38 | 3 | - | - | -
39 | 3 | step | step | step
40 | 4 | - | - | -
41 | 4 | - | - | -
42 | 4 | - | - | -
43 | 4 | step | - | -

I agree that stepping if the next optimizer is in a new epoch is the better choice because it more closely approximates stepping at the end of an epoch.

rb-determined-ai (approved):
great catch on this!

rb-determined-ai:
We should probably have some release notes for this fix.

harness/determined/pytorch/_pytorch_trial.py
armandmcqueen: Effective batch size has to be `instantaneous_batch_size * aggregation_frequency`. That's the definition and I don't think you can use gradient aggregation correctly without understanding that. 

Option 3 seems like the right approach to me with a feature/docs for randomling duplicate data in the dataset to make the size of the dataset evenly divisible by the effective batch size.

I'm a little confused about Option 1. If I have 101 pieces of data training on 10 GPUs (per-GPU batch size = 1) for 5 epochs, how many data points will I train on? 550? 510?

PR #2274, opened 2021-04-24
fix: improve rendering for uPlot chart with empty data [DET-5330]

apizzini:
## Description

This is just preventing renderings when chart data has no X values (that's where the main layout imperfection are happening).

If there are X values but no Y values, the chart will still be rendered, but chart layout is ok.
Cases like this, where possible, needs to be handled on a per-chart basis, because the issue of having no Y data can have 2 sources:
- no series or empty series
- series with data but hidden

A good example is the "GPU Hours by Label" chart in [the 2nd point of this page](http://localhost:3000/det/cluster/historical-usage?after=2021-01-01&before=2021-01-31&group-by=day) and the relative code fix in this MR.

## Test Plan

Those 2 URLs will cause 2 different types of empty chart to be rendered:
- [ ] [chart with no X values](http://localhost:3000/det/experiments/1455)
- [ ] [cluster chart that has X values but no visible Y values](http://localhost:3000/det/cluster/historical-usage?after=2021-01-01&before=2021-01-31&group-by=day)

## Checklist

### DIFF
--- a/webui/react/src/components/UPlotChart.tsx
+++ b/webui/react/src/components/UPlotChart.tsx
@@ -1,7 +1,8 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { throttle } from 'throttle-debounce';
 import uPlot, { AlignedData } from 'uplot';
 
+import Message, { MessageType } from 'components/Message';
 import useResize from 'hooks/useResize';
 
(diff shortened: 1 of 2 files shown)


hkang1 (approved):
Looks awesome 👍 

Screen Shot 2021-04-26 at 11 04 53 AM
Screen Shot 2021-04-26 at 11 04 59 AM

PR #2277, opened 2021-04-26
fix: add support for dynamic section content via css [DET-5299]

hkang1:
## Description

Trial logs does not render properly on `Safari 14.0.3` (`13.1.1` works fine)
Screen Shot 2021-04-26 at 10 20 30 AM

The problem is due to `Safari 14.0.3` not respecting the CSS height of `100%`. `LogViewerTimestamp` attempts to dynamically fetch the height of the space available to set height of the visible log viewer, but on `Safari 14.0.3` it was returning `0` due to CSS issues.

The solution is to adjust the CSS to work on Safari (and the other browsers) so that the height can correctly and dynamically be fetched.

Fixed:
Screen Shot 2021-04-26 at 10 20 18 AM

## Test Plan

- [ ] Check that the Trial logs render properly in Safari `14.0.3`
- [ ] Check that the Trial logs render properly in Safari `13.1.1` or older
- [ ] Check that the Trial logs render properly in Chrome
- [ ] Check that the Trial logs render properly in Firefox

## Checklist

### DIFF
--- a/docs/release-notes/2277-fix-safari-trial-logs.txt
+++ b/docs/release-notes/2277-fix-safari-trial-logs.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  WebUI: Fix the issue of trial logs not rendering properly on Safari
+   14.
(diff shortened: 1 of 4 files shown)


apizzini (approved):
thanks for fixing this :)

tested on safari 13.1.1 (i told you the last version 13.11.1, it's actually 13.1.1).

PR #2279, opened 2021-04-26
chore: apply react strict mode and upgrade to React 17 [DET-5325]

hkang1:
## Description

`React Strict Mode` is going to help us migrate to `React v17`.
- apply `React Strict Mode` and fix all errors and warnings.
- upgrade to React 17
- check for any additional errors or warnings to fix

[React Strict Mode Reference](https://reactjs.org/docs/strict-mode.html)

## Test Plan

- [ ] run `npm start` and check to make sure all the pages still work
- [ ] run `npm run storybook` and make sure it runs without errors
- [ ] run `npm run build-storybook` and make sure it runs without errors

## Checklist


webui/react/package.json
```diff
@@ -35,7 +35,7 @@
     "query-string": "^6.13.5",
     "react": "^16.14.0",
     "react-dom": "^16.14.0",
-    "react-helmet": "^6.1.0",
```
hkang1, line 38: `react-helmet` is considered [inactive](https://github.com/nfl/react-helmet/issues/548#issuecomment-665351178)

webui/react/package.json
```diff
@@ -82,7 +82,6 @@
     "@types/plotly.js": "^1.50.21",
     "@types/react": "^16.9.52",
     "@types/react-dom": "^16.9.8",
-    "@types/react-helmet": "^6.1.0",
```
hkang1, line 85: `@types/react-helmet-async` is not needed as it [provides its own definitions](https://www.npmjs.com/package/@types/react-helmet-async).

webui/react/package.json
```diff
@@ -33,9 +33,9 @@
     "portable-fetch": "^3.0.0",
     "pretty-bytes": "^5.4.1",
     "query-string": "^6.13.5",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
-    "react-helmet": "^6.1.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
hkang1, line 37: decided to upgrade here to see how it went and surprisingly smooth because everything was resolved during the `Strict Mode` application.

[React 17 Migration Guide](https://reactjs.org/blog/2020/10/20/react-v17.html)

webui/react/package.json
```diff
@@ -33,9 +33,9 @@
     "portable-fetch": "^3.0.0",
     "pretty-bytes": "^5.4.1",
     "query-string": "^6.13.5",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
-    "react-helmet": "^6.1.0",
+    "react": "^17.0.2",
```
apizzini, line 36: There looks to be something off in the [LogViewer storybook](http://localhost:9009/?path=/story/logviewer--default) which is broken in this branch, but not on master?

webui/react/package.json
```diff
@@ -33,9 +33,9 @@
     "portable-fetch": "^3.0.0",
```
apizzini, line 33: We should consider updating the other deps too?

Schermata 2021-04-27 alle 09 11 00

apizzini:
This is looking good aside for the logviewer issue, but I would consider landing this after the release just so that we can have people play around with the WebUI for 2 weeks before releasing it.

hkang1:
> This is looking good aside for the logviewer issue, we should probably consider landing this after the release just so that we can have people play around with the WebUI for 2 weeks before releasing it.

👍  Yeah not looking to rush this one, because issues probably will popup up over time as it's a major upgrade.

webui/react/package.json
hkang1: I have a separate ticket / story for the task of upgrading the rest just in case it became a huge ordeal. But want to tackle this soon too.

webui/react/package.json
hkang1: great find!

Turned out `LogViewer` is using `Page` which requires the new `HelmetProvider` as a decorator. Newest commit should fix this issue.

apizzini (approved):
looks great! 👍

PR #2281, opened 2021-04-26
fix: squelch "response already committed" master log message

neilconway:
This was emitted when creating new experiments. The cause is that
postExperiment() calls echo.Context#JSON(http.StatusCreated, ...), but it is
wrapped by api.Route(), which calls echo.Context#JSON() itself, with a different
status code. The second call is ignored but results in printing the ugly log
message.

To fix this, we can either

  (a) change the HTTP status code for POST'ing an experiment to
      be http.StatusOK, instead of http.StatusCreated, or
  (b) change postExperiment() to not use api.Route(), or
  (c) make api.Route() smart enough to support different HTTP response codes.

This commit takes option A, which results in a change to the REST API. However,
switching from status 201 to 200 is probably not a change that will upset many
API clients.

## Checklist

### DIFF
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -434,7 +434,6 @@ func (m *Master) postExperiment(c echo.Context) (interface{}, error) {
 	}
 
 	dbExp, validateOnly, taskSpec, err := m.parseCreateExperiment(&params)
-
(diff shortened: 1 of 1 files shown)


armandmcqueen (approved):
LGTM!

PR #2282, opened 2021-04-27
feat: add support for preemption in Kubernetes [DET-5135]

eecsliu:
## Description
In conjunction with the kubernetes coscheduler plugin, this change facilitates preemption in kubernetes by allowing experiments to be paused. Experiments are paused when their pods are tagged with the "determined-preemption" label. 

## Test Plan
Tested manually on kubernetes with the coscheduler running alongside determined.

## Checklist

### DIFF
--- a/master/internal/kubernetes/preemptions.go
+++ b/master/internal/kubernetes/preemptions.go
@@ -1,8 +1,9 @@
 package kubernetes
 
 import (
-	"github.com/determined-ai/determined/master/pkg/actor"
 	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/actor"
 
 	k8sV1 "k8s.io/api/core/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/master/internal/kubernetes/preemptions.go
+++ b/master/internal/kubernetes/preemptions.go
@@ -1,0 +1,79 @@
+package kubernetes
+
+import (
+	"github.com/determined-ai/determined/master/pkg/actor"
+	"github.com/pkg/errors"
+
+	k8sV1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sClient "k8s.io/client-go/kubernetes"
+)
+
+// Messages that are sent to the preemption listener.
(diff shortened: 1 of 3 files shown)


master/internal/kubernetes/preemptions.go
```diff
@@ -0,0 +1,80 @@
...
+func (p *preemptionListener) startPreemptionListener(ctx *actor.Context) error {
+	watch, err := p.clientSet.CoreV1().Pods(p.namespace).Watch(
+		metaV1.ListOptions{LabelSelector: "determined-preemption"})
+	if err != nil {
+		return errors.Wrap(err, "error initializing preemption watch")
+	}
+
+	ctx.Log().Info("preemption listener is starting")
+	for pod := range watch.ResultChan() {
+		newPod, ok := pod.Object.(*k8sV1.Pod)
```
mackrorysd, line 68: I'm worried about getting ourselves stuck in a state where the master restarts right here. I want to check in with @stoksc about how good we are about avoiding cases like this to know if this is really an immediate deal-breaker, but I think we should check for the labels on start-up too and deal with them.

mackrorysd (changes requested):

master/internal/kubernetes/preemptions.go
```diff
@@ -57,6 +57,17 @@ func (p *preemptionListener) Receive(ctx *actor.Context) error {
...
 func (p *preemptionListener) startPreemptionListener(ctx *actor.Context) error {
+	// check if there are pods to preempt on startup
+	pods, err := p.clientSet.CoreV1().Pods(p.namespace).List(
+		metaV1.ListOptions{LabelSelector: "determined-preemption"})
+	if err != nil {
+		return errors.Wrap(err,
+			"error in initializing preemption listener: checking for pods to preempt")
+	}
+	for _, pod := range pods.Items {
+		ctx.Tell(p.podsHandler, podPreemption{podName: pod.Name})
```
mackrorysd, line 68: Why do we not need *k8sV1.Pod type assertion here if we do down below?

eecsliu: I believe the watch returns an object of type `Event`, which wraps whatever object gets sent. `List` on the other hand returns a list of pods, already properly typed.

mackrorysd (approved):

PR #2283, opened 2021-04-27
fix: wait for uPlot to be ready to setData or setSize [DET-5343]

apizzini:
## Description

While reworking `<LearningCurveChart>` I discovered a couple issues on the <UPlotChart> component.
Preferred to split this in a dedicated MR that can land before the release gets cut.

## Checklist

### DIFF
--- a/webui/react/src/components/UPlotChart.tsx
+++ b/webui/react/src/components/UPlotChart.tsx
@@ -49,7 +49,7 @@ const UPlotChart: React.FC<Props> = ({ data, options }: Props) => {
    */
   useEffect(() => {
     if (!chart || !data) return;
-    setTimeout(() => chart.setData(data), 1000);
+    chart.setData(data);
   }, [ chart, data ]);
 
   /*
--- a/webui/react/src/components/UPlotChart.tsx
+++ b/webui/react/src/components/UPlotChart.tsx
@@ -19,36 +19,37 @@ const SCROLL_THROTTLE_TIME = 500;
 const UPlotChart: React.FC<Props> = ({ data, options }: Props) => {
   const [ chart, setChart ] = useState<uPlot>();
   const chartDivRef = useRef<HTMLDivElement>(null);
+  const hasData: boolean = useMemo(() => !!(data && data[0].length > 0), [ data ]);
 
   /*
    * Chart setup.
    */
   useEffect(() => {
-    if (!chartDivRef.current || !options) return;
+    if (!chartDivRef.current || !hasData || !options) return;
 
     const optionsExtended = uPlot.assign(
       {
         cursor: { drag: { dist: 5, uni: null, x: true, y: true } },
+        hooks: { ready: [ (chart: uPlot) => setChart(chart) ] },
         width: chartDivRef.current.offsetWidth,
       },
       options,
     );
 
     const plotChart = new uPlot(optionsExtended as uPlot.Options, [ [] ], chartDivRef.current);
-    setChart(plotChart);
 
     return () => {
       setChart(undefined);
       plotChart.destroy();
     };
-  }, [ chartDivRef, options ]);
+  }, [ chartDivRef, hasData, options ]);
 
   /*
    * Chart data.
    */
   useEffect(() => {
     if (!chart || !data) return;
-    chart.setData(data);
+    setTimeout(() => chart.setData(data), 1000);
   }, [ chart, data ]);
 
   /*
@@ -56,7 +57,7 @@ const UPlotChart: React.FC<Props> = ({ data, options }: Props) => {
    */
   const resize = useResize(chartDivRef);
   useEffect(() => {
-    if (!chart || !options?.height) return;
+    if (!chart || !options?.height || !resize.width) return;
     chart.setSize({ height: options.height, width: resize.width });
   }, [ chart, options?.height, resize ]);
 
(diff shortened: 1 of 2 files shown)


webui/react/src/components/UPlotChart.tsx
```diff
@@ -19,29 +19,30 @@ const SCROLL_THROTTLE_TIME = 500;
...
       {
         cursor: { drag: { dist: 5, uni: null, x: true, y: true } },
+        hooks: { ready: [ (chart: uPlot) => setChart(chart) ] },
         width: chartDivRef.current.offsetWidth,
       },
       options,
     );
 
     const plotChart = new uPlot(optionsExtended as uPlot.Options, [ [] ], chartDivRef.current);
-    setChart(plotChart);
```
apizzini, line 38: We might `setChart` and trigger the other `useEffect` before the chart is ready.
Moved to the ready hook.

webui/react/src/components/UPlotChart.tsx
```diff
@@ -83,18 +90,11 @@ const UPlotChart: React.FC<Props> = ({ data, options }: Props) => {
...
-   * Don't plot the chart if there are no X values.
-   */
-  const hasData: boolean = useMemo(() => {
-    return !!(data && data[0].length > 0);
-  }, [ data ]);
-
-  if (!hasData) {
-    return <Message title="No data to plot." type={MessageType.Empty} />;
-  }
+  useImperativeHandle(ref, () => chart, [ chart ]);
```
apizzini, line 93: this will be used in `<LearningCurveChart>` to set focused series when highlighting the table.

hkang1: **question:** would passing in the highlighted series index as a prop work as an alternative?

apizzini: I just think it makes more sense to expose the uPlot instance to be able to access `setSeries` and all those similar methods for stuff that will be tweaked on a per-case basis.

Aiming to have a huge component handling all the tweaks we potentially are gonna need on all the kind of charts look a bit aggressive to me.

On the other hand if we ever will build another chart that needs the same logic that should be shared somehow (no code duplication).

webui/react/src/components/UPlotChart.tsx
```diff
@@ -12,36 +14,41 @@ export interface Options extends Omit<uPlot.Options, 'width'> {
...
 const SCROLL_THROTTLE_TIME = 500;
 
-const UPlotChart: React.FC<Props> = ({ data, options }: Props) => {
+const UPlotChart: React.FC<Props> = forwardRef((
+  { data, options }: Props,
+  ref?: React.Ref<uPlot|undefined>,
+) => {
   const [ chart, setChart ] = useState<uPlot>();
   const chartDivRef = useRef<HTMLDivElement>(null);
+  const hasData: boolean = useMemo(() => !!(data && data[0].length > 0), [ data ]);
```
hkang1, line 28: **suggestion:** I just ran into a problem with `LearningCurveChart` where `uPlot` reports a problem if the series in the `data` don't match up in length with the `xValues`. So something like this will report an error:

```
[
  [ 1, 2, 3 ],
  [],
  [],
]
```

this will result in:
Screen Shot 2021-04-27 at 12 07 28 PM
Screen Shot 2021-04-27 at 12 07 39 PM

The reason this can happen is because sometimes series data can come in a bit later than xValues. Can we update this `hasData` to check that every row of data (xValues, and series data) all have the same length and not empty to be considered true?

hkang1 (approved):
LGTM! one non-blocking suggestion on beefing up what is considered `hasData` to reduce console errors.

webui/react/src/components/UPlotChart.tsx
apizzini: Do you have instructions to replicate the issue quickly? Just to be sure i properly address the fix when working on this.

webui/react/src/components/UPlotChart/tooltipsPlugin.ts
```diff
@@ -115,25 +112,29 @@ export const tooltipsPlugin = ({ getXTooltipHeader, getXTooltipYLabels }: Props
...
-          hide();
+        if (!left || left < 0 || !top || top < 0 || idx == null) {
+          if (displayedIdx) hide();
           return;
         }
 
-        if (idx != null && idx !== displayedIdx) {
-          showIdx(uPlot, idx);
+        if (idx !== displayedIdx) {
+          const hasXValue = !!uPlot.series.find((serie, serieId) => (
```
apizzini, line 125: i pushed the fix we discussed yesterday here (not showing the line - and the tooltip - when the tooltip would be empty).

can test on [this page ](http://localhost:3000/det/experiments/330/trials/381)with just "accuracy" turned on.

webui/react/src/components/UPlotChart.tsx
```diff
@@ -12,36 +14,50 @@ export interface Options extends Omit<uPlot.Options, 'width'> {
...
+  { data, options }: Props,
+  ref?: React.Ref<uPlot|undefined>,
+) => {
   const [ chart, setChart ] = useState<uPlot>();
   const chartDivRef = useRef<HTMLDivElement>(null);
 
+  const hasData: boolean = useMemo(() => {
+    // no x values
+    if (!data || data[0].length === 0) return false;
+    // series values length not matching x values length
```
apizzini, line 32: here's the fix you asked.

hkang1: that looks great thank you 🙏

webui/react/src/components/UPlotChart/tooltipsPlugin.ts
hkang1: that worked beautifully!

webui/react/src/components/UPlotChart.tsx
hkang1: [test on this](http://localhost:3000/det/experiments/1289/visualization) once `LearningCurveChart` migrates over to `UPlotChart`

PR #2287, opened 2021-04-27
fix: only warn for non-numeric np.dtypes [DET-5288]

rb-determined-ai:
Numerous users have run into mysterious error messages due to having
numpy arrays of python objects or strings.

It turns out numpy has support for string arrays that is useful to NLP
users, so there's not necessarily good reason to enforce that users use
python lists instead of numpy arrays for non-numeric data.

We can just handle it better.

The allowable dtypes for pytorch come from:

    pytorch/torch/csrc/utils/tensor_numpy.cpp

The corresponding kinds are derived from:

    numpy/numpy/core/numerictypes.py

### DIFF
--- a/harness/determined/pytorch/_data.py
+++ b/harness/determined/pytorch/_data.py
@@ -394,13 +394,17 @@ def to_device(
 
     if isinstance(data, dict):
         return {k: to_device(v, device, warned_types) for k, v in data.items()}  # type: ignore
-    if isinstance(data, list):
+    elif isinstance(data, list):
         return [to_device(d, device, warned_types) for d in data]  # type: ignore
-    if isinstance(data, tuple):
+    elif isinstance(data, tuple):
         return tuple(to_device(d, device, warned_types) for d in data)  # type: ignore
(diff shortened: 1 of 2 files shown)


harness/determined/pytorch/_data.py
```diff
@@ -409,4 +413,4 @@ def to_device(
             f"Was not able to move data item of type '{type(data).__name__}' to device."
         )
 
-    return data
+    return data  # type:ignore
```
ioga, line 416: I agree with the type checker - this function doesn't return `TorchData` as its type signature suggests.

ioga (approved):
fix lgtm, but if this was a public API, then the type signature should be more reliable.

harness/determined/pytorch/_data.py
rb-determined-ai: Yeah, I noticed that too.  I'm really unsure how it didn't need a `type: ignore` before, since it was wrong before as well.

I chose to kick that can down the road in this case.

PR #2293, opened 2021-04-28
chore: Helm notes should recognize preemption scheduler

mackrorysd:
## Description

In connection with @eecsliu's recent changes - just a missed clause in the NOTES file.

## Test Plan

Ran Helm locally


eecsliu (approved):
👩‍🍳😗👌

PR #2295, opened 2021-04-28
chore: improve master logging

neilconway:
Two main changes:

  * Reduce gRPC log level from INFO to WARN. gRPC INFO log messages are
    almost universally about cryptic implementation details, such as

      pickfirstBalancer: HandleSubConnStateChange: 0xc0006a2010, {READY <nil>}

  * Squelch a log message ("initializing endpoints for agents") that does
    not seem useful.

Along the way, fix various typos and improve log message consistency.

## Checklist

### DIFF
--- a/master/internal/grpcutil/api.go
+++ b/master/internal/grpcutil/api.go
@@ -26,22 +26,28 @@ const jsonPretty = "application/json+pretty"
 
 // NewGRPCServer creates a Determined gRPC service.
 func NewGRPCServer(db *db.PgDB, srv proto.DeterminedServer) *grpc.Server {
-	logger := logrus.NewEntry(logrus.StandardLogger())
+	// In go-grpc, the INFO log level is used primarily for debugging
+	// purposes, so omit INFO messages from the master log.
+	logger := logrus.New()
+	logger.SetLevel(logrus.WarnLevel)
+
+	logEntry := logrus.NewEntry(logger)
+	grpclogrus.ReplaceGrpcLogger(logEntry)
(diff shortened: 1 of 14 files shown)


neilconway:
@stoksc Not sure offhand if it is okay to create multiple logrus loggers...

stoksc (approved):
lgtm

stoksc:
from an example in the logrus docs:
> // Create a new instance of the logger. You can have any number of instances.

I just assumed, based on the thinking `logrus.New()` would be a weird function to have otherwise.

stoksc:
Huh wait, maybe that's multiple instances of the same logger.. I guess I'll read better.

master/internal/grpcutil/api.go
```diff
@@ -26,22 +26,25 @@ const jsonPretty = "application/json+pretty"
 
 // NewGRPCServer creates a Determined gRPC service.
 func NewGRPCServer(db *db.PgDB, srv proto.DeterminedServer) *grpc.Server {
-	logger := logrus.NewEntry(logrus.StandardLogger())
+	logger := logrus.New()
```
stoksc, line 29: I spent some time reading the logrus code - afaict it seems totally fine to have two logrus loggers but I'm always happy to be surprised..

neilconway: Yep, that is my reading as well -- assuming the underlying `io.Writer` is safe to write to concurrently, which I think should be the case here with `os.Stderr`.

PR #2296, opened 2021-04-28
chore: add support for building with Golang race detector

neilconway:
This is not enormously useful, because of how our actor system
works. But it still might have some value when debugging concurrency
issues or interactions with external libraries.

## Checklist

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -21,6 +21,14 @@ build:
 		-o build/determined-agent \
 		./cmd/determined-agent
 
+.PHONY: build-race
+build-race:
+	go build \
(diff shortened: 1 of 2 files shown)


stoksc (approved):
lgtm. I've added this flag and removed so many times (along with tons of flags to `go test` manually).. will be useful

stoksc:
> This is not enormously useful, because of how our actor system
works. But it still might have some value when debugging concurrency
issues or interactions with external libraries.

fwiw the last time I turned this on I found some bugs with how we were asynchronously registering echo endpoints deep in the actor system during startup and parts of echo's internals weren't threadsafe.

PR #2298, opened 2021-04-28
chore: remove end_time as required from json-schema

mh-determined:
## Description
When an experiment is created it only has start_time, and end_time is not set. The experiment stays in 'Pause' state before it is activated. Swagger generates an exception on the response as the Experiment requires end_time not to be 'None'.
 
Getting a list of experiments also generates this exception if some experiments are in active/pause state.

## Checklist


hamidzr (approved):
LGTM

hamidzr:
thanks @mh-determined

PR #2300, opened 2021-04-29
docs: make experiment config page top level

armandmcqueen:
## Description

Make the experiment config doc page a top-level section in the table of contents by moving it out of the `references` folder and into the root folder.

Screen Shot 2021-04-28 at 9 27 17 PM

## Commentary (optional)

When you search for "experiment configuration" the experiment config page result isn't even above the fold any more

Screen Shot 2021-04-28 at 9 40 56 PM

## Moving out of References section

This pulls the experiment config out of the `references` section, which isn't ideal, but Sphinx really makes customizing the ToC hard. All the other options I tried were worse. 

If you tell the ToC to include the experiment config as a top level section, the ToC tries to highlight and expand both Experiment Configuration and References sections in the sidebar at the same time when you are in the expconf page, which looks bad. 

Screen Shot 2021-04-28 at 9 37 33 PM

If you duplicate the file, Sphinx will complain and you need to make all the page/section references have unique names.

If there's a Sphinx expert around maybe they can do better, but if not I think this is an improvement to what we have now.

## Checklist


neilconway:
@armandmcqueen Thanks for thinking about this! Just so I understand the rationale -- the root problem we're trying to solve here is discoverability of the experiment config, not search results as such, right? i.e., if the search is crappy (it is), then I'd argue for fixing that rather than doing what is proposed here.

armandmcqueen:
I would say the core problem is that the experiment config is hidden too deep in the documentation for how often I expect people to refer to it and it gets more and more tucked away as more docs are added. It's a bit about discoverability and a bit about ease of reference.

It would be good if we had data to measure popularity of various docs pages - do we have that?

The search results getting worse and worse was just the trigger for me to finally do something about it.

neilconway:
> It would be good if we had data to measure popularity of various docs pages - do we have that?

Yep, we have Google Analytics for the docs site -- let me add you.

> I would say the core problem is that the experiment config is hidden too deep in the documentation for how often I expect people to refer to it and it gets more and more tucked away as more docs are added.

Yeah -- I agree it would be nice to make it more accessible, but this solution seems a bit ugly and asymmetric to me (why treat experiments differently than other workloads or the cluster config page?). I think adding links/cross-references to the experiment config from other docs pages helps. But I'm fine with this as an interim measure.

neilconway (approved):

armandmcqueen:
The usage data shows that experiment config is the most accessed reference page, but the main thing we need to promote is installation docs. I don't think this approach is worth doing for the experiment config, since it won't generalize to the more important docs. Better to do an actual 'frequently-accessed pages' feature

PR #2302, opened 2021-04-29
chore: move LearningCurveChart to use uPlot shared component [DET-5331]

apizzini:
## Description

The main accomplishment here is making the closer point tooltip and focusing a re-usable plugin for uPlot and making code cleaner while keeping it performant.

## Test Plan

- [ ] focusing/unfocusing series (via hovering table and chart data points)
- [ ] clicking data points on chart (open trial in same window or new window)
- [ ] tooltip positioning and text

## Checklist


webui/react/src/routes/utils.ts
```diff
@@ -68,14 +68,18 @@ export const handlePath = (
...
+    if (isShiftClick(event) || options.popout) {
       openBlank(href);
     } else {
       routeAll(href);
     }
   }
 };
 
+export const isShiftClick = (event: MouseEvent|React.MouseEvent): boolean => {
+  return event.button === 1 || event.metaKey || event.ctrlKey;
```
hkang1, line 80: **nit:** `event.button === 1` is really the middle mouse button (sometimes the clicking of the wheel button).

webui/react/src/routes/utils.ts
```diff
@@ -68,14 +68,18 @@ export const handlePath = (
...
-    if (event.button === 1 || event.metaKey || event.ctrlKey || options.popout) {
+    if (isShiftClick(event) || options.popout) {
       openBlank(href);
     } else {
       routeAll(href);
     }
   }
 };
 
+export const isShiftClick = (event: MouseEvent|React.MouseEvent): boolean => {
```
hkang1, line 79: **suggestion:** hmmm, I wonder if there is a better place for this utility. `utils/browser.ts` or a new one for `utils/event.ts`?

hkang1, line 79: **suggestion:** maybe something more prescriptive like `isNewTabEvent` or `isNewTabClick`?

webui/react/src/routes/utils.ts
apizzini: you're telling me that just because is the middle mouse button doesn't match the isShiftClick naming?

Not sure if there's a better name for this method.

webui/react/src/routes/utils.ts
apizzini: isNewTabClickEvent?

and there was another comment to move this in another util file, totally fine for me.

webui/react/src/routes/utils.ts
hkang1: yeah exactly

webui/react/src/routes/utils.ts
hkang1: yeah not sure what or which util file it should go into, but I don't feel strongly about leaving it here for now or moving it.

hkang1 (approved):
LGTM! Awesome work in fulling extracting all the uPlots into `UPlotChart.tsx` and creating the specialized plugins for cases!!

PR #2306, opened 2021-04-30
fix: improvements to automatic pod spec configurator

eecsliu:
## Description
There were two issues with the automatic podspec labeler in determined - it labeled GC tasks as requiring a GPU node, which is incorrect, and it would forgo automatic setup if the user manually provided a scheduler name (but is missing some important labels). Now GC tasks require 0 nodes and should be immediately scheduled, while the automatic podspec setup has been tweaked to be more forgiving.

## Checklist


master/internal/kubernetes/spec.go
```diff
@@ -175,8 +176,8 @@ func (p *pod) configureVolumes(
 }
 
 func (p *pod) modifyPodSpec(newPod *k8sV1.Pod, scheduler string) {
-	if scheduler != "coscheduler" || newPod.Spec.SchedulerName != "" ||
-		p.taskSpec.Description() == "cmd" {
+	if scheduler != coscheduler || p.taskSpec.Description() == "cmd" ||
+		(newPod.Spec.SchedulerName != "" && newPod.Spec.SchedulerName != coscheduler) {
```
mackrorysd, line 180: Nit: cmd is another magic string we should try to factor out of here and  task_spec.go.

Also - why is preemption not included here?

mackrorysd (changes requested):

master/internal/kubernetes/spec.go
eecsliu: will change cmd. Since preemption and coscheduling are the same on determined's side, I just had the preemption option basically just point to the coscheduler. Should I make it more verbose?

mackrorysd: Oh I had missed that. It's not new on that line anyway... I feel like we need distinct terms for coscheduler, coscheduler-with-preemption, and the combination. But that can be a separate discussion.

eecsliu: Since we're introducing another scheduler, I'm going to refactor this part of the code to be more friendly to different scheduler types.

eecsliu: Also, `Description()` will always return "cmd", "gc", or an experiment name.

master/internal/kubernetes/spec.go
```diff
@@ -175,16 +181,27 @@ func (p *pod) configureVolumes(
 }
 
 func (p *pod) modifyPodSpec(newPod *k8sV1.Pod, scheduler string) {
-	if scheduler != "coscheduler" || newPod.Spec.SchedulerName != "" ||
-		p.taskSpec.Description() == "cmd" {
+	if p.taskSpec.Description() == cmdTask {
+		return
+	}
+
+	if scheduler == coscheduler || scheduler == preemptionScheduler {
```
mackrorysd, line 188: I think the refactor is WAY clearer - thanks. I hoped Github would show this as a "suggestion" as part of a review (a-la Google Docs), but it looks like it literally added a commit. What do you think of just combining the 2 conditionals for the 2 scheduler types? For now they're identical, the conditional is still super clear (IMO anyway), and this reduces the code a little bit. If you're agreeable, I really like this.

mackrorysd (approved):

PR #2307, opened 2021-04-30
feat: add experiment notes & name [DET-5352]

hamidzr:
## Description

- introduce notes and name fields to experiment configuration.
- set name for old experiment configs based on description

misc
- fix a typo in fork flow
- make description filter case insensitive
- ~fix an issue with det e describe when label is None~ separate PR https://github.com/determined-ai/determined/pull/2411/files

## Test Plan

- test-migration.sh
- run against latest master db
- try patching description, note, and name
- create experiment with and without name/description

## Commentary (optional)

migration perf: 500 experiments ~5s on my local machine. If this scales linearly it'd translate to 100 seconds for 10k experiments
- how to deal with experiments that have a description set to config and are used to seeing that be used in our system as name in places such as the UI
- existing experiments in the database with description and no name

I explore some solutions in this PR to these questions. The upcoming experiment config versioning alongside experiment config schema seems to gives us a way to address both of these.

## Checklist

- [x] take out test commit/code


master/internal/api_experiment.go
```diff
@@ -590,6 +590,7 @@ func (a *apiServer) CreateExperiment(
 		return nil, status.Errorf(codes.Internal, "failed to get the user: %s", err)
 	}
 
+	// TODO save originial user submission req.Config
```
hamidzr, line 593: https://github.com/determined-ai/determined/pull/2332

hamidzr:
- how to deal with experiments that have description set to config and are used to seeing that be used in our system as name in places such as the UI
- existing experiments in the database with description and no name

I explore some solutions in this PR to these questions. The upcoming experiment config versioning alongside experiment config schema seems to gives us a way to address both of these.

hamidzr:
@stoksc  could you do a quick review so we can chat about the comments I left in code and in the commentary section?

master/static/srv/get_experiments.sql
```diff
@@ -1,6 +1,8 @@
 WITH filtered_exps AS (
     SELECT
         e.id AS id,
+        COALESCE(NULLIF(e.config->>'name', ''), NULLIF(e.config->>'description', ''), 'Experiment ' ||  e.id) AS name,
+        e.config->>'notes' AS notes,
```
stoksc, line 5: Thinking about it now, should notes be in the config? Should notes be transferred on a fork, or should they start over. It feels like maybe they should start over but we should ask an AMLE. wdyt?

master/static/srv/get_experiment.sql
```diff
@@ -1,5 +1,9 @@
 SELECT
     e.id AS id,
+    /* TODO We could either do this and rely that we go through the apis that use this query
+     (eg update cli) or migrate older experiments in database (one way). Look at model defaults */
+    COALESCE(NULLIF(e.config->>'name', ''), NULLIF(e.config->>'description', ''), 'Experiment ' ||  e.id) AS name,
```
stoksc, line 5: This logic could happen here, but I think it's probably better to write a shim for the exp conf and trust that all old expconfs are shimmed on startup, rather than scattering shimming logic all over the db. e.g. what happens when another query reads name and forgets this? we have the capability to just correct the data model so we probably should.

master/static/srv/get_experiments.sql
hamidzr: the idea was that they'd have the option of setting it in their experiment config. although since we want to support annotations for notes that might be a bit tricky to manually embed in a yaml or json config file.

master/static/srv/get_experiments.sql
```diff
@@ -1,10 +1,15 @@
...
     SELECT
         e.id AS id,
+        e.config->>'name' AS name,
         e.config->>'description' AS description,
         e.config->'labels' AS labels,
         e.config->'resources'->>'resource_pool' AS resource_pool,
         e.config->'searcher'->'name' as searcher_type,
+        CASE
+            WHEN NULLIF(e.notes, '') IS NULL THEN NULL
+            ELSE 'omitted'
```
hamidzr, line 11: wdy think @stoksc

master/static/srv/get_experiment.sql
hamidzr: I added a migration to do this.

test-migration.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# make -C tools start-db
+
+psql -h localhost -U postgres -d postgres <<EOF
```
stoksc, line 5: do you want to add something like this?

master/static/srv/get_experiments.sql
```diff
@@ -1,10 +1,15 @@
 WITH filtered_exps AS (
     SELECT
         e.id AS id,
+        e.config->>'name' AS name,
         e.config->>'description' AS description,
         e.config->'labels' AS labels,
         e.config->'resources'->>'resource_pool' AS resource_pool,
         e.config->'searcher'->'name' as searcher_type,
+        CASE
```
stoksc, line 9: I'm confused by this, if the notes as '' then they're not omitted right, they're just not there? wouldn't this be the query to always just return 'omitted' if we're not going to return them from list endpoints?

master/static/migrations/20210511152412_description-to-name.up.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE experiments
```
stoksc, line 1: whether we're shimming in a migrations or shimming on query, i dont think we should be shimming in the database going forward, but this is probably fine for just this.

master/internal/api_experiment.go
```diff
@@ -67,18 +67,12 @@ func (a *apiServer) GetExperiment(
 		return nil, err
 	}
 
-	confBytes, err := a.m.db.ExperimentConfigRaw(int(req.ExperimentId))
+	config, err := a.m.db.ExperimentConfig(int(req.ExperimentId))
```
stoksc, line 70: what's the purpose of this? it was this way for a reason - all old exp confs can deser into a `map[string]interface{}` but some will fail deser'ing into the current expconf model.

master/internal/api_experiment.go
```diff
@@ -499,6 +502,7 @@ func (a *apiServer) PatchExperiment(
 		"patch_experiment",
 		req.Experiment.Id,
 		marshalledPatches,
+		exp.Notes,
```
stoksc, line 505: rewriting the notes every time is a bit disappointing but i don't think it'll cause issues.

master/static/srv/get_experiments.sql
hamidzr: I used this to encode whether the experiment has a note or not but we probably don't want users to rely on that 🤔  if so we can just set them all to 'omitted'

stoksc: Ah I get it now, duh

test-migration.sh
hamidzr: I'll take this out when everything is done. I don't think we keep migration tests anywhere in the repo do we?

master/static/srv/get_experiments.sql
hamidzr: keep?

master/internal/api_experiment.go
hamidzr: aah I didn't realize that I'll revert.

master/static/migrations/20210511152412_description-to-name.up.sql
hamidzr: hmm what would be some alternatives?

hamidzr:
@hkang1  or @apizzini  could you please review the webui changes?

apizzini:
WebUI side looks good, the only thing I think can probably be improved is that we're now showing experiment name in the "dashboard" and "experiments" (experiment list) pages, while in the "experiment detail" page we're showing an "Experiment <ID> (<description>)" title. Probably better to at least switch description and name there?

hamidzr:
> WebUI side looks good, the only thing I think can probably be improved is that we're now showing experiment name in the "dashboard" and "experiments" (experiment list) pages, while in the "experiment detail" page we're showing an "Experiment ()" title. Probably better to at least switch description and name there?

I'll update that

master/internal/api_experiment.go
```diff
@@ -206,7 +208,8 @@ func (a *apiServer) GetExperiments(
 		archivedExpr,
 		userFilterExpr,
 		labelFilterExpr,
-		req.Description,
+		fmt.Sprintf("%%%s%%", req.Description),
```
stoksc, line 211: If this is for WHERE x LIKE '%...%' in sql i'd probably push that down to the query and just use this to pass the param.

master/internal/api_experiment.go
```diff
@@ -500,34 +503,63 @@ func (a *apiServer) PatchExperiment(
 	}
 
 	paths := req.UpdateMask.GetPaths()
+	updateNotes := false
 	for _, path := range paths {
 		switch {
-		case path == "description":
-			exp.Description = req.Experiment.Description
+		case path == "name":
+			isEmpty, err := regexp.Match(`^ *$`, []byte(req.Experiment.Name))
```
stoksc, line 510: Why not just `if len(strings.TrimSpace(req.Experiment.Name)) == 0`?

master/pkg/model/defaults.go
```diff
@@ -31,12 +31,10 @@ const (
 
 // DefaultExperimentConfig returns a new default experiment config.
 func DefaultExperimentConfig(taskContainerDefaults *TaskContainerDefaultsConfig) ExperimentConfig {
```
stoksc, line 33: I could be wrong but I feel like this shouldn't be used anymore, I see it mostly just used by tests. cc @rb-determined-ai ?

master/internal/core_experiment.go
```diff
@@ -290,9 +290,7 @@ func (m *Master) patchExperiment(c echo.Context) (interface{}, error) {
 		dbExp.Config.SetResources(resources)
 	}
 	if patch.Description != nil {
```
stoksc, line 292: do you know if anything still calls this echo patch?

stoksc (commented):
looks pretty good. are we runtime defaulting name? I looked for it but couldn't find it

hamidzr:
> looks pretty good. are we runtime defaulting name? I looked for it but couldn't find it

I used the definition from expconf. should be in default.go

```
	conf := schemas.WithDefaults(expconf.ExperimentConfig{}).(expconf.ExperimentConfig)

	defaultConfig := ExperimentConfig{
		Name: conf.RawName.String(),
```

master/internal/api_experiment.go
hamidzr: huh I didn't know about that. 👍🏼

master/pkg/model/defaults.go
hamidzr: hmm this is not used any less than it was used before the pr, but after expconf pr, and If I'm not mistaken commands use it but only two parts of it so we can probably simplify it.

hamidzr: used to impl db.Scanner interface for experimentconf and in a bunch of tests too so I'd keep it out of this pr but I can remove the reference from the commands config.

master/internal/core_experiment.go
hamidzr: our cli does. I moved our cli usage for description and name to use the new API. 
once we add support for what 

```
		Resources *struct {
			MaxSlots api.MaybeInt `json:"max_slots"`
			Weight   *float64     `json:"weight"`
		} `json:"resources"`
		CheckpointStorage *struct {
			SaveExperimentBest int `json:"save_experiment_best"`
			SaveTrialBest      int `json:"save_trial_best"`
			SaveTrialLatest    int `json:"save_trial_latest"`
		} `json:"checkpoint_storage"`
```
is offering in our new apis we should be able to move off of this.

master/internal/command/command.go
```diff
@@ -50,17 +50,11 @@ type commandOwner struct {
...
 // does not specify any configuration options.
 func DefaultConfig(taskContainerDefaults *model.TaskContainerDefaultsConfig) model.CommandConfig {
-	expConf := model.DefaultExperimentConfig(taskContainerDefaults)
+	resources := model.DefaultResourcesConfig(taskContainerDefaults)
+	resources.Slots = 1
 	return model.CommandConfig{
-		Resources: model.ResourcesConfig{
-			Slots:  1,
-			Weight: 1,
-			// SlotsPerTrial is not used by commands. They prefer Slots instead.
```
stoksc, line 58: 🙃 👍

master/pkg/model/defaults.go
```diff
@@ -29,14 +29,52 @@ const (
...
+	config := ResourcesConfig{
+		SlotsPerTrial:  1,
+		Weight:         1,
+		NativeParallel: false,
+	}
+	if taskContainerDefaults == nil {
+		return config
+	}
+
+	config.Devices = taskContainerDefaults.Devices
```
stoksc, line 43: this isn't the subject of this pr but this behavior is questionable (multiple containers on the same agent could all get access to all the devices on the cluster, this is really a scheduler deficiency).

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -81,15 +82,15 @@ func (e *ExperimentConfigV0) Scan(src interface{}) error {
 	return nil
 }
 
-// Description is a container struct for handling runtime defaults. It has to be a container so that
+// Name is a container struct for handling runtime defaults. It has to be a container so that
 // it can be responsible for allocating the nil pointer if one is not provided.  It would be nice if
-// you could use `type Description *string` but go won't let you create methods on such a type.
-type Description struct {
+// you could use `type Name *string` but go won't let you create methods on such a type.
+type Name struct {
```
stoksc, line 88: maybe ask @rb-determined-ai how he feels about this/run it by him since he's landing the last bit of expconf/fixing it up now

master/static/srv/get_experiments.sql
```diff
@@ -28,9 +33,10 @@ WITH filtered_exps AS (
                     ELSE e.config->'labels' END
                 ))
             )
-        AND ($5 = '' OR POSITION($5 IN (e.config->>'description')) > 0)
+        AND ($5 = '' OR (e.config->>'description') ILIKE  ('%%' || $5 || '%%'))
```
stoksc, line 36: what is %% (i'm only aware of % and i dont think doubling what i know would change behavior)?

stoksc (commented):
looks great. we should probably have an extended chat about

> hmm, do you have any alternatives in mind that would result in updating the database? maybe load, shim, push to the database one by one?

if you have time

harness/determined/common/api/experiment.py
```diff
@@ -14,6 +14,12 @@
 from determined.common.api import request as req
 
 
+def patch_experiment_v1(master_url: str, exp_id: int, patch_doc: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
hamidzr: >  python sdk?

Do you mean the new APIs? Ideally, I'd want us to use the python-sdk that @mh-determined  worked on or directly use the swagger bindings instead of manually writing apis like this. both of those would require bigger changes that I'd like to keep out of this PR. This is following other uses of calls to `/api/v1/X` in our cli code which isn't as ideal as using the generated code or the sdk directly to me but it's a step toward it.

master/internal/api_experiment.go
```diff
@@ -500,34 +502,59 @@ func (a *apiServer) PatchExperiment(
...
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to marshal experiment patches")
 	}
 
-	if _, err := a.m.db.RawQuery(
-		"patch_experiment",
-		req.Experiment.Id,
-		marshalledPatches,
-	); err != nil {
+	if updateNotes {
```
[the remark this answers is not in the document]
hamidzr: as in run 2 separate queries? 

that does seem cleaner and less error-prone. Your point is that these would outweigh the perf cost? I agree.

harness/determined/common/api/experiment.py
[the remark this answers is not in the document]
stoksc: > Ideally, I'd want us to use the python-sdk that @mh-determined worked on or directly use the swagger bindings instead of manually writing apis like this

this is exactly what i mean.

> This is following other uses of calls to /api/v1/X in our cli code which isn't as ideal as using the generated code or the SDK directly to me but it's a step toward it.

yeah, i was just curious when you thought the next step should be.

master/static/srv/get_experiments.sql
hamidzr: I needed to escape it otherwise it'd get substituted and it's because of sprintf there

```
func (db *PgDB) QueryProtof(
	queryName string, args []interface{}, v interface{}, params ...interface{}) error {
	query := db.queries.getOrLoad(queryName)
	if len(args) > 0 {
		query = fmt.Sprintf(query, args...)
	}

```

 hence escaping it with go syntax otherwise it'd interfere with the orderby string. 
related to args vs params substitution in our 

` select  ('%' || 'asdf' || '%' ) as test from users where username LIKE ('%' || ''|| '%');` works fine in pgsql but through our queryproto:

`determined-db_1       | 2021-05-26 21:58:17.026 UTC [216] STATEMENT:  select  ('%!'(string=id ASC) || 'asdf' || '%!'(MISSING) ) as test from users where username LIKE ('%!'(MISSING) || ''|| '%!'(MISSING))`

master/internal/api_experiment.go
```diff
@@ -500,34 +502,59 @@ func (a *apiServer) PatchExperiment(
...
 		Description string   `json:"description"`
+		Name        string   `json:"name"`
+	}
+	patches := experimentPatch{
+		Labels:      exp.Labels,
+		Description: exp.Description,
+		Name:        exp.Name,
 	}
-	patches := experimentPatch{Description: exp.Description, Labels: exp.Labels}
 	marshalledPatches, err := json.Marshal(patches)
```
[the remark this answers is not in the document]
hamidzr: isn't marshaled the correct spelling?

master/pkg/schemas/expconf/experiment_config.go
hamidzr: we have chatted about it. let me know if you have any other comments @rb-determined-ai

stoksc: ok I just forget everything, sgtm.

master/static/srv/get_experiments.sql
stoksc: oh yeah, duh, query proto _f_. ok, i was just curious if this was some new thing I didn't know lol

stoksc (approved):
lgtm

master/internal/api_experiment.go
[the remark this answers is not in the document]
stoksc: either is fine but yeah that was my point, I think it is cleaner.

harness/determined/common/api/experiment.py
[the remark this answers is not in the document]
hamidzr: could we consider this tech debt and do a tech-debt week on?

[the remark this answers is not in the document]
hamidzr: I added an entry in the sheet

PR #2308, opened 2021-04-30
feat: omnibar initial support [DET-5374]

hamidzr:
## Description

- add an omnibar
- added utilties to the storage class for marshalling and unmarshalling, `keys`, `reset`, and `fork`
- add a basic script with express to serve the UI and redirect all routes to `index.html` to let react-router handle them.

## Checklist


webui/react/TODO.md
```diff
@@ -0,0 +1,60 @@
+# Product Week
```
hamidzr, line 1: I'll remove this unless we want something like this as opposed to inline todos. (or maybe we rather not leave any TODOs in git)

webui/react/src/omnibar/Component.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { omnibar } = useStore();
+  const globalKeyHandler = {
+    HIDE_OMNIBAR: (): void => {
+      storeDispatch({ type: StoreAction.HideOmnibar });
+    },
+    SHOW_OMNIBAR: (): void => storeDispatch({ type: StoreAction.ShowOmnibar }),
+  };
+
+  useEffect(() => {
+    exposeStore({ dispatch: storeDispatch, state: { omnibar } });
```
hamidzr, line 26: Not the safest thing to do but gives us access to the store without turning different omnibar trees into React components.

webui/react/TODO.md
hamidzr: https://determinedai.atlassian.net/browse/DET-5382

webui/react/package.json
```diff
@@ -24,18 +24,21 @@
     "dayjs": "^1.10.4",
     "eventemitter3": "^4.0.7",
     "fp-ts": "^2.8.4",
+    "fuse.js": "^6.4.6",
     "history": "^4.10.1",
     "humanize-duration": "^3.24.0",
     "io-ts": "^2.2.11",
     "js-sha512": "^0.8.0",
     "js-yaml": "^3.14.0",
+    "omnibar": "github:hamidzr/omnibar",
```
hkang1, line 33: capturing what we talked about in terms of packaging the omnibar:
- start by bringing in the entire omnibar project into `determined` codebase
- making the necessary customizations to the omnibar project
- keep the omnibar package mostly isolated (including the tweaks) so we are able to break it out into a separate repo of its own once it has stablized

hkang1:
Some visual recommendations:2021-05-12 at 2 32 PM
2021-05-12 at 1 20 PM

Some observed quirks:
- escape to dismiss the omnibar doesn't work for me (even right after triggering the omnibar with nothing typed)

UI suggestions:
- after user sees the help modal, when the user dismisses the modal (via okay button, escape key, enter key), re-focus on the omnibar and clear out the bar input, so the user can go right into using the omnibar again
- for help modal, for options that take in addition param, show the param. e.g. `goto experiment <id>` and `kill notebook <id>`
- break up combo words for commands. e.g. `pauseExperiment` to `pause experiment`
- consistent command pattern of "action" followed by "noun" (or could be other formats, but what you have makes sense IMO). e.g. instead of `dev serverAddress reset`, `reset serverAddress`. 
- using `Tab` to select the highlighted option (after using the arrow keys) but NOT execute the option. Where as `Enter` would actually fire off the command if applicable. So if user typed `hel` and the highlighted option is `help`, hitting `Enter` would bring up the help modal, whereas `Tab` would just show `help ` typed into the omnibar but not yet fired. So thinking of `Tab` as just an option selector, and `Enter` as both an option selector and also execution of the command.
- upon a match where there are options being shown, show the full command instead of just the next available word. So if I type `goto `, it would show `goto **experiment**`, `goto **cluster**`, etc (with the shared portion showing in lighter great, so the auto complete portion would be more emphasized)
- change "no sub branches: exit" to something like "no options matched". IMO, also thinking that it might be better to prevent `Enter` from dismissing the omnibar (only allowing escape key at this point)

Nice to haves:
- help modal to have direct links for each option that can then take the user back to the omnibar with the selected option partially filled out
- support for keyboard shortcuts for available options (added a screenshot below)

webui/react/README.md
```diff
@@ -41,7 +41,7 @@ The page will automatically load and display new changes via [Hot Module Replace
 - `SERVER_ADDRESS`: If set, directs the WebUI to find the Determined cluster at this address.
 This allows users to host the WebUI on a separate server from Determined. This would need the target
 server to allow requests coming from the domain hosting the WebUI, aka CORS.
-- `PUBLIC_URL`: Indicates where the React assets are being served from relative to the root of the webserver.
+- `PUBLIC_URL`: Indicates where the React assets are being served from relative to the root of the webserver. Set this variable to an empty string to serve from `/`.
```
hkang1, line 44: 👍

webui/react/src/App.tsx
```diff
@@ -1,36 +1,47 @@
...
+  const globalKeyHandler = {
+    SHOW_OMNIBAR:
+    (): void => storeDispatch({ type: StoreAction.ShowOmnibar }),
+  };
+
   useRouteTracker();
   useTheme();
 
   // Poll every 10 minutes
+
```
hkang1, line 44: lint will catch this I'm sure

webui/react/src/App.tsx
```diff
@@ -1,36 +1,47 @@
...
+  const storeDispatch = useStoreDispatch();
 
   const fetchInfo = useFetchInfo(canceler);
 
   useKeyTracker();
+  const globalKeyHandler = {
+    SHOW_OMNIBAR:
+    (): void => storeDispatch({ type: StoreAction.ShowOmnibar }),
+  };
+
```
hkang1, line 39: **nit** move this above `useKeyHandler`?

webui/react/package.json
```diff
@@ -24,18 +24,21 @@
...
     "plotly.js": "^1.56.0",
     "portable-fetch": "^3.0.0",
     "pretty-bytes": "^5.4.1",
     "query-string": "^6.13.5",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-helmet-async": "^1.0.9",
+    "react-hotkeys": "^2.0.0",
```
hkang1, line 41: **question:** thoughts on using the `useKeyTracker` event bus versus including a `react-hotkeys` library? `GalleryModal` has an example on how to use our `useKeyTracker` on the listening side.

`react-hotkeys` is not a big bundle by any means, but for the simple use case of just changing a store bit when a key combo is detected, the `useKeyTracker` should suffice?

Screen Shot 2021-05-12 at 3 27 49 PM

webui/react/src/contexts/Store.tsx
```diff
@@ -16,11 +16,16 @@ interface UI {
...
+interface OmnibarState {
+  isShowing: boolean;
+}
+
+export interface State {
   agents: Agent[];
   auth: Auth & { checked: boolean };
   cluster: ClusterOverview;
   info: DeterminedInfo;
+  omnibar: OmnibarState
```
hkang1, line 28: **suggestion:** Roll the omnibar state into the `ui: UI` section?

webui/react/src/omnibar/AsyncTree.ts
```diff
@@ -0,0 +1,85 @@
...
+  onCustomInput?: (input: string) => Eventually<Children>;
+  options?: Children | ComputedChildren; // leaf nodes have no children
+}
+
+export const isLeafNode = (node: TreeNode): node is LeafNode =>
+  'onAction' in node && !('options' in node);
+export const isNLNode = (node: TreeNode): node is NLNode =>
+  !('onAction' in node) && ('options' in node || 'onCustomInput' in node);
+export const isTreeNode = (node: TreeNode): node is TreeNode =>
+  'title' in node && node.title !== undefined && (isLeafNode(node) || isNLNode(node));
```
hkang1, line 32: **chore:** sounds like this is probably not needed based on our conversation

webui/react/src/omnibar/AsyncTree.ts
```diff
@@ -0,0 +1,85 @@
...
+export type TreeNode = LeafNode | NLNode;
+export type ComputedChildren = (arg?: NLNode) => Eventually<Children>
+export type FinalAction = (node?: LeafNode) => Eventually<void>;
+
+export interface LeafNode extends BaseNode {
+  onAction: FinalAction;
+}
+
+// Non-leaf Node
+export interface NLNode extends BaseNode {
```
hkang1, line 22: **suggestion:** `NonLeafNode`? `BranchNode`?

webui/react/src/omnibar/Component.module.scss
```diff
@@ -0,0 +1,21 @@
+.base {
+  background-color: rgba(0, 0, 0, 0.7);
```
hkang1, line 2: **style:** `rgba(0, 0, 0, 0.45)` to match Ant Design's overlay for their modal

webui/react/src/omnibar/Component.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  left: 0;
+  position: absolute;
+  top: 0;
+  width: 100%;
+
+  .bar {
+    left: 50%;
+    min-width: 50rem;
+    position: absolute;
+    top: 50%;
```
hkang1, line 13: **style suggestion:** `33.333%`

webui/react/src/omnibar/Component.module.scss
```diff
@@ -0,0 +1,21 @@
...
+    transform: translate(-50%, -50%);
+    width: calc(max(40vw, 300px));
```
hkang1, line 15: **style suggestion:** adding a mobile break point so it looks proper on a mobile device:
```
@media only screen and (max-width: $breakpoint-mobile) {
  width: ...
  top: ...
}
```
Haven't played around too much but for mobile, it might be better to have the bar right at the top and have a padding all around both the omnibar and the options that show. The options would be nice if it has a max height of 100vh minus the omnibar height and also scroll if it spills past the screen height.

webui/react/src/omnibar/Component.tsx
hkang1: **suggestion:** this feels like an anti-pattern. How about defining a `const handleAction = useCallback(..., []);` inside `Component.tsx` and in there put all the logic that is in `Tree.onAction`. Taking a look at `Tree.ts` more closely, it seems like it is a good candidate to split up into half into `omnibar/utils` and half directly into `Component.tsx`. Especially `onAction` and `extension` since they are React specific and don't really need to be extracted.

webui/react/src/omnibar/Tree.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+    } else {
+      // trigger the query.
+      input.value = input.value + SEPARATOR;
+      query(input.value);
+    }
+
+  }
+  return Promise.resolve();
+};
```
hkang1, line 105: **suggestion:** a bit of repeating, but these 3 things should probably just be in `omnibar/Components.tsx` directly, or they can be injected as a hook e.g. `useOmnibarTree` because they have React dependencies. I'm leaning more towards direct approach because `omnibar/Component` itself is tiny and there's no reusability concerns here.

hkang1: The rest can go inside an omnibar utility file.

webui/react/src/omnibar/modifiers/anchor/BaseRenderer.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react';
```
hkang1, line 1: **nit:** `omnibar/modifier/anchor` path seems a bit odd... what about just `src/omnibar/OmnibarRenderer.tsx`?

webui/react/src/omnibar/modifiers/anchor/BaseRenderer.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const COLORS = {
+  BLACK: '#000',
+  BLUE: '#00f',
+  DARKGRAY: '#ddd',
+  GRAY: '#eee',
+  GREEN: '#0f0',
+  RED: '#f00',
+  WHITE: '#fff',
+};
```
hkang1, line 15: **question:** can we use our css var theme colors?

webui/react/src/omnibar/exposedStore.ts
```diff
@@ -0,0 +1,14 @@
...
+interface ES {
+  dispatch: Dispatch<Action>,
+  state: Partial<State>
+}
+
+export let store: null | ES = null;
+
+export const exposeStore = (x: ES): void => {
+  store = x;
+};
```
hkang1, line 14: I think we can avoid this

webui/react/src/omnibar/sampleTree.ts
```diff
@@ -0,0 +1,196 @@
+import { displayHelp, parseIds, visitAction } from 'omnibar/actions';
```
hkang1, line 1: **nit:** rename file to something more like `decisionTree.ts`?

webui/react/src/utils/storage.ts
```diff
@@ -76,11 +81,44 @@ export class Storage {
     this.store.setItem(path, item);
   }
 
+  keys(): string[] {
+    const prefix = this.pathKeys.join(this.delimiter) + this.delimiter;
+    return this.store.keys()
```
hkang1, line 86: **question:** why not just use `Object.keys(this.store)` here and avoid having to prototype?

webui/react/src/omnibar/actions.tsx
```diff
@@ -0,0 +1,44 @@
+import { message, Modal } from 'antd';
```
hkang1, line 1: **nit:** rename to `src/omnibar/actions/utils.ts`?

webui/react/src/omnibar/Component.tsx
```diff
@@ -0,0 +1,50 @@
+import OmnibarNpm from 'omnibar';
```
hkang1, line 1: **nit:** rename to `OmniBar.tsx`?

webui/react/src/omnibar/AsyncTree.ts
hkang1, line 22: **suggestion:** move these typescript definitions to `src/omnibar/types.ts`?

webui/react/package.json
hamidzr: I think both TS and eslint are gonna be very unhappy.
we can ignore eslint for the whole directory but ts would be more challenging. I'll look into this more.

webui/react/src/omnibar/modifiers/anchor/BaseRenderer.tsx
hamidzr: I was keeping same path the package used but we don't need to do that. I like your suggestion, how about we name it ItemRenderer as it only renders a single item/option

webui/react/src/omnibar/actions.tsx
hamidzr: I restructure this and nodes directory into `trees`

webui/react/src/omnibar/sampleTree.ts
hamidzr: I restructured them into `trees`  dir how does that sound?

webui/react/src/App.tsx
hamidzr: I don't think we have such a lint rule. I removed the added line

webui/react/package.json
hamidzr: i converted it and it was pretty nice to use. + fixed the escape issue

webui/react/src/omnibar/AsyncTree.ts
hamidzr: I restructured them under tree-extension dir

webui/react/src/omnibar/Tree.ts
hamidzr: > Some observed quirks:
> 
> escape to dismiss the omnibar doesn't work for me (even right after triggering the omnibar with nothing typed)

should be fixed now. please check if cmd/ctrl + space works on mac to launch the bar

> UI suggestions:
 
> after user sees the help modal, when the user dismisses the modal (via okay button, escape key, enter key), re-focus on the omnibar and clear out the bar input, so the user can go right into using the omnibar again
> help modal to have direct links for each option that can then take the user back to the omnibar with the selected option partially filled out
> for help modal, for options that take in addition param, show the param. e.g. goto experiment <id> and kill notebook <id>

> break up combo words for commands. e.g. pauseExperiment to pause experiment

the only downside I see to this is that it'd be a bit slower to access as you can't get the full picture without getting the pause fully in. (that might be where tab to complete is helpful since enter to complete can risk executing something). This might also be addressed by allowing the user at each step to search through the fully resolved branches (same output as help) instead of just the immediate children (REF A)

> consistent command pattern of "action" followed by "noun" (or could be other formats, but what you have makes sense IMO). e.g. instead of dev serverAddress reset, reset serverAddress.

This makes sense. I mainly put it the other way to 1. better organize it: eg all serverAddress operations are under serverAddress 2. keep the tree narrower closer to the root 3. better discoverability: by looking at the nodes under dev you can't know what `dev reset` is but on the other hand looking at `dev serverAddress` you can guess operations relating to serverAddress can be found there.

> using Tab to select the highlighted option (after using the arrow keys) but NOT execute the option. Where as Enter would actually fire off the command if applicable. So if user typed hel and the highlighted option is help, hitting Enter would bring up the help modal, whereas Tab would just show help typed into the omnibar but not yet fired. So thinking of Tab as just an option selector, and Enter as both an option selector and also execution of the command.

> upon a match where there are options being shown, show the full command instead of just the next available word. So if I type goto , it would show goto **experiment**, goto **cluster**, etc (with the shared portion showing in lighter great, so the auto complete portion would be more emphasized)

doable. will need to make sure if the command gets the overflow is handled reasonably

> change "no sub branches: exit" to something like "no options matched". IMO, also thinking that it might be better to prevent Enter from dismissing the omnibar (only allowing escape key at this point)

Hmm do you think it probably serves our users better if it always stays around as opposed to sometime auto closing after a commnad? :thinking

> Nice to haves:
> 
> support for keyboard shortcuts for available options (added a screenshot below)



considering the suggestions, the work estimates, and impact estimates, I'd prioritize these 3 in this order to work on first.

- search against fully resolved branches
- merge mousehover and is selected
- break commands such as pauseExperiment

webui/react/src/omnibar/Component.module.scss
hamidzr: hmm I wouldn't expect mobile users to be using this 🤔  how would they even trigger this? unless we add a special case for them

webui/react/src/utils/storage.ts
hamidzr: the idea is that this gives each store a chance to implement their own `.keys`. However looking at the two store we currently have what you're suggesting would work as well. should I update it?

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const storeDispatch = useStoreDispatch();
+
+  const hideBar = useCallback(
+    () => storeDispatch({ type: StoreAction.HideOmnibar }),
+    [ storeDispatch ],
+  );
+
+  const onAction = useCallback((item, query) => {
+    // ideally we could useRef here but can we w/o modifying the omnibar package?
+    const input: HTMLInputElement|null = document.querySelector('#omnibar input[type="text"]');
```
hamidzr, line 22: any ideas on how we could avoid this? an option is to push a ref down to the omnibar package but that'd involve changes to omnibar package

hkang1:
### Bugs

- The z-index tweaking covers the help modal now. If there isn't an easy solution with the ordering, worse case you can have it as before so the notification appears on the lower right.

Screen Shot 2021-05-17 at 9 16 26 PM
 
- There is something weird going on with fuzzy match when it comes to the numbers. So when I try to match `goto experiment 1455`, if I type `goto experiment 145` and I'm on the option for `1455` and I hit enter, it changes it to `455` instead.
  
  Screen Shot 2021-05-19 at 9 27 58 AM

webui/react/src/omnibar/Component.module.scss
hkang1: lol that's a great point. So used to supporting all breakpoints, without consideration of how they can actually even trigger it.

webui/react/src/omnibar/Tree.ts
hkang1: > > Some observed quirks:
> > escape to dismiss the omnibar doesn't work for me (even right after triggering the omnibar with nothing typed)
> 
> should be fixed now. please check if cmd/ctrl + space works on mac to launch the bar
> 
works great!

> > UI suggestions:
> 
> > after user sees the help modal, when the user dismisses the modal (via okay button, escape key, enter key), re-focus on the omnibar and clear out the bar input, so the user can go right into using the omnibar again
> > help modal to have direct links for each option that can then take the user back to the omnibar with the selected option partially filled out
> > for help modal, for options that take in addition param, show the param. e.g. goto experiment  and kill notebook
> 
> > break up combo words for commands. e.g. pauseExperiment to pause experiment
> 
> the only downside I see to this is that it'd be a bit slower to access as you can't get the full picture without getting the pause fully in. (that might be where tab to complete is helpful since enter to complete can risk executing something). This might also be addressed by allowing the user at each step to search through the fully resolved branches (same output as help) instead of just the immediate children (REF A)
> 
After playing around with this more, I understand the limitations. The ideal thing would be to explore beyond just the one level deep and see more options that can potentially match. `pau` to be able to show options `pause experiement` and `pause trial` (made up). But that's probably not trivial to implement, but maybe we can provide that functionality as a future improvement?

> > consistent command pattern of "action" followed by "noun" (or could be other formats, but what you have makes sense IMO). e.g. instead of dev serverAddress reset, reset serverAddress.
> 
> This makes sense. I mainly put it the other way to 1. better organize it: eg all serverAddress operations are under serverAddress 2. keep the tree narrower closer to the root 3. better discoverability: by looking at the nodes under dev you can't know what `dev reset` is but on the other hand looking at `dev serverAddress` you can guess operations relating to serverAddress can be found there.
> 
I think this is related to the previous point, where if we can extend our look-ahead algorithm, we won't have to break a pattern to make it more useable.

> > using Tab to select the highlighted option (after using the arrow keys) but NOT execute the option. Where as Enter would actually fire off the command if applicable. So if user typed hel and the highlighted option is help, hitting Enter would bring up the help modal, whereas Tab would just show help typed into the omnibar but not yet fired. So thinking of Tab as just an option selector, and Enter as both an option selector and also execution of the command.
> 
> this is a missing feature from the omnibar and to add it I think would require a nontrivial change to the package.
> 
I see, yep not worth to do right now.

> > upon a match where there are options being shown, show the full command instead of just the next available word. So if I type goto , it would show goto **experiment**, goto **cluster**, etc (with the shared portion showing in lighter great, so the auto complete portion would be more emphasized)
> 
> doable but I'm not sure how much it'll add to the user experience. I think we could use the space better if we showed a description of what it would do instead of repeating what the user has just typed in the omnibar and can see right above the presented options.
> 
I think the full command match is more important to have when we are fuzzy matching across multiple node levels, and not just one level deep. If the user thinks of a command from a noun, it can better aid them. For example if they want to know what they can do with an experiment, they can type `exp` and the full commands `kill **exp**eriment <id>`, `pause **exp**eriment <id>`, `pause all **exp**eriments`, etc will show up. The motivation for this is to really make the omnibar more flexible and not force the user to think about what they want to do in a specific approach (which we have the WebUI to be there for).

> > change "no sub branches: exit" to something like "no options matched". IMO, also thinking that it might be better to prevent Enter from dismissing the omnibar (only allowing escape key at this point)
> 
> Hmm do you think it probably serves our users better if it always stays around as opposed to sometime auto closing after a commenad? :thinking . If escape wasn't so far away from the home row i'd feel better about not auto closing. maybe if I modify our hotkeys to allow closing with ctrl+space it'd help here
> 
This one is subjective and after looking at other omnibars, they generally always have a go-to / default option so there is not an `unmatched` option. So the way I'm thinking about this is that the user came to the omnibar for a reason. And if they go through a flow and ends up not finding a path, the question I asked myself is, would the user want to just end it there or try another path? I was leaning more towards the 2nd option. They also have the option click on any part of the overlay to dismiss in addition to the escape key.

> > Nice to haves:
> > support for keyboard shortcuts for available options (added a screenshot below)
> 
I saw the return icon thanks! 👍  we can place other shortcuts in the future!

> TODOs considering the suggestions, the work estimates, and impact estimates:
> 
> * [x]  merge mousehover and is selected
> * [x]  maybe remove auto closing
> * [ ]  search against fully resolved branches [could get expensive]
> * [ ]  maybe break commands such as pauseExperiment

webui/react/src/utils/storage.ts
hkang1: I see, no how you have it is good.

webui/react/src/omnibar/Omnibar.tsx
hkang1: just noting what we talked:
- potentially building the string from tree node paths
- using the `value` prop for the omnibar

hkang1 (approved):
This look great and really looking forward to using it! Thanks for working on this on product week!

hamidzr:
reminder: launch tensorboard fromExperiments label FIXME

webui/react/src/omnibar/Tree.ts
hamidzr: > I think the full command match is more important to have when we are fuzzy matching across multiple node levels,

yeah I agree we'll def want it for that

> I was leaning more towards the 2nd option. They also have the option click on any part of the overlay to dismiss in addition to the escape key.

I removed that.

hamidzr:
z index checklist: 

new ui dialog box.
side navigation
antd notifications

webui/react/src/omnibar/Omnibar.tsx
hamidzr: > just noting what we talked:
> 
> potentially building the string from tree node paths

We can do this part. If we don't immediately need this part I think it'll be better if we do this
together with multi level query matching as I suspect these two would affect eachother.

> using the value prop for the omnibar

since we don't have access to let the omnibar's input update this value as the user is typing I couldn't find a good way to use this.

hamidzr: I saved this idea in a code comment

hamidzr:
@hkang1 does the z-index changes make sense?  https://github.com/determined-ai/determined/pull/2308/commits/d25c6ede2a63137372ab74b0115b51f7e6e23dcd

hkang1:
> @hkang1 does the z-index changes make sense? [d25c6ed](https://github.com/determined-ai/determined/commit/d25c6ede2a63137372ab74b0115b51f7e6e23dcd)

This works for me. We'll be talking about how we want to manage z-indices soon enough to address this project-wide

PR #2311, opened 2021-05-03
fix: system metric chart fix to support milliseconds [DET-5348]

apizzini:
## Description

Fixing this chart as per uPlot documentation instructions: https://github.com/leeoniya/uPlot/tree/master/docs#data-format

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/Profiles/SystemMetricChart.tsx
+++ b/webui/react/src/pages/TrialDetails/Profiles/SystemMetricChart.tsx
@@ -40,20 +40,27 @@ const SystemMetricChart: React.FC<Props> = ({ filters, trial }: Props) => {
       axes: [
         {
           space: (self, axisIdx, scaleMin, scaleMax, plotDim) => {
-            const rangeSecs = scaleMax - scaleMin;
-            const pxPerSec = plotDim / rangeSecs;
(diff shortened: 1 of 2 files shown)


hkang1 (approved):
the code looks good, but I wasn't able to turn on system metrics for experiments. Is there a documented way of enabling this feature?

apizzini:
Thanks for checking @armandmcqueen.

The zooming change is not really coming from this PR but from another one we merged some days ago.

I agree that the prev zooming was more intuitive on charts that are displaying time series data.

@hkang1 what do you think about adding a chart option to be able to select if a chart should have (or not) Y zooming enabled? (while keeping X axis zoom always enabled).

armandmcqueen:
Ah in that case this PR is approved, but it would definitely be nice to have that chart zoom mode as an option.

hkang1:
> @hkang1 what do you think about adding a chart option to be able to select if a chart should have (or not) Y zooming enabled? (while keeping X axis zoom always enabled).

Ah I see, that makes sense for this case. 👍

apizzini:
@armandmcqueen, @hkang1: I've created [a separated issue for the zooming fix](https://determinedai.atlassian.net/browse/DET-5395). Merging this.

PR #2315, opened 2021-05-03
feat: support push metric APIs internally [DET-5215]

stoksc:
## Description
This adds a push API for trials to push metrics to the master.

## Test Plan
- [x] hit it a bunch, make sure it works (will be tested automatically by all trial impls eventually).

## Commentary (optional)
This is just the training metric API, I'll begin to build validations and checkpoints on top of it.

## Checklist


proto/src/determined/trial/v1/trial.proto
```diff
@@ -150,3 +150,30 @@ message TrialProfilerMetricsBatch {
...
+    EXITED_REASON_USER_REQUESTED_STOP = 2;
+  }
+  // The reason for the exit.
+  ExitedReason reason = 1;
+}
+
+// The training metrics for some duration of training.
+message TrainingMetrics {
+  // The number of batches trained on when these metrics were reported.
+  int32 batches_trained = 1;
```
shiyuann, line 172: q: are those two fields both required? If they are not set, what happens?

proto/src/determined/api/v1/api.proto
```diff
@@ -509,6 +509,78 @@ service Determined {
       tags: "Internal"
     };
   }
+  // Get the absolute training length until the searcher needs the trial to
+  // report a validation.
+  rpc GetTrialSearcherTrainUntil(GetTrialSearcherTrainUntilRequest)
+      returns (GetTrialSearcherTrainUntilResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/trials/{trial_id}/searcher/train_until"
```
shiyuann, line 517: A trial shouldn't be aware of the existence of the searcher, right? It should just know it needs to early stop.
```suggestion
      get: "/api/v1/trials/{trial_id}/train_length"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -509,6 +509,78 @@ service Determined {
...
+      tags: "Internal"
+    };
+  }
+  // Reports to the searcher that the trial has completed the current
+  // requested amount of training with the given searcher validation
+  // metric.
+  rpc ReportTrialSearcherValidation(ReportTrialSearcherValidationRequest)
+      returns (ReportTrialSearcherValidationResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/trials/{trial_id}/searcher/report_validation"
```
shiyuann, line 529:
```suggestion
      post: "/api/v1/trials/{trial_id}/validation"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -509,6 +509,78 @@ service Determined {
...
+      tags: "Internal"
+    };
+  }
+  // Reports to the searcher that the trial has completed the current
+  // requested amount of training with the given searcher validation
+  // metric.
+  rpc ReportTrialSearcherEarlyExit(ReportTrialSearcherEarlyExitRequest)
+      returns (ReportTrialSearcherEarlyExitResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/trials/{trial_id}/searcher/early_exit"
```
shiyuann, line 542: early stopping is a more commonly used word in ML word if I am right

```suggestion
      post: "/api/v1/trials/{trial_id}/early_stop"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -509,6 +509,78 @@ service Determined {
...
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "Internal"
+    };
+  }
+  // For bookkeeping, updates the progress towards to current requested searcher
+  // training length.
+  rpc ReportTrialProgress(ReportTrialProgressRequest)
+      returns (ReportTrialProgressResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/trials/{trial_id}/searcher/report_progress"
```
shiyuann, line 554: Progress is the nature of a trial not a searcher.
```suggestion
      post: "/api/v1/trials/{trial_id}/progress"
```

master/static/migrations/20210429141534_add-trial-metadata.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE public.trials
+    ADD COLUMN run_start timestamp without time zone NULL,
+    ADD COLUMN metadata jsonb NOT NULL DEFAULT '{}'::jsonb;
+
+AlTER TABLE public.steps
+    ADD COLUMN total_inputs integer NOT NULL DEFAULT 0;
```
shiyuann, line 6: total_inputs is a bit ambiguous. It doesn't immediately deliver it is total_batches or total_inputs of the model

master/internal/trial.go
```diff
@@ -312,7 +312,7 @@ func (t *trial) Receive(ctx *actor.Context) error {
 	case trialWatchPreemption:
 		// Size 2; at most 2 messages can be sent and we don't want to block or lose them.
 		w := make(chan bool, 2)
-		if t.PendingGracefulTermination {
+		if t.PendingGracefulTermination || t.experimentState != model.ActiveState {
```
shiyuann, line 315: can you explain a bit why do you need this line?

master/internal/db/postgres.go
```diff
@@ -1447,6 +1485,114 @@ func (db *PgDB) UpdateStep(
 	if err != nil {
 		return errors.Wrapf(err, "error finding step (%v, %v) to update", trialID, totalBatches)
 	}
+	return db.updateStep(step, newState, metrics)
+}
+
+// AddTrainingMetrics adds a completed step to the database with the given training metrics.
```
shiyuann, line 1491: Can you briefly explain the idea here? why do you need to migrate steps on the fly rather than in migration scipt?

master/internal/db/postgres.go
```diff
@@ -885,61 +888,43 @@ WHERE id = :id`
...
 // DeleteExperiment deletes an existing experiment.
 func (db *PgDB) DeleteExperiment(id int) error {
-	tx, err := db.sql.Begin()
-	if err != nil {
-		return errors.Wrap(err, "error starting transaction")
-	}
-	defer func() {
-		if tx == nil {
-			return
+	return db.withTransaction("delete experiment", func(tx *sqlx.Tx) error {
```
shiyuann, line 891: q: Is this relevant?

master/internal/db/database.go
```diff
@@ -7,5 +7,8 @@ import (
 // ErrNotFound is returned if nothing is found.
 var ErrNotFound = errors.New("not found")
 
+// ErrTooMany is returned if too many rows are affected.
+var ErrTooMany = errors.New("too many")
```
shiyuann, line 11:
```suggestion
var ErrTooManyRowsAffected = errors.New("too many rows are affected")
```

master/internal/api_trials.go
```diff
@@ -546,17 +550,12 @@ func (a *apiServer) TrialPreemptionSignal(
 	}
 
 	id := uuid.New()
-	ret, err := a.askAtDefaultSystem(trial, trialWatchPreemption{id: id})
-	if err != nil {
+	var signal <-chan bool
```
shiyuann, line 553: A mixed use of go channel and actor actually drives me crazy. We can create an actor for each `TrialPreemptionSignalRequest` to listen to the trial actor and make the trial actor notify all of the listening actors of the preemption signals.

shiyuann: You don't have to change it now. I just thought there was a more understandable way to write this but you might disagree.

master/internal/db/postgres.go
stoksc: nah i just made this code use `db.withTransaction` when i changed the sig to take a sqlx.Tx (since it was so easy to do).

master/internal/db/postgres.go
stoksc: There is no migration here, which part are you referring to?

master/internal/trial.go
stoksc: previously if you started listening to preemption signals for a paused trial it would say, no you dont have to preempt but wait and I'll tell you when. a trial paused trial (or anything but active) should always just tell the caller to exit.

master/static/migrations/20210429141534_add-trial-metadata.up.sql
stoksc: yeah I agree. total_records?

proto/src/determined/api/v1/api.proto
stoksc: @rb-determined-ai and I talked about this, he felt the same so yeah, this is fine with me.

proto/src/determined/api/v1/api.proto
stoksc: ah, see, well this one is different, because we're going to have separate APIs to report a searcher validation and to report validation metrics just to be persisted.

stoksc, line 529:
```suggestion
      post: "/api/v1/trials/{trial_id}/searcher/searcher_validation"
```

proto/src/determined/api/v1/api.proto
stoksc: yeah I agree, sgtm

proto/src/determined/trial/v1/trial.proto
stoksc: i believe batches_trained (as a total amount) is required. @rb-determined-ai had more opinions about the python side after what is required/what is not.

master/internal/api_trials.go
stoksc: ah, i had a really long discussion with @rb-determined-ai about this. honestly, channels or actors 🤷 it really doesn't make a huge difference, but then you have an extra actor vs an extra channel for each caller, and you have to write the little actor. after writing a bit more go concurrency code, i think this is probably more approachable.

stoksc: I took the actor approach initially with trial log streaming and backed out of it because it added so much unnecessary complexity compared to channels.

stoksc: The channel code is more "go native". If we hire more go systems engineers, this will be more approachable right off, I bet.

proto/src/determined/api/v1/api.proto
shiyuann: okay. I actually was lazy. This comment contains two things. One thing is the searcher portion of the path. The other thing is that `report_validation` isn't really restful pattern in API design unless we don't care about the restful design general practice.

proto/src/determined/api/v1/api.proto
shiyuann: This progress can also be used to be seen by users although we don't support it, right?

master/static/migrations/20210429141534_add-trial-metadata.up.sql
shiyuann: yeah. that makes sense in using the units to let us know it's the number of examples. but I thought we used batches?

master/internal/trial.go
shiyuann: is this conflict with our plan to change the states of the experiments/trials?

master/internal/db/postgres.go
shiyuann: sry. I might misunderstand the logic here. Because in order to add a training metrics, you roll back all the training metrics and validations. At least the validations seem very irrelevant. Why do you need to add interdependency here?

proto/src/determined/api/v1/api.proto
stoksc: > The other thing is that report_validation isn't really restful pattern in API design unless we don't care about the restful design general practice.

What precisely do you mean by restful pattern? Whatever you mean, I just meant it to be a plain old RPC. I had CQRS in mind a bit, with RPCs that affect the state of the system and queries that retrieve it. I do like the REST pattern, i just dont think it fits everything perfectly (some how k8s has made it work though, I guess that's a good counterexample that it can be made to literally fit anything lol).

master/internal/db/postgres.go
shiyuann: I was asking why deleting experiments relevant to supporting push metrics API. why do you need to change the code in this function?

master/internal/api_trials.go
shiyuann: > but then you have an extra actor vs an extra channel for each caller, and you have to write the little actor. after writing a bit more go concurrency code, i think this is probably more approachable.

The idea of using actor is to avoid adding synchronization (channel blocking) so that one day it wouldn't get out of control. I think adding one actor wouldn't add more effort in this case. It just needs more thinking. We used to use channels everywhere and moved to an actor system for a reason.

> I took the actor approach initially with trial log streaming and backed out of it because it added so much unnecessary complexity compared to channels.

IDK the log streaming but this is definitely not the case here. Adding one actor type alone doesn't add complexity compared to the complexity added by one synchronization point. It's just hard to understand unless you look at the code in all places that involve this synchronization.

> The channel code is more "go native". If we hire more go systems engineers, this will be more approachable right off, I bet.

This is definitely not true. I guess the "go native" way isn't really appropriate in our applications. The mixed use of "go native" channels and actors just makes things more confusing.

shiyuann: @stoksc can we at least use typed channels here so that we could use search to help understand the codebase?

proto/src/determined/api/v1/api.proto
shiyuann: talked offline. we will use the REST pattern.

proto/src/determined/api/v1/api.proto
stoksc: > If it's early stopped, it's early stopped in the middle of the whole planned precedure.

yeah i suppose you could report progress as a percent of max_length and if it is early stopped it just never reaches 100%? is this what you mean.

master/static/migrations/20210429141534_add-trial-metadata.up.sql
stoksc: i think sprinkle proposes to have both available, but this one is optional

master/internal/trial.go
stoksc: > is this conflict with our plan to change the states of the experiments/trials 

I have no plan to do this, but I did have a plan to add substates under active.

> Why isn't it handled in the termination of the trial?

because the caller may not be listening yet when the trial terminates.

master/internal/db/postgres.go
stoksc: talk offline, gonna change all this and store multiple runs

master/internal/db/postgres.go
stoksc: i'm confused, i answered that question above.

> nah i just made this code use db.withTransaction when i changed the sig to take a sqlx.Tx (since it was so easy to do).

nah is it not _strictly_ relevant but i just broke the signature of `db.deleteSnapshotsForExperiment` to require a `sqlx.Tx` not a `sql.Tx`so i moved it around to use `db.withTransaction` instead. you can checkout the code and change `withTransaction` to use a `sql.Tx` and you'll see why I made this change

master/internal/api_trials.go
stoksc: > The idea of using actor is to avoid adding synchronization (channel blocking) so that one day as you add more channels it wouldn't get out of control.

There is no synchronization here, we use a buffered channel that is the size of the max number of channels. The flipside of the actor system, and one we luckily don't hit very often, is that it's essentially like using `make(chan interface{}, ∞)` so we have no backpressure mechanism. If you wanted to, a send in a for loop to a slow actor would likely OOM the system.

> The other benefit is that since actor messages are strictly typed you can just search the message type to know how different parts of the actors work together.

This totally isn't true and is one of my biggest pet peeves about the actor system - there are multiple actors that receive a `model.State` from multiple places. It is impossible to statically reason about where those messages come from the way you describe. At least with my code, I send a specifically typed actor message (trialWatchPreemptionSignal that responds with a chan bool. It's fairly easily to trace where that chan comes from and what is in it.

> I think adding one actor wouldn't add more effort in this case. It just needs more thinking. We used to use channels everywhere and moved to an actor system for a reason.

I didn't know this, but I think for a case as simple as this, this is simpler personally. For long running processes that need to send and receive messages in an addressable way, actors totally make sense, but this is just propagating a bool 2 times.

> IDK the log streaming but this is definitely not the case here. Adding one actor type alone doesn't add complexity compared to the complexity added by one synchronization point. It's just hard to understand unless you look at the code in all places that involve this synchronization.

like I said, there is no synchronization point to do with channels like you describe. the handler asks the actor for a chan to receive signals on (the _ask_ is a synchronization point between the handler and the actor) and then the trial sends and handler receives at whim on the channel without any possible blocking.

> can we at least use typed channels here so that we could search to help understand the codebase?

what do you mean typed channels? a channel that sends a `type PreemptSignal bool` or something instead of bool?

stoksc: > We used to use channels everywhere and moved to an actor system for a reason.

Out of curiosity I git blamed pedl to see and found no evidence of this. It looked like we used globals accessed by API endpoints under a mutex and moved straight to an actor system.

stoksc: Some small bits too:
* it's much nicer IMO too to `select` off a grpc `ctx.Done()` and whatever is chan feeding data to your API endpoint (and whatever other signals), than to need to check `ctx.Err()` or make a go routine that selects off `ctx.Done()` and tells your intermediary actor.
* it is good to have to explicitly handle backpressure (channels have bounded cap), than "actor magic" where someone could easily oom the master filling up an actor's inbox.

proto/src/determined/api/v1/api.proto
shiyuann: exactly what I mean.

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,78 @@ message TrialEarlyExit {
...
+  google.protobuf.Struct metrics = 5;
+}
+
+// The checkpoint metadata for some trial checkpoint.
+message CheckpointMetadata {
+  // XXX: Clean up required on all these.
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      required: [
+        "trial_id",
```
rb-determined-ai, line 201: is there a plan for how we will migrate checkpoints away from depending on training-specific values, like trial_id?

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,78 @@ message TrialEarlyExit {
...
+    }
+  };
+  // The trial associated with these metrics.
+  int32 trial_id = 1;
+  // The trial run associated with these metrics.
+  int32 trial_run_id = 2;
+  // The number of batches trained on when these metrics were reported.
+  int32 batches_trained = 3;
+  // The number of batches trained on when these metrics were reported.
+  int32 records_trained = 4;
```
rb-determined-ai, line 169: why not add an optional epochs trained as well?  That is the third x-axis that would be very natural in our current system.

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,78 @@ message TrialEarlyExit {
...
+    json_schema: {
+      required: [ "trial_id", "trial_run_id", "batches_trained", "metrics" ]
+    }
+  };
+  // The trial associated with these metrics.
+  int32 trial_id = 1;
+  // The trial run associated with these metrics.
+  int32 trial_run_id = 2;
+  // The number of batches trained on when these metrics were reported.
+  int32 batches_trained = 3;
```
shiyuann, line 167: As discussed, this should be 
```suggestion
  int32 total_batches = 3;
```

proto/src/determined/trial/v1/trial.proto
shiyuann, line 169: As discussed, this should be 
```suggestion
  int32 total_records = 4;
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,78 @@ message TrialEarlyExit {
...
```
shiyuann, line 188: same

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,78 @@ message TrialEarlyExit {
...
+  // The number of batches trained on when these metrics were reported.
+  int32 records_trained = 4;
```
shiyuann, line 190: same

master/internal/db/postgres.go
```diff
@@ -1441,6 +1452,211 @@ func (db *PgDB) UpdateStep(
...
+	))
+`, trialID).Scan(&endTime); err != nil {
+		return time.Time{}, errors.Wrap(err, "getting prior training metrics")
+	}
+	return endTime, nil
+}
+
+// UpdateStepByPriorBatches updates an existing step. Fields that are nil or zero are not
+// updated.  end_time is set if the step moves to a terminal state.
+func (db *PgDB) UpdateStepByPriorBatches(
```
shiyuann, line 1647: This needs to be removed since we want to be consistent in using `total_batches` as the indexes.

master/internal/db/postgres.go
```diff
@@ -1430,6 +1426,21 @@ WHERE trial_id = $1 AND total_batches = $2`, &step, trialID, totalBatches); err
 	return &step, nil
 }
 
+// StepByPriorBatches looks up a step by (TrialID, PriorBatches) pair,
+// returning an error if none exists.
+func (db *PgDB) StepByPriorBatches(trialID, priorBatches int) (*model.Step, error) {
```
shiyuann, line 1431: This one should be removed too

master/internal/db/postgres.go
```diff
@@ -1441,6 +1452,211 @@ func (db *PgDB) UpdateStep(
...
+WHERE trial_id = $1
+`, m.TrialId).Scan(&cRunID); {
+		case err != nil:
+			return errors.Wrap(err, "querying current run")
+		case int(m.TrialRunId) < cRunID:
+			return api.AsErrBadRequest("stale training metrics for run %d during %d", m.TrialRunId, cRunID)
+		}
+
+		if _, err := tx.ExecContext(ctx, `
+UPDATE raw_steps SET archived = true
```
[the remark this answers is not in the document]
stoksc: > I think using archived is not as good as just simply overriding the metrics by those of the latest run.

This is "simply overriding the metrics by those of the latest run". How would you do it? Just with delete the same as this update?

master/internal/db/postgres.go
```diff
@@ -1,13 +1,18 @@
...
 	"database/sql"
 	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
 	"time"
 
+	"github.com/determined-ai/determined/master/internal/api"
+
+	"github.com/determined-ai/determined/proto/pkg/trialv1"
```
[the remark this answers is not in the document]
stoksc: eh, i'd rather do exactly what `goimports` does and no more, I don't see the point in doing more.

master/internal/db/postgres.go
```diff
@@ -1,13 +1,18 @@
...
 import (
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
 	"time"
 
+	"github.com/determined-ai/determined/master/internal/api"
```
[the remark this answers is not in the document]
stoksc: it wouldn't compile if it did

master/pkg/model/experiment.go
```diff
@@ -295,8 +295,10 @@ func NewTrial(
 // Step represents a row from the `steps` table.
 type Step struct {
 	TrialID      int        `db:"trial_id"`
+	TrialRunID   int        `db:"trial_run_id"`
 	ID           int        `db:"id"`
 	TotalBatches int        `db:"total_batches"`
+	TotalInputs  int        `db:"total_inputs"`
```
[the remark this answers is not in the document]
stoksc: yeah just haven't done the rename yet

proto/src/determined/api/v1/api.proto
```diff
@@ -556,6 +556,39 @@ service Determined {
       tags: "Internal"
     };
   }
+  // Record training metrics for specified training.
+  rpc ReportTrialTrainingMetrics(ReportTrialTrainingMetricsRequest)
+      returns (ReportTrialTrainingMetricsResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/trials/{training_metrics.trial_id}/training_metrics"
```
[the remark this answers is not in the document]
stoksc: talked offline, keep it this way for now while it is internal?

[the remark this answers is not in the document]
shiyuann: keeping just as is is okay with me

master/internal/db/postgres.go
```diff
@@ -1441,6 +1452,211 @@ func (db *PgDB) UpdateStep(
...
+		}); err != nil {
+			return errors.Wrap(err, "inserting checkpoint metadata")
+		}
+		return nil
+	})
+}
+
+// derivePriorWorkloadEndTime approximates the start time of currently reported metrics since
+// resource allocation uses these times.
+func derivePriorWorkloadEndTime(ctx context.Context, tx *sqlx.Tx, trialID int) (time.Time, error) {
```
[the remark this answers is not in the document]
stoksc: cc @rb-determined-ai again

stoksc:
> Names: batches_trained, total_batches, and prior_batches. And make a ticket to push for consistency afterward.

> Not using a separate column archived since setting all prior metrics to be achieved when you just want to add one metrics sounds like digging a performance hole for the future us to easily step in. More importantly, the overriding chain should be explicit and we don't need additional states to maintain this chain in favor of preventing potential inconsistency.There are some magics in the adding training metrics function. We should avoid doing those magics.

stoksc:
resp to PR comment @shiyuann 

> Names: batches_trained, total_batches, and prior_batches. And make a ticket to push for consistency afterward.

Is this just to make a ticket or change them all right now?

> Not using a separate column archived since setting all prior metrics to be achieved when you just want to add one metrics sounds like digging a performance hole for the future us to easily step in. More importantly, the overriding chain should be explicit and we don't need additional states to maintain this chain in favor of preventing potential inconsistency.

Specifically to 

> setting all prior metrics to be achieved when you just want to add one metrics sounds like digging a performance hole for the future us to easily step in

This is no more costly than rollback trial and it happens at the same time/serves the same purpose. To

> More importantly, the overriding chain should be explicit and we don't need additional states to maintain this chain in favor of preventing potential inconsistency.

Are you advocating for an explicit rollback endpoint? The additional state is just in lieu of a DELETE api - since this is a public facing API, i think a POST that can permanently delete a bunch of data is a disaster but an archive bothers me less.

> There are some magics in the adding training metrics function. We should avoid doing those magics.

Not sure which these are?

master/internal/db/postgres.go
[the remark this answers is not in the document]
stoksc: > When the API needs to query the metrics, it just get all the metrics of the largest run id.

This was actually my first iteration on the PR, but I wrote the query and it is was too costly, so I decided to use archived so I can query all valid steps by one bool instead. Here was the query (well, a function to replace steps -> steps_from_valid_runs() in every query, with some massaging to get queries with a `FROM steps WHERE trial_id = $1` to have to `trial_id` filter pushed down into the function.. but even then some hpviz queries didn't cooperate):

```
CREATE OR REPLACE FUNCTION steps_from_valid_runs ()
    RETURNS SETOF steps
    LANGUAGE SQL IMMUTABLE
AS $$
	SELECT s.*
	FROM steps s
	INNER JOIN (
		SELECT
			id,
			trial_id
		FROM (
			SELECT
				id,
				run_id,
				trial_id,
	        	max(s.run_id) OVER (PARTITION BY s.trial_id ORDER BY s.total_batches ASC) AS max_run_id
			FROM steps s
		) q1
		WHERE q1.run_id = q1.max_run_id
	 ) q ON s.id = q.id and s.trial_id = q.trial_id;
$$;
```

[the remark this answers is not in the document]
stoksc: But the heart of the query
```
SELECT *
FROM (
	SELECT *,
		max(s.run_id) OVER (PARTITION BY s.trial_id ORDER BY s.total_batches ASC) AS max_run_id
	FROM steps s
) q1
WHERE q1.run_id = q1.max_run_id
```
wasn't too bad cost wise, it was just really annoying to find a way to write it in a drop in way where I could guarantee we would blow a query plan up.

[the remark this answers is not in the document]
stoksc: Once I realized I could just use an archive column, it was so much simpler I just ditched this approach.

master/internal/db/postgres.go
```diff
@@ -1441,6 +1452,211 @@ func (db *PgDB) UpdateStep(
...
+UPDATE raw_steps SET archived = true
+WHERE trial_id = $1
+  AND trial_run_id < $2
+  AND total_batches >= $3;
+`, m.TrialId, m.TrialRunId, m.BatchesTrained); err != nil {
+			return errors.Wrap(err, "archiving training metrics")
+		}
+
+		if _, err := tx.ExecContext(ctx, `
+UPDATE raw_validations SET archived = true
```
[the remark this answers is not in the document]
stoksc: same reply here

master/internal/db/postgres.go
```diff
@@ -1441,6 +1452,211 @@ func (db *PgDB) UpdateStep(
...
+		// TODO(DET-5210): This can go away when step ID does.
+		var id int
+		if err := tx.QueryRowxContext(ctx, `
+SELECT max(id)
+FROM steps
+WHERE trial_id = $1`, m.TrialId).Scan(&id); err != nil {
+			return errors.Wrap(err, "querying next id")
+		}
+
+		startTime, err := derivePriorWorkloadEndTime(ctx, tx, int(m.TrialId))
```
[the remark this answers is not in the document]
stoksc: replied in thread on slack, I agree, I think now the SDK should handle deriving it

master/internal/db/postgres.go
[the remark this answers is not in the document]
rb-determined-ai: Resolution here (from slack) was that we still don't want to delete old metrics implicitly when you post metrics from a new run.  And the steps to mark things as archived is equally efficient as marking them deleted, but it makes it trivially easy to pull out the logically merged metrics from multiple runs at read time.

praise: I think this solution is very elegant, once I realized how the archived column worked.

master/internal/db/postgres.go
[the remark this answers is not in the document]
rb-determined-ai: resolution on this (from slack) was that we agree that `start_time` shouldn't be calculated from step start/end times, and it will be easy to recalculate it on new information at another time.  The only real reason to keep start_time around was for resource allocation, which read the steps table, but that information can be more correctly calculated from the new runs table.

Also, by leaving the derived time here (which is correct in the current system, albeit fragile) we can land this work without having to rewrite resource allocation logic before landing this PR.

proto/src/determined/trial/v1/trial.proto
stoksc: i think we will remove the "required" here which is a backwards compatible change. other than that, i think we can figure it out when we get there.

master/static/migrations/20210513134524_updates-for-push-api.down.sql
```diff
@@ -0,0 +1,30 @@
...
+    DROP COLUMN metadata;
+
+DROP TYPE public.run_type;
+
+DROP TABLE public.runs;
+
+AlTER TABLE public.steps
+    DROP COLUMN total_inputs,
+    DROP COLUMN trial_run_id,
+    ADD CONSTRAINT steps_trial_total_batches_unique UNIQUE (trial_id, total_batches);
```
shiyuann, line 11: Shouldn't this also include `trial_run_id`?

master/static/migrations/20210513134524_updates-for-push-api.down.sql
```diff
@@ -0,0 +1,30 @@
...
+
+ALTER TABLE public.raw_steps RENAME TO steps;
+DROP VIEW steps;
+
+ALTER TABLE public.validations
+    DROP COLUMN total_inputs,
+    DROP COLUMN trial_run_id,
+    ADD CONSTRAINT validations_trial_total_batches_unique UNIQUE (trial_id, total_batches);
```
shiyuann, line 19: same

master/static/migrations/20210513134524_updates-for-push-api.down.sql
```diff
@@ -0,0 +1,30 @@
...
+
+ALTER TABLE public.raw_validations RENAME TO validations;
+DROP VIEW steps;
+
+ALTER TABLE public.checkpoints
+    DROP COLUMN total_inputs,
+    DROP COLUMN trial_run_id,
+    ADD CONSTRAINT checkpoints_trial_total_batches_unique UNIQUE (trial_id, total_batches);
```
shiyuann, line 27: same

master/internal/trial.go
```diff
@@ -1122,6 +1126,10 @@ func (t *trial) terminated(ctx *actor.Context) {
 
 	terminationSent := t.TerminationSent
 
+	if err := t.db.CompleteTrialRun(t.id, t.RunID); err != nil {
```
shiyuann, line 1129: Probably should add a reset operation to restarting a trial. Otherwise, there might be trials that don't have `end_time`

shiyuann (commented):
The only things left are:
1. there are some places that should use `total_records` rather than `total_inputs`.
2. removing end time from steps, validations, ...
3. fail the trial if writing to trial run table fails.

master/internal/trial.go
stoksc: can you clarify what you mean?

stoksc:
@shiyuann for this
> removing end time from steps, validations, ...

what do you mean?

master/internal/trial.go
shiyuann: Sorry I was confusing. This `terminated` function might not be called at all if a trial actor fails and exits. I was trying to think of one case that causes trial not to continue after restarting the master.

stoksc: > is if there are any cases that might fail the trial (if the trial actor exits) when it should just restart a trial run.

ah yeah.. hmm. i suppose if the actor panics we could end up with open ended runs which is bad. i'll add a cleanup query to end all open runs in actor.PostStop.

proto/src/determined/trial/v1/trial.proto
stoksc: wait, removing required in a response though isn't backwards compatible.. hmm. idk, i think we're gonna need a breaking change for that tbh, either that or we make v2 apis when we promote this out of internal

shiyuann (approved):
Can we also run GPU tests before merging this PR?

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,83 @@ message TrialEarlyExit {
   // The reason for the exit.
   ExitedReason reason = 1;
 }
+
+// The training metrics for some duration of training.
+message TrainingMetrics {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      required: [ "trial_id", "trial_run_id", "batches_trained", "metrics" ]
```
stoksc, line 159:
```suggestion
      required: [ "trial_id", "trial_run_id", "total_batches", "metrics" ]
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -148,3 +151,83 @@ message TrialEarlyExit {
...
+  google.protobuf.Struct metrics = 6;
+  // The batch metrics for this bit of training.
+  repeated google.protobuf.Struct batch_metrics = 7;
+}
+
+// The validation metrics at some point of training.
+message ValidationMetrics {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: {
+      required: [ "trial_id", "trial_run_id", "batches_trained", "metrics" ]
```
stoksc, line 182:
```suggestion
      required: [ "trial_id", "trial_run_id", "total_batches", "metrics" ]
```

PR #2316, opened 2021-05-03
ci: move CUDA 11 testing to more available GPUs

mackrorysd:
## Description

A100's have been unavailable most of the last week, they're expensive, and they're not strictly necessary to catch significant regressions in the CUDA 11 images (now that all our GKE clusters are on newer CUDA drivers). 

Basically this makes the CUDA 11 setup step almost identical to the vanilla GKE set up step except it applies the CUDA 11 images as the default, etc. The termination step is now identical so we just use a single definition now.

## Test Plan

https://app.circleci.com/pipelines/github/determined-ai/determined?branch=cuda11

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,20 +462,6 @@ commands:
           command: |
             gcloud container clusters delete <<parameters.cluster-id>> --quiet --region=<<parameters.region>>
 
-  terminate-gke-cluster-cuda-11:
-    parameters:
-      cluster-id:
-        type: string
-      zone:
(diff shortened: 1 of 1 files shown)


mackrorysd:
CI failure is just caused by the websockets incompatible change

ioga (approved):

shiyuann (approved):

mackrorysd:
Override the CI - changes were tested one the affected jobs. Other failures are unrelated and due to websockets API change (being addressed independently)

PR #2318, opened 2021-05-04
chore: add experiment name to breadcrumb on trial detail page [DET-5284]

armandmcqueen:
## Description

Include the experiment description in the WebUI breadcrumbs.

Removed the breadcrumbs from the Experiment Detail page.

Screen Shot 2021-06-02 at 12 01 07 PM

Added the experiment description to the Trial Detail page breadcrumbs that is truncated when it exceeds ~40~ 30 characters.

Screen Shot 2021-06-02 at 12 01 27 PM

Added a tooltip to the breadcrumb when the description is truncated.

Screen Shot 2021-06-02 at 12 01 40 PM

## Commentary (optional)

I played with mobile a bit. This implementation doesn't look very good. 

Screen Shot 2021-06-02 at 12 01 50 PM

Max length of 40 looks better than 30 on desktop IMO, but it really breaks the mobile view. The implementation looks better on mobile if we set the max description length of 15, but that isn't a good value for desktop. If the react code can differentiate between mobile and desktop, we could improve this, but I'm not sure how to do that/if that's a good practice.

However, there seems to be other problems with the header on mobile, so I don't think this PR makes anything worse

Screen Shot 2021-06-02 at 12 05 44 PM
Screen Shot 2021-06-02 at 12 05 52 PM

We could rethink how we do breadcrumbs as the breadcrumbs are currently very redundant on the Trial Detail page.

Screen Shot 2021-06-02 at 12 07 46 PM

Mobile looks much better if we just don't render the `/ Trial 71` part of the breadcrumb (forgot to take a screenshot of the better mobile look).

However, the breadcrumbs are also used by the Task Logs page where the info is possibly not redundant.

Screen Shot 2021-06-02 at 12 19 25 PM

If we don't think `/ Command 76b6` is important information, we can change the breadcrumb rendering to not render the last entry. All we would need to do is play with CSS since the black text color is currently based on `.ant-breadcrumb > span:last-child`.

### TODO

- Add release notes

## Checklist

### DIFF
--- a/webui/react/src/components/PageHeader.tsx
+++ b/webui/react/src/components/PageHeader.tsx
@@ -1,5 +1,4 @@
 import { Breadcrumb, Tooltip } from 'antd';
-import { Route } from 'antd/es/breadcrumb/Breadcrumb';
 import React from 'react';
 
 import { CommonProps } from 'types';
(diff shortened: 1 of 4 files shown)


webui/react/src/components/PageHeader.tsx
```diff
@@ -17,12 +17,49 @@ export interface Props extends CommonProps {
   title?: string;
 }
 
+const tooLongExperimentDescription = 40;
+
+const isExperimentIdWithDescCrumb = (breadcrumbName: string) => {
+  const crumbElements = breadcrumbName.split(' ');
+  // The experiment description may have spaces in it, so there should be at least 3 elements
+  if (crumbElements.length < 3
+      || crumbElements[0] !== 'Experiment'
```
apizzini, line 26: To make breadcrumb Tooltips something re-usable we could:
- extend the `BreadCrumbRoute` and add a `breadcrumbTooltip: string` property
- here, if there's a `breadcrumbTooltip`, render the `<Tooltip>` component
- in `TrialDetails.tsx` pass this additional prop for the exp breadcrumb item and keep the related logic there (add max 40 chars of description to `breadcrumbName`)

armandmcqueen: That's a much better implementation. Updated the PR to use that approach

armandmcqueen:
@hkang1 @apizzini  Updated the PR to have the improved implementation. Also added some info to the Commentary section about how this looks on mobile and how we might want to adjust the way we do breadcrumbs to look better on mobile.

apizzini (approved):
Great cleanup, thanks!

PR #2319, opened 2021-05-04
chore: update Docker images and AMIs for yogadl update

mackrorysd:
## Description

Update version of yogadl to work around websockets API change


shiyuann (approved):

mackrorysd:
Force pushed after discussion with Ilia - we've been doing other testing out of band. Affected by unrelated issues so proceeding with a release we'd like to make soon.

PR #2320, opened 2021-05-04
fix: improve tqdm rendering in the web ui

armandmcqueen:
## Description

In the WebUI, guess if the logs contain a TQDM line and skip rendering that line if so.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewerTimestamp.tsx
+++ b/webui/react/src/components/LogViewerTimestamp.tsx
@@ -121,6 +121,31 @@ const LogViewerTimestamp: React.FC<Props> = ({
   const addLogs = useCallback((addedLogs: TrialLog[], isPrepend = false): void => {
     const newLogs = addedLogs
       .map(log => {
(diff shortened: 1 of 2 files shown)


webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -121,8 +121,17 @@ const LogViewerTimestamp: React.FC<Props> = ({
   const addLogs = useCallback((addedLogs: TrialLog[], isPrepend = false): void => {
     const newLogs = addedLogs
       .map(log => {
+        let hide = false;
```
hkang1, line 124: **suggestion:** small refactor:

```
.map(log => {
  // If the line is probably a TQDM line, hide it
  const hide = log.message.includes('\r');
  const formattedTime = log.time ? formatDatetime(log.time, DATETIME_FORMAT) : '';
  return { ...log, formattedTime, hide };
})
.filter(logEntry => !logEntry.hide)
```

hkang1 (approved):
Thanks for making the trial logs more legible!

PR #2321, opened 2021-05-04
feat: delete some old commands APIs

shiyuann:
## Description

Delete some old APIs for commands, including:

- Launching, listing, deleting Commands.
- Launching, listing, deleting TensorBoards.
- Launching, listing, deleting Shells.
- Launching, listing, deleting Notebooks.

The event streaming APIs are kept.

## Test Plan

- Manually test all the related WebUI and CLI features.
- E2E test pipeline.

## Commentary (optional)

N/A

## Checklist


harness/determined/cli/shell.py
```diff
@@ -36,46 +33,44 @@ def start_shell(args: Namespace) -> None:
...
         args.master,
-        "shells",
+        "api/v1/shells",
         config,
         args.template,
         context_path=args.context,
         data=data,
-    )
+    )["shell"]
+    print(resp)
```
naren-determined, line 42: Not sure if this was used for debugging purposes, but it should be removed.

harness/determined/cli/shell.py
```diff
@@ -36,46 +33,44 @@ def start_shell(args: Namespace) -> None:
...
 
     if args.detach:
         print(resp["id"])
         return
 
-    command = None
+    ready = False
```
naren-determined, line 48: This logic below seems a bit complicated to me with the condition variable you are setting here. Why do you need to break out of the `with api.ws(args.master, "shells/{}/events".format(resp["id"])) as ws:` before calling `_open_shell`?

master/internal/api_command.go
```diff
@@ -47,14 +49,14 @@ func (a *apiServer) prepareLaunchParams(ctx context.Context, req *protoCommandPa
 	if req.Config != nil {
 		configBytes, err := protojson.Marshal(req.Config)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrapf(err, "failed to parse config: %s", configBytes)
```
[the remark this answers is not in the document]
shiyuann: No. `req.Config` is a `*pstruct.Struct`. `configBytes` is marshaled string.

harness/determined/cli/shell.py
shiyuann: good catch. removed it

harness/determined/cli/shell.py
shiyuann: I didn't change the logic of using the condition variable. I just removed the use of the `snapshot` field, which belongs to the old API. The logic is waiting for the state of a shell to change to running. It used to get the shell id from the websocket event. It is just changed to make a different call to the master.

naren-determined (approved):
Lgtm!

proto/src/determined/tensorboard/v1/tensorboard.proto
```diff
@@ -43,4 +44,8 @@ message Tensorboard {
   string service_address = 11;
   // The name of the resource pool the command was created in
   string resource_pool = 12;
+  // The config;
+  google.protobuf.Struct config = 13;
```
hamidzr, line 48: we've kept config out of the list endpoints for experiments as it could end up accountable for the majority of the response size and instead only provide it for when a specific experiment is requested. I realize we probably won't accumulate as much commands as we would experiments but wdy think about keeping consistent with that approach?

```
// Response to GetExperimentRequest.
message GetExperimentResponse {
    json_schema: { required: [ "experiment", "config" ] }
  };
  // The requested experiment.
  determined.experiment.v1.Experiment experiment = 1;
  // The experiment config.
  google.protobuf.Struct config = 2;
}
```

also reminder to update the `required` annotation if config is always expected to be there

proto/src/determined/shell/v1/shell.proto
```diff
@@ -41,4 +42,12 @@ message Shell {
   string username = 10;
   // The name of the resource pool the Shell was created in
   string resource_pool = 11;
+  // The config;
+  google.protobuf.Struct config = 12;
```
hamidzr, line 46: ditto (and the other 2 command/tasks

proto/src/determined/api/v1/tensorboard.proto
```diff
@@ -33,6 +33,8 @@ message GetTensorboardsRequest {
   int32 offset = 3;
   // Limit the number of tensorboards. A value of 0 denotes no limit.
   int32 limit = 4;
+  // Limit tensorboards to those that are owned by the specified users.
+  repeated string users = 5;
```
hamidzr, line 37: 👍🏼

proto/src/determined/notebook/v1/notebook.proto
```diff
@@ -39,4 +40,8 @@ message Notebook {
   string service_address = 11;
   // The name of the resource pool the Notebook was created in
   string resource_pool = 12;
+  // The config of the notebook;
+  google.protobuf.Struct config = 13;
+  // The exit status;
+  string exit_status = 14;
```
hamidzr, line 46: question: why not an `int` ? is this because we can't encode lack of a exist code?.

nit: wdy about exit code IMO it's a bit more explicit?

hamidzr: update: I see exist_status is defind as a string in our master code. in that case it'd make sense to me to keep it as string especially if we're not sure if it takes non numeric values.

master/internal/command/command.go
```diff
@@ -419,24 +415,45 @@ func (c *command) toCommand(ctx *actor.Context) *commandv1.Command {
...
+		ResourcePool:   c.config.Resources.ResourcePool,
+		Config:         protoutils.ToStruct(c.config),
+		ExitStatus:     exitStatus,
+		Addresses:      addresses,
+		AgentUserGroup: protoutils.ToStruct(c.agentUserGroup),
 	}
 }
 
 func (c *command) toTensorboard(ctx *actor.Context) *tensorboardv1.Tensorboard {
+	exitStatus := protoutils.DefaultStringValue
```
hamidzr, line 452: could we encode no status as `""` for our external facing apis? IMO it'd be more natural to parse than some string. wdy think

hamidzr (commented):
Nice work. I added some comments mostly focusing on the external-facing api changes

PR #2322, opened 2021-05-04
chore: release shouldn't check

stoksc:
## Checklist


ioga:
I may be late to the party but perhaps this problem stems from an excessive duplication of circleci code for `package-and-push-system-{dev,rc,release}`, so they drift away over time. instead, we could have one parametrized job where the parameter should just control whether it runs `make {publish-dev,publish,release}`

stoksc:
@ioga this moved to another PR but that's a good idea, I was just making the smallest change possible to fix for now. i'll do that.

PR #2323, opened 2021-05-04
build: always run go mod tidy after go mod download

shiyuann:
## Checklist


stoksc (commented):
lgtm with 1 q: can you make `package-and-push-system-rc` run all this that `package-and-push-system-release` runs too? 2nd q: can we just not run `go mod download` at all? Instead of tidy'ing the small changes it makes, I dont think it is necessary.

shiyuann:
@stoksc they all run `go-get-deps`, which runs `make -C master/agent/proto get-deps`, which runs the changes in this PR

stoksc:
@shiyuann the rc plan doesn't run agent and master `check` target though.

PR #2327, opened 2021-05-04
feat: tolerate missing GPU stats when running under MIG [DET-5387]

mackrorysd:
## Description

These statistics are not available when running on a partition of GPU (NVIDIA MIG) unless you're running with more privileges. They do not appear to be critical, so we just apply sensible defaults to avoid divide-by-zero, etc. and log a warning.

## Test Plan

Tested on a cluster with MIG devices: was able to provision 7 slots for every actual GPU, and run distributed training successfully.

### DIFF
--- a/harness/determined/gpu.py
+++ b/harness/determined/gpu.py
@@ -22,6 +22,19 @@ class GPU(NamedTuple):
     memoryUtil: float
 
 
+warned_fields = set()
+
+
+def float_or_default(fields: dict, key: str, default: float) -> float:
+    try:
+        return float(fields[key])
+    except ValueError:
+        if key not in warned_fields:
+            warned_fields.add(key)
+            logging.warning(f"Unable to get {key} from nvidia-smi")
+        return default
+
+
 def get_gpus() -> List[GPU]:
     try:
         proc = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


stoksc (approved):
lgtm

harness/determined/gpu.py
```diff
@@ -22,6 +22,19 @@ class GPU(NamedTuple):
     memoryUtil: float
 
 
+warned_fields = set()
```
shiyuann, line 25: what is this warned fields for? I didn't see it was used anywhere.

harness/determined/gpu.py
```diff
@@ -48,12 +61,13 @@ def get_gpus() -> List[GPU]:
                     GPU(
                         id=int(fields["index"]),
                         uuid=fields["uuid"].strip(),
-                        load=float(fields["utilization.gpu"]) / 100,
-                        memoryUtil=float(fields["memory.used"]) / float(fields["memory.total"]),
+                        load=float_or_default(fields, "utilization.gpu", 0.0) / 100,
```
shiyuann, line 64: Is a partial GPU one item in the list or one whole GPU is one item in the list?

mackrorysd: Each partial GPU is one item in the list. As far as Determined can tell, a partial GPU is a whole GPU, except that it will end up having fewer compute resources, less memory, and you won't (in Kubernetes at least) have the ability to look up utilization.

harness/determined/gpu.py
mackrorysd: It's used on line 32 and 33. Basically if we encounter a problem reading a certain metric, we'll log a warning. But we don't want to log the same warning again every time we hit that issue. We do a similar thing in a few other places already.

harness/determined/gpu.py
mackrorysd: Just to be clear, though, NVIDIA does not recommend using multipe partial GPUs for a single task. I tested using distributed training with multiple partial GPUs, and it works, but the recommend configuration would be to just use a whole GPU in that case. I can see MIG being used for a massive hyperparameter search with single-slot trials, or perhaps an inference-heavy workload. I don't think it will be useful for most distributed training.

PR #2328, opened 2021-05-04
chore: add searcher type to output of experiment APIs

armandmcqueen:
## Description

Include the searcher type as part of the experiment LIST and experiment GET APIs.

## Commentary (optional)

@hamidzr Should I make `searcher_type` a required field for the API? 

## Checklist

### DIFF
--- a/proto/src/determined/experiment/v1/experiment.proto
+++ b/proto/src/determined/experiment/v1/experiment.proto
@@ -68,6 +68,8 @@ message Experiment {
   string username = 10;
   // The resource pool the experiment was created in
   string resource_pool = 11;
+  // The type of searcher for the experiment
+  string searcher_type = 12;
 }
 
 // ValidationHistoryEntry is a single entry for a validation history for an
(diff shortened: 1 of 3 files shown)


proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -68,6 +68,8 @@ message Experiment {
   string username = 10;
   // The resource pool the experiment was created in
   string resource_pool = 11;
+  // The type of searcher for the experiment
+  string searcher_type = 12;
```
hamidzr, line 72: this is also used in the singular version of the API `experiments/:id` which uses another SQL file than the one you modified. could you update that one as well? 
we could probably refactor to avoid the need for almost duplicated query (sql files) to get ExperimentV1 object.

armandmcqueen: Fixed

hamidzr (approved):
LGTM. if this field would always be available. we should also mark it as "required" for swagger. (L38)

PR #2332, opened 2021-05-04
feat: store the original user submitted experiment config in db

hamidzr:
## Description

keep a record of the exact user submitted config (prior to applying defaults or any shimming)

## Test Plan

create an experiment with det cli and check the database

`select original_config, config, id from experiments;`

## Checklist

### DIFF
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -415,9 +415,8 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams) (
 	}
 
 	dbExp, err := model.NewExperiment(
-		config, modelBytes, params.ParentID, params.Archived,
+		config, params.ConfigBytes, modelBytes, params.ParentID, params.Archived,
 		params.GitRemote, params.GitCommit, params.GitCommitter, params.GitCommitDate)
-	dbExp.OriginalConfig = params.ConfigBytes
 	return dbExp, params.ValidateOnly, &taskSpec, err
 }
 
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -417,7 +417,7 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams) (
 	dbExp, err := model.NewExperiment(
 		config, modelBytes, params.ParentID, params.Archived,
 		params.GitRemote, params.GitCommit, params.GitCommitter, params.GitCommitDate)
-	dbExp.OriginalConfig = []byte(params.ConfigBytes)
+	dbExp.OriginalConfig = params.ConfigBytes
 	return dbExp, params.ValidateOnly, &taskSpec, err
 }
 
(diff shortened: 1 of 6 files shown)


master/internal/core_experiment.go
```diff
@@ -417,6 +417,7 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams) (
 	dbExp, err := model.NewExperiment(
 		config, modelBytes, params.ParentID, params.Archived,
 		params.GitRemote, params.GitCommit, params.GitCommitter, params.GitCommitDate)
+	dbExp.OriginalConfig = params.ConfigBytes
```
stoksc, line 420: why not have NewExperiment handle this instead?

hamidzr: good call. I was assuming it is being used elsewhere without direct user config input but it is not being used that way.

hamidzr: I updated it

master/internal/core_experiment.go
```diff
@@ -341,7 +341,7 @@ func (m *Master) patchExperiment(c echo.Context) (interface{}, error) {
 
 // CreateExperimentParams defines a request to create an experiment.
 type CreateExperimentParams struct {
-	ConfigBytes   string          `json:"experiment_config"`
+	ConfigBytes   *string         `json:"experiment_config"`
```
stoksc, line 344: why should we make this a pointer just to deref it sort of unsafely all over?

stoksc (commented):
one comment otherwise lgtm

master/internal/core_experiment.go
hamidzr: Hmm other than trying to save a few copies I was thinking this could allow us to potentially save an experiment with no user config 🤔 Maybe something we could use for forks?  But I also forgot to check for nil pointers

hamidzr: this doesn't seem to be on a perf critical code path so it might be better to stick with the simpler solution and copy?

stoksc: go strings are copy on modification afaik so sholdn't be any copies since this value is read only.

stoksc: even so yeah, i think it's fine

master/static/migrations/20210504130449_add-exp-original-config.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE public.experiments ADD COLUMN original_config character varying;
```
stoksc, line 1:
```suggestion
ALTER TABLE public.experiments ADD COLUMN original_config text;
```

stoksc (approved):
lgtm. potentially could be convenient to convert the yaml to json and store it as jsonb but i cant think of a reason we should ever be working with the original config in sql and no in go

hamidzr:
> lgtm. potentially could be convenient to convert the yaml to json and store it as jsonb but i cant think of a reason we should ever be working with the original config in sql and no in go

thanks for the review. That could be useful 🤔 The idea with going with varchar is avoid restricting the column type here eg what if we support other types of user config in future, or if that process introduced a small loss of information that we're unaware of

PR #2333, opened 2021-05-04
feat: add ArchiveSelectFilter replacing toggle for seeing archived experiments [DET-3925]

EmilyBonar:
## Description
Replaced "Show Archived" toggle in the experiments list with a dropdown [DET-3925]

## Test Plan
Before:
Screen Shot 2021-05-04 at 4 53 53 PM
After:
Screen Shot 2021-05-04 at 4 53 53 PM

[DET-3925]: https://determinedai.atlassian.net/browse/DET-3925

### DIFF
--- a/webui/react/src/components/ArchiveSelectFilter.stories.tsx
+++ b/webui/react/src/components/ArchiveSelectFilter.stories.tsx
@@ -12,5 +12,5 @@ export const Default = (): React.ReactNode => {
   const [ currentValue, setCurrentValue ] = useState<ArchiveFilters>('unarchived');
   return <ArchiveSelectFilter
     value={currentValue}
-    onChange={(newValue) => setCurrentValue(newValue as ArchiveFilters)} />;
+    onChange={(newValue) => setCurrentValue(newValue)} />;
 };
--- a/webui/react/src/components/ArchiveSelectFilter.stories.tsx
+++ b/webui/react/src/components/ArchiveSelectFilter.stories.tsx
@@ -1,0 +1,16 @@
+import React, { useState } from 'react';
+
+import ArchiveSelectFilter from 'components/ArchiveSelectFilter';
+import { ArchiveFilters } from 'types';
+
+export default {
+  component: ArchiveSelectFilter,
+  title: 'ArchiveSelectFilter',
+};
+
+export const Default = (): React.ReactNode => {
+  const [ currentValue, setCurrentValue ] = useState<ArchiveFilters>('unarchived');
+  return <ArchiveSelectFilter
+    value={currentValue}
+    onChange={(newValue) => setCurrentValue(newValue as ArchiveFilters)} />;
+};
--- a/webui/react/src/components/ArchiveSelectFilter.tsx
+++ b/webui/react/src/components/ArchiveSelectFilter.tsx
@@ -3,13 +3,15 @@ import { SelectValue } from 'antd/es/select';
 import React, { useCallback } from 'react';
 
 import SelectFilter from 'components/SelectFilter';
+import { ALL_VALUE, ArchiveFilters } from 'types';
 
 const { Option } = Select;
-
 interface Props {
-  onChange?: (value: string) => void;
-  value?: string;
+  onChange?: (value: ArchiveFilters) => void;
+  value?: ArchiveFilters;
 }
+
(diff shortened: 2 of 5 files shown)


webui/react/src/types.ts
```diff
@@ -457,7 +457,7 @@ export type TaskType = CommandType | 'Experiment';
 
 export interface ExperimentFilters {
   labels?: string[];
-  showArchived: boolean;
+  showArchived: string;
```
hamidzr, line 460: nit: could we use a stronger type than string here? maybe using the ALL_VALUE

webui/react/src/components/ArchiveSelectFilter.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { SelectValue } from 'antd/es/select';
+import React, { useCallback } from 'react';
+
+import SelectFilter from 'components/SelectFilter';
+
+const { Option } = Select;
+
+interface Props {
+  onChange?: (value: string) => void;
+  value?: string;
```
hamidzr, line 11: ditto on using a stronger type

webui/react/src/components/ArchiveSelectFilter.tsx
```diff
@@ -0,0 +1,36 @@
...
+    onChange(newValue.toString());
+  }, [ onChange ]);
+  return (
+    <SelectFilter
+      dropdownMatchSelectWidth={130}
+      label="Show"
+      style={{ minWidth: 100 }}
+      value={value}
+      onSelect={handleSelect}>
+      <Option key="All" value="All">All</Option>
```
hamidzr, line 29: nit: there is an ALL_VALUE constant in types module you could use here.

Also for the option values we try to stick to lowercase (it shows up the parameter lists if you're sharing a url)

hamidzr (commented):
This is looking good. I added some code-style comments but no blockers. This might be a good time to also add a storybook for this new component and get familiar with that. We're hoping to use those later for visual testing as well.

webui/react/src/components/ArchiveSelectFilter.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default {
+  component: ArchiveSelectFilter,
+  title: 'ArchiveSelectFilter',
+};
+
+export const Default = (): React.ReactNode => {
+  const [ currentValue, setCurrentValue ] = useState<ArchiveFilters>('unarchived');
+  return <ArchiveSelectFilter
+    value={currentValue}
+    onChange={(newValue) => setCurrentValue(newValue as ArchiveFilters)} />;
```
hamidzr, line 15: probably `as ArchiveFilters` wouldn't be needed

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -67,7 +67,7 @@ const STORAGE_LIMIT_KEY = 'limit';
 const STORAGE_SORTER_KEY = 'sorter';
 
 const defaultFilters: ExperimentFilters = {
-  showArchived: false,
+  showArchived: 'unarchived' as ArchiveFilters,
```
hamidzr, line 70: ditto

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -162,8 +162,9 @@ const ExperimentList: React.FC = () => {
 
     // archived
     const archived = urlSearchParams.get('archived');
-    if (archived != null) {
-      filters.showArchived = (archived === '1');
+    if (archived !== null &&
+      (archived === ALL_VALUE || archived === 'unarchived' || archived === 'archived')) {
```
hamidzr, line 166: nit: you could probably use the same check in  https://github.com/determined-ai/determined/pull/2333/files#diff-71021c13439e62579692aa701d15bba3fadc33b92a3a703a36fcb575a0615e67R22 (in your handleSelect)

hamidzr (commented):
one last blocking issue: 
we store user preferences in local storage and previously this field was stored as a boolean thus if a user has this preference saved when they upgrade they'll run into something like the screenshot below. Our users have run into similar issues before and so far we've solved it by asking them to reset their localstorage. I'm not sure what the best way to address this would be let's chat

hamidzr (approved):
LGTM great work.

PR #2342, opened 2021-05-05
chore: edits to expconf before enabling it

rb-determined-ai:
This PR contains the first 6 commits of #2310 without the massive "use expconf everywhere" commits.

Probably they are easier to review separately.


rb-determined-ai:
This PR doesn't have the generated code yet, for reviewability.

stoksc (approved):
lgtm

PR #2344, opened 2021-05-06
chore: migrate to use total_batches more in HP search viz.

mackrorysd:
## Description

These queries pre-date the total_batches column but it's definitely a better fit for the usage.

## Test Plan

All queries are covered by test_metrics.py


stoksc (approved):

master/internal/db/postgres_experiments.go
```diff
@@ -388,16 +388,16 @@ func (db *PgDB) ValidationMetricsSeries(trialID int32, startTime time.Time, metr
 	err error) {
 	rows, err := db.sql.Query(`
 SELECT 
-  (prior_batches_processed + num_batches) AS batches,
+  v.total_batches AS batches,
```
stoksc, line 391: 👌 ended up just catching this in my branch. glad you wrote such thorough tests.

PR #2345, opened 2021-05-06
chore: drop extra trial batches columns [DET-5405]

stoksc:
## Description
This change drops `steps.prior_batches_processed` and `steps.num_batches` in favor of always using `steps.total_batches_processed`.

## Test Plan
- [x] ensure no existing functionality is broken

## Checklist


proto/src/determined/trial/v1/trial.proto
```diff
@@ -32,23 +25,15 @@ message CheckpointWorkload {
   // Dictionary of file paths to file sizes in bytes of all files in the
   // checkpoint.
   map<string, int64> resources = 5;
-  // Number of batches in this workload.
-  int32 num_batches = 6;
-  // Number of batches processed prior to this workload.
-  int32 prior_batches_processed = 7;
+  // Total number of batches as of this workload's completion.
+  int32 total_batches = 6;
```
stoksc, line 29: This is technically a breaking change, though I think the only alternative is deprecating it and then breaking it next release, not this one.

proto/src/determined/trial/v1/trial.proto
```diff
@@ -59,12 +44,10 @@ message MetricsWorkload {
   determined.experiment.v1.State state = 3;
   // Metrics.
   google.protobuf.Struct metrics = 4;
-  // Number of inputs processed.
+  // DEPRECATED: Number of inputs processed.
```
stoksc, line 47:
```suggestion
  // Number of inputs processed.
```

master/pkg/model/experiment.go
```diff
@@ -294,41 +294,35 @@ func NewTrial(
...
-		State:                 CompletedState,
-		StartTime:             now,
-		EndTime:               &now,
-		NumBatches:            0,
-		PriorBatchesProcessed: 0,
+		TrialID:   trialID,
+		ID:        stepID,
+		State:     CompletedState,
+		StartTime: now,
+		EndTime:   &now,
```
hkang1, line 325: **question:** should this have the following line?
```
TotalBatches: 0,
```

master/static/srv/proto_get_trial_workloads.sql
```diff
@@ -38,18 +34,16 @@ checkpoints_vt AS (
...
         c.resources
       FROM checkpoints c
         INNER JOIN steps s ON c.trial_id = s.trial_id
         AND c.total_batches = s.total_batches
       WHERE c.trial_id = $1
     ) AS r1
 )
-SELECT v.validation::jsonb - 'total_batches' AS validation,
-  t.training::jsonb - 'total_batches' AS training,
-  c.checkpoint::jsonb - 'total_batches' AS checkpoint
```
hkang1, line 52: lol, it was really going out of the way to remove `total_batches`!

webui/react/src/utils/types.ts
```diff
@@ -281,10 +281,13 @@ export const getMetricValue = (workload?: Workload, metricName?: string): number
...
 
 export const getBatchNumber = (
-  data: {batch: number} | {numBatches: number, priorBatchesProcessed: number},
+  data: {batch: number} | {numBatches: number, priorBatchesProcessed: number}
+  | {totalBatches: number},
 ): number => {
   if ('batch' in data) {
     return data.batch;
+  } else if ('totalBatches' in data) {
+    return data.totalBatches;
```
hkang1, line 290: **suggestion:** I would just get rid of this `getBatchNumber` function entirely and change the only file that references this function `src/components/CheckpointModal.tsx`
* remove `getBatchNumber` from line 11 (remove the import from `utils/types`)
* change line 87 to be `const totalBatchesProcessed = checkpoint.totalBatches;`

hkang1 (approved):
WebUI portion looks good. One small suggestion to simplify (happy to do it myself also if you don't want to do it in this PR).

master/static/srv/proto_get_trial_workloads.sql
stoksc: lol seriously

master/pkg/model/experiment.go
stoksc: This is implied in go, you could make this change and keep NumBatches/PriorBatchesProcessed with no effect.

proto/src/determined/trial/v1/trial.proto
rb-determined-ai: @hamidzr wdyt?

rb-determined-ai (commented):
lgtm, but I don't want to be the one making calls about breaking changes in the rest APIs

webui/react/src/utils/types.ts
stoksc: @hkang1 the input checkpoint for that CheckpointModal is `CheckpointWorkloadExtended | CheckpointDetail`, of which `CheckpointDetail` doesn't have totalBatches, so I think that was the point? I'll just leave this to you if that's ok..

proto/src/determined/trial/v1/trial.proto
```diff
@@ -59,12 +44,10 @@ message MetricsWorkload {
...
   int32 num_inputs = 5;
-  // Number of batches in this workload.
-  int32 num_batches = 6;
-  // Number of batches processed prior to this workload.
-  int32 prior_batches_processed = 7;
+  // Total number of batches as of this workload's completion.
+  int32 total_batches = 6;
```
hamidzr, line 50: should we add the deprecated option as well for num_inputs https://developers.google.com/protocol-buffers/docs/proto3#options 

also, this probably won't have any effet since we're not directly using protobuf but, let's avoid reusing the old field numbers?

proto/src/determined/trial/v1/trial.proto
hamidzr: hmm we haven't set up a deprecation or API versioning strategy yet to follow in these situations. In lieu of those, we should make sure we put the changes in our release notes as breaking changes and use the API change GitHub label. In terms of our own usage, we migrate relatively easily but we don't really have a way of telling what our external users are doing with the APIs..  Since we still have that beta tag in our APIs I think we'd be okay to just remove. If we're not looking to immediately remove support for these fields that we are removing from the system leaving them exposed as deprecated for a release and removing them later doesn't seem viable.

hamidzr: also ditto on reusing field numbers

webui/react/src/utils/types.ts
hkang1: ah yeah, you're right. how you have it is great 👍

Issue #1805, opened 2021-01-18
Errors in log during agent shutdown on AWS


wbwatkinson:
The following is a snippet from the AWS logs... essentially, the master node is shutting down the 8x GPU agent due to inactivity. It appears that everything shutdown correctly, but the logs reported websocket and http errors.

As far as I can tell, everything worked correctly, but the logged information indicates that something was wrong, when there probably wasn't anything wrong.

```bash
[2021-01-18, 16:20:03] resources are released for /experiment-34-checkpoint-gc  id="default" resource-pool="default" system="master" type="ResourcePool"
<info> [2021-01-18, 16:30:09] decided to terminate 1 instances: ,i-0d2fc018489ab2f24 (reason: long idle)  id="provisioner" resource-pool="default" system="master" type="Provisioner"
<info> [2021-01-18, 16:30:09] terminated 1/1 EC2 instances: i-0d2fc018489ab2f24 (Terminating)  id="provisioner" resource-pool="default" system="master" type="Provisioner"
<error> [2021-01-18, 16:30:10] error while actor was running  error="websocket: close 1006 (abnormal closure): unexpected EOF" id="websocket-393a233c-3a95-4aad-b362-3ac57c244a5c" system="master" type="websocketActor"
<error> [2021-01-18, 16:30:10] websocket: close 1006 (abnormal closure): unexpected EOF
<error> [2021-01-18, 16:30:10] http: connection has been hijacked
<error> [2021-01-18, 16:30:10] error while actor was running  error="child failed: /agents/i-0d2fc018489ab2f24/websocket-393a233c-3a95-4aad-b362-3ac57c244a5c: websocket: close 1006 (abnormal closure): unexpected EOF" id="i-0d2fc018489ab2f24" system="master" type="agent"
<info> [2021-01-18, 16:30:10] removing device: gpu5 (Tesla K80) (i-0d2fc018489ab2f24)  id="default" resource-pool="default" system="master" type="ResourcePool"
<info> [2021-01-18, 16:30:10] removing device: gpu6 (Tesla K80) (i-0d2fc018489ab2f24)  id="default" resource-pool="default" system="master" type="ResourcePool"
<info> [2021-01-18, 16:30:10] removing device: gpu7 (Tesla K80) (i-0d2fc018489ab2f24)  id="default" resource-pool="default" system="master" type="ResourcePool"
[...]

PR #2351, opened 2021-05-07
fix: hide expected network errors when nodes are terminated [DET-4822]

mackrorysd:
## Description

An "abnormal" closure is actually exactly what the websockets library returns in the case where an idle node is being scaled down. Simply including this closure in IsClosureError() prevents this very routine case from spamming multiple lines into the error logging.

## Test Plan

Tested scale-down on a cluster - logs are much cleaner but do indicate the time at which a node is scaled down (so we can differentiate this case from any actual abnormal failures that would happen).


mackrorysd:
This fixes the behavior reported in #1805

PR #2355, opened 2021-05-10
chore: hook ProfilerAgent into harness [DET-5204]

armandmcqueen:
## Description

Hook the ProfilerAgent into the harness.

## Checklist


armandmcqueen:
This PR can be reviewed. 

Please do not comment on the debug logs. They will be completely removed before merging, they are just useful while I manually test. Only distributed training still needs to be tested, so this is ready to review.

harness/determined/keras/callbacks.py
```diff
@@ -701,3 +701,22 @@ def on_train_workload_begin(
...
+        self.current_batch = 0
+        self.prof = prof
+
+    def get_state(self):
+        return {"current_batch": self.current_batch}
+
+    def load_state(self, state: Any) -> None:
+        self.current_batch = state["current_batch"]
+
+    def on_train_batch_begin(self, batch: int, logs: Optional[Dict] = None) -> None:
```
armandmcqueen, line 720: @shiyuann `logs` is part of the abstract class's contract here so mypy complains if I don't include it in the signature.

e2e_tests/tests/fixtures/metric_maker/metric_maker.py
```diff
@@ -68,17 +68,51 @@ class MetricMaker(det.CallbackTrialController):
     based on hyperparameters.
     """
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__init__(*args, **kwargs)
+    def __init__(
+        self,
+        prof: profiler.ProfilerAgent,
```
rb-determined-ai, line 73: I think you have expanded a lot of `*args, **kwargs`, which is a sign that you have broken a pattern that previously worked.  Looking at the ordering of the `prof` argument, it looks inconsistent across various TrialController classes, which is probably the problem.

\<aside>

You are sort of wrestling with an architectural issue in the harness; we've had this problem where we pass way too many parameters through way too many function calls ever since I did the refactor of the harness a year ago.  It's definitely an antipattern, but it's not an antipattern I had an answer for until we decided on push architecture.  I *think* that antipattern is going to be resolved, at least partially, during the harness refactor I'm working on right now for push architecture.

The root of the problem in today's harness is that the part of the system which needs the most context is the lowest layer of the system, the TrialController.  So it ends up needing all of this information passed into it, since all the information is instantiated elsewhere.

A cleaner pattern would be to let the TrialController own more of the training.  After launch layer happens, it will; the horovod launch layer will instantiate the workers, and inside each worker `harness.py` will figure out just enough information to load the user's Trial class, and call TrialController.from_trial().  That will have much fewer args, because a bunch of the information will then be instantiated *inside* the TrialController.  The overall design should be analagous to where the generic API where after the launch layer, the user's workers each call `ctx = det.generic.init()` to instantiate the same collection of information that we currently have.

This will all be much easier with push architecture because all of the logic which currently is embedded in processing a stream of workloads will instead be part of the generic API's calls for reporting metrics and checkpoints.

\</aside>

So I don't think you should reorder the `prof` arg to fix the problem, I think you should just instantiate the profiler in the base `TrialController.__init__()` so all TrialController subclasses can access it as `self.prof`.  Then, (because we separated the `prof.start()` from `prof.__init__()`) we can put the `with prof():` context manager in each trial controller's `run()` call... or actually only  in the three you care about, which are PyTorchTrialController, TFKerasTrialController, and EstimatorTrialController.  You'll have to tweak the logic very slightly, so that e.g. the `local_rank=0` worker collects system metrics for each compute node.

For reference, a similar effort is underway to move the storage manager into the TrialController.  With the storage manager there, the WorkloadManager can go away entirely, and we will have an easy transition to using the new push architecture endpoints.

If you are curious, I'd be happy to explain more about the upcoming architecture changes via zoom.

harness/determined/_experiment_config.py
```diff
@@ -26,6 +26,22 @@ def slots_per_trial(self) -> int:
     def experiment_seed(self) -> int:
         return int(self.get("reproducibility", {}).get("experiment_seed", 0))
 
+    def profiling_enabled(self) -> bool:
+        return bool(self.get("profiling", {}).get("enabled", False))
+
+    def profiling_interval(self) -> Tuple[int, Optional[int]]:
+        end_after_batch = self.get("profiling", {}).get("end_after_batch", None)
```
rb-determined-ai, line 33: I don't understand why you are keying off of `end_after_batch` rather than `enabled`?  And isn't `end_after_batch = 0` a totally valid value for `end_after_batch`?

harness/determined/keras/callbacks.py
rb-determined-ai: I think what shiyuan was probably getting at was that you should call it `_: Optional[Dict] = None` to make it clear it's not being used.

harness/determined/_experiment_config.py
stoksc: fixed

e2e_tests/tests/fixtures/metric_maker/metric_maker.py
stoksc: (at the design, still getting it working with tests and stuff)

harness/determined/keras/callbacks.py
stoksc: also fixed

harness/determined/_trial_controller.py
```diff
@@ -41,13 +42,21 @@ def __init__(
...
         self.workloads = workloads
         self.load_path = load_path
         self.rendezvous_info = rendezvous_info
         self.hvd_config = hvd_config
+        self.prof = prof
+
+        self.prof = profiler.ProfilerAgent.from_env(
+            env,
+            rendezvous_info.get_rank(),
+            int(os.getenv("HOROVOD_LOCAL_RANK", 0)),
```
stoksc, line 58: afaict this is the same env where we were getting it before so it just be there/just fine

harness/determined/estimator/_estimator_trial.py
```diff
@@ -694,21 +702,24 @@ def _init_run_config(self, config: tf.estimator.RunConfig) -> tf.estimator.RunCo
...
-                "Training loop exited unexpectedly but without throwing any errors. This is "
-                "possibly due to either setting train_spec.max_steps to a non-None value or due to "
-                "a user callback causing the training loop to exit, which is not supported at this "
-                "time."
-            )
-        finally:
-            for callback in self.train_hooks:
-                if isinstance(callback, estimator.RunHook):
-                    callback.on_trial_close()
+        # would've done a decator for all these runs instead but then i need to pop
```
stoksc, line 705: don't forget to rm this comment

armandmcqueen:
IMO we should remove the args and kwargs. Every single one of those APIs is a strictly-defined, typed API that will fail if the arguments are not passed in in the correct order. There's no reason to use args or kwargs - it just turns a well documented and clear API into an opaque one.

stoksc:
> IMO we should remove the args and kwargs. All of those APIs is a strictly-defined, typed API that will fail if the arguments are not passed in in the correct order. There's no reason to use args or kwargs - it just turns a well documented and clear API into an opaque one.

I agree with that, I've always disliked them. I think more and more they will go away with this upcoming refactor.

harness/determined/keras/callbacks.py
```diff
@@ -701,3 +701,23 @@ def on_train_workload_begin(
...
+
+    def get_state(self) -> Dict:
+        return {"current_batch": self.current_batch}
+
+    def load_state(self, state: Any) -> None:
+        self.current_batch = state["current_batch"]
+
+    def on_train_batch_begin(self, batch: int, _: Optional[Dict] = None) -> None:
+        # Brad: why not just this?
+        # self.prof.update_batch_idx(batch)
```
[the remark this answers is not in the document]
rb-determined-ai: the `_` arg is an epoch index; `batch` is not absolute.

harness/determined/_trial_controller.py
rb-determined-ai: please do not read `HOROVOD_LOCAL_RANK`, use `context.distributed.get_rank()` instead.

harness/determined/_swagger/client/api_client.py
```diff
@@ -14,18 +14,18 @@
 import datetime
 import json
 import mimetypes
-from multiprocessing.pool import ThreadPool
```
rb-determined-ai, line 17: Please discard changes from `harness/determined/_swagger/` and rebase onto tip-of-master; there was a bug in `make fmt` that caused this

rb-determined-ai (commented):
Thank you, this approach is going to make the push architecture work much easier.

rb-determined-ai:
>  > IMO we should remove the args and kwargs. All of those APIs is a strictly-defined, typed API that will fail if the arguments are not passed in in the correct order. There's no reason to use args or kwargs - it just turns a well documented and clear API into an opaque one.
>
> I agree with that, I've always disliked them. I think more and more they will go away with this upcoming refactor.

Honestly, extending an existing function using `*args, **kwargs` is very idomatic python.  It does require the developer to understand the class hierarchy a bit more, but it avoids issues where e.g. all of the subclasses of a function take the same parameters but in different order.  Personally I think that's worth it, but I can see the opacity counter-argument.  TBH I would probably not be happy with a pattern like this in a user-facing API, so I agree it's not very good.

Overall though, I think we probably all agree that there are a lot of objects that are initialized with a long and pretty similar list of arguments, and that will hopefully get cleaned up in the current push architecture harness refactor.

stoksc:
going to close in favor of #2348 since it is confusing to reviewers since they alter the same code and the code is _much_ more compact now.

PR #2356, opened 2021-05-10
test: always calling read() before calling wait()

shiyuann:
## Description

The tests seem to be fine without this change. But technically you should always call `read` before calling `wait` on a process otherwise it would block forever. `wait()` might be a blocking call if your CLI prints out an empty line, which happens on my MacOS.

## Checklist


stoksc (approved):
lgtm, `pexpect.spawn.wait()` docs clearly says it will block if there is unread output. confirmed on my mac.

PR #2357, opened 2021-05-10
fix: limit metrics size to 100mb

azhou-determined:
enforce max metric size object to 500MB


harness/determined/common/constants.py
```diff
@@ -17,6 +17,11 @@
 # We subtract one megabyte to account for any message envelope size we may have.
 MAX_CONTEXT_SIZE = MAX_ENCODED_SIZE - (1 * 1024 * 1024)
 
+# The maximum size of a workload metrics object is capped at 500MB. Metrics
+# are specified by the model definition and persisted after each workload
+# training step is completed.
+MAX_METRICS_SIZE = 500 * (1 << 20)
```
eecsliu, line 23: Lets change this to 100 MB, unless Hoang told you otherwise?

harness/determined/util.py
```diff
@@ -144,6 +143,14 @@ def wrap_metrics(
         return {"metrics": metrics, "stop_requested": stop_requested, "invalid_hp": invalid_hp}
 
 
+def sizeof_dict(dict: Dict[str, Any]) -> int:
+    """
+    Serializes dict object to json and calculates size in bytes.
+    """
+    dict_json = json_encode(dict)
```
eecsliu, line 150: were you able to see how long this operation took? Any noticeable performance impact to experiments?

eecsliu (changes requested):
Looks mostly good - just checking on the max file size and the runtime.

harness/determined/util.py
azhou-determined: weren't any noticeable impacts when i ran the experiments. timed the harness in python, the entire sizeof_dict method takes around 7ms.

harness/determined/common/constants.py
azhou-determined: my bad, forgot to change it. done.

harness/determined/layers/_workload_manager.py
```diff
@@ -141,6 +140,12 @@ def _respond(in_response: workload.Response) -> None:
 
             in_response = cast(workload.Metrics, in_response)
             metrics = in_response["metrics"]
+
+            metrics_size = det.util.sizeof_dict(metrics)
```
shiyuann, line 144: blocking: I tried to time json-encoding 100MB object, which takes 27ms. It doesn't hurt much but I'd check the size in `_socket_manager.py`, where calls `json_encode`. Then you don't need to encode the metrics twice unnecessarily.

azhou-determined: the response in _socket_manager.py includes other data that isn't directly what's recorded on a column-basis to the database. it includes some overhead for other columns: 
`{'type': 'WORKLOAD_COMPLETED', 'workload': <RUN_STEP (100 Batches): (1,1,10)>, 'start_time': datetime.datetime(2021, 5, 17, 19, 35, 34, 97884, tzinfo=datetime.timezone.utc), 'end_time': datetime.datetime(2021, 5, 17, 19, 35, 37, 558983, tzinfo=datetime.timezone.utc), 'metrics': {...metrics }`
not a huge deal, but not 100% accurate in terms of the size of the object being persisted to the column. i guess i could encode the actual metrics piece of that json separately, then manually stitch it together to send in _socket_manager respond, but that seems confusing from a code perspective. what do you think?

PR #2359, opened 2021-05-11
feat: add a preview parameter to the Notebook launching API [DET-5368]

shiyuann:
## Description

Add a preview parameter to the Notebook launching API. This API returns a full configuration populated with the template and the Notebook configuration.

## Test Plan

- Manually test it.

## Checklist


master/internal/api_command.go
```diff
@@ -31,36 +34,122 @@ type protoCommandParams struct {
...
+			return nil, nil, errors.Wrapf(
+				err,
+				"unable to parse the config in the parameters: %s",
+				string(configBytes),
+			)
+		}
+	}
+
+	// mustBeZeroSlot indicates that this type of command may never use more than
+	// zero slots (as of Jan 2021, this is only Tensorboards). This is important
```
naren-determined, line 71: Instead of putting a date here, can we put a Determined version?

naren-determined (approved):
Made one comment that I think should be changed, but approved b/c that is minor and everything else looks good!

master/internal/api_command.go
shiyuann: This is a comment that is moved from other places in the codebase. I can't really find the exact version to which it belongs.  Anyways, this should be refactored later. I'd not worry about it

master/internal/api_tensorboard.go
```diff
@@ -69,24 +70,22 @@ func (a *apiServer) LaunchTensorboard(
...
 
 	tensorboardConfig := command.TensorboardRequest{
-		CommandParams: cmdParams,
+		CommandParams: params,
 		ExperimentIDs: experimentIds,
 		TrialIDs:      trialIds,
 	}
-	tensorboardLaunchReq := command.TensorboardRequestWithUser{
-		Tensorboard: tensorboardConfig,
-		User:        user,
```
hamidzr, line 88: question: is the user portion of this type used anywhere else? it seems we can get rid of it.

hamidzr (approved):

EmilyBonar (approved):

master/internal/api_tensorboard.go
shiyuann: It is not gotten rid of. It is merged into the `Tensorboard` field. I just reused the same go struct.

hamidzr: aah I didn't explain it well I meant something like this https://github.com/determined-ai/determined/pull/2377 how does this look?

shiyuann: It looks good! I approved it

PR #2360, opened 2021-05-11
fix: profiles timings graph data conversion filling empty data [DET-5433]

apizzini:
## Description

WebUI was expecting all data points (where with data point I mean the values included in a batch or a timestamp returned by the API) to specify a value for all the metrics.

That's not happening (eg. batch 1 might have 3 metric values, batch 2 might have 1 metric value).

Added a fix to default to `0` when a value is not specified.

Note that we could use `null` too (difference between `0` and `null` is that uPlot will interpret `null` as a missing value and, depending on the chart config, display a gap between points).

## Checklist


leeoniya:
there is an alignment utility in uPlot for this called `uPlot.join()`. you can see it used in this demo: https://github.com/leeoniya/uPlot/blob/master/demos/path-gap-clip.html

hkang1 (approved):
The code looks good, thanks for the quick fix!

PR #2361, opened 2021-05-11
chore: fixes eslint error

emily-roses:
## Description
While setting up eslint, I received the following error:
```
ESLint: Error while loading rule 'jest/no-deprecated-functions': Unable to detect Jest version - please ensure jest package is installed, or otherwise set version explicitly
```
Making this change to eslintrc.js fixes this error and allows eslint to function correctly.


hamidzr (approved):
LGTM. I can't repro the issue but this seems harmless.

PR #2364, opened 2021-05-12
chore: move trial hyperparameters in a dedicated trial page tab [DET-5412]

apizzini:
## Description

This is a simple change needed to be able to remove the overview info box in the next MRs related to the trial detail page.

In the next weeks we will work on a better visualization for those.

BEFORE:

Schermata 2021-05-12 alle 10 56 39
Schermata 2021-05-12 alle 10 56 33

AFTER:

Schermata 2021-05-12 alle 10 56 06
Schermata 2021-05-12 alle 10 55 59

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
+++ b/webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
@@ -1,27 +1,53 @@
-import { List } from 'antd';
-import React from 'react';
+import { ColumnType } from 'antd/es/table';
+import React, { useMemo } from 'react';
 
-import Section from 'components/Section';
+import ResponsiveTable from 'components/ResponsiveTable';
 import { ExperimentBase, TrialDetails } from 'types';
 import { isObject } from 'utils/data';
+import { alphanumericSorter } from 'utils/sort';
 
 export interface Props {
   experiment: ExperimentBase;
   trial: TrialDetails;
 }
 
+interface HyperParameter {
+  hyperparameter: string,
+  value: string,
+}
+
 const TrialDetailsHyperparameters: React.FC<Props> = ({ trial }: Props) => {
+  const columns: ColumnType<HyperParameter>[] = useMemo(() => [
+    {
+      dataIndex: 'hyperparameter',
+      defaultSortOrder: 'ascend',
+      key: 'hyperparameter',
+      sorter: (a: HyperParameter, b: HyperParameter) =>
+        alphanumericSorter(a.hyperparameter, b.hyperparameter),
+      title: 'Hyperparameter',
+    },
+    {
+      dataIndex: 'value',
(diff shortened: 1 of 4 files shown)


webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export interface Props {
+  experiment: ExperimentBase;
+  trial: TrialDetails;
+}
+
+const TrialDetailsHyperparameters: React.FC<Props> = ({ trial }: Props) => {
+  return (
+    <Section bodyBorder bodyNoPadding>
+      <List
```
apizzini, line 16: @hkang1 we could consider displaying those in a Table instead of a List. Do you think that's better?

In case:
- I guess the table should not be sortable?
- What headers could we use for the 2 colums? "Hyperparameter" and "Value"?

hkang1: @apizzini I think that's a great idea. The 2 columns you proposed sounds good, and then make just `Hyperparameter` column sortable (as alphabetical)

hkang1 (approved):
LGTM! The table suggestion is a welcome addition.

webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
```diff
@@ -0,0 +1,54 @@
...
+      sorter: (a: HyperParameter, b: HyperParameter) =>
+        alphanumericSorter(a.hyperparameter, b.hyperparameter),
+      title: 'Hyperparameter',
+    },
+    {
+      dataIndex: 'value',
+      key: 'value',
+      title: 'Value',
+    },
+  ], []);
```
hkang1, line 34: **super nit:** this is a total personal preference thing so feel free to ignore. A space after this line even though it's variable declarations, just because it's a series of large `useMemo` blocks. So using a space to break these blocks up a bit. Similar to how you was organize `useCallbacks`.

webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
```diff
@@ -0,0 +1,54 @@
...
+        value: isObject(value) ? JSON.stringify(value, null, 2) : value.toString(),
+      };
+    });
+  }, [ trial.hparams ]);
+
+  return (
+    <ResponsiveTable
+      columns={columns}
+      dataSource={dataSource}
+      pagination={false}
```
hkang1, line 49: **style suggestion:** add a `size="small"` so it's more consistent with our other tables.

hkang1 (approved):
LGTM. Thanks for updating to a table! Left two small comments.

PR #2365, opened 2021-05-12
chore: move trial logs in a trial detail page tab [DET-5410]

apizzini:
## Description

BEFORE:

Schermata 2021-05-12 alle 11 28 18
Schermata 2021-05-12 alle 11 28 08

AFTER:

Schermata 2021-05-12 alle 11 27 46

## Checklist


webui/react/src/routes/routes.ts
```diff
@@ -1,18 +1,6 @@
...
-  {
-    id: 'trialLogs',
-    needAuth: true,
-    path: '/experiments/:experimentId/trials/:trialId/logs',
-    title: 'Trial Logs',
-  },
-  {
-    id: 'trialLogs',
-    needAuth: true,
-    path: '/trials/:trialId/logs',
```
apizzini, line 13: @hkang1 do you think we can drop the support for this old route?

Otherwise we need to create a component just to handle the redirect for this (need to call API to get the trial to have the experimentId and then redirect).

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -445,53 +442,53 @@ const LogViewerTimestamp: React.FC<Props> = ({
...
+      bodyBorder
+      bodyDynamic
+      bodyNoPadding
+      filters={FilterComponent && <FilterComponent
+        filter={filter}
+        filterOptions={filterOptions}
+        onChange={setFilter}
+      />}
+      maxHeight
+      options={logOptions}
```
apizzini, line 455: I don't really like where options appear in the page, but don't have a better idea at the moment.

The only alternative that comes to my mind is to move them inside the "filters" area, maybe separated from the "filters" by a border.

Schermata 2021-05-12 alle 11 31 18

hkang1: I think putting them inside the filter area might make that already busy area a bit harder to process, especially when they are actions that don't deal with filtering.

I don't have any great ideas for these buttons either, but one thing that can help with just general positioning is just to right alight these buttons for now.

webui/react/src/routes/routes.ts
hkang1: @apizzini should changing line ~25 from:
```
path: '/trials/:trialId',
```
to
```
path: '/trials/:trialId/:tab?',
```
I just tried it and links like `http://localhost:3000/det/trials/2031/logs` worked great.

hkang1 (approved):
This is shaping up, nice work! Couple of comments in response to your questions.

webui/react/src/components/LogViewerTimestamp.tsx
```diff
@@ -223,15 +220,15 @@ const LogViewerTimestamp: React.FC<Props> = ({
...
+        message: 'Available logs Copied',
       });
     } catch (e) {
       notification.warn({
         description: e.message,
         message: 'Unable to Copy to Clipboard',
       });
     }
-  }, [ logs, props.pageProps.title ]);
+  }, [ logs ]);
```
hkang1, line 231: 👍

hkang1 (approved):
Sweet, changes work great!

webui/react/src/components/LogViewer.module.scss
```diff
@@ -17,6 +17,7 @@
     font-family: 'Source Code Pro', monospace;
     font-size: var(--theme-sizes-font-small);
     height: 100%;
+    min-height: 500px;
```
hkang1, line 20: go ahead and land this, the fix for this is a bit more complicated (need to apply flex box all the way through to the logviewer from section). We'll address that separately.

hkang1 (approved):
go ahead and land this, the fix for this is a bit more complicated (need to apply flex box all the way through to the logviewer from section). We'll address that separately.

PR #2368, opened 2021-05-12
chore: disable eslint for storybook

EmilyBonar:
## Description
Our configuration of eslint and storybook's disagree on formatting. This disables storybook's version in order to resolve that disagreement.


EmilyBonar:
My problem was real, but this was not the right solution

PR #2373, opened 2021-05-13
chore: add zmq-based IPC to the DistributedContext [DET-5484]

rb-determined-ai:
This is part of a multi-step process to refactor the harness code in
favor of push architecture:

Goal:

*    Introduce Python WorkloadSequencer that uses Push Architecture.

That can't happen until:

*    The WorkloadManager is broken into simple function calls for
    mutating workload responses.  These simple functions (one after each
    of training, validation, and checkpointing) can be called right
    before making calls to the push API, rather than after responding
    with a workload.Response on a stream of workloads.

That can't happen until:

*    The StorageManager is no longer a part of the WorkloadManager.  The
    StorageManager is the only part of the WorkloadManager that doesn't
    fit neatly into the aforementioned three function calls, since
    storage_mgr.store_path() is a contextmanager.  However, if the
    TrialController were responsible for the StorageManager rather than
    the WorkloadManager, the extra `args: List[Any]` parameter in the
    workload stream would disappear.

Now, it would be nice to package all of that functionality on a
GenericContext of some sort, in preparation for the eventual Generic
API.  It makes sense that the GenericContext would be the place that all
metrics and checkpoint reporting would go through, so that would be the
place where after_training(), after_validation(), and
after_checkpointing() would each be called.  It also makes sense that
the GenericContext would expose the StorageManager to user code.

But nothing can be packaged into the GenericContext until the
GenericContext has ZMQ communications.  For example, a
GenericContext.maybe_load_checkpoint() can't function without
cross-worker synchronization; only the local_rank=0 worker each compute
node should actually load the checkpoint, and the other workers must
wait.

We have all of the tools for ZMQ communication in the harness, and we
even have the ports allocated for it.  But they have always been
attached to the TrialController, rather than the DistributedContext.
That is because right after the refactor, I believed/enforced that the
Context objects were purely for light-weight API call exposures and
didn't have any beefy logic in them.  However, PyTorchContext has proven
that putting logic in the Context object is actually a great strategy,
and the upcoming Generic API is really just the logical conclusion of
that realization.

So as a first step, the ownership of the ZMQ sockets is passing to the
DistributedContext (which will eventually be a member of the
GenericContext).

### DIFF
--- a/harness/determined/_train_context.py
+++ b/harness/determined/_train_context.py
@@ -1,9 +1,9 @@
 import abc
 import logging
-from typing import Any, Callable, Dict, Optional, cast
+from typing import Any, Callable, Dict, List, Optional, cast
 
 import determined as det
-from determined import horovod
+from determined import constants, horovod, ipc
 
 
 class _TrainContext(metaclass=abc.ABCMeta):
@@ -12,10 +12,16 @@ class _TrainContext(metaclass=abc.ABCMeta):
     These methods should be made available to both Native and Trial APIs.
     """
 
-    def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext):
-        self.env = env  # type: det.EnvContext
-        self.hvd_config = hvd_config  # type: horovod.HorovodContext
-        self.distributed = DistributedContext(env, hvd_config)
+    def __init__(
+        self,
+        env: det.EnvContext,
+        hvd_config: horovod.HorovodContext,
+        rendezvous_info: det.RendezvousInfo,
+    ) -> None:
+        self.env = env
+        self.hvd_config = hvd_config
+        self.rendezvous_info = rendezvous_info
+        self.distributed = DistributedContext(env, hvd_config, rendezvous_info)
         self._stop_requested = False
 
     @classmethod
@@ -58,7 +64,7 @@ class _TrainContext(metaclass=abc.ABCMeta):
             config=config,
             limit_gpus=1,
         )
-        return cls(env_context, hvd_config)
+        return cls(env_context, hvd_config, rendezvous_info)
 
     def get_experiment_config(self) -> Dict[str, Any]:
         """
@@ -152,8 +158,13 @@ class TrialContext(_TrainContext):
     The context passed to the User's ``Trial.__init__()`` will inherit from this class.
     """
 
-    def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext):
-        super().__init__(env, hvd_config)
+    def __init__(
+        self,
+        env: det.EnvContext,
+        hvd_config: horovod.HorovodContext,
+        rendezvous_info: det.RendezvousInfo,
+    ) -> None:
+        super().__init__(env, hvd_config, rendezvous_info)
 
 
 class NativeContext(_TrainContext):
@@ -163,8 +174,13 @@ class NativeContext(_TrainContext):
     The context returned by the ``init()`` function will inherit from this class.
     """
 
-    def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext):
-        super().__init__(env, hvd_config)
+    def __init__(
+        self,
+        env: det.EnvContext,
+        hvd_config: horovod.HorovodContext,
+        rendezvous_info: det.RendezvousInfo,
+    ) -> None:
+        super().__init__(env, hvd_config, rendezvous_info)
         self._train_fn = None  # type: Optional[Callable[[], None]]
 
     def _set_train_fn(self, train_fn: Callable[[], None]) -> None:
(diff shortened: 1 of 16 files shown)


rb-determined-ai:
I should also mention that this PR is a step in the direction of finally deleting the PolyHorovod object, since the places where framework-agnostic horovod is required are getting fewer and fewer.  By the time we have a fully Generic API, we will not need it at all (since users will not need it).

harness/determined/_trial_controller.py
```diff
@@ -249,82 +249,5 @@ def __init__(self, *args: List[Any], **kwargs: Dict[str, Any]) -> None:
...
+            f"TrialController initialized on rank {rank}, using hvd: {self.hvd_config.use}."
         )
-
-        # Initialize communication directly between training processes.
-        self.train_process_comm_chief = None  # type: Optional[ipc.ZMQBroadcastServer]
-        self.train_process_comm_worker = None  # type: Optional[ipc.ZMQBroadcastClient]
-        if self.hvd_config.use:
-            self._initialize_train_process_comm()
-
-    def _initialize_train_process_comm(self) -> None:
```
shiyuann, line 262: :D I actually wrote some similar changes in my `LightningTrial` code. I am so glad to see those changes happen here. Those functions just make no sense to be in the trial controller.

harness/determined/cli/experiment.py
```diff
@@ -174,7 +174,7 @@ def local_experiment(args: Namespace) -> None:
     determined.common.set_logger(bool(experiment_config.get("debug", False)))
 
     with det._local_execution_manager(args.model_def.resolve()):
-        trial_class = load.load_trial_implementation(experiment_config["entrypoint"])
+        trial_class = load.trial_class_from_entrypoint(experiment_config["entrypoint"])
```
shiyuann, line 177: prasie: I like the new name. :D

harness/determined/estimator/_estimator_context.py
```diff
@@ -34,8 +34,13 @@ class EstimatorContext(estimator._EstimatorReducerContext):
     workflow that uses the ``tf.estimator`` API.
     """
 
-    def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext) -> None:
-        super().__init__()
+    def __init__(
+        self,
+        env: det.EnvContext,
+        hvd_config: horovod.HorovodContext,
+        allgather_fn: Callable[[Any], List[Any]],
```
shiyuann, line 41: :D This makes the inheritance relationship between `EstimatorContext` and `EstimatorTrialContext` looks dumb since `TrialContext` inherits `TrainContext`, which contains the allgather function. I understand this is just workaround to make the od code path work. So I won't let you refactor the whole class inheritance relationships.

harness/determined/estimator/_estimator_trial.py
```diff
@@ -770,34 +768,17 @@ def compute_validation_metrics(self) -> workload.Response:
...
-        else:
-            self.train_process_comm_worker = cast(
-                ipc.ZMQBroadcastClient, self.train_process_comm_worker
-            )
-            logging.debug(f"Worker {hvd.rank()} sending metrics.")
-            self.train_process_comm_worker.send(metrics)
-            # Synchronize with the chief so that there is no risk of accidentally calling send()
-            # for a future gather before all workers have called send() on this gather.
-            _ = self.train_process_comm_worker.recv()
+        all_metrics = self.context.distributed._zmq_gather(metrics)
```
shiyuann, line 771: praise: nice refactor

harness/determined/_train_context.py
```diff
@@ -152,8 +158,13 @@ class TrialContext(_TrainContext):
...
+    def __init__(
+        self,
+        env: det.EnvContext,
+        hvd_config: horovod.HorovodContext,
+        rendezvous_info: det.RendezvousInfo,
+    ) -> None:
+        super().__init__(env, hvd_config, rendezvous_info)
 
 
 class NativeContext(_TrainContext):
```
shiyuann, line 170: Shouldn't this be removed? I thought we have already removed the native API.

shiyuann (approved):
This is actually a PR that makes me happier after reviewing. Just one question: what blocks us removing the Native API?

harness/determined/_train_context.py
rb-determined-ai: No, it is not removed.  It is only deprecated.  I have stated that I would like to remove `det.experimental.keras.init()` and `det.experimental.estimator.init()` for the `0.16.0` release, but that won't happen until right before the `0.16.0` release.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -931,7 +931,7 @@ def _compute_validation_metrics(self) -> workload.Response:
             # Use a global ZMQ barrier here because we have observed cases where hvd.allreduce
             # may hang when called minutes apart by different workers which may happen if
             # workers complete evaluation at different speeds.
-            self._global_barrier()
+            _ = self.context.distributed._zmq_gather(None)
```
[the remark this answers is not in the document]
rb-determined-ai: Why bother, while they are all private functions?  (we can discuss the public API later, when we expose a Generic API)

harness/determined/_trial_controller.py
rb-determined-ai: Agreed.  (sorry it took me so long to come around, haha)

harness/determined/estimator/_estimator_context.py
rb-determined-ai: Yeah.  I didn't spend too much time refactoring the inheritance pattern here, largely because of the Native Context laying around.  It's only the Native context that even distinguishes between the `EstimatorContext` and the `EstimatorTrialContext`, so after Native is gone there will only be `EstimatorContext`.

rb-determined-ai:
all tests passed, merging.

PR #2376, opened 2021-05-13
fix: `det deploy gcp` support for terraform 0.15 [DET-5449]

ioga:
## Description
Terraform 0.15 [doesn't support](https://www.terraform.io/docs/cli/commands/plan.html#passing-a-different-configuration-directory) passing the config directory as the last argument anymore; they suggest using -chdir instead, but that doesn't work in older terraform (e.g. 0.13). Since we explicitly specify the tfstate path anyway, change the subprocess `cwd` instead.

Also, remove `shell=True`. Now the users will be able to have directory names with spaces in them.

## Test Plan
Tested with Terraform 0.13 and 0.15.

## Checklist

### DIFF
--- a/docs/release-notes/2376-fix-gcp-terraform.txt
+++ b/docs/release-notes/2376-fix-gcp-terraform.txt
@@ -1,0 +1,5 @@
+:orphan:
+
+**Bug Fixes**
+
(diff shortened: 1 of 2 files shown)


dzhu:
Good fixes, thanks!

dzhu (approved):

PR #2378, opened 2021-05-14
refactor: experiment list native filters [DET-5389]

hkang1:
## Description

- Move filtering capability into native Ant Design filter column. (Labels, State, Archived, User)
- Separate labels into separate column for enabling filtering.
- Enable show more for labels columns to help control # of labels to show from the start. (show +x more option if more than 1 label, can change later easily)
- With the exception of Archived, support multiple option filter (with OR logic).
- Validate filters before making API calls to ensure valid values are being sent
- Visually show which filter columns have filters applied

Before:
Screen Shot 2021-05-13 at 5 29 07 PM

After:
2021-05-14 at 12 29 PM

Not in scope:
- move experiment list page search into ID and Name column

## Test Plan

- [ ] able to add and delete experiment labels from experiment list page
- [ ] check that the dashboard page still displays and filters correctly
- [ ] check filtering mechanism works properly for `Labels`, `States`, `Archived` and `Users` columns
- [ ] check that the column filtering correctly updates the experiment list page URL with the correct params
- [ ] check experiment list page still reads the url params correctly for filter values
- [ ] check that show `+[X] more` option displays properly for when there are more than 1 label and expands properly. Also ensure adding and editing labels still work.
- [ ] check experiment list page column sorting still functions

## Checklist


webui/react/src/pages/ExperimentList.tsx
```diff
@@ -353,53 +410,68 @@ const ExperimentList: React.FC = () => {
 
     const newColumns = [ ...defaultColumns ].map(column => {
       column.sortOrder = null;
+
       if (column.key === sorter.key) column.sortOrder = sorter.descend ? 'descend' : 'ascend';
-      if (column.key === V1GetExperimentsRequestSortBy.DESCRIPTION) column.render = nameRenderer;
+      if (column.key === 'labels') {
```
apizzini, line 415: have you considered moving all the definitions of the columns here instead of importing from an external file and mapping on it to apply tweaks?

apizzini (commented):
This looks great, the only thing I don't really like is that if filters are "reset" or "cleared" manually (deselecting all the items i activated previously), the Icon doesn't disable, it stays active. Can we do something about this?

apizzini (approved):

webui/react/src/pages/ExperimentList.tsx
hkang1: lol, yeah at this point, there's so much modifications, the external import doesn't make sense.

hkang1:
> This looks great, the only thing I don't really like is that if filters are "reset" or "cleared" manually (deselecting all the items i activated previously), the Icon doesn't disable, it stays active. Can we do something about this?

thanks for finding and reporting that issue, should be good to go now.

hkang1:
@cla-bot check

PR #2382, opened 2021-05-14
chore: unrevert and fix "actually use expconf in the master"

rb-determined-ai:
## Description

fix add schemas.Defaultable and schemas.Copyable

The reflect-based behavior of schemas.WithDefaults and schemas.Merge
works great on plain-old-data types which have no private fields, but it
pukes on private fields.

There already was a schemas.Mergable interface for customizing the
merging behaviors of objects.  This change introduces the
schemas.Defaultable and schemas.Copyable interfaces, which have
analagous effects on schemas.WithDefaults and schemas.cpy (unexported).

schemas.Copyable is very easy to understand.  For example, the new
expconf.Pod (which is type-defined to k8sV1.Pod) simply calls
k8sV1.Pod.DeepCopy().

schemas.RuntimeDefaultable has been removed in favor of
schemas.Defaultable.  Defaultable makes the reflect code a little
easier and is more like the Mergable and Copyable
interfaces than RuntimeDefaultable was.  The flipside is that the
runtime default behavior of the ExperimentConfig.Description is slightly
more complex.  Overall the tradeoff seems appropriate; the general code
(schemas.WithDefaults()) got more general and clearer, and the specific
challenges of the calling code are handled in the calling code.

Overall, the motivation for this change was to protect against panics
when interfacing with the k8sV1.Pod object.  The only other external
object in the expconf project, `AuthConfig`, is reflect-friendly and
does not require modification.

## Test Plan

I will run GKE tests before thinking this works.

### DIFF
--- a/master/pkg/schemas/copy.go
+++ b/master/pkg/schemas/copy.go
@@ -5,11 +5,29 @@ import (
 	"reflect"
 )
 
+// Copyable means an object can have custom behvaiors for schemas.Copy.  Mostly this is useful for
+// working around types which we do not own and which schemas.Copy() would puke on.
+type Copyable interface {
+	// Copy should return the same type.  It must not be defined as a method against a pointer of
+	// the type or it will not work.
+	Copy() interface{}
+}
+
 // cpy is for deep copying, but it will only work on "nice" objects, which should include our
 // schema objects.  Useful to other reflect code.
 func cpy(v reflect.Value) reflect.Value {
 	// fmt.Printf("cpy(%T)\n", v.Interface())
 	var out reflect.Value
+
+	// Detect Copyables, but disallow pointers.  The reason is that if you have a method like:
+	//    func (t Thing) Copy() interface{}
+	// then Copy() will return a plain Thing object, but a pointer to a Thing will still be treated
+	// as copyable.  Then, schemas.Copy(&t).(*Thing) would panic because it returns the wrong type.
+	if v.Kind() != reflect.Ptr {
+		if copyable, ok := v.Interface().(Copyable); ok {
+			return reflect.ValueOf(copyable.Copy())
+		}
+	}
 
 	switch v.Kind() {
 	case reflect.Ptr:
--- a/master/pkg/schemas/merge.go
+++ b/master/pkg/schemas/merge.go
@@ -7,7 +7,8 @@ import (
 
 // Mergable means an object can have custom behvaiors for schemas.Merge.
 type Mergable interface {
-	// Merge should take a struct and return the same struct.
+	// Merge should take a struct and return the same struct.  It must not be defined as a method
+	// against a pointer of the type or it will not work.
 	Merge(interface{}) interface{}
 }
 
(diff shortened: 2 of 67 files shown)


rb-determined-ai:
maybe this is obvious, but please only review the fix commit.  The other is just a revert of a revert.

master/pkg/schemas/copy.go
```diff
@@ -5,12 +5,30 @@ import (
...
 // cpy is for deep copying, but it will only work on "nice" objects, which should include our
 // schema objects.  Useful to other reflect code.
 func cpy(v reflect.Value) reflect.Value {
 	// fmt.Printf("cpy(%T)\n", v.Interface())
 	var out reflect.Value
 
+	// Detect Copyables, but disallow pointers.  The reason is that if you have a method like:
+	//    func (t Thing) Copy() interface{}
+	// then Copy() will return a plain Thing object, but a pointer to a Thing will still be treated
+	// as copyable.  Then, schemas.Copy(&t).(*Thing) would panic because it returns the wrong type.
```
stoksc, line 25: Seems to me a method `func (t Thing) Copy() interface{}` should always operate on non-pointer receivers not the opposite?

master/pkg/schemas/merge.go
```diff
@@ -84,9 +85,15 @@ func merge(obj reflect.Value, src reflect.Value, name string) reflect.Value {
...
 
-	// Detect Mergables.
-	if mergeable, ok := obj.Interface().(Mergable); ok {
-		return reflect.ValueOf(mergeable.Merge(src.Interface()))
+	// Detect Mergables, but disallow pointers.  The reason is that if you have a method like:
+	//    func (t Thing) Merge(other interface{}) interface{}
+	// then Merge() will return a plain Thing object, but a pointer to a Thing will still be treated
+	// as mergable.  Then, schemas.Merge(&t, &t2).(*Thing) would panic because it returns the wrong
+	// type.
+	if obj.Kind() != reflect.Ptr {
```
stoksc, line 93: q: if we have a mergable and a non-pointer here, should we panic instead?

stoksc (approved):
lgtm just some q's

master/pkg/schemas/merge.go
stoksc: `obj.Kind() != reflect.Ptr` == brad cant read

master/pkg/schemas/copy.go
stoksc: `obj.Kind() != reflect.Ptr` == brad cant read

master/pkg/schemas/merge.go
stoksc: obj != src == brad cant read

master/pkg/schemas/copy.go
rb-determined-ai: Is that not what I'm saying?  Suppose you have a
```go
func (t Thing) Copy() interface{}
```
then you have to be careful as you are recursing through something like:
```go
type SuperThing struct {
    RawThing *Thing
}
```
then you'll find that:
```
_, ok =: superThing.RawThing.(Copyable)
```
Will return `ok`.  But that's a problem!  Because you will have accidentally converted a pointer to a non-pointer as part of `schemas.Copy()`.

master/pkg/schemas/merge.go
rb-determined-ai: We have other checks to make sure the types are exactly right.  Also, we don't modify either of `src` or `obj` when we merge, we return a clean copy.

rb-determined-ai:
passed GKE tests.  Talked with Bradley, and he's on board with the comments.  Landing now.

PR #2383, opened 2021-05-17
feat: add SlotsPerAgent in resource pool API

shiyuann:
## Description

Add a SlotsPerAgent field in the resource pool API. This returns the slot number on each dynamic agent. If there is no dynamic agents configured, return -1.

## Test Plan

- CI
- Manually check the resource pool field.

## Checklist

### DIFF
--- a/docs/release-notes/2383-slots-per-agent.txt
+++ b/docs/release-notes/2383-slots-per-agent.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**Improvements**
+
+-  API: return a SlotsPerAgent field in the resource pool API. This
+   returns the slot number on each dynamic agent on AWS, GCP, or K8s. If
+   no dynamic agent is configured, return -1.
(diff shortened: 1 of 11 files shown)


EmilyBonar (approved):
LGTM

armandmcqueen:
This is overloading the meaning of `SlotsPerAgent=0` - here it can mean either static agents and CPU-only agents. I think we should try to differentiate them. Either explicitly return information about whether the pool is using dynamic agents (`bool DynamicAgents`) so clients can do the logic client side (i.e. only look at `SlotsPerAgent` if `DynamicAgents==True`) or a more obvious default when there are static agents (e.g. `SlotsPerAgent=-1`)

Otherwise it looks good

shiyuann:
@armandmcqueen I found we actually have a field `type` returning AWS, GCP, K8S, or static. This can be used to meet the requirement you bring up.

PR #2386, opened 2021-05-17
docs: add instructions to use dtrain workflow for inference with PyTorch

ahj-determined:
## Description

Add documentation to walk users through how they can use the existing dtrain workflow to run distributed batch inference with PyTorch.

## Checklist


examples/computer_vision/cifar10_pytorch_inference/model_def.py
```diff
@@ -0,0 +1,95 @@
...
+        float((predictions.argmax(1) == labels.to(torch.long)).sum()) / predictions.shape[0]
+    )
+
+class CIFARTrial(PyTorchTrial):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        self.context = context
+        self.download_directory = tempfile.mkdtemp()
+
+        # TODO: Load your trained model
+        self.model = self.context.wrap_model(resnet.resnet18(pretrained=True))
```
liamcli, line 40: i don't think the `resnet.py` file and the predownloaded checkpoint are necessary.  

instead, let's follow the steps here to finetune a pretrained model from torchvision so we don't need to include model code and checkpoint in this example.
https://colab.research.google.com/github/pytorch/tutorials/blob/gh-pages/_downloads/df1f5ef1c1a8e1a111e88281b27829fe/finetuning_torchvision_models_tutorial.ipynb#scrollTo=n4pbCfUO0wWs

examples/computer_vision/cifar10_pytorch_inference/model_def.py
```diff
@@ -0,0 +1,95 @@
...
+    return (  # type: ignore
+        float((predictions.argmax(1) == labels.to(torch.long)).sum()) / predictions.shape[0]
+    )
+
+class CIFARTrial(PyTorchTrial):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        self.context = context
+        self.download_directory = tempfile.mkdtemp()
+
+        # TODO: Load your trained model
```
liamcli, line 39: let's add a comment here about ways you can load a model in determined.  e.g. load checkpoint from model registry, load checkpoint from previous experiment, or specifying uuid in experiment config

examples/computer_vision/cifar10_pytorch_inference/model_def.py
```diff
@@ -0,0 +1,95 @@
...
+        # TODO: Perform your evaluation step
+        batch = cast(Tuple[torch.Tensor, torch.Tensor], batch)
+        data, labels = batch
+        output = self.model(data)
+
+        # TODO: Optionally log or save outputs to persistent store
+        print(output)
+        print(labels)
+        '''
+        with open("/path/to/output.txt", "w+") as f:
```
liamcli, line 67: i think it's probably better to write a numpy array instead of writing to a text file.  what we can do is use a custom metric reducer to gather the predictions from across all gpus and then convert that to numpy array and do np.save

liamcli (changes requested):
let me know if you want to discuss any of the comments here

examples/computer_vision/cifar10_pytorch_inference/model_def.py
ahj-determined: Yeah, I agree that it'd be nice if we didn't have to upload a checkpoint. If we did it this way, wouldn't the last layer be untrained and therefore give us bad predictions? Or maybe it's good enough for the purposes for an example.

liamcli: the last layer was trained to classify imagenet classes so we need to remove that layer and add a layer for cifar10 classes.

ahj-determined: Yeah, I'm just wondering if it's a problem that the new cifar10 classifier layer is untrained. I can just try it at see what the accuracy is

ahj-determined: @liamcli I've used the pretrained imagenet in the most recent commit. However, without additional finetuning, this gets a validation error of 88%. With the CIFAR10-trained checkpoint we got an error of ~6%. Do you think we should revert and go back to uploading a large checkpoint or go with the unfinetuned model? I could also use mobilenet pretrained on cifar10 if we wanted to upload a _smaller_ checkpoint.

ahj-determined:
@liamcli I've addressed the above comments. Let me know what you think.

examples/computer_vision/cifar10_pytorch_inference/model_def.py
```diff
@@ -0,0 +1,143 @@
...
+	# from determined.experimental import Determined
+	# checkpoint = Determined().get_experiment(id).top_checkpoint()
+	# model = checkpoint.load()
+
+	### Specify a UUID with `source_trial_id` in the experiment config
+
+	### Load a model that was not trained by Determined
+        self.model = self.context.wrap_model(resnet.resnet18(pretrained=True))
+
+        # IGNORE: Dummy optimizer that needs to be specified but is unused
```
liamcli, line 90: this line looks duplicated

docs/how-to/distributed-training.txt
```diff
@@ -142,6 +142,30 @@ distributed training:
    can also be used to check if any other tasks are using GPUs and
    preventing your experiment from using all the GPUs on a machine.
 
+***********************
+ Distributed Inference
+***********************
+
+PyTorch users can also use the existing distributed training workflow
+with PytorchTrial to accelerate their inference workloads. This workflow
```
liamcli, line 150: let's point to the inference example in the doc here.

ahj-determined: Good catch, added

examples/computer_vision/cifar10_pytorch_inference/model_def.py
ahj-determined: Fixed

liamcli (approved):
lgtm

ahj-determined:
> lgtm once tests pass

Thanks for the review, @liamcli !

PR #2387, opened 2021-05-17
fix: add max size limit metrics [DET-4878] [DET-4783]

azhou-determined:
enforce max metric size object to 100MB

new branch because of a mess trying to fix author on previous branch commits


harness/determined/util.py
```diff
@@ -144,6 +143,14 @@ def wrap_metrics(
         return {"metrics": metrics, "stop_requested": stop_requested, "invalid_hp": invalid_hp}
 
 
+def sizeof_dict(dict: Dict[str, Any]) -> int:
+    """
+    Serializes dict object to json and calculates size in bytes.
+    """
+    dict_json = json_encode(dict)
```
eecsliu, line 150: blocking: What @shiyuann mentioned in the previous PR was potentially moving this check to `_socket_manager.py`. His comment here:

I tried to time json-encoding 100MB object, which takes 27ms. It doesn't hurt much but I'd check the size in _socket_manager.py, where calls json_encode. Then you don't need to encode the metrics twice unnecessarily.

eecsliu (changes requested):
@shiyuann's comment added

harness/determined/util.py
azhou-determined: adding my comment from other PR:

the response in _socket_manager.py includes other data that isn't directly what's recorded on a column-basis to the database. it includes some overhead for other columns:
{'type': 'WORKLOAD_COMPLETED', 'workload': <RUN_STEP (100 Batches): (1,1,10)>, 'start_time': datetime.datetime(2021, 5, 17, 19, 35, 34, 97884, tzinfo=datetime.timezone.utc), 'end_time': datetime.datetime(2021, 5, 17, 19, 35, 37, 558983, tzinfo=datetime.timezone.utc), 'metrics': {...metrics }
not a huge deal, but not 100% accurate in terms of the size of the object being persisted to the column. i guess i could encode the actual metrics piece of that json separately, then manually stitch it together to send in _socket_manager respond, but that seems confusing from a code perspective. what do you guys think?

shiyuann: > the response in _socket_manager.py includes other data that isn't directly what's recorded on a column-basis to the database. it includes some overhead for other columns:
{'type': 'WORKLOAD_COMPLETED', 'workload': <RUN_STEP (100 Batches): (1,1,10)>, 'start_time': datetime.datetime(2021, 5, 17, 19, 35, 34, 97884, tzinfo=datetime.timezone.utc), 'end_time': datetime.datetime(2021, 5, 17, 19, 35, 37, 558983, tzinfo=datetime.timezone.utc), 'metrics': {...metrics }
not a huge deal, but not 100% accurate in terms of the size of the object being persisted to the column. 

@azhou-determined We don't need to be 100% accurate if I understand it correctly. We just need to ensure the master won't be crashed on saving metrics to the database. `Metrics` should take up most of the time in serializing the message.

> i guess i could encode the actual metrics piece of that json separately, then manually stitch it together to send in _socket_manager respond, but that seems confusing from a code perspective. what do you think?

This doesn't seem to be necessary. We actually handle encoding in [here](https://github.com/determined-ai/determined/blob/8560a969b07332aac08bf64de7bc95e6ccc16716/harness/determined/layers/_socket_manager.py#L200) in `_socket_manager.py`. Just need to move the check you wanted to add to this place. What do you think of this approach?

azhou-determined: it doesn't need to be 100% accurate, you're right. i guess the overhead doesn't matter since it's just a few lines/bytes. i'll change it

azhou-determined: @shiyuann @eecsliu one other thing though, the tensorboard writer syncs in _workload_manager before the message is sent to _socket_manager, meaning for a huge metrics object if we check at _socket_manager, it would still attempt to sync with the metrics writer for tensorboard. is this a potential issue?

shiyuann: should be fine since you fail the trial when the metric size exceeds the limit.

azhou-determined: makes sense. moved the check to socket manager

eecsliu (approved):
👩‍🍳 😗 👌

PR #2392, opened 2021-05-18
feat: experiment list search [DET-5460]

hkang1:
## Description

Migrate experiment list search (only applicable to experiment name) to antd table native search filter.

Before:
Screen Shot 2021-05-18 at 12 26 39 PM

After:
Screen Shot 2021-05-18 at 12 26 26 PM

## Checklist


webui/react/src/components/TableFilterSearch.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleSearch = useCallback(() => {
+    if (onSearch) onSearch(search);
+    confirm();
+  }, [ confirm, onSearch, search ]);
+
+  return (
+    <div className={css.base}>
+      <div className={css.search}>
+        <Input
```
apizzini, line 41: I think it would be better if pressing the "Enter" key would automatically trigger filtering.

webui/react/src/components/TableBatch.module.scss
```diff
@@ -1,8 +1,10 @@
 .base {
   height: 0;
   opacity: 0;
+  pointer-events: none;
   position: relative;
   transition: height 0.2s ease-in-out, opacity 0.2s ease-in-out;
+  z-index: 3;
```
apizzini, line 7: Why 3 and not 2? (the only other z-index in this file is 1).

If that's due to another file z-index we should probably use variables for z-index values so that we can understand the related logic. The reasoning behind this is simply that is hard to debug z-index issues (or find the right value for a new one) if those are spread among different files.. With variables we could keep them close.

webui/react/src/components/TableFilterSearch.tsx
apizzini, line 41: another nit thing would be to make sure that the text input is auto-focused after clicking the search icon and making this component render?

apizzini (approved):
This looks good!

Left 3 small comments mostly to:
- improve UX
- evaluate a better way to handle z-indexes

webui/react/src/components/TableBatch.module.scss
hkang1: the variable approach makes sense to me. I'll add this to the webUI AG topic

PR #2393, opened 2021-05-18
fix: wire up support for plain-string image config

rb-determined-ai:
## Test Plan

Tested manually and added test case.


ioga (approved):

rb-determined-ai:
only failed a flaky webui test, merging anyway.

PR #2394, opened 2021-05-18
ci: temporarily remove flaky tests

hkang1:
## Checklist

### DIFF
--- a/webui/tests/specs/02-experiments.spec
+++ b/webui/tests/specs/02-experiments.spec
@@ -24,13 +24,13 @@ Specification to test the experiment list page.
   |Kill               |false   |
 * Toggle all table row selection
 
-## Filter experiments by archived
+// ## Filter experiments by archived
 
-* Scroll table to the "right"
-* Filter table header "Archived" with option "Archived"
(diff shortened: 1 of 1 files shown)


shiyuann (approved):

webui/tests/specs/02-experiments.spec
```diff
@@ -24,13 +24,13 @@ Specification to test the experiment list page.
   |Kill               |false   |
 * Toggle all table row selection
 
-## Filter experiments by archived
+// ## Filter experiments by archived
```
RehanSD, line 27: Is there a reason we are using `//` instead of `#`?

RehanSD (commented):
Looks good to me! Left one quick clarification question, but other than that I think it can be merged!

webui/tests/specs/02-experiments.spec
hkang1: for gauge tests, it uses a special markdown and `#` is treated as a title/section

RehanSD: Thank you!

PR #2398, opened 2021-05-19
feat: add launch notebook modal [DET-5376] [DET-5377] [DET-5380] [DET-5378] [DET-5379] [DET-5375]

EmilyBonar:
## Description
Created the new modal for launching notebooks in the webui and hooked into APIs.

This addresses tickets [DET-5376] [DET-5377] [DET-5380] [DET-5378] [DET-5379] [DET-5375].

## Test Plan
Notebook Launch
Manual testing. Attached user flow gif.

## Checklist

[DET-5376]: https://determinedai.atlassian.net/browse/DET-5376
[DET-5377]: https://determinedai.atlassian.net/browse/DET-5377
[DET-5380]: https://determinedai.atlassian.net/browse/DET-5380
[DET-5378]: https://determinedai.atlassian.net/browse/DET-5378
[DET-5379]: https://determinedai.atlassian.net/browse/DET-5379

[DET-5375]: https://determinedai.atlassian.net/browse/DET-5375

### DIFF
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -2,6 +2,7 @@ import { Button, Col, InputNumber, Modal, Row } from 'antd';
 import { Form, Input, Select } from 'antd';
 import { ModalProps } from 'antd/es/modal/Modal';
 import React, { useCallback, useEffect, useState } from 'react';
+import MonacoEditor from 'react-monaco-editor';
 
 import { getResourcePools, getTemplates } from 'services/api';
 import { ResourcePool, Template } from 'types';
@@ -44,6 +45,7 @@ const NotebookModal: React.FC<Props> = (
   { visible = false, ...props }: Props,
 ) => {
   const [ showFullConfig, setShowFullConfig ] = useState(false);
+  const [ fullConfig, setFullConfig ] = useState('');
   const [ templates, setTemplates ] = useState<Template[]>([]);
   const [ resourcePools, setResourcePools ] = useState<ResourcePool[]>([]);
   const [ resourceTypeOptions, setResourceTypeOptions ] =
@@ -68,8 +70,13 @@ const NotebookModal: React.FC<Props> = (
   useEffect(()=> {
     if(showFullConfig) {
       null; //call api to generate configuration
+      //setFullConfig(value)
     }
   }, [ showFullConfig ]);
+
+  const handleConfigChange = useCallback((value) => {
+    setFullConfig(value);
+  },[]);
 
   const handleSecondary = useCallback(() => {
     setShowFullConfig(show => !show);
@@ -77,10 +84,14 @@ const NotebookModal: React.FC<Props> = (
 
   const handleCreateEnvironment = useCallback(
     (values) =>{
-      if(values.template !== '') {
-        launchNotebook(values.resourceType === 'GPU'? values.slots : 0, values.template);
+      if (showFullConfig) {
+        //launchNotebook with full config
       } else {
-        launchNotebook(values.resourceType === 'GPU'? values.slots : 0);
+        if(values.template !== '') {
+          launchNotebook(values.resourceType === 'GPU'? values.slots : 0, values.template);
+        } else {
+          launchNotebook(values.resourceType === 'GPU'? values.slots : 0);
+        }
       }
     },
     [ ],
@@ -144,7 +155,11 @@ const NotebookModal: React.FC<Props> = (
           Read about notebook settings
           </Link>
         </div>
-        <Input.TextArea defaultValue='' />
+        <MonacoEditor
+          height={400}
+          language='yaml'
+          value={fullConfig}
+          onChange={handleConfigChange} />
       </> :
       <Form form={form} initialValues={{ slots:1 }} labelCol={{ span:8 }}>
         <Row justify='end'>
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -190,6 +190,17 @@ const NotebookFullConfig:React.FC<FullConfigProps> = (
   </Form>;
 };
 
+interface LabelledLineProps {
+  content: JSX.Element;
+  label: string;
+}
+
+const LabelledLine: React.FC<LabelledLineProps> = (
+  { label, content }: LabelledLineProps,
+) => {
+  return <div className={css.line}><p>{label}</p>{content}</div>;
+};
+
 const NotebookForm:React.FC<FormProps> = (
   { onChange, fields }: FormProps,
 ) => {
@@ -240,44 +251,59 @@ const NotebookForm:React.FC<FormProps> = (
     fetchResourcePools();
   }, [ fetchResourcePools ]);
 
-  return (<>
-    <Select
-      allowClear
-      placeholder="No template (optional)"
-      value={fields.template}
-      onChange={(value) => onChange({ key: 'template', value: value?.toString() })}>
-      {templates.map(temp =>
-        <Option key={temp.name} value={temp.name}>{temp.name}</Option>)}
-    </Select>
-    <Input
-      placeholder="Name"
-      value={fields.name}
-      onChange={(value) => onChange({ key: 'name', value: value.target.value })} />
-    <Select
-      allowClear
-      placeholder="Pick the best option"
-      value={fields.pool}
-      onChange={(value) => onChange({ key: 'pool', value: value?.toString() })}>
-      {resourcePools.map(pool =>
-        <Option key={pool.name} value={pool.name}>{pool.name}</Option>)}
-    </Select>
+  return (<div className={css.form}>
+    <LabelledLine
+      content = {
+        <Select
+          allowClear
+          placeholder="No template (optional)"
+          value={fields.template}
+          onChange={(value) => onChange({ key: 'template', value: value?.toString() })}>
+          {templates.map(temp =>
+            <Option key={temp.name} value={temp.name}>{temp.name}</Option>)}
+        </Select>}
+      label="Template" />
+    <LabelledLine
+      content = {
+        <Input
+          placeholder="Name"
+          value={fields.name}
+          onChange={(value) => onChange({ key: 'name', value: value.target.value })} />}
+      label="Name" />
+    <LabelledLine
+      content = {
+        <Select
+          allowClear
+          placeholder="Pick the best option"
+          value={fields.pool}
+          onChange={(value) => onChange({ key: 'pool', value: value?.toString() })}>
+          {resourcePools.map(pool =>
+            <Option key={pool.name} value={pool.name}>{pool.name}</Option>)}
+        </Select>}
+      label="Resource Pool" />
     {resourceInfo.showResourceType &&
-      <RadioGroup
-        options={[ { id: ResourceType.CPU, label: ResourceType.CPU },
-          { id: ResourceType.GPU, label: ResourceType.GPU } ]}
-        value={fields.type}
-        onChange={(value) => {
-          onChange({ key: 'type', value: value });
-          onChange({ key: 'slots', value: value === ResourceType.CPU? 0: fields.slots });
-        }} />}
+    <LabelledLine
+      content = {
+        <RadioGroup
+          options={[ { id: ResourceType.CPU, label: ResourceType.CPU },
+            { id: ResourceType.GPU, label: ResourceType.GPU } ]}
+          value={fields.type}
+          onChange={(value) => {
+            onChange({ key: 'type', value: value });
+            onChange({ key: 'slots', value: value === ResourceType.CPU? 0: fields.slots });
+          }} />}
+      label="Type" />}
     {fields.type === ResourceType.GPU &&
-      <InputNumber
-        defaultValue={1}
-        min={1}
-        value={fields.slots}
-        onChange={(value) => onChange({ key: 'slots', value: value })} />
+    <LabelledLine
+      content = {
+        <InputNumber
+          defaultValue={1}
+          min={1}
+          value={fields.slots}
+          onChange={(value) => onChange({ key: 'slots', value: value })} />}
+      label="Slots" />
     }
-  </>);
+  </div>);
 };
 
 export default NotebookModal;
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -20,11 +20,12 @@ const STORAGE_PATH = 'notebook-launch';
 const STORAGE_KEY = 'notebook-config';
 
 interface Props extends ModalProps {
+  onLaunch?: () => void;
   visible?: boolean;
 }
 
 const NotebookModal: React.FC<Props> = (
-  { visible = false, ...props }: Props,
+  { visible = false, onLaunch = (() => null), ...props }: Props,
 ) => {
   const storage = useStorage(STORAGE_PATH);
   const [ showFullConfig, setShowFullConfig ] = useState(false);
@@ -91,8 +92,9 @@ const NotebookModal: React.FC<Props> = (
           values.pool,
         );
       }
+      onLaunch();
     },
-    [ showFullConfig, form ],
+    [ showFullConfig, form, onLaunch ],
   );
 
   const handleResourcePoolUpdate = useCallback((e) => {
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -34,9 +34,7 @@ const NotebookModal: React.FC<Props> = (
   const [ showFullConfig, setShowFullConfig ] = useState(false);
   const [ templates, setTemplates ] = useState<Template[]>([]);
   const [ resourcePools, setResourcePools ] = useState<ResourcePool[]>([]);
-  const [ showResourceType, setShowResourceType ] =useState(
-    storage.getWithDefault(STORAGE_KEY, { type: undefined }).type !== ResourceType.CPU,
-  );
+  const [ showResourceType, setShowResourceType ] = useState<boolean>(true);
   const [ resourceType, setResourceType ] = useState<ResourceType | undefined>(
     storage.getWithDefault(STORAGE_KEY, { type: undefined }).type,
   );
@@ -115,7 +113,10 @@ const NotebookModal: React.FC<Props> = (
 
   const handleResourcePoolUpdate = useCallback((selectedPoolName: string) => {
     const selectedPool = resourcePools.find(pool => pool.name === selectedPoolName);
-    if (!selectedPool) return;
+    if (!selectedPool) {
+      setShowResourceType(true);
+      return;
+    }
     const hasCPUCapacity = selectedPool.cpuContainerCapacityPerAgent > 0;
     const hasGPUCapacity = selectedPool.slotsAvailable > 0
       || (selectedPool.slotsPerAgent && selectedPool.slotsPerAgent > 0);
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -171,7 +171,7 @@ const NotebookModal: React.FC<Props> = (
           value={JSON.stringify(fullConfig, null, 2)}
           onChange={handleConfigChange} />
       </> :
-      <Form form={form} initialValues={{ slots:1 }} labelCol={{ span:8 }}>
+      <Form form={form} initialValues={{ pool: '', slots:1, template: '' }} labelCol={{ span:8 }}>
         <Row justify='end'>
           <Item
             label='Type'
@@ -180,7 +180,7 @@ const NotebookModal: React.FC<Props> = (
             rules={[ { message: 'Please choose a resource type', required: true } ]}>
             <RadioGroup
               options={resourceTypeOptions}
-              onChange={(e) => handleTypeUpdate(e)} />
+              onChange={handleTypeUpdate} />
           </Item>
           <Col span={11}>
             { resourceType === 'GPU'?
@@ -195,31 +195,28 @@ const NotebookModal: React.FC<Props> = (
           </Col>
         </Row>
         <Item label='Notebook Template' name='template'>
-          <Dropdown options={templates.map(template => template.name)} />
+          <Select
+            style={{ minWidth:120 }}>
+            <Option key='' value=''>---Empty---</Option>
+            {templates.map(temp =>
+              <Option key={temp.name} value={temp.name}>{temp.name}</Option>)}
+          </Select>
         </Item>
         <Item label='Name' name="name">
           <Input placeholder='Name' />
         </Item>
         <Item label='Resource Pool' name="pool">
-          <Dropdown
-            options={resourcePools.map(pool => pool.name)}
-            onChange={handleResourcePoolUpdate} />
+          <Select
+            style={{ minWidth:120 }}
+            onChange={handleResourcePoolUpdate}>
+            <Option key='' value=''>---Empty---</Option>
+            {resourcePools.map(pool =>
+              <Option key={pool.name} value={pool.name}>{pool.name}</Option>)}
+          </Select>
         </Item>
       </Form>
     }
   </Modal>;
-};
-
-interface DropdownProps {
-  onChange?: (e: string) => void;
-  options?: string[];
-}
-
-const Dropdown: React.FC<DropdownProps> = ({ options, onChange }: DropdownProps) => {
-  return options? <Select defaultValue='' style={{ minWidth:120 }} onChange={onChange}>
-    <Option key='empty' value=''>---Empty---</Option>
-    {options.map(option => <Option key={option} value={option}>{option}</Option>)}
-  </Select> : null;
 };
 
 export default NotebookModal;
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -75,82 +75,6 @@ interface ResourceInfo {
   showResourceType: boolean;
 }
 
-/*
-const resourcePools = [ {
-  agentDockerImage: '',
-  agentDockerNetwork: '',
-  agentDockerRuntime: '',
-  agentFluentImage: '',
-  containerStartupScript: '',
-  cpuContainerCapacity: 100,
-  cpuContainerCapacityPerAgent: 100,
-  cpuContainersRunning: 0,
-  defaultCpuPool: true,
-  defaultGpuPool: true,
-  description: '',
-  details: {
-    aws: null,
-    gcp: null,
-    priorityScheduler: null,
-  },
-  imageId: '',
-  instanceType: '',
-  location: 'on-prem',
-  masterCertName: '',
-  masterUrl: '',
-  maxAgents: 0,
-  maxAgentStartingPeriod: 0,
-  maxIdleAgentPeriod: 0,
-  minAgents: 0,
-  name: 'both',
-  numAgents: 1,
-  preemptible: false,
-  schedulerFittingPolicy: 'FITTING_POLICY_BEST',
-  schedulerType: 'SCHEDULER_TYPE_FAIR_SHARE',
-  slotsAvailable: 1,
-  slotsPerAgent: -1,
-  slotsUsed: 0,
-  startupScript: '',
-  type: 'RESOURCE_POOL_TYPE_STATIC',
-}, {
-  agentDockerImage: '',
-  agentDockerNetwork: '',
-  agentDockerRuntime: '',
-  agentFluentImage: '',
-  containerStartupScript: '',
-  cpuContainerCapacity: 100,
-  cpuContainerCapacityPerAgent: 100,
-  cpuContainersRunning: 0,
-  defaultCpuPool: true,
-  defaultGpuPool: true,
-  description: '',
-  details: {
-    aws: null,
-    gcp: null,
-    priorityScheduler: null,
-  },
-  imageId: '',
-  instanceType: '',
-  location: 'on-prem',
-  masterCertName: '',
-  masterUrl: '',
-  maxAgents: 0,
-  maxAgentStartingPeriod: 0,
-  maxIdleAgentPeriod: 0,
-  minAgents: 0,
-  name: 'cpu',
-  numAgents: 1,
-  preemptible: false,
-  schedulerFittingPolicy: 'FITTING_POLICY_BEST',
-  schedulerType: 'SCHEDULER_TYPE_FAIR_SHARE',
-  slotsAvailable: -1,
-  slotsPerAgent: -1,
-  slotsUsed: 0,
-  startupScript: '',
-  type: 'RESOURCE_POOL_TYPE_STATIC',
-} ];
-*/
-
 const NotebookModal: React.FC<NotebookModalProps> = (
   { visible = false, onLaunch, ...props }: NotebookModalProps,
 ) => {
@@ -315,10 +239,6 @@ const NotebookForm:React.FC<FormProps> = (
   useEffect(() => {
     fetchResourcePools();
   }, [ fetchResourcePools ]);
-
-  useEffect(() => {
-    console.log(fields);
-  }, [ fields ]);
 
   return (<>
     <Select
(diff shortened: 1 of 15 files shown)


webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+          slots:1,
+          template: '',
+          type: undefined,
+        })}
+        labelCol={{ span:8 }}
+        onValuesChange={storeConfig}>
+        <Item label='Notebook Template' name='template'>
+          <Select
+            style={{ minWidth:120 }}>
+            <Option key='' value=''>---Empty---</Option>
```
hkang1, line 188: **suggestion:** instead of an empty option, use the `placeholder` prop on `Select`. something like `placeholder="Select a notebook template"`

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+            {templates.map(temp =>
+              <Option key={temp.name} value={temp.name}>{temp.name}</Option>)}
+          </Select>
+        </Item>
+        <Item label='Name' name="name">
```
hkang1, line 193: **nit:** use double quotes for string props `"Name"` instead of `'Name'` for consistency (will look into if this is something we can auto lint)

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+          <Input placeholder='Name' />
+        </Item>
+        <Item label='Resource Pool' name="pool">
+          <Select
+            style={{ minWidth:120 }}
+            onChange={handleResourcePoolUpdate}>
+            <Option key='' value=''>---Empty---</Option>
```
hkang1, line 200: ditto - something like `placeholder="Select a resource pool"`

webui/react/src/contexts/Store.tsx
```diff
@@ -60,6 +63,7 @@ type Action =
 | { type: StoreAction.SetAuth; value: Auth }
 | { type: StoreAction.SetAuthCheck }
 | { type: StoreAction.SetInfo; value: DeterminedInfo }
+| { type: StoreAction.SetNotebookConfig; value: NotebookConfig }
```
hkang1, line 66: **question:** looks like this isn't really being used? maybe old code that just needs to be pruned? same with the store action in `NotebookModalContainer` for the NotebookModal story

hkang1: basically the notebook config state can probably persist at the modal component level and not have to be accessible or shared beyond that boundary

webui/react/src/components/NotebookModal.tsx
EmilyBonar: Choosing the empty option has semantic meaning in this case - it means that you want to use the default template. Using a placeholder would mean that if you choose a template and then change your mind, you wouldn't be able to go back to the default, as the placeholder disappears once something else is chosen.

webui/react/src/services/api.ts
```diff
@@ -163,10 +163,18 @@ export const killTensorboard = generateDetApi<
   CommandIdParams, Api.V1KillTensorboardResponse, void
 >(Config.killTensorboard);
 
+export const getTemplates = generateDetApi<
```
hkang1, line 166: **nit:** Not 100% sure on this but I think there are a couple of different types of template (not in the WebUI yet) but might be better to be explicit now and just call it `getTaskTemplates`?

webui/react/src/components/NotebookModal.tsx
EmilyBonar: Ditto above. I think being able to consciously choose to use the default resource pool is valuable.

webui/react/src/components/NotebookModal.tsx
EmilyBonar: It looks like I've used both throughout the code. Should I switch all JSX string props to double quotes?

webui/react/src/components/NotebookModal.tsx
hkang1: I see. If that's the case, I would recommend the option label as `Default Task Template` or something to that effect so the user gets a bit more info of what they are selecting.

webui/react/src/components/NotebookModal.tsx
hkang1: after our offline discussion, we will want this to a be required field, so the placeholder approach for this field still makes sense (unlike the template select fiele)

webui/react/src/contexts/Store.tsx
EmilyBonar: Storybook yells at me that "useStore must be used within a StoreProvider" if I don't do this, so I added it.

EmilyBonar: The tickets call for form options except for name to be saved in local storage, I can take that out if you'd prefer though.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+        const config = await previewNotebook(
+          values.slots,
+          values.template,
+          values.name,
+          values.pool,
+        );
+        form.setFieldsValue({ config: yaml.dump(config) });
+      }
+    };
+    fetchConfig(form.getFieldsValue(true));
```
hkang1, line 66: **chore:** we probably don't want to call `fetchConfig` at all if `showFullConfig` is not true. Move the `if (showFullConfig) {` to wrap around the `fetchConfig` function call.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const [ resourceTypeOptions, setResourceTypeOptions ] =
+    useState<{id:string, label:string}[]>([ { id:'CPU', label:'CPU' }, { id:'GPU', label:'GPU' } ]);
+  const [ resourceType, setResourceType ] =
+    useState(storage.getWithDefault(STORAGE_KEY, { type: undefined }).type);
+  const [ form ] = Form.useForm();
+
+  useEffect(() => {
+    const fetchTemplates = async () => {
+      setTemplates(await getTemplates({}));
+    };
```
hkang1, line 43: **chore:** let's redefine these fetch functions with `useCallback` hooks outside of the `useEffect` and add try/catch blocks. Otherwise if the api fails, the thrown errors will break the react app.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    fetchTemplates();
+  }, []);
+
+  useEffect(() => {
+    const fetchResourcePools = async () => {
+      setResourcePools( await getResourcePools({}));
+    };
```
hkang1, line 50: ditto (same as the `fetchTemplates` comment)

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    fetchResourcePools();
+  }, []);
+
+  useEffect(()=> {
+    const fetchConfig = async (values: NotebookConfig) => {
```
hkang1, line 55: ditto (same as the `fetchTemplates` comment)

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+const NotebookModal: React.FC<Props> = (
+  { visible = false, onLaunch = (() => null), ...props }: Props,
+) => {
+  const storage = useStorage(STORAGE_PATH);
+  const [ showFullConfig, setShowFullConfig ] = useState(false);
+  const [ templates, setTemplates ] = useState<Template[]>([]);
+  const [ resourcePools, setResourcePools ] = useState<ResourcePool[]>([]);
+  const [ resourceTypeOptions, setResourceTypeOptions ] =
+    useState<{id:string, label:string}[]>([ { id:'CPU', label:'CPU' }, { id:'GPU', label:'GPU' } ]);
```
hkang1, line 35: **suggestion:** use `ResourceType` from `types` for `CPU` and `GPU` as opposed to string values to centralize resource type definitions.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+const STORAGE_PATH = 'notebook-launch';
+const STORAGE_KEY = 'notebook-config';
+
+interface Props extends ModalProps {
+  onLaunch?: () => void;
+  visible?: boolean;
+}
+
+const NotebookModal: React.FC<Props> = (
+  { visible = false, onLaunch = (() => null), ...props }: Props,
```
hkang1, line 28: **suggestion:** my preference is to leave it undefined (not setting a default), and wherever you are running `onLaunch` to do `if (onLaunch) onLaunch();`. The idea is that checking whether `onLaunch` is defined or not is a cheaper operation than having JS to schedule the `onLaunch` function call into the event loop and creating a call stack for it, even though it simply returns a `null`.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    {...props}>
+    {showFullConfig?
+      <Form form={form}>
+        <div style={{
+          backgroundColor:'rgb(230,230,230)',
+          border:'1px solid rgb(200,200,200)',
+          marginBottom: '4px',
+          padding: 2,
+          textAlign:'center',
+        }}
```
hkang1, line 147: **suggestion:** create a `NotebookModal.module.scss` file to extract and isolate styles and use CSS theme vars wherever possible. So it would be something like:
`NotebookModal.module.scss`
```
.note {
  background-color: var(--theme-colors-monochrome-16),
  ...,
}
```
`NotebookModal.tsx`
```
import css from './NotebookModal.module.scss';
...
<div className={css.note}>
  ...
</div>
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,234 @@
+import { Button, Col, InputNumber, Modal, Row } from 'antd';
+import { Form, Input, Select } from 'antd';
+import { ModalProps } from 'antd/es/modal/Modal';
+import yaml from 'js-yaml';
+import React, { useCallback, useEffect, useState } from 'react';
+import MonacoEditor from 'react-monaco-editor';
```
hkang1, line 6: **suggestion:** take a look at `ExperimentInfoBox.tsx` on how `monaco-editor` gets imported via `React.lazy`. This is because `monaco-editor` is a really really big vendor package in terms of bundle size and we only want to load it when absolutely necessary.
```
const MonacoEditor = React.lazy(() => import('react-monaco-editor'));

...

// in JSX
        <React.Suspense
          fallback={<div className={css.loading}><Spinner className="minHeight" /></div>}>
          <MonacoEditor {...props} />
        </React.Suspense>
```
`React.Suspense` will basically show a loader while our WebUI loads a separate bundle for `monaco-editor`

webui/react/src/components/NotebookModal.tsx
hkang1: 🙏

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+              <Item
+                label="Number of Slots"
+                labelCol={{ span:14 }}
+                name="slots"
+                rules={[ { message: 'Please choose a number of slots', required: true } ]}>
+                <InputNumber min={1} />
+              </Item> : null
+            }
+          </Col>
+        </Row>
```
hkang1, line 241: **chore:** let me know if you want to discuss this one in more detail:

2021-05-20 at 1 28 PM

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+            <Item
+              label="Type"
+              labelCol={{ span:8 }}
+              name="type"
+              rules={[ { message: 'Select a resource type', required: true } ]}>
+              <RadioGroup
+                options={resourceTypeOptions}
+                onChange={handleTypeUpdate} />
+            </Item>
+          </Col>
```
hkang1, line 229: **style:** minor adjustment for mobile view

2021-05-20 at 10 53 AM

this looks ok when both options are present but looks off when it's only `Type` (no Number of Slots field)

hkang1 (changes requested):
Nice work with the notebook persistence project! Most of this looks great! Sprinkled some suggestions and comments through out the PR. Let me know if you want to go over anything specific in more detail.

hkang1:
love the GIF preview!

webui/react/src/services/decoder.ts
```diff
@@ -147,6 +147,10 @@ export const mapV1Tensorboard =
     };
   };
 
+export const mapV1Template = (template: Sdk.V1Template): types.Template => {
+  return { config: template.config || {}, name: template.name || '' };
```
hkang1, line 151: @shiyuann just want to confirm that changing `proto/src/determined/template/v1/template.proto`to the following is ok (making the template `name` and `config` required in the JSON schema):
```
syntax = "proto3";

package determined.template.v1;
option go_package = "github.com/determined-ai/determined/proto/pkg/templatev1";

import "google/protobuf/struct.proto";
import "protoc-gen-swagger/options/annotations.proto";

// Templates move settings that are shared by many experiments into a single
// YAML file.
message Template {
    json_schema: { required: [ "name", "config" ] }
  };
  // The name of the template.
  string name = 1;
  // The template value.
  google.protobuf.Struct config = 4;
}
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+          values.name,
+          values.pool,
+        );
+      }
+      if (onLaunch) onLaunch();
+    },
+    [ showFullConfig, form, onLaunch, resourceType ],
+  );
+
+  const handleResourcePoolUpdate = useCallback((e: string) => {
```
hkang1, line 112: **suggestion:** rename `e` to `selectedPoolName` so it's a bit more descriptive of what the parameter holds.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+    const pool = resourcePools.find(pool => pool.name === e);
```
hkang1, line 113: suggestion: rename e to selectedPoolName so it's a bit more descriptive of what the parameter holds. rename `const pool` to `const selectedPool`

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+    if (pool){
```
hkang1, line 114: **suggestion:** instead of a nested check, do an early exit so it flattens the code. For example, instead of:
```
if (a) {
  // do stuff
}
```
do:
```
if (!a) return;
// do stuff
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+      if (pool.cpuContainerCapacityPerAgent > 0 && pool.slotsPerAgent && pool.slotsPerAgent > 0) {
+        setShowResourceType(true);
+      } else if (pool.cpuContainerCapacityPerAgent > 0) {
+        setResourceType(ResourceType.CPU);
+        setShowResourceType(false);
+      } else if (pool.slotsPerAgent && pool.slotsPerAgent > 0) {
+        setResourceType(ResourceType.GPU);
+        setShowResourceType(false);
+      }
```
hkang1, line 123: **suggestion:** define the checks as a variable so do you don't perform the same check multiple times. So instead of:
```
if (a > 0 && b > 0) {
} else if (a > 0) {
} else if (b > 0) {
}
```
do:
```
const aNotZero = a > 0;
const bNotZero = b > 0;
if (aNotZero && bNotZero) {
} else if (aNotZero) {
} else if (bNotZero) {
}
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+      if (onLaunch) onLaunch();
+    },
+    [ showFullConfig, form, onLaunch, resourceType ],
+  );
+
+  const handleResourcePoolUpdate = useCallback((e: string) => {
+    const pool = resourcePools.find(pool => pool.name === e);
+    if (pool){
+      if (pool.cpuContainerCapacityPerAgent > 0 && pool.slotsPerAgent && pool.slotsPerAgent > 0) {
```
hkang1, line 115: **suggestion:** I checked the default resource pool and saw that `slotsPerAgent` can be set to `-1` to indicate that there is GPU capacity but just no ceiling. So we may have to do the check something more like:
```
const hasCPUCapacity = selectedPool.cpuContainerCapacityPerAgent !== 0;
const hasGPUCapacity = selectedPool.slotsPerAgent !== 0;
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+        setResourceType(ResourceType.CPU);
+        setShowResourceType(false);
+      } else if (pool.slotsPerAgent && pool.slotsPerAgent > 0) {
+        setResourceType(ResourceType.GPU);
+        setShowResourceType(false);
+      }
+    }
+  },[ resourcePools ]);
+
+  const handleTypeUpdate = useCallback((e) => {
```
hkang1, line 127: **nit:** rename `e` to something more descriptive (e.g. `selectedResourceType`)

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+}
+
+const NotebookModal: React.FC<Props> = (
+  { visible = false, onLaunch, ...props }: Props,
+) => {
+  const storage = useStorage(STORAGE_PATH);
+  const [ showFullConfig, setShowFullConfig ] = useState(false);
+  const [ templates, setTemplates ] = useState<Template[]>([]);
+  const [ resourcePools, setResourcePools ] = useState<ResourcePool[]>([]);
+  const [ showResourceType, setShowResourceType ] = useState(true);
```
hkang1, line 37: **chore:** this should be `false` to begin with and once the conditions are satisfied below, it can be set to `true`.

2021-05-24 at 8 17 AM

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+            <MonacoEditor
+              height={400}
+              language="yaml"
+              options={{
+                minimap: { enabled: false },
+                scrollBeyondLastLine: false,
+                selectOnLineNumbers: true,
+              }} />
+          </Item>
+        </React.Suspense>
```
hkang1, line 176: nice work with lazy loading!

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+    visible={visible}
+    {...props}>
+    {showFullConfig ?
+      <Form form={form}>
+        <div className={css.note}
+        >
+          <Link external path="/docs/reference/command-notebook-config.html">
+          Read about notebook settings
+          </Link>
+        </div><React.Suspense
```
hkang1, line 153: **nit:** newline between `</div>` and `<React.Suspense`

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+        {showResourceType && <Item
+          label="Type"
+          name="type"
+          rules={[ { message: 'Select a resource type', required: true } ]}>
+          <RadioGroup
+            options={[ { id:ResourceType.CPU, label:ResourceType.CPU },
+              { id:ResourceType.GPU, label:ResourceType.GPU } ]}
+            onChange={handleTypeUpdate} />
+        </Item>}
+        { resourceType === 'GPU' ?
```
hkang1, line 217: *nit:** remove the space before `resourceType`

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+        values.pool,
+      );
+      form.setFieldsValue({ config: yaml.dump(config) });
+    } catch {}
+  }, [ form ]);
+
+  useEffect(()=> {
+    if (showFullConfig){
+      fetchConfig();
+    }
```
hkang1, line 78: **super nit:** feel free to ignore this. I personally like the simple one liners in these types of situations to condense the code a little bit, since it's pretty easy to read too:
```
if (showFullConfig) fetchConfig();
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+    storage.set(STORAGE_KEY,values);
+  }, [ storage ]);
+
+  const handleSecondary = useCallback(() => {
+    if (showFullConfig) {
+      setShowFullConfig(show => !show);
+    } else {
+      form.validateFields().then(() => setShowFullConfig(show => !show)).catch();
+    }
+  },[ form, showFullConfig ]);
```
hkang1, line 92: **suggestion:** rewrite as an `async/await` form:
```
const handleSecondary = useCallback(async () => {
  if (showFullConfig) {
    setShowFullConfig(false);
  } else {
    try {
      await form.validateFields();
      setShowFullConfig(true);
    } catch (e) {}
  }
}, [ form, showFullConfig ]);
```
Also note that we don't need to check the previous state of the `showFullConfig` variable because we already checked if it was true or false.

webui/react/src/components/NotebookModal.tsx
shiyuann: @hkang1 Resoruce pools should actually be optional. Sorry for the recent change. We used to say we can label resource pools to be default CPU/GPU pools. But we recently decided that there might be many different types of resource pools, not limited to CPU and GPU. It might not be ideal to choose any resource pool to be the default so I think it's okay to allow it to be empty, which is the same behavior as the CLI.

webui/react/src/components/NotebookModal.tsx
shiyuann: @hkang1 we should check if the value be larger than 0 rather than `!== 0`. `slotsPerAgent == -1` now means we don't have dynamic agents configured but there still might be static agents that have GPU slots.

webui/react/src/services/decoder.ts
shiyuann: @hkang1 @EmilyBonar I think it's ok.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+                selectOnLineNumbers: true,
+              }} />
+          </Item>
+        </React.Suspense>
+      </Form> :
+      <Form
+        form={form}
+        initialValues={storage.getWithDefault(STORAGE_KEY, {
+          slots:1,
+          template: 'default',
```
hkang1, line 186: **chore:** this be `undefined` for the initial value. `default` may not actually map to a resource pool?

2021-05-25 at 10 39 AM

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+  useEffect(() => {
+    fetchResourcePools();
+  }, [ fetchResourcePools ]);
+
+  const fetchConfig = useCallback(async () => {
+    try {
+      const values: NotebookConfig = form.getFieldsValue(true);
+      const config = await previewNotebook(
+        values.type === ResourceType.CPU ? 0 : values.slots,
+        values.template === 'default'? undefined : values.template,
```
hkang1, line 67: **chore:** extract `default` as a constant up near the top of this file, since there are probably ~3 references to this from what I've seen.

hkang1 (approved):
Nice work with all the changes @EmilyBonar ! This looks great. One final non-blocking small constant extraction, but once you make that change, go ahead and land this!

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+      if (onLaunch) onLaunch();
+    },
+    [ showFullConfig, form, onLaunch, resourceType ],
+  );
+
+  const handleResourcePoolUpdate = useCallback((selectedPoolName: string) => {
+    const selectedPool = resourcePools.find(pool => pool.name === selectedPoolName);
+    if (!selectedPool) return;
+    const hasCPUCapacity = selectedPool.cpuContainerCapacityPerAgent > 0;
+    const hasGPUCapacity = selectedPool.slotsPerAgent && selectedPool.slotsPerAgent > 0;
```
shiyuann, line 117: This should be `slotsAvailable > 0` (for static agents) or `slotsPerAgent > 0` (for dynamic agents).

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+    if (hasCPUCapacity && hasGPUCapacity) {
+      setShowResourceType(true);
+    } else if (hasCPUCapacity) {
+      setResourceType(ResourceType.CPU);
+      setShowResourceType(false);
```
shiyuann, line 122: If there is CPU capacity, the `resources.slots` should be filled with 0 rather than 1.

EmilyBonar: If there is only CPU capacity, the slots field doesn't show up and it is treated as 0 slots.

webui/react/src/components/NotebookModal.tsx
EmilyBonar: How do I know if it is a static vs dynamic agent? Also, what is the behavior if someone tries to use more slots than are available?

shiyuann: Actually, you don't need to know if the pool has static agents or dynamic agents. If `slotsAvailable > 0`, it means agents of more than one slots are connected. `slotsPerAgent > 0` means dynamic agents of more than one slots are configured. `slotsAvailable > 0` alone cannot determine if there might be more dynamic agents of more than one slots connected. So `slotsAvailable > 0 or slotsPerAgent > 0` would be the expression to determine if there exist GPU capacity.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+) => {
+  const [ field, setField ] = useState([ { name: 'config', value: '' } ]);
+
+  useEffect(() => {
+    setField([ { name: 'config', value: yaml.dump(config) } ]);
+  }, [ config ]);
+
+  const extractConfig = useCallback((fieldData) => {
+    onChange(JSON.parse(fieldData[0].value));
+  }, [ onChange ]);
```
hkang1, line 154: **chore:** if you go to `edit full config` mode and then select all config text and delete it, the React app will crash. I can't remember the specific details but I think `JSON.parse` is the problem here. Shouldn't this be something like `yaml.dump(fieldData[0].value)`?

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+  return <Form
+    fields={field}
+    onFieldsChange={(_, allFields) => {
+      extractConfig(allFields);
+    }}>
```
hkang1, line 160: **nit:** just point this `extractConfig` and rename `extractConfig` to something like `handleConfigChange` and that function can look something like:
```
const handleConfigChange = useCallback((_, allFields) => {
  if (!Array.isArray(allFields) || allFields.length === 0) return;
  const configString = allFields[0].value;
  try {
    const config = yaml.dump(configString);
    onChange(config);
  } catch (e) {}
}, []);
```
```

File: webui/react/src/components/NotebookModal.tsx [hunk 35]
```diff
@@ -0,0 +1,309 @@
...
+      <Item
+        name="config"
+        noStyle
+        rules={[ { message: 'Invalid YAML', required: true }, () => ({
+          validator(_, value) {
+            try {
+              yaml.load(value);
+              return Promise.resolve();
+            } catch(err) {
+              return Promise.reject(new Error('Invalid YAML'));
```
hkang1, line 176: **question:** does `err` hold any info on where the parsing error might be in the YAML config? It might be better to show that error if provides the user clues as to how to fix the YAML as these config files can get large for certain users.

webui/react/src/omnibar/tree-extension/trees/index.ts
```diff
@@ -164,11 +164,11 @@ const root: NonLeafNode = {
         {
           options: [
             {
-              onAction: () => launchNotebook(0),
+              onAction: () => launchNotebook(undefined, 0),
               title: 'zeroSlot',
             },
             {
-              onAction: () => launchNotebook(1),
+              onAction: () => launchNotebook(undefined, 1),
```
hkang1, line 171: nice good catch!

webui/react/src/utils/task.ts
```diff
@@ -1,16 +1,37 @@
...
+      });
+    } else {
+      notebook = await apiLaunchNotebook({
+        config: {
+          description: name === '' ? undefined : name,
+          resources: { resource_pool: pool === '' ? undefined : pool, slots },
+        },
+        templateName: templateName === '' ? undefined : templateName,
+      });
+    }
```
hkang1, line 34: **suggestion:** this portion can be a bit more succinct like this:
```
const notebook = await apiLaunchNotebook({
  config: config || {
    description: name === '' ? undefined : name,
    resources: { resource_pool: pool === '' ? undefined : pool, slots },
  },
  templateName: templateName === '' ? undefined : templateName,
});
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+    key: keyof NotebookConfig,
+    value: string | number | undefined
+  }>)
+
+function reducer(
+  state: NotebookConfig,
+  action: {key: keyof NotebookConfig, value: string | number | undefined},
+): NotebookConfig {
+  return { ...state, [action.key]: action.value };
+}
```
hkang1, line 36: **suggestion:** we can refactor this later, but take a look at the pattern we use for `useReducer` in `ExperimentVisualizationFilters.tsx` (particularly the `ActionType`, `Action` and `reducer`). It's a bit more verbose, but the biggest benefit of our existing pattern is that the dispatch is much more strongly typed and will prevent incorrect usages. For example, it's possible to set the `name` property of notebook config with a `number` type even though `name` is a `string` type. The other benefit is that the reducer can also perform more complex operations on the state change, though not needed in this case.

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+      <NotebookFullConfig config={config} onChange={(newConfig) => setConfig(newConfig)} /> :
+      <NotebookForm fields={fields} onChange={dispatch} />
+    }
+  </Modal>;
+};
+
+const NotebookFullConfig:React.FC<FullConfigProps> = (
+  { config, onChange }:FullConfigProps,
+) => {
+  const [ field, setField ] = useState([ { name: 'config', value: '' } ]);
```
hkang1, line 146: **nit:** plural since it has two fields `name` and `value` defined?
`const [ fields, setFields ] = ...`

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+        type="primary"
+        onClick={handleCreateEnvironment}>Launch</Button>
+    </>}
+    title="Launch JupyterLab"
+    visible={visible}
+    width={540}
+    {...props}>
+    {showFullConfig ?
+      <NotebookFullConfig config={config} onChange={(newConfig) => setConfig(newConfig)} /> :
+      <NotebookForm fields={fields} onChange={dispatch} />
```
hkang1, line 138: **suggestion:** Let's leave this change for later (after this PR lands).
Looking at `NotebookModal` it's really large at this point with several large components defined within this file (`NotebookFullConfig`, `NotebookForm`, `NotebookModal`). I would recommend breaking this up and reorganizing as such:
```
src/components/NotebookModal.tsx
src/components/NotebookModal/NotebookFullConfig.tsx
src/components/NotebookModal/NotebookForm.tsx
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+  const storeConfig = useCallback((values: NotebookConfig) => {
+    delete values.name;
+    storage.set(STORAGE_KEY, values);
+  }, [ storage ]);
+
+  useEffect(() => {
+    storeConfig(state);
+  }, [ state, storeConfig ]);
+
```
hkang1, line 53: **suggestion:** Let's leave this for later (after landing this PR).
I think I roughly understand the motivation of wanting to group the form related state code and the related storage code roughly together. However it's fairly small and the reusability aspect of it is not compelling enough to stand on its own. This can probably just sit inside of `NotebookModal` component.

hkang1 (changes requested):
It's getting really close! Last remaining item of resolving the uncaught error when clearing out the config form.

webui/react/src/components/NotebookModal.tsx
EmilyBonar: I was thinking of it as singular since that is describing the single object of the config textbox.

hkang1 (approved):
LGTM! tried out a few scenarios and it worked great. The targeted YAML errors are great and storybook ran great.

PR #2399, opened 2021-05-19
chore: make pip happy again

rb-determined-ai:
## Description

websocket-client just released 1.0, which conflicts with docker-compose,
which breaks the determined install with older version of pip.

See https://app.circleci.com/pipelines/github/determined-ai/determined/12416/workflows/4d41102c-4e72-46e9-842f-6226121f2429/jobs/360714 for an example.

### DIFF
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -25,6 +25,7 @@ setup(
         "pyzmq>=18.1.0",
         "yogadl==0.1.4",
         # Common:
+        "backoff",
         "google-cloud-storage>=1.20.0",
         # google-cloud-core 1.4.2 breaks our windows cli tests for python 3.5.
         "google-cloud-core<1.4.2",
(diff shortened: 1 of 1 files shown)


harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -797,6 +797,7 @@
         "image": {
             "type": [
                 "object",
+                "string",
```
ioga, line 800: I assume you're going to rebase away from that schema change.

harness/setup.py
```diff
@@ -25,6 +25,7 @@
         "pyzmq>=18.1.0",
         "yogadl==0.1.4",
         # Common:
+        "backoff",
```
ioga, line 28: why this?

ioga (approved):

harness/determined/common/schemas/expconf/_gen.py
rb-determined-ai: oops, will fix

harness/setup.py
rb-determined-ai: `backoff` was added recently and not sorted correctly according to the comments in the file.

PR #2401, opened 2021-05-19
chore: only return the port binding appropriate for the proxy [DET-5495]

stoksc:
## Description
This change alters how we derive container addresses. When docker changed to make ipv6 show up by default again it threw us for a loop. My hasty fix wasn't quite correct but also introduced with it a bug that also seemed to be docker port related. This lead me to hackily fix the bug of the bug fix rather than realize the root cause, since a revert of the original fix of bug caused the docker ipv6 change to not allow any containers to run. This change fixes both issues by always just returning a single container address when the container is bound to 0.0.0.0 or :: based on whether the proxy is ipv4 or ipv6.

## Test Plan
- [x] test on an affected machine

## Checklist

### DIFF
--- a/master/pkg/agent/master_message.go
+++ b/master/pkg/agent/master_message.go
@@ -62,12 +62,7 @@ type ContainerStarted struct {
 // started information.
 func (c ContainerStarted) Addresses() []container.Address {
 	proxy := c.ProxyAddress
-
-	proxyIP, _, err := net.ParseCIDR(proxy)
-	if err != nil {
-		panic(errors.Wrapf(err, "failed to parse ip %s", proxy))
-	}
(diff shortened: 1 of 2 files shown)


master/pkg/agent/master_message.go
```diff
@@ -61,6 +62,8 @@ type ContainerStarted struct {
 // started information.
 func (c ContainerStarted) Addresses() []container.Address {
 	proxy := c.ProxyAddress
+	proxyIsIPv4 := len(net.ParseIP(proxy)) == net.IPv4len
```
ioga, line 65:
```suggestion
	proxyIsIPv4 := net.ParseIP(proxy).To4() != nil
```

ioga (approved):

PR #2403, opened 2021-05-19
fix: correct resource pool pagination and make sort sticky [DET-5482]

hkang1:
## Description

- add support for customizable pagination limits for resource pools table
- save the user's last sort order on resource pool

## Checklist


armandmcqueen (approved):
🚀

PR #2405, opened 2021-05-19
fix: avoid rc dev release mismatch notifications

hkang1:
## Description

UPDATE: decided to disable this notification all together for development environments, as it produces more false positives than actual helpful messaging.

2021-05-19 at 2 10 PM

## Checklist

### DIFF
--- a/webui/react/src/App.tsx
+++ b/webui/react/src/App.tsx
@@ -40,10 +40,9 @@ const AppView: React.FC = () => {
 
     /*
      * Check to make sure the WebUI version matches the platform version.
-     * If the release is a release candidate or a dev tag, skip this notification.
+     * Skip this check for development version.
      */
-    const versionPieces = info.version.split('.') || [];
-    if (versionPieces.length === 3 && info.version !== process.env.VERSION) {
+    if (!process.env.IS_DEV && info.version !== process.env.VERSION) {
       const btn = <Button type="primary" onClick={refreshPage}>Update Now</Button>;
       const message = 'New WebUI Version';
       const description = <div>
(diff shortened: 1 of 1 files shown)


webui/react/src/App.tsx
```diff
@@ -36,8 +36,13 @@ const AppView: React.FC = () => {
...
-    if (info.version !== process.env.VERSION) {
+    /*
+     * Check to make sure the WebUI version matches the platform version.
+     * If the release is a release candidate or a dev tag, skip this notification.
+     * For example, check will run for `0.15.5` but not `0.15.5.dev3`.
+     */
+    const versionChunks = info.version.split('.') || [];
+    if (versionChunks.length === 3 && info.version !== process.env.VERSION) {
       const btn = <Button type="primary" onClick={refreshPage}>Update Now</Button>;
       const message = 'New WebUI Version';
```
hamidzr, line 47: sorta unrelated: could we run into a case where the UI is newer than the cluster version? 1) we try to detect older or newer 2) we generalize the messaging?

webui/react/src/App.tsx
```diff
@@ -36,8 +36,13 @@ const AppView: React.FC = () => {
...
     setupAnalytics(info);
 
-    // Check to make sure the WebUI version matches the platform version.
-    if (info.version !== process.env.VERSION) {
+    /*
+     * Check to make sure the WebUI version matches the platform version.
+     * If the release is a release candidate or a dev tag, skip this notification.
+     * For example, check will run for `0.15.5` but not `0.15.5.dev3`.
+     */
+    const versionChunks = info.version.split('.') || [];
```
hamidzr, line 44: suggestion I'd rely as least as possible on an untyped value and lean toward a direction where if the format changed we minimize the incorrect behavior aka fall back to showing the message for mismatched versions. considering this maybe checking the string for existence of `dev` would suffice? Kinda depends on whether we want to lean toward to showing this message or avoiding it.

hamidzr (approved):

armandmcqueen (approved):

PR #2407, opened 2021-05-20
feat: `det shell` option to show ssh command for use in IDE [DET-5462]

ioga:
## Description
- switch tasks to use `/bin/bash` as a default shell for non-root users.
  we already use `/bin/bash` for root users, and this is required for
  VSCode remote access to work properly.
- use `det shell show_ssh_command` or `--show-ssh-command` option for
  `start` and `open` to print underlying ssh command.
- based on https://github.com/hoanghphan/determined/tree/ide-integration

## Test Plan

- requires #2401 to work on recent local docker.

## Checklist

### DIFF
--- a/harness/determined/cli/shell.py
+++ b/harness/determined/cli/shell.py
@@ -1,4 +1,6 @@
 import getpass
+import os
+import shutil
 import subprocess
 import sys
 import tempfile
@@ -6,6 +8,7 @@ from argparse import ONE_OR_MORE, FileType, Namespace
 from pathlib import Path
 from typing import Any, Dict, List
 
+import appdirs
 from termcolor import colored
 
 from determined.cli import command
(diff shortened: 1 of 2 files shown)


master/pkg/tasks/task.go
```diff
@@ -44,7 +44,7 @@ func workDirArchive(aug *model.AgentUserGroup) container.RunArchive {
 // non-root container features will not work properly.
 func injectUserArchive(aug *model.AgentUserGroup) container.RunArchive {
 	passwdBytes := []byte(
-		fmt.Sprintf("%v:x:%v:%v::%v:/bin/sh\n", aug.User, aug.UID, aug.GID, ContainerWorkDir),
+		fmt.Sprintf("%v:x:%v:%v::%v:/bin/bash\n", aug.User, aug.UID, aug.GID, ContainerWorkDir),
```
rb-determined-ai, line 47: wait, is _this_ why vscode shell was failing?  How on earth did you figure that out?

harness/determined/cli/shell.py
```diff
@@ -54,29 +56,62 @@ def start_shell(args: Namespace) -> None:
...
+
+        key_path = cache_dir / "key"
+        keyfile = (cache_dir / "key").open("w")
+        key_path.chmod(0o600)
+
+        if isinstance(cert_bundle_path, str):
+            retained_cert_bundle_path = cache_dir / "cert_bundle"
+            shutil.copyfile(str(cert_bundle_path), retained_cert_bundle_path)
+            retained_cert_bundle_path.chmod(0o600)
+            cert_bundle_path = retained_cert_bundle_path
```
rb-determined-ai, line 98: this seems like the wrong place for this logic.  Seems like if the permissions of the cert bundle are wrong, we should set them at a higher level

harness/determined/cli/shell.py
```diff
@@ -54,29 +56,62 @@ def start_shell(args: Namespace) -> None:
...
+    print_only: bool = False,
+) -> None:
+    cert_bundle_path = request.get_master_cert_bundle()
+
+    if retain_keys_and_print:
+        cache_dir = Path(appdirs.user_cache_dir("determined")) / "shell" / shell["id"]
+        if not cache_dir.exists():
+            cache_dir.mkdir(parents=True)
+
+        key_path = cache_dir / "key"
```
rb-determined-ai, line 90: can we include the uuid in the key path somewhere?  I think relying on only one shell at a time is going to break in a hurry.

harness/determined/cli/shell.py
```diff
@@ -54,29 +56,62 @@ def start_shell(args: Namespace) -> None:
...
+        if isinstance(cert_bundle_path, str):
+            retained_cert_bundle_path = cache_dir / "cert_bundle"
+            shutil.copyfile(str(cert_bundle_path), retained_cert_bundle_path)
+            retained_cert_bundle_path.chmod(0o600)
+            cert_bundle_path = retained_cert_bundle_path
+    else:
+        keyfile = tempfile.NamedTemporaryFile("w")
+        key_path = keyfile.name
+
+    with keyfile:
```
rb-determined-ai, line 103: It might be nice to just write a contextmanager that either yields an open tempfile or an open key file.

harness/determined/cli/shell.py
```diff
@@ -54,29 +56,62 @@ def start_shell(args: Namespace) -> None:
...
+    check_eq(shell["state"], "STATE_RUNNING", "Shell must be in a running state")
+    _open_shell(args.master, shell, args.ssh_opts, retain_keys_and_print=True, print_only=True)
+
+
+def _open_shell(
+    master: str,
+    shell: Dict[str, Any],
+    additional_opts: List[str],
+    retain_keys_and_print: bool = False,
+    print_only: bool = False,
```
rb-determined-ai, line 81: can we make these two new args required?

master/pkg/tasks/task.go
ioga: yes. vscode plugin logs had `-sh: 4: function: not found` in them.

harness/determined/cli/shell.py
ioga: original cert is already 0o600; replaced with shutil.copy2 to simply preserve these permissions and not set them explicitly.

harness/determined/cli/shell.py
ioga: there's `shell["id"]` already in `cache_dir`

harness/determined/cli/shell.py
ioga: moved key and cert stuff into helpers. IO already acts as a contextmanager so we don't need to wrap it explicitly.

harness/determined/cli/shell.py
rb-determined-ai: ah, totally missed the copy there.

rb-determined-ai (approved):
nice work

PR #2409, opened 2021-05-20
fix: make profiling schema more lenient [DET-5497]

armandmcqueen:
## Description

Users should be able to set `profiling.enabled: true` in the experiment config and fall back to intelligent defaults on starting batch and duration of profiling. The current experiment config doesn't allow only `enabled: true` because it wants to validate that `begin_on_batch <= end_after_batch`. This PR changes the experiment config schema to allow one of both of the fields to be unset when `enabled: true`

This is fine because:
- if both are unset, the default behavior of starting at batch 0 and running for 5 minutes is good
- if start is unset but the end is set, it will default to 0 (previously this was handled by logic in the experiment config, but this PR moves that logic into the schema itself). Then the end doesn't matter because it should always be >= 0 (do we need to add a check to enforce this? it's pretty common sense)
- if the end is unset, but the start is set, the behavior is to profile for 5 minutes which is good behavior.

## Commentary (optional)

I don't understand at all how the 'conditional' extension works (e.g. what is `properties.enabled.const: true` doing), so I'm sure this code isn't right. Let me know the right way to fix it.

## Checklist

### DIFF
--- a/schemas/expconf/v0/profiling.json
+++ b/schemas/expconf/v0/profiling.json
@@ -18,7 +18,7 @@
                 "integer",
                 "null"
             ],
-            "default": null,
+            "default": 0,
             "minimum": 0
         },
         "end_after_batch": {
@@ -31,10 +31,12 @@
         }
     },
     "conditional": {
-        "$comment": "when enabled=true, assert begin <= end",
+        "$comment": "when enabled=true and both begin and end are present, assert begin <= end",
         "when": {
             "required": [
-                "enabled"
+                "enabled",
+                "begin_on_batch",
+                "end_after_batch"
             ],
             "properties": {
                 "enabled": {
--- a/schemas/expconf/v0/profiling.json
+++ b/schemas/expconf/v0/profiling.json
@@ -30,38 +30,9 @@
             "minimum": 0
         }
     },
-    "conditional": {
-        "$comment": "when enabled=true and both begin and end are present, assert begin <= end",
-        "when": {
-            "required": [
-                "enabled",
-                "begin_on_batch",
-                "end_after_batch"
-            ],
-            "properties": {
-                "enabled": {
-                    "const": true
-                }
-            }
-        },
-        "enforce": {
-            "required": [
-                "begin_on_batch",
-                "end_after_batch"
-            ],
-            "propeties": {
-                "begin_on_batch": {
-                    "type": "integer"
-                },
-                "end_after_batch": {
-                    "type": "integer"
-                }
-            },
-            "compareProperties": {
-                "type": "a<=b",
-                "a": "begin_on_batch",
-                "b": "end_after_batch"
-            }
-        }
+    "compareProperties": {
+        "type": "a<=b",
+        "a": "begin_on_batch",
+        "b": "end_after_batch"
     }
 }
--- a/schemas/test_cases/v0/misc.yaml
+++ b/schemas/test_cases/v0/misc.yaml
@@ -130,8 +130,7 @@
 - name: profiling is valid when empty
   matches:
     - http://determined.ai/schemas/expconf/v0/profiling.json
-  case:
-#    enabled: false
+  case: {}
 
 - name: profiling is valid when only enabled is set
   matches:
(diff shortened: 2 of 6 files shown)


schemas/expconf/v0/profiling.json
```diff
@@ -31,10 +31,12 @@
         }
     },
     "conditional": {
-        "$comment": "when enabled=true, assert begin <= end",
+        "$comment": "when enabled=true and both begin and end are present, assert begin <= end",
```
rb-determined-ai, line 34: Ok, so based on the behavior you are describing, I actually don't think the conditional needs to be present at all.

Thinking about it, the comment says `when enabled=true, assert begin <= end`... but actually that's not accurate anymore.  Actually the `compareProperties` subschema could be outside of this conditional and work just fine.  `compareProperties` already ignores situations where one or both properties are not defined, so it doesn't need to be inside a `conditional` at all.

Really, what the old schema did was make both properties required only when `enabled == true`.  (since you asked, the `properties.enabled = {const: true}` thing only matches when the `enabled` property matches the constant value `true`.  The `when` clause also _required_ `enabled`, otherwise an empty object would still match, because that's how json-schema matching works).

But, if the correct behavior is for `begin_on_batch` and `end_after_batch` to *never* be required, you can basically replace the whole `condition` clause with just the `compareProperties` clause, and adjust the defaults accordingly.

Don't forget to edit the existing tests and add more tests in `schemas/tests_cases/v0/misc.yaml`; the profiling test cases are all in a clump at the bottom.

rb-determined-ai: (thinking more about it, I think the `conditional` clause _was_ initially necessary before `end_on_batch` was re-written as `end_after_batch`, and the old `compareProperties` was of `type: "a<b"` which failed on pre-expconf schemas that had zero defaults)

schemas/test_cases/v0/misc.yaml
```diff
@@ -132,7 +132,15 @@
...
+    begin_on_batch: 0
+    end_after_batch: null
+
+- name: profiling is valid when only enabled is set
+  matches:
+    - http://determined.ai/schemas/expconf/v0/profiling.json
+  case:
+    enabled: true
+    begin_on_batch: 0
     end_after_batch: null
```
rb-determined-ai, line 144: You should delete these lines to actually test "when only enabled is set"

rb-determined-ai (commented):
I love when things get simpler!  Nice.

schemas/test_cases/v0/misc.yaml
armandmcqueen: Fixed. 

IMO this is sort of an undocumented DSL and it's not super easy to figure out how to use it correctly. We should add documentation on the DSL syntax.

schemas/test_cases/v0/misc.yaml
```diff
@@ -132,9 +132,15 @@
     - http://determined.ai/schemas/expconf/v0/profiling.json
   case:
     enabled: false
-    begin_on_batch: null
+    begin_on_batch: 0
```
rb-determined-ai, line 135: actually, the "valid when empty" test should also not include `begin_on_batch`, or probably `end_on_batch`.  But if they are both `null` that would be equivalent I guess.  One of the overall rules (limitations?) of expconf is that `null` and `not provided` are not really distinguishable in go.  But I do think that not having them is a bit clearer.

(even though it was I who wrote it using `null` originally...)

rb-determined-ai (approved):

schemas/test_cases/v0/misc.yaml
armandmcqueen: This test should be valid if no fields are set whatsoever. How would I write that test case?

rb-determined-ai (approved):
nice!  Thank you for catching this error before we cut a release with the expconf changes.

schemas/test_cases/v0/misc.yaml
rb-determined-ai: I think to get an empty struct you would want:
```yaml
case: {}
```
Because what you have:
```yaml
case:
```
is actually equivalent to:
```yaml
case: null
```

And that is not a valid `profiling.json` schema at all.

Allowing `null` to match `profiling.json` would be wrong, because the "optionalness" of a struct should be declared by the parent struct that includes it, rather than by the struct itself.  So the experiment config could can say "the profiling field is either null or it matches profiling.json".  You don't need to add that test case though because it should already be tested by the minimal valid experiment test case in test_cases/v0/experiment.yaml

PR #2410, opened 2021-05-20
fix: merge logic for union-type configs [DET-5486]

rb-determined-ai:
Previously, schemas.Merge() would happily merge a shared_fs checkpoint
storage and an s3 checkpoint storage into one, totally invalid,
frankenstorage.

The problem was also present on other union types.

## Test Plan

New tests were added.


stoksc (approved):
lgtm

hamidzr (approved):
👍🏼

PR #2415, opened 2021-05-21
chore: update OS and other language in Terraform modules [DET-4276]

mackrorysd:
## Description

This updates a lingering reference to Ubuntu 16.04 (which has EOL'd and has been replaced by Ubuntu 20.04 elsewhere in our stack). It also eliminates warnings related to interpolation syntax that has been deprecated for quite a few Terraform versions now, and adds a source for the Google provider (idiomatic in more recent versions).

## Test Plan

Put a cluster up and down, confirmed the OS version, and did some basic functionality tests (scaling up and down, running an experiment). Confirmed the warnings are gone. This was done on Terraform 0.13.5 - which is also a little bit behind. We don't document a required Terraform version, but this confirms we're not introducing any unreasonable requirement for the newness of the version.

### DIFF
--- a/harness/determined/deploy/gcp/terraform/main.tf
+++ b/harness/determined/deploy/gcp/terraform/main.tf
@@ -1,4 +1,13 @@
 // Configure GCP provider
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/google"
+      version = "~> 3.44.0"
+    }
+  }
+}
+
 provider "google" {
   credentials = var.keypath != null ? file(var.keypath) : null
   project = var.project_id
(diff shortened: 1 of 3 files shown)


harness/determined/deploy/gcp/terraform/main.tf
```diff
@@ -1,4 +1,13 @@
 // Configure GCP provider
+terraform {
+  required_providers {
+    aws = {
```
ioga, line 4:
```suggestion
    gcp = {
```

ioga (approved):

harness/determined/deploy/gcp/terraform/main.tf
mackrorysd: :facepalm:

PR #2421, opened 2021-05-21
chore: use form GET to reload page

hkang1:
## Description

users are still reporting that the reload does not work with the current refresh page setup.
use the form POST method to ensure that the WebUI can fetch the new react page from the server

## Checklist


hkang1:
decided having the server return a `etag` in the header to distinguish whether the served files are up to date or not is a better more reliable approach than a web approach to solve this.

PR #2423, opened 2021-05-24
fix: prevent spot resource pool contention [DET-5349]

armandmcqueen:
## Description

There is a bug where spot requests and spot instances in AWS do not include the resource pool as part of the tags. This leads to issues where one spot resource pool creates instances, but all of the spot resource pools think they are responsible for managing the instance. This causes spot instances to get shut down because some resource pools think they are idle. This PR ensures that we include the resource pool as one of the tags to prevent this contention. 

This PR also ensures that this change doesn't lead to orphaned spot instances by deleting all spot requests and instances that do not include the `determined-resource-pool` tag when the provisioner starts up.

## Checklist

### DIFF
--- a/master/internal/provisioner/aws_spot_legacy.go
+++ b/master/internal/provisioner/aws_spot_legacy.go
@@ -1,0 +1,204 @@
+package provisioner
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	"github.com/determined-ai/determined/master/pkg/actor"
+)
+
+// There was a bug where spot requests and instances were not being tagged with
+// the resource pool, leading to multiple resource pools trying to manage the
+// same instances. The code now uses a resource pool tag, but that means that
+// instances with the old tag format are ignored. To make sure we aren't leaving
+// orphaned instances after a version upgrade, we identify instances with the old
+// format and clean them up. Because the spot API is eventually consistent, we
+// repeat the cleanup after 5 minutes. This function can be removed once all users
+// are on versions newer than 0.15.5
+func (c *awsCluster) cleanupLegacySpotInstances(ctx *actor.Context) {
+
+	cleanup := func(ctx *actor.Context) {
+		ctx.Log().
+			WithField("codepath", "spotLegacy").
+			Infof("Starting Legacy Spot cleanup operation")
+		// List spot requests with the old format
+		activeSpotReqs, err := c.legacyListActiveSpotInstanceRequests(ctx)
+		if err != nil {
+			ctx.Log().
+				WithField("codepath", "spotLegacy").
+				WithError(err).
+				Debugf("cannot list active spot requests")
+		} else {
+			ctx.Log().
+				WithField("codepath", "spotLegacy").
+				Infof("Terminating %d active spot instance requests", activeSpotReqs.numReqs())
+			// Delete spot requests with old format
+			_, err2 := c.terminateSpotInstanceRequests(ctx, activeSpotReqs.idsAsListOfPointers(), false)
+			if err2 != nil {
+				ctx.Log().
+					WithField("codepath", "spotLegacy").
+					WithError(err).
+					Debugf("unable to terminate legacy spot instance requests")
+			}
+
+			// Delete any instance associated with the active requests
+			instancesToTerminate := newSetOfStrings()
+			for _, req := range activeSpotReqs.iter() {
+				if req.InstanceID != nil {
+					instancesToTerminate.add(*req.InstanceID)
+				}
+			}
+			ctx.Log().
+				WithField("codepath", "spotLegacy").
+				Infof("Terminating %d active spot instances", instancesToTerminate.length())
+			_, err3 := c.terminateInstances(instancesToTerminate.asListOfPointers())
+			if err3 != nil {
+				ctx.Log().
+					WithField("codepath", "spotLegacy").
+					WithError(err).
+					Debugf("unable to terminate instances associated with legacy spot instance requests")
+			}
+		}
+
+		ctx.Log().
+			WithField("codepath", "spotLegacy").
+			Infof("Listing CanceledButInstanceRunning requests")
+		canceledButInstanceRunningSpotReqs, err := c.legacyListCanceledButInstanceRunningSpotRequests(ctx)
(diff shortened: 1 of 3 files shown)


master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,205 @@
...
+// instances with the old tag format are ignored. To make sure we aren't leaving
+// orphaned instances after a version upgrade, we identify instances with the old
+// format and clean them up. Because the spot API is eventually consistent, we
+// repeat the cleanup after 5 minutes. This function can be removed once all users
+// are on versions newer than 0.15.5.
+func (c *awsCluster) cleanupLegacySpotInstances(ctx *actor.Context) {
+	cleanup := func(ctx *actor.Context) {
+		ctx.Log().
+			WithField("codepath", "spotLegacy").
+			Infof("Starting Legacy Spot cleanup operation")
```
armandmcqueen, line 26: I will remove all of the `Infof` log lines after I have manually tested

stoksc (commented):
This pretty much lgtm, i had some nits; they're not mandatory (this is write once code) but otherwise i'd request them. mostly just curious about the retry logic/why? can we return a multierror from the cleanup and only retry then? also i think it's a good principle to not start go routines in a library-like function call but let the user do the go'ing and retrying instead.

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,205 @@
...
+		// List spot requests with the old format
+		activeSpotReqs, err := c.legacyListActiveSpotInstanceRequests(ctx)
+		if err != nil {
+			ctx.Log().
+				WithField("codepath", "spotLegacy").
```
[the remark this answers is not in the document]
armandmcqueen: Fixed

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,205 @@
...
+					aws.String("open"),
+					aws.String("active"),
+				},
+			},
+		},
+	}
+
+	response, err := c.client.DescribeSpotInstanceRequests(input)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
armandmcqueen: Agreed. I fixed the code here as well as the place where the code was originally copied from.

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,205 @@
...
+			WithField("codepath", "spotLegacy").
+			Infof("Starting Legacy Spot cleanup operation")
+		// List spot requests with the old format
+		activeSpotReqs, err := c.legacyListActiveSpotInstanceRequests(ctx)
+		if err != nil {
+			ctx.Log().
+				WithField("codepath", "spotLegacy").
+				WithError(err).
+				Debugf("cannot list active spot requests")
+		} else {
```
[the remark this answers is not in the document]
armandmcqueen: Fixed

armandmcqueen:
I addressed the comments. I also updated the log lines to what I thin they should be in production. There is one `Info` that always runs to say that legacy spot cleanup is happening and then 3 other `Info`s that will only run if there were legacy spot requests/instances that need to be cleaned up. Everything else is now `Debug` or `Error`.

The "try again after 5 minutes" logic is because the AWS Spot APIs are eventually consistent. It can take some time for newly created Spot Requests to show up in the API so if we try to just do cleanup immediately when the master starts, we may not be able to see spot requests that exit. 

It might also take time for the state of spot requests to get updated - e.g. it might say pending, when it has actually created an instance, so we try to cancel the spot requests, but we need to go back later to confirm that there isn't a spot instance running that we need to shut down (I haven't confirmed this specific case happens, but I have run into a bunch of similar problems with the Spot API being eventual consistent). More detail on how spot works can be found here: https://github.com/determined-ai/determined/blob/master/master/internal/provisioner/aws_spot.go#L30

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,216 @@
...
+			Errorf("unable to terminate active legacy spot requests")
+	}
+
+	// Delete spot instances associated with the active requests
+	instancesToTerminate := newSetOfStrings()
+	for _, req := range activeSpotReqs.iter() {
+		if req.InstanceID != nil {
+			instancesToTerminate.add(*req.InstanceID)
+		}
+	}
```
stoksc, line 74: why not `instancesToTerminate := activeSpotReqs.instanceIds()`?

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,216 @@
...
+	return true
+}
+
+func (c *awsCluster) legacyListCanceledButInstanceRunningSpotRequests(
+	ctx *actor.Context,
+) (reqs *setOfSpotRequests, err error) {
+	input := &ec2.DescribeSpotInstanceRequestsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name: aws.String(fmt.Sprintf("tag:%s", c.TagKey)),
```
stoksc, line 140: Not just here but it seems like sharing this tagging logic between spot/non-spot as much as possible could've prevented this bug.

stoksc (approved):
lgtm. i get the eventually consistency bit with spots, it's much clearer now how it's written / with the comment.

master/internal/provisioner/aws_spot_legacy.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		loggerSpotLegacy.
+			WithError(err).
+			Debugf("cannot terminate EC2 instances associated with canceled spot requests")
+	}
+}
+
+func isLegacy(request *ec2.SpotInstanceRequest) bool {
+	for _, tag := range request.Tags {
+		if *tag.Key == "determined-resource-pool" {
```
stoksc, line 127: I feel like this deref is probably always safe (unless maybe the response is malformed) but i'd feel better if we checked it first since this would crash the provisioner.

master/internal/provisioner/aws_spot_legacy.go
armandmcqueen: fixed

master/internal/provisioner/aws_spot_legacy.go
armandmcqueen: fixed

PR #2432, opened 2021-05-25
feat: Add `eventually` extension to schema [DET-5520]

RehanSD:
## Description
`eventually` allows for two-step validation, by only enforcing the specified subschemas during the completeness validation phase. 

One use case is when it is necessary to enforce a `oneOf` on two fields that are `eventuallyRequired`. If the `oneOf` is evaluated during the sanity validation phase, it will always fail, if for example, the user is using cluster default values, but if validation for this subschema is held off until completeness validation, it will validate correctly.

## Checklist

### DIFF
--- a/harness/determined/_swagger/client/api_client.py
+++ b/harness/determined/_swagger/client/api_client.py
@@ -14,18 +14,18 @@ from __future__ import absolute_import
 import datetime
 import json
 import mimetypes
-from multiprocessing.pool import ThreadPool
 import os
 import re
 import tempfile
+from multiprocessing.pool import ThreadPool
 
 # python 2 and python 3 compatibility library
 import six
 from six.moves.urllib.parse import quote
 
-from determined._swagger.client.configuration import Configuration
 import determined._swagger.client.models
 from determined._swagger.client import rest
+from determined._swagger.client.configuration import Configuration
 
 
 class ApiClient(object):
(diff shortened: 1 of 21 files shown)


RehanSD:
Do we need release notes for this change @rb-determined-ai ?

harness/tests/common/test_schemas.py
```diff
@@ -85,7 +85,7 @@ def run_matches(self) -> None:
         if not self.matches:
```
RehanSD, line 85: Should we add completeness validation checks to this @rb-determined-ai ?

harness/determined/common/schemas/extensions.py
```diff
@@ -301,6 +301,24 @@ def eventuallyRequired(
...
+    schema: Dict,
+) -> Iterator[jsonschema.ValidationError]:
+    """
+    eventually allows for two-step validation, by only enforcing the specified subschemas
+    during the completeness validation phase. This is a requirement specific to Determined.
+
+    One use case is when it is necessary to enforce a `oneOf` on two fields that are
+    `eventuallyRequired`. If the `oneOf` is evaluated during the sanity validation phase, it will
+    always fail, if for example, the user is using cluster default values, but if validation
+    for this subschema is held off until completeness validation, it will validate correctly.
```
rb-determined-ai, line 317: Please include an `Example` clause like the docstrings of the other extensions.  This file is basically all the docs we have on these extensions...

harness/determined/_swagger/client/api_client.py
```diff
@@ -14,18 +14,18 @@
 import datetime
```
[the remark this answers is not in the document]
RehanSD: #2440

rb-determined-ai (approved):
Great work!

PR #2434, opened 2021-05-25
fix: omit internal fields in previewed notebook [DET-5523]

shiyuann:
## Description

The previewed notebook configuration goes down the same path populating the full configuration for the actual creation. However, some fields might be null and they are still populated. They should be removed in the previewed configuration.

Also, `resources.slots_per_trial` is an internal field since the experiment configuration and the command configuration shared the same resource configuration go struct. This PR also removes this field since now the experiment configuration goes down a new path that doesn't use the previous go struct.

I noticed there are tests that still use the old experiment configuration go struct but the production code does not use it. So the experiment workflow should just work.

## Test Plan

- CI
- Manually run `det notebook start --preview`

## Checklist

### DIFF
--- a/master/pkg/model/experiment_config.go
+++ b/master/pkg/model/experiment_config.go
@@ -216,13 +216,11 @@ func (d *DeviceConfig) UnmarshalJSON(data []byte) error {
 
 // ResourcesConfig configures resource usage for an experiment, command, notebook, or tensorboard.
 type ResourcesConfig struct {
-	// Slots is used by commands while trials use SlotsPerTrial.
 	Slots int `json:"slots,omitempty"`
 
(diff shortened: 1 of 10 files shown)


master/pkg/model/experiment_config.go
```diff
@@ -272,10 +269,7 @@ func ValidatePrioritySetting(priority *int) []error {
 // Validate implements the check.Validatable interface.
 func (r ResourcesConfig) Validate() []error {
 	errs := []error{
-		check.GreaterThan(r.SlotsPerTrial, 0, "slots_per_trial must be > 0"),
```
rb-determined-ai, line 275: Maybe check that `r.Slots` is greather than 0?

rb-determined-ai (approved):
Looks good.

PR #2435, opened 2021-05-25
fix: resource pool not saved in the DB [DET-5485]

shiyuann:
## Description

This bug is introduced in [here](https://github.com/determined-ai/determined/commit/afdb74955b5ec498095767b1d81d81756b6da510#diff-d92b0050d5549a1efe86c524037f8a2709d6ebf8ce2c63a777f63744bc8a0b00R103). When we filled in the resource pool value, we did that on a copied go struct. This is fixed by changing it to use a pointer.

## Test Plan

Manually create one experiment and check the WebUI

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -108,11 +108,8 @@ type (
 func newExperiment(master *Master, expModel *model.Experiment, taskSpec *tasks.TaskSpec) (
 	*experiment, error,
 ) {
-	conf := expModel.Config
+	conf := &expModel.Config
 
-	// Validate the ResourcePool setting.  The reason to do it now and not in postExperiment like
-	// all the other validations is that the resource pool should be revalidated every time the
-	// master restarts.
 	resources := conf.Resources()
 	poolName := resources.ResourcePool()
 	if err := sproto.ValidateRP(master.system, poolName); err != nil {


docs/release-notes/2435-fix-resource-pool-not-saved.txt
```diff
@@ -0,0 +1,5 @@
+:orphan:
```
rb-determined-ai, line 1: I don't know if the release note is necessary; it seems to me like the bug was introduced by expconf, since the old code path updated `expModel.Config.Resources` directly.

master/internal/experiment.go
```diff
@@ -108,7 +108,7 @@ type (
 func newExperiment(master *Master, expModel *model.Experiment, taskSpec *tasks.TaskSpec) (
 	*experiment, error,
 ) {
-	conf := expModel.Config
+	conf := &expModel.Config
 
 	// Validate the ResourcePool setting.  The reason to do it now and not in postExperiment like
 	// all the other validations is that the resource pool should be revalidated every time the
```
rb-determined-ai, line 114: Can you delete this comment?  It isn't my code that was buggy; it was my understanding of what the old code did (as this comment explains).

Side-note, now that I understand correctly what the behavior is, there is a large cleanup that can be done.  For example, all the other default filling happens in a `parseCreateExperiment`, and this should too.

(but I'm fine landing this fix now, since it is correct as-written, and I'll do the followup fix in the next few days)

armandmcqueen (approved):

docs/release-notes/2435-fix-resource-pool-not-saved.txt
shiyuann: I think this is necessary since it's a bug introduced before the last release. The WebUI doesn't show the resource pool if it is filled automatically.

rb-determined-ai: I just checked.  This is not a bug on 0.15.5.

See https://github.com/determined-ai/determined/commit/afdb74955b5ec498095767b1d81d81756b6da510#diff-d92b0050d5549a1efe86c524037f8a2709d6ebf8ce2c63a777f63744bc8a0b00R117

PR #2442, opened 2021-05-25
docs: notify users of coscheduler behavior [DET-5150]

eecsliu:
## Description
Updates the docs to reflect that the coscheduler has long waiting time if attempting to schedule jobs requiring the full cluster.

## Checklist

### DIFF
--- a/docs/topic-guides/system-concepts/scheduling.txt
+++ b/docs/topic-guides/system-concepts/scheduling.txt
@@ -155,16 +155,26 @@ The coscheduling plugin is in beta and is therefore not enabled by
 default. To enable it, edit ``values.yaml`` in the Determined Helm chart
 to set the ``defaultScheduler`` field to ``coscheduler``.
 
-Importantly, the coscheduling plugin does not work with Kubernetes'
-cluster autoscaling feature: static node pools must be used to achieve
-gang scheduling. Also, while the plugin allocates resources to jobs
-based on their priority, it does not support preemption. For example, if
-the cluster is full of low priority jobs and a new high priority job is
-submitted, the high priority job will not be scheduled until one of the
-low priority jobs finishes. Lastly, Determined's capability to
-automatically set pod labels is restricted to GPU experiments;
-Determined does not currently set labels for CPU experiments or user
-commands.
+There are several limitations to the coscheduling plugin to be aware of:
+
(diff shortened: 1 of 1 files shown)


docs/topic-guides/system-concepts/scheduling.txt
```diff
@@ -155,16 +155,26 @@ The coscheduling plugin is in beta and is therefore not enabled by
...
-automatically set pod labels is restricted to GPU experiments;
-Determined does not currently set labels for CPU experiments or user
-commands.
+There are several limitations to the coscheduling plugin to be aware of:
+
+#. The coscheduling plugin does not work with Kubernetes' cluster
+   autoscaling feature. Static node pools must be used to achieve gang
+   scheduling
+
+#. The plugin does not support preemption. For example, if the cluster
```
mackrorysd, line 164: Side note: I feel like we shouldn't even advertise priority support when we don't support preemption. What are the odds that priority actually has any impact on scheduling decisions? Pretty tiny. Don't have to address this now since it's not a "regression" per se. Just thinking out loud.

docs/topic-guides/system-concepts/scheduling.txt
```diff
@@ -155,16 +155,26 @@ The coscheduling plugin is in beta and is therefore not enabled by
...
+   is full of low priority jobs and a new high priority job is
+   submitted, the high priority job will not be scheduled until one of
+   the low priority jobs finishes.
+
+#. Determined is able to automatically set pod labels, but only for GPU
```
mackrorysd, line 169: I prefer the way this was worded before: IMO it should be a stronger statement that we will set pod labels required to schedule GPUs, but nothing else. The fact that we are *able* to automatically set pod labels is of no other consequence to the user.

mackrorysd (changes requested):
Looks good otherwise.

docs/topic-guides/system-concepts/scheduling.txt
eecsliu: Priority for the coscheduler? Your musings make sense to me, but one thing is that the default scheduler is supposed to take priority into account anyways (albeit less deliberately than our priority scheduler) and even support preemption too. I feel like it might be necessary to clarify this as it differs from default k8s scheduler behavior.

mackrorysd (approved):

PR #2443, opened 2021-05-25
chore: make tensorboard and checkpoint gc expconf reads backwards compatible

stoksc:
## Description
This change has two parts: first, it re-adds the atrocity `ExperimentByIDWithUnparsableFieldsDummied` that attempts to parse and expconf from the database and falls back to parsing an incomplete version of it. It also changes checkpoint GC to never try to save the maybe invalid expconf.

## Test Plan
- [ ] run tbs and delete for old, old experiments

## Commentary (optional)
Yep.

## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -86,7 +86,7 @@ func (a *apiServer) DeleteExperiment(
 	_ context.Context, req *apiv1.DeleteExperimentRequest,
 ) (*apiv1.DeleteExperimentResponse, error) {
 	expID := int(req.ExperimentId)
-	exp, err := a.m.db.ExperimentByID(expID)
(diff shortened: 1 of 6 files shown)


master/internal/api_experiment.go
```diff
@@ -113,23 +113,21 @@ func (a *apiServer) DeleteExperiment(
...
 	}
 
-	storage := exp.Config.CheckpointStorage()
-	storage.SetSaveExperimentBest(0)
-	storage.SetSaveTrialBest(0)
-	storage.SetSaveTrialLatest(0)
-	exp.Config.SetCheckpointStorage(storage)
-
+	// TODO(now): This is bad.
 	if sErr := a.m.db.SaveExperimentConfig(exp); sErr != nil {
```
rb-determined-ai, line 117: Why is this not removed?  I don't even see any updates being made to it?

PR #2444, opened 2021-05-26
fix: introduce HistoricConfig for tensorboard and checkpooint gc [DET-5533]

rb-determined-ai:
## Description

Based on #2443, but I hit some git error trying to push to the original PR.  Running tests here.

### DIFF
--- a/master/internal/command/tensorboard_manager.go
+++ b/master/internal/command/tensorboard_manager.go
@@ -50,8 +50,9 @@ type TensorboardRequest struct {
 }
 
 type tensorboardConfig struct {
-	model.Experiment
-	TrialIDs []int
+	Config       expconf.LegacyConfig
+	ExperimentID int
+	TrialIDs     []int
 }
 
 type tensorboardManager struct {
@@ -253,14 +254,14 @@ func (t *tensorboardManager) newTensorBoard(
 
 		if len(exp.TrialIDs) == 0 {
 			expDir := fmt.Sprintf("%s/%s/tensorboard/experiment/%d/",
-				logBasePath, params.TaskSpec.ClusterID, exp.ID)
+				logBasePath, params.TaskSpec.ClusterID, exp.ExperimentID)
 			logDirs = append(logDirs, expDir)
 			continue
 		}
 
 		for _, id := range exp.TrialIDs {
 			trialDir := fmt.Sprintf("trial_%d:%s/%s/tensorboard/experiment/%d/trial/%d/",
-				id, logBasePath, params.TaskSpec.ClusterID, exp.ID, id)
+				id, logBasePath, params.TaskSpec.ClusterID, exp.ExperimentID, id)
 
 			logDirs = append(logDirs, trialDir)
 		}
@@ -268,7 +269,7 @@ func (t *tensorboardManager) newTensorBoard(
 
 	// Get the most recent experiment config as raw json and add it to the container. This
 	// is used to determine if the experiment is backed by S3.
-	mostRecentExpID := exps[len(exps)-1].ID
+	mostRecentExpID := exps[len(exps)-1].ExperimentID
 	confBytes, err := t.db.ExperimentConfigRaw(mostRecentExpID)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error loading raw experiment config: %d", mostRecentExpID)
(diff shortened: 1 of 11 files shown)


master/internal/command/tensorboard_manager.go
```diff
@@ -50,8 +50,9 @@ type TensorboardRequest struct {
 }
 
 type tensorboardConfig struct {
-	model.Experiment
-	TrialIDs []int
+	Config       expconf.HistoricalConfig
+	ExperimentID int
```
stoksc, line 54: 👍 thank you..

master/pkg/schemas/expconf/historical.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return out, err
+	}
+	out.envvars = ev
+
+	return out, nil
+}
+
+// Scan implements the db.Scanner interface.
+func (h *HistoricalConfig) Scan(src interface{}) error {
```
stoksc, line 164: not really much of a difference, but i prefer querying bytes and putting this logic in the db rather than hiding db specific logic (impl scan) in expconf.

master/pkg/schemas/expconf/historical.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	bm = schemas.WithDefaults(bm).(BindMountsConfig)
+	if err = schemas.IsComplete(bm); err != nil {
+		return bm, errors.Wrap(err, "historical bind mounts is incomplete")
+	}
+	return bm, nil
+}
+
+func getEnvironmentVariables(raw map[string]interface{}) (EnvironmentVariablesMap, error) {
+	ev := EnvironmentVariablesMap{}
```
stoksc, line 93: how far are we from schemas being magical enough to just unmarshal the bytes into `HistoricalConfig` and then 
checking sane bytes/iscomplete/defaulting as usually after deser? i guess it just would require adding a historical config schema?

stoksc (commented):
this checkpoint gc refactor is just stellar. i will do a more thorough review tomorrow morning, just a first pass.

master/pkg/schemas/expconf/historical.go
rb-determined-ai: Sure, I can do that.  The anonymous struct to invoke the Scan was a little weird anyway

master/pkg/schemas/expconf/historical.go
rb-determined-ai: I don't think it's a good idea to try to write a json schema HistoricalConfig.  I think in the future, we can check the version field and unmarshal that version, but the v0 handling is always going to be tricky.

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -81,6 +81,14 @@ func (e *ExperimentConfigV0) Scan(src interface{}) error {
 	return nil
 }
 
+// AsHistorical converts a current ExperimentConfig to a (limited capacity) HistoricalConfig.
+func (e ExperimentConfig) AsHistorical() HistoricalConfig {
+	return HistoricalConfig{
+		checkpointStorage: e.CheckpointStorage(),
+		bindMounts:        e.BindMounts(),
```
stoksc, line 88: no envvars?

rb-determined-ai: Just found/fixed that as part of the `Historical` -> `Legacy` rename.

master/pkg/schemas/expconf/historical.go
rb-determined-ai: this is done now.

master/pkg/schemas/expconf/legacy.go
```diff
@@ -0,0 +1,175 @@
...
+	// Validate fully before passing anything out.
+	if err = schemas.IsComplete(cs); err != nil {
+		return cs, errors.Wrap(err, "legacy checkpoint storage is incomplete")
+	}
+
+	return cs, nil
+}
+
+func getBindMounts(raw map[string]interface{}) (BindMountsConfig, error) {
+	bm := BindMountsConfig{}
```
stoksc, line 71:
```suggestion
	var bm BindMountsConfig
```
only because it's really confusing to read a slice init'd that way

stoksc (approved):
lgtm, killer tests

master/pkg/schemas/expconf/legacy.go
rb-determined-ai: Sure.  Though I would like to point out that these initialization techniques are materially different.

stoksc: > Though I would like to point out that these initialization techniques are materially different.

Wait, how?

stoksc: also i dont actually care, you can leave it

rb-determined-ai: > Wait, how?

The `var` syntax creates a `nil` map, the `{}` syntax creates a non-nil map.  Which is typically what I expect `{}` to do.

PR #2445, opened 2021-05-26
fix: allow removed searchers in configs only [DET-5526]

rb-determined-ai:
## Description

Restore the old searcher config objects and schemas, but don't allow them into new experiments.


harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -2228,6 +2228,214 @@
...
+            "minimum": 1,
+            "default": 5
+        },
+        "max_length": {
+            "type": [
+                "object",
+                "null"
+            ],
+            "default": null,
+            "optionalRef": "http://determined.ai/schemas/expconf/v0/check-positive-length.json"
```
mackrorysd, line 2399: What is the actual difference in the constraints between max_rungs and max_length?

rb-determined-ai: I think `max_rungs` affects the shape of the tournament trees that are created and `max_length` is the total training length of the longest-running trial

mackrorysd: Oh nevermind - I looked at the check-positive-length.json, but missed ITS reference to length.json.

mackrorysd (approved):

PR #2451, opened 2021-05-26
fix: prevent concurrent updates to a single expconf object [DET-5543]

rb-determined-ai:
Use a deep copy when passing an experiment config from one actor to
another.


stoksc (approved):
lgtm, though this is a lot of copying to introduce. absolutely necessary with actors though.

shiyuann (approved):
A bit aggressive but looks good!

PR #2452, opened 2021-05-26
revert: revert added example model (tf classification)

azhou-determined:
revert changes made to add an example model (tensorflow multiclass text classification)


eecsliu (approved):
🚀

PR #2453, opened 2021-05-26
fix: synchronize pods actor startup in k8s resource manager [DET-5536]

shiyuann:
## Description

Previously, when the k8s pods actor starts up, it sends a message to the k8s resource manager actor, which then creates a mock agent. Before the mock agent is created, no allocation requests should be handled in the k8s resource manager actor. However, there is no synchronization to ensure the k8s resource manager actor receives `SetPods` message first.

This PR moves creating the k8s pods actor into the `PreStart` of the k8s resource manager actor so the order is enforced.

## Test Plan

CI

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/kubernetes_resource_manager.go
+++ b/master/internal/resourcemanagers/kubernetes_resource_manager.go
@@ -2,13 +2,15 @@ package resourcemanagers
 
 import (
 	"github.com/google/uuid"
+	"github.com/labstack/echo/v4"
 
+	"github.com/determined-ai/determined/master/internal/kubernetes"
 	"github.com/determined-ai/determined/master/internal/sproto"
 	"github.com/determined-ai/determined/master/pkg/actor"
 	"github.com/determined-ai/determined/master/pkg/actor/actors"
-	"github.com/determined-ai/determined/master/pkg/check"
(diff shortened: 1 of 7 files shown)


master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -55,10 +68,19 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
 	case actor.PreStart:
 		actors.NotifyAfter(ctx, actionCoolDown, schedulerTick{})
 
-	case sproto.SetPods:
-		check.Panic(check.True(k.agent == nil, "should only set pods once"))
+		podsActor := kubernetes.Initialize(
```
eecsliu, line 71: non-blocking clarification: is this pods actor or a resource manager being created?

eecsliu (approved):
👩‍🍳😗👌

master/internal/resourcemanagers/kubernetes_resource_manager.go
shiyuann: This is to create the pods actor when "Prestarting" the k8s resource manager.

PR #2455, opened 2021-05-27
fix: update Buf image and CLI usage

dzhu:
## Description

The new version of Buf has moved around how its CLI is organized, so
update to the now-recommended commands.

Also update the image with `make gen-buf-image`, since the new version
complains otherwise.

## Test Plan

- [x] `make -C proto check`


RehanSD (approved):
looks good to me!

stoksc (approved):

PR #2458, opened 2021-05-27
chore: dont close trial on socket close

stoksc:
## Description
This change updates the master code to _not_ forcibly close trials on socket disconnects. In the event of a socket from the harness disconnecting, the harness code should exit. Forcibly killing hung code should be left to the use of `det t kill` or similar, since hung code could also not be hung, but just awaiting some important cleanup.

## Test Plan
- [x] run stuff and make sure trials end up in ok states still

## Checklist

### DIFF
--- a/master/internal/trial.go
+++ b/master/internal/trial.go
@@ -439,9 +439,8 @@ func (t *trial) runningReceive(ctx *actor.Context) error {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


stoksc:
fwiw, the bug wasn't even an abnormal closure so this is just a noop.. it just adds a log for it.

PR #2460, opened 2021-05-27
fix: use user preferences when no search params are present [DET-5537]

hkang1:
## Description

Problem: when viewing the Experiment List page via the nav side bar, the route is set to `/experiments` without any search params. For this case, it currently grabs all of the experiments (essentially no filters). Instead, we want to use user preferences if they are available.

## Test Plan

- [ ] apply some experiment list filters such as `labels`, `states` and or `users`, then click on `Dashboard` then click back to `Experiments`. The previously selected filters should be exactly the same.

## Checklist


hamidzr (approved):
🚀

hamidzr:
thx for fixing this

PR #2462, opened 2021-05-27
chore: move run increment to allocation not termination [DET-5559, DET-5450]

stoksc:
## Description
This change just moves "starting a new run" to when resource are allocated from when we were terminated. It really should've happened in trial reset previously, but this change makes more semantic sense to me: a run isn't started until it is.. started?

## Test Plan
- [ ] run tests, also make sure resets dont print error messages

## Checklist

### DIFF
--- a/master/internal/trial.go
+++ b/master/internal/trial.go
@@ -378,6 +378,13 @@ func (t *trial) Receive(ctx *actor.Context) error {
 				name = fmt.Sprintf("Trial (Experiment %d)", t.experiment.ID)
 			}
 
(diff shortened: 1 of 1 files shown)


shiyuann:
Shouldn't it increment run id even before sending a resource request?

stoksc:
> Shouldn't it increment run id even before sending a resource request?

it could, not sure it makes a massive difference here. I thought it we didn't start and thus wouldn't complete anything there was no reason to count it as a run, but i suppose it makes sense. I saw both options and just picked one.

shiyuann (approved):

PR #2464, opened 2021-05-27
chore: remove trial old messages and consolidate others

stoksc:
## Checklist

### DIFF
--- a/master/internal/trial.go
+++ b/master/internal/trial.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"encoding/json"
 	"fmt"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
 	"sort"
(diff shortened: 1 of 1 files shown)


master/internal/trial.go
```diff
@@ -439,14 +437,13 @@ func (t *trial) runningReceive(ctx *actor.Context) error {
 			return err
 		}
 
-	case actor.ChildFailed:
```
shiyuann, line 442: What is this one to fix again?

stoksc: Well this one doesn't usually happen anyone because we started swallowing abnormal closures in the websocket actor. Now it would need to be a websocket actor panic. I prob will just revert this though and clean it up later.

shiyuann (approved):

PR #2467, opened 2021-05-27
chore: add testing for `eventually` schema [DET-5560]

RehanSD:
## Description
This PR adds testing for the `eventually` schema extension, adds completeness validation testing, and splits the `errors` schema tests to separate into `sanity` and `completeness` errors.

## Checklist


harness/determined/common/schemas/expconf/_validate.py
```diff
@@ -14,7 +14,9 @@ def make_validator(url: Optional[str] = None, complete: Optional[bool] = False)
         url = "http://determined.ai/schemas/expconf/v1/experiment.json"
 
     global _validators
-    if url in _validators:
+    # Need a new validator to be made for completeness validation. If we
+    # used the cached sanity validator, we wouldn't check `eventuallyRequire`, etc.
+    if url in _validators and not complete:
```
RehanSD, line 19: @rb-determined-ai instead of removing caching altogether for `completeness` validators, should we perhaps cache it as `url+"complete"`?

harness/tests/common/test_schemas.py
```diff
@@ -100,31 +105,46 @@ def run_completeness(self) -> None:
...
 
-    def run_errors(self) -> None:
-        if not self.errors:
+    def run_completeness_errors(self) -> None:
+        if not self.completeness_errors:
             return
-        for url, expected in self.errors.items():
+        self.run_errors(self.completeness_errors, test_type="completeness")
+
+    def run_errors(self, error_cases: Any, test_type: Optional[str] = "sanity") -> None:
```
rb-determined-ai, line 124: don't use default arguments for "internal" helper functions like this.  Make it required so it is explicit what the caller is trying to specify, rather than have half the callers depend on defaults to communicate their intent.

rb-determined-ai: possibly also call this `_run_errors` to distinguish it from the "higher" functions that invoke this one.

harness/tests/common/test_schemas.py
```diff
@@ -100,31 +105,46 @@ def run_completeness(self) -> None:
...
-                    msg = f"while testing '{self.name}', expected to match the pattern\n"
-                    msg += f"    {exp}\n"
+                    msg = f"while testing '{self.name}' for {test_type}, expected to match the"
+                    msg += f" pattern\n    {exp}\n"
                     msg += "but it was not found in any of\n    "
                     msg += "\n    ".join(errors)
                     raise ValueError(msg)
 
     def run_defaulted(self) -> None:
         if not self.defaulted:
```
rb-determined-ai, line 144: can you mimic the all-or-nothing assertion for `merge_as/merge_src/merged` here for `default_as/defaulted`?  Just as a little defense against people accidentally writing tests that have a `default_as` without a `defaulted`, which would silently not really run at all.

harness/determined/common/schemas/expconf/_validate.py
rb-determined-ai: just use two separate caches, `_sanity_validators` and `_completeness_validators`, or let `_validators` be a map of bools to maps of urls to validators, like `_validators[optional][url]` or something like that.

rb-determined-ai (commented):
python side looks good.  Overall, I think this is a big improvement to the testing infrastructure

master/pkg/schemas/expconf/schema_test.go
```diff
@@ -81,24 +84,41 @@ func (tc SchemaTestCase) CheckCompleteAs(t *testing.T) {
...
+}
+
+func (tc SchemaTestCase) CheckCompletenessErrors(t *testing.T) {
+	if tc.CompletenessErrors == nil {
+		return
+	}
+	checkErrors(tc, t, tc.CompletenessErrors, "completeness")
+}
+
+func checkErrors(tc SchemaTestCase, t *testing.T, errors *map[string][]string, testType string) {
```
rb-determined-ai, line 101: This seems like it should be a method on `tc SchemaTestCase` like the others.  (but lower-casing the `c` of `checkErrors` is good, because that distinguishes it from the other "higher-level" functions, the way that the lead `_` of `_run_errors` does in python)

rb-determined-ai (approved):
Looks good.

PR #2469, opened 2021-05-28
chore: improve performance of tfevent file filtering

rb-determined-ai:
I observed that as tfevent files build up, just selecting which files to
examine can actually be a bit of a bottleneck.  The current strategy
seems to have a minimum cost of about 1ms per 100 files (calling
stat() on 1500 event files takes my system 0.012s).  The previous
implementation kept a complex index of how large each file had been
before uploading it, which quadrupled the time for choosing files to
around 0.045 seconds.  The new strategy, which just chooses all modified
files since the last sync time, has basically no overhead over the cost
of stat'ing each file.

The only strategy I can think of that would be faster would be to hook
into inotify to have the operating system watch the list of files for
you.

### DIFF
--- a/harness/determined/tensorboard/base.py
+++ b/harness/determined/tensorboard/base.py
@@ -1,7 +1,8 @@
 import abc
 import logging
 import pathlib
-from typing import Dict, List
+import time
+from typing import List
 
 
 class TensorboardManager(metaclass=abc.ABCMeta):
@@ -22,11 +23,17 @@ class TensorboardManager(metaclass=abc.ABCMeta):
     ) -> None:
         self.base_path = base_path
         self.sync_path = sync_path
-        self._synced_event_sizes: Dict[pathlib.Path, int] = {}
+        self.last_sync = 0.0
 
-    def list_tfevents(self) -> List[pathlib.Path]:
+    def list_tfevents(self, since: float) -> List[pathlib.Path]:
         """
-        list_tfevents returns tfevent file names located in the base_path directory.
+        list_tfevents returns tfevent file names located in the base_path directory that have been
+        modified since a certain time.
+
+        If many tfevent files have been created, the syscall to stat on each of them can be quite
+        expensive, taking on the order of 1ms for every 100 files. Each file gets stat'd every call
+        to this function, which can be a bottleneck late in training or in local training when many
+        tfevent files exist but few or none need to be resynced.
         """
 
         if not self.base_path.exists():
@@ -36,20 +43,12 @@ class TensorboardManager(metaclass=abc.ABCMeta):
             )
             return []
 
-        return [f.resolve() for f in self.base_path.glob("**/*tfevents*")]
+        return [f for f in self.base_path.glob("**/*tfevents*") if f.stat().st_mtime > since]
 
     def to_sync(self) -> List[pathlib.Path]:
-        """
-        to_sync returns tfevent files that have not been exported to the persistent
-        storage backend.
-        """
-
-        sync_paths = []
-        for path in self.list_tfevents():
-            if path not in self._synced_event_sizes:
-                sync_paths.append(path)
-            elif path.stat().st_size > self._synced_event_sizes[path]:
-                sync_paths.append(path)
+        sync_start = time.time()
+        sync_paths = self.list_tfevents(self.last_sync)
+        self.last_sync = sync_start
 
         return sync_paths
 
--- a/harness/determined/tensorboard/gcs.py
+++ b/harness/determined/tensorboard/gcs.py
@@ -31,7 +31,6 @@ class GCSTensorboardManager(base.TensorboardManager):
             logging.debug(f"Uploading to GCS: {blob_name}")
 
             blob.upload_from_filename(str(path))
-            self._synced_event_sizes[path] = path.stat().st_size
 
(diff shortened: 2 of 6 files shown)


harness/determined/tensorboard/base.py
```diff
@@ -36,20 +44,16 @@ def list_tfevents(self) -> List[pathlib.Path]:
...
-        sync_paths = []
-        for path in self.list_tfevents():
-            if path not in self._synced_event_sizes:
-                sync_paths.append(path)
-            elif path.stat().st_size > self._synced_event_sizes[path]:
-                sync_paths.append(path)
+    def to_sync(self) -> List[pathlib.Path]:
+        sync_start = time.time()
+        sync_paths = self.list_tfevents(self.last_sync)
+        self.last_sync = start
```
dzhu, line 56: `sync_start`, I presume?

harness/determined/tensorboard/base.py
```diff
@@ -36,20 +44,16 @@ def list_tfevents(self) -> List[pathlib.Path]:
...
+        def stat_each(paths):
+            for f in paths:
+                yield (f, f.stat())
 
-    def to_sync(self) -> List[pathlib.Path]:
-        """
-        to_sync returns tfevent files that have not been exported to the persistent
-        storage backend.
-        """
+        return [f for f, s in stat_each(self.base_path.glob("**/*tfevents*")) if s.st_mtime > since]
```
dzhu, line 51: Haven't tested it, but I believe this can replace the several lines above.

```python
        return [f for f in self.base_path.glob("**/*tfevents*") if f.stat().st_mtime > since]
```

harness/determined/tensorboard/base.py
```diff
@@ -23,10 +24,17 @@ def __init__(
         self.base_path = base_path
         self.sync_path = sync_path
         self._synced_event_sizes: Dict[pathlib.Path, int] = {}
```
dzhu, line 26: I think all references to this field can be removed now.

harness/determined/tensorboard/base.py
rb-determined-ai: oh right!  Originally I thought I had to return the stat output too, which is why I introduced the extra layer of iterator (didn't want to stat twice)

harness/determined/tensorboard/base.py
rb-determined-ai: yes

rb-determined-ai:
@dzhu does the approach seem sane to you, of using time rather than size?  I can't really think of how it can go wrong under reasonable conditions.

dzhu:
Yeah, I think it makes sense. Only thing I'd want to check: nothing should be writing to those files concurrently with this code checking them, right? I thought a bit and I don't think that would cause any issues due to weird timing, but it'd be nicest not to worry at all.

harness/determined/tensorboard/base.py
```diff
@@ -22,11 +23,17 @@ def __init__(
     ) -> None:
         self.base_path = base_path
         self.sync_path = sync_path
-        self._synced_event_sizes: Dict[pathlib.Path, int] = {}
+        self.last_sync = float(0)
```
dzhu, line 26: I think this is a bit more idiomatic:
```suggestion
        self.last_sync = 0.0
```

harness/determined/tensorboard/base.py
```diff
@@ -22,11 +23,17 @@ def __init__(
...
 
-    def list_tfevents(self) -> List[pathlib.Path]:
+    def list_tfevents(self, since: float) -> List[pathlib.Path]:
         """
-        list_tfevents returns tfevent file names located in the base_path directory.
+        list_tfevents returns tfevent file names located in the base_path directory, which have been
```
dzhu, line 30:
```suggestion
        list_tfevents returns tfevent file names located in the base_path directory that have been
```

rb-determined-ai:
No, nothing should be writing to them concurrently.  But just in case such a thing ever happens, that's why I set the last_sync time based on the time *before* examining the filesystem.

dzhu (approved):
Alternatively, is it possible at all to determine from here whether a file is done being written to? Then we could move old files, delete them, or remember their names, and so not have to stat them at all.

rb-determined-ai:
> Alternatively, is it possible at all to determine from here whether a file is done being written to? Then we could move old files, delete them, or remember their names, and so not have to stat them at all.

Not that I know of.  I think I might trust an inotify-based implementation of that over trying to understand tensorboard metric writers anyway, though.

rb-determined-ai:
You know who should take a look at this? @eecsliu, does this change look safe to you?

harness/determined/tensorboard/base.py
rb-determined-ai: :facepalm:

eecsliu:
I _really_ like this strategy. The problem I see here is that I think the metric writers are asynchronous, and I've seen it cause missing metrics when it wasn't able to flush to disk in time. That should be an easy fix of forcing the metric writer to flush here.

rb-determined-ai:
@eecsliu and I were not sure what the behavior was for modified files with open write descriptors, but we experimented with it and it appears that any time data is flushed to the file the modified time is updated automatically by the OS.  So we think this strategy is safe.

PR #2472, opened 2021-05-29
fix: timing metric chart x-axis tick off [DET-5501]

apizzini:
## Description

In some cases the result computed through the factor is too large.
Fixing this by setting a max value of 70.

So now ticks space will always be between 35 and 70.

## Checklist


hkang1 (approved):
Took me a while to figure this portion out but LGTM!
So it looks like it will try find the largest divisor possible given the size of the chart and use that to set tick space, but if it's greater than 70px, then it caps it there.

Before:
Screen Shot 2021-06-01 at 9 44 59 AM

After:
Screen Shot 2021-06-01 at 9 45 08 AM

apizzini:
> So it looks like it will try find the largest divisor possible given the size of the chart and use that to set tick space, but if it's greater than 70px, then it caps it there.

@hkang1 correct.

PR #2473, opened 2021-05-29
fix: send all batches from profiler [DET-5566]

stoksc:
## Description
This change ensures that all batches are sent on shutdown.

## Test Plan
- [x] test on a real cluster

## Commentary (optional)
This is the PR I'm going to add tests over.

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -15,7 +15,7 @@ from determined.common.api import TrialProfilerMetricsBatch
 
 SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
 TIMING_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_TIMING"
-
+MAX_COLLECTION_SECONDS = 300
 LOG_NAMESPACE = "determined-profiler"
 
 try:
@@ -100,17 +100,20 @@ class ProfilerAgent:
 
         # If the ProfilingAgent is disabled, don't waste resources by creating useless threads
         if self.is_enabled:
-            # Set up timer thread to stop collecting after 5 minutes
-            self.max_collection_seconds = 300
-            self.shutdown_timer = PreemptibleTimer(
-                self.max_collection_seconds, self._end_collection
-            )
-
-            # Set up the thread responsible for making API calls
+            # Set up sender thread with the number of producers to expect shutdowns from.
             self.send_queue = (
                 queue.Queue()
             )  # type: """queue.Queue[Union[List[TrialProfilerMetricsBatch], ShutdownMessage]]"""
-            self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+
+            # Set up timer thread to stop collecting after 5 minutes
+            self.shutdown_timer = PreemptibleTimer(MAX_COLLECTION_SECONDS, self._end_collection)
+
+            # Set up the thread responsible for making API calls
+            if self.sysmetrics_is_enabled or self.timings_is_enabled:
+                num_producers = int(self.sysmetrics_is_enabled) + int(self.timings_is_enabled)
+                self.sender_thread = ProfilerSenderThread(
+                    self.send_queue, self.master_url, num_producers
+                )
 
             if self.sysmetrics_is_enabled:
                 self.sys_metric_collector_thread = SysMetricCollectorThread(
@@ -281,7 +284,6 @@ class ProfilerAgent:
                 self.timings_batcher_thread.send_shutdown_signal()
                 self.timings_batcher_thread.join()
 
-            self.sender_thread.send_shutdown_signal()
             self.sender_thread.join()
 
             self.has_finished = True
@@ -389,7 +391,6 @@ class SysMetricCollectorThread(threading.Thread):
     MEASUREMENT_INTERVAL = 0.1
 
     def __init__(self, trial_id: str, agent_id: str, send_queue: queue.Queue):
-
         self.current_batch_idx = 0
         self.send_queue = send_queue
         self.control_queue: "queue.Queue[Union['StartMessage', 'ShutdownMessage']]" = queue.Queue()
@@ -420,6 +421,7 @@ class SysMetricCollectorThread(threading.Thread):
         # Do nothing while we wait for a StartMessage
         msg = self.control_queue.get()
         if isinstance(msg, ShutdownMessage):
+            self.send_queue.put(ShutdownMessage())
             return
 
         # Do initial measurement for rate-based collectors
@@ -441,7 +443,8 @@ class SysMetricCollectorThread(threading.Thread):
                 try:
                     msg = self.control_queue.get(timeout=sleep_time)
                     if isinstance(msg, ShutdownMessage):
-                        # Drop any partial batches if we receive a shutdown
+                        self.send_queue.put(self.current_batch.convert_to_post_format())
+                        self.send_queue.put(ShutdownMessage())
                         return
                 except queue.Empty:
                     pass
@@ -527,6 +530,7 @@ class TimingsBatcherThread(threading.Thread):
             if isinstance(msg, StartMessage):
                 break
             if isinstance(msg, ShutdownMessage):
+                self.send_queue.put(ShutdownMessage())
                 return
             else:
                 # Ignore any Timings that are received before StartMessage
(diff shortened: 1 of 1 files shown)


harness/determined/profiler.py
```diff
@@ -546,8 +512,10 @@ def run(self) -> None:
                 message = self.inbound_queue.get(timeout=timeout)
                 if isinstance(message, ShutdownMessage):
                     # Drop any partial batches if we receive a shutdown
```
stoksc, line 514:
```suggestion
```

harness/determined/profiler.py
```diff
@@ -229,11 +232,12 @@ def record_timing(self, metric_name: str) -> Iterator[None]:
...
 
-        timing = Timing(metric_name, self.current_batch_idx)
-        timing.start()
+        start_time = time.perf_counter()
         yield
-        timing.end()
-        self.timings_batcher_queue.put(timing)
+        dur = time.perf_counter() - start_time
+        self.timings_batcher_queue.put(
+            NamedMeasurement(metric_name, datetime.datetime.utcnow(), self.current_batch_idx, dur)
```
armandmcqueen, line 239: This puts the timestamp as some time after the timing has completed. I think the associated timestamp for an event should be the start time of the event.

harness/determined/profiler.py
```diff
@@ -15,7 +15,7 @@
 
 SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
 TIMING_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_TIMING"
-
+MAX_COLLECTION_SECONDS = 300
```
armandmcqueen, line 18: Why are we moving this to a global variable? I think it is quite likely that this is something we will want to configure and so it should remain a property of the `ProfilingAgent`

harness/determined/profiler.py
```diff
@@ -281,51 +285,11 @@ def _end_collection(self) -> None:
                 self.timings_batcher_thread.send_shutdown_signal()
                 self.timings_batcher_thread.join()
 
-            self.sender_thread.send_shutdown_signal()
             self.sender_thread.join()
 
             self.has_finished = True
 
 
-class Timing:
```
armandmcqueen, line 290: IMO we should keep Timings or something similar so it is easy to give users a well-defined way to provide their own timings without relying on our context manager - particularly as we add non-PyTorch frameworks or let users add their own as generic APIs

harness/determined/profiler.py
stoksc: I think making it a property is easy enough once we want it configurable; it was just easier for me to find this way.

harness/determined/profiler.py
stoksc: Yeah, noticed that. Not sure if perf counters converts well to a datetime, maybe we use time.time after all.

harness/determined/profiler.py
stoksc: Yeah I agree we'll need a nice user facing API, we could probably have an extended discussion about the right API, but I'm not sure if we should try to have it here since this is not user facing yet. I was just trying to save some LOC. If you still want this I can add it back.

harness/determined/profiler.py
```diff
@@ -101,16 +107,13 @@ def __init__(
...
-                self.max_collection_seconds, self._end_collection
-            )
+            self.shutdown_timer = PreemptibleTimer(MAX_COLLECTION_SECONDS, self._end_collection)
 
             # Set up the thread responsible for making API calls
             self.send_queue = (
                 queue.Queue()
             )  # type: """queue.Queue[Union[List[TrialProfilerMetricsBatch], ShutdownMessage]]"""
-            self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+            self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url, 2)
```
armandmcqueen, line 116: Won't `sender_thread` shutdown hang forever on all GPUs except for `global_rank=0` since one or both of the `sys_metric_collector_thread` and `timings_batcher_thread` won't exist?

armandmcqueen: That wasn't quite accurate. It will hang for all GPUs that are `local_rank=0`, but `global_rank != 0` because the timings batcher thread won't exist and so the sender thread will never shut down. All other GPUs will work correcty because they are either `global_rank=0` which this code was written for or they are not `local_rank=0` in which case `is_enabled` will be `False`.

stoksc: yeh, I'll try to handle this end case. I'd rather this than drop partial batches because I think then we won't end up actually respecting start on / end after batch.

stoksc: fixed

harness/determined/profiler.py
```diff
@@ -579,21 +585,23 @@ def run(self) -> None:
...
 
-    def send_shutdown_signal(self) -> None:
-        self.inbound_queue.put(ShutdownMessage())
-
     def run(self) -> None:
         while True:
             message = self.inbound_queue.get()
             if isinstance(message, ShutdownMessage):
+                self.producers_shutdown += 1
+            if self.num_producers == self.producers_shutdown:
```
armandmcqueen, line 604: nit: I would indent this check so it only happens when a `ShutdownMessage` is received. Same outcome, but a little easier to follow IMO. Up to you.

harness/determined/profiler.py
```diff
@@ -100,17 +100,20 @@ def __init__(
...
             self.send_queue = (
                 queue.Queue()
             )  # type: """queue.Queue[Union[List[TrialProfilerMetricsBatch], ShutdownMessage]]"""
-            self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+
+            # Set up timer thread to stop collecting after 5 minutes
+            self.shutdown_timer = PreemptibleTimer(MAX_COLLECTION_SECONDS, self._end_collection)
+
+            # Set up the thread responsible for making API calls
+            if self.sysmetrics_is_enabled or self.timings_is_enabled:
```
armandmcqueen, line 112: This check is already handled by `if self.is_enabled` (`is_enabled` will be `False` if both `sysmetrics_is_enabled` and `timings_is_enabled` are `False`). However, I forgot that it worked like that when I wrote my original comment and you didn't notice it when you were writing the code, so maybe it is unintuitive and should be changed. Or just drop this `if`

harness/determined/profiler.py
```diff
@@ -545,6 +549,8 @@ def run(self) -> None:
                 message = self.inbound_queue.get(timeout=timeout)
                 if isinstance(message, ShutdownMessage):
                     # Drop any partial batches if we receive a shutdown
```
armandmcqueen, line 551:
```suggestion
```

harness/determined/profiler.py
```diff
@@ -100,17 +100,20 @@ def __init__(
...
-            )
-
-            # Set up the thread responsible for making API calls
+            # Set up sender thread with the number of producers to expect shutdowns from.
             self.send_queue = (
                 queue.Queue()
             )  # type: """queue.Queue[Union[List[TrialProfilerMetricsBatch], ShutdownMessage]]"""
-            self.sender_thread = ProfilerSenderThread(self.send_queue, self.master_url)
+
+            # Set up timer thread to stop collecting after 5 minutes
```
armandmcqueen, line 108:
```suggestion
            # Set up timer thread to stop collecting after MAX_COLLECTION_SECONDS
```

harness/determined/profiler.py
```diff
@@ -100,17 +100,20 @@ def __init__(
...
+            self.shutdown_timer = PreemptibleTimer(MAX_COLLECTION_SECONDS, self._end_collection)
```
armandmcqueen, line 109: nit: is there a reason creating the `shutdown_timer` was moved down to be between `self.send_queue` and creating `self.sender_thread`?

armandmcqueen (approved):
A couple of minor comments, but otherwise it looks good!

harness/determined/profiler.py
stoksc: I will fix. I literally just was moving where send queue was made around so that it would be within line length since i couldn't figure out how to write a multiline `# type` comment

harness/determined/profiler.py
stoksc: kinda just reordered stuff to make it look better/more logical

PR #2475, opened 2021-06-01
chore: measure profiler timings with time.time()

armandmcqueen:
## Description

Change `ProfilerAgent` to record timings using `time.time()` instead of `datetime.utcnow()`. For timing small segments of code, this will make the measurements more accurate as `utcnow()` does conversion work that was previously included as part of the measurement. 

`time.time()` was chosen over `time.time_ns()` because that was introduced in Python 3.7 and at the time of coding we needed to support 3.6. It was chosen over `time.perf_counter()` and `time.monotonic()` because they are only useful for relative measurements (i.e. calculating durations). We need to record a timestamp to send to the DB so using those clocks would require us to store a timestamp in addition to the starting perf_counter/monotonic value. That seems to complicate the code more than the benefits are worth - we don't think the additional granularity of `perf_counter` is useful (the webui displays millisecond granularity) and we don't expect the system clock to change during training.

Also change all datetimes to be explicitly UTC so they are timezone aware.

## Commentary (optional)

The PEP that introduced `time_ns` is [here](https://www.python.org/dev/peps/pep-0564) and it provides some useful information around performance and precision, although it dates to 2017. TLDR: given that we don't care about a couple of hundred nanoseconds, `time.time()` should be fine.

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -1,9 +1,9 @@
 import contextlib
-import datetime
 import logging
 import queue
 import threading
 import time
+from datetime import datetime, timezone
 from types import TracebackType
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union, cast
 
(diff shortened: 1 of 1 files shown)


stoksc:
@armandmcqueen we should chat, i hit this in https://github.com/determined-ai/determined/pull/2473

stoksc:
if we go with this one, i think we should use time.monotonic() not time.time()

stoksc (approved):
lgtm

PR #2478, opened 2021-06-01
fix: add init_invalid_hp to master [DET-5569]

naren-determined:
## Description

Currently, when an invalid hyperparameter is detected by the trial it gracefully exits causing a checkpoint after the initial workload is completed. However, if an invalid hp is detected in the trial __init__(), we want to hard exit and not have the trial checkpoint a workload. This fix distinguishes invalid HPs raised in the init vs. after a trial has run at least one training step.

## Test Plan

Test by running https://github.com/determined-ai/determined/tree/master/examples/features/hp_constraints_mnist_pytorch and triggering InvalidHPs in the __init__() and train_batch().

## Checklist


rb-determined-ai:
@naren-determined, is the workload sequence the only reason to distinguish these two failures?

We are currently refactoring the harness pretty aggressively for push arch (https://github.com/determined-ai/determined/pull/2470) and afterwards the whole workload sequencer will go into the python code.  So if it is possible to wait to fix this until then, I think that would be preferable.

What is the buggy behavior here, by the way?  There is no description in the PR at all.

naren-determined:
> @naren-determined, is the workload sequence the only reason to distinguish these two failures?
> 
> We are currently refactoring the harness pretty aggressively for push arch (#2470) and afterwards the whole workload sequencer will go into the python code. So if it is possible to wait to fix this until then, I think that would be preferable.
> 
> What is the buggy behavior here, by the way? There is no description in the PR at all.

Added a description to the PR. At the moment, the invalid hp feature is broken if there is an invalid HP used in the init. So between the experiments we need to run and the blog post coming out, I don't think waiting for a few weeks is a good idea.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -816,6 +817,7 @@ def _control_loop(self) -> None:
                             {},
                             self.context.get_stop_requested(),
                             invalid_hp=True,
+                            init_invalid_hp=False,
```
stoksc, line 820: any reason these aren't all optional and false by default?

master/pkg/searcher/searcher.go
```diff
@@ -107,6 +107,8 @@ func (s *Searcher) TrialExitedEarly(
 	switch exitedReason {
 	case workload.InvalidHP:
 		delete(s.TrialProgress, requestID)
+	case workload.InitInvalidHP:
+		delete(s.TrialProgress, requestID)
```
stoksc, line 111:
```suggestion
	case workload.InvalidHP, workload.InitInvalidHP:
		delete(s.TrialProgress, requestID)
```

master/internal/trial.go
```diff
@@ -705,6 +706,10 @@ func (t *trial) processCompletedWorkload(ctx *actor.Context, msg workload.Comple
 		reason := *msg.ExitedReason
 		ctx.Log().Infof("exiting trial early from %v with reason %v", msg.Workload, reason)
 
+		if reason == workload.InitInvalidHP {
+			t.TerminateNoCheckpoint = true
```
stoksc, line 710: Can you leave a comment on the decl of this field that describes its purpose? Something like "... this flag marks an immediate exit that is unrecoverable"? Maybe be nice to just call it `UnrecoverableExit` or somthing (the more general/accurate what it does, the more likely it will be reused properly if needed).

stoksc (approved):
lgtm just minor comments.

master/internal/trial.go
stoksc: But, actually don't worry about it because I'm going to refactor this shortly.

PR #2480, opened 2021-06-01
chore: remove native init() functions [DET-5574]

rb-determined-ai:
## Description

Remove native functions that were deprecated in 0.13.5.

Also refactor the `load` module to now be much simpler.

Added e2e and unit test for det.experimental.create.


harness/determined/load/_native.py
```diff
@@ -0,0 +1,151 @@
+import contextlib
```
shiyuann, line 1: Why do we still have this file?

rb-determined-ai: Because we are not removing `det.experimental.create()` yet, due to customers who are actively using it.

rb-determined-ai: added some tests for the `det.experimental.create()` codepath; previously there were none at all.

PR #2481, opened 2021-06-01
test: add e2e tests for profiling features [DET-5245]

stoksc:
## Description
This change adds tests to cover all the observability features.

## Test Plan
- [ ] make sure tests pass

## Commentary (optional)
unfortunately, estimators is failing and it's a genuine failure so i have to fix that first.

## Checklist

### DIFF
--- a/e2e_tests/tests/experiment/test_profiling.py
+++ b/e2e_tests/tests/experiment/test_profiling.py
@@ -1,0 +1,206 @@
+import multiprocessing as mp
+import tempfile
+from typing import Any, Dict, Optional, Sequence, Tuple
+from urllib.parse import urlencode
+
+import pytest
+import simplejson
+import yaml
+
+import determined.common.api.authentication as auth
+from determined.common import api
+from tests import config as conf
+from tests import experiment as exp
+
+Result = Optional[Sequence[Any]]
+
+
+@pytest.mark.e2e_gpu  # type: ignore
+@pytest.mark.timeout(600)  # type: ignore
+@pytest.mark.parametrize(  # type: ignore
+    "framework_base_experiment,framework_timings_enabled",
+    [
+        ("tutorials/mnist_pytorch", True),
+        ("tutorials/fashion_mnist_tf_keras", False),
+        ("computer_vision/mnist_estimator", False),
+    ],
+)
+def test_streaming_observability_metrics_apis(
+    framework_base_experiment: str, framework_timings_enabled: bool
+) -> None:
+    auth.initialize_session(conf.make_master_url(), try_reauth=True)
+
+    # To test observability, we kick off an experiment with profiling.enabled = true,
+    # then we stream the observability metric APIs.
+
+    config_path = conf.tutorials_path(f"../{framework_base_experiment}/const.yaml")
+    model_def_path = conf.tutorials_path(f"../{framework_base_experiment}")
+
+    config_obj = conf.load_config(config_path)
+    config_obj = conf.set_profiling_enabled(config_obj)
+    with tempfile.NamedTemporaryFile() as tf:
+        with open(tf.name, "w") as f:
+            yaml.dump(config_obj, f)
+        experiment_id = exp.create_experiment(
+            tf.name,
+            model_def_path,
+        )
+
+    exp.wait_for_experiment_state(experiment_id, "COMPLETED")
+    trials = exp.experiment_trials(experiment_id)
+    trial_id = trials[0]["id"]
+
+    pool = mp.pool.ThreadPool(processes=3)
+    metric_labels_thread = pool.apply_async(
+        request_profiling_metric_labels, (trial_id, framework_timings_enabled)
+    )
+    gpu_util_metrics_thread = pool.apply_async(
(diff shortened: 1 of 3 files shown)


e2e_tests/tests/experiment/test_profiling.py
```diff
@@ -0,0 +1,204 @@
...
+        experiment_id = exp.create_experiment(
+            tf.name,
+            model_def_path,
+        )
+
+    exp.wait_for_experiment_state(experiment_id, "COMPLETED")
+    trials = exp.experiment_trials(experiment_id)
+    trial_id = trials[0]["id"]
+
+    pool = mp.pool.ThreadPool(processes=3)
```
armandmcqueen, line 51: I would suggest using [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) to create threadpools as a context manager. 

`mp.pool.ThreadPool` predates `concurrent.futures` and the docs now [recommend using `concurrent.futures` instead](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.ThreadPool)

e2e_tests/tests/experiment/test_profiling.py
```diff
@@ -0,0 +1,204 @@
...
+        pytorch_timings_thread = pool.apply_async(
+            request_profiling_pytorch_timing_metrics, (trial_id, "train_batch")
+        )
+
+    metric_labels_results = metric_labels_thread.get()
+    gpu_util_metrics_result = gpu_util_metrics_thread.get()
+    if framework_timings_enabled:
+        pytorch_timings_result = pytorch_timings_thread.get()
+
+    if metric_labels_results is not None:
```
armandmcqueen, line 68: This is Golang-style code implemented in Python. I don't recommend it. Definitely don't do Golang-style error handling where `None` means no error. Either use exceptions or write code that doesn't have errors, just outcomes.

Python code is generally more straightforward and only rarely dips into parallelism when it's really needed. Pythonic code is typically serial, simple, and explicit.

To make this code fit better in Python world, I would get rid of the threadpool and write one simple test function with a couple of helper functions for readability.

e.g.

```python
with api.get(conf.make_master_url(), "api/v1/trials/{}/profiler/available_series".format(trial_id), stream=True) as r:
    for line in r.iter_lines():
        label_batch = simplejson.loads(line)["result"]["labels"]
        missing_labels_for_batch = missing_labels(label_batch, timing_enabled=framework_timings_enabled)
        if len(missing_labels_for_batch) > 0:
           pytest.fail(f"failed to collect metric labels. expected completed experiment to all labels, but some where missing: {missing_labels}")

with api.get(conf.make_master_url(), "api/v1/trials/{}/profiler/metrics?{}".format(trial_id, to_query_params(PROFILER_METRIC_TYPE_SYSTEM, metric_name)), stream=True) as r:
    for line in r.iter_lines():            
        batch = simplejson.loads(line)["result"]["batch"]
        num_values = len(batch["values"])
        num_batch_indexes = len(batch["batches"])
        num_timestamps = len(batch["timestamps"])
        if num_values != num_batch_indexes or num_batch_indexes != num_timestamps:
            pytest.fail("blah blah blah)
        if num_values == 0:           
            pytest.fail("blah blah blah)

# etc
```
Python code heavily prioritizes simplicity of code over performance.

If you really need to do multithreading, you could add a `ValidationResult` where you can explicitly check stuff like:
```python
if result.failed:
    pytest.fail("label validation failed due to {result.error}
```
But I really wouldn't do it unless the performance differences are on the order of minutes

stoksc: TLDR why it looks like this, I just copy pasta'd the tests from `test_metrics.py` from HP viz and originally it _needed_ to be parallel because I was going to test that the data actually streamed in properly but then I decided that was overkill and left it this way. I'll just remove the process pool, it's essentially not used as is.

stoksc: > Python code heavily prioritizes simplicity of code over performance.

TBH I would be since I'm waiting until the exp is created, the process is likely closed to less efficient here.

e2e_tests/tests/experiment/test_profiling.py
stoksc: Going to just remove the pool

e2e_tests/tests/experiment/test_profiling.py
```diff
@@ -0,0 +1,204 @@
...
+        "api/v1/trials/{}/profiler/available_series".format(trial_id),
+        stream=True,
+    ) as r:
+        for line in r.iter_lines():
+            labels = simplejson.loads(line)["result"]["labels"]
+            # Since the result from available-series is logically monotonic,
+            # if we see all the labels we can be sure this is true going forward.
+            missing_labels = labels_missing(labels)
+            if len(missing_labels) > 0:
+                return f"expected completed experiment to have missing labels: {missing_labels}"
```
stoksc, line 116:
```suggestion
                return f"expected completed experiment to have all labels but some are missing: {missing_labels}"
```

armandmcqueen (approved):
LGTM!

PR #2482, opened 2021-06-01
fix: ProfilingAgent serializing timestamps incorrectly

armandmcqueen:
## Description

Fix bug where some datetimes were being serialized in a valid format, while others were not and causing POSTs to fail.

Python's `datetime.datetime` objects can be naive or timezone-aware. When `isoformat()` is called on them, the output is different depending on whether it is timezone-aware. During the initial implementation, all `datetime`s were naive, which meant the serialized form did not include any timezone information. We appended 'Z' to all timestamp strings to indicate that they are in UTC. PR #2475 changed most of the datetimes to be timezone-aware so appending 'Z' to all timestamps led to "2021-06-01T21:41:23.411254+00:00Z" instead of the intended "2021-06-01T21:41:23.411254Z". Now that we are working with timezone aware datetimes, the best way to serialize them is to just call `isoformat()` to get "2021-06-01T21:41:23.411254+00:00" which is valid.

However, this exposed the fact that `datetime.utcnow()` actually creates a datetime that is timezone-unaware, which caused new errors. So all uses of `datetime.utcnow()` were replaced with the preferred `datetime.now(timezone.utc)` which is timezone aware. And now the serialization logic performs a check to ensure that all datetimes are timezone-aware and in UTC to avoid this headache in the future.

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -3,7 +3,7 @@ import logging
 import queue
 import threading
 import time
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from types import TracebackType
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union, cast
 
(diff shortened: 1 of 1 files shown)


harness/determined/profiler.py
```diff
@@ -762,14 +773,14 @@ def convert_to_post_format(self) -> List[TrialProfilerMetricsBatch]:
 class SimpleCpuUtilCollector:
     def measure(self, batch_idx: int) -> Measurement:
         cpu_util = psutil.cpu_percent()
-        timestamp = datetime.utcnow()
+        timestamp = datetime.now(timezone.utc)
```
stoksc, line 776: perhaps the API should take just take unix timestamp nanos instead?

stoksc (approved):
lgtm for now, left a comment for a large discussion

harness/determined/profiler.py
armandmcqueen: I don't have an opinion either way. Let's discuss at standup.

PR #2484, opened 2021-06-02
chore: add non-streaming APIs for trial profiler endpoints

armandmcqueen:
## Description

Add the `follow` param to the Profiling APIs to indicate whether the API should stream results until the trial is complete.

## Commentary (optional)

The available series changes are required for the ProfilingAgent to know whether profiling results have already been recorded for the trial so it can know to not add more profiling data.

I added the flag to both APIs because some tools don't have good support for streaming (e.g. Swagger) and it is really convenient to have non-streaming versions of the APIs.

## Checklist

### DIFF
--- a/harness/determined/_swagger/client/api/profiler_api.py
+++ b/harness/determined/_swagger/client/api/profiler_api.py
@@ -43,7 +43,7 @@ class ProfilerApi(object):
 
         :param async_req bool
         :param int trial_id: The requested trial's id. (required)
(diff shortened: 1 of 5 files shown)


stoksc (approved):
all this new swagger stuff is a mystery to me but otherwise lgtm. the follow false on the streamer will just terminate on the first empty batch.

apizzini (approved):
LGTM

PR #2486, opened 2021-06-02
chore: disable profiling after restart [DET-5424]

armandmcqueen:
## Description

Disable profiling if profiling data already exists in the API. This avoids problems around multiple data points with the same batch index, timeseries graphs with large gaps, and confusing data caused by GPU warmup if the user didn't notice the restart.

Also change the POST API function to be passed in to the ProfilingAgent constructor instead of being hardcoded. Making all functions that communicate with the master API easy to overwrite will make it easier to write unit tests without a running master.

Additionally, fix missing task token authentication logic for streaming APIs and unify auth logic across unary and streaming APIs.

## Checklist

### DIFF
--- a/harness/determined/common/api/profiler.py
+++ b/harness/determined/common/api/profiler.py
@@ -1,7 +1,6 @@
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import backoff
-import requests
 from requests.exceptions import RequestException
 
 from determined.common import api
@@ -49,6 +48,15 @@ def post_trial_profiler_metrics_batches(
     )
 
 
+class TrialProfilerSeriesLabels:
+    def __init__(self, trial_id: int, name: str, agent_id: str, gpu_uuid: str, metric_type: str):
+        self.trial_id = str(trial_id)
+        self.name = name
+        self.agent_id = agent_id
+        self.gpu_uuid = gpu_uuid if gpu_uuid != "" else None  # type: Optional[str]
+        self.metric_type = metric_type
+
+
 @backoff.on_exception(  # type: ignore
     backoff.constant,
     RequestException,
@@ -58,7 +66,7 @@ def post_trial_profiler_metrics_batches(
 def get_trial_profiler_available_series(
     master_url: str,
     trial_id: str,
-) -> None:
+) -> List[TrialProfilerSeriesLabels]:
     """
     Get available profiler series for a trial. This uses the non-streaming version of the API
     """
@@ -66,6 +74,17 @@ def get_trial_profiler_available_series(
     response = api.get(
         host=master_url,
         path=f"/api/v1/trials/{trial_id}/profiler/available_series",
-        params={"follow": follow}
+        params={"follow": follow},
     )
-    print(response)
+    j = response.json()
+    labels = [
+        TrialProfilerSeriesLabels(
+            trial_id=ld["trialId"],
+            name=ld["name"],
+            agent_id=ld["agentId"],
+            gpu_uuid=ld["gpuUuid"],
+            metric_type=ld["metricType"],
+        )
+        for ld in j["result"]["labels"]
+    ]
+    return labels
--- a/harness/determined/common/api/profiler.py
+++ b/harness/determined/common/api/profiler.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List
 
 import backoff
+import requests
 from requests.exceptions import RequestException
 
 from determined.common import api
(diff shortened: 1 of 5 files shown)


harness/determined/profiler.py
```diff
@@ -81,22 +101,33 @@ def __init__(
...
+        self.global_rank = global_rank
+        self.local_rank = local_rank
         self.start_on_batch = start_on_batch
         self.end_after_batch = end_after_batch
-        self.local_rank = local_rank
-        self.global_rank = global_rank
-
-        self.profiling_is_enabled_in_experiment_config = profiling_is_enabled
+        self.send_batch_fn = send_batch_fn
+        self.check_data_already_exists_fn = check_data_exists_fn
```
stoksc, line 117: Is this going to be a convenience for testing?

harness/determined/profiler.py
```diff
@@ -81,22 +101,33 @@ def __init__(
...
         self.has_finished = False
 
         self.shutdown_lock = threading.Lock()
+        self.disabled_due_to_preexisting_metrics = self.check_data_already_exists_fn(
+            self.master_url, self.trial_id
+        )
+        if self.disabled_due_to_preexisting_metrics and self.global_rank == 0:
+            logging.warning(
+                f"{LOG_NAMESPACE}: ProfilerAgent is disabled because profiling data for this "
+                f"trial already exists in the DB. No additional profiling data will be generated."
```
stoksc, line 129:
```suggestion
                f"trial already exists. No additional profiling data is generated after a restart."
```

harness/determined/profiler.py
```diff
@@ -48,6 +48,18 @@ class ShutdownMessage:
     pass
 
 
+def profiling_metrics_exist(master_url: str, trial_id: str) -> bool:
+    """
+    Return True if there are already profiling metrics for the trial.
+    """
+    # TODO: Implement
+    raise NotImplementedError
```
stoksc, line 56: Not sure is this was the plan after that PR yesterday, this could just call `GET:master_url/api/v1/trial/:trial_id/profiler/available_series?follow` and exit on any return, right?

harness/determined/profiler.py
armandmcqueen: Yeah, writing unit tests is much easier if we can replace the functions that communicate with the master when creating the `ProfilingAgent`

harness/determined/profiler.py
armandmcqueen: Updated the code

master/internal/grpcutil/auth.go
```diff
@@ -40,7 +40,7 @@ var (
 	// ErrInvalidCredentials notifies that the provided credentials are invalid or missing.
 	ErrInvalidCredentials = status.Error(codes.Unauthenticated, "invalid credentials")
 	// ErrTokenMissing notifies that the bearer token could not be found.
-	ErrTokenMissing = status.Error(codes.InvalidArgument, "token missing")
+	ErrTokenMissing = status.Error(codes.Unauthenticated, "token missing")
```
armandmcqueen, line 43: The `InvalidArgument` code was very confusing to me when debugging

stoksc: Note to check: this feels like the kinda thing to be careful about (e.g. make sure the CLI doesn't rely on a 400 in a bad way somewhere).

armandmcqueen: I searched the codebase for "400", "bad request" and "badrequest". The only relevant thing I saw was that the CLI will handle displaying the error slightly differently (displaying a traceback instead of printing the Exception) - https://github.com/determined-ai/determined/blob/master/harness/determined/cli/cli.py#L283.

But since that is how we currently handle unauthenticated errors, I think it's fine? Or we could add Unauthenticated errors to the list of errors where we print out the Python Exception

harness/determined/common/api/profiler.py
```diff
@@ -46,3 +46,45 @@ def post_trial_profiler_metrics_batches(
...
+    giveup=lambda e: e.response is not None and e.response.status_code < 500,
+)
+def get_trial_profiler_available_series(
+    master_url: str,
+    trial_id: str,
+) -> List[TrialProfilerSeriesLabels]:
+    """
+    Get available profiler series for a trial. This uses the non-streaming version of the API
+    """
+    follow = False
```
stoksc, line 73: maybe just make this an optional param?

master/test/integration/api/api_trials_test.go
```diff
@@ -247,7 +247,7 @@ func trialProfilerMetricsAvailableSeriesTests(
 	ctx, _ := context.WithTimeout(creds, time.Minute)
 	tlCl, err := cl.GetTrialProfilerAvailableSeries(ctx, &apiv1.GetTrialProfilerAvailableSeriesRequest{
 		TrialId: int32(trial.ID),
-		Follow: true,
+		Follow:  true,
```
stoksc, line 250: yeah, turns out this isn't linted because of the build tag.. oops.

master/internal/grpcutil/auth.go
```diff
@@ -105,13 +105,33 @@ func GetUser(ctx context.Context, d *db.PgDB) (*model.User, *model.UserSession,
...
+	}
+	return nil
+}
+
 func streamAuthInterceptor(db *db.PgDB) grpc.StreamServerInterceptor {
 	return func(
 		srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler,
 	) error {
-		if _, _, err := GetUser(ss.Context(), db); err != nil {
+		err := auth(ss.Context(), db, info.FullMethod)
```
stoksc, line 130: good catch

master/internal/grpcutil/auth.go
stoksc: Yeah, looks fine to me.

stoksc (approved):

harness/determined/common/api/profiler.py
armandmcqueen: I hardcoded it to `False` because I think the interface for a streaming API should probably be streaming-centric, but I'm not sure exactly what it should be (maybe it returns a generator?). That problem is best solved when someone has a use case for it..

PR #2487, opened 2021-06-02
chore: improved pynvml usage by profiler [DET-5394]

armandmcqueen:
## Description

Wrap `pynvml` in a class that checks if `pynvml` is available and if `pynvml` is working. Abstracts away these checks from the user and just exposes a `is_available` boolean. Also wraps the various pynmvl APIs so that they are now typed and have some simple caching.

Also improves the design of the GPU memory/util collectors so that they will not cause errors if used when there are not GPUs (they just return an empty dict of measurements). 

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -24,6 +24,13 @@ class PynvmlWrapperError(Exception):
 
 
 class PynvmlWrapper:
+    """
+    Class to wrap pynvml. Handle checks around whether the nvidia management
+    library is available, whether the pynvml bindings are installed, and
+    whether actual utilization/memory operations are successful (they may not
+    be in some cases, for example when using MIGs).
+    """
+
     def __init__(self) -> None:
         self._pynvml = None  # type: Optional[Any]
         self._device_count = None  # type: Optional[int]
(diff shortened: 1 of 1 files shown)


harness/determined/profiler.py
```diff
@@ -18,19 +18,96 @@
...
+                    f"{LOG_NAMESPACE}: pynvml is functional, but failed to pass functionality "
+                    f"test due to exception. Not collecting GPU metrics. Exception details: {e}"
+                )
+        except ModuleNotFoundError:
+            logging.info(f"{LOG_NAMESPACE}: pynvml not found. Not collecting GPU metrics")
+        except pynvml.NVMLError_LibraryNotFound:
+            logging.info(
+                f"{LOG_NAMESPACE}: pynvml LibraryNotFound error. Not collecting GPU metrics"
+            )
+        except Exception as e:
```
stoksc, line 54: Is this even possible? I think we just shouldn't even catch this one maybe? Idk, I guess this is strictly better since it tells users exactly where/what happened.

harness/determined/profiler.py
```diff
@@ -18,19 +18,96 @@
...
+            raise PynvmlWrapperError(
+                "Tried to call a pynvml operation but pynvml is either unavailable or not "
+                "functional. Code should check pynvml_is_working before calling any operations."
+            )
+
+    def nvml_get_device_count(self) -> int:
+        self._safety_check()
+        self._pynvml = cast(Any, self._pynvml)
+
+        if self._device_count is None:
```
stoksc, line 78: could we just set this in init and make this a prop?

stoksc (approved):
lgtm. probably a bit further than I would've gone (with the `PynvmlWrapper`), but it's good.

harness/determined/profiler.py
armandmcqueen: fixed

PR #2489, opened 2021-06-02
docs: add db migration troubleshooting section [DET-5493]

stoksc:
## Description
Add some documentation around migration failures.

## Test Plan
- [ ] compile and 👁️ docs

## Checklist

### DIFF
--- a/docs/how-to/installation/upgrades-troubleshoot.txt
+++ b/docs/how-to/installation/upgrades-troubleshoot.txt
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
 
    docker run --gpus all --rm nvidia/cuda:9.0-runtime nvidia-smi
    docker run --gpus all --rm nvidia/cuda:10.0-runtime nvidia-smi
+
+*****************************
+ Database Migration Failures
+*****************************
+
+.. code::
+
+   Dirty database version A_LONG_NUMBER. Fix and force version.
+
+If you see the above error message, a database migration was likely
+interrupted while running and the database is now in a dirty state.
+
+Make sure you backup the database and temporarily shutdown the master
+before proceeding further.
+
+To fix this error message, locate the up migration with a suffix of
+``.up.sql`` and a migration number matching A_LONG_NUMBER in the error
+message in `here
+<https://github.com/determined-ai/determined/tree/master/master/static/migrations>_`
+and carefully run the SQL within the file manually against the database
+used by Determined. For convenience, all the information needed to
+connect except the password can be found with:
+
+.. code::
+
+   det master config | jq .db
+
+If this proceeds successfully, then mark the migration as successful by
+running the following SQL:
+
+.. code::
+
+   UPDATE schema_migrations SET dirty = false
+
+And restart the master. Otherwise, please seek assistance in the
+community slack.
--- a/docs/how-to/installation/upgrades-troubleshoot.txt
+++ b/docs/how-to/installation/upgrades-troubleshoot.txt
@@ -138,12 +138,12 @@ be able to use Determined as long as you use Docker images based on CUDA
 
 .. code::
 
-   Dirty database version A_LONG_NUMBER. Fix and force version.
+   Dirty database version <a long number>. Fix and force version.
 
 If you see the above error message, a database migration was likely
 interrupted while running and the database is now in a dirty state.
 
-Make sure you backup the database and temporarily shutdown the master
+Make sure you back up the database and temporarily shut down the master
 before proceeding further.
 
 To fix this error message, locate the up migration with a suffix of
@@ -163,7 +163,7 @@ running the following SQL:
 
 .. code::
 
-   UPDATE schema_migrations SET dirty = false
+   UPDATE schema_migrations SET dirty = false;
 
 And restart the master. Otherwise, please seek assistance in the
 community slack.
--- a/docs/how-to/installation/upgrades-troubleshoot.txt
+++ b/docs/how-to/installation/upgrades-troubleshoot.txt
@@ -147,8 +147,8 @@ Make sure you back up the database and temporarily shut down the master
 before proceeding further.
 
 To fix this error message, locate the up migration with a suffix of
-``.up.sql`` and a migration number matching A_LONG_NUMBER in the error
-message in `here
+``.up.sql`` and a prefix matching the long number in the error
+message in `this directory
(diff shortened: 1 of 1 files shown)


docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+*****************************
+
+.. code::
+
+   Dirty database version A_LONG_NUMBER. Fix and force version.
+
+If you see the above error message, a database migration was likely
+interrupted while running and the database is now in a dirty state.
+
+Make sure you backup the database and temporarily shutdown the master
```
dzhu, line 146:
```suggestion
Make sure you back up the database and temporarily shut down the master
```

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+before proceeding further.
+
+To fix this error message, locate the up migration with a suffix of
```
dzhu, line 149: In general, doesn't this process depend on exactly what the migration is doing? Like if there are two long statements and the first one finishes, they're probably going to need to dig in and extract part of the migration code to run. Maybe this is the closest thing that's worth putting into a general troubleshooting doc, though. Can't exactly teach everybody to debug arbitrary migrations right here.

Also, as a possible alternative process (for the cases when the migration needs to be rerun in its entirety): I would go into the database and check that it matches the pre-migration state, then set the version to the previous version and `dirty` to false and restart the master, so it'll automatically retry that migration and then just continue from there. Feels a bit nicer to me than running a possibly large migration interactively. Think that would make sense to mention?

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+.. code::
+
+   det master config | jq .db
+
+If this proceeds successfully, then mark the migration as successful by
+running the following SQL:
+
+.. code::
+
+   UPDATE schema_migrations SET dirty = false
```
dzhu, line 166:
```suggestion
   UPDATE schema_migrations SET dirty = false;
```

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
    docker run --gpus all --rm nvidia/cuda:9.0-runtime nvidia-smi
    docker run --gpus all --rm nvidia/cuda:10.0-runtime nvidia-smi
+
+*****************************
+ Database Migration Failures
+*****************************
+
+.. code::
+
+   Dirty database version A_LONG_NUMBER. Fix and force version.
```
dzhu, line 141: Obviously not a hard standard, but this looks a bit more natural to me:
```suggestion
   Dirty database version <a long number>. Fix and force version.
```

docs/how-to/installation/upgrades-troubleshoot.txt
stoksc: > In general, doesn't this process depend on exactly what the migration is doing?

Yeah, I mean, ideally our migrations should either be idempotent or in a transaction (so in a transaction since the former isn't always possible with e.g. renames). Then this would always work.

> I would go into the database and check that it matches the pre-migration state, then set the version to the previous version and dirty to false and restart the master, so it'll automatically retry that migration and then just continue from there.

I've thought about this exactly; the issue is if the migration failed for some reason (like time limit from some admin, idk, I've seen it once before) then it's just going to fail over and over again this way. You could say do this + don't mess with the DB at all while it is running.

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+   Dirty database version <a long number>. Fix and force version.
+
+If you see the above error message, a database migration was likely
+interrupted while running and the database is now in a dirty state.
+
+Make sure you back up the database and temporarily shut down the master
+before proceeding further.
+
+To fix this error message, locate the up migration with a suffix of
+``.up.sql`` and a migration number matching A_LONG_NUMBER in the error
```
dzhu, line 150:
```suggestion
``.up.sql`` and a prefix matching the long number in the error
```

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+
+If this proceeds successfully, then mark the migration as successful by
+running the following SQL:
+
+.. code::
+
+   UPDATE schema_migrations SET dirty = false;
+
+And restart the master. Otherwise, please seek assistance in the
+community slack.
```
dzhu, line 169:
```suggestion
community `Slack
<https://join.slack.com/t/determined-community/shared_invite/zt-cnj7802v-KcVbaUrIzQOwmkmY7gP0Ew>`__.
```

docs/how-to/installation/upgrades-troubleshoot.txt
```diff
@@ -131,3 +131,39 @@ be able to use Determined as long as you use Docker images based on CUDA
...
+
+If you see the above error message, a database migration was likely
+interrupted while running and the database is now in a dirty state.
+
+Make sure you back up the database and temporarily shut down the master
+before proceeding further.
+
+To fix this error message, locate the up migration with a suffix of
+``.up.sql`` and a migration number matching A_LONG_NUMBER in the error
+message in `here
```
dzhu, line 151:
```suggestion
message in `this directory
```

docs/how-to/installation/upgrades-troubleshoot.txt
dzhu: The other way isn't perfect either, but it would be useful to include as an alternative, I think. But anyway I'm happy to get an initial version of this section in and we can iterate on it later.

dzhu (approved):

docs/how-to/installation/upgrades-troubleshoot.txt
stoksc: Yeah, if may be useful as like "if the migration was unintentionally interrupted by the user or failed for some other reason known to be intermittent, try that method, else proceed with a more manual approach.

PR #2494, opened 2021-06-02
chore: fix make -C bindings check

rb-determined-ai:
## Description

Previously `make -C bindings check` combined too many steps into one
make target, and as a result the subshell which checked for differences
in the git repository executed before the code was regenerated.

As a result, there was one new file missing from the repository.

## Test Plan

I manually updated the swagger file and verified that `make -C bindings check` failed.


hamidzr:
@rb-determined-ai is this still applicable?

rb-determined-ai:
@hamidzr nope

PR #2495, opened 2021-06-02
chore: migrate away from spot test assertions [DET-5589]

hkang1:
## Description

After a bit of digging, here's what came up. A typical karma/protractor assertion-based testing and other popular e2e-testing frameworks generally use and rely on assertions to ensure things are in order. With e2e and gauge, we have to change a little bit of how we think about e2e testing. A typical flow might be something like:

* interact with the UI
* check for some assertion to be true

With our setup, the above assertions will occasionally fail. This is because during the actual assertion check, it's possible that there is a timing issue where the UI is still in flight and have not reached the expected state the assertion checks for.

For a concrete example, let's take a look at a test for our dashboard. We have an assertion that checks that upon viewing the dashboard route, there are 7 cards on the board (3 experiments, 2 notebooks and 2 tensorboards). 

```
* Should have "7" recent task cards
```

The implementation for this step looks like this:

```
@Step('Should have <count> recent task cards')
public async checkRecentTasks(count: string) {
  const expectedCount = parseInt(count);
  const cards = await getElements(
    '[class*=TaskCard_base]',
    '[class*=TaskCard_badges] span:first-of-type',
  );
  await assert.strictEqual(cards.length, expectedCount);
}
```

There is a timing issue with this implementation. It is very possible that when gauge runs this step, the page has not fully loaded or rendered the dashboard with all of the expected cards. Instead, a more reliable approach is to use taiko's `waitFor` function to wait for the expected condition to be true. There is a default timeout period of 10s and if it can't resolve in that timeframe, then it will fail.

```
@Step('Should have <count> recent task cards')
public async checkRecentTasks(count: string) {
  await t.waitFor(async () => {
    const expectedCount = parseInt(count);
    const cards = await getElements(
      '[class*=TaskCard_base]',
      '[class*=TaskCard_badges] span:first-of-type',
    );
    return cards.length === expectedCount;
  });
}
```

This approach will continuously check for the expected condition to be true. With the default timeout period of 10s, it gives taiko ample time to ensure that the condition is reached if the expected setup eventually occurs.

## Checklist


EmilyBonar (approved):
Makes sense to me

apizzini (approved):
LGTM.

Out of curiosity, so `await getElements` was completely without logic to await for elements after the page rendered? Was it just checking for elements on the 1st render of the page?

hkang1:
> LGTM.
> 
> Out of curiosity, so `await getElements` was completely without logic to await for elements after the page rendered? Was it just checking for elements on the 1st render of the page?

Good question, and after some investigation, I think I nailed down why this portion was so unpredictable.

What the taiko `elements()` call actually returned was all the matching elements based on a query selector OVER a span of time. I'm not sure how long of a time span. So if you called `t.$('tr').elements()` for a table that only has 4 rows, but the content of the rows changed (like `<Timeago />`), it will actually return a list of elements, but MORE than 4 elements. Because when the content of the element changes, it gets added to the list. So the `getElements` function attempts to address this first problem with a secondary param to help the helper function to remove duplicate elements that are essentially the same but have changed over time.

The second problem was that it grabs elements over a span of time, but that time seemed arbitrary and didn't depend on just one factor. Changing content within any part of the element, and number of render times that gets call, and some time frame, etc. So it's very possible and likely (as we saw with the flakes) that within the span of time that taiko attempts to grab the element, it doesn't actually find them. The `waitFor` with the default `10s` wait seems to have done the job of ensuring whatever UI dependency or expectations are set before moving on with the test.

apizzini:
Thanks for the explanation!

PR #2496, opened 2021-06-02
chore: update eslint-no-unused-vars to handle special cases

hkang1:
## Description

This change will allow unused vars during an object deconstruction. For example, this current causes an issue:
```
const { first, ...rest } = obj;
return rest;
```
This will be an issue if you don't use the `first` variable.

This PR will allow for unused variables in the case of an object deconstruction if you use the spread operator.

## Checklist


EmilyBonar (approved):
LGTM!

PR #2499, opened 2021-06-02
chore: fix TF version detection and RNG usage in test

mackrorysd:
# Description

This came up in @azhou-determined 's testing of making TensorFlow 2 the default. Turns out this test was only ever running against the default, and additional logic is required for it to run on the right version in the TF1/TF2 tests. And .uniform() may have changed since I last tested this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,11 +97,11 @@ commands:
       - when:
           condition: <<parameters.tf1>>
           steps:
-            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-606fd02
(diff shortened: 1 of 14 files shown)


azhou-determined:
you're going to rebase off of master right? looks like your changes are using my personal docker registry test images :D

mackrorysd:
> you're going to rebase off of master right? looks like your changes are using my personal docker registry test images :D

Yes, that's correct. I'm testing that it'll fix your branch and work with TF2 again. I already know that it works with TF1, but yes this is a draft. I'll submit it to master when I know it works, and then you can rebase your branch on that and it should work.

PR #2500, opened 2021-06-03
chore: fix TF version detection and RNG usage in test

mackrorysd:
## Description

This came up in @azhou-determined 's testing of making TensorFlow 2 the default. Turns out this test was only ever running against the default, and additional logic is required for it to run on the right version in the TF1/TF2 tests. And .uniform() may have changed since I last tested this.

## Test Plan

This is covered by CI


azhou-determined:
looks good to me. thanks for the fix.

azhou-determined (approved):

PR #2502, opened 2021-06-03
fix: generating printable master config does not alter original

rb-determined-ai:
## Description

Since expconf, any S3 checkpoint storage configured in the master
which included the access key and access secret would be mangled while
generating a printable config, and experiments generated from that
master would no longer work.

## Test Plan

Added a unit test to ensure that generating a printable config does not alter the original.

### DIFF
--- a/master/internal/config.go
+++ b/master/internal/config.go
@@ -96,7 +96,7 @@ func (c Config) Printable() ([]byte, error) {
 	c.Telemetry.SegmentMasterKey = hiddenValue
 	c.Telemetry.SegmentWebUIKey = hiddenValue
 
-	c.CheckpointStorage.Printable()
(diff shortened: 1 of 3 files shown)


master/internal/config.go
```diff
@@ -96,7 +96,7 @@ func (c Config) Printable() ([]byte, error) {
 	c.Telemetry.SegmentMasterKey = hiddenValue
 	c.Telemetry.SegmentWebUIKey = hiddenValue
 
-	c.CheckpointStorage.Printable()
+	c.CheckpointStorage = c.CheckpointStorage.Printable()
```
stoksc, line 99: ouch..

stoksc (commented):
dang.. lgtm.

stoksc (approved):

PR #2503, opened 2021-06-03
chore: remove scheduler and provision backwards compatibility support [DET-5590]

shiyuann:
## Checklist


armandmcqueen:
Why are we removing backwards compatibility?

It looks like lots of the tests are failing. Are they still relying on the old format?

shiyuann:
@armandmcqueen It's because we want to encourage users to use resource pools rather than one single resource pool and we have announced deprecation twice. 

Some of the tests are relying on the old format. I'll fix those tests.

PR #2504, opened 2021-06-03
chore: update urllib3

mackrorysd:
This was highlighted by dependabot. After looking into why it was pinned the way it was, the previous incompatibilities requiring a very specific range of some of these dependencies appear to no longer be an issue. As long as we're past certain minimum versions, things should work better.

### DIFF
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -43,7 +43,7 @@ setup(
         "requests>=2.22.0",
         # botocore>1.19.0 has stricter urllib3 requirements than boto3, and pip will not reliably
         # resolve it until the --use-feature=2020-resolver behavior in pip 20.3, so we list it here.
-        "urllib3>=1.25.4,<1.26",
+        "urllib3>=1.26.5
         # CLI:
         "argcomplete>=1.9.4",
         "gitpython>=3.1.3",
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -37,12 +37,10 @@ setup(
         "pathspec>=0.6.0",
         "simplejson",
         "termcolor>=1.1.0",
-        # boto3 1.14.11+ has consistent urllib3 requirements which we have to manually resolve.
-        "boto3>=1.14.11",
-        # requests<2.22.0 requires urllib3<1.25, which is incompatible with boto3>=1.14.11
+        # previous releases of boto3 (botocore specifically), urllib3 and requests have had
+        # incompatible interactions that required pinning the versions of these libraries.
+        "boto3>=1.17.87",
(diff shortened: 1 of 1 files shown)


harness/setup.py
```diff
@@ -37,13 +37,11 @@
         "pathspec>=0.6.0",
         "simplejson",
         "termcolor>=1.1.0",
-        # boto3 1.14.11+ has consistent urllib3 requirements which we have to manually resolve.
-        "boto3>=1.14.11",
-        # requests<2.22.0 requires urllib3<1.25, which is incompatible with boto3>=1.14.11
+        # previous releases of boto3 (botocore specifically), urllib3 and requests have had
+        # incompatible interactions that required pinning the versions of these libraries.
+        "boto3>=1.17.87",
```
mackrorysd, line 42: For commentary: as of the previous release of botocore (1.19), they were pushing patches to immediately fix the latest releases of urllib3, so I'm confident the previous issues are resolved - pegging boto to at least the current latest. Not sure if requests needs to be re-thought, but it's a >= anyway. Just updated the comment. Maybe we can remove the pinning at some point.

rb-determined-ai: Can we rewrite these comments from first principles, rather than referring to how it used to be?  So it seems like the decision process here is:

- `urllib3` has some minimum version due to security issues in older versions
- `boto3` has some minimum version due to the `urllib3` minimum version

And given the minimum version of urllib3, do we also need to require a specific version of `requests`?

it looks like `setup.py` in `requests` (tag `v2.22.0) has requirement: `urllib3>=1.21.1,<1.26,!=1.25.0,!=1.25.1`, which is incompatible with `urllib3>=1.26.5`.  I think we need to require `requests>=2.25.0` if we want to _require_ a safe urllib3 version.

Lastly, is it better to require a safe urllib3 version, or just to not disallow it?  i.e. do we care if a user wants to run in an insecure python environment?  I feel it is not our job to police that (though obviously our own environments should be made secure).

mackrorysd: Yeah good point - I honestly didn't like it as I was writing it :D Since they appear to have fixed the rigid requirements and actively work to quickly work with the latest now, I think we can just depend on boto3 and things should transitively work themselves out. I tried that in the latest patch with a clean build and a reset conda environment, and I picked up the required urllib3 version and a recent version of requests.

rb-determined-ai (approved):
Nice.  I think by the time that requests and boto3 disagree again, we can probably start relying on the 2020 resolver, and we probably won't have to deal with this again.

PR #2506, opened 2021-06-03
feat: support .tgz model definitions

rb-determined-ai:
The determined-master has always taken an unorthodox
tarball-over-json-like file format for specifying context directories
(notably for POST experiment, but also for all varieties of commands).

The new format is to post the contents of a gzipped tarball directly.

In all cases, the master supports both the old and the new formats via
different arguments.  This is important because it is fairly common for
users to run old cli's against new masters, and it is only a small
amount of shimming code to give those users an excellent experience.

### DIFF
--- a/master/internal/api_command.go
+++ b/master/internal/api_command.go
@@ -19,12 +19,12 @@ import (
 	"github.com/determined-ai/determined/master/internal/grpcutil"
 	"github.com/determined-ai/determined/master/internal/sproto"
 	"github.com/determined-ai/determined/master/pkg/actor"
+	"github.com/determined-ai/determined/master/pkg/archive"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/pkg/protoutils"
 	"github.com/determined-ai/determined/master/pkg/tasks"
 	"github.com/determined-ai/determined/proto/pkg/apiv1"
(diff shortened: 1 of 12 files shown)


rb-determined-ai:
(the cli and python sdk changes will be part of the PR, but I wanted to check in with this part before doing that part)

stoksc:
@rb-determined-ai what's the plan for this/sending the tgz encoded file?

rb-determined-ai:
@stoksc given the current limits we have with grpc, I think the only sensible thing to do here is to use a base64-encoded json file.

I actually think we should make it a `data:` url (see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)) as @ioga suggested, so that some day we can transparently support multipart-form or application/json encoding.

But in any case, I won't be attempting to land this PR before we cut 0.16.0.  I think no matter what this PR shows that it's pretty darn easy to support either upload format, just as different named parameters.  So if it can be 100% backwards compatible with little effort, there is no reason to push hard to land it with the python sdk.

stoksc:
cool, well the PR lgtm @rb-determined-ai, minus renaming to indicate its a b64 encoded tgz / decoding it, if you wanted to land it.

rb-determined-ai:
closing this until further notice.  It's still too low-priority to get done right now.

PR #2508, opened 2021-06-04
feat: support nested HPs [DET-4785]

liamcli:
## Description
This PR modifies the experiment config for the `hyperparameters` section to support nested hps.  Changes in this PR are as follows:
- Update marshaling and unmarshaling of HyperparametersV0 to be compatible with the new structure.  
- Update grid and PBT searchers to work with nested hps.  
- Add tests for new nested structure.

## Test Plan
- [x] Tests for nested hps and impacted searchers all pass.
- [x] Confirm through logs that the behavior is as expected when using nested hps.

## Commentary (optional)
WebUI support for nested hps will need to implemented separately.  

## Checklist


master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -12,7 +13,117 @@ const GlobalBatchSize = "global_batch_size"
...
+		case HyperparametersV0:
+			flattenHPs(tHP, prefix+key+".", target)
+		default:
+			target[prefix+key] = element
+		}
+	}
+}
+
+// UnflattenHPs undos the FlattenHPs function to restore the nested structure.
+func UnflattenHPs(h HyperparametersV0) HyperparametersV0 {
```
rb-determined-ai, line 39: This seems awfully dangerous, because after mangling a user namespace (hyperparameter key names) we are now attempting to unmangle it and the system will try to operate on the result.

Mangling it for e.g. display purposes (in the webui maybe) is reasonable, but the mangled version should only ever be used for cosmetic things.

schemas/expconf/v0/hyperparameters.json
```diff
@@ -18,6 +18,13 @@
         }
     },
     "additionalProperties": {
-        "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameter.json"
+        "anyOf": [
+            {
+                "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameter.json"
+            },
+            {
+                "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameters.json"
```
rb-determined-ai, line 26: This will _almost_ never be valid, because it looks to me like each nesting would have to have a `global_batch_size` for this to be true.

I think the recursion needs to be added inside of `hyperparameter.json`

Fun fact: I originally planned on landing nested hp as part of expconf when I started.

Other fun fact, I just realized that I _forgot to fully remove nested hp from expconf_.  You can see that `hyperparameter.json` _already has a recursion_.

If you want to also allow lists in nested hp (which seems only logical to me), you can do that with the following additional diff to `hyperparameter.json`:

```diff
             {
                 "unionKey": "const:type=categorical",
                 "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameter-categorical.json"
             },
+            {
+                "unionKey": "type:array",
+                "type": "array",
+                "items": {
+                    "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameter.json"
+                }
+            },
             {
                 "unionKey": "always",
                 "type": "object",
                 "checks": {
                     "if a hyperparameter object's [\"type\"] is set, it must be one of \"int\", \"double\", \"log\", const\", or \"categorical\"": {
                         "properties": {
                             "type": false
                         }
                     }
                 },
                 "additionalProperties": {
                     "$ref": "http://determined.ai/schemas/expconf/v0/hyperparameter.json"
                 }
             },
             {
                 "unionKey": "never",
                 "not": {
-                    "type": "object"
+                    "type": [
+                        "object",
+                        "array"
+                    ]
                 }
             }
```

(that part I wrote a long time ago and did remember to remove)

rb-determined-ai (commented):
Please also include these three test cases I wrote for `schemas/test_cases/v0/hyperparameters.yaml`:

```yaml
- name: implicit const hyperparameter (valid, implicit)
  sane_as:
    - http://determined.ai/schemas/expconf/v0/hyperparameter.json
  case:
    nest:
      const_implicit_empty:
      const_implicit_null: null
      const_implicit_bool: true
      const_implicit_int: 1234
      const_implicit_float: 1234.5678
      const_implicit_list:
        - 1
        - fish
        - 2
        - fish
      const_implicit_dict:
        red: fish
        blue: fish

- name: list, const hyperparameter (valid, explicit)
  sane_as:
    - http://determined.ai/schemas/expconf/v0/hyperparameter.json
  case:
    - type: const
      val: asdf
    - type: const
      val: [1, "fish", 2, "fish"]
    - type: const
      val: {"red": "fish", "blue": "fish"}

- name: nested hyperparameters (invalid, to show paths in errors)
  sanity_errors:
    http://determined.ai/schemas/expconf/v0/hyperparameter.json:
      - "<config>\\[0\\]\\[0\\]:"
      - "<config>\\[0\\]\\[1\\]:"
      - "<config>\\[1\\].one.a:"
      - "<config>\\[1\\].one.b\\[0\\]:"
      - "<config>\\[1\\].one.b\\[1\\]:"
  case:
    - - type: categorical
      - type: categorical
    - one:
        a:
          type: categorical
        b:
          - type: categorical
          - type: categorical
```

(this is another thing that got removed when I decided to not land nested hp as part of expconf)

master/pkg/schemas/expconf/hyperparameters_config.go
stoksc: FWIW, we did run into this exactly at some point in the master config, viper mangles configs exactly the same as is done here and so `.` in a podspec was unacceptable (even though it's in keys frequently). You could just use a more unique pattern to mangle (I changed viper to use `..`), but prob not mangling at all is better.

liamcli: Thanks.  I removed `UnflattenHPs` but still use an internal method `unflattenSample` for generating the hp grid in grid search.  I did adopt `..` as the separator for more uniqueness.

master/pkg/schemas/expconf/experiment_config_test.go
```diff
@@ -10,6 +10,159 @@ import (
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			jByts, err := schemas.JSONFromYaml(tc.Bytes)
+			assert.NilError(t, err)
+			config, err := ParseAnyExperimentConfigYAML(jByts)
+			assert.NilError(t, err)
+			err = schemas.IsComplete(config)
```
rb-determined-ai, line 161: Can you just add nested hp to the 'full experiment' test in `schemas/test_cases/v0/experiment.yaml`?

All this does is unmarshal and check IsComplete, but it only does it for one language.  Putting the test in `schemas/test_cases` will ensure that when the python sdk supports ExperimentConfig objects that we have all the same test coverage on both languages.

master/pkg/schemas/expconf/experiment_config_test.go
```diff
@@ -10,6 +10,159 @@ import (
...
+type ExpConfigTestCase struct {
+	Name     string
+	Bytes    []byte
+	Expected ExperimentConfig
+}
+
+func TestExpConfig(t *testing.T) {
+	testCases := []ExpConfigTestCase{
+		{
+			Name: "Test experimnet config with nested hps",
```
rb-determined-ai, line 22: spelling

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,6 +47,7 @@ type HyperparameterV0 struct {
 	RawDoubleHyperparameter      *DoubleHyperparameterV0      `union:"type,double" json:"-"`
 	RawLogHyperparameter         *LogHyperparameterV0         `union:"type,log" json:"-"`
 	RawCategoricalHyperparameter *CategoricalHyperparameterV0 `union:"type,categorical" json:"-"`
+	RawNestedHyperparameter      *map[string]HyperparameterV0 `union:"type,object" json:"-"`
```
rb-determined-ai, line 50: The `union:"type,object"` thing here would indicate that it was possible to have a map of hyperparameters like:
```yaml
hyperparameters:
  global_batch_size: 8
  nest:
    type: object
    value:
      a: 1
      b: 2
```

It might not be bad to allow that, but right now we don't.

Using the union package this way is a bit of a hack though, since you are including the union tag to benefit from some behaviors but you are manually avoiding other behaviors of the union tag (marshaling and unmarshaling).

I am actually ok with the way it is but it would be nice to include a comment above `RawNestedHyperparameter` that explains why you are assigning it a `union` tag even though you never want it to marshal or unmarshal with the usual union logic.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -37,14 +63,62 @@ func (h *HyperparameterV0) UnmarshalJSON(data []byte) error {
...
+	}
+	return output, nil
+}
+
 // MarshalJSON implements the json.Marshaler interface.
 func (h HyperparameterV0) MarshalJSON() ([]byte, error) {
+	if h.RawNestedHyperparameter != nil {
+		hps, err := hpsToMap(h)
+		if err != nil {
+			return []byte("null"), err
```
rb-determined-ai, line 118: I think it is preferable to return `nil, err` rather than `<valid json>, err`, for readability reasons.  I had a wtf moment where I thought you were dropping the error silently and marshaling everything as `"null"`, before looking closer.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -37,14 +63,62 @@ func (h *HyperparameterV0) UnmarshalJSON(data []byte) error {
...
+			output[key] = unionHP
+		}
+	}
+	return output, nil
+}
+
 // MarshalJSON implements the json.Marshaler interface.
 func (h HyperparameterV0) MarshalJSON() ([]byte, error) {
+	if h.RawNestedHyperparameter != nil {
+		hps, err := hpsToMap(h)
```
rb-determined-ai, line 116: I don't understand what `hpsToMap` is for.  Can you just `return json.Marshal(*h.RawNestedHyperparameter)` here?

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -37,14 +63,62 @@ func (h *HyperparameterV0) UnmarshalJSON(data []byte) error {
 		return err
 	}
 	if _, ok := parsed.(map[string]interface{}); ok {
+		parsedMap := parsed.(map[string]interface{})
```
rb-determined-ai, line 66: try:
```go
if tParsed, ok := parsed.(map[string]interface{}); ok {
    hpType, hasType := parsedMap["type"]
```
the `if _, ok` from before was dropping exactly the value you are looking for.

rb-determined-ai (commented):
let's chat offline about the desired unmarshaling behavior and how we can make it happen.  You're not too far off right now.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,6 +47,7 @@ type HyperparameterV0 struct {
 	RawDoubleHyperparameter      *DoubleHyperparameterV0      `union:"type,double" json:"-"`
 	RawLogHyperparameter         *LogHyperparameterV0         `union:"type,log" json:"-"`
 	RawCategoricalHyperparameter *CategoricalHyperparameterV0 `union:"type,categorical" json:"-"`
+	RawNestedHyperparameter      *map[string]HyperparameterV0 `union:"type,object" json:"-"`
 }
 
 // Merge prevents recursive merging of hyperparameters.
```
rb-determined-ai, line 53: I actually think it might be reasonable to allow the NestedHyperparameter should be merged recursively.  What do you think?

master/pkg/searcher/grid.go
```diff
@@ -106,36 +106,94 @@ func (s *gridSearch) trialClosed(ctx context, _ model.RequestID) ([]Operation, e
...
+	return samples
+}
+
+// Returns gridded values for a single hyperparameter.
+func getHPValueSet(h expconf.Hyperparameter) [][]interface{} {
+	var valueSets [][]interface{}
+	if h.RawNestedHyperparameter != nil {
+		for _, nestedHP := range *h.RawNestedHyperparameter {
+			valueSets = append(valueSets, getHPValueSet(nestedHP)...)
+		}
```
rb-determined-ai, line 124: is it a problem that this ordering is non-deterministic?

I *think* so, because it seems like the recursion is happening in two separate places, but I can't tell.

Either way, probably sorting the keys is a good idea to make the code safe for future reuse.

master/pkg/schemas/expconf/hyperparameters_config_test.go
```diff
@@ -0,0 +1,30 @@
+package expconf
+
+import (
+	"encoding/json"
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestNestedHPs(t *testing.T) {
```
rb-determined-ai, line 10: it's not clear to me that this test tests anything not already tested in `schema_test.go`, where we also do round-trip-through-json testing.  If this test is in fact redundant, we should remove it.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,6 +47,8 @@ type HyperparameterV0 struct {
 	RawDoubleHyperparameter      *DoubleHyperparameterV0      `union:"type,double" json:"-"`
 	RawLogHyperparameter         *LogHyperparameterV0         `union:"type,log" json:"-"`
 	RawCategoricalHyperparameter *CategoricalHyperparameterV0 `union:"type,categorical" json:"-"`
+	// This option does not actually go through union marshaling and unmarshaling logic.
```
rb-determined-ai, line 50: Don't be afraid to leave more detail!

If I walked up to this code without any background understanding, I would not know right away "why did they bring up this detail at all?"

Can you go into more detail about "we're using the `union` tag due to XYZ, even though we specifically avoid the `union` behavior for marshaling/unmarshaling this field"

rb-determined-ai (commented):
looking very good

master/pkg/searcher/grid.go
liamcli: We won't have issues with nondeterministic ordering because we store the `valueSet` index for a given hyperparameter in `hpToInd` and use that to map values to a nested `hparamSample`.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -36,7 +62,25 @@ func (h *HyperparameterV0) UnmarshalJSON(data []byte) error {
...
+			}
+			for key := range parsedMap {
+				bytes, _ := json.Marshal(parsedMap[key])
+				var hp HyperparameterV0
+				err := json.Unmarshal(bytes, &hp)
+				if err != nil {
+					return err
+				}
+				nestedHPs[key] = hp
+			}
```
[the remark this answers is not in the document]
liamcli: this part is from outdated commit

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,11 +47,48 @@ type HyperparameterV0 struct {
...
 
-// Merge prevents recursive merging of hyperparameters.
+// Merge prevents recursive merging of hyperparameters unless h
+// is a nested hyperparameter.  When h is a nested hyperparameter,
+// we merge fields of h and other into a single map.
 func (h HyperparameterV0) Merge(other interface{}) interface{} {
-	// Never merge partial hyperparameters.
+	// Only merge nested hyperparameters.
+	if h.RawNestedHyperparameter != nil {
+		target := *h.RawNestedHyperparameter
```
rb-determined-ai, line 77: Let's return a new HyperparameterV0 with a RawNestedHyperparameter which is totally new map.

Ultimately, `z = x.Merge(y)` should not alter either `x` or `y`.

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,11 +47,48 @@ type HyperparameterV0 struct {
...
+		source := *other.(HyperparameterV0).RawNestedHyperparameter
```
rb-determined-ai, line 78: I don't think it's safe to assume that the other hyperparameter has a RawNestedHyperparameter field; you should probably be checking that.  Probably you can include that in the top-level if statement.

(however it _is_ safe to assume that `other interface{}` is actually a `HyperparameterV0`, because merging differently-typed objects would be insane and is not allowed by `schemas.Merge()`)

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,11 +47,48 @@ type HyperparameterV0 struct {
...
+		for key, val := range target {
+			_, inSource := source[key]
```
rb-determined-ai, line 80: capture the value of `source[key]` here rather than look it up again in the if statement

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,11 +47,48 @@ type HyperparameterV0 struct {
...
+			if inSource {
+				target[key] = val.Merge(source[key]).(HyperparameterV0)
+			}
+		}
+		for key, val := range source {
+			_, inTarget := target[key]
```
rb-determined-ai, line 86: ditto on capturing the value

master/pkg/schemas/expconf/hyperparameters_config.go
```diff
@@ -22,11 +47,53 @@ type HyperparameterV0 struct {
...
+			}
+		}
+		for key, val := range source {
+			if _, inTarget := target[key]; !inTarget {
+				newNestedHP[key] = val
+			}
+		}
+		var newHP HyperparameterV0
+		newHP.RawNestedHyperparameter = &newNestedHP
+		return newHP
```
rb-determined-ai, line 95: These three lines could be `return HyperparameterV0{.RawNestedHyperparameter: &newNestedHP}`

rb-determined-ai (approved):
nice work!

PR #2509, opened 2021-06-04
chore: clean up task spec

shiyuann:
## Description

A very boring internal cleanup PR.

## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -113,17 +113,14 @@ func (a *apiServer) DeleteExperiment(
 	}
 
 	agentUserGroup, err := a.m.db.AgentUserGroup(*exp.OwnerID)
-	switch {
-	case err != nil:
+	if err != nil {
 		return nil, errors.Errorf("cannot find user and group for experiment")
-	case agentUserGroup == nil:
-		agentUserGroup = &a.m.config.Security.DefaultTask
 	}
 
 	addr := actor.Addr(fmt.Sprintf("delete-checkpoint-gc-%s", uuid.New().String()))
+	taskSpec := a.m.taskSpecMaker.MakeTaskSpec("", agentUserGroup)
 	if gcErr := a.m.system.MustActorOf(addr, &checkpointGCTask{
-		agentUserGroup:     agentUserGroup,
-		taskSpec:           a.m.taskSpec,
+		taskSpec:           &taskSpec,
 		rm:                 a.m.rm,
 		db:                 a.m.db,
 		experiment:         exp,
@@ -643,7 +640,17 @@ func (a *apiServer) CreateExperiment(
 		detParams.ParentID = &parentID
 	}
 
-	dbExp, validateOnly, taskSpec, err := a.m.parseCreateExperiment(&detParams)
+	user, _, err := grpcutil.GetUser(ctx, a.m.db)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get the user: %s", err)
+	}
+
+	agentUserGroup, err := a.m.db.AgentUserGroup(user.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	dbExp, validateOnly, taskSpec, err := a.m.parseCreateExperiment(&detParams, agentUserGroup)
 
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "invalid experiment: %s", err)
@@ -651,11 +658,6 @@ func (a *apiServer) CreateExperiment(
 
 	if validateOnly {
 		return &apiv1.CreateExperimentResponse{}, nil
-	}
-
-	user, _, err := grpcutil.GetUser(ctx, a.m.db)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get the user: %s", err)
 	}
 
 	dbExp.OwnerID = &user.ID
--- a/master/internal/command/command_manager.go
+++ b/master/internal/command/command_manager.go
@@ -26,9 +26,6 @@ var shellFormEntrypoint = []string{"/bin/sh", "-c"}
 
 type commandManager struct {
 	db *db.PgDB
-
-	defaultAgentUserGroup model.AgentUserGroup
-	makeTaskSpec          tasks.MakeTaskSpecFn
 }
 
 // CommandLaunchRequest describes a request to launch a new command.
@@ -101,16 +98,19 @@ func (c *commandManager) newCommand(params *CommandParams) *command {
 	}
 	setPodSpec(config, params.TaskSpec.TaskContainerDefaults)
 
+	params.TaskSpec.SetInner(&tasks.StartCommand{
+		Config:    *config,
+		UserFiles: params.UserFiles,
+	})
+
 	return &command{
-		taskID:    sproto.NewTaskID(),
-		config:    *params.FullConfig,
-		userFiles: params.UserFiles,
+		taskID: sproto.NewTaskID(),
+		config: *params.FullConfig,
 		owner: commandOwner{
 			ID:       params.User.ID,
 			Username: params.User.Username,
 		},
-		agentUserGroup: params.AgentUserGroup,
-		taskSpec:       params.TaskSpec,
+		taskSpec: params.TaskSpec,
 
 		db: c.db,
 	}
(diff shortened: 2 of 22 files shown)


master/pkg/tasks/task_spec.go
```diff
@@ -131,413 +138,18 @@ func (t *TaskSpec) EnvVars() map[string]string {
...
-	return 0
-}
-
-// ResourcesConfig implements InnerSpec.
-func (s StartTrial) ResourcesConfig() expconf.ResourcesConfig {
-	return s.ExperimentConfig.Resources()
-}
-
 // TaskSpecMaker is used to make task specs.
 type TaskSpecMaker struct {
```
rb-determined-ai: Why is this not just in `core.go`?  It seems like the only `TaskSpecMaker` object is attached to the `master` object.

master/pkg/model/task_container_defaults.go
```diff
@@ -20,20 +19,28 @@ import (
...
+
+	// These can be overridden by the environment.* field in the tasks.
+	Image            *RuntimeItem      `json:"image,omitempty"`
+	RegistryAuth     *types.AuthConfig `json:"registry_auth,omitempty"`
+	ForcePullImage   bool              `json:"force_pull_image,omitempty"`
+	AddCapabilities  []string          `json:"add_capabilities"`
+	DropCapabilities []string          `json:"drop_capabilities"`
+
+	// This can be overridden by the resources.shm_size field configured in the tasks.
+	ShmSizeBytes int64 `json:"shm_size_bytes,omitempty"`
```
rb-determined-ai, line 40: It looks like in the final form of this PR, `ShmSizeBytes` is literally never used.  It should be merged into the config object when we merge other fields into the experiment config, and then we should always read it form there.

master/pkg/tasks/trial.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+
+	return mounts
+}
+
+// UseFluentLogging implements InnerSpec.
+func (s StartTrial) UseFluentLogging() bool { return true }
+
+// UseHostMode implements InnerSpec.
```
rb-determined-ai, line 217: Could we rename `UseHostMode` to be `ForceHostMode`?  That is what it actually means.

master/pkg/model/task_container_defaults.go
rb-determined-ai: Nope!  I messed up my `git grep`'ing.  This is fine.

master/internal/command/command_manager.go
```diff
@@ -26,9 +25,6 @@ var shellFormEntrypoint = []string{"/bin/sh", "-c"}
 
 type commandManager struct {
 	db *db.PgDB
-
-	defaultAgentUserGroup model.AgentUserGroup
```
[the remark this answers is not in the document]
shiyuann: no, it was not used at all

master/pkg/tasks/trial.go
shiyuann: Why do you need to force it? It seems like just using it would be effective

rb-determined-ai: Because the value of `UseHostMode` is only used when it is true.  That is why `ForceHostMode` is a better name for what it actually does.

master/pkg/tasks/task_spec.go
shiyuann: Everything can be moved to `core.go` if you don't split code into packages and everything would be only used in `core.go`. :

master/pkg/model/task_container_defaults.go
```diff
@@ -20,20 +20,31 @@ import (
...
-	GPUPodSpec             *k8sV1.Pod            `json:"gpu_pod_spec"`
-	Image                  *RuntimeItem          `json:"image,omitempty"`
-	RegistryAuth           *types.AuthConfig     `json:"registry_auth,omitempty"`
-	ForcePullImage         bool                  `json:"force_pull_image,omitempty"`
-
-	AddCapabilities  []string      `json:"add_capabilities"`
-	DropCapabilities []string      `json:"drop_capabilities"`
-	Devices          DevicesConfig `json:"devices"`
+
+	// These can be overridden by the environment.pod_spec field in the tasks.
```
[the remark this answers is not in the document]
shiyuann: okay.

master/pkg/tasks/trial.go
shiyuann: This should apply to other fields. like `UseFluentLogging`. It can also be `ForceFluentLogging`.

rb-determined-ai: This is what I'm trying to say:
```go
        // master/pkg/task/task.go::ToContainerSpec()
	network := t.TaskContainerDefaults.NetworkMode
	if t.UseHostMode() {
		network = hostMode
	}
```

That is, `UseHostMode` doesn't mean that we "use host mode".  It means that we _override the host mode setting from the task container defaults_.  If it was called `ForceHostMode()` it would be a better name.

That's not true for `UseFluentLogging()`.  UseFluentLogging literally returns if we should use fluent logging or not:
```
	spec := container.Spec{
		// ...
		RunSpec: container.RunSpec{
			// ...
			UseFluentLogging: t.UseFluentLogging(),
		},
	}

File: master/pkg/tasks/task_spec.go (hunk 1 above)
Comment (rb-determined-ai) [reply master/pkg/tasks/task_spec.go]: But literally the only thing that uses this seems to be the master object...

**Review by rb-determined-ai — APPROVED** [review]
This is much, much nicer than before.

File: master/pkg/model/task_container_defaults.go [hunk 6]
```diff
@@ -20,20 +19,28 @@ import (
 
 // TaskContainerDefaultsConfig configures docker defaults for all containers.
 type TaskContainerDefaultsConfig struct {
+	// These can only be configured in task_container_defaults.
```
stoksc, line 22: ++

master/internal/api_experiment.go
```diff
@@ -610,7 +607,17 @@ func (a *apiServer) CreateExperiment(
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get the user: %s", err)
+	}
+
+	agentUserGroup, err := a.m.db.AgentUserGroup(user.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	dbExp, validateOnly, taskSpec, err := a.m.parseCreateExperiment(&detParams, agentUserGroup)
```
stoksc, line 620: can below be a switch now?

master/internal/command/notebook_manager.go
```diff
@@ -184,31 +181,36 @@ func (n *notebookManager) newNotebook(params *CommandParams) (*command, error) {
...
-			params.AgentUserGroup.OwnedArchiveItem(jupyterDir, nil, 0700, tar.TypeDir),
-			params.AgentUserGroup.OwnedArchiveItem(jupyterConfigDir, nil, 0700, tar.TypeDir),
-			params.AgentUserGroup.OwnedArchiveItem(jupyterDataDir, nil, 0700, tar.TypeDir),
-			params.AgentUserGroup.OwnedArchiveItem(jupyterRuntimeDir, nil, 0700, tar.TypeDir),
-			params.AgentUserGroup.OwnedArchiveItem(
+	taskSpec := params.TaskSpec
+	taskSpec.SetInner(&tasks.StartCommand{
+		Config:    *config,
+		UserFiles: params.UserFiles,
+		AdditionalFiles: archive.Archive{
```
stoksc, line 188: do you think it makes sense to break `StartCommand` into `StartNotebook`, `Start...` and push this into that?

stoksc (commented):
going to leave this for now before I get sidetracked again. just some general q's the refactor looks really good. going to continue reviewing.

master/pkg/tasks/task_spec.go
```diff
@@ -61,26 +45,44 @@ type InnerSpec interface {
...
 }
 
+// SetRuntimeInfo sets the runtime information.
+func (t *TaskSpec) SetRuntimeInfo(taskID string, taskToken string) {
+	t.TaskID = taskID
+	t.TaskToken = taskToken
+}
+
+// SetContainerInfo sets the container information.
+func (t *TaskSpec) SetContainerInfo(containerID string, devices []device.Device) {
```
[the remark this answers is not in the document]
stoksc: I see this is called for each allocation, essentially customizing the task spec for each container (and it's passed by not as a pointer). Hadn't read the otherside of this code when I commented. Still fine but a bit weird in a task spec tbh, now that I think about it.

stoksc (approved):
this is amazing, thanks @shiyuann

PR #2510, opened 2021-06-04
fix: python api create experiment bug [DET-5601]

liamcli:
## Description
Fix how we package up experiment directory for `create_experiment` to correctly handle nested directories


shiyuann (approved):

harness/determined/common/experimental/determined.py
```diff
@@ -15,17 +15,18 @@
 def _path_to_files(path: pathlib.Path) -> List[V1File]:
     files = []
     for item in context.read_context(path)[0]:
-        content = item["content"].decode("utf-8")
-        file = V1File(
-            path=item["path"],
-            type=item["type"],
-            content=content,
```
rb-determined-ai, line 22: I think the if statement should be here, on the `content=` line?  I'm reasonably confident that even though it's called a V1File, you can't omit directories.

liamcli: in my tests, it returns the right structure in the model definition tar without including dir items.  i think it'll create the directories as needed for files with nested paths.  but, if a user expects an empty directory to exist in the model definition then this would not handle correctly.  

i tried passing None for content when it's a directory and also modifying `ContextItem.dict` to return content for all types but both error when trying to create `V1File`.

liamcli: ok, i added the dirtyle to the set of things we return content for in `ContextItem.dict`. it handles empty dirs now.

PR #2513, opened 2021-06-04
chore: allow non-singleton Authentication

rb-determined-ai:
## Description

The authentication.Authentication object was written as a singleton
object for the CLI.  Since the CLI never talked to multiple masters or
wrote to multiple masters, that pattern worked great.

However, the upcoming Python SDK will need more flexibility than that.
Rather than duplicate the existing authentication logic, we can refactor
what exists now to be a bit more flexible so that the Python SDK can
reuse what is already written.

Additionally, this PR introduces an explicit Cert object, and configures all
Python SDK API calls to pass explicit Authentication and Cert objects.

## Test Plan

I am new to the cli authentication logic.  I ran all the following tests manually:

- authentication bootstrapping is automatic
- broken tokens in auth.json get cleaned up automatically
- make sure the fallback to "regular determined user" still works.
- change password for yourself -> token update in file
- change password for other user -> no token update in file
- login works
- logout works
- logout unsets active user
- logout removes token

### DIFF
--- a/e2e_tests/tests/test_users.py
+++ b/e2e_tests/tests/test_users.py
@@ -23,10 +23,13 @@ ADMIN_CREDENTIALS = authentication.Credentials("admin", "")
 
 
 @pytest.fixture(scope="session")  # type: ignore
-def auth() -> authentication.Authentication:
-    authentication.cli_auth = None
-    yield authentication.cli_auth
-    authentication.cli_auth = None
+def clean_auth() -> None:
+    """
+    clean_auth is a session-level fixture that ensures that we run tests with no preconfigured
+    authentication, and that any settings we save during tests are cleaned up afterwards.
+    """
+    authentication.TokenStore.delete_token_cache()
+    yield None
     authentication.TokenStore.delete_token_cache()
 
 
@@ -177,7 +180,7 @@ def extract_id_and_owner_from_exp_list(output: str) -> List[Tuple[int, str]]:
 
 
 @pytest.mark.e2e_cpu  # type: ignore
-def test_logout(auth: authentication.Authentication) -> None:
+def test_logout(clean_auth: None) -> None:
     creds = create_test_user(ADMIN_CREDENTIALS, True)
 
     # Set Determined passwordto something in order to disable auto-login.
@@ -224,7 +227,7 @@ def test_logout(auth: authentication.Authentication) -> None:
 
 
 @pytest.mark.e2e_cpu  # type: ignore
-def test_activate_deactivate(auth: authentication.Authentication) -> None:
+def test_activate_deactivate(clean_auth: None) -> None:
     creds = create_test_user(ADMIN_CREDENTIALS, True)
 
     # Make sure we can log in as the user.
@@ -247,7 +250,7 @@ def test_activate_deactivate(auth: authentication.Authentication) -> None:
 
 
 @pytest.mark.e2e_cpu  # type: ignore
-def test_change_password(auth: authentication.Authentication) -> None:
+def test_change_password(clean_auth: None) -> None:
     # Create a user without a password.
     creds = create_test_user(ADMIN_CREDENTIALS, False)
 
@@ -264,7 +267,7 @@ def test_change_password(auth: authentication.Authentication) -> None:
 
 
 @pytest.mark.e2e_cpu  # type: ignore
-def test_experiment_creation_and_listing(auth: authentication.Authentication) -> None:
+def test_experiment_creation_and_listing(clean_auth: None) -> None:
     # Create 2 users.
     creds1 = create_test_user(ADMIN_CREDENTIALS, True)
 
@@ -389,7 +392,7 @@ def kill_tensorboards(*tensorboard_ids: str) -> None:
 
 
 @pytest.mark.e2e_cpu  # type: ignore
-def test_notebook_creation_and_listing(auth: authentication.Authentication) -> None:
+def test_notebook_creation_and_listing(clean_auth: None) -> None:
     creds1 = create_test_user(ADMIN_CREDENTIALS, True)
     creds2 = create_test_user(ADMIN_CREDENTIALS, True)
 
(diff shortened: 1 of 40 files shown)


rb-determined-ai:
@shiyuann there are 3 commits here, but only the first is really ready for review.

rb-determined-ai:
> @shiyuann there are 3 commits here, but only the first is really ready for review.

I think this PR is ready now.

e2e_tests/tests/test_users.py
```diff
@@ -12,26 +12,26 @@
...
 
 @pytest.fixture(scope="session")  # type: ignore
-def auth() -> Authentication:
-    a = Authentication.instance()
-    a.reset_session()
-    yield a
-    TokenStore.delete_token_cache()
+def auth() -> authentication.Authentication:
+    authentication.cli_auth = None
+    yield authentication.cli_auth
```
shiyuann, line 28: Does this always yield None?

rb-determined-ai: Yes it does.  I can clean that up though.

shiyuann: It's better to clean that up in this PR I think. It looks weird to set it to be None, yield a None, and set it to be None again. Doesn't make straight sense to me.

rb-determined-ai: will do.

harness/determined/common/api/request.py
```diff
@@ -128,8 +67,16 @@ def do_request(
     body: Optional[Dict[str, Any]] = None,
     headers: Optional[Dict[str, str]] = None,
     authenticated: bool = True,
+    auth: Optional[authentication.Authentication] = None,
+    cert: Optional[certs.Cert] = None,
     stream: bool = False,
 ) -> requests.Response:
+    # If no explicit Authentication object was provided, use the cli's singleton Authentication.
+    if auth is None:
+        auth = authentication.cli_auth
```
shiyuann, line 76: This `cli_auth` is actually used by non-cli functions?

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,220 @@
...
+        The name we use to verify the master certificate.
+        """
+        return self._name
+
+
+cli_cert = None  # type: Optional[Cert]
+
+
+class _BrokenCertStore(Exception):
+    pass
```
shiyuann, line 70: Add a doc string to this to explain which would cause this to happen

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,220 @@
...
+        path.unlink()
+        return {}
+
+    return cast(Dict[str, str], store)
+
+
+@contextlib.contextmanager
+def _modifiable_store(path: pathlib.Path) -> Iterator[Dict["str", "str"]]:
+    """
+    Yields the appropriate store that can be modified, and the modified result will be written
```
shiyuann, line 110: can you define a type `CertStore`? When reading this file, I took some brainpower to understand naming.

shiyuann: Also, this docstring isn't helpful to deliver why a store needs to be modified.

```suggestion
    Yields the cert store that exists in the provided path. This cert store instance can be modified on the fly and the modified result will persist in the provided path.
```

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,220 @@
...
+
+    except _BrokenCertStore:
+        path.unlink()
+        return {}
+
+    return cast(Dict[str, str], store)
+
+
+@contextlib.contextmanager
+def _modifiable_store(path: pathlib.Path) -> Iterator[Dict["str", "str"]]:
```
shiyuann, line 108: how does `persistent_store` sound to you. I think this function mainly persists changes to the cert store file.

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,220 @@
...
+        store[url] = cert_pem
+
+
+def delete_cert(path: pathlib.Path, url: str) -> None:
+    with _modifiable_store(path) as store:
+        if url in store:
+            del store[url]
+
+
+def maybe_shim_old_cert_store(
```
shiyuann, line 151: what is this function for?

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # No need for try/finally, because we don't update the file after failures.
+        yield store
+
+        with temp.open("w") as f:
+            json.dump(store, f, indent=4, sort_keys=True)
+        temp.rename(path)
+
+
+def get_cert(path: pathlib.Path, url: str) -> Optional[str]:
```
shiyuann, line 132: Does `get_cert` just return the path to the cert file?

harness/determined/common/api/request.py
rb-determined-ai: Yes, but that should change in the near future.  As far as I know, only the data layer and the profiler currently make API calls through here.

harness/determined/common/api/certs.py
rb-determined-ai: Refactoring this to have the same behavior as the token store; I like that behavior much better.

Also renaming to `api.errors.CorruptCertificateCacheException`, which I think is probably clearer.

harness/determined/common/api/certs.py
rb-determined-ai: Sure, if you like.

harness/determined/common/api/certs.py
rb-determined-ai: Added a docstring:
```python
    """
    maybe_shim_old_cert_store will detect when an old v0 cert store is present and will shim it to
    a v1 cert store.
    """
```

harness/determined/common/api/certs.py
rb-determined-ai: No, it is returning the content of the certificate.  Added a docstring.

harness/determined/common/api/certs.py
rb-determined-ai: sounds fine

harness/determined/common/api/certs.py
```diff
@@ -0,0 +1,228 @@
...
+        with self._persistent_store() as store:
+            store[url] = cert_pem
+
+    def delete_cert(self, url: str) -> None:
+        with self._persistent_store() as store:
+            if url in store:
+                del store[url]
+
+
+def _load_cert_store(path: pathlib.Path) -> Dict[str, str]:
```
shiyuann, line 120: this should return `CertStore` since it is now defined as a separate class?

shiyuann (approved):
Looks good!

harness/determined/common/api/certs.py
rb-determined-ai: Renamed this and attached it as a method to the CertStore object

PR #2514, opened 2021-06-04
feat: CPU-only training backend [DET-5575, DET-5576, DET-5577, DET-5578, DET-5579, DET-5585, DET-5586]

ioga:
## Description

Add support CPU-based execution throughout the product:
- Rename GPU/CPU tasks/resource pools to compute/aux, while retaining backwards compatibility for user-facing configuration and CLI flags. Change `gpus` to `slots` where appropriate.
- CircleCI tests for the new flows. Some of our `distributed` tests have a hard requirement on CUDA (e.g. by virtue of using `torch.cuda` in the code). Added the `gpu_required` pytest mark for these to give an ability to exclude them.
- Handful of changes in horovod treatment to support this setup.
- Get rid of a no longer supported `native_parallel` flow for distributed training in keras.
- Refactor kube message to determined human-readable messages.
- Each CPU agent provides a single slot. No instance partitioning for now. Kube nodes can be partitioned by setting a finer-grained `slotResourceRequests`.

## Test Plan

Added circleci checks for:
- cpu-only gke cluster (master branch and per request on upstream branches), and 
- cpu-only distributed training on 16 c5.2xlarge's (nightly and per-request on upstream branches)
Added unit tests for:
- master config compatibility.

Manual testing: 
- `det deploy gcp`: use `--gpu-num 0`
- `det deploy aws`: use `--compute-agent-instance-type` set to a gpu-less type (e.g. `c5.2xlarge`).
- on-prem: auto slot type.
- kube: set `slotType: cpu` and `slotResourceRequests:` in helm chart appropriately.

## Commentary (optional)

*Breaking Change* REST API: rename `*gpu*` and `*cpu*` fields in `ResourcePool` object to `*compute**` and `*aux*`.
Since the only known consumer of that is WebUI, I believe this is acceptable.

## Checklist


proto/src/determined/resourcepool/v1/resourcepool.proto
```diff
@@ -50,6 +50,16 @@ enum FittingPolicy {
   FITTING_POLICY_KUBERNETES = 3;
 }
 
+// Device type backing the resource pool slots.
+enum SlotType {
```
hamidzr, line 54: could we share this with v1device.Type?

harness/determined/deploy/aws/templates/efs.yaml
```diff
@@ -697,15 +697,15 @@ Resources:
 
             cat << EOF >> /usr/local/determined/etc/master.yaml
             resource_pools:
-              - pool_name: cpu-pool
+              - pool_name: aux-pool
                 max_cpu_containers_per_agent: ${MaxCPUContainersPerAgent}
```
armandmcqueen, line 701: Update this to use the new config field name?

harness/determined/deploy/aws/templates/efs.yaml
```diff
@@ -761,6 +761,7 @@ Resources:
                   ssh_key_name: ${Keypair}
                   tag_key: det-${AWS::StackName}
                   tag_value: det-agent-${AWS::StackName}
+                  cpu_slots: true
```
armandmcqueen, line 764: I think this field naming is confusing. To me it seemed like `cpu_slots=True` is forcing slots to use CPUs, but it looks like the field is actually specifying whether to use CPU slots if GPUs aren't available. IMO it needs a clearer name.

The new field should also be documented.

harness/determined/deploy/aws/templates/efs.yaml
```diff
@@ -734,15 +734,15 @@ Resources:
                     mkdir -p /mnt/efs
                     mount -t efs ${EFS}:/ /mnt/efs
 
-              - pool_name: gpu-pool
+              - pool_name: compute-pool
```
armandmcqueen, line 737: We need to be very careful here. How will this work around upgrading from a cluster that has experiments on `gpu-pool`? Resource pools were never designed to be removed. Does this lead to a crash if an active experiment expects `gpu-pool` to exist?

harness/determined/deploy/aws/templates/efs.yaml
```diff
@@ -682,8 +682,8 @@ Resources:
 
             resource_manager:
               type: agent
-              default_cpu_resource_pool: cpu-pool
-              default_gpu_resource_pool: gpu-pool
+              default_cpu_resource_pool: aux-pool
```
armandmcqueen, line 685: Should we change these to use the new config field names?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -266,11 +266,14 @@ def _filter_named_parameters(self, optimizer: torch.optim.Optimizer) -> List:
...
-            torch.cuda.set_device(self.device)
+            if self.n_gpus > 0:
+                check.gt(self.n_gpus, 0)
+                # We launch a horovod process per GPU. Each process
+                # needs to bind to a unique GPU.
+                self.device = torch.device(hvd.local_rank())
+                torch.cuda.set_device(self.device)
+            else:
+                self.device = torch.device("cpu")
         elif self.n_gpus > 0:
```
armandmcqueen, line 277: Is this multi-GPU, non-horovod codepath still valid?

harness/determined/deploy/gcp/cli.py
```diff
@@ -263,10 +267,12 @@ def handle_up(args: argparse.Namespace) -> None:
                             help="password for master database",
                         ),
                         Arg(
+                            "--max-aux-containers-per-agent",
                             "--max-cpu-containers-per-agent",
-                            type=str,
-                            default=constants.defaults.MAX_CPU_CONTAINERS_PER_AGENT,
-                            help="max CPU containers running for agents in the CPU resource pool",
+                            type=int,
```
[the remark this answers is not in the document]
ioga: sort of: it was `int` in aws, but `str` here, and I've realigned them, making the validation more correct here.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -225,6 +219,10 @@ def compile_model(
                 compile_args.arguments["optimizer"]
             )
 
+        # context.model is Optional[Model]. This assert signals to mypy it can't
+        # be none because we check that in `from_trial`.
+        assert context.model is not None
```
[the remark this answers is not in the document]
ioga: official docs recommend assertion, and we use that style elsewhere in the harness:
https://mypy.readthedocs.io/en/stable/common_issues.html#unexpected-errors-about-none-and-or-optional-types

master/internal/provisioner/agent_setup_test.go
```diff
@@ -45,15 +45,15 @@ echo "#### PRINTING STARTUP SCRIPT END ####"
...
+if $force_gpus
 then
     echo "#### Starting agent with GPUs"
     docker_args+=(--gpus all)
     docker_args+=(-e DET_SLOT_TYPE=gpu)
 else
-    echo "#### Starting agent with only CPUs"
-    docker_args+=(-e DET_SLOT_TYPE=none)
+    echo "#### Starting agent with auto slots"
+    docker_args+=(-e DET_SLOT_TYPE=auto)
```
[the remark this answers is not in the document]
ioga: changed the code to be more explicit here.

master/internal/resourcemanagers/resource_manager_config.go
```diff
@@ -58,35 +65,73 @@ func (a *AgentResourceManagerConfig) UnmarshalJSON(data []byte) error {
...
+
+	if a.DefaultComputeResourcePool == "" {
+		a.DefaultComputeResourcePool = defaultResourcePoolName
 	}
-	if a.DefaultCPUResourcePool == "" {
-		a.DefaultCPUResourcePool = defaultResourcePoolName
+	if a.DefaultAuxResourcePool == "" {
+		a.DefaultAuxResourcePool = defaultResourcePoolName
 	}
+	a.DefaultCPUResourcePool = ""
```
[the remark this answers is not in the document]
ioga: zeroing out deprecated fields, so the Compute/Aux contain the real values, and CPU/GPU aren't used anywhere except here.

e2e_tests/tests/experiment/test_profiling.py
```diff
@@ -13,6 +13,7 @@
 
 
 @pytest.mark.e2e_gpu  # type: ignore
+@pytest.mark.gpu_required  # type: ignore
```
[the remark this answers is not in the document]
ioga: done

harness/determined/deploy/aws/templates/efs.yaml
ioga: changed to `cpu_slots_allowed` and updated the docs.

harness/determined/deploy/aws/templates/efs.yaml
ioga: It doesn't crash, however pending experiments error out. I've added a `breaking change` note.

harness/determined/pytorch/_pytorch_context.py
ioga: yes, it runs on non-distributed single GPU path.

harness/determined/keras/_tf_keras_trial.py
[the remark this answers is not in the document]
armandmcqueen: That adds an unnecessary runtime check to solve a static analysis problem. `typing.cast` is designed for this use - https://mypy.readthedocs.io/en/stable/casts.html

Non-blocking and not particularly important

harness/determined/pytorch/_pytorch_context.py
armandmcqueen: Oh right, `1` is greater than `0` 🤦‍♂️

armandmcqueen (approved):
Looks good. Great work!

harness/determined/keras/_tf_keras_trial.py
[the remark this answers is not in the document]
ioga: ironically, `assert`s were also designed to be optimized away in python when you run `python -O`. it's just that no one actually seems to use that.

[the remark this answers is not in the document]
armandmcqueen: Interesting, I did not know that

ioga:
admin-landing because of incomplete test-e2e upstream tests.

PR #2515, opened 2021-06-05
fix: trial profiling system metric chart ignoring zero [DET-5505]

apizzini:
## Description

Made a fix because the old code here was returning `null` when the value in the object is zero.

BEFORE:
Schermata 2021-06-05 alle 10 00 50

AFTER:
Schermata 2021-06-05 alle 10 00 29

## Checklist


EmilyBonar (approved):
LGTM

stoksc (approved):

hamidzr (approved):

PR #2517, opened 2021-06-07
feat: provide tensorflow 2.5 image [DET-5522]

ioga:
## Description

Add Tensorflow 2.5 & CUDA 11.2 image.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -581,6 +581,9 @@ commands:
         default: GOOGLE_PROJECT_ID
         description: The Google project ID to connect with via the gcloud CLI
         type: env_var_name
+      environment-image:
+        default: determinedai/environments:cuda-11.0-pytorch-1.7-lightning-1.2-tf-2.4-gpu-0.15.0
+        type: string
     steps:
       - set-cluster-id:
           cluster-id: <<parameters.cluster-id>>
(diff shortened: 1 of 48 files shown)


.circleci/config.yml
```diff
@@ -97,11 +97,11 @@ commands:
       - when:
           condition: <<parameters.tf1>>
           steps:
-            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-606fd02
+            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-da845fc
```
mackrorysd, line 100: Looks like this actually includes @shiyuann 's latest changes, as of "add a github template". Looks like there was a change that got reverted to. Let's have him confirm this is intended for release already.

shiyuann: I think `c1b75818ca76628196c5939f80a6ed4912247db1` and `da845fc89f8ff4b03e2018e407ebf5ff545b3067` should both work.

shiyuann: `254d51127cc540b30a2016081e3e5b4bffa8d292` work too

ioga: his change wasn't intended for that release, but it overwrote 0.15.0, so we've reverted it to overwrite 0.15.0 back, and added a harmless template. so all is correct in my opinion: last material change what the one with horovod cpu.

ioga: I want to use the latest so the "hashed" and "versioned" are the same image.

mackrorysd (approved):

ioga:
(admin-merged because this was an upstream branch, and not all tests were requested)

PR #2520, opened 2021-06-07
fix: allow bumpenvs to update nvcr images in helm charts

eecsliu:
## Description
Previously, the ``update-bumpenvs-yaml`` script would not automatically update the nvcr images in ``bumpenvs.yaml``. This changes alters the way we support nvcr images so that they would be able to be automatically updated.

## Test Plan
Tested manually. To verify that it works, launch determined on k8s while specifying a custom image registry (doesn't matter if it's even valid). After determined has launched, find out the name of the configmap containing the master config and run ``kubectl describe configmap <name of the configmap>``. This should output the contents of the master config. Verify that the custom registry has replaced ``determinedai`` in the default environment images.

## Checklist


shiyuann (approved):
why didn't the bumpenv script not catch it?

eecsliu:
My mistake, the description wasn't great. @ioga caught that the ``bumpenvs.yaml`` would not get updated properly and proposed a different way of doing it so that it wouldn't need any modifications to ``update-bumpenvs-yaml.py``.

helm/charts/determined/templates/master-config.yaml
```diff
@@ -55,6 +55,8 @@ data:
       {{- end }}
       {{- end }}
 
+    {{$cpuImage := (split "/" "determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-606fd02")._1}}
```
ioga, line 58: does that need `{{-` and `-}}` to avoid another empty line? lgtm otherwise.

eecsliu: I tested this several times and adding the ``{{-`` or ``-}}`` to that would cause there to be no newline at all. It ends up looking the most correct as is.

ioga:
I may have caused a conflict by merging #2517, please update.

PR #2525, opened 2021-06-08
fix: prevent zoom reset if chart is already zoomed [DET-5514]

apizzini:
## Description

To do this we're comparing the displayed min/max (uPlot.scales) and the available min/max (uPlot.data -> for efficiency we don't scan the whole data but rely on min/max set on each uPlot.series).

This is working only for X zooms because the comparison is not straight forward on Y values: uPlot is adding some margin on top of the plotted series.

To test this out you can fake some flowing-in data by applying those diffs on TimingMetricChart.tsx:
```
diff --git a/webui/react/src/pages/TrialDetails/Profiles/TimingMetricChart.tsx b/webui/react/src/pages/TrialDetails/Profiles/TimingMetricChart.tsx
index e1059ca1..442651bd 100644
--- a/webui/react/src/pages/TrialDetails/Profiles/TimingMetricChart.tsx
+++ b/webui/react/src/pages/TrialDetails/Profiles/TimingMetricChart.tsx
@@ -1,6 +1,5 @@
 import { Alert } from 'antd';
-import React, { useMemo } from 'react';
-import uPlot, { AlignedData } from 'uplot';
+import React, { useEffect, useMemo, useState } from 'react';import uPlot, { AlignedData } from 'uplot';

 import UPlotChart, { Options } from 'components/UPlotChart';
 import {
@@ -15,6 +14,17 @@ export interface Props {
 }

 const TimingMetricChart: React.FC<Props> = ({ timingMetrics }: Props) => {
+  const [ data, setData ] = useState<AlignedData>();
+
+  useEffect(() => {
+    const inData: AlignedData = [ [ 0 ], [ 0 ] ];
+    setInterval(() => {
+      inData[0].push(inData[0].length);
+      inData[1].push(inData[1].length);
+      setData([ ...inData ]);
+    }, 2500);
+  }, []);
+
   const chartData: AlignedData = useMemo(() => {
     return convertMetricsToUplotData(timingMetrics.dataByBatch, timingMetrics.names);
   }, [ timingMetrics ]);
@@ -36,12 +46,12 @@ const TimingMetricChart: React.FC<Props> = ({ timingMetrics }: Props) => {
       scales: { x: { time: false } },
       series: [
         { label: 'Batch' },
-        ...timingMetrics.names.map((name, index) => ({
-          label: name,
+        {
+          label: '000',
           points: { show: false },
-          stroke: glasbeyColor(index),
+          stroke: glasbeyColor(0),
           width: 2,
-        })),
+        },
       ],
       tzDate: ts => uPlot.tzDate(new Date(ts * 1e3), 'Etc/UTC'),
     };
@@ -57,7 +67,7 @@ const TimingMetricChart: React.FC<Props> = ({ timingMetrics }: Props) => {
     );
   }

-  return <UPlotChart data={chartData} options={chartOptions} />;
+  return <UPlotChart data={data} options={chartOptions} />;
 };

 export default TimingMetricChart;
```

## Checklist


webui/react/src/components/UPlotChart.tsx
```diff
@@ -11,6 +11,11 @@ export interface Options extends Omit<uPlot.Options, 'width'> {
   width?: number;
 }
 
+interface SerieMaxMin {
```
EmilyBonar, line 14: Not sure about calling it a Serie, though if we already use that term elsewhere then it's alright. Not sure about a better option either. `SingleSeriesMaxMin`?

webui/react/src/components/UPlotChart.tsx
```diff
@@ -25,6 +30,7 @@ const UPlotChart: React.FC<Props> = forwardRef((
 ) => {
   const [ chart, setChart ] = useState<uPlot>();
   const chartDivRef = useRef<HTMLDivElement>(null);
+  const isZoomed = useRef<boolean>(false);
```
EmilyBonar, line 33: Is there a reason this is a ref instead of just useState?

webui/react/src/components/UPlotChart.tsx
```diff
@@ -44,10 +50,38 @@ const UPlotChart: React.FC<Props> = forwardRef((
...
+          ready: [ (chart: uPlot) => setChart(chart) ],
+          setScale: [ (uPlot: uPlot, scaleKey: string) => {
+            // need to ignore y because uPlot is adding margins to avoid having
+            // series line touching the max top of the chart.
+            if (scaleKey !== 'x') return;
+
+            const currentMax: number|undefined = uPlot.scales[scaleKey]?.max;
+            const currentMin: number|undefined = uPlot.scales[scaleKey]?.min;
+            const scaleSeries = uPlot.series.filter(serie => serie.scale === scaleKey);
+            let max: number|undefined = scaleMaxMin[scaleKey]?.max;
```
EmilyBonar, line 68: Since we've already returned if scaleKey !== 'x', would it make sense to just use scaleMaxMin['x'] directly, or pull 'x' out as a constant?

EmilyBonar (approved):
Overall looks good.

webui/react/src/components/UPlotChart.tsx
apizzini: Yes, to avoid triggering re-renders.

Setting a useState causes a re-render (which might be fine) but is not needed in a case like this.

useRef doesn't :)

webui/react/src/components/UPlotChart.tsx
apizzini: Oh good catch, that should be ScaleMaxMin, and yes, adding a Single works :)

Will change this

webui/react/src/components/UPlotChart.tsx
apizzini: Technically this could work for Y axis too, so keeping this dynamic is fine.

The only thing stopping me from enabling this behaviour on Y axis is that uPlot is adding some "border" between the max value and the end of the chart. I'll try to understand that better and see if we can enable Y checks too.

stoksc (approved):
approved on functionality, checked it out and tested it. only nit was if i zoomed, then unzoomed, it treated the unzoom as zoomed still and i had to double click again to re-render the rest of the data.

PR #2528, opened 2021-06-08
fix: stop parsing notebook config on every edit [DET-5605]

EmilyBonar:
## Description
The full notebook config was being parsed as yaml after every edit, leading to unwanted results such as deleting the number of slots and having null automatically filled in it's place. This fixes that problem by only parsing the config when launching a new notebook and treating it as text beforehand. This PR also adds word wrapping to the editor.

## Checklist

### DIFF
--- a/webui/react/src/components/NotebookModal.tsx
+++ b/webui/react/src/components/NotebookModal.tsx
@@ -210,6 +210,8 @@ const NotebookFullConfig:React.FC<FullConfigProps> = (
               minimap: { enabled: false },
               scrollBeyondLastLine: false,
               selectOnLineNumbers: true,
+              wordWrap: 'on',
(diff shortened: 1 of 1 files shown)


webui/react/src/components/NotebookModal.tsx
```diff
@@ -211,6 +210,8 @@ const NotebookFullConfig:React.FC<FullConfigProps> = (
               minimap: { enabled: false },
               scrollBeyondLastLine: false,
               selectOnLineNumbers: true,
+              wordWrap: 'on',
+              wrappingIndent: 'indent',
```
hamidzr, line 214: suggestion: I'd check that these match our other monaco usages in the ui

hamidzr (approved):
LGTM

PR #2532, opened 2021-06-08
feat: expose a default Determined in det.experimental.client

rb-determined-ai:
## Description

Expose symbols previously found in `det.experimental.determined`, `det.experimental.experiment`, etc, all inside of `det.experimental.client`.

Then introduce a `det.experimental.client.login()`, and `det.experimental.client.create_experiment()`, etc.

### DIFF
--- a/harness/determined/experimental/client.py
+++ b/harness/determined/experimental/client.py
@@ -1,0 +1,154 @@
+import functools
+import pathlib
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from determined.common.experimental.checkpoint import Checkpoint
+from determined.common.experimental.determined import Determined
+from determined.common.experimental.experiment import ExperimentReference
+from determined.common.experimental.model import Model, ModelOrderBy, ModelSortBy
+from determined.common.experimental.trial import TrialReference
+
+_determined = None  # Optional[Determined]
+
+
+def _require_singleton(fn: Callable) -> Callable:
+    @functools.wraps(fn)
+    def _fn(*args: Any, **kwargs: Any) -> Any:
+        global _determined
+        if _determined is None:
+            _determined = Determined()
+        return fn(*args, **kwargs)
+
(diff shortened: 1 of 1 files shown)


harness/determined/experimental/client.py
```diff
@@ -0,0 +1,154 @@
...
+import pathlib
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from determined.common.experimental.checkpoint import Checkpoint
+from determined.common.experimental.determined import Determined
+from determined.common.experimental.experiment import ExperimentReference
+from determined.common.experimental.model import Model, ModelOrderBy, ModelSortBy
+from determined.common.experimental.trial import TrialReference
+
+_determined = None  # Optional[Determined]
```
ioga, line 11:
```suggestion
_determined = None  # type: Optional[Determined]
```

harness/determined/experimental/client.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Create an experiment with config parameters and model directory. The function
+    returns :class:`~determined.experimental.ExperimentReference` of the experiment.
+
+    Arguments:
+        config(string, pathlib.Path, dictionary): experiment config filename (.yaml)
+            or a dict.
+        model_dir(string): directory containing model definition.
+    """
+    assert _determined is not None
```
ioga, line 73: I'd say these asserts are redundant. when `_determined` is in fact `None`, next line will throw NoneType AttributeError, which is just as readable and developer-friendly.

ioga (approved):

harness/determined/experimental/client.py
rb-determined-ai: These asserts are really just for mypy.  They're impossible to reach, because the decorator guarantees they are not hit.

rb-determined-ai:
only webui test failed, landing this now.

PR #2534, opened 2021-06-08
feat: pull tensorboard images from experiment configs

azhou-determined:
Currently Tensorboard starts with default CPU and GPU images defined in const.go. this pulls images from experiment configs (which return defaults if not custom-defined)


master/internal/core.go
```diff
@@ -629,7 +629,6 @@ func (m *Master) Run(ctx context.Context) error {
 			return c.Request().URL.Path == "/debug/prom/metrics"
 		}
 	}
-	m.echo.Use(middleware.GzipWithConfig(gzipConfig))
```
eecsliu, line 632: blocking: Does this still work with tensorboard 1?

eecsliu (commented):
Looks mostly good! There might be some extraneous things that I commented on.

eecsliu (approved):
👩‍🍳😗👌

PR #2537, opened 2021-06-08
chore: collect samples per second for pytorch and keras [DET-5515]

stoksc:
## Description
This change adds "misc" metrics and collects samples_per_second as the first one, so we can display it in the webui again, but not rely on getting it from a step's metrics (since those will go away shortly). It also heavily refactors the profiler code to have a single "batcher" type and a single "batcher thread" type, instead of one specifically for timings/system/misc metrics.

## Test Plan
- [x] run e2es written for profiling,
- [x] run the code and made sure samples per second are collected.

## Commentary (optional)
This code is not yet "always on". I'm going to leave that for a follow up PR / discuss it at ml-ag.

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -4,6 +4,7 @@ import queue
 import threading
 import time
 from datetime import datetime, timedelta, timezone
+from enum import Enum
 from types import TracebackType
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union, cast
 
@@ -13,8 +14,6 @@ import determined as det
 from determined.common import api, check
 from determined.common.api import TrialProfilerMetricsBatch
 
-SYSTEM_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_SYSTEM"
-TIMING_METRIC_TYPE_ENUM = "PROFILER_METRIC_TYPE_TIMING"
 MAX_COLLECTION_SECONDS = 300
 LOG_NAMESPACE = "determined-profiler"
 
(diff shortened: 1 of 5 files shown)


harness/determined/profiler.py
```diff
@@ -581,50 +630,55 @@ def run(self) -> None:
...
-                self.send_queue.put(self.current_batch.convert_to_post_format())
-                self.current_batch.clear()
+                self.send_queue.put(self.current_batch.consume())
                 batch_start_time = time.time()
 
 
-class TimingsBatcherThread(threading.Thread):
+class MetricsBatcherThread(threading.Thread):
     """
     This is a thread that exists solely so that we can batch Timings and ship them to the
```
armandmcqueen, line 683: Out-of-date comment

harness/determined/profiler.py
```diff
@@ -704,11 +758,87 @@ def run(self) -> None:
...
                 batch_start_time = time.time()
 
 
+class MetricBatch:
+    def __init__(self, trial_id: str, agent_id: str) -> None:
+        self.trial_id = trial_id
+        self.agent_id = agent_id
+        self.batch = {}  # type: Dict[Tuple[MetricType, str, str], List[Measurement]]
+
+    def append(
```
armandmcqueen, line 771: nit: I personally wouldn't call this `append` because `append` is a threadsafe operation in Python while this operation is not. Not very important, so feel free to leave it as `append` if you think that's the best name.

harness/determined/profiler.py
```diff
@@ -704,11 +758,87 @@ def run(self) -> None:
...
-                self.current_batch.clear()
+                self.send_queue.put(self.metrics_batch.consume())
                 batch_start_time = time.time()
 
 
+class MetricBatch:
+    def __init__(self, trial_id: str, agent_id: str) -> None:
+        self.trial_id = trial_id
+        self.agent_id = agent_id
+        self.batch = {}  # type: Dict[Tuple[MetricType, str, str], List[Measurement]]
```
armandmcqueen, line 769: This is a much better design!

armandmcqueen (changes requested):
The code looks good, but it's missing the backend changes.

stoksc:
> The code looks good, but it's missing the backend changes.

Which backend changes exactly?

armandmcqueen:
> > The code looks good, but it's missing the backend changes.
> 
> Which backend changes exactly?

The API's ability to accept the new MISC type of metric. Don't you need to add the MISC type [here](https://github.com/determined-ai/determined/blob/master/proto/src/determined/trial/v1/trial.proto#L104)?

armandmcqueen (approved):

PR #2538, opened 2021-06-08
ci: clean up persistent volumes on GKE

mackrorysd:
## Description

The retain policy is already set to "Delete" for the PVs, but if you don't actually remove the pods and claims before destroying the cluster, they don't actually get freed up.

## Test Plan

Ran the K8s tests in an upstream branch - appears the disks got cleaned up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -473,6 +473,9 @@ commands:
       region:
         type: string
     steps:
+      - helm/delete-helm-release:
+          helm-version: v3.2.4
+          release-name: ci
       # Use a run instead of `gke/delete-cluster` because circle CI orbs do not support `when`.
       - run:
           name: Terminate GKE Cluster


.circleci/config.yml
```diff
@@ -473,6 +473,9 @@ commands:
       region:
         type: string
     steps:
+      - helm/delete-helm-release:
+          helm-version: v3.2.4
+          release-name: ci
```
ioga, line 478: does that need `when: always`?

ioga (approved):
Some clusters don't go through the termination steps in circleci (even with `when: always`), which is why we have the terminator. So I believe we also need to add that or PVC deletion to it.

.circleci/config.yml
mackrorysd: Yup!

eecsliu (approved):

PR #2541, opened 2021-06-08
chore: remove swagger-generated python code

rb-determined-ai:
## Description

The swagger-generated code provided some amount of type safety,
but at the cost of:
- 40k lines of code, causing 0.25s of load time
- breaking the tls features the non-swagger api client had solved
- increasing complexity of the build system

This PR deletes the swagger bindings in favor of just using
the existing REST API client.


harness/setup.py
```diff
@@ -57,11 +54,6 @@
         # docker-compose has a requirement not properly propagated with semi-old pip installations;
         # so we expose that requirement here.
         "websocket-client<1",
-        # Swagger-codegen: python requirements
-        "certifi>=2017.4.17",
```
ioga, line 61: wait, don't we still use certifi in https://github.com/determined-ai/determined/blob/d0adc51c96a778d5f786aa334f783fc79290982c/harness/determined/common/api/certs.py#L10

ioga (approved):

harness/setup.py
rb-determined-ai: Yup, I guess we do.  I think it was an unlisted dependency before.  I'll pull that back in explicitly.

PR #2545, opened 2021-06-09
fix: use model instead of schema struct for de-duping

azhou-determined:
there's code that uses a map to de-dup objects, but raw expconf objects contain pointers, which will be inserted into maps as unique objects as long as the object addresses are unique. fix is to use the api wrapper for the raw expconf object which calls appropriate getters and setters for the actual field values.


rb-determined-ai (commented):
wow, nice catch!

Do you know, is it safe to dedupe based on the `container_path`?  What happens if you try to mount two directories to the same path in docker?  I think that may be a better way forwards.

All of the model config objects are going to get deleted fairly soon, so we may as well fix this in a way that will last.

azhou-determined:
> wow, nice catch!
> 
> Do you know, is it safe to dedupe based on the `container_path`? What happens if you try to mount two directories to the same path in docker? I think that may be a better way forwards.
> 
> All of the model config objects are going to get deleted fairly soon, so we may as well fix this in a way that will last.

It's safe to dedup based on container path. I'll make the change.

azhou-determined:
> wow, nice catch!
> 
> Do you know, is it safe to dedupe based on the `container_path`? What happens if you try to mount two directories to the same path in docker? I think that may be a better way forwards.
> 
> All of the model config objects are going to get deleted fairly soon, so we may as well fix this in a way that will last.

changed to dedup based on container path.

master/internal/command/tensorboard_manager.go
```diff
@@ -200,7 +200,7 @@ func (t *tensorboardManager) newTensorBoard(
 				RawHostPath:      c.HostPath(),
 				RawPropagation:   ptrs.StringPtr(model.DefaultSharedFSPropagation),
 			}).(expconf.BindMount)
-			uniqMounts[sharedFSMount] = true
+			uniqMounts[sharedFSMount.RawContainerPath] = model.ToModelBindMount(sharedFSMount)
```
rb-determined-ai, line 203: Can you use `.ContainerPath()` instead/  The `Raw*` API should only be used when you are first parsing an object.

azhou-determined: good point. changed.

rb-determined-ai (approved):

PR #2547, opened 2021-06-09
docs: add python sdk docs

naren-determined:
## Description

Add documentation for Python SDK checkpoint export functionality.

## Checklist


harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -24,6 +24,11 @@ class CheckpointState(enum.Enum):
 
 class Checkpoint(object):
     """
+    .. note::
+
+        The preferred path for checkpoint export APIs is
+        ``determined.experimental.client.get_checkpoint()``.
```
rb-determined-ai, line 30: I think the wording of these "preferred path" docs is unclear to me.

Can we make them say something like:

```
A Checkpoint object is usually obtained from ``determined.experimental.client.get_checkpoint()``.
```

I also don't think that it needs to be a `.. note::` block.

docs/reference/api/client.txt
```diff
@@ -0,0 +1,142 @@
+.. _client:
+
+################################
+ determined.experimental.client
+################################
+
+..
+   automodule: determined.experimental.client
```
rb-determined-ai, line 8: I believe this is just a comment in ReST notation.

docs/reference/api/client.txt
```diff
@@ -0,0 +1,142 @@
...
+   :member-order: bysource
+
+****************
+ ``Get Models``
+****************
+
+.. automodule:: determined.experimental.client
+   :noindex:
+   :members: get_models
+   :member-order: bysource
```
rb-determined-ai, line 79: I believe all of these lines could just be something like:

```rest
************
 ``Client``
************

.. automodule:: determined.experimental.client
   :members:
   :member-order: bysource
```

I don't see a lot of value in exposing each function via a separate automodule section.

docs/reference/api/client.txt
```diff
@@ -0,0 +1,142 @@
...
+
+****************
+ ``Checkpoint``
+****************
+
+.. autoclass:: determined.experimental.Checkpoint
```
rb-determined-ai, line 85: It looks like if we run change each of these lines to:

```rest
.. autoclass:: determined.experimental.client.Checkpoint
```

sphinx will still find the objects, and it will display their names as `determined.experimental.client.Checkpoint`, which seems good

harness/determined/experimental/client.py
```diff
@@ -34,27 +34,37 @@ def login(
...
        ``login(user=...)`` for any of those users without putting your password in your code.
+
+    Arguments:
+        master (string, optional): The URL of the Determined master.
+            If this argument is not specified, the environment variables
+            DET_MASTER and DET_MASTER_ADDR will be checked for the master URL in that order.
+        user (string, optional): The Determined username used for
+            authentication. (default: ``determined``)
+        password (string, optional): The password associated with the Determined
+            user.
```
rb-determined-ai, line 67: can you remove the word `Determined`?  That sounds a bit like it wants the name for the user whose username is `determined`.

harness/determined/experimental/client.py
```diff
@@ -34,27 +34,37 @@ def login(
...
     """
```
rb-determined-ai, line 68: Additional arguments:

```
cert_path (string, optional): A path to a custom PEM-encoded certificate, against which to validate the master.  (default: ``None``)
cert_name (string, optional): The name of the master hostname to use during certificate validation.  Normally this is taken from the master URL, but there may be cases where the master is exposed on multiple networks that this value might need to be overridden.  (default: ``None``)
noverify: disable all TLS verification entirely.  (default: ``False``)
```

docs/reference/api/client.txt
naren-determined: If you do this, then each function won't have it's own place to click on and won't be exposed at the top-level Python API docs under `determined.experimental.client`. See the attached pic.
Screen Shot 2021-06-09 at 1 28 29 PM

rb-determined-ai: That doesn't sound like a big deal to me.

harness/determined/common/experimental/experiment.py
```diff
@@ -51,6 +51,10 @@ def __init__(self, raw: Any):
 
 class ExperimentReference:
     """
+    An ExperimentReference object is usually used in conjunction with
```
rb-determined-ai, line 54: "used in conjunction with" doesn't make much sense, I think you should say 'is usually obtained from' or 'is usually returned by' or something.

docs/how-to/use-trained-models.txt
```diff
@@ -375,6 +375,6 @@ The checkpoint can then be loaded in Python with this code:
  Next Steps
```
rb-determined-ai, line 375: Looks like this file has some old checkpoint export APIs embedded in it, can we update those?

rb-determined-ai (commented):
Other files which seem to have the old checkpoint export APIs:

- examples/computer_vision/cifar10_pytorch_inference/model_def.py
- examples/gan/dcgan_tf_keras/export.py

docs/how-to/use-trained-models.txt
```diff
@@ -55,16 +53,14 @@ order with respect to the specified metric.
 
 .. code:: python
 
-   from determined.experimental import Determined
+   from determined.experimental.client import get_experiment
```
rb-determined-ai, line 56: can you make this `from determined.experimental import client`?

examples/gan/dcgan_tf_keras/export.py
```diff
@@ -25,9 +25,7 @@ def generate_and_plot_images(generator: tf.keras.Sequential, noise_dim: int) ->
 
 
 def export_model(experiment_id: int, master_url: str) -> tf.keras.Model:
```
rb-determined-ai, line 27: You will have to remove `master_url` as a parameter, and invoke `client.login()` somewhere (probably in main())

docs/how-to/use-trained-models.txt
```diff
@@ -115,42 +115,46 @@ selecting checkpoints.
...
+loads it into memory in a Python process, as shown in the following
+snippet.
 
 .. code:: python
 
-   from determined.experimental import Determined
+   from determined.experimental import client
 
-   checkpoint = Determined().get_experiment(id).top_checkpoint()
+   client.login()
```
rb-determined-ai, line 156: can we remove these `client.login()` lines?  They're often not necessary, and when they are necessary it's unlikely that  an empty `login()` call will work.

A better strategy is to point to the overall `client` docs right at the top of the document and let that be enough.

docs/how-to/use-trained-models.txt
```diff
@@ -367,14 +375,14 @@ The checkpoint can then be loaded in Python with this code:
 
 .. code:: python
 
-   from determined.experimental import Checkpoint
+   from determined.experimental.client import Checkpoint
 
    model = Checkpoint.load_from_path("local_checkpoint")
```
rb-determined-ai, line 380: should be `client.Checkpoint`

examples/README.md
```diff
@@ -2,7 +2,7 @@
...
 | Tutorials | <p>[mnist_pytorch](tutorials/mnist_pytorch)<br>[fashion_mnist_tf_keras](tutorials/fashion_mnist_tf_keras)</p> | <p>MNIST<br>Fashion MNIST</p> | <p>PyTorch<br>TensorFlow (tf.keras)</p> |
-| Computer Vision | <p>[cifar10_pytorch](computer_vision/cifar10_pytorch)<br>[fasterrcnn_coco_pytorch](computer_vision/fasterrcnn_coco_pytorch)<br>[mmdetection_pytorch](computer_vision/mmdetection_pytorch)<br>[detr_coco_pytorch](computer_vision/detr_coco_pytorch)<br>[deformabledetr_coco_pytorch](computer_vision/deformabledetr_coco_pytorch)<br>[mnist_estimator](computer_vision/mnist_estimator)<br>[mnist_tf_layers](computer_vision/mnist_tf_layers)<br>[cifar10_tf_keras](computer_vision/cifar10_tf_keras)<br>[iris_tf_keras](computer_vision/iris_tf_keras)<br>[unets_tf_keras](computer_vision/unets_tf_keras)</p> | <p>CIFAR-10<br>Penn-Fudan Dataset<br>COCO<br>COCO<br>COCO<br>MNIST<br>MNIST<br>CIFAR-10<br>Iris Dataset<br>Oxford-IIIT Pet Dataset</p> | <p>PyTorch<br>PyTorch<br>PyTorch<br>PyTorch<br>PyTorch<br>TensorFlow (Estimator API)<br>TensorFlow (tf.layers via Estimator API)<br>TensorFlow (tf.keras)<br>TensorFlow (tf.keras)<br>TensorFlow (tf.keras)</p> |
```
rb-determined-ai, line 5: um... what happened here?

naren-determined: This is me adding the `cifar10_pytorch_inference` example to the README b/c I noticed it wasn't there.

docs/how-to/use-trained-models.txt
```diff
@@ -14,15 +14,16 @@ This guide discusses:
 #. Storing additional user-defined metadata in a checkpoint.
 #. Using the Determined CLI to download checkpoints to disk.
 
+For specific on the checkpoint API, see
```
rb-determined-ai, line 17: Reword?  Maybe something like:

"The Checkpoint Export API is a subset of the features found in ..."

rb-determined-ai (approved):
great work, thank you!

PR #2548, opened 2021-06-09
chore: improve default_load in certs.py

rb-determined-ai:
Previously, you were not able to do:

    client.login(
        master="https://localhost:8443",
        cert_name="self_signed_name",
    )

and still rely on the trust-on-first-use certificate store in
certs.json, because passing any explicit certificate-related args
prevented loading certs.json entirely.

The old logic could be summarized as:

 - If any explicit certificate args are set, use only explicit args.
 - If not, use the path at DET_MASTER_CERT_FILE.
 - If none, try loading from certs.json.

The new logic can be summarized as:

 - If noverify is set explicitly, or via DET_MASTER_CERT_FILE=noverify,
   disable verification entirely.
 - Otherwise, look for certificate contents in the following places:
    - the explicitliy configured path, if one is provided
    - at DET_MASTER_CERT_FILE, if that is set
    - in certs.json
 - Then look for a certificate name in the following places:
    - the explicitly configure name, if one is provided
    - as DET_MASTER_CERT_NAME, if that is set

The result is that the python sdk is a bit more usable when working with
self-signed certificates.

An alternative would be to save the validation hostname in certs.json at
the moment that you trust a certificate.

### DIFF
--- a/harness/determined/common/api/certs.py
+++ b/harness/determined/common/api/certs.py
@@ -183,23 +183,44 @@ def default_load(
     Having all of the default loading logic in one function makes it easy to invoke the same logic
     in both the cli and the python sdk.
     """
-    # Any explicit args causes us to ignore environment variables and defaults.
-    if explicit_path or explicit_cert_name or explicit_noverify:
-        if explicit_path:
-            with open(explicit_path, "r") as f:
-                cert_pem = f.read()  # type: Optional[str]
-        else:
-            cert_pem = None
-        return Cert(cert_pem=cert_pem, noverify=explicit_noverify, name=explicit_cert_name)
 
-    # Let any environment variable for CERT_FILE override the default store.
+    # Explicit noverify is the easiest case to handle.
+    if explicit_noverify:
+        return Cert(noverify=True)
+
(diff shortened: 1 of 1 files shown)


harness/determined/common/api/certs.py
```diff
@@ -183,39 +183,56 @@ def default_load(
...
+        cert_name = os.environ.get("DET_MASTER_CERT_NAME")
+
+    cert_pem = None  # type: Optional[str]
+    if explicit_path:
+        # Certificate contents from explicit args.
+        with open(explicit_path, "r") as f:
+            cert_pem = f.read()
+    elif env_path:
+        # Certificate contents from environment.
+        if os.path.exists(env_path):
```
ioga, line 223: so if `explicit_path` doesn't exist, we'll error out in the `open` call. maybe we should also fail if `env_path` cert doesn't exist, instead of printing a warning? 
it'd be easier on the user if these are just different ways to set a config option, with the same treatment.

ioga (approved):

rb-determined-ai:
After further discussion, we elected to not land this change at all.

PR #2559, opened 2021-06-10
perf: optimizations to query batching fetch profiler metrics

stoksc:
## Description
This change makes some minor optimizations to the query backing the API to fetch profiler metrics. Usage of `array_to_json` and `QueryProto` were removed in favor of some fancy types from `github.com/jackc/pgtype`. This should help serialization costs on both sides, master and postgresql.

## Test Plan
- [x] call the API, make sure the response is unchanged.

## Checklist


armandmcqueen:
What performance metrics did we measure and how did this change affect those metrics?

stoksc:
@armandmcqueen query execution time uses 10x less CPU this way and average execution time goes from 50ms to sub-millisecond. I noticed the release party db CPU spiked to >50% util from this query and suspected it was the `array_to_json` (that was the only this that would do that in the query). This is stricly less work for the master side too, moving one array into another rather than serializing a json array into another.

armandmcqueen (approved):
LGTM! Should we be using `Queryx` rather than `QueryProto` in general?

stoksc:
> Should we be using Queryx rather than QueryProto in general?

I mean.. hmm. Probably we should prefer it. QueryProto queries into a `map[string]interface{}` then marshals JSON bytes, then finally tries to use `protojson` to unmarshal those bytes into the proto.. it's kinda a mess imo.

PR #2560, opened 2021-06-10
fix: dedup BindMounts and Devices on merge

rb-determined-ai:
Previously, forking an experiment with either devices or bind_mounts set
in the task_container_defaults at the master level would add duplicate
entries to the devices or bind_mounts (since the master injected them
into the original experiment, and injected them again after forking the
original experiment).

This does not dedup duplicates which already existed in the config,
since a schemas.Merge() operation should not actually fix an
already-broken user-provided config.  Instead, it should just never
break a user-provided config.

## Test Plan

Added merge tests for these two objects.

### DIFF
--- a/master/pkg/schemas/expconf/experiment_config.go
+++ b/master/pkg/schemas/expconf/experiment_config.go
@@ -223,7 +223,17 @@ func (b BindMountsConfigV0) Merge(other interface{}) interface{} {
 	// Merge by appending.
 	out := BindMountsConfigV0{}
 	out = append(out, b...)
-	out = append(out, tOther...)
+
+	// Prevent duplicate container paths as a result of the merge.
+	paths := map[string]bool{}
+	for _, mount := range b {
+		paths[mount.ContainerPath()] = true
+	}
+	for _, mount := range tOther {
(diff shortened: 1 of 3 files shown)


stoksc (approved):
makes sense/lgtm

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -241,12 +251,22 @@ type BindMountV0 struct {
...
+func (d DevicesConfigV0) Merge(other interface{}) interface{} {
 	tOther := other.(DevicesConfigV0)
 	// Merge by appending.
 	out := DevicesConfigV0{}
-	out = append(out, b...)
-	out = append(out, tOther...)
+	out = append(out, d...)
+
+	// Prevent duplicate container paths as a result of the merge.
+	paths := map[string]bool{}
```
stoksc, line 261: could drop a comment that `DevicesConfigV0.Merge` prefers the receiver, not the param's devices/bind mounts, if there are multiple devices/bind mounts for a single container path.

rb-determined-ai: oops, I blindly landed this without seeing your comment, sorry.

PR #2561, opened 2021-06-10
fix: erroneous warning with PyTorchCallback

rb-determined-ai:
Previously, PyTorchTrial would issue warnings about
on_validation_step_start and on_validation_step_end anytime you provided
any PyTorchCallback, regardless of whether or not you used the
deprecated methods.


stoksc (approved):
lgtm

PR #2564, opened 2021-06-11
fix: don't return dupes from det model list-versions [DET-5640, DET-4248]

stoksc:
## Description
This change fixes up a bad SQL join that resulted in a query that returned a combination of all rows rather than the intended join.

## Test Plan
- [x] ensure the correct response is returned from `det model list-versions`

## Checklist

### DIFF
--- a/e2e_tests/tests/test_system.py
+++ b/e2e_tests/tests/test_system.py
@@ -376,7 +376,6 @@ def test_end_to_end_adaptive() -> None:
 
 @pytest.mark.e2e_cpu  # type: ignore
 def test_model_registry() -> None:
-    # Given a b
     exp_id = exp.run_basic_test(
         conf.fixtures_path("mnist_pytorch/const-pytorch11.yaml"),
         conf.tutorials_path("mnist_pytorch"),
--- a/e2e_tests/tests/test_system.py
+++ b/e2e_tests/tests/test_system.py
@@ -376,6 +376,7 @@ def test_end_to_end_adaptive() -> None:
 
 @pytest.mark.e2e_cpu  # type: ignore
 def test_model_registry() -> None:
+    # Given a b
     exp_id = exp.run_basic_test(
         conf.fixtures_path("mnist_pytorch/const-pytorch11.yaml"),
         conf.tutorials_path("mnist_pytorch"),
@@ -384,6 +385,7 @@ def test_model_registry() -> None:
 
     d = Determined(conf.make_master_url())
 
+    # Create a model and validate twiddling the metadata.
     mnist = d.create_model("mnist", "simple computer vision model")
     assert mnist.metadata == {}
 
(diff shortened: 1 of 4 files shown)


e2e_tests/tests/test_system.py
```diff
@@ -376,6 +376,7 @@ def test_end_to_end_adaptive() -> None:
 
 @pytest.mark.e2e_cpu  # type: ignore
 def test_model_registry() -> None:
+    # Given a b
```
stoksc, line 379:
```suggestion
```

master/static/srv/get_model_versions.sql
```diff
@@ -36,3 +36,5 @@ SELECT
     version AS version,
     creation_time
```
dzhu, line 37: Maybe
```suggestion
    mv.version,
    m.creation_time
```
for clarity.

master/static/srv/get_model_versions.sql
```diff
@@ -36,3 +36,5 @@ SELECT
...
     FROM c, m, mv
+    WHERE m.name = mv.model_name
```
dzhu, line 39: Hmm, isn't this condition always satisfied due to the WHERE clauses on `m` and `mv`?

stoksc: Yeah suppose it is, didn’t notice that

stoksc: I fixed this up by taking a step back and cleaning the whole query up a bit.. curiously I noticed a `modelv1.ModelVersion` should contain the `modelv1.Model` too but doesn't and never has.. I guess it's never used by the list endpoint, maybe just the get.. kinda funky design there. But the query should be overall more correct now. Could revisit the specifics of the protos later.

master/static/srv/get_model_versions.sql
stoksc: Ended up changing `m.creation_time -> mv.creation_time` since the proto docs indicate that is what should be there.

master/static/srv/get_model_versions.sql
```diff
@@ -33,6 +33,7 @@ c AS (
 )
 SELECT
     to_json(c) AS checkpoint,
-    version AS version,
-    creation_time
-    FROM c, m, mv
+    mv.version,
+    mv.creation_time
```
dzhu, line 37: Nice, I actually checked a database and missed that `model_versions` has a `creation_time` too. Although, is it the same thing? I don't know the model registry code at all. Is it creation time of the original model versus that of this specific version? If there is a difference, this one seems at least as reasonable to include, but I wanted to check if this represents a change.

dzhu (approved):
One minor question about the creation time semantics, but looks good!

master/static/srv/get_model_versions.sql
stoksc: `model_versions.creation_time` is the creation time of the model version, model's is model's, and the proto for the endpoint says the creation time on the response is supposed to be the former https://github.com/determined-ai/determined/blob/master/proto/src/determined/model/v1/model.proto#L43-L44

stoksc: >  but I wanted to check if this represents a change.

yeah, this does represent a change, but i think this way is correct (esp since it's what the docs say). fwiw, too, the CLI doesn't even surface this field, it seems.

dzhu: Sounds good!

PR #2565, opened 2021-06-11
chore: add checkpoint_policy: trial_best

stoksc:
## Description
Simple change to add a `trial_best` checkpoint policy.

## Test Plan
- [x] run e2es, will add a test if we agree on this approach

## Checklist


rb-determined-ai (commented):
I vote we implement it in the python version.

stoksc:
will just await push arch

PR #2573, opened 2021-06-14
chore: add PreemptionWatcher as a client to the Preemption API

rb-determined-ai:
## Description

_PreemptionWatcher uses a separate thread and the selectors library to stream from the Preemption API.

## Test Plan

Tested manually.  It will become part of the e2e test automatically when push architecture lands.

### DIFF
--- a/harness/determined/_preemption.py
+++ b/harness/determined/_preemption.py
@@ -1,0 +1,141 @@
+import json
+import selectors
+import socket
+import threading
+from typing import Any, Optional
+
+from determined.common import api
+from determined.common.api import certs
+
+
+class _PreemptionWatcher(threading.Thread):
+    """
+    _PreemptionWatcher connects to the master and asynchronously waits for a preemption signal.
+
+    _PreemptionWatcher.should_preempt() is non-blocking (after the initial contact is made with the
+    master) and returns a bool indicating if a preemption signal has been received yet.
+
+    Example usage:
+
+    .. code:: python
+
+       with _PreemptionWatcher(master_url, trial_id, cert) as p:
+           print("started!")
+           for i in range(10):
+               if p.should_preempt():
+                   print("preempted!")
+                   break
+               print('no preemption yet, waiting...')
+               time.sleep(1)
+           else:
+               print('finished without preemption signal')
+    """
+
+    def __init__(self, master_url: str, trial_id: int, cert: certs.Cert) -> None:
+        self._master_url = master_url
+        self._trial_id = trial_id
+        self._cert = cert
+
+        # We will use a selector loop as a non-blocking way to stream through the requests library.
+        # We use a socketpair for a cross-platform way to wake up a selector loop mid-stream.
+        self._rsock = None  # type: Optional[socket.socket]
+        self._wsock = None  # type: Optional[socket.socket]
+
+        self._should_preempt = None  # type: Optional[bool]
+
+        self._cond = threading.Condition()
+        super().__init__()
+
+    def run(self) -> None:
+        assert self._rsock is not None
+        assert self._wsock is not None
+
+        with api.get(
+            self._master_url,
+            f"/api/v1/trials/{self._trial_id}/signals/preemption",
+            stream=True,
+            cert=self._cert,
+        ) as r:
+            lines = r.iter_lines()
+
+            # The first response is always immediate, so we always block for it.  This is doubly
+            # important because we can only select on the HTTP connection socket _after_ we call
+            # next(lines) at least once.
+            for line in lines:
+                if not line:
+                    # Empty keepalive message.
+                    continue
+                break
+
+            j = json.loads(line)
+
+            with self._cond:
+                self._should_preempt = j["result"]["preempt"]
+                # Wake the main thread in case it was waiting for the main response.
+                self._cond.notify()
+
+            if self._should_preempt:
+                return
+
+            # It may be a long time between the first response and the second.  We want to be
+            # responsive to .close() calls from the main thread, so we block on input from either
+            # the HTTP connection socket or on our own _rsock socket.
+            with selectors.DefaultSelector() as sel:
+                sel.register(r.raw, selectors.EVENT_READ)
+                sel.register(self._rsock, selectors.EVENT_READ)
+
+                while True:
+                    for key, _ in sel.select():
+                        if key.fileobj == self._rsock:
+                            # self._rsock is only written to when we are supposed to close.
+                            return
+
+                        assert key.fileobj == r.raw
+                        # Calling next() is blocking, but it should only block for a trivial amount
+                        # of time; basically only in the case that one message from the preemption
+                        # API got broken into two network packets.
+                        line = next(lines)
+
+                        if not line:
+                            # Empty keepalive message.
+                            continue
+
+                        j = json.loads(line)
+
+                        self._should_preempt = j["result"]["preempt"]
+
+                        if self._should_preempt:
+                            return
+
+    def start(self) -> None:
+        self._rsock, self._wsock = socket.socketpair()
+        super().start()
+
+    def close(self) -> None:
+        if self._wsock is None:
+            return
+        # Sending any message at all will wake up the selector loop.
+        self._wsock.send(b"quit")
+        self.join()
+        self._rsock.close()
+        self._rsock = None
+        self._wsock.close()
+        self._wsock = None
+
+    def __enter__(self) -> "_PreemptionWatcher":
+        self.start()
+        return self
+
+    def __exit__(self, *_: Any) -> None:
+        self.close()
+
+    def should_preempt(self) -> bool:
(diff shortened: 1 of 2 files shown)


harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+                    for key, _ in sel.select():
+                        if key.fileobj == self._rsock:
+                            # self._rsock is only written to when we are supposed to close.
+                            return
+
+                        assert key.fileobj == r.raw
+                        # Calling next() is blocking, but it should only block for a trivial amount
+                        # of time; basically only in the case that one message from the preemption
+                        # API got broken into two network packets.
+                        # r.raw._connection.sock.setblocking(True)
```
rb-determined-ai, line 98: oops, will clean these up.

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+                # Wake the main thread in case it was waiting for the main response.
+                self._cond.notify()
+
+            if self._should_preempt:
+                return
+
+            # It may be a long time between the first response and the second.  We want to be
+            # responsive to .close() calls from the main thread, so we block in input from either
+            # the HTTP connection socket or on our own _rsock socket, over which we can be alerted
+            # to .close() calls.
```
rb-determined-ai, line 83: will reword a bit

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+        self._should_preempt = None  # type: Optional[bool]
+
+        self._cond = threading.Condition()
+        super().__init__()
+
+    def run(self) -> None:
+        assert self._rsock is not None
+        assert self._wsock is not None
+
+        with api.get(
```
ioga, line 53: this may need to run for the entire duration of a trial, which may be days. so we probably should handle network failures in this.

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+            # The first response is always immediate, so we always block for it.  This is doubly
+            # important because we can only select on the HTTP connection socket _after_ we call
+            # next(lines) at least once.
+            for line in lines:
+                if not line:
+                    # Empty keepalive message.
+                    continue
+                break
+
+            j = json.loads(line)
```
ioga, line 70: perhaps under some network failures, the json content may be "partial", and will fail `json.load`, so I'd handle it similarly to a network failure on the top `api.get`.

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+            f"/api/v1/trials/{self._trial_id}/signals/preemption",
+            stream=True,
+            cert=self._cert,
+        ) as r:
+            lines = r.iter_lines()
+
+            # The first response is always immediate, so we always block for it.  This is doubly
+            # important because we can only select on the HTTP connection socket _after_ we call
+            # next(lines) at least once.
+            for line in lines:
```
ioga, line 64: let's say 
- there's still no first response, so we are blocking here.
- `exit()` is called. it'd stop/return this thread after line 86, but if we are stuck here, it'd be stuck "forever"/until the network timeout.

perhaps all the reading in this method should only be done after the selectors are registered.

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+        self._rsock, self._wsock = socket.socketpair()
+        super().start()
+
+    def close(self) -> None:
+        if self._wsock is None:
+            return
+        # Sending any message at all will wake up the selector loop.
+        self._wsock.send(b"quit")
+        self.join()
+        self._rsock = None
```
ioga, line 123: should we close these raw sockets explicitly instead of relying on GC?

harness/determined/_preemption.py
```diff
@@ -0,0 +1,142 @@
...
+                        if key.fileobj == self._rsock:
+                            # self._rsock is only written to when we are supposed to close.
+                            return
+
+                        assert key.fileobj == r.raw
+                        # Calling next() is blocking, but it should only block for a trivial amount
+                        # of time; basically only in the case that one message from the preemption
+                        # API got broken into two network packets.
+                        # r.raw._connection.sock.setblocking(True)
+                        line = next(lines)
```
ioga, line 99: there's a temptation to just read from `r.raw` and not do `iter_lines()` at all. then you'd have at least some degree of control over `timeout`s on these read operations.

ioga (commented):
General comment: why is this not a websocket-based API? 
Then this client could use `websockets` library and work with messages instead of line-separated json over raw sockets. And we could run it in asyncio event loop on a side instead of the `select`s.

harness/determined/_preemption.py
rb-determined-ai: r.raw is a `HTTPConnection` object, but it does have a `fileno()` method, which is why you can pass it to select

harness/determined/_preemption.py
rb-determined-ai: doh.

harness/determined/_preemption.py
rb-determined-ai: A couple things here:

* if `exit()` was called on the main thread, you'd only have to block until the first round-trip from the master was complete, because the master _must_ send the first response immediately.  That would be pretty trivial in practice, unless the master was not running and there was a firewall dropping ports.  In that case, blocking until a network timeout seems quite reasonable to me.
* it would be nice to not block here, even for one roundtrip to the master, but there's a practical consideration here which is that `select` does not work on the `r.raw` object until after streaming the first request from the master.  I didn't delve deep enough into urllib3 to try to figure out why that was.

harness/determined/_preemption.py
rb-determined-ai: Sounds good. Let's work out "do we want to stream ndjson here or use websockets", and the next version of the code will handle network failures.

rb-determined-ai:
Some thoughts on using websockets.

Previously I said, "I don't care if the master implements this API over longpolling vs websockets vs ndjson".  The more I think about it, the more I think that websockets is the right tool here.

These are my reasons:

* The Preemption API at the master is still internal, and could plausibly be internal forever.  The Python SDK should correspond to puclic REST APIs, but the python Generic API that we eventually release doesn't necessarily need to be public.  It's basically guaranteed that the Generic API has sufficient client-side logic that the server-side API endpoints which back the Generic API are not meaningfully useful to end-users.  So I believe we are free to choose the technology that is easiest for us to implement.
* ndjson, as implemented here, requires us to use the requests library (or rewrite a bunch of authentication logic and tls logic around another http request library), but using requests in a non-blocking way was hard, non-standard, and kinda fragile.  For instance, I have actually no idea why the select strategy only works _after_ the first response is streamed.
* longpolling would be a simpler API to document and understand, and simpler master side, and simpler client-side, and easier for customers to make sense of (though I still think that should not be a big concern).  However, that would still have us using `requests` and select would not work (since there would not be a first response)
* websockets: we already have the auth/tls support for websockets in python, and it would not require any tricks with `select` to implement the client side.  We also already have websockets in the master

So to me, websockets looks like the winner in this case.

harness/determined/_preemption.py
rb-determined-ai: Ok, after poking around a bit, it seems highly likely that the reason `select` is not working for before the first response is because the first response out of `iter_lines` arrives in the same network packet as the header of the HTTP response, and has already been recv'd off of the socket.

We've subclassed Requests before to make that happen.  Maybe there is a way here as well.

rb-determined-ai: I dug even deeper into this.

* a `requests.Response` object contains a `urllib3.HTTPResponse` object
* a `urllib3.HTTPResponse` object contains an `http.client.HTTPResponse` object
* the `http.client.HTTPResponse` object contains an `io.BufferedRWPair`-wrapped socket object.
* the `BufferedRWPair` can consume data off of the wire in an opaque way, making it unreliable to use `select` to decide if there `read()` will block
* PEP 3116 indicates that it's ok to just make the underlying filedescriptor non-blocking, in which case the `BufferedRWPair` [will raise `EAGAIN` errors](https://www.python.org/dev/peps/pep-3116/#non-blocking-i-o)
* If we did that, we can basically count on at least one of the `requests`, `urllib3`, or `http.client` objects becoming unusable due to the libraries assuming IO exceptions are non-recoverable.

TL;DR I'm convinced that using `select` like this is an absolutely terrible idea.

P.S: "doesn't this code work as-written?  Why wouldn't we just use this?"  No, this code would hang forever if the master ever sent the second response so quickly that both streamed responses appear in a single `socket.recv()` call from the operating system.

rb-determined-ai:
Rewritten for a long-polling API.  I'll adjust it for when the long-polling API actually exists.

rb-determined-ai:
Closing this here in favor of review in https://github.com/determined-ai/determined/pull/2771 and testing/landing in the feature branch https://github.com/determined-ai/determined/pull/2727

PR #2576, opened 2021-06-15
chore: upgrade the version of Terraform used to deploy docs

dzhu:
## Description

With the release of Terraform 1.0.0, the old version constraint was no
longer satisfied by the latest Terraform image.

## Test Plan

- [x] Publish docs... it seems to have completed successfully, so it's good, right?

### DIFF
--- a/docs/deploy/main.tf
+++ b/docs/deploy/main.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = "~> 0.12"
+  required_version = "~> 1.0"
 
   backend "s3" {
     bucket = "determined-ai-docs-terraform"


stoksc:
@dzhu as far as i remember .12 to 1 has a lot of breaking syntax changes, does this "just work"?

dzhu:
Yeah, I ran it with 1.0.0 locally and it got through to the end without any errors. (There was one deprecation warning about specifying a provider version in a different way; that should be dealt with at some point, but I'd expect that not to have any functional impact.) That said, https://docs.determined.ai hasn't updated yet for me, but I feel like I keep hearing about there being a lag there. Terraform definitely told me it updated a bunch of files, anyway.

stoksc (approved):

PR #2578, opened 2021-06-17
fix: make header loading spinner more prominent[DET-5667]

apizzini:
## Description

BEFORE:

Schermata 2021-06-17 alle 10 59 31

AFTER:

Schermata 2021-06-17 alle 10 59 15

## Checklist


EmilyBonar (approved):
The changes look good, and I think this will be a bit more visible.

PR #2580, opened 2021-06-17
fix: add download model to visible header buttons [DET-5670]

apizzini:
## Description

That's the 3rd button that will be visible in the header only when width of the page is > 1400px.

Note that with Caleb we decided to keep this simple and use css @media queries to position those buttons.

## Checklist


webui/react/src/components/PageHeaderFoldable.tsx
```diff
@@ -38,6 +38,7 @@ const PageHeaderFoldable: React.FC<Props> = (
   if (options && options.length > 0) {
     if (options.length === 1) dropdownClasses.push(css.optionsDropdownOneChild);
     if (options.length === 2) dropdownClasses.push(css.optionsDropdownTwoChild);
+    if (options.length === 3) dropdownClasses.push(css.optionsDropdownThreeChild);
```
EmilyBonar, line 41: It seems like there should be a better way to do this than all these if statements but it's not coming to me at the moment

EmilyBonar (approved):

PR #2581, opened 2021-06-17
build: remove java dep for deploying webui

### LINKED ISSUE #2591: remove Java dependency from building determined.
as a starting place, we can limit the scope to WebUI. This should allow us to deploy WebUI in environments that do not have Java which is typical for most web hosting solutions

hamidzr:
## Test Plan

- make a proto change and make sure ci breaks if the generated code isn't updated. https://github.com/determined-ai/determined/pull/2581/commits/a43d2771dfe0c46443a60acba4d08a1c1d33abfc

## Commentary (optional)

Fixes #2591 

This will slightly change the dev flow.  If you're making a change that results in the swagger spec files to change you'd need to update the generated code for the webui. If you forget the ci would stop and remind you.

I've added the `build-bindings` target  under `webui/react` to do this. But maybe we'd want to move this under the `bindings` module?  any thoughts?

blocks netlify previews

## Checklist


rb-determined-ai (commented):
seems solid to me.

mackrorysd (approved):

PR #2583, opened 2021-06-17
feat: add trial profiler summary API [DET-5597, DET-5425]

stoksc:
## Description
This change adds a summary API for profiler metrics. It collects the average and a signal to noise ratio for the last 10m of metrics included in the summary.

## Test Plan
- [x] call with curl
- [x] add tests in a bit.

## Checklist

### DIFF
--- a/master/internal/db/postgres_trial_profiler_metrics.go
+++ b/master/internal/db/postgres_trial_profiler_metrics.go
@@ -1,9 +1,12 @@
 package db
 
 import (
+	"context"
+	"fmt"
 	"time"
 
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/proto/pkg/trialv1"
@@ -14,9 +17,9 @@ func (db *PgDB) InsertTrialProfilerMetricsBatch(
 	values []float32, batches []int32, timestamps []time.Time, labels []byte,
 ) error {
 	_, err := db.sql.Exec(`
-INSERT INTO trial_profiler_metrics (values, batches, ts, labels)
-VALUES ($1, $2, $3, $4)
-`, values, batches, timestamps, labels)
+INSERT INTO trial_profiler_metrics (values, batches, ts, ts_range, labels)
+VALUES ($1, $2, $3, tstzrange($4, $5, '[]'), $6)
+`, values, batches, timestamps, timestamps[0], timestamps[len(timestamps)-1], labels)
 	return err
 }
 
@@ -54,4 +57,43 @@ OFFSET $2 LIMIT $3`, labelsJSON, offset, limit)
 		pBatches = append(pBatches, pBatch)
 	}
 	return pBatches, nil
+}
+
+const summaryWindowSeconds = 10 * 60
+
+// GetTrialProfilerMetricSummary gets a summary of profiler metrics.
+func (db *PgDB) GetTrialProfilerMetricSummary(
+	ctx context.Context, labels *trialv1.TrialProfilerMetricLabels,
+) (*trialv1.TrialProfilerMetricSummary, error) {
+	labelsJSON, err := protojson.Marshal(labels)
+	if err != nil {
+		return nil, errors.Wrap(err, "marshaling labels")
+	}
+
+	summary := trialv1.TrialProfilerMetricSummary{
+		Labels: labels,
+	}
+
+	if err := db.sql.QueryRowxContext(ctx, fmt.Sprintf(`
+WITH latest AS (
+	SELECT upper(m.ts_range) AS ts
+	FROM trial_profiler_metrics m
+	WHERE m.labels @> $1::jsonb
+	ORDER BY id DESC
+	LIMIT 1
+)
+SELECT
+  avg(q.v) AS avg,
+  -- classic signal-to-noise ratio
+  (avg(q.v) ^ 2.0) / (stddev(q.v) ^ 2.0) > 5.0 AS stable
+FROM (
+  SELECT unnest(m.values) AS v
+  FROM trial_profiler_metrics m, latest
+  WHERE m.labels @> $1::jsonb
+    AND m.ts_range && tstzrange(latest.ts - interval '%d seconds', latest.ts, '[]')
+) q`, summaryWindowSeconds), labelsJSON).Scan(&summary.Average, &summary.Stable); err != nil {
+		return nil, errors.Wrapf(err, "querying summary data for %s", labelsJSON)
+	}
+
+	return &summary, nil
 }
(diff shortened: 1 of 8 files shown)


proto/src/determined/api/v1/api.proto
```diff
@@ -443,6 +443,16 @@ service Determined {
...
     };
   }
+  // Get the summary of trial profiles.
+  rpc GetTrialProfilerSummary(GetTrialProfilerSummaryRequest)
+      returns (GetTrialProfilerSummaryResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/trials/{trial_id}/profiler/summary"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: [ "Unimplemented" ]
```
stoksc, line 453:
```suggestion
      tags: [ "Internal" ]
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -87,6 +87,29 @@ message Trial {
   CheckpointWorkload best_checkpoint = 10;
 }
 
+// TrialProfilerMetricSummary is a summary of a single metric.
+message TrialProfilerMetricSummary {
```
armandmcqueen, line 91: I don't think this needs to be so generic. IMO the response JSON is really convoluted here and the list of summaries are returned by the API are undocumented. I feel like the API would be more usable if we didn't use this generic approach.

The current response as a JSON is

```json
{
    "summary": {
        "metric_summaries": [
            {
                "average": 1234,
                "stable": true,
                "labels": {
                    "trial_id": 1,
                    "name": "samples_per_second",
                    "agent_id": null,
                    "gpu_uuid": null,
                    "metric_type": "PROFILER_METRIC_TYPE_MISC",
                }
            }
        ]
    }
}
```

The same information represented more succinctly:

```json
{
    "average_throughput": 1234,
    "throughput_is_stable": true
}
```

Or if you want a hierarchy

```json
{
    "throughput": {
        "average": 1234,
        "stable": true
    }
}
```

You shouldn't have to iterate through a list to parse the response from this API - the set of fields we return is static, so we should use a well-defined response schema.

master/internal/db/postgres_trial_profiler_metrics.go
```diff
@@ -55,3 +58,42 @@ OFFSET $2 LIMIT $3`, labelsJSON, offset, limit)
...
+	SELECT upper(m.ts_range) AS ts
+	FROM trial_profiler_metrics m
+	WHERE m.labels @> $1::jsonb
+	ORDER BY id DESC
+	LIMIT 1
+)
+SELECT
+  avg(q.v) AS avg,
+  -- classic signal-to-noise ratio
+  (avg(q.v) ^ 2.0) / (stddev(q.v) ^ 2.0) > 5.0 AS stable
```
armandmcqueen, line 88: Why are we using signal-to-noise ratio here? It doesn't seem like a particularly appropriate measure to me. 

Signal-to-noise will return true on series that are obviously monotonically increasing (e.g. `[200, 210, 220, 230, 240, 250, 260, 270]`).

proto/src/determined/trial/v1/trial.proto
```diff
@@ -87,6 +87,29 @@ message Trial {
...
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [ "float", "labels" ] }
```
armandmcqueen, line 93:
```suggestion
    json_schema: { required: [ "average", "labels" ] }
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -87,6 +87,29 @@ message Trial {
...
+  bool stable = 2;
+  // The labels for this series.
+  TrialProfilerMetricLabels labels = 3;
+}
+
+// TrialProfilerSummary is a summary of the information gather about the trial
+// during profiling.
+message TrialProfilerSummary {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [ "trial_id", "name" ] }
```
armandmcqueen, line 107: I don't understand this line. Where are these fields required to exist in the summary?

proto/src/determined/trial/v1/trial.proto
stoksc: Considering there has been some interest in generic summarizations already, I'd rather keep this in place. now adding GPU util/mem summaries in the backend is a ~3 line code change.

master/internal/db/postgres_trial_profiler_metrics.go
stoksc: To always return true on monotonically increasing series the measure would need a concept of time, which it doesn't, it just looks at a chunk of data without ordering.

proto/src/determined/trial/v1/trial.proto
stoksc: this is just old, i'll remove

stoksc:
```suggestion
    json_schema: { required: [ "metric_summaries" ] }
```

proto/src/determined/trial/v1/trial.proto
armandmcqueen: What about keeping the generic summarization object, but using a map instead of a list? e.g.

```json
{
  "throughout": {
        ... The summary object
    }
}
```
Designing this API to return a fixed size list instead of a map doesn't make a ton of sense to me.

stoksc: sure i can take that

master/internal/db/postgres_trial_profiler_metrics.go
stoksc: i didn't think the slope was very important, just the spread of the window. like, we wouldn't want to fire on everything with a clearly defined slope, unless it was big enough to cause a large spread in the distributed, which this would catch

proto/src/determined/trial/v1/trial.proto
stoksc: well, i will but unfortunately because of grpc's rigidity this means we lose the generic nature internally a bit without some not-worth-it hacks.

master/internal/db/postgres_trial_profiler_metrics.go
stoksc: > this boolean is basically trying to detect whether the slope of the timeseries is flat or not

i think this is the root of disagreement, i just wanted/went with a measure of stability, not flatness

proto/src/determined/trial/v1/trial.proto
stoksc: @armandmcqueen after some thought over the weekend, I'm beginning to think trying to show "stable" is just a mistake.. what about just error bars?

PR #2589, opened 2021-06-21
feat: spin up/down Filestore for det deploy gcp

shiyuann:
## Description

1. Spin up Filestore on GCP when running `det deploy gcp up`.
2. Spin down Filestore on GCP when running `det deploy gcp down`.

## Test Plan

Manually tested it.

## Checklist

### DIFF
--- a/harness/determined/deploy/gcp/cli.py
+++ b/harness/determined/deploy/gcp/cli.py
@@ -34,7 +34,7 @@ def validate_scheduler_type() -> Callable:
 
 
 def deploy_gcp(command: str, args: argparse.Namespace) -> None:
-    # Preprocess the local path to store the states. #
+    # Preprocess the local path to store the states.
 
     # Set local state path as our current working directory. This is a no-op
     # when the --local-state-path arg isn't used. We do this because Terraform
@@ -52,20 +52,20 @@ def deploy_gcp(command: str, args: argparse.Namespace) -> None:
     env = os.environ.copy()
     env["TF_DATA_DIR"] = os.path.join(args.local_state_path, "terraform_data")
 
-    # Initialize determined configurations. #
+    # Initialize determined configurations.
     det_configs = {}
     args_dict = vars(args)
     for arg in args_dict:
         if args_dict[arg] is not None:
             det_configs[arg] = args_dict[arg]
 
-    # Handle down subcommand. #
+    # Handle down subcommand.
     if command == "down":
-        gcp.delete(det_configs, env, args.auto_approve)
+        gcp.delete(det_configs, env, args.no_prompt)
         print("Delete Successful")
         return
 
-    # Handle Up subcommand. #
+    # Handle Up subcommand.
     if (args.cpu_env_image and not args.gpu_env_image) or (
         args.gpu_env_image and not args.cpu_env_image
     ):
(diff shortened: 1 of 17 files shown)


harness/determined/deploy/gcp/gcp.py
```diff
@@ -277,7 +277,10 @@ def delete(configs: Dict, env: Dict) -> None:
     command = ["terraform", "destroy"]
 
     command += ["-input=false"]
-    command += ["-auto-approve"]
+
+    if auto_approve:
+        command += ["-auto-approve"]
```
ioga, line 282: maybe we should also respect this setting in `def terraform_apply` line 124 in this file?

harness/determined/deploy/gcp/cli.py
```diff
@@ -44,23 +46,35 @@ def deploy_gcp(command: str, args: argparse.Namespace) -> None:
     if not os.path.exists(args.local_state_path):
         os.makedirs(args.local_state_path)
     os.chdir(args.local_state_path)
+    print("Using local state path:", args.local_state_path)
```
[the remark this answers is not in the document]
shiyuann: It's supposed to be a user-facing message

harness/determined/deploy/gcp/gcp.py
shiyuann: We don't really need an auto-approve flag for this because the use case is that we don't want things to be accidentally deleted. But we don't really need manual approval for requesting any resources.

harness/determined/deploy/gcp/terraform/outputs.tf
```diff
@@ -42,6 +42,14 @@ output "B5--GPUs-per-Agent" {
   value = "                    ${var.gpu_num}\n"
 }
 
+output "C1--GCS-bucket" {
```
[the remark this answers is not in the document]
shiyuann: I think there is no significance. I guess it was just for printing the message in a prettier way

harness/determined/deploy/gcp/gcp.py
ioga: should we then call the user-facing flag a more generic `-y` / `--yes`, so it doesn't look like it's literally used for terraform?

harness/determined/deploy/gcp/terraform/modules/filestore/main.tf
```diff
@@ -0,0 +1,24 @@
...
+// Random integer to use
+resource "random_integer" "naming_int" {
+  min = 100000
+  max = 999999
+}
+
+// Create Filestore instance
+
+resource "google_filestore_instance" "persistence-filestore" {
+  name = "det-filestore-${var.unique_id}-${random_integer.naming_int.result}"
```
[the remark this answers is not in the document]
shiyuann: The reason is to not conflict with existing resources. I am simply following the same pattern as other modules. We should maybe just use the unique id. But we could refactor that across the whole terraform scripts.

harness/determined/deploy/gcp/gcp.py
shiyuann: Will use `no-prompt`

ioga (approved):

PR #2593, opened 2021-06-21
chore: update ZMQ logic in DistributedContext

rb-determined-ai:
## Description

This PR introduce a series of small refactors to support push architecture.

* It fixes a rare hang in the ZMQBroadcastServer/Client.  The hang
  has always been possible but seemingly was not often hit due to the
  ordering of how we ran things.
* It introduces a node-local ZMQBroadcastServer/Client in the
  DistributedContext.  This will be used in push architecture for things
  like coordinating checkpoint downloads.
* It introduces Unix-socket-based ZMQ communication for the node-local
  sever/client, which is slightly faster and less likely to be affected
  by networking issues.
* It adds a test for the DistributedContext, for each of local/global
  broadcast/gather/allgather functions.

## Test Plan

Test added!

### DIFF
--- a/harness/determined/_train_context.py
+++ b/harness/determined/_train_context.py
@@ -3,7 +3,7 @@ import logging
 import shutil
 import socket
 import tempfile
-from typing import Any, Callable, Dict, List, Optional, cast
+from typing import Any, Callable, Dict, List, Optional
 
 import determined as det
 from determined import constants, horovod, ipc
@@ -24,7 +24,29 @@ class _TrainContext(metaclass=abc.ABCMeta):
         self.env = env
         self.hvd_config = hvd_config
         self.rendezvous_info = rendezvous_info
-        self.distributed = DistributedContext(env, hvd_config, rendezvous_info)
+
+        if hvd_config.use:
+            rank_info = RankInfo(
+                rank=horovod.hvd.rank(),
+                size=horovod.hvd.size(),
+                local_rank=horovod.hvd.local_rank(),
+                local_size=horovod.hvd.local_size(),
+                cross_rank=horovod.hvd.cross_rank(),
+                cross_size=horovod.hvd.cross_size(),
+            )
+        else:
+            rank_info = RankInfo(
+                rank=0,
+                size=1,
+                local_rank=0,
+                local_size=1,
+                cross_rank=0,
+                cross_size=1,
+            )
+
+        self.distributed = DistributedContext(
+            rank_info, rendezvous_info, env.det_trial_unique_port_offset
+        )
         self._stop_requested = False
 
     @classmethod
@@ -190,6 +212,49 @@ class NativeContext(_TrainContext):
         self._train_fn = train_fn
 
 
+class RankInfo:
+    def __init__(
+        self,
+        *,
+        rank: int,
+        size: int,
+        local_rank: int,
+        local_size: int,
+        cross_rank: int,
+        cross_size: int,
+    ) -> None:
+        self._rank = rank
+        self._size = size
+        self._local_rank = local_rank
+        self._local_size = local_size
+        self._cross_rank = cross_rank
+        self._cross_size = cross_size
+
+    @property
+    def rank(self) -> int:
+        return self._rank
+
+    @property
+    def size(self) -> int:
+        return self._size
+
+    @property
+    def local_rank(self) -> int:
+        return self._local_rank
+
+    @property
+    def local_size(self) -> int:
+        return self._local_size
+
+    @property
+    def cross_rank(self) -> int:
+        return self._cross_rank
+
+    @property
+    def cross_size(self) -> int:
+        return self._cross_size
+
+
 class DistributedContext:
     """
     DistributedContext extends all TrialContexts and NativeContexts under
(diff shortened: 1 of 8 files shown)


harness/determined/layers/_worker_process.py
```diff
@@ -60,6 +60,11 @@ def __init__(self, broadcast_client: ipc.ZMQBroadcastClient):
         # checks on it.
         self._broadcast_client.send(ipc.ConnectedMessage(process_id=os.getpid()))
 
+        # Avoid hangs when receiving from the broadcast server.  Run this after sending the
+        # ConnectedMessage so that the broadcast server has a pid for running a healthcheck while
+        # this runs.
+        self._broadcast_client.safe_start()
```
shiyuann, line 66: q: is this path only used to synchronize workloads? Do we communicate workloads that are only generated on the local chief? Sorry I forgot we had concluded. Can we include this answer on why we need a broadcast client in the comment?

harness/determined/_train_context.py
```diff
@@ -196,58 +264,147 @@ class DistributedContext:
...
+            return
+
+        # Why does this work?  Why don't multiple agents in training decide on different ports,
+        # causing them to be unable to communicate?
+        #
+        # Answer: unique port offsets are set on by the determined-master on a per-container basis.
+        # Multi-agent dtrain only allows for one-container-per-node, in which case
+        # unique_port_offset is always 0, and every container in training will agree on the same
+        # port.  Single-agent dtrain always within a single container, so the unique_port_offset
+        # may not be zero but all workers (in that same container) will still agree.
```
shiyuann, line 299:
```suggestion
        # unique_port_offset is used for avoiding collision of ports when using distributed training.
        # It should work even if containers use host network. unique_port_offset is set by the master
        # on a per-container basis. With multi-agent dtrain, we run only one container on each node, 
        # in which case unique_port_offset is 0 for all the containers. With single-agent dtrain, we 
        # run only one container, unique_port_offset might not be zero.
```

harness/tests/test_ipc.py
```diff
@@ -158,3 +163,109 @@ def test_zmq_server_client() -> None:
...
+        the results from each rank.
+        """
+        results = [None] * size  # type: List
+        errors = [None] * size  # type: List
+        threads = []
+
+        for cross_rank, local_rank in itertools.product(range(cross_size), range(local_size)):
+            rank = cross_rank * local_size + local_rank
+
+            def _fn(rank: int, cross_rank: int, local_rank: int) -> None:
```
shiyuann, line 194:
```suggestion
            def _run_fn(rank: int, cross_rank: int, local_rank: int) -> None:
```

harness/tests/test_ipc.py
```diff
@@ -158,3 +163,109 @@ def test_zmq_server_client() -> None:
...
+    # Generate one rendezvous_info per node.
+    rendezvous_info = [
+        det.RendezvousInfo(
+            addrs=["localhost:12345"] * cross_size,
+            rank=i,
+        )
+        for i in range(cross_size)
+    ]
+
+    def do_parallel(fn: Callable) -> List:
```
shiyuann, line 182:
```suggestion
    def do_parallel(fn: Callable[[int, int, int], None]) -> List:
```

harness/determined/_train_context.py
```diff
@@ -21,7 +24,29 @@ def __init__(
         self.env = env
         self.hvd_config = hvd_config
         self.rendezvous_info = rendezvous_info
-        self.distributed = DistributedContext(env, hvd_config, rendezvous_info)
+
+        if hvd_config.use:
```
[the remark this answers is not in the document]
rb-determined-ai: No, the creation of a DistribtedContext should always be in the worker process.  And the worker process is responsible for knowing how to figure out the RankInfo from how it was launched.  That is, if the launch layer calls horovodrun, the workers will create the RankInfo from `hvd`; if the launch layer uses torch-native distribution, the workers should check `torch.distributed`, etc.

harness/determined/_train_context.py
```diff
@@ -196,58 +264,147 @@ class DistributedContext:
 
     def __init__(
         self,
-        env: det.EnvContext,
-        hvd_config: horovod.HorovodContext,
+        rank_info: RankInfo,
```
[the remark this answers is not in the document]
rb-determined-ai: No, because RendezvousInfo is a machine-level data structure that is:
* independent of the launch layer
* consumed by the launch layer

while the RankInfo is a worker-level data structure that is:
* dependent on the launch layer
* consumed by the worker processes

harness/determined/layers/_worker_process.py
rb-determined-ai: On tip-of-master, the workload information flows from the master, through each local chief, to each local worker.  There is no such thing as "workloads generated on the local chief".

I'd rather not change the message because honestly, this call and the corresponding comment has already been deleted further downstream (when workloads originate from the chief worker rather than from the master).

harness/tests/test_ipc.py
rb-determined-ai: I think since `_fn` wraps `fn`, it is a clearer name than `_run_fn`

harness/determined/_train_context.py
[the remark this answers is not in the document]
shiyuann: Does the worker processes need to know whether it runs on Horovod?

harness/tests/test_ipc.py
shiyuann: It's weird a `fn` wraps `_fn`. Shouldn't `fn` wrap `_fn`?

harness/determined/_train_context.py
shiyuann: Okay. I am okay with removing this QA. I was simply suggesting to use declarative sentences rather than QA.

shiyuann (approved):
Looks good!

harness/tests/test_ipc.py
rb-determined-ai: no... `fn` is a parameter, and those shouldn't have leading underscore names.  `_fn` wraps the parameter named `fn`.

shiyuann: I am fine with this name since it's only one place.

rb-determined-ai:
@shiyuann I had to include a workaround for the missing `hvd.cross_rank` and `hvd.cross_size` when using `import horovod.tensorflow.keras as hvd`.  See https://github.com/horovod/horovod/pull/3008 for the upstream fix.  We can remove the workaround after the next released version of horovod is sufficiently wide-spread among our users that it won't be disruptive.

rb-determined-ai:
tests passed, landing this now.

PR #2595, opened 2021-06-22
chore: update webui experiment layout for multi-trial and single-trial [DET-5407]

apizzini:
## Description

BEFORE (same for single-trial and multi-trial):

Schermata 2021-06-28 alle 16 25 49

AFTER (single-trial):

Schermata 2021-06-28 alle 16 25 34

AFTER (multi-trial):

Schermata 2021-06-28 alle 16 25 31

## Checklist


webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -65,6 +46,7 @@ const ExperimentDetails: React.FC = () => {
       ]);
       if (!isEqual(experimentData, experiment)) setExperiment(experimentData);
       if (!isEqual(validationHistory, valHistory)) setValHistory(validationHistory);
+      setIsSingleTrial(experimentData?.config.searcher.name === ExperimentSearcherName.Single);
```
EmilyBonar, line 49: chore: Change this to account for experiments with only 1 trial that aren't named single

EmilyBonar (approved):
Looks good, just fix that one thing and we should be good to go

webui/react/src/pages/ExperimentDetails.tsx
apizzini: i rebased locally and forgot to force push, so my push failed..

PR #2596, opened 2021-06-22
fix: support Docker images with EXPOSE in notebooks/shells/tensorboard [DET-5648]

rb-determined-ai:
## Test Plan

Tested manually.  Note that for the old code to fail, `docker inspect CONTAINER_ID | jq '.[].NetworkSettings'` should show the _wrong_ port last.  The ports seem to be sorted alphanumerically, and we typically pick ports in the 3000 range, so any port that starts with a 4 or higher is a good test port.

Example dockerfile:

```dockerfile
FROM docker.io/determinedai/environments:py-3.7-pytorch-1.7-lightning-1.2-tf-2.4-cpu-3a452bc
EXPOSE 8888
```

```bash
docker build . -t junk && det notebook start --config=environment.image=junk
```

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -74,6 +74,7 @@ type command struct {
 	readinessMessageSent bool
 	metadata             map[string]interface{}
 	serviceAddress       *string
+	assignedPort         int
 
 	registeredTime time.Time
 	task           *sproto.AllocateRequest
@@ -206,8 +207,16 @@ func (c *command) Receive(ctx *actor.Context) error {
 		case msg.Container.State == container.Running:
 			c.addresses = msg.ContainerStarted.Addresses
 
-			names := make([]string, 0, len(c.addresses))
+			var names []string
 			for _, address := range c.addresses {
+				// Only proxy the port we expect to proxy.  If a dockerfile uses an EXPOSE command,
+				// additional addresses will appear her, but currently we only proxy one uuid to one
+				// port, so it doesn't make sense to send multiple proxy.Register messages for a
+				// single ServiceID (only the last one would work).
+				if c.assignedPort == 0 || address.ContainerPort != c.assignedPort {
+					continue
+				}
+
 				// We are keying on task ID instead of container ID. Revisit this when we need to
 				// proxy multi-container tasks or when containers are created prior to being
 				// assigned to an agent.
@@ -220,6 +229,13 @@ func (c *command) Receive(ctx *actor.Context) error {
 					ProxyTCP: c.proxyTCP,
 				})
 				names = append(names, string(c.taskID))
+			}
+			if c.assignedPort == 0 && len(names) > 0 {
+				ctx.Log().Error("Expected to not proxy any ports but proxied one anyway")
+			} else if len(names) != 1 {
+				ctx.Log().Errorf(
+					"Expected to proxy exactly 1 port but proxied %v instead", len(names),
+				)
 			}
 			c.proxyNames = names
 			ctx.Tell(c.eventStream, event{
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -74,7 +74,7 @@ type command struct {
 	readinessMessageSent bool
 	metadata             map[string]interface{}
 	serviceAddress       *string
-	assignedPort         int
+	assignedPort         *int
 
 	registeredTime time.Time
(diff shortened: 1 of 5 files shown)


master/internal/command/command.go
```diff
@@ -206,8 +207,16 @@ func (c *command) Receive(ctx *actor.Context) error {
...
 			c.addresses = msg.ContainerStarted.Addresses
 
-			names := make([]string, 0, len(c.addresses))
+			var names []string
 			for _, address := range c.addresses {
+				// Only proxy the port we expect to proxy.  If a dockerfile uses an EXPOSE command,
+				// additional addresses will appear her, but currently we only proxy one uuid to one
+				// port, so it doesn't make sense to send multiple proxy.Register messages for a
+				// single ServiceID (only the last one would work).
+				if c.assignedPort == 0 || address.ContainerPort != c.assignedPort {
```
stoksc, line 216: this pattern of retrieving the ports we wanted mapped from the container addresses is the exact one used in rendezvous. with trying to bring commands and trials closer, what do you think about extracting this into a helper
`AddressesInRange(start, end int, addresses ...[]sproto.Address)`. since it would conflict my rendezvous work, there is no reason to do it here. but maybe drop a TODO and ticket so it's not forgotten.

master/internal/command/command.go
```diff
@@ -221,6 +230,13 @@ func (c *command) Receive(ctx *actor.Context) error {
 				})
 				names = append(names, string(c.taskID))
 			}
+			if c.assignedPort == 0 && len(names) > 0 {
+				ctx.Log().Error("Expected to not proxy any ports but proxied one anyway")
```
stoksc, line 234:
```suggestion
				ctx.Log().Error("expected to not proxy any ports but proxied one anyway")
```

master/internal/command/command.go
```diff
@@ -221,6 +230,13 @@ func (c *command) Receive(ctx *actor.Context) error {
...
+			} else if len(names) != 1 {
+				ctx.Log().Errorf(
+					"Expected to proxy exactly 1 port but proxied %v instead", len(names),
```
stoksc, line 237:
```suggestion
					"expected to proxy exactly 1 port but proxied %v instead", len(names),
```

master/internal/command/command.go
```diff
@@ -74,6 +74,7 @@ type command struct {
 	readinessMessageSent bool
 	metadata             map[string]interface{}
 	serviceAddress       *string
+	assignedPort         int
```
stoksc, line 77: since 0 port can sometimes mean "auto assign a port" in port world, can we use the less ambiguous *int for the port.

stoksc (approved):
lgtm. i left some minor feedback i'd appreciate if you consider.

master/internal/command/command.go
rb-determined-ai: done

master/internal/command/command.go
rb-determined-ai: https://determinedai.atlassian.net/browse/DET-5682

master/internal/command/command.go
```diff
@@ -230,11 +232,11 @@ func (c *command) Receive(ctx *actor.Context) error {
 				})
 				names = append(names, string(c.taskID))
 			}
-			if c.assignedPort == 0 && len(names) > 0 {
-				ctx.Log().Error("Expected to not proxy any ports but proxied one anyway")
+			if c.assignedPort == nil && len(names) > 0 {
+				ctx.Log().Error("expected to not proxy any ports but proxied one anyway")
 			} else if len(names) != 1 {
 				ctx.Log().Errorf(
```
stoksc, line 238: this is ancillary to this pr, but it seems both of these cases should crash the actor to me.

rb-determined-ai: I think it's better to fail a bit more gracefully.  For example, if we had been that drastic before, then the customer who was hit this originally would have had absolutely no recourse and would not have been able to continue using their notebook by bypassing the master proxy.

It's not like the user isn't going to know that the notebook failed; that will still be obvious.

Though I do agree it's a bad error condition that I would definitely want to see when I asked a customer for the master logs.

stoksc: yeah.. degrading in unanticipated ways just always seems dangerous (scary) to me, but I can see that case at least.

PR #2606, opened 2021-06-22
chore: add loading state profile metrics

### LINKED ISSUE #2605: "no data available" while the profiler is loading is confusing
> 1. while it’s still loading, it says “No data available.“. Probably should show a spinner?

when looking at the profiler the ui shows a "no data available" message depending on how we're interacting with the backend here we should be able to show a loading ui/message instead of a definite no data available.

hamidzr:
## Commentary (optional)

fixes #2605 
TODO can we do this? 

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/Profiles/ProfilesEnabled.tsx
+++ b/webui/react/src/pages/TrialDetails/Profiles/ProfilesEnabled.tsx
@@ -1,7 +1,8 @@
 import { Alert } from 'antd';
-import React, { useEffect } from 'react';
+import React from 'react';
 
 import Section from 'components/Section';
+import Spinner from 'components/Spinner';
 import { useProfilesFilterContext } from 'pages/TrialDetails/Profiles/ProfilesFiltersProvider';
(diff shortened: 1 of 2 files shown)


apizzini:
React code looks good, but we have to involve @stoksc and @armandmcqueen here.

Displaying "no data" until system metrics data loads (and are not empty) was a requirement.

stoksc:
For context @hamidzr , there may not always be timings metrics, and the API isn't aware (it just returns what the harness gives it, it does not know which framework we are on). So the goal was to avoid spinning forever on timings for keras.

stoksc:
I see this is only for system metrics, not timings. This is fine, those should always be there.

stoksc (approved):

hamidzr:
> I see this is only for system metrics, not timings. This is fine, those should always be there (so long as a conf with `profiling.enabled = false` still renders with a link to the docs not a spinner.

yep the ones that have it disabled  still get the following
image

hamidzr:
I'm gonna rebase to avoid any potential conflict with the new header changes that landed after tests ran here.

PR #2610, opened 2021-06-23
feat: support auto bind mount as part of det deploy local [DET-5432]

naren-determined:
## Description

Add support for automatically bind mounting the user's home directory onto task agents for `det deploy local` and a flag for disabling it.

## Test Plan

Test by running: 
1. `det deploy local cluster-up`, open two files in the base directory and ensure any changes to one happen to the other.
2. `det deploy local cluster-up --no-auto-bind-mount` to ensure normal behavior.

## Checklist

### DIFF
--- a/harness/determined/deploy/local/cli.py
+++ b/harness/determined/deploy/local/cli.py
@@ -29,7 +29,7 @@ def handle_cluster_up(args: argparse.Namespace) -> None:
         delete_db=args.delete_db,
         gpu=args.gpu,
         autorestart=(not args.no_autorestart),
-        auto_bind_mount=(not args.no_auto_bind_mount),
+        auto_bind_mount=args.auto_bind_mount,
     )
 
 
@@ -52,7 +52,7 @@ def handle_master_up(args: argparse.Namespace) -> None:
         db_password=args.db_password,
         delete_db=args.delete_db,
         autorestart=(not args.no_autorestart),
-        auto_bind_mount=(not args.no_auto_bind_mount),
+        auto_bind_mount=args.auto_bind_mount,
         cluster_name=args.cluster_name,
     )
 
@@ -159,10 +159,13 @@ args_description = Cmd(
                     help="disable container auto-restart (recommended for local development)",
                     action="store_true",
                 ),
-                Arg(
+                BoolOptArg(
+                    "--auto-bind-mount",
                     "--no-auto-bind-mount",
-                    help="disable bind mounting the user's home directory into task containers",
-                    action="store_true",
+                    dest="auto_bind_mount",
+                    default=True,
+                    true_help="bind mount the user's home directory into task containers",
+                    false_help="disable mounting the user's home directory into task containers",
                 ),
             ],
         ),
@@ -227,10 +230,13 @@ args_description = Cmd(
                     help="disable container auto-restart (recommended for local development)",
                     action="store_true",
                 ),
-                Arg(
+                BoolOptArg(
+                    "--auto-bind-mount",
                     "--no-auto-bind-mount",
-                    help="disable bind mounting the user's home directory into task containers",
-                    action="store_true",
+                    dest="auto_bind_mount",
+                    default=True,
+                    true_help="bind mount the user's home directory into task containers",
+                    false_help="disable mounting the user's home directory into task containers",
                 ),
                 Arg(
                     "--cluster-name",
--- a/harness/determined/deploy/local/cli.py
+++ b/harness/determined/deploy/local/cli.py
@@ -29,6 +29,7 @@ def handle_cluster_up(args: argparse.Namespace) -> None:
         delete_db=args.delete_db,
         gpu=args.gpu,
         autorestart=(not args.no_autorestart),
+        auto_bind_mount=(not args.no_auto_bind_mount),
     )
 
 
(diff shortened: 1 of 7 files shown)


master/cmd/determined-master/root.go
```diff
@@ -137,6 +143,35 @@ func getConfig(configMap map[string]interface{}) (*internal.Config, error) {
...
+	if !fieldExists {
+		return configMap, nil
+	}
+	vBindMount, ok := bindMountField.(string)
+	if !ok {
+		return nil, errors.New("wrong type for bind mount field")
+	}
+
+	if vBindMount != "no" {
+		_, tcdExisted := configMap["task_container_defaults"]
```
ioga, line 157: I am not sure I quite like this, an env variable which overwrites a chunk of config may be a bit fragile.
We've talked about the `master.yaml.template` for aws & gcp last week. Maybe we should adopt a similar approach for det deploy local: instead of passing a set of env variables through docker-compose, have a `master.yaml.template` inside `det deploy` cli, which is filled in with values, and then mounted into the local master container as a file.

having said that, it may be a bigger and more radical change, so I won't block this PR over it.

master/cmd/determined-master/root.go
```diff
@@ -137,6 +143,35 @@ func getConfig(configMap map[string]interface{}) (*internal.Config, error) {
...
+		if !tcdExisted {
+			newTCD := make(map[string]interface{})
+			bindMounts := make([]model.BindMount, 1)
+			bindMounts[0] = model.BindMount{
+				HostPath:      vBindMount,
+				ContainerPath: "/run/determined/shared_fs",
+			}
+			newTCD["bind_mounts"] = bindMounts
+			configMap["task_container_defaults"] = []map[string]interface{}{newTCD}
```
ioga, line 166: if a user has a `--master-config-path`, having this option enabled will wipe out `task-container-defaults` section. 
should we disallow having both `--master-config-path` and this new option together (probably on the CLI side)? or document this behavior?

harness/determined/deploy/local/cli.py
```diff
@@ -157,6 +159,11 @@ def deploy_local(args: argparse.Namespace) -> None:
                     help="disable container auto-restart (recommended for local development)",
                     action="store_true",
                 ),
+                Arg(
```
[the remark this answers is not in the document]
shiyuann: I saw `--gpu` and `--no-gpu` are both by default False. Is it a bug? Because there should be a default value. @ioga

master/cmd/determined-master/root.go
shiyuann: Agreed. Don't think this PR should be blocked. Once we figure out what library or template language to use in that, we could move this environment variable to the new `master.yaml.template`.

harness/determined/deploy/local/cli.py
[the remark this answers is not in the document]
naren-determined: So this is defined in `common` as follows, which means in the code I am setting one of the flags to be True (and I see this behavior in testing). However, the [documentation](https://docs.determined.ai/latest/reference/cli.html?highlight=det%20deploy%20local) lists both as the same value which is weird.

` class BoolOptArg(NamedTuple):
     """Describes a boolean --foo / --no-foo flag pair."""
 
     true_name: str 
     false_name: str 
     dest: str 
     default: bool = False
     true_help: Optional[str] = None
     false_help: Optional[str] = None`

master/cmd/determined-master/root.go
naren-determined: Added code to handle the case when user specified a `--master-config-path`.

shiyuann (approved):
Looks good! I tested with `det deploy local cluster-up` and it worked.

master/cmd/determined-master/root.go
```diff
@@ -137,6 +143,47 @@ func getConfig(configMap map[string]interface{}) (*internal.Config, error) {
...
+			vTCD, ok := tcd.(map[string]interface{})
+			if !ok {
+				return nil, errors.New("wrong type for task_container_defaults field")
+			}
+
+			vTCD["bind_mounts"] = append(
+				vTCD["bind_mounts"].([]model.BindMount),
+				model.BindMount{
+					HostPath:      vBindMount,
+					ContainerPath: "/run/determined/shared_fs",
```
rb-determined-ai, line 178: @naren-determined is this the long-term location we expect to mount this directory to?

naren-determined: @shiyuann can confirm, but this is also the mounting location for AWS/GCP for notebooks.

harness/determined/deploy/local/cli.py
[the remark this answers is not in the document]
ioga: I've finally found some time to poke at this. Argparse prints the "default" value for the underlying field (`dest='gpu'`), not whether the cli flag itself is set or not. So as a result it shows `Default: False` in the help and in the autogenerated documentation for both flags. Interpret it as a final value that'll be used for the pair of flags.

master/cmd/determined-master/root.go
shiyuann: @rb-determined-ai yes. Before we want to change, this will be the path for the shared fs container path

PR #2615, opened 2021-06-24
chore: update environment images before release

mackrorysd:
Updating images. Also, since Ubuntu images for GovCloud need to be added manually, the previous entries caused this run to fail because equivalents couldn't be looked up the same way. We now just gracefully log a warning, reminding the user that this needs to happen (as said in the README).

I'm running all the opt-in tests here: https://app.circleci.com/pipelines/github/determined-ai/determined/13602/workflows/d7bcc6c7-85ee-42c9-816a-3aa1e131f8a5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,11 +97,11 @@ commands:
       - when:
           condition: <<parameters.tf1>>
           steps:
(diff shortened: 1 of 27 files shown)


azhou-determined (approved):
lgtm

mackrorysd:
Something's up with CircleCI considering all the "request" entities as "Running". All tests have passed...

PR #2616, opened 2021-06-24
fix: pull registry_auth from experiment for tensorboard

rb-determined-ai:


rb-determined-ai:
Is this actually a good idea, or is it better to assume that registry auth comes from the cluster config or an explicit tensorboard config?

stoksc (approved):
lgtm

rb-determined-ai:
updated to not override what was provided explicitly by either the cluster registry auth or the tensorboard explicitly-provided config.  I feel better landing this than I felt about the last edition.

PR #2617, opened 2021-06-24
fix: improve det deploy aws messaging for inconsistent stack states [DET-5695]

varlogtim:
## Test Plan

Tested Manually.

### DIFF
--- a/harness/determined/deploy/aws/aws.py
+++ b/harness/determined/deploy/aws/aws.py
@@ -18,6 +18,10 @@ DELETE_MASTER_IGNORE_ERRORS = (
 )
 
 
+class NoStackOutputError(Exception):
+    pass
+
+
 def get_user(boto3_session: boto3.session.Session) -> str:
     sts = boto3_session.client("sts")
     response = sts.get_caller_identity()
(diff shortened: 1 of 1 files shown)


rb-determined-ai:
@cla-bot[bot] check

ioga (approved):
do you think if we're doing `delete`, we should still continue and try to delete the stack (while skipping an attempt to stop the master)? 
I believe this operation is very likely to succeed for `ROLLBACK_` states, and will save the user a trip to the CF console.

Otherwise, the change by itself looks good to me.

varlogtim:
> do you think if we're doing `delete`, we should still continue and try to delete the stack (while skipping an attempt to stop the master)?
> I believe this operation is very likely to succeed for `ROLLBACK_` states, and will save the user a trip to the CF console.

That makes sense to me. I will take a deeper look today.

harness/determined/deploy/aws/aws.py
```diff
@@ -62,8 +66,23 @@ def stop_master(master_id: str, boto3_session: boto3.session.Session, delete: bo
...
+    try:
+        stack_output = get_output(stack_name, boto3_session)
+    except NoStackOutputError:
+        print(
+            f"Stack {stack_name} is in inconsistent state. "
+            "This error is ignored as stack is going to be deleted."
+        )
+        if stack_uses_spot(stack_name, boto3_session):
+            print("Terminating Running Agents and Pending Spot Requests")
+            clean_up_spot(stack_name, boto3_session)
```
ioga, line 80: it seems wrong that `clean_up_spot` doesn't need stack outputs, and reconstructs spot instance tags by itself (instead of using the values from output like `terminate_running_agents`).
so I think we should either
1. remove this block, lines 78-81, because it's unlikely there are any spot instances anyway,
2. or preface this block with a comment why we only delete spot instances (and not normal ones or s3 bucket).

varlogtim: Sounds good. I didn't really like that block being in there twice anyway.

ioga: lgtm, feel free to land when the checks pass

PR #2618, opened 2021-06-24
docs: correct pod spec

rb-determined-ai:
## Description

https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret


eecsliu (approved):
👌👌

PR #2625, opened 2021-06-26
fix: prevent impossible slot requests for notebooks [DET-5690]

ioga:
## Description
- Return API error for "impossible" slot requests.
- Limit max number of slots in the UI.

## Test Plan
Given a GCP, AWS, or Kube cluster with say 4 MaxSlotsPerPod or 4 gpu-nodes:
- In "Launch notebook" UI model, the number of slots for compute will be limited to the max possible number of slots (i.e. 4).
- "Bad" commands such as `det notebook start --config resources.slots=10`, `det shell start --config resources.slots=10` etc. should fail with a comprehensible error about an incorrect # of slots.
- "Good" commands such as `det shell start --config resources.slots=4` should succeed.
On-prem resource pools, e.g. `det deploy local` or local devclusters, will always allow all requests as the max slots are not known. As such, any request should continue working on these clusters.

## Checklist

### DIFF
--- a/master/internal/api/error.go
+++ b/master/internal/api/error.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 
 	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/labstack/echo/v4"
 )
@@ -61,4 +63,22 @@ func AsErrNotFound(msg string, args ...interface{}) error {
 		msg,
 		args...,
 	)
+}
+
+// APIErr2GRPC converts internal api error categories into grpc status.Errors.
+func APIErr2GRPC(err error) error {
+	switch {
+	case errors.Is(err, ErrBadRequest):
+		return status.Errorf(
+			codes.InvalidArgument,
+			err.Error(),
+		)
+	case errors.Is(err, ErrNotFound):
+		return status.Errorf(
+			codes.NotFound,
+			err.Error(),
+		)
+	default:
+		return nil
+	}
 }
--- a/master/internal/api/error.go
+++ b/master/internal/api/error.go
@@ -79,6 +79,6 @@ func APIErr2GRPC(err error) error {
 			err.Error(),
 		)
 	default:
-		return nil
+		return err
 	}
 }
--- a/master/internal/api_command.go
+++ b/master/internal/api_command.go
@@ -199,10 +199,7 @@ func (a *apiServer) LaunchCommand(
 		Data:         req.Data,
 	})
 	if err != nil {
-		if grpcErr := api.APIErr2GRPC(err); grpcErr != nil {
-			return nil, grpcErr
-		}
-		return nil, err
+		return nil, api.APIErr2GRPC(err)
 	}
 
 	commandLaunchReq := command.CommandLaunchRequest{CommandParams: params}
--- a/master/internal/api_command.go
+++ b/master/internal/api_command.go
@@ -94,9 +94,13 @@ func (a *apiServer) makeFullCommandSpec(
 	}
 
 	if config.Resources.Slots > 0 {
-		fillable := sproto.ValidateRPResources(
+		fillable, err := sproto.ValidateRPResources(
 			a.m.system, config.Resources.ResourcePool, config.Resources.Slots)
-		if fillable != nil && !*fillable {
+		if err != nil {
+			return nil, nil, errors.Wrapf(
+				err, "failed to check resource pool resources: ")
+		}
+		if !fillable {
 			return nil, nil, errCommandUnfulfillable
 		}
 	}
--- a/master/internal/api_command.go
+++ b/master/internal/api_command.go
@@ -28,6 +28,7 @@ import (
 )
 
 var commandsAddr = actor.Addr("commands")
+var errCommandUnfulfillable = errors.New("resource request unfulfillable")
 
 type protoCommandParams struct {
 	TemplateName string
@@ -92,6 +93,14 @@ func (a *apiServer) makeFullCommandSpec(
 		}
 	}
 
+	if config.Resources.Slots > 0 {
+		fillable := sproto.ValidateRPResources(
+			a.m.system, config.Resources.ResourcePool, config.Resources.Slots)
+		if fillable != nil && !*fillable {
+			return nil, nil, errCommandUnfulfillable
+		}
+	}
+
 	return &config, &taskSpec, nil
 }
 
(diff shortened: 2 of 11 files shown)


master/internal/api/error.go
```diff
@@ -62,3 +64,21 @@ func AsErrNotFound(msg string, args ...interface{}) error {
...
+			codes.InvalidArgument,
+			err.Error(),
+		)
+	case errors.Is(err, ErrNotFound):
+		return status.Errorf(
+			codes.NotFound,
+			err.Error(),
+		)
+	default:
+		return nil
```
stoksc, line 82: I think default should just return the error, not discard it (so then the calls below are just `if err != nil { return api.APIErr2GRPC(err) }`. My concern is just this way is misuse-able and that is not.

master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -136,8 +139,12 @@ func (a *agentResourceManager) forwardToPool(
 	ctx *actor.Context, resourcePool string, msg actor.Message,
 ) {
 	if a.pools[resourcePool] == nil {
+		sender := "unknown"
+		if ctx.Sender() != nil {
```
stoksc, line 143: can sender ever be nil?

master/internal/resourcemanagers/resource_managers_test.go
```diff
@@ -35,3 +36,30 @@ func TestResourceManagerForwardMessage(t *testing.T) {
...
+			{
+				PoolName:                 defaultResourcePoolName,
+				MaxAuxContainersPerAgent: 100,
+			},
+		},
+	}
+
+	Setup(system, echo.New(), conf, nil, nil)
+	value := sproto.ValidateRPResources(system, defaultResourcePoolName, 1)
+	assert.Assert(t, is.Nil(value))
```
stoksc, line 64: Why not just `assert.Assert(t, value == nil)` or assert? (ask mainly because the rename import is not standard go).

master/internal/sproto/task.go
```diff
@@ -65,3 +65,33 @@ type Allocation interface {
 	Start(ctx *actor.Context, spec tasks.TaskSpec)
 	Kill(ctx *actor.Context)
 }
+
+type (
+	// ValidateCommandResourcesRequest is a message asking resource manager whether the given
+	// resource pool can (or, rather, if it's not impossible to) fulfill the command request
+	// for the given amount of slots.
+	ValidateCommandResourcesRequest struct {
```
stoksc, line 73: I think placement-wise, these probably belong beside `GetTaskSummaries/GetResourceSummary` since they are messages a resource pool handles.

master/internal/sproto/task.go
```diff
@@ -65,3 +65,33 @@ type Allocation interface {
...
+		Slots        int
+	}
+
+	// ValidateCommandResourcesResponse is the response to ValidateCommandResourcesRequest.
+	ValidateCommandResourcesResponse struct {
+		// Fulfillable values:
+		// - nil: unknown
+		// - false: impossible to fulfill
+		// - true: ok
+		Fulfillable *bool
```
stoksc, line 84: I would prefer if this is just a `bool` not `*bool`. Do we really need unknown? On prem agents are still within a resource pool afaik and have slots per agent -1. I don't see any case where we could have a nil `*bool` in the code.

master/internal/sproto/task.go
```diff
@@ -65,3 +65,33 @@ type Allocation interface {
...
+	}
+)
+
+// ValidateRPResources checks if the resource pool can fulfill resource request for single-node
+// notebook/command/shell etc. Returns &true if yes, &false if not, and nil if unknown.
+func ValidateRPResources(system *actor.System, resourcePoolName string, slots int) *bool {
```
stoksc, line 90: I think this should probably return `(bool, error)` and we should check the error on the response since it can return with one. So like 
```
if resp.Error() != nil {
    return false, resp.Error()
}
return resp.(ValidateCommandResourcesResponse).Fulfillable
```

webui/react/src/components/NotebookModal.tsx
```diff
@@ -239,17 +240,20 @@ const NotebookForm:React.FC<FormProps> = (
...
     if (!selectedPool) {
-      return { hasAux: false, hasCompute: false, showResourceType: true };
+      return { hasAux: false, hasCompute: false, maxSlots: 0, showResourceType: true };
     }
     const hasAuxCapacity = selectedPool.auxContainerCapacityPerAgent > 0;
     const hasComputeCapacity = selectedPool.slotsAvailable > 0
       || (!!selectedPool.slotsPerAgent && selectedPool.slotsPerAgent > 0);
+    const maxSlots = hasComputeCapacity ?
+      (selectedPool.slotsPerAgent && selectedPool.slotsPerAgent > 0 ?
+        selectedPool.slotsPerAgent : 2**16) : 0;
```
hamidzr, line 256: nit: I think you could use `undefined` to represent "lack of known max slots" and the input element should accept it and behave niely as well. This works too

master/internal/api_command.go
```diff
@@ -133,6 +146,10 @@ func (a *apiServer) prepareLaunchParams(ctx context.Context, req *protoCommandPa
 	params.FullConfig, params.TaskSpec, err = a.makeFullCommandSpec(
 		configBytes, &req.TemplateName, req.MustZeroSlot)
 	if err != nil {
+		if err == errCommandUnfulfillable {
+			return nil, api.AsErrBadRequest(
+				"resource request unfulfillable, please try requesting less slots")
```
hamidzr, line 151: nit: less resources?

hamidzr (approved):
the ui portion LGTM

master/internal/resourcemanagers/resource_managers_test.go
stoksc: i've looked at using testify, fine with it tbh

master/internal/api/error.go
stoksc: yeah, i see.. i just think this could get misused pretty easily and we could drop the error. in that case i guess it'd prefer 
`return api.APIErr2GRPC(errors.Wrap(err, "failed to prepare launch params"))`
I think the grpc error formed from the `error.Error()` of the passed error still needs to context "failed to prepare launch params" or else the error may just be like "string too long" or something (totally made up example)

stoksc (commented):
lgtm. as far as testing, what was done as far as a test plan goes for on-prem deployments?

stoksc (approved):
lgtm.

master/internal/api/error.go
ioga: okay this works

ioga:
> lgtm. as far as testing, what was done as far as a test plan goes for on-prem deployments?

local devcluster is effectively on-prem deployment.

stoksc:
@ioga 
> local devcluster is effectively on-prem deployment.
i mean specifically in the "test plan"

ioga:
updated the "test plan".

PR #2634, opened 2021-06-29
chore: make experiment description editable via webui [DET-4398]

apizzini:
## Description

used just here right now, but we should adopt it in other places.
ux should more or less respect initial caleb ideas.

Schermata 2021-06-29 alle 14 51 32

## Checklist


hamidzr (commented):
Overall this is looking good. I added some comments, mostly nit picks and a feature suggestion.

Are the package-lock changes intentional?
```
  -  "lockfileVersion": 2,
  +  "lockfileVersion": 1,
```

it'd be nice if we could set enter (or ctrl-enter) to confirm. this would also allow us to nudge people to avoid entering multiline text in certain places eg name or maybe even description.

webui/react/package.json
```diff
@@ -36,6 +36,7 @@
     "pretty-bytes": "^5.4.1",
     "query-string": "^7.0.0",
     "react": "^17.0.2",
+    "react-contenteditable": "^3.3.5",
```
apizzini, line 39: @hamidzr I cannot get `lockfileVersion` to stay 2. That's supported only since npm > 7 and if i use the last versions coming with node 15 or 16 I get either a broken app (`make live`) or a broken deps install (`npm install`).

Lmk if you have suggestions.

webui/react/src/components/InlineTextEdit.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const handleFocus = useCallback(() => {
+    setIsFocused(true);
+  }, []);
+  const handleSetRef = useCallback((el: HTMLElement) => {
+    inputRef.current = el;
+  }, []);
+
+  return (
+    <>
+      <ContentEditable
```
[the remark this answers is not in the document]
apizzini: possible but that's not something we should get in this MR.

apizzini:
@hamidzr thanks for the review, made some of the fixes you suggested and added support for esc and submit keypress.

webui/react/package.json
hamidzr: same here. this looks good thanks for looking into it

webui/react/src/components/InlineTextEdit.tsx
```diff
@@ -0,0 +1,93 @@
...
+    inputRef.current.blur();
+  }, [ value ]);
+  const focus = useCallback(() => {
+    if (!inputRef.current) return;
+    inputRef.current.focus();
+  }, []);
+  const save = useCallback(async () => {
+    setIsSaving(true);
+    setIsChanged(false);
+    await setValue(innerValueRef.current);
```
hamidzr, line 35: blocking: could we also handle request (setValue) failures as well? If my request fails this seems to stay in the spinning state.

hamidzr (approved):
LGTM 
minus the one comment

webui/react/src/components/InlineTextEdit.tsx
apizzini: added, nice catch.

PR #2640, opened 2021-06-30
docs: clarify node 16 incompatibility

bensomers:
## Description
Clarify that node 16 currently does not work - running `make all` will throw errors around installing `libsass`.

## Test Plan

Docs change only.

## Checklist


hamidzr (approved):
🚢 
thanks Ben

hamidzr:
on second thought maybe we should just advertise the latest LTS. 14 is the latest lts  https://nodejs.org/en/download/releases/

CONTRIBUTING.md
```diff
@@ -50,7 +50,7 @@ git clone git@github.com:determined-ai/determined.git
 
 - Go (>= 1.16)
 - Python (>= 3.6, < 3.8)
-- Node (>= 12)
+- Node (>= 12, < 16)
```
hamidzr, line 53: suggestion: I still have some unresolved questions with 15 let's stick with `<15`

CONTRIBUTING.md
```diff
@@ -50,7 +50,7 @@ git clone git@github.com:determined-ai/determined.git
...
 - NPM (>= 6.12)
```
hamidzr, line 54: npm usually comes bundled with node I've never mixed and matched these manually but to be safe let's also put an upper limit here.

the latest active lts comes with `Node.js 14.17.1	Fermium	2021-06-15	8.4.371.23	NPM: 6.14.13	83`

bensomers: I don't know that we need to upper-bound NPM until we actually have evidence it doesn't work on newer releases.(though `6.14.13` is what I used here)

shiyuann:
Does this PR accidentally come with a bunch of unrelated commits?

bensomers:
> Does this PR accidentally come with a bunch of unrelated commits? @bensomers

it's just rebased on top of latest master

PR #2642, opened 2021-06-30
feat: change hyperparameter tab on multi-trial experiment trial pages [DET-5413]

EmilyBonar:
## Description
Before:
Screenshot_2021-06-30 Trial 2507 - Determined(1)

After:
Screenshot_2021-06-30 Trial 2507 - Determined

## Test Plan
We'll do a bug bash to test this manually before release.

## Checklist


webui/react/src/pages/TrialDetails/TrialRangeHyperparameters.module.scss
```diff
@@ -0,0 +1,98 @@
+$color-blue: rgb(0, 155, 222);
```
apizzini, line 1: (this is something i often forget too) we have css vars defining colors and some other variables for measures (margin, border-radius, ...). Would be best to use them here too.

apizzini: eg: this color is `var(--theme-colors-action-normal)`

webui/react/src/pages/TrialDetails.tsx
```diff
@@ -232,7 +232,7 @@ const TrialDetailsComp: React.FC = () => {
           <TrialDetailsOverview experiment={experiment} trial={trial} />
         </TabPane>
         <TabPane key={TabType.Hyperparameters} tab="Hyperparameters">
-          <TrialDetailsHyperparameters experiment={experiment} trial={trial} />
+          <TrialRangeHyperparameters experiment={experiment} trial={trial} />
```
apizzini, line 235: If it's easy to do we could potentially try to check exp config to understand if exp is single trial or not (maybe by moving the check i added on the exp page in a common util fn) and serve this component only if the exp is not single-trial.

User can still access the trial page via URL, I don't think this will happen, so I would do this small change only if it takes a couple of minutes.

webui/react/src/pages/TrialDetails/TrialRangeHyperparameters.tsx
```diff
@@ -0,0 +1,210 @@
...
+          val: String(value.find(ob => ob.name === hp.name)?.value || 0),
+          vals: hp.value.vals?.map(val => String(val)) ||
+          [ String(hp.value.minval || 0), String(hp.value.maxval || 1) ],
+        }
+      );
+    });
+  }, [ experiment.config.hyperparameters, trial.hparams ]);
+
+  return (
+    <div style={{ display: 'flex' }}>
```
apizzini, line 48: IMHO this looks better if wrapped in a `<Section bodyBorder>` (so there's a container with white background)

apizzini: and i think we need to make sure on mobile everything is visible, so either scrollable vertically or horizontally.

webui/react/src/pages/TrialDetails/TrialRangeHyperparameters.module.scss
```diff
@@ -0,0 +1,102 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 10px;
```
apizzini, line 4: non-blocking: as discussed on yesterday's Webui AG.. There are some variables with values like this, it's unclear if we want to enforce using them, but i'm usually asked to use them by Hamid and Caleb.

apizzini (approved):
LGTM, just 1 other comment but is not a blocker.

PR #2644, opened 2021-06-30
feat: improve support for `pytorch_geometric` and custom pytorch batches

ioga:
## Description

There was some interest in OSS slack on `pytorch_geometric` support; the problem is that this library uses custom class which exposes `.to` method (so we can move it to the device just fine), but its `__len__` doesn't represent the number of records in a batch, and we require that value.

Let's allow users to override the batch length calculation in using `PyTorchTrial#calculate_batch_length`.

## Test Plan

This PR includes an example based on `pytorch_geometric`, which is also our first GCN (graph convolutional network) example. It should run to completion on a cluster both in `const` and `distributed` configurations. The standard trials which do not override this field should run normally as well.

## Checklist

### DIFF
--- a/examples/graphs/proteins_pytorch_geometric/const.yaml
+++ b/examples/graphs/proteins_pytorch_geometric/const.yaml
@@ -1,0 +1,19 @@
+name: proteins_pytorch_geometric
+hyperparameters:
+  global_batch_size: 64
+  dataset: PROTEINS
+  lr: 5.e-4
+  weight_decay: 1.e-4
+  hidden_channels: 32
+records_per_epoch: 1001
+searcher:
+  name: single
+  metric: validation_loss
+  max_length:
+    epochs: 300
+  smaller_is_better: true
+entrypoint: model_def:GCNTrial
+environment:
+  image:
+    cpu: determinedai/environments:py-3.7-pytorch-1.9-lightning-1.3-tf-2.4-cpu-b06dafb
+    gpu: determinedai/environments:cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-b06dafb
--- a/examples/graphs/proteins_pytorch_geometric/const.yaml
+++ b/examples/graphs/proteins_pytorch_geometric/const.yaml
@@ -5,7 +5,8 @@ hyperparameters:
   lr: 5.e-4
   weight_decay: 1.e-4
   hidden_channels: 32
-records_per_epoch: 1001
+  training_records: 890
+records_per_epoch: 890
 searcher:
   name: single
   metric: validation_loss
--- a/examples/graphs/proteins_pytorch_geometric/distributed.yaml
+++ b/examples/graphs/proteins_pytorch_geometric/distributed.yaml
@@ -5,6 +5,7 @@ hyperparameters:
   lr: 5.e-4
   weight_decay: 1.e-4
   hidden_channels: 32
+  training_records: 890
 records_per_epoch: 890
 searcher:
   name: single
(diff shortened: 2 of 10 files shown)


shiyuann:
TBH, I think `calculate_ batch_length` is a hack to override the behavior of the `TrialController`. And it's hard to understand as an user interface. We should actually expose `TrialController` to users rather than letting them need to understand a function that is not useful in most cases.

ioga:
>need to understand a function that is not useful in most cases
I don't think they need to understand this function for most cases, just for the ones where they actually need it, like this one. In that regard, it's similar to other more advanced features beyond the bare minimum, such as callbacks or `evaluation_reducer`.

>I think our API designing framework of letting the Trial class be only user-facing and TrialController to be internal is wrong. Any hacks to make up the flaws of this framework are making the API worse.
So what is your proposal? 

a. Move `calculate_batch_length` to `PyTorchTrialController` , so a user would do something like

```
class MyTrialController(PyTorchTrialController):
    def calculate_batch_length(self, batch):
        return batch.num_graphs

class MyTrial(PyTorchTrial):
    trial_controller_class = MyTrialController
```

b. not expose this as abstraction at all, and make them override entire `_train_for_step` and `_compute_validation_metrics`?

examples/graphs/proteins_pytorch_geometric/model_def.py
```diff
@@ -0,0 +1,149 @@
...
+        # other when doing distributed training.
+        download_directory = tempfile.mkdtemp()
+
+        self.dataset = TUDataset(
+            root=download_directory,
+            name=self.context.get_hparam("dataset"),
+            use_node_attr=True,
+        )
+
+        val_ratio = 0.2
```
naren-determined, line 69: Can we put this in the .yaml file config? I don't like hardcoded values like this in the model definition.

naren-determined (commented):
Similar to Ilia's stance, I view `calculate_batch_length` as an advanced feature that is optionally used just like custom reducers. I don't really think exposing the `TrialController` is going to be that useful for the vast majority of users who may not necessarily be using these advanced features and may in fact cause more confusion. Not a fan of options (a) or (b) Ilia laid out.

examples/graphs/proteins_pytorch_geometric/distributed.yaml
```diff
@@ -0,0 +1,21 @@
...
+  hidden_channels: 32
+records_per_epoch: 890
+searcher:
+  name: single
+  metric: validation_loss
+  max_length:
+    epochs: 300
+  smaller_is_better: true
+entrypoint: model_def:GCNTrial
+resources:
```
shiyuann, line 16: remove this?

examples/graphs/proteins_pytorch_geometric/distributed.yaml
```diff
@@ -0,0 +1,21 @@
...
+  slots_per_trial: 4
+environment:
```
shiyuann, line 18: also remove this?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -980,6 +980,36 @@ def evaluate_full_dataset(self, data_loader: torch.utils.data.DataLoader) -> Dic
         """
         pass
 
+    def calculate_batch_length(self, batch: Any) -> int:
```
shiyuann, line 983: rename it to use "get" since it doesn't really calculate anything?

```suggestion
    def get_batch_length(self, batch: Any) -> int:
```

examples/graphs/proteins_pytorch_geometric/const.yaml
```diff
@@ -0,0 +1,19 @@
...
+  hidden_channels: 32
+records_per_epoch: 1001
+searcher:
+  name: single
+  metric: validation_loss
+  max_length:
+    epochs: 300
+  smaller_is_better: true
+entrypoint: model_def:GCNTrial
+environment:
```
shiyuann, line 16: same

shiyuann (approved):
LGTM!

PR #2652, opened 2021-07-01
fix: kube RPs should allow aux tasks [DET-5710]

ioga:
## Description

WebUI uses `auxContainersPerAgent` to check if zero-slot tasks are allowed; Kubernetes does allow these, so its dummy resource pool should advertise itself as such.
Also, we now have `jq` as a requirement so I'm updating the `CONTRIBUTING` docs for good measure.

## Test Plan

Deploy a kube cluster and try to launch a zero-slot notebook via "Launch Notebook" UI.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,6 +57,7 @@ git clone git@github.com:determined-ai/determined.git
 - Protoc (>= 3.0)
 - Java (>= 7)
 - cURL (>= 7)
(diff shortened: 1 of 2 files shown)


shiyuann (approved):

eecsliu (approved):

ioga:
using admin merge due to a flaky webui test.

PR #2653, opened 2021-07-02
fix: improve trial chart tooltip [DET-5712]

apizzini:
## Description

Tooltips are displayed on what uPlot suggests to be the closest (to the cursor) x index.
In this case the chart might have multiple series which are hidden, and uPlot still account them for computing the closest x index.
If the only visible serie has way less data points than hidden ones, we can see what's reported in the ticket.

Solution:
Only provide uPlot series that are visible, so that x index is computed according to available (and visible) data point.

## Checklist


EmilyBonar (approved):
Looks great!

PR #2656, opened 2021-07-02
ci: correct order of artifact release [DET-5195]

mackrorysd:
## Description

This should be straighforward because I do not see how any of the Python packages could be pulled into the things packaged by goreleaser. The problem is that people are able to install the latest CLI and try deploying a cluster before the rest of the system is available.

## Test Plan

Can't really test this well until we do a release, so I'll plan to be actively involved in case anything goes wrong with the next release. Unfortunately CircleCI also doesn't make it easy to find workflows run on a tag, and these jobs are run specifically only on a tag, so I think the only way to see the logs is to run them anew.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2220,13 +2220,6 @@ workflows:
             - build-helm
             - build-proto
 
-      - publish-python-package:
-          matrix:
-            parameters:
-              path: ["harness", "common", "cli", "deploy", "model_hub"]
-          context: determined-production
-          filters: *release-and-rc-filters
-
       - package-and-push-system-rc:
           requires:
             - build-react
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -2241,6 +2234,13 @@ workflows:
           context: determined-production
           filters: *release-filters
 
+      - publish-python-package:
```
ioga, line 2237: hmm I am not sure just changing the order will help here? I believe this needs `requires:` on either `package-and-push-system-rc` or `package-and-push-system-release`, depending on the filter.

dzhu: This is correct, I'm pretty sure.

PR #2662, opened 2021-07-03
chore: clean up task spec

shiyuann:
## Description

Currently, we support agent resource manager and k8s resource manager. We will support Slurm resource manager in the future. Also, we will support other task types, such as inference tasks.

**The Go struct design** is as follows:

1.  Container spec structs: `ContainerdSpec` (now `ContainerSpec`), `K8sSpec` (now `PodSpec`), and `SlurmSpec` will be used to represent all the specifications to start one container.
2. Generic `TaskSpec`: A generic `TaskSpec` is designed to include all the information of a task that does not depend on any specific container platform. It has member functions to generate container spec structs for many containers. (One task might have many containers.). The `TaskSpec` contains the shared logic between all the tasks.
3. Task structs: `CommandSpec`, `GCCkptSpec`, and `TrialSpec` are designed to include all the parameters to create tasks. They also have the member functions to generate `TaskSpec`. They can be used to restart a new same task.
4. We will also have structs for runtime information for tasks and containers. The runtime information together with specs can be used to resume a stopped task.
5. All the spec structs are not supposed to change after being initialized although there are some lazy initialization logics.

This PR is a first step to move to this direction.

## Test Plan

1. Use the e2e tests.
2. Sanity-check all the features.

## Commentary (optional)

N/A

## Checklist


master/internal/resourcemanagers/resource_pool.go
```diff
@@ -366,7 +366,7 @@ func (c containerAllocation) Summary() sproto.ContainerSummary {
 }
 
 // StartContainer notifies the agent to start a container.
-func (c containerAllocation) Start(ctx *actor.Context, spec image.TaskSpec) {
+func (c containerAllocation) Start(ctx *actor.Context, spec image.TaskSpec, rank int) {
```
stoksc, line 369: side note, why is tasks imported as image? doesn't really make sense to me

master/internal/trial.go
```diff
@@ -643,62 +614,28 @@ func (t *trial) processAllocated(
...
-		),
-	}
 	taskToken, err := t.db.StartTaskSession(string(t.task.ID))
 	if err != nil {
 		return errors.Wrap(err, "cannot start a new task session for a trial")
 	}
 	t.preemption = newPreemption()
+
+	trialSpec := &tasks.TrialSpec{
+		Base: *t.taskSpec,
```
stoksc, line 624: i might would embed this.. since the fields in Base and the TrialSpec itself should be mutually exclusive (TrialSpec is just a more specific BaseSpec), this would guarantee this at compile time and sort of hint it to the reader.

master/internal/experiment.go
```diff
@@ -168,6 +167,7 @@ func newExperiment(master *Master, expModel *model.Experiment, taskSpec *tasks.T
 	if agentUserGroup == nil {
 		agentUserGroup = &master.config.Security.DefaultTask
 	}
+	taskSpec.AgentUserGroup = agentUserGroup
```
stoksc, line 170: this snippet is in every task actor and would be really easily for someone to forget. may be nice to make it harder to miss.

stoksc (commented):
looks pretty good, though we should be really careful to look for small bugs. i may try to eye it a again later.

master/internal/sproto/task.go
```diff
@@ -92,6 +92,6 @@ func NewTaskID() TaskID {
 // to start tasks on assigned resources.
 type Allocation interface {
 	Summary() ContainerSummary
-	Start(ctx *actor.Context, spec tasks.TaskSpec)
+	Start(ctx *actor.Context, spec tasks.TaskSpec, rank int)
```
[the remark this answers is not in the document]
shiyuann: I agree it. I think we will need structs for runtime information to help recover /starta task.

master/internal/resourcemanagers/resource_pool.go
shiyuann: I don't know why it is imported as `image`. I just didn't change it. Probably we should change ti

master/internal/trial.go
shiyuann: I think embedding a struct looks like inheritance although there is no such concept in GoLang. `TrialSpec` is not just a more specific BaseSpec but makes a `TaskSpec` on the fly. So I would avoid creating any struct relationship between `TaskSpec` and `TrialSpec`.

master/internal/experiment.go
shiyuann: One of my next steps to move this defaulting logic to one place instead of separating out to everywhere. We will get there.

master/internal/trial.go
stoksc: Embedding a struct isn't really inheritance at all, it's composition. A trial spec is trial specific details composed with the base spec, and they are mutual exclusive. I think embedding makes sense here but that's fine.

master/internal/experiment.go
stoksc: sgtm

master/internal/trial.go
shiyuann: It's inheritance. It inherits all the member functions and you can override any of the functions.

stoksc: It's not inheritance when you just compose two structs. The key detail of inheritance (and what makes it so different/annoying) is the parent-child hierarchy where some methods invoke the parent and child versions, some override, some fall back to the parent and you never know which it will do without digging into the methods themselves. Embedding a struct isn't inheritance but just strictly composition.. the methods on the structs will always do what they always did.. anyway.

stoksc (approved):
lgtm.. i've tried to read this so carefully for small bugs but in the end we'll just need to test a ton. but as i land push i will hit it plenty

master/internal/trial.go
stoksc: well, looks like i'm wrong here, they override.. i swear.. give me some time to figure out why i thought this

master/internal/checkpoint_gc.go
```diff
@@ -48,35 +38,15 @@ func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 		ctx.Tell(t.rm, *t.task)
 
 	case sproto.ResourcesAllocated:
+		ctx.Log().Info("starting checkpoint garbage collection")
```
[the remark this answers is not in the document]
shiyuann: yes. I didn't really change the location of the log. Why do you ask?

master/internal/command/shell_manager.go
```diff
@@ -110,7 +106,7 @@ func (s *shellManager) processLaunchRequest(
 
 func (s *shellManager) newShell(
 	params *CommandParams,
-	keyPair ssh.PrivateAndPublicKeys,
+	keyPair *ssh.PrivateAndPublicKeys,
```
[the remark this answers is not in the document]
shiyuann: Okay. I think I just wanted to pass the `keyPair` through to the field `generatedKeys`. It doesn't really matter actually.

master/internal/trial.go
stoksc: ok at some point i had a refactor with a `tasks.Task` and a `taskspec.Task` and i embedded them and got an error that lead me to believe you couldn't. ok i'm firmly against embedding now, thanks @shiyuann .

master/internal/checkpoint_gc.go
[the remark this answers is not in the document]
stoksc: im just blind, i looked for the log in the old code and read right over it. thought it was a debug line.

PR #2663, opened 2021-07-05
fix: Add legend to Trial metrics graph [DET-5723]

apizzini:
## Description

Made 2 changes:

1. Display the type of metric on the tooltip
Schermata 2021-07-05 alle 14 37 32

2. If only 1 metric is selected, show it as the Y axis label:
Schermata 2021-07-05 alle 14 31 53

I'm not sure if alternatively we want to display the uPlot default legend and drop the select above the chart.

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/TrialChart.tsx
+++ b/webui/react/src/pages/TrialDetails/TrialChart.tsx
@@ -24,6 +24,12 @@ interface Props {
 
 const STORAGE_PATH = 'trial-detail';
 
+const getChartMetricLabel = (metric: MetricName): string => {
+  if (metric.type === 'training') return `[T] ${metric.name}`;
+  if (metric.type === 'validation') return `[V] ${metric.name}`;
+  return metric.name;
+};
+
 const TrialChart: React.FC<Props> = ({
   defaultMetricNames,
   metricNames,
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -86,8 +89,8 @@ const TrialChart: React.FC<Props> = ({
       },
       series: [
         { label: 'Batch' },
-        ...metrics.map((metricName, index) => ({
-          label: metricName.name,
+        ...metrics.map((metric, index) => ({
+          label: `${metric.name} (${metric.type})`,
```
hamidzr, line 93: suggestion: if we can use any elements here it might be good to try and keep it consistent with how we display the type in other places  


if it's text only maybe `[V] metricname` ?

hamidzr (commented):
I can't see the original ask from the ticket but the change looks good to me. 

> I'm not sure if alternatively we want to display the uPlot default legend and drop the select above the chart.

Is this something we've discussed before? would it cover all our current usecases? would it further differentiate uplot and plotly backed charts

apizzini:
The original ask of the ticket is not that specific.

@EmilyBonar @ahj-determined can you please check the description, screenshot and comments on this MR?

ahj-determined:
@apizzini Sorry for not clarifying in the ticket. This looks great to me. Thanks for the change!

webui/react/src/pages/TrialDetails/TrialChart.tsx
hkang1: @apizzini thoughts on this? I agree that this would keep it consistent and make it easier for the user to connect the dots when they see the prefix. Visual div prefix would be preferred, but a text `[V]` or `[T]` is acceptable as well.

hkang1 (approved):
Code looks good! Having the metric type as a tag/prefix would be great to have!

webui/react/src/pages/TrialDetails/TrialChart.tsx
apizzini: Done :)

PR #2665, opened 2021-07-05
fix: Add page elements to Trial page while data is loading [DET-5718]

apizzini:
## Description

Simply shows this in all tabs (except config) instead of an empty page with spinners:
Schermata 2021-07-05 alle 16 07 51

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails.tsx
+++ b/webui/react/src/pages/ExperimentDetails.tsx
@@ -5,6 +5,7 @@ import CreateExperimentModal, { CreateExperimentType } from 'components/CreateEx
 import Message, { MessageType } from 'components/Message';
 import Page from 'components/Page';
 import Spinner from 'components/Spinner';
+import handleError, { ErrorLevel, ErrorType } from 'ErrorHandler';
 import usePolling from 'hooks/usePolling';
 import ExperimentDetailsHeader from 'pages/ExperimentDetails/ExperimentDetailsHeader';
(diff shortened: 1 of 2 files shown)


webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -61,7 +62,26 @@ const ExperimentDetails: React.FC = () => {
...
+        message: e.message,
+        publicMessage: 'Failed to fetch experiment trials.',
+        silent: true,
+        type: ErrorType.Server,
+      });
+    }
+  }, [ canceler, id ]);
+
   const { startPolling, stopPolling } = usePolling(fetchExperimentDetails);
+  const { stopPolling: stopPollingFirstTrialId } = usePolling(fetchFirstTrialId);
```
EmilyBonar, line 84: Is this supposed to be a comma after stopPolling, or is stopPollingFirstTrialId a type?

EmilyBonar (approved):
Looks good

webui/react/src/pages/ExperimentDetails.tsx
apizzini: that's to rename `stopPolling` since it's already used from the line before :)

in this case this get exposed aliased as `stopPollingFirstTrialId`

EmilyBonar: Ah, makes sense!

PR #2666, opened 2021-07-06
fix: correct for possibility of negative numbers in log range [DET-5717]

EmilyBonar:
## Description
If the range of a log type hp plot includes negative numbers, treat it as a double type instead.

## Checklist


apizzini (approved):
lgtm!

PR #2672, opened 2021-07-07
chore: update images for release

mackrorysd:
## Description

The combination of PyTorch 1.9 and the HOROVOD_GPU_ALLREDUCE=NCCL option (which means to use NCCL specifically for the All-Reduce algorithm only) has been resulting in extra processes on the GPUs and consequent memory errors. Configuring NCCL to be used for all operations appears to eliminate this problem and we probably should have been doing that anyway. Also, working around a recent regression in Pillow by just pinning to an older release (https://github.com/python-pillow/Pillow/issues/5571).

See also: https://github.com/determined-ai/environments/pull/110

## Test Plan

https://app.circleci.com/pipelines/github/determined-ai/determined?branch=all_operations_nccl

At the time of this writing, one job is still running but the rest have passed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,11 +97,11 @@ commands:
       - when:
           condition: <<parameters.tf1>>
           steps:
-            - run: docker pull determinedai/environments:py-3.7-pytorch-1.7-tf-1.15-cpu-0af9954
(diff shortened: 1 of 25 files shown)


ioga (approved):

mackrorysd:
Thank you, auto-merge. Circle CI was done last night but apparently the Github side of it doesn't recognize that.

PR #2674, opened 2021-07-07
refactor: allow user to inline edit experiment name [DET-4405]

apizzini:
## Description

Add support for inline editing. Apply to experiment name and experiment description within the experiment header.

https://user-images.githubusercontent.com/220971/127579900-9b317a41-856b-4b2b-8225-92c6d0c9ad31.mov

**NOTE:** Updated from the previous approach of using a `react-contenteditable` component. The new custom approach uses a textarea to eliminate any non-text artifacts (HTML and other rendering artifacts) and dynamically grows the editor, mapping to the content size. Contenteditable elements support rich text but is extremely difficult to manage, due to unpredictable artifacts inside the editor.

## Checklist


hkang1 (approved):
Sorry for the delay, I don't want to block this. This looks good!

I'd like to take a more in depth look at this later. The symptom you found seems to be an intentional attempt at breaking it, which hopefully users will not sabotage their own experiment name. :)

apizzini:
I don't think this is really good to go.

1. Breaking layout on mobile (we can fix).
2. Issue happens easily on a smaller screen with 1440px width (not sure how we could fix this). I see only 2 possible solution: opening an <input> in a modal or tooltip or moving the name away from the header.

Are you really sure the current state is acceptable on prod?

hkang1:
> I don't think this is really good to go.
> 
> 1. Breaking layout on mobile (we can fix).
> 2. Issue happens easily on a smaller screen with 1440px width (not sure how we could fix this). I see only 2 possible solution: opening an  in a modal or tooltip or moving the name away from the header.
> 
> Are you really sure the current state is acceptable on prod?

Ah sorry, I didn't know the mobile view and the smaller screen sizes were the issue, I mostly focused on the screen grab. I'll have to try the smaller mobile views.

PR #2676, opened 2021-07-07
fix: preserve subroutes for experiment detail page

hkang1:
## Description

2 route improvements:
- The hp visualization routes for multi-trial experiments are getting redirected to a base route. For example, `/experiments/123/visualization/hp-parallel-coordinates` get redirected to `/experiments/123`. 
- Multi-trial and single-trial experiment pages reroute to a simpler route if the default tab is selected. Update the route to accurately reflect exactly what tab state the page is in. For example, currently if user is viewing Visualizations for the Multi-trial experiment page, it will change the route (if applicable) to `/experiments/123`. Instead we want `/experiments/123/visualization` for a couple of reasons: 1) we are starting to  sync up and preserve page state in the URL so they can be shareable. 2) avoid complex logic to maintain default routes (the default routes will still work but will redirect to the accurate routes)

## Checklist


EmilyBonar (approved):

apizzini (approved):
LGTM

PR #2684, opened 2021-07-07
fix: avoid passing loginRedirect to routeAll

hamidzr:
## Description

We know that loginRedirect is always set to internal react routes since it originates from a page in our SPA so we set signIn to always treat it as an internal route instead of trying to figure out whether it's an internal route or not.

## Commentary (optional)

Determining whether a string is a path to A) an internal react route vs B) an asset that hosted by react vs C ) a server route quickly and without making async requests is hard
multiple ways to address this: TBD (edited) 

- stop using string routes, use react route objects
- stop using relative paths where possible
- remove the need for routeAll
- if we use routeAll we explicitly tell it if we know it's an internal route which is an equivelant of just using routeToReactUrl
- something else?

## Checklist

### DIFF
--- a/webui/react/src/pages/SignIn.tsx
+++ b/webui/react/src/pages/SignIn.tsx
@@ -50,11 +50,10 @@ const SignIn: React.FC = () => {
 
       // Reroute the authenticated user to the app.
       const loginRedirect = getPath<Location>(location, 'state.loginRedirect');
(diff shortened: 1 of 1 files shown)


hkang1 (approved):
This looks good! Tried it on incognito sessions and also the query `redirect`

Thanks for jumping on this!

hamidzr:
thanks for helping me debug this

PR #2687, opened 2021-07-07
chore: move automount path into /run/determined/workdir

rb-determined-ai:
## Description

At the risk of a name conflict around `shared_fs`, put the automount
mount point inside /run/determined/workdir, to avoid having to expose
the Determined-internal paths in `/run/determined/*` to users.

The potential name conflict is detected during model upload and any
top-level files named `shared_fs` are simply ignored.  This will make it
easy for users to train either locally or on-cluster with the same code
by symlinking the automount directory into their local context directory
as `shared_fs`.  That symlink will simply be ignored when training
on-cluster.

Ultimately there are three competing requirements:

1. We don't want to move the context directory from
   /run/determined/workdir (too many users expect that exact path)

2. We don't want to expose users to the other paths in /run/determined/*

3. We don't want to create a possible name conflict with the user's
   context directory.

The previous approach solved 1 & 3, sacrificing 2.  This approach solves
1 & 2, sacrificing 3.

## Test Plan

- [x] test that `shared_fs` files are being ignored in context directory
- [ ] test that `det deploy local` is working nicely (doesn't work on archlinux)
- [ ] test that `det deploy aws` is working nicely (aws is out of instances?  or we are?)
- [x] test that `det deploy gcp` is working nicely


shiyuann (approved):

rb-determined-ai:
Having trouble testing this one, because `make package && det deploy local cluster-up` doesn't work on archlinux (something something `version 'GLIBC_2.32' not found`), and I'm actually not sure how to test changes to `det deploy aws` or `det deploy gcp`.

@shiyuann or @naren-determined could I get some help with verifying it?  Or do we want to just leave that to the release party?

(I did verify the `shared_fs` ignoring behavior)

rb-determined-ai:
Tested on GCP, AWS seems to be out of instances, and I can't test on det-deploy local on an archlinux machine.  Landing as-is.

PR #2690, opened 2021-07-07
chore: add a safeguard for setting server address through the omnibar

hamidzr:
## Checklist

### DIFF
--- a/webui/react/src/dev.ts
+++ b/webui/react/src/dev.ts
@@ -1,5 +1,6 @@
 /* Tools and tweaks for dev environments */
 import { globalStorage } from 'globalStorage';
+import { paths, routeToReactUrl } from 'routes/utils';
 import * as Api from 'services/api';
 import { updateDetApi } from 'services/apiConfig';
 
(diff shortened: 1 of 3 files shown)


apizzini (approved):
Looks good.

Didn't know about the omnibar.

2 nit thing:
- close the omnibar after the command is submitted?
- when setting the dev server we could potentially reload the page (need to do that manually)?

hamidzr:
> close the omnibar after the command is submitted?

there has been some talks around this one. There is a google document and a jira ticket we're using to collect feedback on it and this is one of them 🤔  (I'll share the links)

hamidzr:
> when setting the dev server we could potentially reload the page (need to do that manually)?

I meant to send the user to the login page but I'm now sending users to the dashboard. I think that should be enough and no reload would be necessary

hamidzr:
I made two changes:
- send the user to the logout route after a server address change
- treat server address reset as a server address change

webui/react/src/dev.ts
```diff
@@ -1,16 +1,27 @@
...
+import { paths, routeToReactUrl, serverAddress } from 'routes/utils';
 import * as Api from 'services/api';
 import { updateDetApi } from 'services/apiConfig';
 
+const onServerAddressChange = () => {
+  updateDetApi({ basePath: serverAddress() });
+  routeToReactUrl(paths.logout());
+};
+
 export const setServerAddress = (address: string): void => {
```
hamidzr, line 12: I"m not adding the same safeguard here to allow for more flexibility when going this route.

PR #2693, opened 2021-07-07
build: let netlify handle node dependency management

hamidzr:
## Description

Let netlify handle grabbing node dependencies to take advantage of thecaching that it provides

## Commentary (optional)

should shave off about 50 seconds from each build

## Checklist

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -14,11 +14,12 @@
   publish = "build/"
 
   # Default build command.
-  command = "make get-deps build"
+  command = "npm run build"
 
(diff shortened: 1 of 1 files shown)


hkang1 (approved):
I don't know enough about this to do a proper review :D

But I'm assuming the `npm ci` gets triggered somewhere else then since it's no longer being done by `make get-deps`. Also are we ok not dealing with bindings either before the build?

hamidzr:
> Also are we ok not dealing with bindings either before the build?

those are now committed to the repo and we don't need to build them to deploy.

> I'm assuming the npm ci gets triggered somewhere else

if netlify detects a change in package-lock it'll proceed to download the new dependencies is what I understand it to be  https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

hamidzr:
It's a pretty safe change and easy to go back on if we notice any issues with it.

PR #2695, opened 2021-07-08
fix: no data available [DET-5734]

hkang1:
## Description

When viewing a single trial experiment, the detail page shows `No data available` during a trial load. There is a special case where the experiment has not progressed enough to actually have trial data, but is still running. To avoid confusing the user into thinking that the trial will forever not have data, show a spinner instead.

Before:
Screen Shot 2021-07-07 at 4 37 01 PM

After:
Screen Shot 2021-07-07 at 4 37 23 PM

## Checklist


apizzini (approved):
LGTM 👍

PR #2701, opened 2021-07-09
fix: remove proteins_pytorch_geometric from nightlies.

ioga:
## Description

Seems like `proteins_pytorch_geometric` example sometimes has convergence issues which breaks nightlies. Remove it for the nightlies for now, pending an investigation. Also set batch_size to 20 to align with the original repo example.

## Checklist

### DIFF
--- a/e2e_tests/tests/nightly/test_capability.py
+++ b/e2e_tests/tests/nightly/test_capability.py
@@ -82,13 +82,3 @@ def test_word_language_lstm_const() -> None:
     config["hyperparameters"]["tied"] = False
 
     exp.run_basic_test_with_temp_config(config, conf.nlp_examples_path("word_language_model"), 1)
-
-
-@pytest.mark.nightly  # type: ignore
-def test_protein_pytorch_geometric() -> None:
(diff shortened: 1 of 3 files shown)


e2e_tests/tests/nightly/test_capability.py
```diff
@@ -82,13 +82,3 @@ def test_word_language_lstm_const() -> None:
     config["hyperparameters"]["tied"] = False
 
     exp.run_basic_test_with_temp_config(config, conf.nlp_examples_path("word_language_model"), 1)
-
```
RehanSD, line 85: Should we comment this out instead of just deleting so we know to return to it eventually?

RehanSD (commented):
LGTM except for the one comment I left. Once we address that I'm happy to merge!

e2e_tests/tests/nightly/test_capability.py
ioga: I think this is what git history is for. you don't want a lot of dead commented code laying around.

RehanSD (approved):
LGTM!

PR #2703, opened 2021-07-09
docs: warn users not to use cpu training with a custom scheduler

eecsliu:
## Checklist

### DIFF
--- a/docs/how-to/installation/kubernetes.txt
+++ b/docs/how-to/installation/kubernetes.txt
@@ -151,8 +151,7 @@ want to use finer-grained ``slotResourceRequests.cpu: 15``, set
 ``maxSlotsPerPod: 2``.
 
 When using CPU-only training, avoid setting the ``defaultScheduler``
-field. CPU training has not been validated with these scheduler options
-and may cause instability the deployment.
+field. CPU training is not supported for these options yet.
 
 Checkpoint Storage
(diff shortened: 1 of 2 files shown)


docs/how-to/installation/kubernetes.txt
```diff
@@ -318,6 +322,10 @@ set ``defaultScheduler`` to ``preemption``.
 
    defaultScheduler: preemption
 
+If you are using cpu training with Kubernetes, avoid setting the
```
ioga, line 325:
```suggestion
If you are using CPU-only training with Kubernetes, avoid setting the
```

docs/how-to/installation/kubernetes.txt
```diff
@@ -318,6 +322,10 @@ set ``defaultScheduler`` to ``preemption``.
...
+``defaultScheduler`` field. Cpu training has not yet been validated for
```
ioga, line 326:
```suggestion
``defaultScheduler`` field. CPU training is not supported for these scheduling options yet.
```

ioga (approved):

docs/how-to/installation/kubernetes.txt
ioga: I don't see these changes applied after the last force-push?..

PR #2707, opened 2021-07-09
fix: bug in eval for dtrain for question-answering example [DET-5756][DET-5757]

liamcli:
## Description
* Distributed training for question answering examples used to have very low f1 scores due to mismatched labels and inputs.  The mismatch stems from how we shard data for distributed eval and subsequently gather outputs for the custom reducer.  We shard by checking whether the index % world_size == rank but then aggregate examples across slots by stacking.  The fix is to provide index for each sample and sort by index to match targets correctly.  
* The ALBERT NLP example is a subset of the possible supported configurations for model-hub question answering.  This PR adds the config for ALBERT with 8 gpus and matches the performance in the ``examples/nlp/albert_squad_pytorch`` example.  

## Test Plan
- [x] Test ``squad_distributed.yaml`` config and verify performance of at least 93 f1 reported at https://github.com/huggingface/transformers/tree/master/examples/pytorch/question-answering
- [x] Test ``squad_v2_albert.yaml`` config and verify we reach close to 88.8 f1 from ``examples/nlp/albert_squad_pytorch/README.md``

## Checklist


.bumpversion.cfg
```diff
@@ -43,3 +43,15 @@ values =
 [bumpversion:file:model_hub/examples/huggingface/text-classification/glue_config.yaml]
 
 [bumpversion:file:model_hub/examples/huggingface/text-classification/xnli_config.yaml]
+
+[bumpversion:file:model_hub/examples/huggingface/question-answering/squad.yaml]
```
ioga, line 47: Maybe `[bumpversion:glob:model_hub/examples/huggingface/*/*.yaml]`? We won't need to bother adding each file then.

naren-determined (commented):
LGTM -- don't have any comments here

.bumpversion.cfg
liamcli: thanks for the suggestion.  i changed bumpversion in #2614

naren-determined (approved):

PR #2717, opened 2021-07-12
fix: cpu only preemption [DET-5763]

eecsliu:
## Description
Kubernetes was preempting the master and DB instance because they had a priority of 0. This was occurring as part of the kubernetes preemption protocol, and not as a result of Determined itself. This fix adds a priority value to the master and db instances, as well as reintroduces content from the reverted pull #2631 (prevent CPU-only training from being blocked by the coscheduler).

## Test Plan
Set up a CPU-only kubernetes cluster and submit a job. First, make sure that the master and DB instances don't die and restart after job submission. Then, make sure the job is able to be scheduled and starts running.

## Checklist


ioga (approved):
lgtm.

PR #2721, opened 2021-07-12
feat: check server reachability on first load

hamidzr:
## Checklist

### DIFF
--- a/webui/react/src/App.tsx
+++ b/webui/react/src/App.tsx
@@ -90,48 +90,44 @@ const AppView: React.FC = () => {
   // Correct the viewport height size when window resize occurs.
   useLayoutEffect(() => correctViewportHeight(), [ resize ]);
 
-  return (
-    <div className={css.base}>
-      <Navigation>
-        <main>
-          <Router routes={appRoutes} />
-          {ui.omnibar.isShowing && <Omnibar />}
-        </main>
-      </Navigation>
-    </div>
-  );
-};
-
-const App: React.FC = () => {
   const [ isAlive, setIsAlive ] = useState<boolean>(true);
 
   useEffect(() => {
     checkServerAlive().then(aIsAlive => {
       if (!aIsAlive) {
         setIsAlive(false);
-        // notification.warn({
-        //   btn: <Button type="primary" onClick={refreshPage}>Retry</Button>,
-        //   description: `We cannot communicate with the server at "${serverAddress()}".
-        //   Please check firewall or server settings.`,
-        //   duration: 0,
-        //   message: 'Server Unreachable',
-        // });
       }
     });
   }, []);
 
-  const message = `We cannot communicate with the server at "${serverAddress()}".
-          Please check firewall or server settings.`;
-
   const UnreachableServerMessage = (
-    <PageMessage message={message} title="Server Unreachable" />
+    <PageMessage title="Server Unreachable">
+      <p>We cannot communicate with the server at {'"'}{serverAddress()}{'"'}.
+      Please check firewall or server settings.</p>
+      <Button onClick={refreshPage}>Retry</Button>
+    </PageMessage>
   );
+
+  return (
+    <div className={css.base}>
+      {isAlive ?
+        <Navigation>
+          <main>
+            <Router routes={appRoutes} />
+          </main>
+        </Navigation>
+        : UnreachableServerMessage }
+      {ui.omnibar.isShowing && <Omnibar />}
+    </div>
+  );
+};
+
+const App: React.FC = () => {
 
   return (
     <HelmetProvider>
       <StoreProvider>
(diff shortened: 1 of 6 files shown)


webui/react/src/App.tsx
```diff
@@ -18,7 +18,7 @@ import appRoutes from 'routes';
 import { correctViewportHeight, refreshPage } from 'utils/browser';
 
 import css from './App.module.scss';
-import { paths } from './routes/utils';
+import { checkServerAlive, paths, serverAddress } from './routes/utils';
```
hkang1, line 21: **chore fix**: Not introduced by the PR but I noticed in `checkServerAlive` there is a small logic that needs to be fixed. This portion needs to be re-written:
```
    attrs.forEach(att => {
      if (!(att in data)) {
        // the server doesn't look like a determined server
        return false;
      }
    });
```
to something like:
```
    for (let attr of attrs) {
      // The server doesn't look like a determined server.
      if (!(attr in data)) return false;
    }
```
The return call in `forEach` doesn't break out of the callback to exit the overall `checkServerAlive` function with a return value. Changing it to `for` should do the trick.

hamidzr: aah good catch!

webui/react/src/App.tsx
```diff
@@ -89,6 +89,20 @@ const AppView: React.FC = () => {
...
+          btn: <Button type="primary" onClick={refreshPage}>Retry</Button>,
+          description: `We cannot communicate with the server at "${serverAddress()}".
+          Please check firewall or server settings.`,
+          duration: 0,
+          message: 'Server Unreachable',
+        });
+      }
+    });
+  }, []);
+
```
hkang1, line 105: **suggestion:** what do you think about displaying in the center of the `App` view the error message and the refresh button instead of having a notification popup on a blank page?

webui/react/src/App.tsx
hkang1: **questions:** couple of questions:
1. Can `checkServerAlive` use `getInfo` already defined in `api.ts`?
2. will `checkServerAlive` be used outside of `App`? Does it need to be reusable vs just having it exist in `App.ts`? (in case referencing `getInfo` causes a circular dependency)

hkang1 (commented):
I think this is a good idea, as it just shows a blank page like you mentioned. Some implementation questions and suggestions.

webui/react/src/App.tsx
hamidzr: hmm I was thinking of the case that we do the same UX with a heart-beat check so the user could be doing something and the server goes away. In that case it might be nicer to avoid covering the whole screen. but if we're addressing just the current use case sending the user to an offline page might be a better experince

webui/react/src/App.tsx
hamidzr: 1. we should be able to. that's a good suggestion. I wanted to be able to quickly change the server address since it's originally used to check whether a given remote server is valid and alive or not. should still be doable when using `getInfo`

2. it currently is used outside of `App` in the omnibar when the user provides a new server address to switch to. I'll watch out for that circular dependency

webui/react/src/App.tsx
hamidzr: added an offline page.

hamidzr:
I added an offline page and it changes how the message is presented to the user and consequently their experience. could you give it a try and see what you think? 

any suggestions on the wording? maybe i should change it to be "we _couldn't_ communi..."

webui/react/src/App.tsx
```diff
@@ -89,19 +90,40 @@ const AppView: React.FC = () => {
...
+    checkServerAlive().then(aIsAlive => {
+      if (!aIsAlive) {
+        setIsAlive(false);
+      }
+    });
+  }, []);
+
+  const UnreachableServerMessage = (
+    <PageMessage title="Server Unreachable">
+      <p>We cannot communicate with the server at {'"'}{serverAddress()}{'"'}.
```
hkang1, line 105: **nit:** small non-blocking suggestions
- `&quot;` instead of `{'"'}`. Just a guess but maybe slight benefit of not requiring extra JSX processing?
- `Unable to communicate` instead of `We cannot communicate`

webui/react/src/App.tsx
```diff
@@ -89,19 +90,40 @@ const AppView: React.FC = () => {
...
+      Please check the firewall and cluster settings.</p>
+      <Button onClick={refreshPage}>Retry</Button>
```
hkang1, line 107: **super nit:** `Try Again` instead of `Retry`? feel free to ignore as usual

hkang1 (approved):
LGTM! great idea with this!

hkang1:
@hamidzr does this need release notes?

PR #2724, opened 2021-07-13
feat: transformers translation example in model-hub [DET-5633]

liamcli:
## Description
Add huggingface transformer's translation example to our model-hub.  

## Test Plan
- [x] Train each experiment configuration and confirm performance is as expected

## Commentary (optional)
The upstream transformers translation example has configurations for three model types: marian, t5, and mbart.  The marian and mbart configurations errored out when I trained to run it myself outside of determined.  

The example configurations provided in this PR fix those issues.  

## Checklist


e2e_tests/tests/model_hub/test_transformers.py
```diff
@@ -185,6 +185,83 @@ def test_text_classification_xnli_amp() -> None:
...
+    config = conf.set_global_batch_size(config, 8)
+    config = conf.set_max_length(config, {"batches": 200})
+    config = conf.set_hparam(config, "use_apex_amp", True)
+    config = set_docker_image(config)
+
+    exp.run_basic_test_with_temp_config(config, example_path, 1)
+
+
+@pytest.mark.model_hub  # type: ignore
+@pytest.mark.model_hub  # type: ignore
```
naren-determined, line 264: Is there a reason for these extra @pytest.mark.model_hub lines here?

model_hub/examples/huggingface/translation/translation_trial.py
```diff
@@ -0,0 +1,348 @@
...
+        self.logger.info("training records: {}".format(train_length))
+        if (
+            "records_per_epoch" in self.exp_config
+            and train_length != self.exp_config["records_per_epoch"]
+        ):
+            self.logger.warning(
+                "number of train records {} does not match records_per_epoch of {}".format(
+                    train_length, self.exp_config["records_per_epoch"]
+                )
+            )
```
naren-determined, line 100: I think this block here is assuming that you are only running for 1 epoch which is true for all of the default configs but not necessarily for all user-provided configs, so should it be:
`train_length != self.exp_config["records_per_epoch"] * self.exp_config["epochs"]`

model_hub/examples/huggingface/translation/translation_trial.py
```diff
@@ -0,0 +1,348 @@
...
+MULTILINGUAL_TOKENIZERS = [
+    transformers.MBartTokenizer,
+    transformers.MBartTokenizerFast,
+    transformers.MBart50Tokenizer,
+    transformers.MBart50TokenizerFast,
+    transformers.M2M100Tokenizer,
+]
+
+
+class TranslationTrial(hf.BaseTransformerTrial):
```
naren-determined, line 46: One thing I noticed that could be helpful here is a link to the docs so people can look at some of the `model_hub.utils` functions that you use here in case they are confused where that comes from. Ditto the other model_hub examples as well.

model_hub/examples/huggingface/translation/translation_trial.py
```diff
@@ -0,0 +1,348 @@
...
+        if self.data_config.pad_to_max_length:
+            collator = transformers.default_data_collator
+        else:
+            collator = transformers.DataCollatorForSeq2Seq(
+                self.tokenizer,
+                model=self.model,
+                label_pad_token_id=label_pad_token_id,
+                pad_to_multiple_of=8 if self.hparams.use_apex_amp else None,
+            )
+        self.collator = lambda x: collator(x).data
```
naren-determined, line 263: Not exactly sure how this would be done in a nice way outside of nesting this function in the __init__() which isn't a great solution, but since `self.collator` is used in other functions would be nice to have that defined in the __init__(). I recognize that in this code `preprocess_function` is guaranteed to be called before the data loaders so functionally it's fine, but just wanted to point it out.

model_hub/examples/huggingface/translation/translation_trial.py
liamcli: i don't think we should be multiplying `records_per_epoch` by the number of epochs.  `train_length` is the length of the dataset so it should be the same as `records_per_epoch` if that field is set in the experiment config.

e2e_tests/tests/model_hub/test_transformers.py
liamcli: removed

model_hub/examples/huggingface/translation/translation_trial.py
liamcli: i moved setting the collator to trial initialization.  the collate functions just need to be created after building the tokenizer.

model_hub/examples/huggingface/translation/translation_trial.py
liamcli: good suggestion, i added a reference for each of the transformer trials

naren-determined (approved):
Lgtm! 🚢

gh-determined-ai (approved):

liamcli:
closing PR since we will support better hf transformers integration with a core API callback soon

PR #2726, opened 2021-07-13
fix: going back in the browser doesn't update the settings of the experiment list view [DET-5625]

apizzini:
## Description

I spend some time trying to support "reloading" the previous state of the table when going back in the browser history.

That's not easy due to having state partially "synced" between ant-table and the page component (the cleanest solution would have been to update the state by parsing the url when it changes, by avoiding to change the state directly from the component and changing it only by parsing the url).

But this simple fix is probably the best one (avoiding to push a new state in the browser history any time the URL changes but you're still in the experiment list page).

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -139,7 +139,7 @@ const ExperimentList: React.FC = () => {
       searchParams.append('user', URL_ALL);
     }
 
-    window.history.pushState(
+    window.history.replaceState(
       {},
       '',
       url.origin + url.pathname + '?' + searchParams.toString(),
--- a/webui/react/src/pages/TrialDetails/Profiles/ProfilesFiltersProvider.tsx
+++ b/webui/react/src/pages/TrialDetails/Profiles/ProfilesFiltersProvider.tsx
@@ -94,7 +94,7 @@ const ProfilesFiltersProvider: React.FC<Props> = ({ children, trial }: Props) =>
       searchParams.append('gpuUuid', filters.gpuUuid);
     }
 
-    window.history.pushState(
+    window.history.replaceState(
       {},
       '',
       url.origin + url.pathname + '?' + searchParams.toString(),


webui/react/src/pages/ExperimentList.tsx
```diff
@@ -139,7 +139,7 @@ const ExperimentList: React.FC = () => {
       searchParams.append('user', URL_ALL);
     }
 
-    window.history.pushState(
+    window.history.replaceState(
```
hkang1, line 142: @apizzini the crux of this problem might be more related to the fact that table page offset is not stored in the URL. We intentionally made that choice of not including table page offset in the URL, So when the user clicks a bunch of different table pages, the URL is all the same, so when the user clicks the back button it just defaults to table page offset of 0 (first table page).

What do you think of the solution to add table page offset to the URL query params as well? That way with the pushing of routes, it'll store the page offset info properly and the back button should naturally behave as expected.

hkang1 (commented):
@apizzini the crux of this problem might be more related to the fact that table page offset is not stored in the URL. We intentionally made that choice of not including table page offset in the URL, So when the user clicks a bunch of different table pages, the URL is all the same, so when the user clicks the back button it just defaults to table page offset of 0 (first table page).

What do you think of the solution to add table page offset to the URL query params as well? That way with the pushing of routes, it'll store the page offset info properly and the back button should naturally behave as expected.

apizzini:
Table offset and limit are currently in the URL.

http://localhost:3000/det/experiments?...&limit=10&offset=30&...

hkang1:
> Table offset and limit are currently in the URL.
> 
> http://localhost:3000/det/experiments?...&limit=10&offset=30&...

🤦 ah that's right.

Let me take this one as I am working on a [PR](https://github.com/determined-ai/determined/pull/2740) that is related to this area and it will address this issue.

hkang1:
If [PR 2740](https://github.com/determined-ai/determined/pull/2740) looks ok, we can close this PR.

PR #2731, opened 2021-07-14
chore: Add back Continue Trial button to single-trial experiment page [DET-5749]

apizzini:
## Description

This should take care of:
- On the Trial page of single-trial experiments, let's add an action to "Continue Trial" to the header
- On the Trial page of single-trial experiments, let's reorder the header buttons to be 1. TensorBoard 2. Download Model 3. Fork 4. Continue Trial
- On the Trial page of a multi-trial experiment, let's reorder the header buttons to be 1. TensorBoard 2. Continue Trial

## Checklist

### DIFF
--- a/webui/react/src/components/ContinueTrial.tsx
+++ b/webui/react/src/components/ContinueTrial.tsx
@@ -1,0 +1,100 @@
+import React, { forwardRef, useCallback, useImperativeHandle, useState } from 'react';
+
+import CreateExperimentModal, { CreateExperimentType } from 'components/CreateExperimentModal';
+import handleError, { ErrorType } from 'ErrorHandler';
+import { paths, routeToReactUrl } from 'routes/utils';
+import { createExperiment } from 'services/api';
+import { ExperimentBase, RawJson, TrialDetails, TrialHyperParameters } from 'types';
+import { clone } from 'utils/data';
+import { trialHParamsToExperimentHParams, upgradeConfig } from 'utils/types';
+
+export interface ContinueTrialHandles {
+  show: () => void;
+}
+
+interface Props {
+  experiment: ExperimentBase;
+  ref?: React.Ref<ContinueTrialHandles>;
+  trial: TrialDetails;
+}
+
+const trialContinueConfig = (
+  experimentConfig: RawJson,
+  trialHparams: TrialHyperParameters,
+  trialId: number,
+): RawJson => {
+  return {
+    ...experimentConfig,
+    hyperparameters: trialHParamsToExperimentHParams(trialHparams),
+    searcher: {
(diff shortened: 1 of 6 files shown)


webui/react/src/pages/ExperimentDetails/ExperimentDetailsHeader.tsx
```diff
@@ -23,14 +24,20 @@ import { openCommand } from 'wait';
 import css from './ExperimentDetailsHeader.module.scss';
 
 interface Props {
+  continueTrialRef: React.RefObject<ContinueTrialHandles>;
```
hkang1, line 27: **suggestion:** What do you think about following the same pattern as `showForkModal: () => void;` instead of passing in a ref?

webui/react/src/components/ContinueTrial.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const show = useCallback(() => {
+    const rawConfig = trialContinueConfig(clone(experiment.configRaw), trial.hparams, trial.id);
+    let newDescription = `Continuation of trial ${trial.id}, experiment ${trial.experimentId}`;
+    if (rawConfig.description) newDescription += ` (${rawConfig.description})`;
+    rawConfig.description = newDescription;
+    upgradeConfig(rawConfig);
+    setContModalConfig(rawConfig);
+    setIsVisible(true);
+  }, [ experiment.configRaw, trial ]);
```
hkang1, line 55: **question:** there might be special circumstances for doing it this way, but thought I ask this:
What if you pass through the `visible` on `CreateExperimentModal` to be a prop on `ContinueTrial` and use an `useEffect` + `usePrevious` to detect when `visible goes from `false` to `true` to do the `rawConfig` processing that's done in this `show` block.

hkang1 (commented):
Nice, the UI works nicely! I had some implementation suggestions and improvements, lemme know what you think and if you have questions.

webui/react/src/pages/ExperimentDetails/ExperimentDetailsHeader.tsx
apizzini: I used the ref to be consistent, but that's good to. Changed.

webui/react/src/components/ContinueTrial.tsx
apizzini: I think this adds un-needed complexity due to the fact that we need to pass both the state and the setState to the component and handle detect the status change.

But if you prefer that it's easy to change.
Pls lmk.

hkang1: as is fine, thanks!

hkang1 (approved):
LGTM

PR #2736, opened 2021-07-15
fix: restrict google-cloud-storage dependency.

ioga:
## Description

Restrict `google-cloud-storage` dependency so that 
a) it doesn't break lint-python
b) wide enough
c) we don't need to add extra additional package versions to the environment images.

## Checklist

### DIFF
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -28,7 +28,7 @@ setup(
         "backoff",
         "certifi",
         "filelock",
-        "google-cloud-storage>=1.20.0",
+        "google-cloud-storage>=1.20.0,<=1.40.0",
         # google-cloud-core 1.4.2 breaks our windows cli tests for python 3.5.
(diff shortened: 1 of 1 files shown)


harness/setup.py
```diff
@@ -28,7 +28,7 @@
         "backoff",
         "certifi",
         "filelock",
-        "google-cloud-storage>=1.20.0",
+        "google-cloud-storage>=1.20.0,<=1.40.0",
```
hkang1, line 31: **nit:** is this worth commenting in the line above that 1.40.x+ is breaking `lint-python` or whatever conflicts it's causing? (like the `google-cloud-core<1.4.2` line)

hkang1 (approved):
LGTM! I imported this change into my PR and it seems to do the trick in fixing `lint-python` in the CI!

rb-determined-ai (approved):

PR #2740, opened 2021-07-16
refactor: list query params [DET-5625]

hkang1:
## Description

Improve the code around retaining page state in the URL query param for experiment list page.

BEFORE:
There are essentially two `useEffects` that manage the query params. One parses and reads the query params and stores them into the state. The other detects changes in the state and updates the URL with the changes. This works fine, but in its current state there are few drawbacks. Because the two effects are essentially the opposite tasks of each other, there needs to be an extra state to ensure one happens after the other to avoid an infinite re-render.

AFTER:
Change the logic so the page states are dynamically determined from a singe step with a priority listing. Pull the state from the following list and in that order:
- URL query param
- user preference
- default value
Next refactor so when changing state, instead of changing it at the React component state level, change it directly in the URL so it will follow a one way cycle, so we don't have to worry about infinite circular logic.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -1,7 +1,9 @@
 import { ExclamationCircleOutlined } from '@ant-design/icons';
 import { Button, Modal } from 'antd';
 import { ColumnsType, FilterDropdownProps, SorterResult } from 'antd/es/table/interface';
+import queryString from 'query-string';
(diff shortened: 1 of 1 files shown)


apizzini (approved):
LGTM!

hkang1:
Closing this PR in lieu of a [more complete set of work](https://github.com/determined-ai/determined/pull/2786).

PR #2742, opened 2021-07-16
chore: add patch to webui proxy

hkang1:
## Description

Our netlify cluster is missing PATCH, so the experiment label API endpoint does not work in the netlify preview.

## Checklist


hamidzr (approved):
thanks for fixing this. I updated the server with this patch

EmilyBonar (approved):

PR #2748, opened 2021-07-19
fix: correct single trial experiment routing [DET-5789]

EmilyBonar:
## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails.tsx
+++ b/webui/react/src/pages/ExperimentDetails.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useParams } from 'react-router';
 
 import { ContinueTrialHandles } from 'components/ContinueTrial';
@@ -148,19 +148,6 @@ const ExperimentDetails: React.FC = () => {
     return () => stopPollingFirstTrialId();
   }, [ firstTrialId, isSingleTrial, stopPollingFirstTrialId ]);
 
-  const ExperimentTabs = useMemo(() => {
-    if (!experiment) return null;
-    if (isSingleTrial) {
-      return <ExperimentSingleTrialTabs
-        continueTrialRef={continueTrialRef}
-        experiment={experiment}
-        trialId={firstTrialId}
-      />;
-    } else if (isSingleTrial === false) {
-      return <ExperimentMultiTrialTabs experiment={experiment} />;
-    } else return null;
-  }, [ experiment, firstTrialId, isSingleTrial ]);
-
   if (isNaN(id)) {
     return <Message title={`Invalid Experiment ID ${experimentId}`} />;
   } else if (pageError) {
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -144,10 +144,23 @@ const ExperimentDetails: React.FC = () => {
...
+      return <ExperimentSingleTrialTabs
+        continueTrialRef={continueTrialRef}
+        experiment={experiment}
+        trialId={firstTrialId}
+      />;
+    } else if (isSingleTrial === false) {
+      return <ExperimentMultiTrialTabs experiment={experiment} />;
+    } else return null;
+  }, [ experiment, firstTrialId, isSingleTrial ]);
+
```
hkang1, line 163: **suggestion:** A simpler change might be to change this line:
```
  } else if (!experiment) {
    return <Spinner tip={`Loading experiment ${experimentId} details...`} />;
  }
```
to
```
  } else if (!experiment || isSingleTrial === undefined) {
    return <Spinner tip={`Loading experiment ${experimentId} details...`} />;
  }
```
Then you can avoid having to modify any of the JSX below because it won't reach that return block and by then `isSingleTrial` will be defined as true or false. Because if it's still `undefined` it will return a spinner.

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -164,21 +177,13 @@ const ExperimentDetails: React.FC = () => {
       headerComponent={<ExperimentDetailsHeader
         experiment={experiment}
         fetchExperimentDetails={fetchExperimentDetails}
-        isSingleTrial={isSingleTrial}
+        isSingleTrial={isSingleTrial || false}
```
hkang1, line 180: **chore:** this shouldn't be needed if the suggestion above is done.

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -164,21 +177,13 @@ const ExperimentDetails: React.FC = () => {
...
-      {isSingleTrial ? (
-        <ExperimentSingleTrialTabs
-          continueTrialRef={continueTrialRef}
-          experiment={experiment}
-          trialId={firstTrialId}
-        />
-      ) : (
-        <ExperimentMultiTrialTabs experiment={experiment} />
-      )}
+      {ExperimentTabs}
```
hkang1, line 186: **chore:** this change shouldn't be needed if the suggestion above is done.

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -170,15 +170,15 @@ const ExperimentDetails: React.FC = () => {
...
           continueTrialRef={continueTrialRef}
           experiment={experiment}
           trialId={firstTrialId}
         />
-      ) : (
-        <ExperimentMultiTrialTabs experiment={experiment} />
-      )}
+        ) : (
+          <ExperimentMultiTrialTabs experiment={experiment} />
+        )}
```
hkang1, line 181: **nit:** revert to exact change to avoid any spacing diffs

hkang1 (approved):
LGTM! yep just did a check on the preview and that did the trick! The route remains stable and unchanging.
One minor spacing nit to avoid extra changes to this PR.

PR #2751, opened 2021-07-20
RFC: naming things

rb-determined-ai:
The RFC proposed here is the combined thoughts and ideas of many people on the team.


rb-determined-ai:
I think this RFC is in its final state.

PR #2753, opened 2021-07-20
fix: pull model definitions into containers [DET-5788]

rb-determined-ai:
## Description

This fixes the model definition duplication problem, which would cause
experiments with model definitions greater than roughly 64MB to never
start.

This also fixes the model-def-too-big-for-config-maps problem when
running on kubernetes.

## Test Plan

Added an e2e test for a large upload, and I plan to run k8s tests to make sure this solution works on non-agents as well.

### DIFF
--- a/harness/determined/exec/fetch_context.py
+++ b/harness/determined/exec/fetch_context.py
@@ -1,0 +1,30 @@
+import base64
+import distutils.util
+import io
+import os
+import tarfile
+
+from determined import constants
+from determined.common.api import certs, request
+
+if __name__ == "__main__":
+    exp_id = os.environ["DET_EXPERIMENT_ID"]
+    master_addr = os.environ["DET_MASTER_ADDR"]
+    master_port = os.environ["DET_MASTER_PORT"]
+    use_tls = distutils.util.strtobool(os.environ.get("DET_USE_TLS", "false"))
+
+    master_url = f"http{'s' if use_tls else ''}://{master_addr}:{master_port}"
+    certs.cli_cert = certs.default_load(master_url=master_url)
+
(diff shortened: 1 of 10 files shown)


shiyuann:
IDK you write React.

rb-determined-ai:
I generate React : )

harness/determined/exec/fetch_context.py
```diff
@@ -0,0 +1,30 @@
...
+from determined.common.api import certs, request
+
+if __name__ == "__main__":
+    exp_id = os.environ["DET_EXPERIMENT_ID"]
+    master_addr = os.environ["DET_MASTER_ADDR"]
+    master_port = os.environ["DET_MASTER_PORT"]
+    use_tls = distutils.util.strtobool(os.environ.get("DET_USE_TLS", "false"))
+
+    master_url = f"http{'s' if use_tls else ''}://{master_addr}:{master_port}"
+    certs.cli_cert = certs.default_load(master_url=master_url)
```
stoksc, line 17: if this preamble is going to appear in many startup scripts, it may be nice to library-ize it. a mistake here that only affects specific customer envs seems easy to write and it would be high impact.

stoksc (approved):
lgtm, behavior-wise seems to basically be a noop, the only order change is now we dl the model into the container after we pip install determined but that shouldn't matter. i'll prob pr the otherwise shortly, to not ignore that ws error.

harness/determined/exec/fetch_context.py
rb-determined-ai: I agree.  I was going to take that on as part of the Generic API work, if that's ok with you.

rb-determined-ai:
All important tests have passed, landing this now.

PR #2754, opened 2021-07-20
fix: k8s available GPUs indicator [DET-5808]

eecsliu:
## Description
The kubernetes GPU number displayed incorrect numbers, seemingly subtracting numbers from the total number of GPUs instead of just from the available number. This was due to kubernetes counting slots differently from IDs, where IDs looped for every pod. Patched by making the ID == slot count. 

## Test Plan
Spin up a kubernetes cluster with autoscaling enabled (https://docs.determined.ai/latest/how-to/installation/setup-gke-cluster.html?highlight=gke). Make sure that the number of slots per node > 1. Start a few experiments and make sure that the number of available GPUs is correct on the dashboard. Also make sure that the number of allocated GPUs on the cluster page also looks correct. 

## Checklist

### DIFF
--- a/master/internal/kubernetes/pods.go
+++ b/master/internal/kubernetes/pods.go
@@ -479,7 +479,7 @@ func (p *pods) summarize(ctx *actor.Context) map[string]model.AgentSummary {
 				}
 
 				slotsSummary[strconv.Itoa(curSlot)] = model.SlotSummary{
-					ID:        strconv.Itoa(i),
+					ID:        strconv.Itoa(curSlot),
 					Device:    device.Device{Type: deviceType},
 					Enabled:   true,
 					Container: podInfo.container,
--- a/master/internal/kubernetes/pods.go
+++ b/master/internal/kubernetes/pods.go
@@ -479,7 +479,7 @@ func (p *pods) summarize(ctx *actor.Context) map[string]model.AgentSummary {
 				}
 
 				slotsSummary[strconv.Itoa(curSlot)] = model.SlotSummary{
-					ID:        strconv.Itoa(curSlot),
+					ID:        strconv.Itoa(i),
 					Device:    device.Device{Type: deviceType},
 					Enabled:   true,
 					Container: podInfo.container,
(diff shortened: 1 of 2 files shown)


RehanSD (approved):
LGTM!

eecsliu:
Another place to put a similar code change would be in `master/pkg/model/agent.go:26` since that is where the code removes the data in the first place. I think the current location is better because it won't affect summaries from the agent architecture, only kubernetes.

ioga (approved):
Thank you for the fix. Re: https://github.com/determined-ai/determined/blob/master/master/pkg/model/agent.go#L26 , should we perhaps explicitly fail or log a warning if there're duplicate "slot ids" in there?

It feels like that serialization code alters the input data in a semantically wrong way, I am not sure there're any legit circumstances where that function will have "duplicate" slots it needs to dedupe, and so maybe it should do that at all and defensively check against it.

eecsliu:
I'm not sure it's something that failing or warning would help solve, since it's invisible to the user anyways. At that point, I think it's probably better to just fix the bug at the serialization step (change the dictionary to use an int key vs the slot ID as a key).

ioga:
It's invisible to the user, but it'll catch our own programming errors.

Regarding rewriting of slot ids at serialization step: that also feels like a data change happening in a serialization. Also, what if we ever add some APIs that operate on the slot *ids*? e.g. schedule my task in that particular slot id because that slot is associated with a particular GPU I want to investigate for a hardware issue. For that to work, preserving slot ids from their generation (in depths of resource managers) would be important.

eecsliu:
It sounds like fixing in the serialization step is the right move then.

> that also feels like a data change happening in a serialization
The real slot ID will now get preserved since it should be part of the object being serialized. The key is now just an integer from the range function.

ioga:
Ideally, `SlotsSummary` then should be a list. 
Other types out there such as https://github.com/determined-ai/determined/blob/master/master/pkg/model/agent.go#L41 use object ids for keys. So `SlotsSummary` type using map as a list would be inconsistent, within the realm of our "web" apis.

eecsliu:
My mistake - I was wrong about the the key of the dictionary. It would just be the same as whatever the keys of the dictionary it's loading from. It looks like in the case of agents it's an address, whereas in kubernetes (which is what I had referenced) is simply an integer converted to a string value. And the kubernetes code is subject to change since I'll be working on it.

PR #2755, opened 2021-07-21
chore: don't drop agent ws failures

stoksc:
## Description
This change updates the agent actor in the master to at least log ws write failures.

## Commentary (optional)
This could, in some cases, get kinda noisy.. that is my only concern. But also not writing these made a bug much harder to fine (I found it quickly mostly by luck).

## Checklist

### DIFF
--- a/master/internal/agent/agent.go
+++ b/master/internal/agent/agent.go
@@ -85,7 +85,7 @@ func (a *agent) Receive(ctx *actor.Context) error {
 
 		wsm := ws.WriteMessage{Message: aproto.AgentMessage{StartContainer: &msg.StartContainer}}
 		if err := ctx.Ask(a.socket, wsm).Error(); err != nil {
-			// TODO(Brad): After push arch, return and handle this error when starting allocations.
+			// TODO(DET-5862): After push arch, return and handle this error when starting allocations.
 			ctx.Log().WithError(err).Error("failed to write start container message")
 		}
 		ctx.Tell(a.slots, msg.StartContainer)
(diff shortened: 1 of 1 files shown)


master/internal/agent/agent.go
```diff
@@ -60,23 +60,34 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
 		ctx.Log().Infof("starting container id: %s slots: %d task handler: %s",
 			msg.StartContainer.Container.ID, len(msg.StartContainer.Container.Devices),
 			msg.TaskActor.Address())
 
-		ctx.Ask(a.socket, ws.WriteMessage{Message: aproto.AgentMessage{
-			StartContainer: &msg.StartContainer,
-		}})
+		wsm := ws.WriteMessage{Message: aproto.AgentMessage{StartContainer: &msg.StartContainer}}
+		if err := ctx.Ask(a.socket, wsm).Error(); err != nil {
+			// TODO(Brad): After push arch, return and handle this error when starting allocations.
```
stoksc, line 88: once i wont just conflict myself, i'm going to make the Tell behind alloc.Start an ask and return allocation start errors to the task actor, so it can decide how to act?

shiyuann: Create a ticket for this plz

shiyuann (approved):

master/internal/agent/agent.go
stoksc:
```suggestion
			// TODO(DET-5862): After push arch, return and handle this error when starting allocations.
```

PR #2757, opened 2021-07-21
fix: fetching trial undefined wait message

hamidzr:
## Commentary (optional)

What messaging would be good when we’re waiting on the trial to appear? `Waiting for the trial` (since we know it’s a single trial) or `waiting for the trial to appear` something else?

there should be a way that the tscompiler can warn about these 🤔 
https://github.com/microsoft/TypeScript/issues/30239

## Checklist


EmilyBonar (approved):
Looks good to me, although I would maybe just say `Waiting for trial...`, leaving out the `the`. Personal preference, up to you.

PR #2759, opened 2021-07-21
fix: /api/v1/experiments/:id/cancel should cancel not kill trials

stoksc:
## Description
This change fixes the new api's cancel endpoint to actually cancel experiments.

## Test Plan
- [x] write a test

## Commentary (optional)
This was broken but tests couldn't catch because the old api worked fine.

## Checklist


hamidzr (approved):
thanks for fixing this 🚢

PR #2761, opened 2021-07-21
fix: add reload route to allow remounting of same pages [DET-5818]

hkang1:
## Description

PROBLEM:
During a reroute that mounts to the same page component, the page component does not unmount. For example, when going from `/experiments/1` to `/experiments/2`, both of these pages mount to `ExperimentDetail` page component and so upon a successful fork, when rerouting to the newly forked experiment, the new page will show the stale data from the previous experiment until the new experiment data loads (if at all). 

SOLUTION:
Create a `/reload` route that takes a `path` query param that we can route to to force an unmount of the component. This essentially "refreshes" the page. This is the preferred approach over a `window.location` reload for a few reasons:
- we don't lose or mess with history we accumulate on our end
- using `window.location` in conjunction with our internal history causes our internal history to diverge from the native history
- we avoid attempting to reload the entire page again 

## Test Plan

- [ ] check that an experiment fork loads a new experiment page without stale data (such as metrics)
- [ ] check that continuing a trial loads a new experiment page without stale data

## Checklist

### DIFF
--- a/docs/release-notes/2761-fix-fork-stale-experiment.txt
+++ b/docs/release-notes/2761-fix-fork-stale-experiment.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  WebUI: fix the issue of new experiments from fork and continue trial
+   showing outdated data.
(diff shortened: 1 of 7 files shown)


hamidzr (approved):
LGTM. 

We also have another [WIP branch](https://github.com/determined-ai/determined/pull/2763) that addresses this in a different way but that's doesn't fully cover the issue yet.

PR #2762, opened 2021-07-21
fix: allow full-sized model definitions through grpc

rb-determined-ai:


e2e_tests/tests/experiment/test_noop.py
```diff
@@ -276,9 +276,9 @@ def test_startup_hook() -> None:
 def test_large_model_def_experiment() -> None:
     with tempfile.TemporaryDirectory() as td:
         shutil.copy(conf.fixtures_path("no_op/model_def.py"), td)
-        # Write a 94MB file into the directory.
-        with open(os.path.join(td, "junk.txt"), "w") as f:
-            f.write("x" * 94 * 1024 * 1024)
+        # Write a 94MB file into the directory.  Use /dev/urandom because it is not compressible.
```
dzhu, line 279: Just to be real pedantic, the docs say `os.urandom` will use the `getrandom` syscall on Linux rather than literally reading from `/dev/urandom`. May as well just say "Use random data", which is the important part.

dzhu (approved):

e2e_tests/tests/experiment/test_noop.py
rb-determined-ai: done.

rb-determined-ai:
All tests that ran have passed, merging.

Issue #1583, opened 2020-11-12
specification of default configs for notebooks created through the web UI


zjorgensenbits:
Feature request: it would be really helpful to be able to specify one or more default configs for the notebook instances that are used when launching notebooks with the button in the web UI. In our situation, being able to specify default mounts and agent labels would be helpful, since we typically mount and NFS share for notebook persistence.

It would be really nice if we could have several presets that are selectable from the dropdown when you click on that button, similarly to how you can select "cpu only" currently.

neilconway:
@zjorgensenbits Thanks for writing this up! I agree that this would be a useful feature. We've started thinking about the best way to implement it -- will keep you posted if we have any updates on the timeline here.

zjorgensenbits:
Any thoughts/progress on this?

vishnu2kmohan:
Update: We'll add the ability to pick a configuration template to the notebook launch dialog and post an update once this feature ships.

zjorgensenbits:
That would be super helpful, thanks!

vishnu2kmohan:
Just a note that we're going to start working on this effort soon @zjorgensenbits, and will update this issue when it's ready.

hamidzr:
Hi, @zjorgensenbits  and thanks for sharing your feedback. We [recently](https://docs.determined.ai/latest/release-notes.html#version-0-16-0) shipped a new and more configurable way of launching Jupyeter Notebooks that allows you to pick a [template](https://docs.determined.ai/latest/reference/config-template.html?highlight=templates) for the Notebook configuration you're running. Does this address what you've expressed? 

![2021-06-21_15-00-24](https://user-images.githubusercontent.com/12127420/122833216-6f12c200-d2a1-11eb-9e5c-ff9302874148.png)

zjorgensenbits:
I will check it out when we update to this version and let you know.  But, based on the screenshot, this is exactly what I had in mind!

vishnu2kmohan:
Hey @zjorgensenbits have you had a chance to upgrade and test 0.16.x out?

Any feedback?

And if it's what you and/or your team were after, may we please close out this issue?

hamidzr:
I'll close this for now but please feel free to reopen if there are any questions or updates.

zjorgensenbits:
Finally had a chance to upgrade and test this and it works great -- exactly what I was hoping for. 
[...]

PR #2765, opened 2021-07-22
refactor: fork and continue trial [DET-5817]

hkang1:
## Description

- Use a hook to capture reusable code around creating experiments, specifically experiment Fork and Continue Trial. Enscapsulate everything that's needed for the Fork and Continue Trial flow to avoid requiring multiple dependency.
- Remove the need for an imperative reference handle (generally an anti-pattern and can be avoided).
- Update the code to use `antd` generative modals to avoid having persisted modals, so the modals can be used in any component with just an import of the hook, without needing to write any JSX or helper code within where the modal is needed.

## Test Plan

- [ ] check that multi-trial experiment fork works
- [ ] check that multi-trial continue trial works
- [ ] check that single trial experiment fork works
- [ ] check that single trial experiment continue trial works
 

## Checklist


webui/react/src/hooks/useCreateExperimentModal.module.scss
```diff
@@ -0,0 +1,42 @@
+.base:global(.ant-modal) {
```
hkang1, line 1: this seems like the most logical place to but css related to the hook... open to other locations

webui/react/src/pages/ExperimentDetails/ExperimentSingleTrialTabs.tsx
```diff
@@ -84,8 +81,10 @@ const ExperimentSingleTrialTabs: React.FC<Props> = (
         { id: experiment.id, limit: 2 },
         { signal: canceler.signal },
       );
-      if (expTrials.trials[0]) {
-        setFirstTrialId(expTrials.trials[0].id);
+      const firstTrial = expTrials.trials[0];
+      if (firstTrial) {
```
hamidzr, line 85: nit: explicitly check against undefined? 0 could be a trial id sometime although it'd not be a valid item in the trials list now. feel free to ignore

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -65,49 +61,19 @@ const ExperimentDetails: React.FC = () => {
...
   const showForkModal = useCallback((): void => {
-    if (experiment?.configRaw) {
-      const rawConfig: RawJson = clone(experiment.configRaw);
-      if (rawConfig.description) rawConfig.description = `Fork of ${rawConfig.description}`;
-      upgradeConfig(rawConfig);
-      setForkModalConfig(rawConfig);
-    }
-    setIsForkModalVisible(true);
-  }, [ experiment?.configRaw ]);
+    if (!experiment) return;
```
hamidzr, line 64: suggestion: make showForkModal optional simliar to show continue

webui/react/src/hooks/useCreateExperimentModal.tsx
```diff
@@ -0,0 +1,350 @@
...
+  ContinueTrial = 'Continue Trial',
+}
+
+interface ShowProps {
+  experiment: ExperimentBase;
+  trial?: TrialDetails;
+  type: CreateExperimentType;
+}
+
+interface ModalState {
```
hamidzr, line 28: 👍🏼

webui/react/src/hooks/useCreateExperimentModal.tsx
```diff
@@ -0,0 +1,350 @@
...
+        parentId: modalState.experiment.id,
+      });
+
+      closeModal();
+
+      // Route to reload path to forcibly remount experiment page.
+      const newPath = paths.experimentDetails(newExperimentId);
+      routeToReactUrl(paths.reload(newPath));
+    } catch (e) {
+      let errorMessage = 'Unable to continue trial with the provided config.';
```
hamidzr, line 152: would this be triggered for fork modals as well?

hamidzr (commented):
reviewed the code it looks great, good work. I added my comments. I'll visually test it next. 

what do you think it would take to support notebook modals through the same hook? could be a modal to handle submitting a YAML file with an alternative form visualization. Not having to redo the same logic in another place should reduce the chance of hitting issues like [this](https://github.com/determined-ai/determined/issues/1583#issuecomment-885784163)

webui/react/src/hooks/useCreateExperimentModal.tsx
hkang1: ah yeah would have to distinguish the two, nice catch!

webui/react/src/pages/ExperimentDetails/ExperimentSingleTrialTabs.tsx
hkang1: not 100% sure but I don't it's possible for `expTrials.trials` to be something other than an array, and if [0] exists, it is going to be an object instead of just a trial id or something to that effect.

hkang1:
> what do you think it would take to support notebook modals through the same hook? could be a modal to handle submitting a YAML file with an alternative form visualization. Not having to redo the same logic in another place should reduce the chance of hitting issues like [this](https://github.com/determined-ai/determined/issues/1583#issuecomment-885784163)

👍  I think this is a good idea. A form/config modal hook. They both follow similar enough of a flow / pattern.

hamidzr (changes requested):
I did a visual inspection:  continue trial functions as expected ✅ 

fork experiment flow (form or full config) doesn't trigger a fork for me. I see an `undefined` logged to the console with error level. 

Also this for single trial experiments.
DeepinScreenshot_select-area_20210727155358

hkang1:
> I did a visual inspection: continue trial functions as expected ✅
> 
> fork experiment flow (form or full config) doesn't trigger a fork for me. I see an `undefined` logged to the console with error level.
> 
> Also this for single trial experiments.
> DeepinScreenshot_select-area_20210727155358

Oops, sorry I forgot to change one part when I was changing from `isContinueTrial` to `isFork`

hamidzr (approved):
🚢

PR #2766, opened 2021-07-22
feat: master yaml templates for `det deploy aws|gcp` [DET-5766]

ioga:
## Description

Allow users to use custom master configuration in `det deploy aws|gcp`. See the docs for usage.

This change is adding another level of indirection to the master instance startup scripts in AWS CloudFormation and GCP terraform. Previously, we've used to generate the master config directly in the startup script. With this change, startup script generates a "template context", i.e. input data for the config template, which users may override. 

To render the template, this change adds a new binary `determined-gotmpl` into the `determined-master` container builds. It's a thin wrapper around golang templates + a few helpers.

**Breaking change:** since the new startup scripts require `determined-gotmpl` in the image, the new `det deploy aws|gcp` will not be able to deploy previous versions of determined.

Since we have a bunch of CF templates, and keeping them in sync is a chore, added a small python utility / Makefile target `aws-master-config-inject` to streamline the template updates.

Added `det master config -o yaml` option to allow printing the master config in yaml format instead of json to make it more easily reusable when writing new config files.

## Test Plan

- Deploy gcp and aws clusters of various types with and without custom `--master-config-template-path` setting.
- Unit test for `determined-gotmpl` is included.

## Checklist


shiyuann:
Will talk offline about this PR. Have some high-level comments.

docs/how-to/installation/aws.txt
```diff
@@ -286,6 +294,84 @@ Listing Clusters
       -  AWS profile used for deploying cluster resources.
       -  default
 
+Printing the default ``master.yaml`` template
+---------------------------------------------
+
+.. code::
+
+   det deploy aws dump-master-yaml-template
```
katport, line 302: I think this should be dump-master-config-template

docs/how-to/installation/aws.txt
```diff
@@ -286,6 +294,84 @@ Listing Clusters
...
+
+Example workflow:
+
+#. Get the default template using
+
+   .. code::
+
+      det deploy aws dump-master-config-template > /path/to/master.yaml.tmpl
+
+#. Customize the template as you see fit by editing it in any text
```
katport, line 327: Lets add a link on the resource pool page that links to this section.

docs/how-to/installation/aws.txt
```diff
@@ -286,6 +294,84 @@ Listing Clusters
...
+          provider:
+            instance_type: {{ .resource_pools.pools.compute_pool.instance_type }}
+            cpu_slots_allowed: true
+            {{- toYaml .resource_pools.aws | nindent 6}}
+
+   Then, append a new section:
+
+   .. code:: yaml
+
+      - pool-name: compute-pool-solo
```
katport, line 357: should be pool_name

master/cmd/determined-gotmpl/main.go
```diff
@@ -0,0 +1,149 @@
...
+
+	data := map[string]interface{}{} // sprig dict functions require map[string]interface{}
+	files := []string{}
+
+	if output != "" {
+		p, err := filepath.Abs(output)
+		if err != nil {
+			log.Fatal(err)
+		}
+		data["Name"] = strings.SplitN(filepath.Base(p), ".", 2)[0]
```
stoksc, line 102: 🤷 
```suggestion
		data["Name"] = strings.TrimSuffix(filepath.Base(p), filepath.Ext(filepath.Base(p)))
```

master/cmd/determined-gotmpl/main.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	if len(files) == 0 {
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	b := RunTmpl(data, files)
+
```
stoksc, line 140: I prefer like, `out := os.Stdout; if output != nil { f, err := os.Open(output)  ... out = f }` .. but not huge. Same for input, mainly to use io.Reader/io.Writer everywhere to not duplicate parsing/outputting logic.

master/cmd/determined-gotmpl/main.go
```diff
@@ -0,0 +1,149 @@
...
+	return b
+}
+
+func main() {
+	flag.Usage = func() {
+		fmt.Fprint(flag.CommandLine.Output(),
+			`Usage: determined-gotmpl [-i data.yaml] [-o output.txt] [key=value] template [template ...]
+
+Based on https://github.com/NateScarlet/gotmpl
```
stoksc, line 84: why not just use gotmpl, as opposed to this binary? i guess there is some customization in here, what does it not support?

harness/determined/deploy/gcp/terraform/modules/compute/main.tf
```diff
@@ -201,14 +131,23 @@ resource "google_compute_instance" "master_instance" {
 
     docker network create ${var.master_docker_network}
 
+    touch /usr/local/determined/etc/master.yaml
+    docker run \
```
stoksc, line 135: since this overhead is on every new install, we should make sure the image is very small even though it's short in comparison to the rest

stoksc (commented):
go code looks fine, even though it isn't super clean (it's very small). this solution logically follows from the problem to me, like, we need to just pass cf/tf a master config, but there is state we don't have at the time the user runs det deploy so we need to template it somewhere, and it has to be where we have all that state, and that has to be while cf/tf is running so in the metadata script or part of the cf/tf (i think tf could just handle this with https://registry.terraform.io/providers/hashicorp/template/latest/docs/data-sources/file but not sure cf can). to propose an alternative, we could have a base cf/tf script that rolls out all the dependencies and queries the state, then does the templating in det deploy and just launches the final master instance with all the information we need, templating it in their with jinja or something. This is slightly beneficial longer term too, that deployment model is probably better suited to no-downtime upgrades. feel free to ping offline if you have thoughts

docs/how-to/installation/aws.txt
```diff
@@ -286,6 +294,84 @@ Listing Clusters
...
+---------------------------------------------
+
+.. code::
+
+   det deploy aws dump-master-yaml-template
+
+.. _aws-master-yaml-template:
+
+******************************
+ Custom master.yaml templates
```
shiyuann, line 307:
```suggestion
 Custom master.yaml templates with ``det deploy aws``
```

shiyuann, line 307: This should lie under  ``det deploy`` tool because the document only works with ``det deploy aws``

docs/how-to/installation/gcp.txt
```diff
@@ -306,6 +314,85 @@ If you deployed with a service account JSON key file, the same
 credentials file will be used for deprovisioning. Otherwise, default
 Google Cloud credentials are used.
 
+.. _gcp-master-yaml-template:
+
+******************************
+ Custom master.yaml templates
```
shiyuann, line 320: same

docs/release-notes/2766-det-deploy-templates.txt
```diff
@@ -0,0 +1,15 @@
+:orphan:
+
+**New Features**
+
+-  Add the ability to use customizable master configuration templates in
+   ``det deploy aws|gcp``.
+
+   -  **Breaking Change:** This version of ``det deploy`` will not be
```
shiyuann, line 8: These two breaking changes should be first-level.

shiyuann:
I don't have any other major comments. It looks good to me. You can assign it to @stoksc next time.

master/cmd/determined-gotmpl/main.go
ioga: that is mostly `gotmpl`, so if I can I'll keep it as is.

master/cmd/determined-gotmpl/main.go
ioga: yaml input and `toYaml` helper, mostly. added a comment.

harness/determined/deploy/gcp/terraform/modules/compute/main.tf
ioga: it's another binary in the same image, so it's slightly larger but we don't pull 2 images

docs/how-to/installation/gcp.txt
ioga: this file is structured differently, so for better or for worse that seems to be correct here.

ioga:
>to propose an alternative, we could have a base cf/tf script that rolls out all the dependencies and queries the state, then does the templating in det deploy and just launches the final master instance with all the information we need, templating it in their with jinja or something. This is slightly beneficial longer term too, that deployment model is probably better suited to no-downtime upgrades. feel free to ping offline if you have thoughts

Shiyuan has suggested a similar approach (if not the same) in his original slack message between three of us. My critique was that "I wanted to preserve an ability to go into CF console and change variables and have it still work."
That is, still allow raw cloudfront or terraform work without `det deploy` as a necessary orchestrator.

stoksc:
I mean, that one requires a script to query cf/tf state and make a vars file, this requires a templating binary and a to submit everything in the form of a template. From the user's point of view, for this I need to retrieve/learn/deal with the template, the other I may need to retrieve/learn/deal with parts of cf/tf. ..hmm. Either way, the user will give us a master.yaml. If it is a master.yaml.tpl, the user is aware of what we are going to substitute, if we just take a master.yaml, we then have to decide the correct way to merge the "template context" that exists from the cloud provider no matter way. On that basis, I think I am in favor of the templating, though I wish it could happen earlier (not in the metadata_startup_script of the ec2).

harness/determined/deploy/gcp/cli.py
```diff
@@ -131,6 +147,17 @@ def handle_up(args: argparse.Namespace) -> None:
     return deploy_gcp("up", args)
 
 
+def handle_dump_master_config_template(args: argparse.Namespace) -> None:
```
stoksc, line 150: i think we could drop this and just use https://registry.terraform.io/providers/hashicorp/local/latest/docs/data-sources/file, which would be a lot nicer.

stoksc (commented):
looks great. two minor cleanup comments and i wanna take it for a whirl but should be good to approve after that.

PR #2767, opened 2021-07-22
chore: remove stale install extras from harness python package.

ioga:
## Description

Install extras are stale and see little use. Instead of keeping this extra maintenance burden, let's get rid of them.

## Test Plan

The python package can still be installed.

## Checklist

### DIFF
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -58,16 +58,6 @@ setup(
         # so we expose that requirement here.
         "websocket-client<1",
     ],
-    extras_require={
-        "tf-115-cuda102": ["tensorflow-gpu==1.15.5"],
-        "tf-115-cpu": ["tensorflow==1.15.5"],
-        "tf-240-cuda102": ["tensorflow-gpu==2.4.1"],
-        "tf-240-cpu": ["tensorflow==2.4.1"],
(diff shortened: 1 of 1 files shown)


mackrorysd:
I personally have no objection. This feature requires a user to basically specify a minor version of dependencies to have us install them, when they could also have the package manager install them by specifying the same minor version. Unless the combination of Determined and specific versions of these libraries triggers OTHER dependencies where we do want to be more involved, I see no benefit. These have gotten out of date multiple times and no one has, as far as I can tell, ever actually noticed unless they were doing a search / replace to bump dependencies.

I would like to give @rb-determined-ai and @hamidzr a chance to weigh in because I know they have some context too.

mackrorysd (approved):
+1 (unless Ryan or Hamid has something to add)

hamidzr (approved):
LGTM. This could potentially break user's use cases but I don't see use advertising for this use anywhere. Should we add a release note?

azhou-determined (approved):

rb-determined-ai:
I'm totally ok with this.  I have never heard of a customer telling me they've installed determined with one of these extras.  

What would an amle say?  @katport, do people actually use these?

katport (approved):

PR #2769, opened 2021-07-22
feat: add `--config` override to `det e create`. [DET-5786]

ioga:
## Description

Add `--config` override to `det e create`, similar to `det notebook` or `det shell`. Now you can change experiment config in ad-hoc fashion, e.g. `det e create const.yaml . --config name=test-final-hack-9`

Also, remove a little bit of dead code in local mode that is irrelevant now that we've merged all python packages into one.

## Test Plan

Added an e2e test.

## Checklist

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -18,7 +18,7 @@ import determined.load
 from determined import _local_execution_manager
 from determined.cli import checkpoint, render
 from determined.cli.command import CONFIG_DESC, parse_config_overrides
-from determined.common import api, constants, context, util, yaml, set_logger
+from determined.common import api, constants, context, set_logger, util, yaml
 from determined.common.api import authentication
 from determined.common.declarative_argparse import Arg, Cmd, Group
 from determined.common.experimental import Determined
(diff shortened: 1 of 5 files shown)


harness/determined/cli/experiment.py
```diff
@@ -9,12 +9,15 @@
 from argparse import FileType, Namespace
 from pathlib import Path
 from pprint import pformat
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 import tabulate
 
-import determined.common
+import determined
```
rb-determined-ai, line 16: Can we stick with the cannonical `import determined as det`?

rb-determined-ai (approved):

harness/determined/cli/experiment.py
ioga: Since we use an "underscored" thing from `import determined`, which doesn't look like a canonical user-facing thing, I've switched it to a direct function import.

ioga:
admin merging due to https://github.com/determined-ai/determined/pull/2781

PR #2775, opened 2021-07-22
feat: experiment delete

### LINKED ISSUE #2752: Add support for experiment DELETE on the WebUI
Blocked on this [PR](https://github.com/determined-ai/determined/pull/2741) landing.

**NOTE** Experiment delete is only available to Admin users.

Experiment List page
- add `Delete` option to overflow menu for each experiment row
- show modal delete confirmation and send Rest API call upon confirmation. The state should change to `DELETING` upon a successful call.

Experiment Detail page
- add `Delete` option to experiment action bar in the header (within the overflow menu).
- show modal delete confirmation and send Rest API call upon confirmation. The state should change to `DELETING` upon a successful call.

https://determinedai.atlassian.net/browse/DET-5244

hamidzr:
## Commentary (optional)

- [x] would this be an admin only feature?
- tradeoff discussion https://determined-community.slack.com/archives/C025A98H82X/p1627494839214600

closes #2752 

## Checklist

### DIFF
--- a/webui/react/src/components/TaskActionDropdown.tsx
+++ b/webui/react/src/components/TaskActionDropdown.tsx
@@ -1,6 +1,7 @@
 import { isNumber } from 'util';
 
-import { Dropdown, Menu } from 'antd';
+import { ExclamationCircleOutlined } from '@ant-design/icons';
+import { Dropdown, Menu, Modal } from 'antd';
 import { MenuInfo } from 'rc-menu/lib/interface';
 import React from 'react';
 
@@ -8,10 +9,10 @@ import Icon from 'components/Icon';
 import handleError, { ErrorLevel, ErrorType } from 'ErrorHandler';
 import { paths } from 'routes/utils';
 import {
-  activateExperiment, archiveExperiment, cancelExperiment, killExperiment, killTask,
-  openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
+  activateExperiment, archiveExperiment, cancelExperiment, deleteExperiment, killExperiment,
+  killTask, openOrCreateTensorboard, pauseExperiment, unarchiveExperiment,
 } from 'services/api';
-import { AnyTask, CommandTask, ExperimentTask, RunState } from 'types';
+import { AnyTask, CommandTask, DetailedUser, ExperimentTask, RunState } from 'types';
 import { capitalize } from 'utils/string';
 import { isExperimentTask } from 'utils/task';
 import { cancellableRunStates, isTaskKillable, terminalRunStates } from 'utils/types';
@@ -24,6 +25,7 @@ export enum Action {
   Activate = 'activate',
   Archive = 'archive',
   Cancel = 'cancel',
+  Delete = 'delete',
   Kill = 'kill',
   Pause = 'pause',
   Tensorboard = 'tensorboard',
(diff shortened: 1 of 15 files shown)


webui/react/src/components/TaskCard.tsx
```diff
@@ -6,14 +6,16 @@ import Icon from 'components/Icon';
...
 import { percent } from 'utils/number';
 import { canBeOpened, isExperimentTask } from 'utils/task';
 import { openCommand } from 'wait';
 
 import css from './TaskCard.module.scss';
 
-const TaskCard: React.FC<RecentTask> = (props: RecentTask) => {
+type Props = AnyTask & RecentEvent & {curUser?: DetailedUser}
+
+const TaskCard: React.FC<Props> = (props: Props) => {
```
EmilyBonar, line 18: nit: destructure props to maintain consistency?

webui/react/src/pages/Dashboard.tsx
```diff
@@ -200,7 +200,10 @@ const Dashboard: React.FC = () => {
         {filteredTasks.length !== 0
           ? (
             <Grid gap={ShirtSize.medium} mode={GridMode.AutoFill}>
-              {filteredTasks.map((props: RecentTask) => <TaskCard key={props.id} {...props} />)}
+              {filteredTasks.map((props: RecentTask) => <TaskCard
+                key={props.id}
+                {...props}
+                curUser={user} />)}
```
EmilyBonar, line 206: nit: reorder props
```
{filteredTasks.map((props: RecentTask) => <TaskCard
    curUser={user}
    key={props.id}
    {...props} />)}
```

webui/react/src/components/TaskActionDropdown.tsx
```diff
@@ -90,6 +95,22 @@ const TaskActionDropdown: React.FC<Props> = ({ task, onComplete }: Props) => {
           if (!isExperiment) break;
           await unarchiveExperiment({ experimentId: id });
           if (onComplete) onComplete(action);
+          break;
+        case Action.Delete:
+          if (!isExperiment) break;
+          Modal.confirm({
```
EmilyBonar, line 101: you could make this modal a component since it looks like you use it twice, here and in ExperimentDetailsHeader

EmilyBonar (approved):
Looks good, just a few comments and nothing breaking

webui/react/src/components/TaskCard.tsx
hamidzr: do you mean to separate curUser out of props so it is a proper Task or to destructure everything

webui/react/src/components/TaskActionDropdown.tsx
hamidzr: I thought about that but then in order to make it a bit more useful, I would have needed to expose basically the same options Modal.confirm takes so I went back to using it directly. We're gonna be using confirm modals more but in order to be able to use it we'd have to expose most of these options. wdy think?

webui/react/src/components/TaskCard.tsx
hamidzr: https://github.com/determined-ai/determined/pull/2775/commits/39667be37ec27ad445ad3587fefa4ede006e12b6

webui/react/src/components/TaskActionDropdown.tsx
EmilyBonar: Yeah, that's true. I think as long as the modals stay small like this it's fine, and just extract them into a component if they start getting their own logic.

webui/react/src/components/TaskCard.tsx
EmilyBonar: I meant destructure everything, but this is fine imo.

PR #2786, opened 2021-07-24
refactor: settings data flow [DET-5625]

hkang1:
## Description

This body of work focus on revamping the user settings on a per page basis. Currently, on a per page-basis, there is a lot of code to perform the following:
- encode and decode settings from the query string
- read/write from storage
- side effects to update state based on storage and query string

Goals of this PR:
- Remove the need for a bunch of boiler plate code for settings
- Convert the settings data flow to be in one direction to avoid the need for complex logic
- Extract a config file per page in order to allow for a simple hook inject + config to support settings seamlessly
- Apply to just the Task List page to see the effects of this PR on a lower impact page

What's happening under the hood:

READ:
detect query string changes => page state (query param > persistent storage > default value) => render (and side effects such as API calls based on page state)

WRITE:
user interaction => update persistent storage (if applicable) => update query string (to trigger READ flow)

## Test Plan

On the Task List page
- [ ] check to make sure table sorting works on all the sortable columns (columns, asc, desc)
- [ ] check that table page size change works
- [ ] check that table pagination works (changing different table page offset)
- [ ] check that table filters works (state, type, user, name search)
- [ ] check that the change the table filters are saved into local storage (change the filters, navigate to another page, navigate back to see that the filters are preserved)
- [ ] check that copying and pasting the task list page URL will reproduce the page state reliably

## Checklist

### DIFF
--- a/webui/react/src/components/NavigationSideBar.tsx
+++ b/webui/react/src/components/NavigationSideBar.tsx
@@ -43,8 +43,14 @@ const settingsConfig: SettingsConfig = {
 const menuConfig = {
   bottom: [
     { icon: 'logs', label: 'Master Logs', path: paths.masterLogs() },
-    { icon: 'docs', label: 'Docs', path: paths.docs(), popout: true },
-    { icon: 'cloud', label: 'API (Beta)', path: paths.docs('/rest-api/'), popout: true },
+    { external: true, icon: 'docs', label: 'Docs', path: paths.docs(), popout: true },
+    {
+      external: true,
+      icon: 'cloud',
+      label: 'API (Beta)',
+      path: paths.docs('/rest-api/'),
+      popout: true,
+    },
   ],
   top: [
     { icon: 'dashboard', label: 'Dashboard', path: paths.dashboard() },
--- a/webui/react/src/components/NavigationSideBar.tsx
+++ b/webui/react/src/components/NavigationSideBar.tsx
@@ -3,8 +3,8 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { useLocation } from 'react-router-dom';
 import { CSSTransition } from 'react-transition-group';
 
-import { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
-import useStorage from 'hooks/useStorage';
+import { useStore } from 'contexts/Store';
+import useSettings, { BaseType, SettingsConfig } from 'hooks/useSettings';
 import { paths } from 'routes/utils';
 import { ResourceType } from 'types';
 
@@ -20,10 +20,41 @@ interface ItemProps extends LinkProps {
   icon: string;
   label: string;
   status?: string;
+  tooltip?: boolean;
 }
 
+interface Settings {
+  navbarCollapsed: boolean;
+}
+
+const settingsConfig: SettingsConfig = {
+  settings: [
+    {
+      defaultValue: false,
+      key: 'navbarCollapsed',
+      skipUrlEncoding: true,
+      storageKey: 'navbarCollapsed',
+      type: { baseType: BaseType.Boolean },
+    },
+  ],
+  storagePath: 'navigation',
+};
+
+const menuConfig = {
+  bottom: [
+    { icon: 'logs', label: 'Master Logs', path: paths.masterLogs() },
+    { icon: 'docs', label: 'Docs', path: paths.docs(), popout: true },
+    { icon: 'cloud', label: 'API (Beta)', path: paths.docs('/rest-api/'), popout: true },
+  ],
+  top: [
+    { icon: 'dashboard', label: 'Dashboard', path: paths.dashboard() },
+    { icon: 'experiment', label: 'Experiments', path: paths.experimentList() },
+    { icon: 'tasks', label: 'Tasks', path: paths.taskList() },
+    { icon: 'cluster', label: 'Cluster', path: paths.cluster() },
+  ],
+};
+
 const NavigationItem: React.FC<ItemProps> = ({ path, status, ...props }: ItemProps) => {
-  const { ui } = useStore();
   const location = useLocation();
   const [ isActive, setIsActive ] = useState(false);
   const classes = [ css.navItem ];
@@ -41,21 +72,17 @@ const NavigationItem: React.FC<ItemProps> = ({ path, status, ...props }: ItemPro
     </Link>
   );
 
-  return ui.chromeCollapsed ? (
+  return props.tooltip ? (
     <Tooltip placement="right" title={props.label}><div>{link}</div></Tooltip>
   ) : link;
 };
-
-const STORAGE_KEY = 'collapsed';
 
 const NavigationSideBar: React.FC = () => {
   // `nodeRef` padding is required for CSSTransition to work with React.StrictMode.
   const nodeRef = useRef(null);
   const { auth, cluster: overview, ui } = useStore();
-  const storeDispatch = useStoreDispatch();
-  const storage = useStorage('navigation');
-  const [ isCollapsed, setIsCollapsed ] = useState(storage.getWithDefault(STORAGE_KEY, false));
   const [ showNotebookModal, setShowNotebookModal ] = useState(false);
+  const { settings, updateSettings } = useSettings<Settings>(settingsConfig);
 
   const showNavigation = auth.isAuthenticated && ui.showChrome;
   const version = process.env.VERSION || '';
@@ -66,15 +93,8 @@ const NavigationSideBar: React.FC = () => {
     undefined : `${overview[ResourceType.ALL].allocation}%`;
 
   const handleCollapse = useCallback(() => {
-    const newCollapsed = !isCollapsed;
-    storage.set(STORAGE_KEY, newCollapsed);
-    setIsCollapsed(newCollapsed);
-  }, [ isCollapsed, storage ]);
-
-  useEffect(() => {
-    const type = isCollapsed ? StoreAction.CollapseUIChrome : StoreAction.ExpandUIChrome;
-    storeDispatch({ type });
-  }, [ isCollapsed, storeDispatch ]);
+    updateSettings({ navbarCollapsed: !settings.navbarCollapsed });
+  }, [ settings.navbarCollapsed, updateSettings ]);
 
   if (!showNavigation) return null;
 
@@ -83,8 +103,8 @@ const NavigationSideBar: React.FC = () => {
       appear={true}
       classNames={{
         appear: css.collapsedAppear,
-        appearActive: isCollapsed ? css.collapsedEnterActive : css.collapsedExitActive,
-        appearDone: isCollapsed ? css.collapsedEnterDone : css.collapsedExitDone,
+        appearActive: settings.navbarCollapsed ? css.collapsedEnterActive : css.collapsedExitActive,
+        appearDone: settings.navbarCollapsed ? css.collapsedEnterDone : css.collapsedExitDone,
         enter: css.collapsedEnter,
         enterActive: css.collapsedEnterActive,
         enterDone: css.collapsedEnterDone,
@@ -92,7 +112,7 @@ const NavigationSideBar: React.FC = () => {
         exitActive: css.collapsedExitActive,
         exitDone: css.collapsedExitDone,
       }}
-      in={isCollapsed}
+      in={settings.navbarCollapsed}
       nodeRef={nodeRef}
       timeout={200}>
       <nav className={css.base} ref={nodeRef}>
@@ -117,12 +137,13 @@ const NavigationSideBar: React.FC = () => {
               <Button
                 className={css.launchButton}
                 onClick={() => setShowNotebookModal(true)}>Launch JupyterLab</Button>
-              {isCollapsed?
+              {settings.navbarCollapsed ? (
                 <Button className={css.launchIcon} onClick={() => setShowNotebookModal(true)}>
                   <Icon
                     name={'add-small'}
                     size="tiny" />
-                </Button> : null}
+                </Button>
+              ) : null}
             </div>
             <NotebookModal
               visible={showNotebookModal}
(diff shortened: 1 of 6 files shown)


webui/react/src/hooks/useSettings.ts
```diff
@@ -0,0 +1,160 @@
...
+        ? [ queryValue ] : queryValue;
+
+      acc[prop.key] = normalizedValue;
+    } catch (e) {
+      if (defaultSettings && defaultSettings[prop.key] !== undefined) {
+        acc[prop.key] = defaultSettings[prop.key];
+      }
+    }
+    return acc;
+  }, {} as GenericSettings) as unknown as T;
```
hkang1, line 98: Open to ideas on how to avoid this `X as unknown as Y` type casting. This goal in general is to cast it to something that's either more specific or something that's more general depending on the use case. The one here specifically is going the direction of more specific. Where we process the query string into a `GenericSettings` but we know for a fact, given the config, it is actually a specific type `T`.

webui/react/src/pages/TaskList.tsx
```diff
@@ -562,17 +404,16 @@ const TaskList: React.FC = () => {
...
-    storage.set(STORAGE_SORTER_KEY, updatedSorter);
-    setSorter(updatedSorter);
-
-    const updatedPagination = {
-      limit: tablePagination.pageSize,
-      offset: (tablePagination.current - 1) * tablePagination.pageSize,
+    const newSettings = {
+      sortDesc: order === 'descend',
+      /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+      sortKey: columnKey as any,
```
hkang1, line 410: don't have a good way to avoid this... any help is appreciated!

webui/react/src/pages/TaskList.settings.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface Settings {
+  search: string;
+  sortDesc: boolean;
+  sortKey: 'id' | 'name' | 'resourcePool' | 'startTime' | 'state' | 'type' | 'username';
+  state: CommandState[] | undefined;
+  tableLimit: number;
+  tableOffset: number;
+  type: CommandType[] | undefined;
+  user: string[] | undefined;
+}
```
hkang1, line 14: I started off using `io-ts` + `io.TypeOf` to create the Typescript definitions, but there are too many walls I've hit. The biggest problem is that as definitions became more complex, the generated Typescript became `any`. Anytime, `io.Mixed` is used `any` gets generated, which renders the Typescript useless.

webui/react/src/hooks/useSettings.ts
```diff
@@ -0,0 +1,167 @@
...
+      }
+      acc[config.key] = value;
+      return acc;
+    }, {} as GenericSettings) as unknown as T;
+  });
+
+  const filterCount = useMemo(() => {
+    return config.settings.reduce((acc, config) => {
+      const settingsValue = (settings as unknown as GenericSettings)[config.key];
+      return acc + (config.isFilter && settingsValue !== config.defaultValue ? 1 : 0);
```
hamidzr, line 132: we might need to check the values here as this comparison would only compare by reference for non premitives and fail for say `[1] === [1]`

hamidzr (commented):
submitting my scattered comments. I'll do another round soon

webui/react/src/hooks/useSettings.ts
```diff
@@ -0,0 +1,167 @@
...
+  Float = 'Float',
+  Integer = 'Integer',
+  String = 'String',
+}
+
+type GenericSettingsType = Primitive | Primitive[] | undefined;
+
+/*
+ * defaultValue   - Must be provided even when `undefined`.
+ * isFilter       - Needed to support filterCount and resetFilters.
```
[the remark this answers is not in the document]
hkang1: yeah good call, the filters should probably just be handled by the respective pages.

webui/react/src/hooks/useSettings.ts
hkang1: great catch! this should be able to handle more than just primitives.

webui/react/src/pages/TaskList.settings.ts
```diff
@@ -0,0 +1,81 @@
...
+  tableOffset: number;
+  type?: CommandType[];
+  user?: string[];
+}
+
+const config: SettingsConfig = {
+  settings: [
+    {
+      defaultValue: '',
+      isFilter: true,
```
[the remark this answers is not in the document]
hkang1: yeah maybe not the place to have it. It is convenient to have it as a property to have hook automatically handle it, but maybe it doesn't quite fit into the settings mold.

webui/react/src/pages/TaskList.tsx
```diff
@@ -58,178 +57,34 @@ interface SourceInfo {
...
-
 const TaskList: React.FC = () => {
   const { users } = useStore();
-  const storage = useStorage(STORAGE_PATH);
-  const initLimit = storage.getWithDefault(STORAGE_LIMIT_KEY, MINIMUM_PAGE_SIZE);
-  const initFilters = storage.getWithDefault(STORAGE_FILTERS_KEY, { ...defaultFilters });
-  const [ isUrlParsed, setIsUrlParsed ] = useState(false);
+  const { filterCount, resetFilters, settings, updateSettings } = useSettings<Settings>(
+    settingsConfig,
+    paths.taskList(),
```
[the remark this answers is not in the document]
hkang1: The hook maintains its own state. Looking at code again, if the path changes, the only thing that gets updated is the `updateSettings` callback. The existing local storage and the user settings state shouldn't get affected.

webui/react/src/pages/TaskList.tsx
```diff
@@ -562,17 +404,16 @@ const TaskList: React.FC = () => {
...
+      /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+      sortKey: columnKey as any,
+      tableLimit: tablePagination.pageSize,
+      tableOffset: (tablePagination.current - 1) * tablePagination.pageSize,
     };
-    storage.set(STORAGE_LIMIT_KEY, tablePagination.pageSize);
-    setPagination(updatedPagination);
-  }, [ columns, setSorter, storage ]);
+    const shouldPush = settings.tableOffset !== newSettings.tableOffset;
+    updateSettings(newSettings, shouldPush);
```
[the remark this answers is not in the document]
hkang1: Interesting... I guess it kind of depends on what setting changes are good to add to the history queue versus replacing it. `tableOffset` seemed like the only one that was worth pushing but I haven't really put too much thought into this part. What do you think?

webui/react/src/pages/TaskList.tsx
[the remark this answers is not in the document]
hkang1: > if this path or the way we generate it changes in any way it'll blow away user settings.

I think I get what you're putting down. Like `/tasks` getting changed to `/task-list`, then URLs that use `/tasks` won't work.

webui/react/src/pages/TaskList.tsx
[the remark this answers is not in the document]
hamidzr: >  kind of depends on what setting changes are good to add to the history queue

aah, I didn't think of it like that. it makes sense.

webui/react/src/pages/TaskList.tsx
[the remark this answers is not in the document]
hamidzr: > > if this path or the way we generate it changes in any way it'll blow away user settings.
> 
> I think I get what you're putting down. Like `/tasks` getting changed to `/task-list`, then URLs that use `/tasks` won't work.

yeah this one

webui/react/src/pages/TaskList.settings.ts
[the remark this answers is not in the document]
hamidzr: I was gonna chat with you about this. maybe we still do later. if it's very convenient to have it maybe we do 🤔

webui/react/src/pages/TaskList.tsx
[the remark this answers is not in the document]
hamidzr: maybe if our `paths` returned the full route instead of just the resolved url we could have used the route id which would be less likely to change but a constant works great as well IMO

webui/react/src/pages/TaskList.settings.ts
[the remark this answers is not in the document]
hkang1: one thought I had around this is to enlarge the scope and maybe think about this as `useUserState`, where it will handles all the states that involve direct user interactions. It will then automatically handle behaviors around filtering, query string, internal state, local storage, and potentially more future hooks.

webui/react/src/pages/TaskList.tsx
```diff
@@ -58,178 +57,34 @@ interface SourceInfo {
...
-};
-
-const URL_ALL = 'all';
-
-const STORAGE_PATH = 'task-list';
-const STORAGE_FILTERS_KEY = 'filters';
-const STORAGE_LIMIT_KEY = 'limit';
-const STORAGE_SORTER_KEY = 'sorter';
+const filterKeys: Array<keyof Settings> = [ 'search', 'state', 'type', 'user' ];
+const settingsBasePath = paths.taskList();
```
hamidzr, line 61: ditto https://github.com/determined-ai/determined/pull/2786#discussion_r681208872

webui/react/src/pages/TaskList.tsx
```diff
@@ -249,6 +104,30 @@ const TaskList: React.FC = () => {
...
+  const filterCount = useMemo(() => {
+    return filterKeys.reduce((acc, key) => {
+      const config = filterConfigMap[key];
+      const settingsValue = settings[key];
+      const settingsIsSet = settingsValue && !isEqual(settingsValue, config.defaultValue);
+      return acc + (settingsIsSet ? 1 : 0);
+    }, 0);
+  }, [ filterConfigMap, settings ]);
+
+  const resetFilters = useCallback(() => {
```
hamidzr, line 123: suggestion: I recall you had a resetFilters in the useSettings hook. if that was a resetSettings that optionally took a set of setting keys? would that  make it reusable here?

webui/react/src/pages/TaskList.tsx
```diff
@@ -249,6 +104,30 @@ const TaskList: React.FC = () => {
...
   }, [ selectedTasks ]);
 
+  const filterConfigMap = useMemo(() => {
+    return filterKeys.reduce((acc, key) => {
+      const config = settingsConfig.settings.find(config => config.key === key);
+      return { ...acc, [key]: config };
+    }, {} as Record<keyof Settings, SettingsConfigProp>);
+  }, []);
+
+  const filterCount = useMemo(() => {
```
hamidzr, line 114: suggestion: similar to the comment below. If the useSettings hook could give you a list of "active" settings IMO that could make this portion reusable without specializing it to filters.

webui/react/src/pages/TaskList.tsx
hamidzr: did you mean to use the `storagePath` from `..settings.ts` file?

hamidzr (commented):
this looks good to me, I added a couple of suggestions and I want to do a bit more final testing. Are the ones you have in the test plans ones that you've gone through or are those todos?

could you please add a sample usage where we want to persist a setting but not utilize sharing (and thus no query parameter updates)

webui/react/src/pages/TaskList.tsx
hkang1: this is a great idea!

webui/react/src/pages/TaskList.tsx
hkang1: totally makes sense!

webui/react/src/pages/TaskList.tsx
hkang1: the `storagePath` is meant for the local storage. Was thinking at one point of doing the same for `settingsBasePath` but on pages with variables in the route paths (like experiment detail `/experiment/<id>`) it would require a bit of processing.

hkang1:
> This looks good to me, I added a couple of suggestions and I want to do a bit more final testing. Are the ones you have in the test plans ones that you've gone through or are those todos?
> 
Yeah I went through those but also listed them for release testing.

> could you please add a sample usage where we want to persist a setting but not utilize sharing (and thus no query parameter updates)

I was able to test this case by enabling the config `skipUrlEncoding` for a couple of the `TaskList.settings.ts` config, but I don't have a real use case scenario for this yet. Any recommendations? Navbar is not a good candidate as `useSettings` is mostly designed on a per page basis and would not work well with different levels of use. The core problem is that the different query params from different uses of `useSettings` at different levels would override each other.

hkang1:
will add some unit tests soon

hamidzr:
> The core problem is that the different query params from different uses of useSettings at different levels would override each other.

I can see this especially if the different levels of settings would have settings that are shared but could it be still problematic if one does not have any sharable settings? In any case, I don't think we need to solve this right now.

hamidzr:
> Yeah I went through those but also listed them for release testing.

👍🏼

webui/react/src/hooks/useSettings.ts
```diff
@@ -0,0 +1,244 @@
...
+  settings: SettingsConfigProp[];
+  storagePath: string;
+}
+
+type GenericSettings = Record<string, GenericSettingsType>;
+
+interface SettingsHook<T> {
+  resetSettings: (keys?: string[]) => void;
+  settings: T;
+  settingsCount: (keys?: string[]) => number;
```
hamidzr, line 49: I could see a usage for this if this was to return the active key names (and not doing the counting), or if we exposed a separate hook that did this: if we wanted to compare two pages together eg for sharing webui state in sharable UI project. This can always be added later when we want to use it.

hamidzr (approved):
great work @hkang1

hkang1:
> > The core problem is that the different query params from different uses of useSettings at different levels would override each other.
> 
> I can see this especially if the different levels of settings would have settings that are shared but could it be still problematic if one does not have any sharable settings? In any case, I don't think we need to solve this right now.

This should be address with the this [new commit](https://github.com/determined-ai/determined/pull/2786/commits/6611d49b1ce1a8ef68fd6bd5cb348ad3c42283f9).

webui/react/src/hooks/useSettings.ts
hkang1: great suggestion! generalized to `activeSettings` to provide more flexibility in usage.

hkang1:
> will add some unit tests soon, but just need to confirm this is the direction we want to go.

too many unit tests to add for full coverage, so covered some core basic ones and one dealing with config specifically so it'll be easier to add more in the future... 😓

hamidzr (approved):
works great. and thanks for adding the unit tests

webui/react/src/components/NavigationSideBar.tsx
```diff
@@ -20,10 +20,47 @@ interface ItemProps extends LinkProps {
...
+  ],
+  storagePath: 'navigation',
+};
+
+const menuConfig = {
+  bottom: [
+    { icon: 'logs', label: 'Master Logs', path: paths.masterLogs() },
+    { external: true, icon: 'docs', label: 'Docs', path: paths.docs(), popout: true },
+    {
+      external: true,
```
hkang1, line 48: @hamidzr oops

PR #2787, opened 2021-07-24
feat: support configuring idle timeout for generic commands [DET-4589]

shiyuann:
## Description

Support configuring idle timeout for generic commands. This supports TensorBoard and will support Notebook in a followup PR. 

## Test Plan

- Manually test it with TensorBoards

## Checklist


docs/release-notes/2773-work-dir.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**New Features**
+
+-  Support configuring working directory in the ``environment.work_dir`` configuration for all task
```
stoksc, line 5: @shiyuann is this PR blocked on another (since title only describes the other feature) or am I to review both? either is fine just curious

shiyuann: It's not blocked by anyone. It's to configure TensorBoard idle timeout, which we already support. But this will also help if we support Notebook idle timeout.

shiyuann: But yes, it will be rebased if #2773 is merged. You can review the last two commits.

master/internal/task/idle_timeout.go
```diff
@@ -0,0 +1,57 @@
...
+		services := ctx.Ask(p.proxyRef, proxy.GetSummary{}).Get().(map[string]proxy.Service)
+
+		service, ok := services[p.TaskID]
+		if !ok {
+			return nil
+		}
+
+		if time.Now().After(service.LastRequested.Add(time.Duration(p.Timeout))) {
+			ctx.Log().Infof("killing %s due to inactivity", p.Description)
+			ctx.Ask(ctx.Self(), p.KillMessage)
```
stoksc, line 47: wait, wouldn't asking yourself anything deadlock immediately? but you dont `.Get()` the resp it why it works. why not just a tell?

master/internal/task/idle_timeout.go
```diff
@@ -0,0 +1,57 @@
...
+
+	TaskID      string
+	Description string
+	Timeout     model.Duration
+	KillMessage actor.Message
+}
+
+// PreStart should be called on task actor PreStart.
+func (p *ProxyIdleTimeoutWatcher) PreStart(ctx *actor.Context) {
+	p.proxyRef = ctx.Self().System().Get(actor.Addr("proxy"))
```
stoksc, line 29: can you put `actor.Addr("proxy")` into a pkg somewhere as a `var proxyAddr = ...` at the top. I think we should do this for all "well known" addresses (rm, trial logger, etc)

master/internal/task/idle_timeout.go
```diff
@@ -0,0 +1,57 @@
...
+	case ProxyIdleTimeoutWatcherTick:
+		services := ctx.Ask(p.proxyRef, proxy.GetSummary{}).Get().(map[string]proxy.Service)
+
+		service, ok := services[p.TaskID]
+		if !ok {
+			return nil
+		}
+
+		if time.Now().After(service.LastRequested.Add(time.Duration(p.Timeout))) {
+			ctx.Log().Infof("killing %s due to inactivity", p.Description)
```
stoksc, line 46: alternative to passing description+kill message, what i've been doing is passing a `func(*actor.Context)` that encapsulates all the state to take a generic action in a subcomponent.

master/pkg/model/environment_config.go
```diff
@@ -24,16 +24,19 @@ const (
 
 // Environment configures the environment of a Determined command or experiment.
 type Environment struct {
-	Image                RuntimeItem  `json:"image"`
-	EnvironmentVariables RuntimeItems `json:"environment_variables,omitempty"`
-
-	Ports          map[string]int    `json:"ports"`
+	Image          RuntimeItem       `json:"image"`
```
stoksc, line 27: appreciate the restructuring, maybe drop comments to describe the mini-sections within the struct?

master/pkg/tasks/task_command.go
```diff
@@ -23,6 +23,8 @@ type GenericCommandSpec struct {
 
 	Port     *int
 	ProxyTCP bool
+
+	WatchTensorBoardTimeout bool
```
stoksc, line 27: is this var accurately named?

stoksc (commented):
mostly lgtm, nice refactor, definitely moving in the right direction

master/internal/task/idle_timeout.go
stoksc: or do it later/make a ticket for this pattern if you agree with it

master/internal/task/idle_timeout.go
shiyuann: good catch. Change it to tell

master/internal/task/idle_timeout.go
shiyuann: agreed. Will do that when we refactor interactive jobs.

master/pkg/tasks/task_command.go
shiyuann: Changed it to `WatchProxyIdleTimeout`

master/pkg/model/environment_config.go
shiyuann: This is done in a different PR. Also, I think it's not very necessary to add comments for mini-sections since they are already self-explained.

master/internal/api_command.go
```diff
@@ -132,6 +132,14 @@ func (a *apiServer) getCommandLaunchParams(ctx context.Context, req *protoComman
 	var userFiles archive.Archive
 	if len(req.Files) > 0 {
 		userFiles = filesToArchive(req.Files)
+
+		workdirSetInReq := config.Environment.WorkDir != nil &&
```
stoksc, line 136: sorry, I'm a bit confused, the case where `config.Environment.WorkDir != nil && *workDirInDefaults == *config.Environment.WorkDir` is ok?

stoksc: i guess my question is, why is this different than the check for experiments? sorry i'm out of the loop product side on this one.

stoksc (approved):
approved (obviously after conflicts/tests pass etc)

master/internal/api_command.go
stoksc: i can't read

master/internal/command/command.go
```diff
@@ -158,7 +196,12 @@ func (c *command) Receive(ctx *actor.Context) error {
 			Notebook: c.toNotebook(ctx),
 			Config:   protoutils.ToStruct(c.Config),
 		})
-
+	case *apiv1.IdleNotebookRequest:
+		if !msg.Idle {
+			now := time.Now()
+			c.lastActivity = &now
```
stoksc, line 202: fwiw, for this annoyance there is `ptrs.TimePtr()` that rb added. also, should this be called like "SetIdleness" ha.. Idk can't think of a great name for it.

stoksc (approved):

master/internal/command/command.go
```diff
@@ -158,7 +196,12 @@ func (c *command) Receive(ctx *actor.Context) error {
 			Notebook: c.toNotebook(ctx),
 			Config:   protoutils.ToStruct(c.Config),
 		})
-
+	case *apiv1.IdleNotebookRequest:
```
stoksc, line 199: my only real critique is that I may call this "IdleTaskRequest" or "IdleAllocationRequest" instead.

shiyuann: This is only under `/api/v1/notebooks/` now and internal. I think it's fine to do so for now.

master/internal/command/command.go
shiyuann: Let's make the names consistent when we refactor this. I think it's fine for now.

master/internal/command/command.go
```diff
@@ -57,6 +58,45 @@ func createGenericCommandActor(
...
+		case spec.WatchProxyIdleTimeout && spec.WatchRunnerIdleTimeout:
+			getLastActivity = func(ctx *actor.Context) *time.Time {
+				proxyRef := ctx.Self().System().Get(actor.Addr("proxy"))
+				services := ctx.Ask(proxyRef, proxy.GetSummary{}).Get().(map[string]proxy.Service)
+				service, ok := services[string(taskID)]
+				if !ok {
+					return nil
+				}
+
+				if cmd.lastActivity != nil && cmd.lastActivity.After(service.LastRequested) {
```
stoksc, line 73: I should've caught this in review @shiyuann but here we have unprotected concurrent access of cmd.lastActivity since it's captured by closure. We're very unlikely to hit it, but i'll fix it in the refactor I'm doing right now for task unif.

stoksc: oh wait, it's not a different actor, nvm. i forgot it just has things that look like actor methods.

PR #2796, opened 2021-07-28
feat: adding profiling metrics to continuous benchmarking

azhou-determined:
## Description
- add profiling capability to continuous metric server

## Checklist


e2e_tests/tests/conftest.py
```diff
@@ -154,3 +155,16 @@ def test_start_timer(request: SubRequest) -> Iterator[None]:
...
+def profile_test(record_property: Callable[[str, object], None]) -> Callable[[int], None]:
+    """
+    Returns a method that allows profiling of test run for certain system metrics
+    and records to JUnit report.
+
+    Currently retrieves metrics by trial (assumes one trial per experiment) using
+    profiler API.
+    """
+
+    return ProfileTest(record_property=record_property).record
```
stoksc, line 170: I think it makes more sense to use just a higher order function that takes record_property and returns ProfileTest.record instead of returning a method on a class.

e2e_tests/tests/conftest.py
```diff
@@ -154,3 +155,16 @@ def test_start_timer(request: SubRequest) -> Iterator[None]:
         # This ends up concatenated to the line pytest prints containing the test file and name.
         print("starting at", time.strftime("%Y-%m-%d %H:%M:%S"))
     yield
+
+
+@pytest.fixture
+def profile_test(record_property: Callable[[str, object], None]) -> Callable[[int], None]:
```
stoksc, line 161: profile_test -> collect_trial_profiles

stoksc (commented):
looks pretty good, just minor style nits

stoksc:
anywhere I can see this live, in-action? (can you link the db/visuals in devops ag or something?)

azhou-determined:
> anywhere I can see this live, in-action? (can you link the db/visuals in devops ag or something?)

good call. going to change the branch off of master to run a few runs on this branch and catch nightlies tonight.

azhou-determined:
> > anywhere I can see this live, in-action? (can you link the db/visuals in devops ag or something?)
> 
> good call. going to change the branch off of master to run a few runs on this branch and catch nightlies tonight.

@stoksc http://54.203.200.130/d/2v40ruMnk/system-metrics?orgId=1&from=now-24h&to=now 
some system metrics dashboards, look back 24h to see the runs i've done in the past few days. login for grafana is username: admin, password: determined

e2e_tests/tests/conftest.py
azhou-determined: hmm... so i want the metrics to be configurable from the conftest config, like: 
`ProfileTest(record_property=record_property, metrics={SysMetricName.DISK_IOPS_METRIC: ["avg", "max"]}).record`
but also keep the fixture that's returned really simple in signature, i.e tests should only need to pass in trial_id in order to use it. is there a cleaner way to do this with higher order functions?

.circleci/scripts/upload_test_results.py
```diff
@@ -24,13 +25,19 @@ def upload_results(status, directory, job_id, access_key, url) -> None:
...
+                   args.access_key, args.url)
 
 
 if __name__ == "__main__":
-    main()
\ No newline at end of file
+    try:
+        main()
+    except ConnectionError as err:
+        print(f"Error connecting to CI server {err}")
```
stoksc, line 43:
```suggestion
        print(f"Error connecting to CI server {err}")

```

e2e_tests/tests/conftest.py
stoksc: I have I don't seen anywhere it is being used like that but if you wanted it configurable you could just have a signature like
```
def recorder_maker(record_property, metrics):
    def record(self, trial_id):
        ...
     return record
```
i mean, i'm not sure there is much of a difference in capturing the state as an instance var in a class vs a closure like there, other than the original style nit that it feels weird to me to return a class instance method without the class.

e2e_tests/tests/requirements.txt
```diff
@@ -9,3 +9,5 @@ tensorflow==2.4.1
 pandas
 pyyaml
 docker
+boto3
```
stoksc, line 12: where does this come from?

stoksc (commented):
this code lgtm but the grafana instance doesn't seem to show any data. maybe (looks like) the X axis on all the charts is time? I think it should just be like, run_id or something. Some way that if there aren't runs for a while I still see the latest.

e2e_tests/tests/requirements.txt
azhou-determined: removed. my IDE was complaining at some point but seems to have stopped after invalidating caches and restarting.

e2e_tests/tests/conftest.py
azhou-determined: makes sense, i like this better too. changed.

stoksc (approved):
lgtm!

PR #2797, opened 2021-07-28
fix: preserve zoom levels between uPlot remounts [DET-5636, DET-5751]

hkang1:
## Description

PROBLEM:
Zoom on the chart does not get preserved on the Learning Curve chart. The Learning Curve chart differs slightly in that new series (new trials) can get added to the chart, which cause the chart options to change, which requires the chart to be remounted (teared down and rebuilt). The zoom gets lost during this remount process.

SOLUTION:
Preserve the zoom (scales) if a zoom is detected, and upon remounting, apply the scales appropriately.

ADDITIONS:
- refactor to save chart as a reference instead of a state.
- refactor to remove the need for a forward reference (generally an anti-pattern)
- fix issue of learning curve not resizing properly.

## Test Plan

- [ ] test the HP-viz learning curve during a live experiment, by zooming in and make sure the zoom doesn't get reset upon a new trial.
- [ ] test the Cluster Historical charts. Zoom in and wait for the data to change and check to make sure the zoom does not change upon new data.
- [ ] test the Profiler charts (same test as above).
- [ ] check to make sure learning curve resizes when you make the browser wider.

## Checklist

### DIFF
--- a/docs/release-notes/2797-preserve-uplot-zoom-between-remounts.txt
+++ b/docs/release-notes/2797-preserve-uplot-zoom-between-remounts.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  WebUI: Prevent zoom from reseting when chart data series gets added.
+-  WebUI: Fix issue of learning curve not resizing properly.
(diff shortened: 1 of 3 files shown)


EmilyBonar (approved):
LGTM

PR #2802, opened 2021-07-29
fix: size and resize the config monico editor [DET-5736]

EmilyBonar:
## Description
I don't love my solution for the mobile issue (I just checked whether the width is under the breakpoint and hardcoded the monaco editor height), but it works.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentConfiguration.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentConfiguration.tsx
@@ -1,8 +1,9 @@
 import yaml from 'js-yaml';
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import MonacoEditor from 'react-monaco-editor';
 
 import Section from 'components/Section';
+import useResize from 'hooks/useResize';
 import { ExperimentBase } from 'types';
 
 interface Props {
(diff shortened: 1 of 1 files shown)


hkang1 (changes requested):
for the auto-resize, it looks like you can pass an argument of `automaticLayout: true` into options, as opposed to having to re-run `layout()` after resize is detected.
https://stackoverflow.com/a/47163918/5402432

Regarding the mobile size, it turns out it's the wrapping `Page` component that is not properly sized. I only know this because I was addressing a different vertical height issue on a different ticket and it's not just the monaco editor's problem. Just leaving the height `100%` as-is is fine. The other ticket should fix the mobile issue when `Page` is corrected.

webui/react/src/pages/ExperimentDetails/ExperimentConfiguration.tsx
```diff
@@ -1,15 +1,23 @@
...
 interface Props {
   experiment: ExperimentBase;
 }
 
 const ExperimentConfiguration: React.FC<Props> = ({ experiment }: Props) => {
+  const editor = useRef<MonacoEditor>(null);
+  const resize = useResize();
+
+  useEffect(() => {
+    editor.current?.editor?.layout();
```
hkang1, line 18: sweet syntax!

hkang1 (approved):
LGTM, discussed the reasons of doing it this way was more efficient than the default MonacoEditor's `automaticLayout: true` property. Resizing it after resize has completed vs polling every so often to resize.

PR #2804, opened 2021-07-30
feat: switch to bash in jupyterlab shells. [DET-5791]

ioga:
## Description

JupyterLab defaults to `/bin/sh` in terminals which is annoying for users. Let's grab the shell from passwd and set `SHELL` to make JupyterLab use the user's preferred shell (in practice, `/bin/bash`).

## Test Plan
Launch a terminal inside JupyterLab.

## Checklist


rb-determined-ai (approved):
Great idea.  I had no idea this is why jupyter was acting like that.

PR #2805, opened 2021-08-04
fix: k8s scheduler deployment fails

eecsliu:
## Description
Previously, the score plugin was added to the end of the list. However, the list needs to be ordered in the same way the functions are declared in order to work. In the current configuration, the coscheduler scheduler deployment fails and determined is not able to schedule jobs when it uses the coscheduler. This fixes the issue

## Checklist

### DIFF
--- a/helm/charts/determined/templates/scheduler-config.yaml
+++ b/helm/charts/determined/templates/scheduler-config.yaml
@@ -25,13 +25,13 @@ data:
           enabled:
             - name: Coscheduling
         {{- end }}
+        score:
+          enabled:
+            - name: Coscheduling
         permit:
           enabled:
             - name: Coscheduling
         unreserve:
-          enabled:
-            - name: Coscheduling
-        score:
           enabled:
             - name: Coscheduling
     # optional plugin configs


mackrorysd (approved):

mackrorysd (commented):
I know I already clicked approve, but it occurs to me that it's far from ideal for the version of our Helm chart to matter so much depending on which version of other stuff you're deploying, when the relevant change is the reordering of functions. Is it not better to just reorder the functions in a new build?

eecsliu:
> Is it not better to just reorder the functions in a new build?

Good point, let me investigate that a little bit and make sure it works.

eecsliu:
Verified that rebuilding the plugin with the reordered function declarations works with this configuration. This also brings up the question - should we be ordering the plugin points in the order of which they're listed in the docs/the order in which they occur? Or is it ok for the order to be arbitrary (but fixed)? 

``score, permit, unreserve`` is the proper order in which they occur. Are we ok with fixing it to ``permit, unreserve, score``?

PR #2815, opened 2021-08-09
tests: add waitFor helper with error retention

hamidzr:
## Description

the way the exiting `waitFor` helper provided by taiko is set up discourages use of `expect` which relies on throwing exceptions. The expect libraires, and similar other testing libraries, provide useful messages around why the test failed. The added utility makes using these exception-based libraries easier.

## Checklist


webui/tests/tests/StepImplementation.ts
```diff
@@ -64,6 +65,42 @@ const goto = async (url: string) => {
...
+// waitFor that takes a throwable function instead of a predicate compared to t.waitFor
+const waitFor = async (throwableFn: () => Eventually<void | Error>, timeout = TIMEOUT) => {
+  let timeOver = false;
+  let timeoutPromise = sleep(timeout).then(() => {
+    timeOver = true;
+  });
+  let lastStatus = undefined;
+  const runSafely = async (fn) => {
+    try {
+      const pr = await fn();
```
hamidzr, line 85: rename. to..?

webui/tests/tests/StepImplementation.ts
```diff
@@ -64,6 +65,42 @@ const goto = async (url: string) => {
...
+      return pr;
+    } catch (e) {
+      return e;
+    }
+  };
+  while (true) {
```
hamidzr, line 91: there might be a more elegant way of doing this. 

goals:
- respect the overall timeout
- return the last failure reason instead of just "timed out"
- exit as soon as the condition passes
- avoid using a defined interval to check.

caveat: the current approach does not throttle the check

webui/tests/tests/StepImplementation.ts
```diff
@@ -26,6 +26,7 @@ const BASE_URL = `${HOST}${BASE_PATH}`;
 const viewports = {
   desktop: { width: 1366, height: 768 },
 };
+const TIMEOUT = 5000;
```
hamidzr, line 29: this differs from the default 10s timeout for taiko.

webui/tests/tests/StepImplementation.ts
hkang1: **nit suggestion:** not 100% sure but I think this should be equal:
```
while (lastStatus === undefined && !timeOver) {
   ...
}
return lastStatus;
```
instead of
```
while (true) {
   ...
   if (timeOver) {
    return lastStatus;
  }
}
```
The only benefit is to just avoid a `while (true) { ... }` even though it is structured to guarantee an exit.

`lastStatus === undefined` to ensure the loop runs at least once even if `timeOver` happens to be `true` before it even gets to the loop. (not sure if that is possible but just to be 100% sure).

webui/tests/tests/StepImplementation.ts
```diff
@@ -64,6 +65,42 @@ const goto = async (url: string) => {
...
+      return e;
+    }
+  };
+  while (true) {
+    const rv = await runSafely(() => Promise.race([timeoutPromise, runSafely(throwableFn)]));
+    if (!(rv instanceof Error)) {
+      return rv;
+    } else {
+      lastStatus = rv;
+    }
```
hkang1, line 97: **nit:** a slight swap to make it easier to read
```
if (rv instanceof Error) {
  lastStatus = rv;
} else {
  return rv;
}
```

webui/tests/tests/StepImplementation.ts
```diff
@@ -64,6 +65,42 @@ const goto = async (url: string) => {
   await t.goto(url, { waitForEvents: ['loadEventFired'] });
 };
 
+type Eventually<T> = T | Promise<T>;
```
hkang1, line 68: sweet! could probably use this in `src/types.ts` wdy think?

hamidzr: we already got it there :)

hamidzr: in utils/types

webui/tests/tests/StepImplementation.ts
hamidzr: I updated this to fix and ensure we throw at the end if we are wrapping up with an error instead of an expected value and that resulted in not needing the timeout anymore. 

assumption: in case timeout is reached before the input function is fully resolved we want to exit with a timeout

are we set up to write and run unit tests for our test code..?

webui/tests/tests/StepImplementation.ts
```diff
@@ -64,6 +65,42 @@ const goto = async (url: string) => {
...
+    timeOver = true;
+  });
+  let lastStatus = undefined;
+  const runSafely = async (fn) => {
+    try {
+      const result = await fn();
+      return result;
+    } catch (e) {
+      return e;
+    }
```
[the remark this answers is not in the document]
hamidzr: hmm I don't want to encourage using it it looks like we're better off if we can avoid doing things like this.

webui/tests/tests/StepImplementation.ts
hamidzr:
```
const TIMEOUT = 5000

type Eventually<T> = T | Promise<T>;

const sleep = (ms = 1000) => {
  return new Promise((resolve) => {
    setTimeout(resolve, ms);
  });
};

// waitFor that takes a throwable function instead of a predicate compared to t.waitFor
const waitFor = async (
  throwableFn: () => Eventually<unknown | Error>,
  timeout = TIMEOUT,
): Promise<unknown> => {
  const timeoutError = new Error('timeout');
  const timeoutPromise = sleep(timeout).then(() => {
    return timeoutError;
  });
  let lastStatus = undefined;
  const runSafely = async (fn: () => Eventually<unknown | Error>): Promise<unknown> => {
    try {
      const result = await fn();
      return result;
    } catch (e) {
      return e;
    }
  };
  while (true) {
    const raceResult = await runSafely(() =>
      Promise.race([timeoutPromise, runSafely(throwableFn)]),
    );
    if (raceResult instanceof Error) {
      if (raceResult === timeoutError) {
        throw lastStatus || timeoutError;
      }
      lastStatus = raceResult;
    } else {
      return raceResult;
    }
  }
};

const testCase1 = () => {
  throw new Error('case 1');
};

const testCase2 = async () => {
  await sleep(1000)
  throw new Error('case 2');
};

const testCase3 = async () => {
  await sleep(10000)
};

let callCount = 0;
const testCase4 = async () => {
  await sleep(1000)
  callCount++
  if (callCount < 3) throw new Error(`nope ${callCount}`)
  return callCount;
};


const test = async () => {
  // rv 3
  let rv = await waitFor(testCase4, 4000);
  console.log('rv is', rv);

  // throw nope 2
  // rv = await waitFor(testCase4, 2500);
  // console.log('rv is', rv);

  // throw timeout
  // rv = await waitFor(testCase4, 200);
  // console.log('rv is', rv);
};

test();
```

hkang1: ah great catch

hamidzr:
I added some unit tests. there is an issue I need to figure out.

PR #2817, opened 2021-08-10
feat: add links to trial pages in trial comparison modal [DET-5850]

EmilyBonar:
## Description
Adds links to the trial page when you click on the trial id.

## Checklist


alexjdetermined:
Suggestion: Can we add in a hover state for the main tag hover to match the checkpoint button? The blue pattern should help indicate that we are "going somewhere".

Also, im not worried about users accidentally clicking and discovering that feature as it is easy to get back to that view and adds in little friction but good nit catch @hamidzr.

PR #2819, opened 2021-08-10
fix: issue with order of install and version check of tensorboard

varlogtim:
## Test Plan

None required.


rb-determined-ai:
Actually, could I defer to @azhou-determined on this one?  I am not sure why we have multiple separate pip installs (seems like we could get away with just one?) and I'm not 100% sure if the ordering of these commands is sensitive in any way.

azhou-determined (approved):
lgtm

PR #2820, opened 2021-08-11
feat: collect sync_optimizer and backwards pass timings [DET-5724]

stoksc:
## Description
This change just adds "accumulated timings" that get aggregated over a batch (since there may be many calls to ctx.backward() or ctx.step_optimizer()), and adds code to collect new metrics depending on this.

## Test Plan
- [x] add a test for edge cases with this code
- [x] other tests already exist

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -145,10 +145,15 @@ class NamedMeasurement(Measurement):
         timestamp: datetime,
         batch_idx: int,
         value: float,
+        accumulated: bool = False,
     ):
         super().__init__(timestamp, batch_idx, value)
         self.metric_type = metric_type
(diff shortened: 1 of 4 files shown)


harness/determined/profiler.py
```diff
@@ -145,10 +145,15 @@ def __init__(
...
         batch_idx: int,
         value: float,
+        accumulated: bool = False,
     ):
         super().__init__(timestamp, batch_idx, value)
         self.metric_type = metric_type
         self.metric_name = metric_name
+        self.accumulated = accumulated
+
+    def id(self) -> str:
```
armandmcqueen, line 155: This could be a property

armandmcqueen (approved):
LGTM

PR #2825, opened 2021-08-11
fix: properly dedup container configs

rb-determined-ai:
## Description

Do the same deduplication that we do for experiments, for commands.


hamidzr (approved):
LGTM. I checked it against the steps here https://github.com/determined-ai/determined/issues/1583#issuecomment-896459817

PR #2827, opened 2021-08-12
feat: add `det agent disable --drain` [DET-5713]

ioga:
## Description
- Allow taking agent out while allowing currently running tasks to finish.

## Test Plan
- e2e test included.

## Checklist


harness/determined/cli/agent.py
```diff
@@ -124,14 +135,40 @@ def patch(args: argparse.Namespace) -> None:
...
+
         for agent_id in agent_ids:
-            path = "agents/{}/slots".format(agent_id)
-            headers = {"Content-Type": "application/merge-patch+json"}
-            payload = {"enabled": enabled}
+            action = "enable" if enabled else "disable"
+            path = f"api/v1/agents/{agent_id}/{action}"
+
+            payload = None
+            if drain_mode:
```
stoksc, line 145: any reason we should allow enable + --drain?

e2e_tests/tests/test_system.py
```diff
@@ -595,6 +595,64 @@ def test_disable_and_enable_slots() -> None:
...
+        None,
+    )
+    exp.wait_for_experiment_state(experiment_id, "ACTIVE")
+    exp.wait_for_experiment_active_workload(experiment_id)
+    exp.wait_for_experiment_workload_progress(experiment_id)
+
+    command = ["det", "-m", conf.make_master_url(), "agent", "disable", "--drain", agent_id]
+    subprocess.check_call(command)
+
+    exp.wait_for_experiment_state(experiment_id, "COMPLETED")
```
stoksc, line 634: I think we should also check the experiment is @ the last step we expected it to complete.

stoksc (commented):
looks pretty good, though I think this has some edge cases that could give us trouble, e.g. 
* if an agent is disabled and set to drain then quickly re-enabled, do we end up over subscribing that agent/slot?
* if the slot is closing forever (actor.PostStop) i think it should still kill task containers left on the slot.

we're in a weird state here where the task (trial/notebook/etc) knows it's running on the slot when we're draining but the resource pool and scheduler have lost all context on it and may make bad decisions because of it.

harness/determined/cli/agent.py
```diff
@@ -54,18 +54,19 @@ def list_agents(args: argparse.Namespace) -> None:
...
+    agent_res = api.get(master, "agents")
 
     agents = agent_res.json()
     tasks = task_res.json()
 
     c_names = {}
     for task in tasks.values():
         for cont in task["containers"]:
-            c_names[cont["id"]] = {"name": task["name"], "id": task["id"]}
+            c_names[cont["id"]] = {"name": task["name"], "id": task.get("task_id") or task["id"]}
```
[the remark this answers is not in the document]
ioga: oops, that fix has leaked from push branch

ioga:
>if an agent is disabled and set to drain then quickly re-enabled, do we end up over subscribing that agent/slot?
no; I've reordered the test to try and ensure that doesn't happen.

>if the slot is closing forever (actor.PostStop) i think it should still kill task containers left on the slot.
makes sense, I've rearranged `patch` to account for that.

master/internal/agent/slot.go
```diff
@@ -109,10 +110,14 @@ func (s *slot) patch(ctx *actor.Context) {
 			add.ContainerID = &s.container.ID
 		}
 		ctx.Tell(s.resourcePool, add)
-	} else if !s.enabled.Enabled() && s.enabled.deviceAdded {
-		s.enabled.deviceAdded = false
+	} else if !s.enabled.Enabled() {
 		remove := sproto.RemoveDevice{DeviceID: s.deviceID(ctx)}
-		ctx.Tell(s.resourcePool, remove)
+
+		if s.enabled.deviceAdded {
```
stoksc, line 116: i guess what i mean is, even if we are draining, if the slot actor stops, we've lost all context on the slot forever. we used to kill task containers in the case (i guess sort of emergency cleanup). i think we still should.

e2e_tests/tests/test_system.py
```diff
@@ -631,22 +631,32 @@ def _fetch_slots() -> List[Dict[str, Any]]:
...
-
     # Try to launch another experiment.
-    experiment_id = exp.create_experiment(
+    experiment_id_no_start = exp.create_experiment(
         conf.fixtures_path("no_op/single-medium-train-step.yaml"),
         conf.fixtures_path("no_op"),
         None,
     )
-    exp.wait_for_experiment_state(experiment_id, "ACTIVE")
+    exp.wait_for_experiment_state(experiment_id_no_start, "ACTIVE")
```
stoksc, line 640: i think to see the bug I'm concerned about, we need to quickly re-enable the slot after setting it to drain. then i think 2 experiments will land on the same slot momentarily

master/internal/agent/slot.go
ioga: I unset draining status above on `PostStop` line 79, and so line 121 should kill the container even if there's no device. is this not sufficient?

e2e_tests/tests/test_system.py
ioga: how about now?

stoksc: what happens without the call to disable --drain the agent again?

master/internal/agent/slot.go
stoksc: ah, missed that, that should do it.

stoksc: maybe drop a comment?

e2e_tests/tests/test_system.py
ioga: the 2nd experiment will start when the first one finishes.

stoksc: can you add that into the test? making sure the 2nd doesn't start while the 1st is active? sorry for asking for so much. that's just what i was concerned about from the start. we're not quite testing it yet.

ioga:
(this PR has a fatal flaw and I'll take some time to address it)

e2e_tests/tests/test_system.py
stoksc: found a bug, x-post here
1. start a 2 slot cluster
2. start a 1 slot experiment with really high resources.weight so you dont it wont get displaced
3. disable --drain its agent
4. start a new plain-old low weight 1 slot experiment
5. it won't get scheduled because fair share thinks the cluster is at capacity (since it only knows of 1 agent with 1 slot and has already offered a slot to the first one)

master/internal/agent/agent.go
```diff
@@ -103,9 +103,19 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
 	case *proto.DisableAgentRequest:
-		ctx.Tell(a.slots, patchSlot{Enabled: false})
+		// Mark current agent as disabled with RP.
+		ctx.Tell(a.resourcePool, sproto.DisableAgent{Agent: ctx.Self(), Drain: msg.Drain})
+		// Update individual slot state.
+		ctx.Tell(a.slots, patchSlot{Enabled: false, Drain: msg.Drain})
+		// Kill both slotted and zero-slot tasks, unless draining.
+		if !msg.Drain {
+			for cid := range a.containers {
+				ctx.Tell(ctx.Self(), sproto.KillTaskContainer{ContainerID: cid})
```
stoksc, line 116: Side note, I don't think the task actors as is are aware of these kills, so they will count to max restarts (for trials). Maybe drop a comment with a ticket?

master/internal/agent/agent.go
```diff
@@ -198,6 +208,11 @@ func (a *agent) containerStateChanged(ctx *actor.Context, sc aproto.ContainerSta
 }
 
 func (a *agent) summarize(ctx *actor.Context) model.AgentSummary {
+	agentState := ctx.Ask(
```
stoksc, line 211: nit: could just store these in the agent struct when we tell the slots/rm about these changes.

master/internal/resourcemanagers/fitting_methods.go
```diff
@@ -54,11 +51,10 @@ func WorstFit(req *sproto.AllocateRequest, agent *agentState) float64 {
 	switch {
 	case agent.numUsedSlots() != 0 || req.SlotsNeeded != 0:
 		return float64(agent.numEmptySlots()) / float64(agent.numSlots())
-	case agent.maxZeroSlotContainers == 0:
+	case agent.numZeroSlots() == 0:
 		return 0.0
 	default:
-		return float64(agent.maxZeroSlotContainers-agent.numZeroSlotContainers()) /
-			float64(agent.maxZeroSlotContainers)
+		return float64(agent.numEmptyZeroSlots()) / float64(agent.numZeroSlots())
```
stoksc, line 57: may be nice to note that this logic only makes sense/works because hard constraints exclude agents that are draining.

master/internal/sproto/agent_resource_manager.go
```diff
@@ -36,6 +36,28 @@ type (
...
+	}
+
+	// DisableAgent disables or drains the agent.
+	DisableAgent struct {
+		Agent *actor.Ref
+		Drain bool
+	}
+
+	// AgentStateRequest asks the cluster for the enabled/draining agent state.
+	AgentStateRequest struct {
```
stoksc, line 52: i do really think it's fine to dupe this state in the agent, but up to you

stoksc (approved):
lgtm, this got pretty dicey, thanks for pulling through. had minor comments here and there but nothing to block.

stoksc:
oh @ioga, before we land this can we decide if devcluster will replace make tools run long term? if so g2g, if not maybe we shouldn't use it in CI. but i think we're g2g.

e2e_tests/tests/experiment/experiment.py
```diff
@@ -181,6 +181,36 @@ def experiment_has_active_workload(experiment_id: int) -> bool:
...
+
+
+def wait_for_experiment_workload_progress(
+    experiment_id: int, max_ticks: int = conf.MAX_TRIAL_BUILD_SECS
+) -> None:
+    for _ in range(conf.MAX_TRIAL_BUILD_SECS):
+        trials = experiment_trials(experiment_id)
+        if len(trials) > 0:
+            only_trial = trials[0]
+            if len(only_trial["steps"]) > 1:
```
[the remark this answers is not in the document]
ioga: >this will return as soon as the trial starts its first workload

I don't think that's true: it's `> 1` so it needs to start the second one.

e2e_tests/tests/test_system.py
```diff
@@ -595,6 +595,251 @@ def test_disable_and_enable_slots() -> None:
...
+    exp.wait_for_experiment_state(experiment_id, "COMPLETED")
+    trials = exp.experiment_trials(experiment_id)
+    assert len(trials) == 1
+    assert len(trials[0]["steps"]) == 5
+
+    # Ensure the slot is empty.
+    slots = _fetch_slots()
+    assert len(slots) == 1
+    assert slots[0]["enabled"] is False
+    assert slots[0]["draining"] is True
```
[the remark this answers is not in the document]
ioga: I suppose I see `draining` more as a configuration option of the thing rather than the current state. so we could do that but it's extra complexity, so I'd skip that if you don't mind.

added the agent state check.

e2e_tests/tests/test_system.py
```diff
@@ -595,6 +595,251 @@ def test_disable_and_enable_slots() -> None:
...
+    assert agent_id1 != agent_id2
+
+    command = ["det", "-m", conf.make_master_url(), "agent", "disable", "--drain", agent_id2]
+    subprocess.check_call(command)
+
+    for command_id in [command_id1, command_id2]:
+        _wait_for_command_state(command_id, "TERMINATED", 60)
+        assert "success" in _get_command_info(command_id)["exitStatus"]
+
+    for agent_id in [agent_id1, agent_id2]:
```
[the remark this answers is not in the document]
ioga: it was getting hairy in this method... so I've added a contextmanager instead which makes this all much nicer.

master/internal/agent/agent.go
ioga: it sounds odd to me to duplicate that state. having `disabled` state both in slots and agents sounds okay because both agents and slots can currently be disabled individually (although I am not sure if disabling slots has actual product value).
and here we'd need to maintain it just for convenience? I feel like that goes against the point of responsibility separation of actors.

master/internal/resourcemanagers/fitting_methods.go
ioga: if it's draining, numEmptyZeroSlots would be zero. so this formula makes sense to me? the hard constraint is already included in it.

e2e_tests/tests/experiment/experiment.py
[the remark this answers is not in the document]
stoksc: yep, i just didn't think very good

e2e_tests/tests/test_system.py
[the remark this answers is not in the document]
stoksc: sgtm

master/internal/agent/agent.go
stoksc: I think we'd maintain it because the agent is the ultimate owner of its state but yeah also for convenience; like, the agent is draining and it just tells the resource pool this so it won't schedule on it, but ultimately its the agent that's draining and knows its draining? in general, the ask's are just a little dicey, too, since they have potential to deadlock.

stoksc: also, that means this ask will happen N times for every call to `/agents`, which is kinda a lot of calls in some cases to go through serial execution, esp for an endpoint each webui calls every 5s? i mean, it probably would be really fast in general still, but eh.

master/internal/resourcemanagers/fitting_methods.go
stoksc: sorry, i left the comment here but i was thinking it on line 39, where `numUsedSlots() > 0` for draining agents and `numEmptySlots()` is 0 so the affinity is 1 go fit to draining agents, but draining agents are excluded by hard constraints because they dont have enough slots or zero slots to satisfy the task.

master/internal/agent/agent.go
ioga: okay fine. I'll leave a TODO to make the entire thing better.

PR #2828, opened 2021-08-13
feat: cross-compile for powerpc64

rb-determined-ai:
## Description

For fetch-cloud-forest.sh, assume that GOOS and GOHOST have been set already, and only detect the special case where `growforest` is put directly in the GOPATH/bin.

## Test Plan

First, the user will confirm if it runs at all.  Then we can talk about potentially some sort of powerpc automated testing.

## Commentary (optional)

An OSS user has been trying to get the `determined-master` and `determined-agent` to build for powerpc64 but has not been successful.  I think that actually supporting powerpc64 is basically free for us.

### DIFF
--- a/agent/.goreleaser.yml
+++ b/agent/.goreleaser.yml
@@ -10,6 +10,7 @@ builds:
       - darwin
     goarch:
       - amd64
+      - ppc64
 
 archives:
   - wrap_in_directory: true
--- a/master/.goreleaser.yml
+++ b/master/.goreleaser.yml
@@ -18,8 +18,7 @@ builds:
       - darwin
     goarch:
       - amd64
-      # Before supporting more architectures than amd64, the inclusion of the
-      # growforest binary would need to be refactored to support the new arch.
+      - ppc64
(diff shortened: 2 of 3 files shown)


dzhu (approved):
Okay, so the plan for now is to manually run this with `GOARCH` set to `ppc64` just as a proof of concept?

rb-determined-ai:
Yeah basically.  I'm not in a hurry to land it... I'd rather the user who I gave the binaries to confirm they work first.

Do you know, would landing this PR also cause us to start shipping powerpc builds every release?  We would almost need a way to mark those as pre-release or something, because while this particular user has access to powerpc tensorflow/pytorch images, we don't currently ship or test any such images, and so the default images in this build would be totally useless.

dzhu:
Yeah, I think the PPC files would automatically get included into the GitHub releases. The files in the release can be [configured](https://goreleaser.com/customization/release) to an extent, but I'm not sure offhand whether that allows you to exclude the built packages (the doc comment refers to "archives", which are only the raw tarballs containing the binary).

PR #2830, opened 2021-08-13
fix: add in 'just a snapshot' msg to not lose progress on restart

stoksc:
## Description
This fixes a bug where we only took trial snapshots on searcher related events. This sounds right from the searcher point of view.. but the snapshots are also how we restore the workload sequencer state, so this caused us to lose all kinds of state on failures.

## Test Plan
- [x] test with a live cluster by running mnist with checkpoint_period = 100, kill after N 100 batches or so and blip the master. the job should resume about where it left off, not from the beginning.

## Commentary (optional)
Oops. I removed a message but I forgot it had a snapshot attached..

## Checklist

### DIFF
--- a/examples/tutorials/mnist_pytorch/const.yaml
+++ b/examples/tutorials/mnist_pytorch/const.yaml
@@ -8,6 +8,8 @@ hyperparameters:
   n_filters2: 64
   dropout1: 0.25
   dropout2: 0.5
+min_checkpoint_period:
+  batches: 100
 searcher:
   name: single
   metric: validation_loss
--- a/examples/tutorials/mnist_pytorch/const.yaml
+++ b/examples/tutorials/mnist_pytorch/const.yaml
@@ -8,8 +8,6 @@ hyperparameters:
   n_filters2: 64
   dropout1: 0.25
   dropout2: 0.5
-min_checkpoint_period:
-  batches: 100
 searcher:
   name: single
   metric: validation_loss
--- a/master/internal/trial.go
+++ b/master/internal/trial.go
@@ -709,6 +709,11 @@ func (t *trial) processCompletedWorkload(ctx *actor.Context, msg workload.Comple
 		ctx.Tell(ctx.Self().Parent(), sendNextWorkload{runID: t.RunID})
 		return nil
 	default:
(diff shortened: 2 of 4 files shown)


master/internal/trial.go
```diff
@@ -709,6 +709,11 @@ func (t *trial) processCompletedWorkload(ctx *actor.Context, msg workload.Comple
 		ctx.Tell(ctx.Self().Parent(), sendNextWorkload{runID: t.RunID})
 		return nil
 	default:
+		if err := t.tellWithSnapshot(ctx, ctx.Self().Parent(), func(s trialSnapshot) interface{} {
```
rb-determined-ai, line 712: Is this triggering a searcher snapshot for _every_ workload?  Wasn't there a reason that we could only snapshot trials when the trial checkpointed?

stoksc: Either should work just fine (every workload was exactly what we did originally), I'll change it to only on checkpoints though, that's smarter.

stoksc: I just pushed a change to only snapshot on checkpoint; it shouldn't have a correctness difference (because of t.reset()) but may be more performant.

examples/tutorials/mnist_pytorch/const.yaml
```diff
@@ -8,6 +8,8 @@ hyperparameters:
   n_filters2: 64
   dropout1: 0.25
   dropout2: 0.5
+min_checkpoint_period:
```
rb-determined-ai, line 11: is this part extraneous?

rb-determined-ai (commented):
approved, with a comment

examples/tutorials/mnist_pytorch/const.yaml
stoksc: oops, yeah this was me testing

examples/tutorials/mnist_pytorch/const.yaml
```diff
@@ -8,6 +8,8 @@ hyperparameters:
...
+  batches: 100
```
stoksc, line 12:
```suggestion
```

rb-determined-ai (approved):

PR #2834, opened 2021-08-16
feat: add the ability to set job priorities on the fly [DET-5863]

dzhu:
## Description

Add `det ... set priority` CLI commands and API support for changing
priorities.

The scheduling algorithm itself already adapted perfectly fine to
changing priorities; all we have to do here is wire up the CLI and API
to let changes be requested from outside.

## Test Plan

- [x] run a cluster with preemption on, start several experiments and play with their priorities, check that the scheduler responds by preempting and scheduling the right things
- [x] check that all the other `det * set priority` commands set the priority

## Commentary

I feel like it's a problem how much repetition it takes to handle all the non-experiment task types...

### DIFF
--- a/docs/release-notes/2834-set-priority.txt
+++ b/docs/release-notes/2834-set-priority.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**New Features**
+
+-  Support changing task priorities using the ``det experiment/command/notebook/tensorboard/shell
+   set priority`` commands.
--- a/harness/determined/cli/command.py
+++ b/harness/determined/cli/command.py
@@ -163,6 +163,19 @@ def kill(args: Namespace) -> None:
 
 
 @authentication.required
+def set_priority(args: Namespace) -> None:
+    id = RemoteTaskGetIDsFunc[args._command](args)  # type: ignore
+    name = RemoteTaskName[args._command]
+
+    try:
+        api_full_path = "api/v1/{}/{}/set_priority".format(RemoteTaskNewAPIs[args._command], id)
+        api.post(args.master, api_full_path, {"priority": args.priority})
+        print(colored("Set priority of {} {} to {}".format(name, id, args.priority), "green"))
+    except api.errors.APIException as e:
+        print(colored("Skipping: {} ({})".format(e, type(e).__name__), "red"))
+
+
+@authentication.required
 def config(args: Namespace) -> None:
     api_full_path = "api/v1/{}/{}".format(RemoteTaskNewAPIs[args._command], args.id)
     res_json = api.get(args.master, api_full_path).json()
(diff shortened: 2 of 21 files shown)


master/internal/resourcemanagers/priority.go
```diff
@@ -162,12 +162,7 @@ func trySchedulingTaskViaPreemption(
 	log.Debugf("trying to schedule task %s by preempting other tasks", allocationRequest.Name)
 
 	for priority := model.MaxUserSchedulingPriority; priority > allocationPriority; priority-- {
-		if _, ok := priorityToScheduledTaskMap[priority]; !ok {
-			continue
-		}
-
-		preemptionCandidates := priorityToScheduledTaskMap[priority]
-		for _, preemptionCandidate := range preemptionCandidates {
+		for _, preemptionCandidate := range priorityToScheduledTaskMap[priority] {
```
dzhu, line 165: Not needed for this change, but I couldn't help
 throwing in another bit of cleanup I noticed.

docs/release-notes/2834-set-priority.txt
```diff
@@ -0,0 +1,5 @@
+:orphan:
+
+**New Features**
+
+-  Support changing experiment priorities on the fly using the `det e set priority` command.
```
shiyuann, line 5:
```suggestion
-  Support changing task priorities using the `det experiment/command/notebook/tensorboard/shell set priority` command.
```

harness/determined/cli/command.py
```diff
@@ -162,6 +162,20 @@ def kill(args: Namespace) -> None:
...
 
+@authentication.required
+def set_priority(args: Namespace) -> None:
+    id = RemoteTaskGetIDsFunc[args._command](args)  # type: ignore
+    name = RemoteTaskName[args._command]
+
+    try:
+        api_full_path = "api/v1/{}/{}/set_priority".format(RemoteTaskNewAPIs[args._command], id)
+        r = api.post(args.master, api_full_path, {"priority": args.priority})
+        print(r)
```
shiyuann, line 173:
```suggestion
```

master/internal/resourcemanagers/priority.go
shiyuann: Does this actually handle errors?

shiyuann (approved):
LGTM!

docs/release-notes/2834-set-priority.txt
dzhu: Oh right, thanks.

harness/determined/cli/command.py
dzhu: Oops!

master/internal/resourcemanagers/priority.go
dzhu: Yep, if the key isn't present in the map, then the index expression will evaluate to `nil`, so the loop will just be a no-op.

PR #2841, opened 2021-08-17
fix: support pre-cross_rank versions of horovod

rb-determined-ai:
The fallback only works with the gloo controller... which is all we
support anyway.

## Test Plan

Tested manually with the `determinedai/environments:cuda-10.0-pytorch-1.4-tf-1.15-gpu-0.8.0` image, where I noticed the problem in the first place.

### DIFF
--- a/harness/determined/horovod.py
+++ b/harness/determined/horovod.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+import os
 import pathlib
 from typing import Any, Dict, List, Optional, cast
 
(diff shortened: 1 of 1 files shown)


harness/determined/horovod.py
```diff
@@ -68,23 +69,22 @@ def __getattr__(self, attr: str) -> Any:
 
     def cross_rank(self) -> Any:
         """
-        Horovod versions <= v0.21.3 accidentally did not expose cross_rank or cross_size for keras.
-
-        See https://github.com/horovod/horovod/pull/3008.
+        When hvd.cross_rank() is not reliably present (version =< v0.22.1) we fall back to reading
```
ioga, line 72:
```suggestion
        When hvd.cross_rank() is not present (version =< v0.22.1) we fall back to reading
```

I thought it was not present *at all* before 0.22.1? This word "reliably" gives me impression as if it is some non-deterministic behavior.

ioga (approved):

harness/determined/horovod.py
rb-determined-ai: No, `horovod.torch.cross_rank()` was available, as was `horovod.tensorflow.cross_rank()`, but there was a bug (which I fixed) where `horovod.keras.cross_rank()` was accidentally unavailable.

But those calls haven't been around for all that long, which I did not realize when I started using `hvd.cross_rank()` in our code.

ioga: makes sense. disregard my suggestion then.

PR #2842, opened 2021-08-18
feat: Add det deploy gke [DET-5752]

RehanSD:
## Description
Add `det deploy gke`.

## Checklist

### DIFF
--- a/harness/determined/deploy/gke/cli.py
+++ b/harness/determined/deploy/gke/cli.py
@@ -1,0 +1,426 @@
+import argparse
+import json
+from pathlib import Path
+import subprocess
+from typing import Callable
+import determined
+
+from determined.common.declarative_argparse import Arg, ArgGroup, Cmd
+from determined.deploy.gke.constants import defaults
+from determined.common.util import safe_load_yaml_with_exceptions
+from determined.common import yaml
+
+
+def validate_accelerator_type() -> Callable:
+    json_value = subprocess.check_output(
+        ["gcloud", "compute", "accelerator-types", "list", "--format=json(name)"]
+    )
+    json_names = json.loads(json_value)
+    valid_accelerator_types = {accelerator["name"] for accelerator in json_names}
+
+    def validate(s: str) -> str:
+        if s in valid_accelerator_types:
+            return s
+        raise argparse.ArgumentTypeError(
+            "Accelerator must be one of {}".format(valid_accelerator_types)
+        )
+
+    return validate
+
+
+def validate_location(isZone: bool = True) -> Callable:
+    def validate(location: str) -> str:
+        try:
+            cmd = ["gcloud", "compute"]
+            if isZone:
+                cmd += ["zones"]
+            else:
+                cmd += ["regions"]
+            cmd += ["describe", location]
+            subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+        except:
+            raise argparse.ArgumentTypeError(
+                "The specified {} {} was not found".format("zone" if isZone else "region", location)
+            )
+        return location
+
+    return validate
+
+
+def validate_machine_type(args: argparse.Namespace, zone: str, key: str = "master") -> None:
+    machine_type = args.master_machine_type if key == "master" else args.agent_machine_type
+    try:
+        subprocess.check_call(
+            ["gcloud", "compute", "machine-types", "describe", machine_type, "--zone", zone],
+            stdout=subprocess.DEVNULL,
+        )
+    except:
+        raise argparse.ArgumentError(
+            machine_type,
+            "The specified machine type {} was not found in the specified zone {}.".format(
+                machine_type, zone
+            ),
+        )
+
+
+def validate_accelerator_zone(args: argparse.Namespace, zone: str) -> None:
+    try:
+        subprocess.check_call(
+            [
+                "gcloud",
+                "compute",
+                "accelerator-types",
+                "describe",
+                args.gpu_type,
+                "--zone",
+                zone,
+            ],
+            stdout=subprocess.DEVNULL,
+        )
+    except:
+        raise argparse.ArgumentError(
+            args.gpu_type,
+            "The specified accelerator type {} was not found in the specified zone {}.".format(
+                args.gpu_type, zone
+            ),
+        )
+
+
+def validate_args(args: argparse.Namespace) -> None:
+    if args.master_machine_type != "n1-standard-16":
+        validate_machine_type(args, args.zone)
+    if args.agent_machine_type != "n1-standard-32":
+        validate_machine_type(args, args.zone, key="agent")
+    validate_accelerator_zone(args, args.zone)
+    if args.multi_node_pool:
+        if args.cpu_node_pool_name is None:
+            raise argparse.ArgumentError(
+                args.cpu_node_pool_name,
+                "--cpu-node-pool-name must be specified if using multiple node pools "
+                + "(--multi-node-pool flag is set).",
+            )
+
+
+def create_cluster(args: argparse.Namespace) -> None:
+    region = "-".join(args.zone.split("-")[:-1])
+    cmd = [
+        "gcloud",
+        "container",
+        "clusters",
+        "create",
+        args.cluster_name,
+        "--region",
+        region,
+        "--node-locations",
+        args.zone,
+        "--num-nodes=1",
+        "--machine-type={}".format(args.master_machine_type),
+    ]
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+    create_nodepools(region, args)
+    if not args.no_make_bucket:
+        cmd = ["gsutil", "mb", "gs://{}".format(args.gcs_bucket)]
+        subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+
+
+def create_nodepools(region: str, args: argparse.Namespace) -> None:
+    cmd = [
+        "gcloud",
+        "container",
+        "node-pools",
+        "create",
+        args.gpu_node_pool_name,
+        "--cluster",
+        args.cluster_name,
+        "--accelerator",
+        "type={},count={}".format(args.gpu_type, args.gpus_per_node),
+        "--zone",
+        region,
+        "--num-nodes=0",
+        "--enable-autoscaling",
+        "--min-nodes=0",
+        "--max-nodes={}".format(args.max_gpu_nodes),
+        "--machine-type={}".format(args.agent_machine_type),
+        "--scopes=storage-full,cloud-platform",
+        "--node-labels=accelerator_type=gpu",
+    ]
+    if args.multi_node_pool:
+        cmd += ["--node-taints=gpuAvailable=True:NoSchedule"]
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+    cmd = [
+        "kubectl",
+        "apply",
+        "-f",
+        defaults.K8S_NVIDIA_DAEMON,
+    ]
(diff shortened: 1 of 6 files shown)


harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+
+def validate_accelerator_type(s: str) -> None:
+    json_value = subprocess.check_output(
+        ["gcloud", "compute", "accelerator-types", "list", "--format=json(name)"]
+    )
+    json_names = json.loads(json_value)
+    valid_accelerator_types = {accelerator["name"] for accelerator in json_names}
+
+    if s not in valid_accelerator_types:
+        raise argparse.ArgumentTypeError(
```
ioga, line 35: I think `ArgumentError` might be more appropriate here: `ArgumentTypeError` is supposed to be for data types such as date vs int etc.
accelerator-type seems to just be a "value" in that sense.

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+    try:
+        cmd = ["gcloud", "compute"]
+        if isZone:
+            cmd += ["zones"]
+        else:
+            cmd += ["regions"]
+        cmd += ["describe", location]
+        subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    except:
+        raise argparse.ArgumentTypeError(
```
ioga, line 50: also probably `ArgumentError`.

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+                "machine-types",
+                "describe",
+                machine_type,
+                "--zone",
+                zone,
+            ],
+            stdout=subprocess.DEVNULL,
+        )
+    except:
+        raise ValueError(
```
ioga, line 70: doesn't seem like you differentiate `ValueError` and `ArgumentTypeError`/`ArgumentError`, so perhaps just stick to one of them?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+                + "(--multi-node-pool flag is set).",
+            )
+    if args.gpu_coscheduler and args.preemption:
+        raise ValueError(
+            "--gpu-coscheduler and --preemptive-scheduler are mutually exclusive and cannot both be"
+            " specified",
+        )
+    if args.cpu_coscheduler and not args.gpu_coscheduler:
+        raise ValueError(
+            "To enable coscheduling on the CPU Node Pool, coscheduling must be enabled on the GPU"
```
ioga, line 120: then perhaps just one flag for `--coscheduling` will be sufficient? or are there plans to split these?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+        )
+    if args.cpu_coscheduler and not args.multi_node_pool:
+        raise ValueError(
+            "To enable coscheduling on the CPU Node Pool, multiple node pools must be used."
+            + " --cpu-coscheduler cannot be specified if --multi-node-pool isn't as well."
+        )
+
+
+def create_cluster(args: argparse.Namespace) -> None:
+    region = "-".join(args.zone.split("-")[:-1])
```
ioga, line 131: perhaps `s.rsplit('-', 1)[0]`?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+        cmd += ["--node-taints=gpuAvailable=True:NoSchedule"]
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+    cmd = [
+        "kubectl",
+        "apply",
+        "-f",
+        defaults.K8S_NVIDIA_DAEMON,
+    ]
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+    if args.multi_node_pool:
```
ioga, line 190: nit: maybe suggest add some single blank lines between major "stages" in the long functions, e.g. a blank line between gpu and cpu sections here.

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+            cpu_pod_spec["spec"]["schedulerName"] = "coscheduler"
+        helm_values["taskContainerDefaults"]["cpuPodSpec"] = cpu_pod_spec
+    with (helm_dir / "values.yaml").open("w") as f:
+        yaml.round_trip_dump(helm_values, f)
+
+
+def handle_up(args: argparse.Namespace) -> None:
+    try:
+        validate_args(args)
+    except Exception as e:
```
ioga, line 281: perhaps be more specific and only catch the type you throw, `ValueError`/`ArgumentError`?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+        cprint(
+            (
+                "Autoscaling is not supported with the lightweight coscheduling plugin and so a CPU"
+                + " node pool with {} nodes will be statically allocated. This can be changed by"
+                + "  using the default Kubernetes scheduler, or specifying a different value for"
+                + " --max-cpu-nodes."
+            ).format(args.max_cpu_nodes),
+            "yellow",
+        )
+    create_cluster(args)
```
ioga, line 306: what happens when you try to "upgrade" existing cluster?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+                            "--gpus-per-node",
+                            type=int,
+                            default=defaults.GPUS_PER_NODE,
+                            required=False,
+                            help="number of GPUs per node",
+                        ),
+                        Arg(
+                            "--helm-dir",
+                            type=str,
+                            default="helm/charts/determined",
```
ioga, line 422: does it work when your current working directory is not the repo root? 
if not, can we have the helm resources location "automatically discovered" to make it work?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+            "up",
+            handle_up,
+            "create gke cluster",
+            [
+                ArgGroup(
+                    "required named arguments",
+                    None,
+                    [
+                        Arg(
+                            "--cluster-name",
```
ioga, line 377: IIRC we've been using `--cluster-id` for `det deploy gcp|aws`, can we do the same here?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+from termcolor import cprint
+
+import determined
+from determined.common import yaml
+from determined.common.declarative_argparse import Arg, ArgGroup, Cmd, Group
+from determined.common.util import safe_load_yaml_with_exceptions
+from determined.deploy.gke.constants import defaults
+
+
+def make_spec(task_container_defaults: dict, key: str) -> dict:
```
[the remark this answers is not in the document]
RehanSD: Fixed - originally had `Dict` but changed it to `dict` to avoid some PyLance typing complaints.

harness/determined/deploy/gke/cli.py
RehanSD: If using a GPU cluster, I think its possible to use the coscheduler for the GPU but not the CPU, which is why they're separate.

harness/determined/deploy/gke/cli.py
RehanSD: I ended up breaking it into multiple methods, which helps with adding cpu only clusters!

harness/determined/deploy/gke/cli.py
RehanSD: I'm not sure I follow the question - do you mean update the settings or the helm chart?

ioga: either: a user runs `det deploy gke` with the same cluster id and the same or different flags. won't gcloud for cluster create fail in that case?

RehanSD: Ah yeah - that would be an issue but only if they're using the same project. I'm not sure it makes sense for the same user to be able to create multiple gke clusters with the same name - or did you mean for that to be a shortcut to tear down the existing cluster, and create a new one with the different flags?

ioga: >did you mean for that to be a shortcut to tear down the existing cluster, and create a new one with the different flags?
yes. or we can just show an error "cannot do that, please delete the cluster first".

when only helm chart parameters are updated, and we're upgrading the existing cluster, then we can skip the gke cluster portion, and just do `helm upgrade`.

harness/determined/deploy/gke/constants.py
```diff
@@ -0,0 +1,9 @@
+class defaults:
+    GPU_TYPE = "nvidia-tesla-k80"
+    GPUS_PER_NODE = 4
+    ZONE = "us-west1-b"
+    MASTER_MACHINE_TYPE = "n1-standard-16"
+    AGENT_MACHINE_TYPE = "n1-standard-32"
+    MAX_GPU_NODES = 4
+    MAX_CPU_NODES = 4
+    K8S_NVIDIA_DAEMON = "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nvidia-driver-installer/cos/daemonset-preloaded.yaml"
```
mackrorysd, line 9: Possibly the most useless code review comment of all time: there's a bunch of possibility for this floating around, it's badly documented which one you should use for which versions, etc., but this one is by far the best one to be using as far as I know. So... good job? :smile:

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,551 @@
...
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+    create_nodepools(region, args)
+    if not args.no_make_bucket:
+        cmd = ["gsutil", "mb", "gs://{}".format(args.gcs_bucket)]
+        subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+
+
+def create_nodepools(region: str, args: argparse.Namespace) -> None:
+    cmd = [
+        "gcloud",
```
mackrorysd, line 154: For the purpose of knowledge-sharing: this mostly comes up with A100s at the moment, but sometimes this will appear to fail because of capacity, even with 0 min nodes, and if so at a later point you could try to actually provision nodes and possibly be successful (or vice versa). I'm torn on whether or not the right UX is to actually treat that as a fatal error or just raise a warning about capacity.

harness/determined/deploy/gke/cli.py
mackrorysd: There's also a couple of different flavors of coscheduler. There's the one that just provides gang scheduling, there's an even more modified one that has more of a notion of priority, and another one that allows preemption. I'm not sure how many we plan to support long-term or if these are just superceding each other. @eecsliu Can you clarify?

harness/determined/deploy/gke/cli.py
ioga: following up: what happens if you run the same creation command twice? does that error look reasonable?

ioga (commented):
a few minor comments before we ship it, but looks good overall.

RehanSD:
The code has been written - the main concern here is testing. There are a number of flags that the user can set, that haven't been tested in conjunction. Primarily - testing that CPU only mode works, testing that the different schedulers are enabled and work, and checking that mutual exclusion constraints hold (i.e. that setting the cpu co scheduling flag without setting the gpu co scheduling flag raises an error, etc.)

PR #2845, opened 2021-08-18
fix: switch container runtime on slot type

stoksc:
## Description
Previous we set the default container runtime to `nvidia` regardless of slot type for the agent. This change moves us to selecting `runc` or `nvidia` based on the slot type, since nvidia-based GPU images won't launch on CPU agents with the `nvidia` runtime when drivers/etc are already baked in (it tries to run an container start hook and fails).

## Test Plan
- [x] run mnist const on a GPU cluster, launch a tensorboard. it should run fine without any finagling.
- [x] e2e tests

## Commentary (optional)
Seems like just an oversight we weren't doing this before.

## Checklist

### DIFF
--- a/master/pkg/tasks/task.go
+++ b/master/pkg/tasks/task.go
@@ -29,6 +29,12 @@ const (
 	certPath          = "/run/determined/etc/ssl/master.crt"
 )
 
+const (
+	// Container runtimes.
+	nvidiaContainerRuntime = "nvidia"
+	runc                   = "runc"
+)
+
 // TaskSpec defines the spec of a task.
 type TaskSpec struct {
 	// Fields that are only for task logics.
@@ -124,6 +130,14 @@ func (t *TaskSpec) ToDockerSpec() container.Spec {
 	}
 	envVars = append(envVars, env.EnvironmentVariables().For(deviceType)...)
 
+	containerRuntime := ""
+	switch deviceType {
+	case device.CPU, device.ZeroSlot:
+		containerRuntime = runc
+	case device.GPU:
+		containerRuntime = nvidiaContainerRuntime
+	}
+
 	network := t.TaskContainerDefaults.NetworkMode
 	if t.UseHostMode {
 		network = hostMode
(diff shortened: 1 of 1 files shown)


master/pkg/tasks/task.go
```diff
@@ -124,6 +130,14 @@ func (t *TaskSpec) ToDockerSpec() container.Spec {
 	}
 	envVars = append(envVars, env.EnvironmentVariables().For(deviceType)...)
 
+	containerRuntime := ""
+	switch deviceType {
+	case device.CPU, device.ZeroSlot:
+		containerRuntime = runc
+	case device.GPU:
+		containerRuntime = nvidiaContainerRuntime
+	}
```
ioga, line 139: safety default?

```
default:
    panic(fmt.Sprintf("bad device type: %s", deviceType))
```

ioga (approved):
lgtm

master/pkg/tasks/task.go
stoksc: for now it should just run the default docker container runtime if there is an unrecognized device, which i think is fine. anyway, i think we'll panic elsewhere if it was invalid.

master/pkg/tasks/task.go
```diff
@@ -124,6 +130,14 @@ func (t *TaskSpec) ToDockerSpec() container.Spec {
 	}
 	envVars = append(envVars, env.EnvironmentVariables().For(deviceType)...)
 
+	containerRuntime := ""
+	switch deviceType {
+	case device.CPU, device.ZeroSlot:
+		containerRuntime = runc
+	case device.GPU:
+		containerRuntime = nvidiaContainerRuntime
```
stoksc, line 138:
```suggestion
		containerRuntime = nvidiaContainerRuntime
	default:
		panic(fmt.Sprintf("bad device type: %s", deviceType))
```

master/pkg/tasks/task.go
stoksc: eh supposed it cant hurt.

ioga: I remember using some [functional] programming language where one 
1. had to make a string enum 
2. do a switch on that strongly typed enum (instead of raw const/literal strings) 
3. compiler would complain if you didn't have a `case` for each possible type value, or a default. 

it was opinionated but helpful chunk of static analysis.

stoksc: yeah, i definitely like that. i think go help some linter for that (found it https://github.com/nishanths/exhaustive)

PR #2847, opened 2021-08-18
fix: make entrypoint startup-hook.sh eval consistent [DET-5874]

varlogtim:
It is possible for a user to change out of the WORKING_DIR while
inside of startup-hook.sh. This will cause import errors.

[DET-5874]

## Test Plan

Manually tested.


rb-determined-ai:
sorry for the long delay on this one.  I think there's a couple things here to discuss.

* We started adding these `pushd`/`cd` directives a long time ago.  Pretty sure that when we did that, we weren't setting the workdir for the overall container.  We do that right now, unconditionally, so we should just simplify these to be e.g.:

  ```sh
    test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
  ```

  The original intention of the `popd` was not to undo what the user did in their startup hook, but to return to wherever-the-container started.  Again, since we set workdir for the container, there's no point in a `popd` at all.

  > It is possible for a user to change out of the WORKING_DIR

  Yes, that is possible.  That is fine; their script should either include its own popd or they should cd into a sutiable directory that won't have import errors.  It's not our job to anticipate what they might do in startup hook and undo it.  In fact, the very purpose of the startup-hook exists to let users make changes to the starting conditions in the container, and we should not be attempting to undo those changes.

* @shiyuann is adding support for changing the workdir of the container as part of the experiment/command configs.  After that support lands, I expect we will still be setting the workdir for the container unconditionally, in which case I think the strategy above (not pushd'ing into WORKDIR and not popd'ing out of it) is still valid.  But Shiyuan should also be aware of this change since it's pretty relevant to his current work.

master/static/srv/shell-entrypoint.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 "$DET_PYTHON_EXECUTABLE" -m pip install -q --user /opt/determined/wheels/determined*.whl
 
-pushd ${WORKING_DIR} && test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}" && popd
+cd ${WORKING_DIR} && test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
```
rb-determined-ai, line 23: remove `cd ${WORKING_DIR} &&`, as I mentioned in my comment.

Also remove the WORKING_DIR variable entirely while you are at it.

varlogtim: Done!

master/static/srv/entrypoint.sh
```diff
@@ -46,13 +45,14 @@ fi
 # training, so we try to query the user system for a valid HOME, or default to
 # the working directory otherwise.
 if [ "$HOME" = "/" ] ; then
-    HOME="$(set -o pipefail; getent passwd "$(whoami)" | cut -d: -f6)" || HOME="$WORKING_DIR"
+    HOME="$(set -o pipefail; getent passwd "$(whoami)" | cut -d: -f6)" || HOME=$(pwd)
```
rb-determined-ai, line 48: Can we do `"$PWD"` instead of `$(pwd)`?

master/static/srv/entrypoint.sh
```diff
@@ -46,13 +45,14 @@ fi
 # training, so we try to query the user system for a valid HOME, or default to
 # the working directory otherwise.
 if [ "$HOME" = "/" ] ; then
-    HOME="$(set -o pipefail; getent passwd "$(whoami)" | cut -d: -f6)" || HOME="$WORKING_DIR"
+    HOME="$(set -o pipefail; getent passwd "$(whoami)" | cut -d: -f6)" || HOME=$PWD
```
rb-determined-ai, line 48: sorry to send it back one more time... but you need to use quotes when dereferencing a variable.

PR #2727, opened 2021-07-13
feat: Push Arch Feature Branch

rb-determined-ai:
## Description

The whole feature branch for Push Architecture.  This exists for testing and dev deployments.

## Merging Plan

This PR will be merged when it is fully reviewed, tested, and dogfooded.

## Reviewing Plan

The python commits are available for individual review at https://github.com/determined-ai/determined/pull/2771.

PR #2855, opened 2021-08-19
feat: push architecture (webui)

hamidzr:
## Description

- remove workload states
- display trial runner state
- remove dependence on `num_inputs`

## Commentary (optional)

I separated the WebUI portion of the push architecture project (aka sprinkle API) into another PR to hopefully make it easier to review. https://github.com/determined-ai/determined/pull/2855 . 

tests will fail and the react code will not compile due to missing API changes. If you'd like to see the full picture take a look at the parent pr here https://github.com/determined-ai/determined/pull/2727

## Checklist


webui/react/src/services/decoder.ts
```diff
@@ -398,8 +395,11 @@ export const decodeTrialResponseToTrialDetails = (
     }));
   }
 
+  const EMPTY_STATES = new Set([ 'UNSPECIFIED', '', undefined ]);
+
   return {
     ...trialItem,
+    runnerState: EMPTY_STATES.has(data.trial.runnerState) ? undefined : data.trial.runnerState,
```
hkang1, line 402: **question:** not 100% sure but I think `data.trial.runnerState` should probably be `data.trial.state`?

hkang1: some extra typescript errors popped up after fixing this, but I ended up with something ugly like this lol:
```
  const EMPTY_STATES = new Set([ 'UNSPECIFIED', '', undefined ]);
  const trialState: string = data.trial.state as unknown as string;

  return {
    ...trialItem,
    runnerState: EMPTY_STATES.has(trialState) ? undefined : trialState,
    workloads: workloads || [],
  };
```

hkang1: **question:** can `data.trial.state` be something other than the enumerable values such as empty string or `undefined`? If not it can be some more simple:
```
  const isEmptyState = data.trial.state === Sdk.Determinedexperimentv1State.UNSPECIFIED;

  return {
    ...trialItem,
    runnerState: isEmptyState ? undefined : data.trial.state as unknown as string,
    workloads: workloads || [],
  };
```

webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
```diff
@@ -50,6 +49,11 @@ const TrialInfoBox: React.FC<Props> = ({ trial, experiment }: Props) => {
   return (
     <Section>
       <Grid gap={ShirtSize.medium} minItemWidth={180} mode={GridMode.AutoFill}>
+        {trial.runnerState &&
+          <OverviewStats title="Last Runner State">
+            {trial.runnerState}
```
hkang1, line 54: **suggestion:** what do you think about using the `<Badge>` component instead of pure text here?

hkang1: Screen Shot 2021-08-24 at 9 20 41 AM

webui/react/src/utils/workload.ts
```diff
@@ -1,12 +1,12 @@
-import { CheckpointState, CheckpointWorkload, MetricsWorkload, Step,
+import { CheckpointWorkload, MetricsWorkload, Step,
   WorkloadWrapper } from '../types';
 
 export const hasCheckpointStep = (step: Step): boolean => {
-  return !!step.checkpoint && step.checkpoint.state !== CheckpointState.Deleted;
+  return !!step.checkpoint;
```
hkang1, line 5: **question:** help me understand this part. Is it that we no longer support checkpoint deletion or should checkpoints be shown even if deleted?

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.table.tsx
```diff
@@ -18,17 +18,6 @@ export const columns: ColumnType<Step>[] = [
...
-    key: 'state',
-    render: (text: string, record: Step, index: number): React.ReactNode =>
-      stateRenderer(text, record.training, index),
-    sorter: (a: Step, b: Step): number => {
-      // Previously the list of steps would list steps by training steps which
-      // all have RunState type as their state.
-      return runStateSorter(a.training.state, b.training.state);
-    },
-    title: 'State',
-  },
```
hkang1, line 31: yes! good riddance, thanks for dropping the workload state column!

hkang1 (commented):
This looks great! There's one area that needs fixing (I think) and a question regarding checkpoint delete.

I ran into a weird issue where when I activate an experiment, after it runs, the `Last Running State` doesn't popup. This is most likely not caused by this PR and is an existing problem. I think we made the call that state change + not re-triggering polling was considered ok? Maybe we want to reconsider that as this one is more obvious now. More specifically, when we activate a paused experiment, we should maybe start polling? Another alternative would be to use the `reload` reroute to trigger the polling.

webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
hamidzr: the text that comes through would be anything that user sets but we can stil wrap that in a badge 🤔 
let me see how a badge would look with some potential values

webui/react/src/services/decoder.ts
hamidzr: it can be an empty string for sure. but we shouldn't really get into undefined. we could at the decoder level make it undefined if it's in any of the empty states and that might make the usage easier.

hamidzr: > question: not 100% sure but I think data.trial.runnerState should probably be data.trial.state?

wouldn't that get confusing with the existing trial state?

webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
hamidzr: I thought more about this. with our badging we usually specify meaningful colors to states. here we can't really do that since it's user defined.  wdy think?

webui/react/src/utils/workload.ts
hamidzr: I chatted with @stoksc  and as a result, I brought back the checkpoint state.

webui/react/src/services/decoder.ts
hkang1: What I meant was `data.trial` is of type `Trialv1Trial` which has `state` as a property but not `runnerState`:
Screen Shot 2021-08-31 at 9 46 11 AM

webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
hkang1: oh I had no idea the runner state is going to be user defined! hmmm yeah if that's the case the colors don't mean anything to them, it's purely our experiment state we expose to them. Does this mean `Runner State` will co-exist with our internal `State` or does it replace it (thinking what would happen to experiment list page and all the pages that reference states like `Active`, `Paused`, etc).

hkang1: but yeah, leaving them as how you have it (just text) makes sense to me.

hkang1 (commented):
I just have that [build issue](https://github.com/determined-ai/determined/pull/2855/files#r699456548) that needs resolving (not sure if it's my setup), but the rest looks great

webui/react/src/services/decoder.ts
hamidzr: aah, this is expected this branch is missing the backend and api changes that are needed.

hamidzr: could you check with the main push architecture branch see if that works fine for you? (the tests pass so at least should be fine)

hamidzr:
> I just have that [build issue](https://github.com/determined-ai/determined/pull/2855/files#r699456548) that needs resolving (not sure if it's my setup), but the rest looks great


thanks for the re-review. replied to the build issue

hamidzr:
@hkang1  I realized I had missed a commit from the push branch here. I brought it over https://github.com/determined-ai/determined/pull/2855/commits/7042aa89db335406aead2df393e5ca2b20245b70

hkang1 (approved):
This looks great, thanks for clarifying and adding webui push architecture support!

hkang1:
adding some notes from our async chat (not necessary to add to this PR but a nice-to-have):
- add ellipsis overflow to status info box

hamidzr:
> adding some notes from our async chat (not necessary to add to this PR but a nice-to-have):
> 
> * add ellipsis overflow to status info box

I ended up putting this fix separately since 1. it's a valid useful patch on its own and 2. didn't want to force update this and the other push pr. https://github.com/determined-ai/determined/pull/2899

PR #2863, opened 2021-08-23
fix: uPlot to show values of 0

EmilyBonar:
## Description
uPlot was treating values of 0 as the same as not having any data, which was incorrect. This makes it display 0 as a valid value.

## Test Plan
Test by making an experiment return a metric with a value of 0, then view it on the Metrics graph on the Trial Details page (or Experiment Overview for single-trial experiments).

## Checklist

### DIFF
--- a/docs/release-notes/2863-fix-uplot-0-values.txt
+++ b/docs/release-notes/2863-fix-uplot-0-values.txt
@@ -1,0 +1,5 @@
+:orphan:
+
+**Bug Fixes**
+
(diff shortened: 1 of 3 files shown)


hkang1 (approved):
This looks good. There's one more area that I think needs to update in order to detect tooltip:

in `tooltipsPlugin.ts` around line 126:
```
            serie.scale !== 'x' && serie.show && uPlot.data[serieId][idx] != null
```

PR #2864, opened 2021-08-23
chore: make MetricsBatcherThread safer

rb-determined-ai:
## Description
chore: make MetricsBatcherThread safer:

 - Avoid hanging when exceptions are raised
 - Avoid accidentally calling queue.get() with timeout <= 0
 - Avoid crashing on empty metrics
 - Avoid publishing empty metrics altogether

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -212,6 +212,16 @@ class ShutdownMessage:
     pass
 
 
+def yield_timeouts_until(deadline: float) -> Iterator[float]:
+    """Never yield a timeout which is 0 or less; that would cause queue.get() to hang."""
+    while True:
+        timeout = deadline - time.time()
+        if timeout <= 0:
+            raise StopIteration()
+        else:
+            yield timeout
+
+
 def profiling_metrics_exist(master_url: str, trial_id: str) -> bool:
     """
     Return True if there are already profiling metrics for the trial.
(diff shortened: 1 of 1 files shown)


harness/determined/profiler.py
```diff
@@ -760,6 +763,7 @@ def run(self) -> None:
                     for msr in self.accumulating_measurements.values():
                         self.metrics_batch.append(msr.metric_type, msr.metric_name, msr)
                     self.accumulating_measurements = {}
+                    break
```
stoksc, line 766: why this break? we may not necessarily have timed out yet, right?

harness/determined/profiler.py
```diff
@@ -768,21 +772,16 @@ def run(self) -> None:
...
-            batch_start_time = cast(float, batch_start_time)
-            if time.time() - batch_start_time > self.FLUSH_INTERVAL:
+            # Timeout met, or an explicit flush.
+            if not self.metrics_batch.isempty():
                 self.send_queue.put(self.metrics_batch.consume())
-                batch_start_time = time.time()
+            deadline = time.time() + self.FLUSH_INTERVAL
+
+    def run(self) -> None:
+        try:
```
stoksc, line 781: yeah, this is smarter. so i guess if this profiler crashed the sending thread hung waiting for the shutdown message?

harness/determined/profiler.py
rb-determined-ai: Maybe I misunderstood the former code.  This is not intended to force a flush then?

harness/determined/profiler.py
rb-determined-ai: Yes, that was causing the hangs in our push e2e tests.

harness/determined/profiler.py
rb-determined-ai: fixed.

rb-determined-ai:
not sure what's going on with netlify, but I'm landing this now.

PR #2882, opened 2021-08-30
fix: update batch inference example

rb-determined-ai:
## Description

There's some issues with our documented hack for batch inference.

The fixes are rather complex and would vary user-by-user, so I'm not quite sure what's the best way forwards here.

### DIFF
--- a/examples/computer_vision/cifar10_pytorch_inference/model_def.py
+++ b/examples/computer_vision/cifar10_pytorch_inference/model_def.py
@@ -48,9 +48,16 @@ class PredictionsReducer(MetricReducer):
     def update(self, predictions):
         # We are responsible for calling update() as part of our train_batch() and evaluate_batch()
         # methods, which means we can specify any arguments we wish.
+        # XXX: If users wanted to configure streaming writers to start writing predictions as soon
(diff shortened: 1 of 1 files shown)


examples/computer_vision/cifar10_pytorch_inference/model_def.py
```diff
@@ -84,6 +93,8 @@ def __init__(self, context: PyTorchTrialContext) -> None:
 
         # TODO: Load your trained model. Below are example approaches.
 
+        # XXX: this is a bad idea; users frequently try to load checkpoints via checkpoint export
+        #      but due to DET-3448 that is highly likely to result in opaque, difficult errors.
```
ahj-determined, line 97: We could remove this, but I'm worried that folks are going to be trying to load their model programmatically anyway. Should we keep the code but add a user-facing warning, similar to the comment here?

Do we have an alternative way of loading in a checkpoint that's not problematic that we could suggest?

rb-determined-ai: Yes, setting `searcher.source_checkpoint_uuid` should work fine for checkpoints in the system.

rb-determined-ai:
closing due to staleness

PR #2886, opened 2021-08-30
chore: add support for throughput profile chart [DET-5596, DET-5732, DET-5913, DET-5923]

hkang1:
## Description

- Add throughput chart to profiler.
- Order the charts as Throughput, Timing Metrics then System Metrics.
- Remove 5 min window and wrap chart to whatever data available.
- Enable cursor sync between the charts.
- Prevent system metrics from re-rendering when changing filters.

Screen Shot 2021-09-01 at 8 48 09 AM

## Checklist


webui/react/src/pages/TrialDetails/Profiles/ProfilesEnabled.tsx
```diff
@@ -1,66 +1,215 @@
...
+        bodyNoPadding
+        filters={<SystemMetricFilter />}
+        loading={metrics[MetricType.System].isLoading}
+        title="System Metrics">
+        <UPlotChart
+          data={chartInfo[MetricType.System].data}
+          options={chartInfo[MetricType.System].options}
+          style={CHART_STYLE}
+        />
       </Section>
```
EmilyBonar, line 212: bug: I don't know what's causing it, but when I switch system metrics, I am taken back up to the top of the page, which I didn't expect.

EmilyBonar (approved):
Looks good!

webui/react/src/pages/TrialDetails/Profiles/ProfilesEnabled.tsx
hkang1: oooo... this one is a doozie... it's not obvious but all 3 charts have an interdependency because they are now synced so when the filters update the system metrics, both throughput and timing metrics will have to re-render. So ultimately the entire section re-renders, where we lose the scroll position...

hkang1: ended up using a `useScroll` hook to remember the last scroll position and restoring it when chart changes.

Issue #2630, opened 2021-06-28
Docker determined-agent:latest tag not defined


BlueskyFR:
First of all I am very sorry for posting this here because I did not know where to put it.

In some doc pages, it is mentionned to pull `determinedai/determined-agent` using these commands:
![image](https://user-images.githubusercontent.com/25421460/123694969-474fcb00-d85a-11eb-9f73-b09d97a9cfd0.png)
![image](https://user-images.githubusercontent.com/25421460/123694990-4f0f6f80-d85a-11eb-8bff-40522cbbaa21.png)

However, these command will fail with a manifest error from Docker because the `:latest` tag (`determinedai/determined-agent:latest`) is not defined on the image.
I suggest for it to be equals to the latest stable release, so `0.16.0` as of now.

It is important IMO as Docker tries to read `:latest` by default when nothing is specified, and thus allows your customers to always run the latest version.

BlueskyFR:
Also, please note that the command shown on the first screen:
```bash
docker run \
  -v `pwd`/agent-config.yaml:/etc/determined/agent-config.yaml \
  determinedai/determined-agent
  --config-file /etc/determined/agent-config.yaml
```
Is false for 3 reasons:

1. A backslash is missing at line 3
2. The `--config-file` option is not supported by the determined-agent in the 0.16.0 release
3. The .yaml filename should be `agent.yaml` instead of `agent.yaml`

vishnu2kmohan:
Thanks for the report @BlueskyFR, we'll fix this up.

BlueskyFR:
Still no news?

vishnu2kmohan:
Thanks to @dzhu, we will have `latest` tags pushed, for convenience, starting with the 0.16.5 release, expected later this week or early next week.

Thanks for the feedback @BlueskyFR!

PR #2887, opened 2021-08-30
docs: update k8s version to 1.19 >= and <= 1.21

varlogtim:


vishnu2kmohan:
@varlogtim Just checking; do we need any updates to the helm charts to support K8s >=1.19 and <=1.21?

docs/how-to/installation/kubernetes.txt
```diff
@@ -16,7 +16,8 @@ For general information about using Determined with Kubernetes, refer the
 Before installing Determined on a Kubernetes cluster, please ensure that the following prerequisites
 are satisfied:
 
--  The Kubernetes cluster should be running Kubernetes >= 1.15
+-  The Kubernetes cluster should be running Kubernetes versoin >= 1.19 and <= 1.21 (though later
```
vishnu2kmohan, line 19:
```suggestion
-  The Kubernetes cluster should be running Kubernetes version >= 1.19 and <= 1.21, though later
```

vishnu2kmohan (changes requested):
Minor nits, looks good otherwise!

docs/how-to/installation/kubernetes.txt
```diff
@@ -16,7 +16,8 @@ For general information about using Determined with Kubernetes, refer the
...
+   versions may work)
```
vishnu2kmohan, line 20:
```suggestion
   versions may work.
```

vishnu2kmohan (changes requested):
One more minor suggestion.

ioga:
By the way, our CI uses 1.18, so I'd expect it to work as well.

varlogtim:
> @varlogtim Just checking; do we need any updates to the helm charts to support K8s >=1.19 and <=1.21?

Good catch @vishnu2kmohan . It actually looks like the helm chart links, at least in the generated docs.determined.ai, don't appear to work either.

docs/how-to/installation/setup-eks-cluster.txt
```diff
@@ -238,7 +249,7 @@ file such as ``determined-autoscaler.yaml``:
              value: "true"
              effect: NoSchedule
          containers:
-           - image: k8s.gcr.io/autoscaling/cluster-autoscaler:v1.17.3
+           - image: <cluster-autoscaler-image>
```
eecsliu, line 252: non-blocking: should we reiterate the link to autoscalers here? I have a feeling people might not read...

eecsliu (approved):
👩‍🍳😗👌

varlogtim:
@vishnu2kmohan I think some how, through force pushing changes, your last suggestion was lost. Could you re-review? Also, I would like to get your thoughts on Eric's comment.

docs/how-to/installation/setup-eks-cluster.txt
varlogtim: Done!

vishnu2kmohan:
FYI - I chatted to @eecsliu about improving our release process to add/publish a `latest` (via a symlink to the) helm chart so we don't have to constantly bump the version in the docs when there's an update (similar to [this](https://github.com/determined-ai/determined/issues/2630) - cc @dzhu).

varlogtim:
> FYI - I chatted to @eecsliu about improving our release process to add/publish a latest (via a symlink to the) helm chart so we don't have to constantly bump the version in the docs when there's an update (similar to this - cc @dzhu).
Got it. I think I should be able to add this into the review today and remove the other code. This is a much better solution, in my opinion.

vishnu2kmohan (approved):

PR #2894, opened 2021-09-01
fix: change "stop and save checkpoint" from checkbox to button [DET-5837]

EmilyBonar:
## Description
Before:
index
After:
Screenshot 2021-09-01 at 09-21-56 Experiment 1817 - Determined

Rationale:
This removes a click for users to save existing work, while not adding a click for previous default behavior.

## Checklist

### DIFF
--- a/webui/react/src/components/StopExperimentModal.tsx
+++ b/webui/react/src/components/StopExperimentModal.tsx
@@ -17,7 +17,7 @@ interface Props extends Omit<ModalFuncProps, 'type'> {
 }
 
 const StopExperimentModal: React.FC<Props> = ({ experiment, onClose }: Props) => {
-  const [ type, setType ] = useState<ActionType>(ActionType.Kill);
+  const [ type, setType ] = useState<ActionType>(ActionType.Cancel);
 
   const onOk = useCallback(async () => {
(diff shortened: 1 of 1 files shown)


hkang1 (approved):
Code looks good! Going to pull in @alexjdetermined regarding the style. With 3 buttons now and the 2 confirmation buttons as "destructive" choices, what do you recommend we do for styling?

alexjdetermined:
> Code looks good! Going to pull in @alexjdetermined regarding the style. With 3 buttons now and the 2 confirmation buttons as "destructive" choices, what do you recommend we do for styling?

Few options:

Opt A: (**my recommendation**) keep the action buttons to two but have an "opt-out" checkbox for capturing checkpoint before stopping. Optionally (not sure if the effort is worth the value), if they uncheck that option we can show a message of "Any progress on incomplete workflows will be lost."

Opt B: Separate the destructive option from the primary action button group. 

image

EmilyBonar:
Changed the design to Option A
Screenshot 2021-09-07 at 10-13-52 Experiment 1817 - Determined
Screenshot 2021-09-07 at 10-14-09 Experiment 1817 - Determined

PR #2898, opened 2021-09-01
build: remove unneeded build-bindings dependencies

hamidzr:
## Description

since we now commit our generated code for WebUI tests don't need to depend on `build-bindings`
anymore

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1852,7 +1852,6 @@ workflows:
 
       - test-e2e-webui:
           requires:
-            - build-bindings
             - package-and-push-system-local
 
       - deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1760,13 +1760,7 @@ workflows:
   test-unit:
     jobs:
       - test-unit-go
-      - build-proto
-      - build-bindings:
-          requires:
-            - build-proto
-      - test-unit-react:
-          requires:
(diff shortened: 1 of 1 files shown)


hamidzr:
> if we commit the generated code, can we rip out build-bindings altogether from everywhere in ci? perhaps except for lint-react which intentionally builds them to check they've been rebuilt

yeah good catch. we can probably do that let me see.

hamidzr:
leaving on in build-e2e to test that we can build bindings.
this is tested in lint-react but it's cheap and might be good
to have its own representation.

We also want to testing bindings out of react and into the bindings
module in a separate pr.

ioga:
> 
> 
> leaving on in build-e2e to test that we can build bindings.
> this is tested in lint-react but it's cheap and might be good
> to have its own representation.
> 
> We also want to testing bindings out of react and into the bindings
> module in a separate pr.

I'd remove it from `test-e2e`, I really think having them in `lint` is sufficient. `lint` runs on every commit as well, why burn extra circleci hours.

hamidzr:
hmm if we had a "build to make sure it builds" workflow I'd put it there alongside build storybook

ioga:
the lint requirement "make sure the build is up to date" ends up being strictly stronger than "make sure it builds".

ioga (approved):

PR #2900, opened 2021-09-02
feat: add a `make devcluster` target

rb-determined-ai:


tools/devcluster.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+# Three stages: db, master, and agent.
+stages:
+  - db:
+      port: 5432
+      db_name: determined
+      password: postgres
+      container_name: determined_db
+      image_name: "postgres:10.14"
+
```
ioga, line 17: maybe trim all unnecessary & default configuration and just make this minimal? 
"advanced" users with custom configs won't use `make devcluster`, and they are capable of going to devcluster docs to see the examples.

ioga (approved):

tools/devcluster.yaml
rb-determined-ai: The purpose of expanding default values here isn't for discoverability though.  I think it's bad practice for this to rely on defaults in this file.  Reason being is that defaults can change, and it's important that `make devcluster` be as robust as possible, far into the future.

If I have to check out determined version `upstream/master~8months` I want `make devcluster` to just work, even if I'm on a much newer devcluster that services `upstream/master` with its defaults.

It might make sense to leave the `stage.db.image_name` out, since that should be ok to bump without negative effects.  Are there other specific defaults that you would like to see removed?

ioga: it seems to me then you have to nail down the devcluster version to specific 1.0.0. Even if we try not to break things before 2.0, we don't know how behavior can change.

rb-determined-ai: No, I think that's the beauty of the `devcluster<2` specification.  You just make sure that `devcluster` keeps backwards compatibility to 1.0.  When that's not possible, only then do you bump devcluster version to 2.0.

But minor improvements, like if devcluster handled the dangling `determined_db` instance better, could go into v1.1, and that version of devcluster could operate with this version of determined just fine.

The goal is that improvements-of-life in devcluster continue to provide benefit to developers, even if you have to go back and play with an old version of determined.  Actually, this was one of the major goals I started with... before devcluster, I had a huge pile of copy-on-write versioned scripts that could start old versions of our software (which was less stable, so there were major differences in scripts).  The purpose of extensive configurability of devcluster was to allow one tool with one user experience to operate on a large swath of determined versions.

ioga: fine.

PR #2903, opened 2021-09-02
chore: log warnings on critical failures

stoksc:
## Description
This change adds scary and obvious messages when the Determined resource manager crashes. Previously the master used the default "rootActor" which ignored all messages. This adds a root actor just for the master that warns on failures of critical sub-components of Determined and, if it returns an error, it has the potential to crash the root actor (though it doesn't do that quite yet). Previously, there were incidents where the `pods` actor crash didn't result in a system crash and we never recovered. The pods actor crashes were addressed in #2880, this is just the next in a line of improvements.

Nice to have future work after this PR would be to handle cleaning up stray agents/pods on startup (easy) or recovering them properly (hard) and causing RM crashes to crash the master (returning an error from this root actor).

## Test Plan
- [x] crash the RM in weird ways, make sure the system dies.

## Checklist

### DIFF
--- a/master/internal/root.go
+++ b/master/internal/root.go
@@ -1,0 +1,36 @@
+package internal
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/internal/sproto"
+	"github.com/determined-ai/determined/master/pkg/actor"
+)
+
+// root manages the lifecycle of all actors in the Determined master and
+// defines a supervision strategy specifically for the master.
+func root(ctx *actor.Context) error {
+	switch msg := ctx.Message().(type) {
+	case actor.PreStart, actor.PostStop:
+		return nil
+	case actor.ChildFailed:
+		switch msg.Child.Address() {
+		case sproto.K8sRMAddr, sproto.AgentRMAddr:
+			return errors.Wrapf(msg.Error, "resource mananger crashed, exiting")
+		case sproto.PodsAddr, sproto.AgentsAddr:
+			return errors.New("pods or agents manager crashed, exiting")
+		}
+		return nil
+	case actor.ChildStopped:
+		switch msg.Child.Address() {
+		case sproto.K8sRMAddr, sproto.AgentRMAddr:
+			return errors.New("resource manager stopped, exiting")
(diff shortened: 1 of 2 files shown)


master/internal/root.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/internal/sproto"
+	"github.com/determined-ai/determined/master/pkg/actor"
+)
+
+// root manages the lifecycle of all actors in the Determined master and
+// defines a supervision strategy specifically for the master.
+func root(ctx *actor.Context) error {
```
stoksc, line 12: my only concern with this is a retry loop where we spawn agents, crash and lose them over and over again. we might should land some sort of agent/task recovery first

stoksc: It may be better, rather than closing, to just print massive warnings that say to clean up resource, restart and report the error to us

master/internal/root.go
```diff
@@ -0,0 +1,36 @@
...
+	switch msg := ctx.Message().(type) {
+	case actor.PreStart, actor.PostStop:
+		return nil
+	case actor.ChildFailed:
+		switch msg.Child.Address() {
+		case sproto.K8sRMAddr, sproto.AgentRMAddr:
+			return errors.Wrapf(msg.Error, "resource mananger crashed, exiting")
+		case sproto.PodsAddr, sproto.AgentsAddr:
+			return errors.New("pods or agents manager crashed, exiting")
```
dzhu, line 21: Maybe wrap here as well for consistency? Though, actually, right now the error is going to show up three times (once when the actor fails initially, once when the root actor fails, and finally from the new `log.withError` in `core.go`), so it might look nicer to skip the wrap entirely instead.

master/internal/root.go
dzhu: Yeah, I think I would very slightly lean toward just logging. Just keeping it a little bit less potentially disruptive as a first step. I think either way is fine, though.

dzhu (approved):

master/internal/root.go
stoksc: Yah, I'll start with a log.. just incase.

PR #2910, opened 2021-09-03
chore: rewrite primary resource allocation query over public.allocations [DET-5972]

stoksc:
## Description
This change fixes up the primary query behind our historical resource allocation features to use the `public.allocations` table instead. Currently only trial allocations write to the table, but shortly all task types will.

## Test Plan
- [x] run the aggregation query with some experiments, validate the results.

## Checklist


examples/tutorials/mnist_pytorch/const.yaml
```diff
@@ -1,3 +1,5 @@
+labels:
+  - 'okness'
```
stoksc, line 2:
```suggestion
```

master/static/migrations/20210913111416_backfill-last-day-of-allocations.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Nothing to do for down, the tables didn't exist prior to this version and they are dropped
+-- by other migrations.
```
stoksc, line 2:
```suggestion
-- by other migrations.

```

master/static/migrations/20210913111416_backfill-last-day-of-allocations.up.sql
```diff
@@ -0,0 +1,76 @@
+UPDATE
```
stoksc, line 1: i did this separately because i'd rather it fail separately if it is going to

master/static/srv/update_aggregated_allocation.sql
```diff
@@ -5,67 +5,45 @@ WITH const AS (
...
-        AND trials.experiment_id = experiments.id
-        AND (
-            -- Sometimes experiments end uncleanly, leaving workloads permanently without end
-            -- times. Anything in that state must be excluded or it'll show up in all the statistics
-            -- forevermore.
-            upper(all_workloads.range) IS NOT NULL
-            OR experiments.state = 'ACTIVE'
-        )
+        const.period && a.range
+        AND upper(a.range) IS NOT NULL
```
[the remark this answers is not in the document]
stoksc: yeah, i suppose that's true, missed the importance of the ACTIVE check. i've been _really, really_ careful to not leave allocations in ACTIVE. it still worries me though. sort of impossible to know DB side though if it's really stuck or just not closed out.

master/static/srv/update_aggregated_allocation.sql
```diff
@@ -5,67 +5,45 @@ WITH const AS (
...
-        ) * (
-            experiments.config -> 'resources' ->> 'slots_per_trial'
-        ) :: float AS seconds
+                upper(const.period * a.range) - lower(const.period * a.range)
+        ) * (a.slots) :: float AS seconds
     FROM
         (
             SELECT
-                trial_id,
+                *,
```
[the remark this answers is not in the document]
stoksc: I'm really in favor of * here, since semantically this is "select all allocations with the range precalculated".. it's essentially just a view of the allocations table. if we didn't use *, this would need to be changed any time the table changed.

master/static/srv/update_aggregated_allocation.sql
```diff
@@ -5,67 +5,45 @@ WITH const AS (
...
+                allocations
+        ) AS a,
         const
     WHERE
         -- `&&` determines whether the ranges overlap.
-        const.period && all_workloads.range
-        AND all_workloads.trial_id = trials.id
-        AND trials.experiment_id = experiments.id
-        AND (
-            -- Sometimes experiments end uncleanly, leaving workloads permanently without end
```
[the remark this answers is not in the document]
stoksc: just removed the condition entirely.

master/static/srv/update_aggregated_allocation.sql
```diff
@@ -5,67 +5,44 @@ WITH const AS (
...
-            -- times. Anything in that state must be excluded or it'll show up in all the statistics
-            -- forevermore.
-            upper(all_workloads.range) IS NOT NULL
-            OR experiments.state = 'ACTIVE'
-        )
+        const.period && a.range
```
dzhu, line 30: Alright, if there's no way now to know whether something is still active or not, this is good. Would such a thing be feasible to add at some point?

master/pkg/model/task.go
```diff
@@ -20,13 +14,21 @@ func NewAllocationID(name string) AllocationID {
...
 	AllocationID AllocationID `db:"allocation_id"`
+	TaskID       TaskID       `db:"task_id"`
+	Slots        int          `db:"slots"`
+	AgentLabel   string       `db:"agent_label"`
 	ResourcePool string       `db:"resource_pool"`
 	StartTime    time.Time    `db:"start_time"`
 	EndTime      *time.Time   `db:"end_time"`
 }
 
+// TaskID is the unique ID of a task among all tasks.
```
dzhu, line 26: Put these above the struct, next to `AllocationID`/`NewAllocationID`?

dzhu (approved):

master/static/srv/update_aggregated_allocation.sql
stoksc: I mean, the way I see it, if we can't guarantee to write the end time, we can't guarantee to write some state change to 'final' that means the end time should be assumed invalid and it should not be aggregated (e.g. like the state `experiment.state = 'ERRORED' AND upper(all_workloads.range) IS NULL` previously would've been caused by a trial actor crash and excluded, it's entirely possible in the event of database network issues that the trial _and_ experiment actor could've crashed and left us as `experiment.state = 'ACTIVE' AND upper(all_workloads.range) IS NULL` and the workloads would've counted forever right?

I think the best we can do, probably, is to mark all active jobs/tasks/allocations end_time = current_timestamp on a master restart, as a "just in case" sort of measure.

stoksc: To get the old behavior (2 db writes need to fail to cause issues), we could take greatest(upper(allocation_range), upper(task_range)). I just assumed db write failures would be highly correlated in such a short time span so that complexity wouldn't really be worth the hassle.

master/pkg/model/task.go
stoksc: I was just keeping all task stuff together and all allocation stuff together in the file

master/static/srv/update_aggregated_allocation.sql
dzhu: Yeah, clearing out stuff on restart sounds good enough to me and is roughly what I was imagining. Can't do anything until the next restart anyway. I think that matches well enough with what happens with an experiment, where if it stays in ACTIVE because of a master crash, then the next time the master starts, it'll either crash and move out of ACTIVE soon or continue running and eventually finish.

stoksc: I'll go ahead and just tack that onto this pr, since it's just running `update allocations set end_time = current_timestamp where end_time is null;`

master/pkg/model/task.go
stoksc: I did ^ this reordering in https://github.com/determined-ai/determined/pull/2938/files already so I'll just leave it as it and then resolve conflicts to keep it that way, if that's ok. It should land shortly too.

PR #2912, opened 2021-09-03
feat: remove start_time from all workload types [DET-5979]

stoksc:
## Description
This change removes start time for steps, validations and checkpoints and, as a result, obsoletes the hack code to derive start time, since it is no longer reported by the harness. Additionally, step ID was dropped and replaced by a plain old SERIAL since it is also no longer reported by the harness.

Originally after the harness stopped reporting start time, it was derived as a hack in a painful way (but accurately, for the time being) in an attempt to not break the reporting done by the historical resource allocation project, but this has been worked around permanently by aggregating over `public.allocations` instead. Step ID was never really needed, but was kept "just in case" until it was clear after push architecture it was not needed at all.

## Test Plan
- [x] test migrations thoroughly.
- [x] test experiments that write metrics, make sure they are ok
- [x] test weird old stuff that would use step ID potentially (mainly old CLI commands)

## Commentary (optional)
This change should be a no-op after the changes to the aggregation queries land.

## Checklist


stoksc:
This is blocked on #2910 and #2911 but can be reviewed separately.

stoksc:
@shiyuann not sure what is up with react lint (it's passing locally) but everything else should be g2g. i'll figure it out before I land, may just need @hkang1 's PR that goes with this, idk.

shiyuann:
Isn't this user-facing change?

stoksc:
@shiyuann yeah it is, I've run it by @vishnu2kmohan (and I think several others) over the course of push arch. This is just one of those things that was going to happen as a result of push arch (even when we had that long discussion forever ago about pushing metrics vs workloads, and whether to push start time, once we decided against it, I did the "hack" to derive start time for the time being but i think i promised you on pr i would eventually rewrite historical allocation and just remove the hack and start time once and for all).

shiyuann (approved):
Looks good

shiyuann:
I mean you need to write release note for it @stoksc

hkang1 (approved):
Ah yeah sorry I missed these `Start Times`. LGTM!

PR #2914, opened 2021-09-03
chore: fix master/agent Docker image vulnerabilities [DET-5926]

stoksc:
## Description
This change installs `unattended-upgrades` and runs it once in the master and agent Docker images to ensure base OS packages are up to date.

## Test Plan
- [x] build images for master/agent
- [x] check no extra packages are pulled in (image size same as previous releases)
- [x] check affected packages are using fixed patch version listed in CVE.

## Commentary (optional)
Replaced upgrade of perl-base with this since this upgrades it.

## Checklist

### DIFF
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -2,7 +2,11 @@ FROM debian:10.3-slim
 MAINTAINER Determined AI <hello@determined.ai>
 
 RUN apt-get update -yq \
-    && DEBIAN_FRONTEND=noninteractive apt-get dist-upgrade -yq --no-install-recommends \
+    && apt-get upgrade -s \
+        | grep "^Inst" \
+        | grep -e "Debian" -e "Debian-Security" \
+        | awk -F " " {'print $2'} \
+        | DEBIAN_FRONTEND=noninteractive xargs apt-get upgrade -yq \
     && rm -rf /var/lib/apt/lists/*
 
 # For non-CUDA based images, these variables must be explicitly set for the Nvidia runtime to expose
--- a/agent/Dockerfile
+++ b/agent/Dockerfile
@@ -2,9 +2,7 @@ FROM debian:10.3-slim
 MAINTAINER Determined AI <hello@determined.ai>
 
 RUN apt-get update -yq \
-    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-        unattended-upgrades \
-    && unattended-upgrades \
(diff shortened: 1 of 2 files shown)


stoksc:
@ioga took another pass not using `unattended-upgrades`. It's 228MB (original) to 252MB (this) to 282MB (unattended), just on the size of `unattended` (confirmed this installs the same stuff, but it still installs packages we don't have). I'll take one last hack to make it better.

agent/Dockerfile
```diff
@@ -1,7 +1,14 @@
 FROM debian:10.3-slim
```
mackrorysd, line 1: Huh - I wonder why we're not using Ubuntu here? Nothing against Debian, but we're using Ubuntu pretty much everywhere else and it seems to be the best-supported by NVIDIA. I suspect they have more resources to be keeping Ubuntu base images, etc. up to date too. I wonder if we should switch.

stoksc: I was suggesting earlier we could at least use `debian:10-slim`, since it's up to date already.

stoksc: But @ioga mentioned some potential troubles with the same approach in envs.

ioga: we had the same issues with nvidia base images not having security upgrades couple months before. newer tags also didn't matter/help for nvidia. they just didn't update them on time.

stoksc: yep, confirmed latest release of `debian:10-slim` doesn't have the lz4 patch, so we'd need to do this anyway.

stoksc: ah but 10.10 does, I guess 10 != 10.10, which I assumed. oops.

master/Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM debian:10.3-slim
 MAINTAINER Determined AI <hello@determined.ai>
 
 RUN apt-get update -yq \
-    && apt-get install -y ca-certificates perl-base \
+    && apt-get upgrade -s \
+        | grep "^Inst" \
+        | grep -e "Debian" -e "Debian-Security" \
+        | awk -F " " {'print $2'} \
+        | DEBIAN_FRONTEND=noninteractive xargs apt-get upgrade -yq \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
```
stoksc, line 10:
```suggestion
    && DEBIAN_FRONTEND=noninteractive apt-get install -yq \
```

ioga (approved):

stoksc:
@mackrorysd i'll land this for now in the event we get busy, but i'll also look into just switching us to ubuntu in a bit.

PR #2922, opened 2021-09-07
ci: unpin pip version, improve py venv cache key.

ioga:
## Description

- Continuing #2920, I've found that the problem wasn't caused by setuptools 58 itself, but a conflict between it and the old circleci cache.
- To fix that, we bump the "major version" of the cache and add `pip freeze --all` to it, because it includes the versions of pip, wheel, and setuptools themselves (in contrast with the plain `pip freeze`).
- `pip<20` restriction comes from a while ago, and it seems we can modernize it.
- `setuptools<58` is no longer necessary.

## Test Plan

CI is green, and also runs fresh pip & setuptools.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,7 +266,7 @@ commands:
             - "/tmp/venv"
       - run: python --version
       - run: pip --version
-      - run: pip freeze
+      - run: pip freeze --all
       # Allow this to fail, but it is useful for debugging.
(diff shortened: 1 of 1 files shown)


dzhu (approved):
Thanks for figuring this out!

Not critical, but, while we're here, could you also add `--all` to the existing instances of `pip freeze` in the CircleCI config? Might've saved a few seconds in looking into this.

PR #2923, opened 2021-09-07
chore: add sso provider detection from master info and enable sso ui [DET-5937]

hkang1:
## Description

Update OSS to detect if `ssoProviders` property is set on `/api/v1/master` and enable SAML sign in (via Okta).

Previously this code sat in our enterprise only version but we decided to bring all these features and delineations into the OSS repo. The goal is to have the cluster the WebUI is connected to, to determine what capabilities the cluster can support and turning these features on and off based on availability via the API. The basic idea is the following:

1. WebUI loads a generic form
2. WebUI fetches GET `/api/v1/master` to see what type of cluster setup it is talking to
3. Based on the previous API call, the WebUI turns on/off features the cluster is capable of handling.

In our case, we are simply turning on Okta single-sign-on if the `ssoProviders` field is returned from `/api/v1/master` and includes `okta` as one of the sso providers.

## Checklist

### DIFF
--- a/proto/src/determined/api/v1/master.proto
+++ b/proto/src/determined/api/v1/master.proto
@@ -10,6 +10,17 @@ import "protoc-gen-swagger/options/annotations.proto";
 import "determined/log/v1/log.proto";
 import "determined/master/v1/master.proto";
 
+// Describe one SSO provider.
+message SSOProvider {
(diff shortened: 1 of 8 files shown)


hkang1:
wrong DET number

PR #2927, opened 2021-09-08
fix: kubernetes link with agent user [DET-5907]

eecsliu:
## Description
Attempting to run as a non-root user has previously caused kubernetes to fail, erroring out during the mounting volumes step. This fix addresses the issue by allowing the fluent bit sidecar to run as a root user. The main determined container will still run as non-root. 

## Test Plan
launch a kubernetes cluster and use `det user link-with-agent-user ...` to assign a uid and gid to the determined user. Submit an experiment and monitor. The experiment should complete without issue. 

## Checklist

### DIFF
--- a/master/internal/kubernetes/spec.go
+++ b/master/internal/kubernetes/spec.go
@@ -418,7 +418,6 @@ func (p *pod) createPodSpec(ctx *actor.Context, scheduler string) error {
 			Command:         fluentArgs,
 			Image:           "fluent/fluent-bit:1.6",
 			ImagePullPolicy: configureImagePullPolicy(spec.Environment),
-			SecurityContext: configureSecurityContext(spec.AgentUserGroup),
 			VolumeMounts:    loggingMounts,
 			WorkingDir:      fluentBaseDir,
 		})


mackrorysd (approved):
Let's definitely run this by @vishnu2kmohan (saw he was already on the review). This does add reasonable value (user-submitted code runs as that user, not root) in a case where things aren't working at all, so I'm +1'ing the patch anyway. I don't think we should do anything that requires distributing our own fluent-bit image, so if it just plain requires root, I think that's acceptable. But if you haven't already, I think we should do a little digging and see if we can still give fluent-bit what it needs without giving it a blank check. In this example it's running as a daemonset not a sidecar container, but it does hint that fluent-bit can run as non-root in at least some scenarios, so I'd like to make sure we explore the options here: https://github.com/fluent/helm-charts/blob/be60fec0ddd360c95d21778a44de4e45b475b200/charts/fluent-bit/values.yaml. I can do some digging tomorrow.

This keeps user-submitted code running as the user, and running with lower permissions. I suspect we might be able to

PR #2928, opened 2021-09-08
fix: make uPlot axis expand to show new data when not zoomed in [DET-5941]

EmilyBonar:
## Checklist

### DIFF
--- a/webui/react/src/components/UPlotChart.tsx
+++ b/webui/react/src/components/UPlotChart.tsx
@@ -5,6 +5,7 @@ import uPlot, { AlignedData } from 'uplot';
 import Message, { MessageType } from 'components/Message';
 import useResize from 'hooks/useResize';
 import { RecordKey } from 'types';
+import { distance } from 'utils/chart';
 
 export interface Options extends Omit<uPlot.Options, 'width'> {
   width?: number;
@@ -30,7 +31,7 @@ const UPlotChart: React.FC<Props> = ({ data, focusIndex, options }: Props) => {
   const scalesRef = useRef<Record<RecordKey, uPlot.Scale>>();
   const scalesZoomData = useRef<Record<string, ScaleZoomData>>({});
   const isZoomed = useRef<boolean>(false);
-  const mousePosition = useRef<[number, number]>([ 0, 0 ]);
+  const mousePosition = useRef<[number, number]>();
 
   const [ hasData, normalizedData ] = useMemo(() => {
     const chartData: unknown[][] = data || [];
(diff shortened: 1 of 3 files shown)


webui/react/src/components/UPlotChart.tsx
```diff
@@ -55,7 +57,33 @@ const UPlotChart: React.FC<Props> = ({ data, focusIndex, options }: Props) => {
...
+                handler(e);
+              };
+            },
+            mouseup: (_uPlot: uPlot, _target: EventTarget, handler: (e: Event) => void) => {
+              return (e: MouseEvent) => {
+                const distX = e.clientX-mousePosition.current[0];
+                const distY = e.clientY-mousePosition.current[1];
+                if (Math.sqrt(distX**2 + distY**2) > 5) {
+                  isZoomed.current = true;
+                }
```
hkang1, line 80: **chore:** under `src/utils/chart.ts` there is a `distance` function that we can use here.

webui/react/src/components/UPlotChart.tsx
```diff
@@ -55,7 +57,33 @@ const UPlotChart: React.FC<Props> = ({ data, focusIndex, options }: Props) => {
...
+                handler(e);
```
hkang1, line 81: **suggestion:** above this line, on a mouseup, I think it makes sense to clear out `mousePosition.current`

webui/react/src/components/UPlotChart.tsx
```diff
@@ -29,6 +29,8 @@ const UPlotChart: React.FC<Props> = ({ data, focusIndex, options }: Props) => {
   const chartDivRef = useRef<HTMLDivElement>(null);
   const scalesRef = useRef<Record<RecordKey, uPlot.Scale>>();
   const scalesZoomData = useRef<Record<string, ScaleZoomData>>({});
+  const isZoomed = useRef<boolean>(false);
+  const mousePosition = useRef<[number, number]>([ 0, 0 ]);
```
hkang1, line 33: **suggestion:** `mousePosition` should be able to be `undefined` by default. A default value of `[ 0, 0 ]` could potentially cause problems because it assumes the mouse always starts at that location.
```
const mousePosition = useRef<[ number, number ]>();
```

hkang1 (commented):
This does a great job of detecting zoom! Some minor suggestions to make it more robust.

One issue that was previously reported is when the user changes the series (which metrics to show), the current setup will still preserve the zoom when series changes. So if the user adds more metrics to display, and assuming they previously zoomed in, the adding of new metrics will not change the zoomed view. I'm thinking upon a metric change, we set the `isZoomed.current` back to false.

The question is how do we do that... One idea is to read the chart `options.series` and extract all the `label` from each series and if that changes, then reset the `isZoomed.current` back to `false`...

hkang1 (approved):
LGTM! nice work w/ the plugins!

PR #2929, opened 2021-09-08
chore: add 1.17 golang build syntax

hamidzr:
## Description

> //go:build lines
> The go command now understands //go:build lines and prefers them over // +build lines. The new syntax uses boolean expressions, just like Go, and should be less error-prone. As of this release, the new syntax is fully supported, and all Go files should be updated to have both forms with the same meaning. To aid in migration, gofmt now automatically synchronizes the two forms. For more details on the syntax and migration plan, see https://golang.org/design/draft-gobuild.

## Commentary (optional)

I haven't dug through go1.17 changelog but with it I get this set of changes every time I build. Would it be safe to commit these?

https://golang.org/doc/go1.17

## Checklist

### DIFF
--- a/agent/internal/fluent_intg_test.go
+++ b/agent/internal/fluent_intg_test.go
@@ -1,4 +1,5 @@
-//+build integration
+//go:build integration
+// +build integration
 
 package internal
 
--- a/agent/tools.go
+++ b/agent/tools.go
@@ -1,3 +1,4 @@
+//go:build tools
 // +build tools
 
 package tools
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -2,9 +2,10 @@ package command
 
 import (
 	"fmt"
-	"github.com/determined-ai/determined/master/pkg/ptrs"
 	"net/url"
 	"time"
+
+	"github.com/determined-ai/determined/master/pkg/ptrs"
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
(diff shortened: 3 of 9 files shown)


agent/internal/fluent_intg_test.go
```diff
@@ -1,4 +1,5 @@
-//+build integration
+//go:build integration
+// +build integration
```
stoksc, line 2: I was under the impression `go:build` superseded the `+build` syntax, mainly because the DSL for `+build` was error prone. In that case, don't we not need both?

hamidzr: hmm they mention
> , and all Go files should be updated to have both forms with the same meaning. 

and go fmt does this. maybe we keep them untill most people are on 1.17?

stoksc: if we keep both, people will get to delay upgrading their compiler, so i almost prefer dropping the old one.. but yeah, we can keep both, for the odd case if a user is building determined on their own.

stoksc (approved):

stoksc:
I will bump us to go 1.17 shortly unless you want to, to go with this.

agent/internal/fluent_intg_test.go
hamidzr: IMO it makes sense to go with their recommendation here at least until go 1.17 is a bit more common.

PR #2930, opened 2021-09-08
chore: job queue api placeholder

hamidzr:
## Description

- add external APIs and types use for the project

## Commentary (optional)

There are some comments I marked with TODO that are specifically open to discussion

## Checklist

- [x] include job summaries for other job types
- [x] add swagger annotations (required, optional) => leave off until system changes are done. only the clients would be affected by this

### DIFF
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -678,7 +678,7 @@ service Determined {
   // Get job queue stats for a resource pool.
   rpc GetJobQueueStats(GetQueueStatsRequest) returns (GetQueueStatsResponse) {
     option (google.api.http) = {
-      get: "/api/v1/jobs/stats/{resource_pool}"
+      get: "/api/v1/jobs/stats/{resource_pools}"
     };
     option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
       tags: "Jobs"
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -664,7 +664,6 @@ service Determined {
 
   /* Jobs */
 
-  // TODO jobs that are actively waiting to get scheduled.
   // Get a list of jobs in queue.
   rpc GetJobs(GetJobsRequest) returns (GetJobsResponse) {
     option (google.api.http) = {
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -676,7 +676,8 @@ service Determined {
   }
 
   // Get job queue stats for a resource pool.
-  rpc GetJobQueueStats(GetQueueStatsRequest) returns (GetQueueStatsResponse) {
+  rpc GetJobQueueStats(GetJobQueueStatsRequest)
+      returns (GetJobQueueStatsResponse) {
     option (google.api.http) = {
       get: "/api/v1/jobs/stats/{resource_pools}"
     };
@@ -686,7 +687,7 @@ service Determined {
   }
 
   // Control the job queue.
-  rpc UpdateJobQueue(UpdateQueueRequest) returns (UpdateQueueResponse) {
+  rpc UpdateJobQueue(UpdateJobQueueRequest) returns (UpdateJobQueueResponse) {
     option (google.api.http) = {
       post: "/api/v1/jobs/queue-ctl"
     };
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -11,6 +11,7 @@ import "determined/api/v1/auth.proto";
 import "determined/api/v1/checkpoint.proto";
 import "determined/api/v1/command.proto";
 import "determined/api/v1/experiment.proto";
+import "determined/api/v1/job.proto";
 import "determined/api/v1/master.proto";
 import "determined/api/v1/model.proto";
 import "determined/api/v1/notebook.proto";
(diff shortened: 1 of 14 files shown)


eecsliu (approved):
Looks mostly good to me!

proto/src/determined/resourcepool/v1/resourcepool.proto
```diff
@@ -328,4 +334,7 @@ message ResourcePoolPrioritySchedulerDetail {
   // The priority that is assigned to tasks that do not explicitly specify a
   // priority.
   int32 default_priority = 2;
+
+  // List of available priorities for K8 (if applicable).
+  repeated K8PriorityClass k8_priorities = 3;
```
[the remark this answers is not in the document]
hamidzr: for posterity we chatted about this. let keep this in to remind us that we need this.  if we realise we can create priority classes on the fly then we won't need to expose this.

proto/src/determined/job/v1/job.proto
```diff
@@ -0,0 +1,79 @@
...
+  // The username of the user who submitted the job.
+  string user          = 4;
+  string resource_pool = 5;
+  bool is_preemptible  = 6;
+  int32 priority       = 7;
+  int32 weight         = 8;
+}
+
+// Describes a message to control jobs in a queue.
+message QueueControl {
```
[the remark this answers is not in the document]
hamidzr: let me know if you come up with someting you like better. we can always change later.

proto/src/determined/api/v1/api.proto
```diff
@@ -661,6 +662,39 @@ service Determined {
     };
   }
 
+  /* Jobs */
+
+  // Get a list of jobs in queue.
+  rpc GetJobs(GetJobsRequest) returns (GetJobsResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/jobs"
```
stoksc, line 670:
```suggestion
      get: "/api/v1/resource-pools/{pool_id}/queue"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -661,6 +662,39 @@ service Determined {
...
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "Jobs"
+    };
+  }
+
+  // Get job queue stats for a resource pool.
+  rpc GetJobQueueStats(GetJobQueueStatsRequest)
+      returns (GetJobQueueStatsResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/jobs/stats/{resource_pools}"
```
stoksc, line 681:
```suggestion
      get: "/api/v1/resource-pools/{pool_id}/queue/stats"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -661,6 +662,39 @@ service Determined {
...
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "Jobs"
+    };
+  }
+
+  // Control the job queue.
+  rpc UpdateJobQueue(UpdateJobQueueRequest) returns (UpdateJobQueueResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/jobs/queue-ctl"
```
stoksc, line 691:
```suggestion
      post: "/api/v1/resource-pools/{pool_id}/queue_action"
```

stoksc (commented):
this is my take on these APIs. i may be contradicting myself here, i can't remember, sorry it i am.

after thinking on it some more though, the "job queue" isn't the list of all jobs, it's a view of all jobs that reside on a particular pool in order of precedence. I think we should reserve `/api/v1/jobs` for "the list of all jobs, past and present", since this may exist one day, and put these endpoints under resource pools since the view is owned by them. i scheduled some time for tomorrow so that we can get on the same page; whatever we decide, I want to make sure my similar-ish changes are on the same page.

hamidzr:
> this is my take on these APIs. i may be contradicting myself here, i can't remember, sorry if i am.
> 
> after thinking on it some more though, the "job queue" isn't the list of all jobs, it's a view of all jobs that reside on a particular pool in order of precedence. I think we should reserve `/api/v1/jobs` for "the list of all jobs, past and present", since this may exist one day, and put these endpoints under resource pools since the view is owned by them. i scheduled some time for tomorrow so that we can get on the same page; whatever we decide, I want to make sure my similar-ish changes are on the same page.


Having these APIs under resource pools makes more sense to me if I'm just looking at the current state and the project's scope and I think both can work out [[ref1](https://github.com/determined-ai/determined/pull/2930/commits/1f2883a399806e6cafbce367643875be09c0dd6d#diff-6b9604657643382d7fa162b046021c03f6487c7975b67233e6219e2ec9eced5cL43), [ref2](https://www.dropbox.com/s/0mv0z761xsx089r/apis.md?dl=0)]. For what's best for a potential future state and which would be most consistent with our other projects let's chat over slack. 
Either way, the changes are limited to the routes we expose and implementation wise the only part they'd change is how  resource_pool parameter is passed on and read thus not really affecting the progress of the project.

stoksc:
> Either way, the changes are limited to the routes we expose and implementation wise the only part they'd change is how resource_pool parameter is passed on and read thus not really affecting the progress of the project.

I agree they shouldn't block the progress, but we may as well commit what we think it right now rather than putting "fixing it" on the backburner for some other time. It seems from the channel as long as you're ok there is consensus on moving the routes.

hamidzr:
> > Either way, the changes are limited to the routes we expose and implementation wise the only part they'd change is how resource_pool parameter is passed on and read thus not really affecting the progress of the project.
> 
> I agree they shouldn't block the progress, but we may as well commit what we think it right now rather than putting "fixing it" on the backburner for some other time. It seems from the channel as long as you're ok there is consensus on moving the routes.

👍🏼. I'd want to move some of these definitions around when I update the routes. since it's going into a feature branch I'll land this to keep the history and the review separate.

PR #2933, opened 2021-09-09
fix: scroll trial ids with values in trial comparison [DET-5918]

EmilyBonar:
## Description
Fixes header so that the trial ids scroll with values, while remaining sticky to the top of the table. Also makes the row labels sticky.
trial comp

## Checklist


hkang1 (commented):
Nice this is much better with the trial header following the content!

Reporting some visual oddities I encountered:
The dropdown is behind `headerRow` (I think)
Screen Shot 2021-09-09 at 10 19 16 PM
2021-09-09 at 10 19 PM
2021-09-09 at 10 21 PM
2021-09-09 at 10 23 PM

hkang1 (approved):
LGTM, added a some style tweaks. Nice work with the sticky headers (both x and y axis directions!)

PR #2936, opened 2021-09-09
feat: det deploy: prefer /etc/determined to /usr/local/determined

rb-determined-ai:
## Description

Remove the final vestiges of the old `/usr/local/determined` directory.  Use the FHS-compliant `/etc/determined/` instead.

## Test Plan

I'll manually some cloud deployments with and without ssl.

## Commentary (optional)

By my count the following filepaths were affected:

```
/usr/local/determined/etc/master.yaml          -> /etc/determined/master.yaml
/usr/local/determined/etc/master.yaml.context  -> /etc/determined/master.yaml.context
/usr/local/determined/etc/master.yaml.tmpl     -> /etc/determined/master.yaml.tmpl
/usr/local/determined/etc/db_ssl_root_cert.pem -> /etc/determined/db_ssl_root_cert.pem
/usr/local/determined/{etc,}/master.crt        -> /etc/determined/master.crt
/usr/local/determined/{etc,}/master.key        -> /etc/determined/master.key
/usr/local/determined/container_startup_script -> /etc/determined/container_startup_script
/usr/local/determined/startup_script           -> /etc/determined/startup_script
```

To me, those names seem decent, though I considered that maybe `startup_script` and `container_startup_script` could use some clarifying as to what they are starting up.

Are there release party deployment scripts I should edit too?


dzhu:
This is great! I'll update the release party scripts.

harness/determined/deploy/local/mount.yaml
```diff
@@ -3,4 +3,4 @@ version: "3.7"
 services:
   determined-master:
     volumes:
-      -  ${DET_MASTER_CONFIG:-/usr/local/determined/etc/master.yaml}:/etc/determined/master.yaml
+      -  ${DET_MASTER_CONFIG:-/etc/master.yaml}:/etc/determined/master.yaml
```
ioga, line 6:
```suggestion
      -  ${DET_MASTER_CONFIG}:/etc/determined/master.yaml
```

first I was going to suggest that this variable should default to `/etc/determined/master.yaml`, but looking at the code, it seems this mount.yaml is not used unless DET_MASTER_CONFIG is set to something real. so maybe this default is not a thing at all?

ioga (approved):

ioga:
I understand nothing, but to me this also warrants a "User-facing API change" label because we've guided some users (OSS people using `det deploy aws` who want to slightly edit the config before #2766) to edit the config at `/usr/local/determined/`. Now it's in the new place! It's breaking change for them.

rb-determined-ai:
It's not an API change per-se, but we are aiming to land it on a minor version bump since it's definitely a breaking change of sorts.  Label added.

harness/determined/deploy/local/mount.yaml
rb-determined-ai: hey, that's a good point.  That makes one less breaking change than I was thinking.

What do you think about automatically mounting `/etc/master.yaml` when it exists?  That seems like reasonable behavior to me...

rb-determined-ai:
@ioga what about `/etc/determined/agent_vm_startup_script` and `/etc/determined/agent_container_startup_script`?   Instead of `/etc/determined/{,container_}startup_script`?

harness/determined/deploy/local/mount.yaml
ioga: it doesn't really sound appealing to me.

1. this behavior didn't exist before, so using `/etc/determined/master.yaml` as a path probably makes more sense.
2. "native" determined master from a linux package running on the host will use the same config file.
3. which means if you run native and then try `det deploy local`, you'll suddenly conflate their configs and run into expected side effects.

ioga:
> @ioga what about `/etc/determined/agent_vm_startup_script` and `/etc/determined/agent_container_startup_script`? Instead of `/etc/determined/{,container_}startup_script`?

I believe we use `container_startup_script` as a key in master config for aws/gcp providers. So to stay in alignment, we'd need to rename it and `startup_script` to `agent_vm_startup_script` as well, and probably do that in a backwards compatible way.

But then you were also pushing to run master & agent natively instead of containers in aws/gcp... which effectively kills `container_startup_script` for aws/gcp. And this'll be breaking - I doubt you can simply start running former `container_startup_script` on the host vm after normal `startup_script`.

rb-determined-ai:
> But then you were also pushing to run master & agent natively instead of containers in aws/gcp

well, that's a long ways out.  I have a pile of pre-0.17.0 changes to make and that's not even close to being on the table.

> which effectively kills container_startup_script for aws/gcp

I'm thinking though that the people who would be affected by altering the container_startup_script
Do you think it is a good idea to put _this_ breaking change off until _that_breaking change?

ioga:
> Do you think it is a good idea to put _this_ breaking change off until _that_breaking change?

Both the renaming and native master-agent are both breaking changes that don't have a customer asking for them. I'd batch them into a project called "Dynamic Agents UX improvements v2" and do them all at once. Then we don't make an external impression that we break the system little by little every month, and internally we have an impression that department is "done".

PR #2940, opened 2021-09-10
feat: change the default deployment type for det deploy aws

shiyuann:
## Checklist


rb-determined-ai:
@shiyuann what testing did you do on this change?

shiyuann:
@rb-determined-ai no testing is needed because it's only changing the default value of a flag? Testing `det deploy aws --deployment-type efs` is already covered in other PRs.

rb-determined-ai:
Testing is **absolutely** needed.  What happens when a user with a 0.16.5 cluster of the default type tries to upgrade their cluster to the 0.17.0 version of determined?  We'll silently change the cloudformation template that we use, right?  What does that do to their cluster?

rb-determined-ai:
When I deployed a `simple` cluster, used it a bit, and then redeployed as an `efs` cluster, I completely lost my database.  That would be catastrophic if we started doing that to users.

shiyuann:
@rb-determined-ai what do you mean by re-deploying? I don't think you can update an existing deployment?

rb-determined-ai:
Yeah, I thought you just did another `det deploy aws up` with all the args you had before, plus any modifications you want to make.

@hoang-hpe, is that accurate?  It's definitely possible to upgrade the determined version of a det deploy cluster, right?

hoang-hpe:
> Yeah, I thought you just did another `det deploy aws up` with all the args you had before, plus any modifications you want to make.
> 
> @hoang-hpe, is that accurate? It's definitely possible to upgrade the determined version of a det deploy cluster, right?

Yes you should be able to update an existing cluster by running the same `det deploy aws up` command with a new `det-version` or with an updated `determined` package.

rb-determined-ai:
So what is the plan here?  The safe strategies that I can think of are:
1. change the default behavior from `simple` to: detect when the cluster already exists, and use `simple` if it does or use `efs` if it doesn't
2. update the simple template itself to include an EFS (or, equivalently, introduce a new template called `simple-efs` or something and make that the default, leaving the `simple` template alone)

My guess is that 1 is the better option for us, but I'm too familiar with this cloudformation technology.

shiyuann:
@rb-determined-ai I tested it. The DB won't lose the data.

shiyuann:
> So what is the plan here? The safe strategies that I can think of are:
> 1. change the default behavior from simple to: detect when the cluster already exists, and use simple if it does or use efs if it doesn't 
> 2. update the simple template itself to include an EFS (or, equivalently, introduce a new template called simple-efs or something and make that the default, leaving the simple template alone)
> My guess is that 1 is the better option for us, but I'm too familiar with this cloudformation technology.

1 is not feasible. You can't really detect what type a stack is. You may analyze the cloudformation yaml it uses but it is definitely not a good approach.
2 is bad too. Because essentially you are removing the simple deployment and creating a duplicate efs template.

Changing the default deployment type would be just fine. The cloudformation will figure out connecting the right DB itself.

rb-determined-ai:
I just tested it again, and got the same result.  Here are my repro steps:

```
det deploy aws up --det-version 0.17.0rc0 --cluster-id rb-delete-me --deployment-type simple --keypair rb-aws
export DET_MASTER=...

# run an experiment
det e create const.yaml -f .

# verify experiment is there
det e list -a

# switch deploytment type
det deploy aws up --det-version 0.17.0rc0 --cluster-id rb-delete-me --deployment-type efs --keypair rb-aws
export DET_MASTER=...

# verify experiment is missing
det e list -a

# verify a new experiment is created with experiment_id=1
det e create const.yaml -f .
```

Same result as before.  We can't ship this change like this.

PR #2942, opened 2021-09-10
fix: update logic on when query url should be overridden

hkang1:
## Description

The logic of when a query url should be overridden with default settings needs to be corrected. Currently it replaces if the current url query params don't match the settings state and current settings exists. The correct approach is to check to see if any of the settings are set in the query within the keys specified within a config and if nothing is set, THEN override it with the current state.

## Test Plan

- [ ] load multi-trial experiment and navigate to `Trials` tab. Select multiple trials and hit reload page. Upon a refresh, the previously selected trials should remain.

## Checklist


hamidzr:
> - load multi-trial experiment and navigate to Trials tab. Select multiple trials and hit reload page. Upon a refresh, the previously selected trials should remain.￼

just confirming that this is the case in master as well. Could you describe a test case for this PR? would that be something we could add as a unit test

hamidzr (approved):
:shipit:

hkang1:
> > * load multi-trial experiment and navigate to Trials tab. Select multiple trials and hit reload page. Upon a refresh, the previously selected trials should remain.￼
> 
> confirming that this passes(works) in master as well as on the preview built on this branch.. Could you describe a test case for this PR? would that be something we could add as a unit test

Great idea. Added a test case for reading url query params and verifying the read in settings value.

PR #2955, opened 2021-09-14
chore: add documentation for model-hub mmdetection [DET-5924]

liamcli:
## Description
Add documentation for model-hub mmdetection.  

Also fixed pytorch and torchvision versions that were out-of-date in circleci config so that they now match the versions for our default gpu image.  

Release note 2792 covers this PR as well.

## Test Plan
Made sure e2e tests passed in feature branch.  

## Checklist


model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -81,7 +91,6 @@ def __init__(self, context: det_torch.PyTorchTrialContext) -> None:
             mmcv.runner.build_optimizer(self.model, self.cfg.optimizer)
         )
         self.model.zero_grad()
-        self.optimizer.zero_grad()
```
shiyuann, line 84: Is it ok to just remove this?

model_hub/examples/mmdetection/fasterrcnn.yaml
```diff
@@ -15,35 +15,35 @@ data:
...
   max_length:
-    batches: 87850
+    batches: 87960
   smaller_is_better: false
-max_restarts: 0
+max_restarts: 5
 environment:
   image:
-    gpu: determinedai/model-hub-mmdetection
+    gpu: determinedai/model-hub-mmdetection:0.17.0.dev0
```
shiyuann, line 44: Should use commit id.

shiyuann (commented):
I don't have too much context on what needs to be supported in the model hub. The change in the documentation and CI looks good.

model_hub/examples/mmdetection/fasterrcnn.yaml
liamcli: we push tip of master to the dev tag every commit.  on release, the dev suffix is removed.  the assumption is that users will checkout a release tag or just download the example from our doc.

model_hub/model_hub/mmdetection/_trial.py
liamcli: optimizer parameters are a subset of model parameters so model.zero_grad() will cover everything.

model_hub/examples/mmdetection/fasterrcnn.yaml
shiyuann: > on release, the dev suffix is removed.

This won't be done automatically. Is the plan to remove the tag?

liamcli: This is done automatically as part of bumpenvs run during release process. It's one reason we need the dev version here.

shiyuann (approved):

PR #2956, opened 2021-09-15
feat: allow kubernetes to use priority from exp config

eecsliu:
## Description
Previously we relied on static priority classes that we create with helm install. With this change, determined can now dynamically create and destroy priority classes, allowing the user to specify a priority number in the experiment config. Now only two priority classes will be created during installation: a determined system priority for high priority tasks like master and gc, and a medium priority to serve as a default priority level.

NOTE: so far, this has only been tested with 0.16.5 becuase I've had troubles getting 0.17 to build. The mechanisms for priorityclass creation and deletion shouldn't change though.

## Test Plan
spin up a kubernetes cluster, launch an experiment with the priority field populated. (`resources: priority: x`) Check the list of priorityclasses and make sure that the priority class named `<container-id>-priorityclass` has been created. At the end of the experiment check that the priority class has been deleted.

Repeat with an hp search experiment and make sure that it can handle multiple trials.

Repeat again but this time, specify a priority class in the podSpec. It should use the priority class defined in the podSpec instead of generating a new one.  Similarly, not specifying a priority will result in it using the `determined-medium-priority` class. 

## Checklist

### DIFF
--- a/docs/topic-guides/system-concepts/scheduling.txt
+++ b/docs/topic-guides/system-concepts/scheduling.txt
@@ -190,18 +190,25 @@ and Determined can only automatically set labels for GPU experiments.
 Determined provides a default priority class, ``determined-medium-priority`` that has a priority of
 ``50`` and is used for all tasks. If users want to set a different priority level for an experiment,
 they may either specify a priority in the ``resources`` field of the experiment config or create a
-priorityclass and specify it in the ``pod_spec`` of the config, as shown below. Also, if using a
-cluster with tainted nodes or labels, users must specify the tolerations or node selectors in the
-``pod_spec``:
+priorityClass and specify it in the ``pod_spec`` of the config. If both are specified, the specified
+priorityClass will take precedence over the priority field.
+
+Additionally, if using a cluster with tainted nodes or labels, users must specify the tolerations or
+node selectors in the ``pod_spec``. We recommend using both tolerations and node selectors to better
+constrain where your experiments can run, especially on clusters that contain multiple GPU types.
+
+Below is an example that illustrates how to set priorities, tolerations, and node selectors.
 
 .. code:: yaml
 
+   resources:
+      priority: 42 # priorityClass, if set, takes precedence over this value
    environment:
       pod_spec:
          apiVersion: v1
          kind: Pod
          spec:
-            priorityClassName: determined-medium-priority
+            priorityClassName: determined-medium-priority # don't set if using priority value
             nodeSelector:
                key: value
             tolerations:
@@ -209,6 +216,3 @@ cluster with tainted nodes or labels, users must specify the tolerations or node
                operator: "Equal"
                value: "value"
                effect: "NoSchedule"
-
-We recommend using both tolerations and node selectors to better constrain where your experiments
-can run, especially on clusters that contain multiple GPU types.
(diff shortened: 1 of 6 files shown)


eecsliu:
Standby - I forgot to add some docs changes and the helm version bump

docs/topic-guides/system-concepts/scheduling.txt
```diff
@@ -187,8 +187,12 @@ This plugin is also in beta and is not enabled by default. Similar to the cosche
 enabled by setting the ``defaultScheduler`` field to ``preemption``. Autoscaling is not supported
 and Determined can only automatically set labels for GPU experiments.
 
-If using a cluster with tainted nodes or labels, users must specify the tolerations or node
-selectors in the experiment config:
+Determined provides a default priority class, ``determined-medium-priority`` that has a priority of
```
mackrorysd, line 190: Can we keep the discussion of priority separate from the discussion of taints / tolerations? Very distinct topics - easy to miss the other important detail if you know you don't care about one.

helm/charts/determined/templates/priority-classes.yaml
```diff
@@ -7,26 +7,26 @@ preemptionPolicy: Never
...
 ---
+{{- if .Values.defaultScheduler}}
+{{- $schedulerType := .Values.defaultScheduler | trim}}
+{{- if or (eq $schedulerType "coscheduler") (eq $schedulerType "preemption")}}
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: determined-medium-priority
-value: 50
+  name: determined-high-priority
```
mackrorysd, line 25: I don't understand why determined-high-priority is still here, if not for backwards compatibility, nor why it's now inside the conditional but medium isn't. Could you enlighten me?

master/internal/kubernetes/pods.go
```diff
@@ -400,6 +400,9 @@ func (p *pods) cleanUpPodHandler(ctx *actor.Context, podHandler *actor.Ref) erro
 		return errors.Errorf("unknown pod handler being deleted %s", podHandler.Address())
 	}
 
+	name := fmt.Sprintf("%s-priorityclass", podInfo.containerID)
+	_ = p.clientSet.SchedulingV1().PriorityClasses().Delete(name, &metaV1.DeleteOptions{})
```
mackrorysd, line 404: Should at least log an error here in case of failure, even if it shouldn't happen in theory.

helm/charts/determined/templates/priority-classes.yaml
eecsliu: Good point, we actually don't need `determined-high-priority` or `determined-low-priority` anymore. I can just straight up remove them from the helm chart.

master/internal/kubernetes/pods.go
eecsliu: Will add a log statement. 

There is the case where if a user doesn't specify a priority level, we'll automatically user `determined-medium-priority`. If this happens, there is nothing to delete and would return an "error". I can tailor the log statement accordingly.

mackrorysd: Oh, hmm... if it's not feasible to pass the information of whether or not we used the default option through to this point, I'd just log it at debug level so that the information exists but is ignored in the routine case.

mackrorysd (approved):
Looks good to me, but let's add that log line either at debug level (or, if we can tell if there was supposed to be something to delete), warn.

hamidzr (approved):

PR #2957, opened 2021-09-15
fix: implement boto3 wrapper to allow refreshable credentials [DET-5690]

azhou-determined:
## Description

A user bug was reported in which an experiment crashed agent due to S3 credentials being expired. boto3 does not automatically refresh shared credentials that not managed by IAM (in ~/.aws/credentials) so the solution is to implement a wrapper around boto3 that makes use of their `RefreshableCredentials` object which will fetch every time it expires. 

Expiration time has been set at 60 min for now, since that is how long it takes for AWS tokens to expire. boto3 will attempt to refresh 15 min before tokens are set to expire, but these values can be configured. 

This fix was tested by spinning up a local master, configuring an experiment to store on S3 (`checkpoint_storage: type: s3 bucket: bucket-name`), mounting a local .aws/credentials file to the agent/experiment container, and verifying that the training loop fetches the new credentials within expiration time. 

## Checklist


harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,93 @@
...
+    METHOD = "managed-refresh-cred"
+
+    def __init__(self, credential_provider: CredentialProvider, refresh_after: int = 60) -> None:
+        super().__init__()
+        self.refresh_after = timedelta(minutes=refresh_after)
+        self.credential_provider = credential_provider
+
+    def load(self) -> Optional[RefreshableCredentials]:
+        credentials = self._credentials()
+        return credentials and RefreshableCredentials(
```
rb-determined-ai, line 61: I've never seen `and` used like this before, that's kinda cool.

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+    def _credentials(self) -> Credentials:
+        return self.credential_provider.load()
+
+    def _expiry_time(self) -> Any:
+        return _local_now() + self.refresh_after
+
+
+def initialize_boto3() -> None:
```
rb-determined-ai, line 91: Can we rename this to describe what it's actually doing?

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,93 @@
...
+
+class Boto3Manager:
+    """
+    Wrapper class around boto3 that allows registering of custom credential providers
+    and provides methods for setting the default boto3 session or fetching a new session
+    with the new context
+    """
+
+    def __init__(self) -> None:
+        self.session = get_session()
```
rb-determined-ai, line 25: What do you think of using a different strategy, where instead of setting the default session to one we've configured, can we just reconfigure the existing default session? That way if two hacks like this ever have to coexist, we've done our part to work with the other one.

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,93 @@
...
+        return credentials and RefreshableCredentials(
+            method=self.METHOD,
+            refresh_using=self._refresh,
+            access_key=credentials.access_key,
+            secret_key=credentials.secret_key,
+            token=credentials.token,
+            expiry_time=self._expiry_time(),
+        )
+
+    def _refresh(self) -> Dict:
```
rb-determined-ai, line 70: I think what you've done here looks a lot like how RefreshableCredentials is used by the other parts of `botocore.credentials`.  But I think that it's not the right behavior for this situation.  I don't really think there should be a delay between when the files are rewritten and when we start using the new files.  The next time you try to read from the Credentials, I think the Credentials should refresh themselves from the file.  In that case you'd want to subclass Credentials with a different behavior from RefreshableCredentials.

Well... you could simulate the same behavior with a very short expiry time but where you use the timeout to check if the `SharedCredentialProvider.creds_filename` has been updated on disk, rather than always reloading every expiry time.  In practice that would probably be equivalent because you'd likely want a short backoff time with the first strategy anyway, rather than checking disk for every single request.  Ok, so I don't think you actually need to subclass Credentials.

But can we set the expiry time in such a way that we check for changes every minute or every 15 seconds or something?  Not reload the file every 15 seconds, but check if it's been modified on disk and reload if it has.  The user is presumably doing something like every 55 minutes, and if we have a timeout of 45 min, eventually a situation will arise where we check 5 minutes before they change it, which will leave us using an expired session for 40 minutes (assuming what they set expires 60 minutes after they set it, so old credentials are only valid for 5 minutes after they update to new ones).

rb-determined-ai (commented):
boto3 is very confusing.  I had to read a lot of code to review this.

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            self._check_time = self._next_check_time()
+            if self._refresh_needed():
+                logging.info("credential file changes detected, refreshing credentials")
+                credentials = self._refresh_using()
+                self.access_key = credentials.access_key
+                self.secret_key = credentials.secret_key
+                self.token = credentials.token
+
+    def _next_check_time(self) -> Any:
+        return _local_now() + self._check_every
```
rb-determined-ai, line 100: This is a bit of a style preference thing, not a blocker or anything, but I think this helper method is too small to be helpful.

If I read:
```python
self._check_time = self._next_check_time()
```
I have to go look at `self._next_check_time()` to know what it does.

If I read:
```python
self._check_time = time.time() + self._backoff_time
```
I know immediately what it does.  And `time.time() + self._backoff_time` is so simple and easy I don't feel like it needs to have a helper function for the two times you would use it.

Helper functions aren't zero-cost; when used excessively they make simple code which could be understood linearly no longer possible to understand linearly (you have to jump around reading helper functions instead).

Again, this is style thing more than a hard blocker.  This is just my take.

rb-determined-ai (commented):
Overall its looking good

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,135 @@
...
+
+    METHOD = "managed-refresh-cred"
+
+    def __init__(self, credential_provider: SharedCredentialProvider, check_every: int = 2) -> None:
+        super().__init__()
+        self.check_every = check_every
+        self.credential_provider = credential_provider
+
+    def load(self) -> Credentials:
+        return RefreshableSharedCredentials(
```
rb-determined-ai, line 34: I think you still need the `credentials and RefreshableSharedCredentials()` bit here, because returning None from this function is how the CredentialResolver knows to move on and call `load()` on the next CredentialProvider.

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,135 @@
...
+        path = os.path.expandvars(path)
+        path = os.path.expanduser(path)
+        return path
+
+    def _credentials_modified_time(self) -> float:
+        credentials_file = self._credentials_file()
+        return os.stat(credentials_file).st_mtime
+
+    def _refresh_needed(self) -> bool:
+        return self._credentials_modified_time() > self._last_loaded
```
rb-determined-ai, line 72: this is a really slight nit, but maybe return true if `self._credentials_modified_time` is different at all?  Probably would never matter I guess.

rb-determined-ai (commented):
looks good, but there's a couple questions I asked that you haven't responded to yet

harness/determined/common/storage/boto3_credential_manager.py
```diff
@@ -0,0 +1,148 @@
...
+    managed_credential_provider = RefreshedCredentialProvider(
+        credential_provider=credential_provider
+    )
+    credential_resolver.insert_before(
+        name=provider_name, credential_provider=managed_credential_provider
+    )
+
+
+def initialize_boto3_credential_providers() -> None:
+    session = get_session()
```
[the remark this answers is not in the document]
rb-determined-ai: Yeah, I think you make a good point.  I'm ok with landing it as-is and seeing if we ever hit an issue.

rb-determined-ai (approved):
looks great!

PR #2961, opened 2021-09-16
chore: update job queue route paths

hamidzr:
## Checklist

### DIFF
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -668,7 +668,7 @@ service Determined {
   // Get a list of jobs in queue.
   rpc GetJobs(GetJobsRequest) returns (GetJobsResponse) {
     option (google.api.http) = {
-      get: "/api/v1/resource-pools/{pool_id}/queue"
+      get: "/api/v1/resource-pools/queue"
     };
     option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
       tags: "Jobs"
@@ -680,8 +680,7 @@ service Determined {
       returns (GetJobQueueStatsResponse) {
     option (google.api.http) = {
       // FIXME could be queue-stats
-      // TODO how do we get stats for multiple resource pools? we have view that needs it
-      get: "/api/v1/resource-pools/{pool_id}/queue/stats"
+      get: "/api/v1/resource-pools/queue/stats"
     };
     option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
       tags: "Jobs"
@@ -691,7 +690,7 @@ service Determined {
   // Control the job queue.
   rpc UpdateJobQueue(UpdateJobQueueRequest) returns (UpdateJobQueueResponse) {
     option (google.api.http) = {
-      post: "/api/v1/resource-pools/{pool_id}/queue-ctl"
+      post: "/api/v1/resource-pools/queue/queue-ctl"
     };
     option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
       tags: "Jobs"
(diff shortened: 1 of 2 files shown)


proto/src/determined/api/v1/api.proto
```diff
@@ -678,7 +679,9 @@ service Determined {
   rpc GetJobQueueStats(GetJobQueueStatsRequest)
       returns (GetJobQueueStatsResponse) {
     option (google.api.http) = {
-      get: "/api/v1/jobs/stats/{resource_pools}"
+      // FIXME could be queue-stats
+      // TODO how do we get stats for multiple resource pools? we have view that needs it
```
hamidzr, line 683: @stoksc suggestions? ideally, we'd like to avoid setting up the clients to open on request per RP (UI design in figma in the ERD)

could do /resource-pools/stats but that'd conflict with {pool_id} unless we nest that under queue. 

sorta similar issue, but less important, for batch queue control actions across resourcepools

stoksc: I think `/api/v1/resource-pools/queues/stats?pool_id=default_cpu&pool_id=default_gpu` is a fine scheme (or comma separated `pool_ids=...` since some client take params as a map and dont properly handle multiples).

hamidzr: that works we'd need to update the get part so it doesn't conflict with 
 `get: "/api/v1/resource-pools/{pool_id}/queue"`

> that'd conflict with {pool_id} unless we nest that under queue

maybe `/api/v1/resource-pools/queues/` with pool_id as query parameter?

stoksc: I did ponder `/api/v1/resource-pools/default,aux,gpu/queue/states` but idk, that's pretty non-standard.

proto/src/determined/api/v1/api.proto
```diff
@@ -688,7 +689,7 @@ service Determined {
   // Control the job queue.
   rpc UpdateJobQueue(UpdateJobQueueRequest) returns (UpdateJobQueueResponse) {
     option (google.api.http) = {
-      post: "/api/v1/jobs/queue-ctl"
+      post: "/api/v1/resource-pools/queue/queue-ctl"
```
hamidzr, line 692: 🤔 could be a post on /rp/queue but it's not exactly creating a queue. 
maybe a patch would fit?  `patch /rp/queue`

proto/src/determined/api/v1/api.proto
hamidzr: I updated the routes.

proto/src/determined/api/v1/api.proto
hamidzr: just a patch against the queue is looking better to me. thoughts?

stoksc: i had thought just posting a queue_action? patching the queue makes sense to me too (probably more than my dumb post idea)

hamidzr: let's go with a patch now

proto/src/determined/api/v1/api.proto
```diff
@@ -662,12 +662,12 @@ service Determined {
...
   }
 
-  /* Jobs */
+  /* Jobs Queue */
 
   // Get a list of jobs in queue.
   rpc GetJobs(GetJobsRequest) returns (GetJobsResponse) {
     option (google.api.http) = {
-      get: "/api/v1/jobs"
+      get: "/api/v1/resource-pools/queue"
```
stoksc, line 670: Since this can return multiple queues, right?
```suggestion
      get: "/api/v1/resource-pools/queues"
```

stoksc (approved):
lgtm, nit is optional

PR #2966, opened 2021-09-16
chore: bump SSH Keys to RSA 2048-bit

varlogtim:


vishnu2kmohan:
I seem to recall that there was a good reason why we chose 1024 bit, and that it had to do with a drastic slowdown in key generation, preventing workers in large distributed training jobs from rendezvousing with each other over SSH before the Horovod timeout expired? cc @rb-determined-ai

vishnu2kmohan:
[Yep!](https://github.com/determined-ai/determined/commit/3d3ca30057dfb57e64e536dee14c09bec14a8e7d)

> When we have an experiment with thousands of trials, we generate an ssh
key for each trial. This becomes very expensive, since ssh keygen is a
cpu intensive operation. By reducing the keysize, the performance
generating the keys moves from ~1.4 s/op to ~0.02 s/op, which is enough
solve the present impact. Security concerns were taken into
consideration but the worry is negated by the key only being used for
a short duration - usually much less than a day. The time to crack
an 1024-bit RSA key is much longer given even extraordinary
circumstances.

varlogtim:
> When we have an experiment with thousands of trials, we generate an ssh
key for each trial.
Is there a reason why we generate unique keys for each trial?
If we generated the key per experiment and passed the keys to the trial this would be constant time regardless of the number of trials.

stoksc:
Technically (now) with determined agents this generation is deferred until we are allocated, but on k8s it will happen immediately on pod submission so k8s restarts could lock up the master like ssh keygen used to.

varlogtim:
Ohh... https://github.com/determined-ai/determined/pull/144

PR #2969, opened 2021-09-17
docs: Add a checkbox to the PR template

bensomers:
## Description
Extremely tiny change adding a single checkbox to the PR template.

## Test Plan
N/A

## Commentary (optional)
Jointly decided per internal process review meeting

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,7 +26,7 @@ You may intentionally leave this section blank and remove the title.
 
 
 ## Checklist
-
+- [ ] Changes have been manually QA'd
 - [ ] User-facing API changes need the "User-facing API Change" label.
 - [ ] Release notes should be added as a separate file under `docs/release-notes/`.
 See [Release Note](../docs/release-notes/README.md) for details.


.github/pull_request_template.md
```diff
@@ -26,7 +26,7 @@ You may intentionally leave this section blank and remove the title.
 
 
 ## Checklist
-
+- [ ] Changes have been manually QA'd
```
dzhu, line 29:
```suggestion
- [ ] Changes have been manually QAed.
```
(Or QAd? :shrug: Not sure there's a dominant rule, but anyway I subscribe to the school of keeping apostrophes for possessives only. Or you could sidestep the issue with _gone through manual QA_.)

dzhu (approved):

loksonarius (approved):

.github/pull_request_template.md
dannysauer: > there's
...
> keeping apostrophes for possessives only

I dunno; seems like contractions are OK too. ;)  Besides, there doesn't seem to be great agreement upon whether acronyms are regular or irregular (sleep/slept) verbs, so the original feels like a defensible move here. After looking around for a while, it seems that style guides generally approve the use of `'d` after an acronym or initialism ("QA" is an initialism) when it is broadly recognized by the target audience as a verb.  I think it's fair to say that the target audience does recognize `QA`'s use as a verb in this context.

Here's a tweet from the AP Stylebook on "OK", which is short for okay, but does use the apostrophe for the all-caps shortened form.
https://twitter.com/APStylebook/status/1550534545329782785?s=20&t=3TicYX2AhiAMTczmy43ebQ

The Chicago Manual of Style suggests "OD'd" for "overdosed", and [Merriam Webster](https://www.merriam-webster.com/dictionary/od) shows OD'ing, OD's for plural, and either OD'd or ODed (though the apostrophe form is preferred there as well) for past tense.

Based on that set of references, I'm approving and merging this before it hits a year old next month.  Maybe a new PR if someone really prefers `Changes have undergone manual QA`.  The checkboxes really should all be presented in the same tense, just like a bulleted list, so perhaps a new PR could make that change at the same time.

PR #2973, opened 2021-09-17
feat: jq - set up job position and job id

hamidzr:
## Description

- add scheduler interface for reporting ordered allocations
- introduce job id and persist it
- attach job info onto allocationRequests
- collect and present metadata needed for the job queue

## Test Plan

- added go unit test
- examine the queue `./prepqueue.sh && curl-da.sh "${DET_MASTER}/api/v1/resource-pools/queues?resource_pools=default" -s | jq`

## Commentary (optional)

- [x] intro the min needed from job id (attached to allocationRequest)
- [x] persist job id
- [x] add a new interface to Scheduler to return ordered allocations.
  - [x] implement this for priority scheduler
- [x] sql: transaction foreign key constraint
- [x] sql: sync migration.down
- [x] jobid db: existing entities without job id: only experiments; use eid w/o persisting
- [x] tests
- [x] would agent labels affect the job order?
- [x] experiment/trial snapshot restore
- [x] are alloc req enough to represent jobs? case: an unterminated job with no alloc request?
- [x] fix up lint complaints

## Checklist

### DIFF
--- a/agent/go.sum
+++ b/agent/go.sum
@@ -463,6 +463,9 @@ github.com/googleapis/gax-go/v2 v2.0.5/go.mod h1:DWXyrwAJ9X0FpwwEdw+IPEYBICEFu5m
 github.com/googleapis/gnostic v0.0.0-20170729233727-0c5108395e2d h1:7XGaL1e6bYS1yIonGp9761ExpPPV1ui0SAC59Yube9k=
 github.com/googleapis/gnostic v0.0.0-20170729233727-0c5108395e2d/go.mod h1:sJBsCZ4ayReDTBIg8b9dl28c5xFWyhBTVRp3pOg5EKY=
 github.com/gookit/color v1.2.4/go.mod h1:AhIE+pS6D4Ql0SQWbBeXPHw7gY0/sjHoA4s/n1KB7xg=
+github.com/goombaio/orderedmap v0.0.0-20180924084748-ba921b7e2419/go.mod h1:YKu81H3RSd1cFh0d7NhvUoTtUC9IY/vBX0WUQb1/o4Y=
+github.com/goombaio/orderedset v0.0.0-20180925151225-8e67b20a9b77 h1:4dvq1tGHn1Y9KSRY0OZ24Khki4+4U+ZrA//YYsdUlJU=
+github.com/goombaio/orderedset v0.0.0-20180925151225-8e67b20a9b77/go.mod h1:HPelMYpOyy0XvglpBbmZ3krZpwaHmszj/vQNlnETPTM=
 github.com/gophercloud/gophercloud v0.1.0 h1:P/nh25+rzXouhytV2pUHBb65fnds26Ghl8/391+sT5o=
 github.com/gophercloud/gophercloud v0.1.0/go.mod h1:vxM41WHh5uqHVBMZHzuwNOHh8XEoIEcSTewFxm1c5g8=
 github.com/gopherjs/gopherjs v0.0.0-20181017120253-0766667cb4d1 h1:EGx4pi6eqNxGaHF6qqu48+N2wcFQ5qg5FXgOdqsJ5d8=
--- a/master/internal/api_job.go
+++ b/master/internal/api_job.go
@@ -48,9 +48,10 @@ func (a *apiServer) GetJobs(
 		resp.Jobs = append(resp.Jobs, &jobv1.Job{
 			Summary: &jobv1.JobSummary{
 				JobId: string(job.JobID),
+				// State: // look at AllocationState
 			},
 			EntityId: job.EntityID,
-			Type:     jobv1.Type_TYPE_COMMAND,
+			Type:     job.JobType.Proto(),
 			// Type:     "TYPE_" + string(job.JobType), // TODO
 		})
 	}
--- a/master/internal/api_job.go
+++ b/master/internal/api_job.go
@@ -44,15 +44,14 @@ func (a *apiServer) GetJobs(
 	if err != nil {
 		return nil, err
 	}
-	for index, val := range jobs.Values() {
+	for _, val := range jobs.Values() {
 		summary, ok := val.(resourcemanagers.JobSummary)
 		if !ok {
 			continue // FIXME or error?
 		}
 		resp.Jobs = append(resp.Jobs, &jobv1.Job{
 			Summary: &jobv1.JobSummary{
-				JobId:         fmt.Sprint(summary.JobID),
-				QueuePosition: int32(index),
+				JobId: fmt.Sprint(summary.JobID),
 			},
 			EntityId: summary.EntityID,
 			// Type: "TYPE_"+summary.JobType, // TODO
--- a/master/internal/api_job.go
+++ b/master/internal/api_job.go
@@ -2,12 +2,9 @@ package internal
 
 import (
 	"context"
-	"fmt"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-
-	"github.com/goombaio/orderedset"
 
 	"github.com/determined-ai/determined/master/internal/resourcemanagers"
 	"github.com/determined-ai/determined/master/internal/sproto"
@@ -25,7 +22,7 @@ func (a *apiServer) GetJobs(
 	// a.sort(resp.Jobs, req.OrderBy, req.SortBy, apiv1.GetJobsRequest_SORT_BY_QUEUE_POSITION)
 	// resp, a.paginate(&resp.Pagination, &resp.Jobs, req.Pagination.Offset, req.Pagination.Limit)
 
-	var jobs orderedset.OrderedSet
+	var jobs []*sproto.JobSummary
 	resp = &apiv1.GetJobsResponse{}
 
 	// TODO loop over all resource pools in the request
@@ -44,17 +41,17 @@ func (a *apiServer) GetJobs(
 	if err != nil {
 		return nil, err
 	}
-	for _, val := range jobs.Values() {
-		summary, ok := val.(resourcemanagers.JobSummary)
-		if !ok {
-			continue // FIXME or error?
+	for _, job := range jobs {
+		if job == nil {
+			panic("received an empty job summary") // this shouldn't happen
 		}
 		resp.Jobs = append(resp.Jobs, &jobv1.Job{
 			Summary: &jobv1.JobSummary{
-				JobId: fmt.Sprint(summary.JobID),
+				JobId: string(job.JobID),
 			},
-			EntityId: summary.EntityID,
-			// Type: "TYPE_"+summary.JobType, // TODO
+			EntityId: job.EntityID,
+			Type:     jobv1.Type_TYPE_COMMAND,
+			// Type:     "TYPE_" + string(job.JobType), // TODO
 		})
 	}
 	return resp, nil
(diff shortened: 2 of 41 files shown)


agent/go.sum
```diff
@@ -463,6 +463,9 @@ github.com/googleapis/gax-go/v2 v2.0.5/go.mod h1:DWXyrwAJ9X0FpwwEdw+IPEYBICEFu5m
 github.com/googleapis/gnostic v0.0.0-20170729233727-0c5108395e2d h1:7XGaL1e6bYS1yIonGp9761ExpPPV1ui0SAC59Yube9k=
 github.com/googleapis/gnostic v0.0.0-20170729233727-0c5108395e2d/go.mod h1:sJBsCZ4ayReDTBIg8b9dl28c5xFWyhBTVRp3pOg5EKY=
 github.com/gookit/color v1.2.4/go.mod h1:AhIE+pS6D4Ql0SQWbBeXPHw7gY0/sjHoA4s/n1KB7xg=
+github.com/goombaio/orderedmap v0.0.0-20180924084748-ba921b7e2419/go.mod h1:YKu81H3RSd1cFh0d7NhvUoTtUC9IY/vBX0WUQb1/o4Y=
```
hamidzr, line 466: extra?

master/internal/api_job.go
```diff
@@ -15,9 +21,33 @@ var notImplementedError = status.Error(codes.Unimplemented, "API not implemented
 func (a *apiServer) GetJobs(
 	_ context.Context, req *apiv1.GetJobsRequest,
 ) (resp *apiv1.GetJobsResponse, err error) {
+	// TODO pagination and filtering
 	// a.sort(resp.Jobs, req.OrderBy, req.SortBy, apiv1.GetJobsRequest_SORT_BY_QUEUE_POSITION)
 	// resp, a.paginate(&resp.Pagination, &resp.Jobs, req.Pagination.Offset, req.Pagination.Limit)
-	return nil, notImplementedError
+
```
hamidzr, line 27: not the focus of this pr

hamidzr:
```
(determined) hrz ➜ determined git:(jq-qposition) gd $(git merge-base jq-qposition upstream/job-queue) . | grep -E '(todo|fixme|warn|check) '
```
we could chat about the remaining points below when you got some time to review. and would probably be better left for another pr.
- [x]   // TODO pagination and filtering
- [x]       // TODO loop over all resource pools in the request
- [x]                       // TODO why do we persist the task when the underlying command isn't persisted? or is it now persisted
- [x]// FIXME haven't decided if resource manager actor should be responsible for this or not
- [x]func getV1Jobs( // TODO rename
- [x]       // TODO remove me. for testing only.
- [x]// TODO remove me
- [ ]       // WARN scheduled here means that resources are allocated.
- [x]       // FIXME there is gotta be a friendlier version of this.
- [x]       // TODO remove logs
- [x]       // TODO consider agent labels
- [x]               // TODO snapshot restore
- [x]// TODO here or in model/job.go
- [x]       return string(id) // FIXME does this add any value? is this a common interface in go?
- [x]BEGIN; -- FIXME costly? migrations cover this case.
- [x]-- FIXME or we limit these to non-terminal experiments only?
- [ ]-- TODO do we need to persist task association? no..
```

File: master/internal/db/postgres_jobs.go [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/jmoiron/sqlx"
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// addJob persists the existence of a task from a tx.
+func addJob(tx *sqlx.Tx, t *model.Job) error {
```
eecsliu, line 11: I think we have to delete jobs from this table that are completed or deleted. Will this be added later?

master/internal/resourcemanagers/fair_share.go
```diff
@@ -51,6 +51,12 @@ func (f *fairShare) Schedule(rp *ResourcePool) ([]*sproto.AllocateRequest, []*ac
 	return fairshareSchedule(rp.taskList, rp.groups, rp.agents, rp.fittingMethod)
 }
 
+func (f *fairShare) OrderedAllocations(
+	rp *ResourcePool,
+) (reqs []*sproto.AllocateRequest) {
+	return reqs
```
eecsliu, line 57: This is WIP right? Else I'm flagging it as incorrect :)

master/internal/resourcemanagers/job.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/determined-ai/determined/master/internal/sproto"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/proto/pkg/jobv1"
+)
+
+// FIXME haven't decided if resource manager actor should be responsible for this or not
+// we don't want a separate actor do we? could be useful for streaming job endpoints
+type GetJobOrder struct{}
+
+func filterAllocateRequests(reqs AllocReqs) AllocReqs {
```
eecsliu, line 15: clarification: does this just filter allocation requests for if their jobs exist?

master/internal/resourcemanagers/job.go
```diff
@@ -0,0 +1,92 @@
...
+func allocReqsToJobSummaries(reqs AllocReqs) (summaries []*sproto.JobSummary) {
+	for _, req := range filterAllocateRequests(reqs) {
+		summaries = append(summaries, req.Job)
+	}
+	return summaries
+}
+
+// allocReqsToJobOrder convertes sorted allocation requests to job order.
+func allocReqsToJobOrder(reqs []*sproto.AllocateRequest) (jobIds []string) {
+	for _, req := range filterAllocateRequests(reqs) {
```
eecsliu, line 40: Is `filterAllocateRequests` supposed to be sorted?

master/internal/resourcemanagers/job.go
```diff
@@ -0,0 +1,92 @@
...
+	job = &jobv1.Job{
+		Summary: &jobv1.JobSummary{
+			JobId: string(req.Job.JobID),
+			State: req.Job.State.Proto(), // look at AllocationState
+		},
+		EntityId:       req.Job.EntityID,
+		Type:           req.Job.JobType.Proto(),
+		IsPreemptible:  req.Preemptible,
+		ResourcePool:   req.ResourcePool,
+		User:           "hamid", // TODO
```
eecsliu, line 63: nice

prepqueue.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+det e create $PROJECT_ROOT/e2e_tests/tests/fixtures/no_op/adaptive.yaml $PROJECT_ROOT/e2e_tests/tests/fixtures/no_op > /dev/null &
```
eecsliu, line 3: This is interesting...why do we have to do this to prep for queue?

eecsliu (commented):
Have some questions, but I don't have major objections

master/internal/api_job.go
```diff
@@ -15,9 +18,44 @@ var notImplementedError = status.Error(codes.Unimplemented, "API not implemented
...
+	resp = &apiv1.GetJobsResponse{}
+
+	// TODO loop over all resource pools in the request
+	if len(req.ResourcePools) < 1 {
+		return nil, status.Error(codes.InvalidArgument, "missing resource_pools parameter") // FIXME
+	}
+
+	switch {
+	case sproto.UseAgentRM(a.m.system):
+		err = a.actorRequest(sproto.AgentRMAddr.Child(req.ResourcePools[0]), resourcemanagers.GetJobOrder{}, &jobs)
```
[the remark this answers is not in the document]
hamidzr: the final API exposure isn't the point of the pr and I just included the first one for a demo and to keep the pr scope smaller.

master/internal/command/shell_manager.go
```diff
@@ -35,7 +35,7 @@ func (s *shellManager) Receive(ctx *actor.Context) error {
 
 	case tasks.GenericCommandSpec:
 		taskID := model.TaskID(uuid.New().String())
```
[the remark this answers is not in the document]
hamidzr: nope, I just updated what I ran into. `NewTaskID` was already there and not added by this pr but it's a good idea to update the references. 👍🏼

master/internal/db/postgres_jobs.go
hamidzr: we had a [discussion](https://determined-community.slack.com/archives/C0269E3SMJP/p1630365373008500?thread_ts=1630362854.004500&cid=C0269E3SMJP) about this in #queue-management the plan is to leave and never delete for future use.

master/internal/resourcemanagers/fair_share.go
hamidzr: wdy mean, are you saying I couldn't have solved this in one line of code? : ) 
joking aside it's just like this because of scope (out of scope)

master/internal/resourcemanagers/job.go
hamidzr: i should add a comment to the code
1. filters allocations that are not associated with a job
2. if a job has many allocReqs it would only keep the one that's most representative of the final job state. [more reading] with the current state of the schedulers we can achieve this without having to introduce a new struct eg doing additional computation for coming up with the desired `schedulingstate` w/o sacrificing anything.

master/internal/resourcemanagers/job.go
hamidzr: yeah. that func expects a sorted list of allocations, (and outputs a sorted one) but has no way of checking it. maybe I should clarify this somewhere.

hamidzr: any suggestions?

master/internal/resourcemanagers/job.go
hamidzr: haven't decided on a way to grab this yet. suggestions? I'm planning on putting it as out of scope for the PR so we can land and use the rest of the work.

prepqueue.sh
hamidzr: ahh it's just a convenience for populating the queue when developing. will go away before landing in master.

master/internal/db/postgres_jobs.go
eecsliu: ah I didn't pay attention to that. I'll need to think about this a bit more - at first I thought it wasn't a good idea, but I'm now thinking it's not that bad. Esp. if we only refer to active jobs in the first place.

master/internal/resourcemanagers/job.go
eecsliu: hmmm maybe I'm not reading this correctly, but it doesn't look to me like we can assume that the returned req list is sorted

master/internal/command/shell_manager.go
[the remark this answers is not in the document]
hamidzr: done. `fd -t f .go master | xargs sed 's/TaskID(uuid.New().String())/NewTaskID()/' -i` to the rescue.

master/internal/db/postgres_jobs.go
eecsliu: conclusion: probably not very bad like I initially thought

hamidzr: let's continue in the channel if needed

eecsliu (approved):
👌👌👌

PR #2976, opened 2021-09-20
fix: make PIDServer send SIGKILL after waiting on SIGTERM

rb-determined-ai:
In other words, try even harder to avoid hangs.

## Test Plan

I had a bug in some other half-written feature that was causing the `ZMQBroadcastServer` to hang in ways that it normally wouldn't, which resulted in hangs.  This fixed the hangs that resulted.

### DIFF
--- a/harness/determined/ipc.py
+++ b/harness/determined/ipc.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import selectors
 import signal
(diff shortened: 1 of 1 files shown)


harness/determined/ipc.py
```diff
@@ -682,12 +682,22 @@ def health_check() -> None:
                 # Let things finish logging, exiting on their own, etc.
                 time.sleep(grace_period)
                 p.send_signal(on_fail)
+                if on_fail != signal.SIGKILL:
+                    try:
+                        return p.wait(timeout=10)
+                    except subprocess.TimeoutExpired:
+                        p.send_signal(signal.SIGKILL)
```
ioga, line 689: will it be useful to log force-SIGKILL'ing because a) other software tends to do that b) it's informative of how bad things are?

ioga (approved):

harness/determined/ipc.py
rb-determined-ai: good idea

PR #2978, opened 2021-09-20
feat: Add Checkpoint Storage prefix for S3  [DET-5968]

varlogtim:
DET-5968

## Test Plan

Test cases updated.
Manually tested invalid prefix, i.e., starting and/or ending with /

## Checklist

- [?] Release notes should be added as a separate file under `docs/release-notes/`.

### DIFF
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -1,7 +1,8 @@
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Optional
 
 import pytest
+import yaml
 
 from tests import command as cmd
 from tests import config as conf
@@ -32,25 +33,29 @@ entrypoint: model_def:NoOpTrial
 """
 
 
-def s3_config(num_trials: int, secrets: Dict[str, str]) -> str:
-    return f"""
-description: noop_random
-checkpoint_storage:
-  type: s3
-  access_key: {secrets["INTEGRATIONS_S3_ACCESS_KEY"]}
-  secret_key: {secrets["INTEGRATIONS_S3_SECRET_KEY"]}
-  bucket: {secrets["INTEGRATIONS_S3_BUCKET"]}
-hyperparameters:
-  global_batch_size: 1
-searcher:
-  metric: validation_error
-  smaller_is_better: true
-  name: random
-  max_trials: {num_trials}
-  max_length:
-    batches: 100
-entrypoint: model_def:NoOpTrial
-"""
+def s3_config(num_trials: int, secrets: Dict[str, str], prefix: Optional[str] = None) -> str:
+    config_dict = {
+        "description": "noop_random",
+        "checkpoint_storage": {
+            "type": "s3",
+            "access_key": secrets["INTEGRATIONS_S3_ACCESS_KEY"],
+            "secret_key": secrets["INTEGRATIONS_S3_SECRET_KEY"],
+            "bucket": secrets["INTEGRATIONS_S3_BUCKET"],
+        },
+        "hyperparameters": {"global_batch_size": 1},
+        "searcher": {
+            "metric": "validation_error",
+            "smaller_is_better": True,
+            "name": "random",
+            "max_trials": num_trials,
+            "max_length": {"batches": 100},
+        },
+        "entrypoint": "model_def:NoOpTrial",
+    }
+    if prefix is not None:
+        config_dict["checkpoint_storage"]["prefix"] = prefix  # type: ignore
+
+    return str(yaml.dump(config_dict))
 
 
 @pytest.mark.slow
@@ -81,13 +86,16 @@ def test_start_tensorboard_for_shared_fs_experiment(tmp_path: Path) -> None:
 @pytest.mark.slow
 @pytest.mark.e2e_gpu
 @pytest.mark.tensorflow2
-def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
+@pytest.mark.parametrize("prefix", [None, "my/test/prefix"])
+def test_start_tensorboard_for_s3_experiment(
+    tmp_path: Path, secrets: Dict[str, str], prefix: Optional[str]
+) -> None:
     """
     Start a random experiment configured with the s3 backend, start a
     TensorBoard instance pointed to the experiment, and kill the TensorBoard
     instance.
     """
-    with FileTree(tmp_path, {"config.yaml": s3_config(1, secrets)}) as tree:
+    with FileTree(tmp_path, {"config.yaml": s3_config(1, secrets, prefix)}) as tree:
         config_path = tree.joinpath("config.yaml")
         experiment_id = exp.run_basic_test(
             str(config_path), conf.fixtures_path("no_op"), num_trials
@@ -104,13 +112,13 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str, 
 
 @pytest.mark.slow
 @pytest.mark.e2e_cpu
+@pytest.mark.tensorflow2
 def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
     """
     Start 3 random experiments configured with the s3 and shared_fs backends,
     start a TensorBoard instance pointed to the experiments and some select
     trials, and kill the TensorBoard instance.
     """
-
     with FileTree(
         tmp_path,
         {
@@ -139,6 +147,7 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
         "start",
         str(shared_fs_exp_id),
         str(s3_exp_id),
+        str(multi_trial_exp_id),
         "-t",
         *trial_ids,
         "--no-browser",
@@ -148,7 +157,5 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
         for line in tensorboard.stdout:
             if SERVICE_READY in line:
                 break
-            if AWAITING_METRICS in line:
-                raise AssertionError("Tensorboard did not find metrics")
         else:
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
--- a/e2e_tests/tests/experiment/experiment.py
+++ b/e2e_tests/tests/experiment/experiment.py
@@ -677,13 +677,17 @@ def shared_fs_checkpoint_config() -> Dict[str, str]:
     }
 
 
-def s3_checkpoint_config(secrets: Dict[str, str]) -> Dict[str, str]:
-    return {
+def s3_checkpoint_config(secrets: Dict[str, str], prefix: Optional[str] = None) -> Dict[str, str]:
+    config_dict = {
         "type": "s3",
         "access_key": secrets["INTEGRATIONS_S3_ACCESS_KEY"],
         "secret_key": secrets["INTEGRATIONS_S3_SECRET_KEY"],
         "bucket": secrets["INTEGRATIONS_S3_BUCKET"],
     }
+    if prefix is not None:
+        config_dict["prefix"] = prefix
+
+    return config_dict
 
 
 def s3_checkpoint_config_no_creds() -> Dict[str, str]:
--- a/harness/determined/common/storage/s3.py
+++ b/harness/determined/common/storage/s3.py
@@ -1,6 +1,7 @@
 import contextlib
 import logging
 import os
+import re
 import tempfile
 from typing import Iterator, Optional
 
@@ -9,6 +10,16 @@ import requests
 from determined import errors
 from determined.common import util
 from determined.common.storage.base import StorageManager
+
+
+def normalize_prefix(prefix: Optional[str]) -> str:
+    new_prefix = ""
+    if prefix is not None and prefix != "":
+        banned_patterns = (r"^.*\/\.\.\/.*$", r"^\.\.\/.*", r".*\/\.\.$", r"^\.\.$")
+        if any([re.match(bp, prefix) for bp in banned_patterns]):
+            raise ValueError(f"prefix must not match: {' '.join(banned_patterns)}")
+        new_prefix = os.path.normpath(prefix).lstrip("/")
+    return new_prefix
 
 
 class S3StorageManager(StorageManager):
@@ -22,6 +33,7 @@ class S3StorageManager(StorageManager):
         access_key: Optional[str] = None,
         secret_key: Optional[str] = None,
         endpoint_url: Optional[str] = None,
+        prefix: Optional[str] = None,
         temp_dir: Optional[str] = None,
     ) -> None:
         super().__init__(temp_dir if temp_dir is not None else tempfile.gettempdir())
@@ -39,6 +51,8 @@ class S3StorageManager(StorageManager):
         )
         self.bucket = self.s3.Bucket(self.bucket_name)
 
+        self.prefix = normalize_prefix(prefix)
+
         # Detect if we are talking to minio, because boto3 has a client-side bug parsing the output
         # of the minio server.
         self._use_minio_workaround = False
@@ -54,6 +68,9 @@ class S3StorageManager(StorageManager):
                         "MinIO backend detected.  To work around a boto3 bug, empty directories"
                         "will not be uploaded in checkpoints."
                     )
+
+    def get_storage_prefix(self, storage_id: str) -> str:
+        return os.path.join(self.prefix, storage_id)
 
     def post_store_path(self, storage_id: str, storage_dir: str) -> None:
         """post_store_path uploads the checkpoint to s3 and deletes the original files."""
@@ -78,7 +95,7 @@ class S3StorageManager(StorageManager):
 
     @util.preserve_random_state
     def upload(self, storage_id: str, storage_dir: str) -> None:
-        storage_prefix = storage_id
+        storage_prefix = self.get_storage_prefix(storage_id)
         for rel_path in sorted(self._list_directory(storage_dir)):
             key_name = f"{storage_prefix}/{rel_path}"
             logging.debug(f"Uploading {rel_path} to s3://{self.bucket_name}/{key_name}")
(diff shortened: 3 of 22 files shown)


rb-determined-ai:
> Another thing that would be nice, which I tried to figure out but couldn't, would be if we could give a message to the user other than this doesn't match this regex. Sometimes it is easier to say, "Don't start or end with /" instead of having the user try to decipher a regex.

This does exist, there's a `checks` extension of the form:
```python
"checks": {
  "my custom message": {
    # some sub-schema that should pass, or we print the custom message
  }
}
```

Overall though, what if we just let users put the extension right into their bucket setting?  Like:
```yaml
checkpoint_storage:
  type: aws
  bucket: my-favorite-bucket/my/prefix
```

The way you did it is _exactly_ how I would have done it, but I noticed that @RehanSD added support for prefixes as part of the azure `container` setting, and my mind was blown.

varlogtim:
> checkpoint_storage:
  type: aws
  bucket: my-favorite-bucket/my/prefix

... but then this wouldn't be a `bucket`. I feel like the ability to append a key prefix to a `bucket` field wouldn't be self-evident from reading the interface and would require reading at the documentation in detail.

It would make more sense if the field was called `path` and there was the assumption that the first chunk of text inbetween the first two delimiters was the `bucket` and the rest was the `prefix`. Not saying we should do this, just thinking in text...

@rb-determined-ai I don't agree with adding this to the `bucket` field, but if that is what you want I will make the change.

@vishnu2kmohan Curious if you have an opinion?

RehanSD:
> @rb-determined-ai I don't agree with adding this to the `bucket` field, but if that is what you want I will make the change.

So in the case that prefixes are added, it's not going to be a container in azure either, but it allows you to store your checkpoints in a directory inside a container/bucket - rather than having separate containers/buckets for each of your experiments. I think it's useful in the case that users want to impose their own file structure eg have 1 bucket for determined checkpoints, and store this experiments checkpoints at a particular path.

rb-determined-ai:
Right, you get that feature benefit with either config option. The question is more, "which config option do we like better".

docs/reference/api/experiment-config.txt
```diff
@@ -229,6 +229,9 @@ store such as `MinIO <https://min.io/>`__.
 
 **Optional Fields**
 
+``prefix``
+   The optional path prefix to use. Must not start or end with /.
```
rb-determined-ai, line 233: Why can't we accept leading or trailing `/`?

docs/reference/cluster/helm-config.txt
```diff
@@ -95,24 +95,22 @@ should be configured by editing the ``values.yaml`` and ``Chart.yaml`` files in
...
       -  ``bucket``: The S3 bucket name to use.
-
       -  ``accessKey``: The AWS access key to use.
-
       -  ``secretKey``: The AWS secret key to use.
-
+      -  ``prefix``: The optional path prefix to use. Must not start or end with /.
       -  ``endpointUrl``: The optional endpoint to use for S3 clones, e.g., http://127.0.0.1:8080/.
 
-      -  ``type: azure``: Checkpoints are stored in Microsoft's Azure Blob Storage. Authentication
```
rb-determined-ai, line 105: nice catch on the indent here.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -32,14 +32,23 @@ def shared_fs_config(num_trials: int) -> str:
 """
 
 
-def s3_config(num_trials: int, secrets: Dict[str, str]) -> str:
-    return f"""
+def s3_config(num_trials: int, secrets: Dict[str, str], prefix: Union[str, None] = None) -> str:
```
rb-determined-ai, line 35: Please use `Optional[str]` instead of `Union[str, None]`.  They're equivalent in meaning but `Optional[str]` is the canonincal option.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -78,10 +89,9 @@ def test_start_tensorboard_for_shared_fs_experiment(tmp_path: Path) -> None:
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
 
 
-@pytest.mark.slow  # type: ignore
-@pytest.mark.e2e_gpu  # type: ignore
-@pytest.mark.tensorflow2  # type: ignore
-def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
+def start_tensorboard_for_s3_experiment(
+    tmp_path: Path, secrets: Dict[str, str], prefix: Union[str, None] = None
```
rb-determined-ai, line 93: ditto

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -102,6 +112,22 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str,
...
+    start_tensorboard_for_s3_experiment(tmp_path, secrets)
+
+
+@pytest.mark.slow  # type: ignore
+@pytest.mark.e2e_gpu  # type: ignore
+@pytest.mark.tensorflow2  # type: ignore
+def test_start_tensorboard_for_s3_prefix_experiment(
+    tmp_path: Path, secrets: Dict[str, str]
+) -> None:
+    start_tensorboard_for_s3_experiment(tmp_path, secrets, "my/test/prefix")
```
rb-determined-ai, line 128: you can keep the original function and use `@pytest.mark.parameterize(prefix, [None, "my/test/prefix])` do do this a little simpler.

e2e_tests/tests/experiment/experiment.py
```diff
@@ -686,6 +686,16 @@ def s3_checkpoint_config(secrets: Dict[str, str]) -> Dict[str, str]:
     }
 
 
+def s3_checkpoint_config_prefix(secrets: Dict[str, str]) -> Dict[str, str]:
```
rb-determined-ai, line 689: can you just add a `prefix=None` to `s3_checkpoint_config` instead?

e2e_tests/tests/test_system.py
```diff
@@ -92,6 +92,12 @@ def test_gc_checkpoints_s3(secrets: Dict[str, str]) -> None:
     run_gc_checkpoints_test(config)
 
 
+@pytest.mark.e2e_gpu  # type: ignore
+def test_gc_checkpoints_s3_prefix(secrets: Dict[str, str]) -> None:
```
rb-determined-ai, line 96: again, just parameterize the existing test.

harness/determined/common/storage/s3.py
```diff
@@ -77,7 +82,7 @@ def restore_path(self, storage_id: str) -> Iterator[str]:
 
     @util.preserve_random_state
     def upload(self, storage_id: str, storage_dir: str) -> None:
-        storage_prefix = storage_id
+        storage_prefix = f"{self.prefix}{storage_id}"
```
rb-determined-ai, line 85: ... and do `posixpath.join(self.prefix, storage_id)` here.

harness/tests/storage/test_s3.py
```diff
@@ -41,12 +41,16 @@ def live_manager(tmp_path: Path, require_secrets: bool) -> storage.S3StorageMana
 
 
 @pytest.mark.cloud  # type: ignore
-def test_live_s3_lifecycle(live_manager: storage.S3StorageManager, require_secrets: bool) -> None:
+@pytest.mark.parametrize("prefix", ["", "my/test/prefix/"])  # type: ignore
```
rb-determined-ai, line 44: maybe test `None` also

harness/tests/tensorboard/test_s3.py
```diff
@@ -11,6 +11,29 @@
 
 BASE_PATH = pathlib.Path(__file__).resolve().parent.joinpath("fixtures")
 
+test_configs = [
```
rb-determined-ai, line 14: rather than create 3x as many copies of the default_conf, can we just add `@pytest.mark.parameterize` a `prefix` parameter into the `test_s3_build()` test, do something like:
```python
conf = dict(**default_conf)
if prefix is not None:
    conf["prefix"] = prefix
```

schemas/expconf/v0/s3.json
```diff
@@ -42,6 +42,14 @@
             ],
             "default": null
         },
+        "prefix": {
```
rb-determined-ai, line 45: can you add the prefix field to the `s3 checkpoint storage (valid)` test case in `schemas/test_cases/v0/unions.yaml`?

schemas/expconf/v0/s3.json
```diff
@@ -42,6 +42,14 @@
...
+            "type": [
+                "string",
+                "null"
+            ],
+            "pattern": "^([^/].*[^/]|[^/])$",
```
rb-determined-ai, line 50: The only pattern I would bother enforcing is that `..` doesn't appear.  But even that, if the aws http server rejects it, might be unnecessary.

If you do choose to have a pattern here, please add testcases to ensure it's working.

rb-determined-ai (commented):
What's the plan for gcs?

rb-determined-ai:
> @rb-determined-ai I don't particularly like setting the prefix for the manager here. Let me know if you think it is OK, or if you'd like me to change it.

Can you point me to where in the code your are talking about?

harness/tests/storage/test_s3.py
```diff
@@ -41,12 +41,16 @@ def live_manager(tmp_path: Path, require_secrets: bool) -> storage.S3StorageMana
...
+    live_manager: storage.S3StorageManager, require_secrets: bool, prefix: str
+) -> None:
     def post_delete_cb(storage_id: str) -> None:
         """Search s3 directly to ensure that a checkpoint is actually deleted."""
         found = [obj.key for obj in live_manager.bucket.objects.filter(Prefix=storage_id)]
         if found:
             file_list = "    " + "\n    ".join(found)
             raise ValueError(f"found {len(found)} files in bucket after delete:\n{file_list}")
 
+    live_manager.prefix = prefix
```
varlogtim, line 55: @rb-determined-ai ^ this is the line that I don't like. 

This is ungood because I am setting the member var `prefix` from outside of the obj. There is manipulation done to the prefix in the S3StorageManager ctor, which is why it is difficult to test `None` being passed in. Perhaps, I should make it `__prefix`, i.e., private and have a set/getter. My reasoning is that, while we are validating the format of the `prefix` when we parse it from the config, is it possible that someone (or somecode) could use the `S3StorageManager` from outside of the context of an experiment (such as in PyTest)? ... and so, we need some way of validating the format of the prefix inside this class as well.

I am having trouble with using a fixture for the `S3StorageManager`.IMHO, we should be parametrizing the `S3StorageManager`'s creation. Perhaps I am unaware of some PyTest functionality which could help us solve this?

rb-determined-ai: Sure.  Replace the fixture with a function... I don't think the fixture really adds anything of value here.

e2e_tests/tests/command/test_tensorboard.py
varlogtim: done

e2e_tests/tests/test_system.py
varlogtim: done

schemas/expconf/v0/s3.json
varlogtim: Hrm... I guess we could allow a user to put anything here.

I guess we need to define what is supposed to happen given different inputs:
```
The following would all yield: /bucket/a/storage_id/
/a/
a/
/a
a
All of these would yield: "/bucket/storage_id"
/
///
""
None

All these would fail on config parse:
..
/..
/abc/../foo
```
^ Does this behavior seem correct? We are trying to make these some semblance of a unix path, even though I think AWS supports any chars in the key name.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -14,43 +15,45 @@
...
+        "searcher": {
+            "metric": "validation_error",
+            "smaller_is_better": True,
+            "name": "random",
+            "max_trials": num_trials,
+            "max_length": {"batches": 100},
+        },
+        "entrypoint": "model_def:NoOpTrial",
+    }
+    return str(yaml.dump(config_dict))
```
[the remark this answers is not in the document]
varlogtim: done

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -104,12 +112,14 @@ def test_start_tensorboard_for_s3_experiment(tmp_path: Path, secrets: Dict[str,
...
 @pytest.mark.slow  # type: ignore
 @pytest.mark.e2e_cpu  # type: ignore
+@pytest.mark.tensorflow2  # type: ignore
 def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[str, str]) -> None:
     """
     Start 3 random experiments configured with the s3 and shared_fs backends,
     start a TensorBoard instance pointed to the experiments and some select
     trials, and kill the TensorBoard instance.
     """
+    # XXX DET-6094
```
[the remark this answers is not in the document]
varlogtim: This only tests that _some_ metrics load when you specify multiple experiments and trials. It actually doesn't test that _all_ of the trial metrics load. Only one of the specifed experiments works at a time. I feel like the test is misleading. Also, for some reason we run 3 experiments, then were trying to open just two of them - not sure why?

I will remove the comment. It is just odd.

harness/determined/common/storage/s3.py
```diff
@@ -38,6 +39,12 @@ def __init__(
         )
         self.bucket = self.s3.Bucket(self.bucket_name)
 
+        self.prefix = ""
+        if prefix is not None and prefix != "":
+            if ".." in prefix:
```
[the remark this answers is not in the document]
varlogtim: Good call. Updated to match the suggested regexes below.

schemas/expconf/v0/s3.json
```diff
@@ -42,6 +42,14 @@
             ],
             "default": null
         },
+        "prefix": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "pattern": "^[^.]*(?:\\.[^.]+)*$",
```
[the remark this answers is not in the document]
varlogtim: Yeah, I totally forgot about stuff like `asdf..asdf` and also, `blah/.../stuff` <- which I guess is legal. I like your suggestion way better.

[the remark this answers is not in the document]
varlogtim: Also, you test with `grep -E` doesn't work because the regexes here have to be string escaped first. E.g., matching `.` in `grep -E` would be `\.` but here it must be `\\.` because there is escape char overlap.

harness/tests/storage/test_s3.py
```diff
@@ -41,10 +47,48 @@ def live_manager(tmp_path: Path, require_secrets: bool) -> storage.S3StorageMana
...
+        ("foo", "bar", "foo/bar", False),
+        ("/foo", "bar", "foo/bar", False),
+        ("foo/", "bar", "foo/bar", False),
+        ("/foo/", "bar", "foo/bar", False),
+        ("./foo/", "bar", "foo/bar", False),
+        ("/./foo/", "bar", "foo/bar", False),
+        ("/foo/..", "-", "-", True),
+    ],
+)
+def test_storage_prefix_normalization(
```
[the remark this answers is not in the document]
varlogtim: Yup, good call. I can't even call the constructor of the S3StorageManager because it uses the network. I moved this to a helper function so that we can validate it without the network cost.

e2e_tests/tests/test_system.py
```diff
@@ -87,8 +87,9 @@ def test_metric_gathering() -> None:
 
 
 @pytest.mark.e2e_gpu  # type: ignore
-def test_gc_checkpoints_s3(secrets: Dict[str, str]) -> None:
-    config = exp.s3_checkpoint_config(secrets)
+@pytest.mark.parametrize("prefix", [None, "my/test/prefix"])  # type: ignore
```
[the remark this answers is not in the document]
varlogtim: Removed.

harness/tests/storage/test_s3.py
```diff
@@ -41,10 +48,50 @@ def live_manager(tmp_path: Path, require_secrets: bool) -> storage.S3StorageMana
 
 
 @pytest.mark.cloud
-def test_live_s3_lifecycle(live_manager: storage.S3StorageManager, require_secrets: bool) -> None:
+@pytest.mark.parametrize(
```
rb-determined-ai, line 51: Fantastic test.  Excellent coverage of the function, and lightning fast.

harness/tests/test_imports.py
```diff
@@ -47,7 +47,7 @@ def test_imports() -> None:
             bad = []
             for f in found:
                 m = importlib.import_module(f)
-                if hasattr(m, "__file__"):
+                if hasattr(m, "__file__") and m.__file__ is not None:
```
rb-determined-ai, line 50: _weird_.  but ok.

master/pkg/schemas/expconf/storage_config.go
```diff
@@ -131,6 +132,20 @@ type S3ConfigV0 struct {
 	RawAccessKey   *string `json:"access_key"`
 	RawSecretKey   *string `json:"secret_key"`
 	RawEndpointURL *string `json:"endpoint_url"`
+	RawPrefix      *string `json:"prefix"`
+}
+
+// Validate implements the check.Validatable interface.
```
rb-determined-ai, line 138: I don't think you need this.  I think catching it before launching experiments should be sufficient.

Where would this actually get called?

schemas/test_cases/v0/misc.yaml
```diff
@@ -190,3 +190,93 @@
     container: my_container
     connection_string: my_conn_str
     credential: null
+
+- name: s3 checkpoint storage (valid, prefix single dot)
```
rb-determined-ai, line 194: :fire:

rb-determined-ai (approved):
Great work!

master/pkg/schemas/expconf/storage_config.go
varlogtim: Yeah, I couldn't understand these functions at first but after some digging (confirmed with Bradely), when we load up the master.yaml, we walk through all the structs and look for a `Validate()` method and call it. There is some recursive reflection voodoo here: `master/pkg/check/validate.go`, which handles all that.

We need this because if we don't have it and we put some invalid prefix in the `master.yaml`, we won't complain on startup, and will wait to fail until someone tries to launch an experiment ... which is misleading.

PR #2987, opened 2021-09-21
uuids (upstream testing only)

rb-determined-ai:


rb-determined-ai:
upstream tests passed, that's all I needed here.

PR #2990, opened 2021-09-22
fix: tell mypy to ignore azure

rb-determined-ai:


stoksc (approved):

cli/mypy.ini
```diff
@@ -16,3 +16,9 @@ warn_redundant_casts = True
 warn_return_any = True
 warn_unused_configs = True
 warn_unused_ignores = True
+
+[mypy-azure.storage.blob.*]
+; starting in azure 12.9.0:
+; site-packages/azure/storage/blob/_serialize.py:129: error: Type signature has too many arguments
+; site-packages/azure/storage/blob/_blob_client.py:1406: error: Type signature has too few arguments
+follow_imports = skip
```
stoksc, line 24: why not `error`/`silent`? skip seems like the most extreme.

rb-determined-ai: Good question.  I hadn't actually looked at the options before, I just copied from above.

But I just checked, and since this is somebody else's code... `error` just prints noise over code we can't control, and `silent` actually still fails afterwards.

varlogtim (approved):
Lgtm

PR #2991, opened 2021-09-22
chore: recover agent websocket flakes [DET-5935]

stoksc:
## Description
This change adds functionality to recover agent websocket flakes.

## Test Plan
Writing tests right now, going to add a "chaos test" target I think. For manual testing
* add `actors.NotifyAfter(ctx, 10*time.Second, errors.New("connection reset"))` to the websocket prestart
* modify `isClosingError` to treat abnormal closures as socket errors
* submit an experiment and watch it complete, with the agent websocket crashing on both sides every 10 seconds
* or just use socat/nc 🤷 (i also tested this way with `socat TCP-LISTEN:8081,fork,reuseaddr TCP4:localhost:8080` between the master and agent, blipping the tunnel)

## Commentary (optional)
For now, I didn't set the agent to draining. There are some edge cases that are tempting to not handle when we're draining because they're _almost_ not possible (E.G. getting a task scheduled, technically we could get scheduled as we tell the scheduler we're draining, right?); so for the sake of correctness, I handled it this way first. Setting ourselves to drain can be added on top as an improvement.

## Checklist

### DIFF
--- a/master/internal/agent/agent.go
+++ b/master/internal/agent/agent.go
@@ -98,6 +98,14 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
 		}
 		if a.awaitingReconnect {
 			ctx.Log().Info("agent reconnected")
+			// Re-propagate our old state back on successful recovery.
+			if a.enabled {
+				ctx.Tell(a.resourcePool, sproto.EnableAgent{Agent: ctx.Self()})
+			} else {
+				ctx.Tell(a.resourcePool, sproto.DisableAgent{Agent: ctx.Self(), Drain: a.draining})
+			}
+			ctx.Tell(a.slots, patchSlot{Enabled: a.enabled, Drain: a.draining})
+
 			a.awaitingReconnect = false
 			for msg := range a.reconnectBacklog {
 				if err := a.receive(ctx, msg); err != nil {
@@ -192,6 +200,11 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
 		a.socket = nil
 		a.awaitingReconnect = true
 		actors.NotifyAfter(ctx, aproto.AgentReconnectWait, reconnectTimeout{})
+		// Mark ourselves as draining to avoid action on ourselves while we recover. While the
+		// system is technically correct without this, it's better because we avoid any waste
+		// effort scheduling things only to have them suffer AgentErrors later.
+		ctx.Tell(a.resourcePool, sproto.DisableAgent{Agent: ctx.Self(), Drain: true})
+		ctx.Tell(a.slots, patchSlot{Enabled: false, Drain: true})
 	case reconnectTimeout:
 		// Re-enter from actor.ChildFailed.
 		if a.awaitingReconnect {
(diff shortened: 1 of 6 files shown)


master/internal/agent/agent.go
```diff
@@ -106,16 +167,19 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
+		if a.awaitingReconnect {
+			ctx.Respond(errRecovering)
+			return nil
+		}
 
 		ctx.Tell(a.slots, patchSlot{Enabled: true})
 		ctx.Tell(a.resourcePool, sproto.EnableAgent{Agent: ctx.Self()})
 		ctx.Respond(&proto.EnableAgentResponse{Agent: a.summarize(ctx).ToProto()})
 	case *proto.DisableAgentRequest:
-		// Update our state.
```
stoksc, line 116: 🤦 halfway undid a change

master/internal/agent/agent.go
```diff
@@ -68,7 +96,31 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
 		}
+		if a.awaitingReconnect {
+			ctx.Log().Info("agent reconnected")
+			a.awaitingReconnect = false
+
+			// Re-propagate our old state back on successful recovery.
+			if a.enabled {
+				ctx.Tell(a.resourcePool, sproto.EnableAgent{Agent: ctx.Self()})
+			} else {
+				ctx.Tell(a.resourcePool, sproto.DisableAgent{Agent: ctx.Self(), Drain: a.draining})
```
ioga, line 107: So the draining and enabled states here start to differ from the same state in resource pool's structures. We used to have a duplicated state (not accounting for latency between actors passing messages), and now it'll actually be different. I think this adds unnecessary complexity, and I don't like that, I'd store the "previous" state in a separate fields. But I guess we can live with that for now.

master/internal/agent/agent.go
```diff
@@ -68,7 +96,31 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
+			}
+			ctx.Tell(a.slots, patchSlot{Enabled: a.enabled, Drain: a.draining})
+
+			for msg := range a.reconnectBacklog {
+				if err := a.receive(ctx, msg); err != nil {
+					return errors.Wrapf(err, "replaying backlog")
+				}
+			}
+			a.reconnectBacklog = nil
```
ioga, line 116: is that supposed to reset it to empty array? I don't see where it's initialized

ioga (approved):

ioga:
lgtm - it'd take me a week to do the same

master/internal/agent/agent.go
stoksc: since you can append to a nil array in golang, it doesn't need to be initialized. also, yeah, that's why i just set it back to nil.

master/internal/agent/agent.go
stoksc: yeah, earlier in this PR i had an agent state stack, where i pushed the this temp drain on top then pop it off on reconnect (so the state vars were always accurate). i will add that back, i agree this is 'tricky'.

stoksc:
taking a moment before I land to make the agent just give up if it gets a 400 on reconnect.

PR #2994, opened 2021-09-22
fix: always `helm push` latest version

dzhu:
## Description

The version needed to be kept in sync with the chart before, which was
annoying. It turns out `helm push` works fine with the `latest` symlink.

## Test Plan

- [x] run `helm push build/determined-latest.tgz determined` and see that it correctly picks up the versioned file name


eecsliu (approved):
👩‍🍳 😗 👌

PR #2996, opened 2021-09-22
chore: add back some timings that are still valid without a sync

stoksc:
## Commentary (optional)
I went a little too far with the removals. Also, mostly, this makes the webui look better.

## Checklist

### DIFF
--- a/harness/determined/profiler.py
+++ b/harness/determined/profiler.py
@@ -492,6 +492,7 @@ class ProfilerAgent:
             not self.is_enabled
             or not self.timings_is_enabled
             or not self.is_active
+            # Skip recording if this metric requires a sync to be valid and sync is disabled.
             or (not self.sync_timings and requires_sync)
         ):
             yield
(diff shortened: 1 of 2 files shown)


harness/determined/profiler.py
```diff
@@ -492,7 +492,7 @@ def record_timing(
             not self.is_enabled
             or not self.timings_is_enabled
             or not self.is_active
-            or (requires_sync != self.sync_timings)
+            or (not self.sync_timings and requires_sync)
```
armandmcqueen, line 495: This new line was a little hard for me to parse the purpose of. Maybe add a comment here indicating what this conditional is doing.

armandmcqueen (approved):

harness/determined/profiler.py
stoksc:
```suggestion
            # Skip recording if this metric requires a sync to be valid and sync is disabled.
            or (not self.sync_timings and requires_sync)
```

PR #3000, opened 2021-09-23
feat: Implement Model Registry API Backend Read Models (get models/, models/<id>) [DET-5995]

RehanSD:
## Checklist

### DIFF
--- a/master/internal/api_model.go
+++ b/master/internal/api_model.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"bytes"
 	"context"
+	"strconv"
 	"strings"
 	"time"
 
@@ -38,6 +39,8 @@ func (a *apiServer) GetModels(
 		return nil, err
 	}
 
+	allowedUsers := strings.Join(req.users, ",")
+
 	a.filter(&resp.Models, func(i int) bool {
 		v := resp.Models[i]
 
@@ -45,7 +48,27 @@ func (a *apiServer) GetModels(
 			return false
 		}
 
-		return strings.Contains(strings.ToLower(v.Description), strings.ToLower(req.Description))
+		if !strings.Contains(strings.ToLower(v.Description), strings.ToLower(req.Description)) {
+			return false
+		}
+
+		if req.Archived != nil || strconv.FormatBool(req.Archived.Value) != v.Archived {
+			return false
+		}
+
+		if !strings.Contains(allowedUsers, v.Username) {
+			return false
+		}
+		modelLabels := strings.Join(v.Labels, ";")
+		if req.Labels != nil {
+			for _, label := range req.Labels {
+				if !strings.Contains(modelLabels, label) {
+					return false
+				}
+			}
+		}
+
+		return true
 	})
 
 	a.sort(resp.Models, req.OrderBy, req.SortBy, apiv1.GetModelsRequest_SORT_BY_LAST_UPDATED_TIME)
--- a/proto/src/determined/api/v1/model.proto
+++ b/proto/src/determined/api/v1/model.proto
@@ -32,6 +32,8 @@ message GetModelsRequest {
     SORT_BY_CREATION_TIME = 4;
     // Returns models sorted by last updated time.
     SORT_BY_LAST_UPDATED_TIME = 5;
+    // Returns models sorted by number of version.
+    SORT_BY_NUM_VERSIONS = 6;
   }
 
   // Sort the models by the given field.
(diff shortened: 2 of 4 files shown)


master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE public.modelsv2 (
```
ioga, line 1: I am curious, why is it creating new tables instead of altering the old ones?

master/internal/api_model.go
```diff
@@ -38,14 +52,28 @@ func (a *apiServer) GetModels(
 		return nil, err
 	}
 
+	allowedUsers := strings.Join(req.users, ",")
```
mackrorysd, line 55: Huh - am I missing where this gets used? It looks an awful lot like part of one of the contains() approaches that is more concise but less scalable.

master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1
+);
+
+CREATE TABLE public.model_versionsv2 (
+	version integer NOT NULL,
+	model integer NOT NULL REFERENCES public.modelsv2(id),
```
mackrorysd, line 28: TODO: Yeah let's look into why the FOREIGN KEY constraint at the end of the statement wasn't working for you.

master/internal/api_model.go
RehanSD: Nope - this was replaced by `contains()` but I forgot to delete it.

master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
RehanSD: We're changing the primary keys + foreign keys for the tables, and I felt it was easier in that case to just create new ones, copy the data, and delete the old ones. @mackrorysd do you think I should just be altering the tables?

master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE public.modelsv2 (
+	id integer NOT NULL,
+	name character varying UNIQUE NOT NULL,
+	description character varying,
+	metadata jsonb,
+	creation_time timestamp with time zone NOT NULL,
+	last_updated_time timestamp with time zone,
+    num_versions integer DEFAULT 0 NOT NULL,
```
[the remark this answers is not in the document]
RehanSD: Yes, but the tradeoff is storage (one column) vs. database query anytime someone asks for a model - does that tradeoff make sense? If so, I'm happy to do it via a query since it should just be a join and count.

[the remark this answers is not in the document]
ioga: I'd strongly prefer letting postgres do its job and implement this via join-count. I see denormalization as a premature optimization.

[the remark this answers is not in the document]
RehanSD: Sure - will change!

master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
mackrorysd: It does sound easier, but also more risky / messy. But I'll review the specifics of both tables before we meet.

proto/src/determined/api/v1/model.proto
```diff
@@ -49,10 +53,24 @@ message GetModelsRequest {
   string name = 5;
   // Limit the models to those matching the description.
   string description = 6;
+
+  // Limit the models to those with the following labels.
+  repeated string labels = 7;
+
+  // Limit to unarchived models only.
+  google.protobuf.BoolValue archived = 8;
```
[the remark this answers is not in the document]
mackrorysd: +1. The underlying BoolValue class makes it nillable, whereas a naked bool isn't nillable in Go. But in the protobuf definition bool *is* nillable.

master/static/migrations/20210922170815_update-model-registry-get-apis.up.sql
```diff
@@ -0,0 +1,56 @@
...
+    labels text[],
+    readme text,
+    username text,
```
ioga, line 11: forgot one more thing: can username be a foreign key to the user id in `users` table, similar to how `experiments` table has `owner_id`?

proto/src/determined/api/v1/model.proto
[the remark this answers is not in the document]
RehanSD: This actually needs to store 3 values - if its `true` only show `archived` models, if its `false`, only show `unarchived` models, and if its null, show all regardless of `archive` status. If we made it a primitive `bool` it would default to `false` instead of `null`, which is not what we want. This is how the `experiment.proto` file is done as well.

RehanSD:
This PR is a breaking one - we update the model table to use model_id instead of name as a primary key, but this change isn't propagated to all of the model registry methods yet (e.g. PATCH) so its likely that those tests will fail until we fix how PATCH etc is implemented.

PR #3010, opened 2021-09-23
fix: notebook idle check use master port, cert [DET-6013]

naren-determined:
## Description

Fix Notebook Idle timeout check to use master port/cert.

## Checklist

### DIFF
--- a/master/static/srv/check_idle.py
+++ b/master/static/srv/check_idle.py
@@ -4,6 +4,7 @@ import os
 from time import sleep
 from determined.common import api
 from determined.common.api import certs
+import logging
 
 ACTIVE = False
 IDLE = True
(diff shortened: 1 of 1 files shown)


master/static/srv/check_idle.py
```diff
@@ -32,14 +33,18 @@ def main():
...
+                master_url,
                 f"/api/v1/notebooks/{notebook_id}/report_idle",
                 {
                     "notebook_id": notebook_id,
                     "idle": get_execution_state(notebook_server),
                 },
+                cert=cert,
             )
         except Exception as e:
             print(e)
```
rb-determined-ai, line 50: let's make this log a warning with the whole exception, rather than just the error string.  It's too ambiguous like this.

rb-determined-ai (approved):
nice!

PR #3012, opened 2021-09-24
fix: propagate podspec to gc

eecsliu:
## Description
Propagates the podSpec of the experiment to the gc task.

## Test Plan
Launch a kubernetes cluster and use the startup hook to add a sleep statement to `master/static/src/gc-checkpoints-entrypoint.sh` so that you have time to probe the gc pod when it launches. Start and cancel and experiment, and use `kubectl get pod <gc-task-podname> -o yaml` to dump the yaml config of the pod. Visually inspect the yaml to make sure that any podspec options set in the determined exp config have made it to the pod.

## Checklist

### DIFF
--- a/master/pkg/schemas/expconf/environment_config.go
+++ b/master/pkg/schemas/expconf/environment_config.go
@@ -25,13 +25,14 @@ func (p PodSpec) Copy() interface{} {
 func (p PodSpec) Merge(other interface{}) interface{} {
 	out := k8sV1.Pod{}
 	k8sP := k8sV1.Pod(p)
-	k8sOther := k8sV1.Pod(other.(PodSpec))
+	k8sOther := k8sV1.Pod(*other.(*PodSpec))
 	// Copy the low-priority values first.
 	k8sOther.DeepCopyInto(&out)
 	// Overwrite the object with high-priority values.
 	// DeepCopyInto will only copy non-nil values, so this will effectively merge the objects.
 	k8sP.DeepCopyInto(&out)
-	return PodSpec(out)
+	output := PodSpec(out)
+	return &output
 }
 
 // WithDefaults implements the schemas.Defaultable interface.
--- a/master/pkg/schemas/expconf/environment_config.go
+++ b/master/pkg/schemas/expconf/environment_config.go
@@ -25,14 +25,13 @@ func (p PodSpec) Copy() interface{} {
 func (p PodSpec) Merge(other interface{}) interface{} {
 	out := k8sV1.Pod{}
 	k8sP := k8sV1.Pod(p)
-	k8sOther := k8sV1.Pod(*other.(*PodSpec))
+	k8sOther := k8sV1.Pod(other.(PodSpec))
 	// Copy the low-priority values first.
 	k8sOther.DeepCopyInto(&out)
 	// Overwrite the object with high-priority values.
 	// DeepCopyInto will only copy non-nil values, so this will effectively merge the objects.
 	k8sP.DeepCopyInto(&out)
-	output := PodSpec(out)
-	return &output
+	return PodSpec(out)
 }
 
 // WithDefaults implements the schemas.Defaultable interface.
--- a/master/pkg/schemas/expconf/experiment_config.go
+++ b/master/pkg/schemas/expconf/experiment_config.go
@@ -87,12 +87,8 @@ func (e ExperimentConfig) AsLegacy() LegacyConfig {
 	return LegacyConfig{
 		checkpointStorage: schemas.Copy(e.CheckpointStorage()).(CheckpointStorageConfig),
 		bindMounts:        schemas.Copy(e.BindMounts()).(BindMountsConfig),
-		envvars: schemas.Copy(
-			e.Environment().EnvironmentVariables(),
-		).(EnvironmentVariablesMap),
-		podSpec: schemas.Copy(
-			e.Environment().PodSpec(),
-		).(*PodSpec),
+		envvars:           schemas.Copy(e.Environment().EnvironmentVariables()).(EnvironmentVariablesMap),
+		podSpec:           schemas.Copy(e.Environment().PodSpec()).(*PodSpec),
 	}
 }
 
--- a/master/pkg/schemas/expconf/experiment_config.go
+++ b/master/pkg/schemas/expconf/experiment_config.go
@@ -90,6 +90,9 @@ func (e ExperimentConfig) AsLegacy() LegacyConfig {
 		envvars: schemas.Copy(
 			e.Environment().EnvironmentVariables(),
 		).(EnvironmentVariablesMap),
+		podSpec: schemas.Copy(
+			e.Environment().PodSpec(),
+		).(*PodSpec),
 	}
 }
 
--- a/master/pkg/schemas/expconf/legacy.go
+++ b/master/pkg/schemas/expconf/legacy.go
@@ -39,6 +39,7 @@ func (h LegacyConfig) EnvironmentVariables() EnvironmentVariablesMap {
 	return h.envvars
 }
 
+// PodSpec returns a current k8s PodSpec from a LegacyConfig.
 func (h LegacyConfig) PodSpec() *PodSpec {
 	return h.podSpec
 }
@@ -141,6 +142,31 @@ func getEnvironmentVariables(raw map[string]interface{}) (EnvironmentVariablesMa
 	return ev, nil
 }
 
+func getPodSpec(raw map[string]interface{}) (*PodSpec, error) {
+	ps := &PodSpec{}
+
+	envOnly := raw["environment"]
+	if envOnly == nil {
+		return nil, nil
+	}
+
+	evOnly := envOnly.(map[string]interface{})["pod_spec"]
+	if evOnly == nil {
+		return nil, nil
+	}
+
+	rawBytes, err := json.Marshal(evOnly)
+	if err != nil {
+		return ps, errors.Wrap(err, "unable to remarshal environment as json")
+	}
+
+	if err = json.Unmarshal(rawBytes, ps); err != nil {
+		return ps, errors.Wrap(err, "unable to unmarshal environment variables bytes")
+	}
+
+	return ps, nil
+}
+
 // ParseLegacyConfigJSON parses bytes that represent an experiment config that was once valid
 // but might no longer be shimmable to the current version of ExperimentConfig.
 func ParseLegacyConfigJSON(byts []byte) (LegacyConfig, error) {
@@ -175,6 +201,12 @@ func ParseLegacyConfigJSON(byts []byte) (LegacyConfig, error) {
 		return out, err
 	}
 	out.envvars = ev
+
+	ps, err := getPodSpec(raw)
+	if err != nil {
+		return out, err
+	}
+	out.podSpec = ps
 
 	return out, nil
 }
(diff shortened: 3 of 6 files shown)


master/pkg/tasks/task_gc.go
```diff
@@ -21,6 +21,7 @@ type GCCkptSpec struct {
 	Base TaskSpec
 
 	ExperimentID       int
+	Environment        expconf.EnvironmentConfig
```
rb-determined-ai, line 24: This is not going to be possible in all cases (look for other instances that a `tasks.GCCkptSpec` is instantiated).

The reason that checkpoint GC takes a LegacyConfig is that checkpoint gc should _always_ be able to run, even when the experiment against which it is running was configured with some experiment config version that is very old.

You'll have to add a `LegacyConfig.PodSpec()` method (see `pkg/schemas/expconf/legacy.go`).

master/pkg/tasks/task_gc.go
```diff
@@ -46,6 +47,16 @@ func (g GCCkptSpec) ToTaskSpec(allocationToken string) TaskSpec {
...
 
+	if g.Environment.RawPodSpec != nil {
+		resSpec := res.Environment.PodSpec()
+		if resSpec == nil {
+			res.Environment.RawPodSpec = g.Environment.PodSpec()
+		} else {
+			newSpec := resSpec.Merge(g.Environment.PodSpec()).(*expconf.PodSpec)
+			res.Environment.RawPodSpec = newSpec
+		}
+	}
```
rb-determined-ai, line 58: Use `schemas.Merge()` and it will handle these nil cases automatically

rb-determined-ai (commented):
Were there were some dangerous fields that we didn't want to pass straight into the podspec?

eecsliu:
The only dangerous fields I found so far (coscheduler label and gpu resource requests) are set in `master/internal/kubernetes/spec.go`. I checked and these fields overwrite what's in the base podspec, which is correct behavior.

master/pkg/tasks/task_gc.go
```diff
@@ -46,6 +46,10 @@ func (g GCCkptSpec) ToTaskSpec(allocationToken string) TaskSpec {
 	}
 	res.Environment = schemas.WithDefaults(env).(expconf.EnvironmentConfig)
 
+	resSpec := res.Environment.PodSpec()
+	newSpec := resSpec.Merge(g.LegacyConfig.PodSpec()).(*expconf.PodSpec)
+	res.Environment.RawPodSpec = newSpec
```
rb-determined-ai, line 51: Don't use the `Raw` API here, use the `SetPodSpec()`.  The Raw API is only for when you have not finished parsing something.

master/pkg/schemas/expconf/legacy.go
```diff
@@ -21,6 +21,7 @@ type LegacyConfig struct {
 	checkpointStorage CheckpointStorageConfig
 	bindMounts        BindMountsConfig
 	envvars           EnvironmentVariablesMap
+	podSpec           *PodSpec
```
rb-determined-ai, line 24: Could you add tests to this in `legacy_test.go`?

master/pkg/tasks/task_gc.go
```diff
@@ -46,6 +46,9 @@ func (g GCCkptSpec) ToTaskSpec(allocationToken string) TaskSpec {
 	}
 	res.Environment = schemas.WithDefaults(env).(expconf.EnvironmentConfig)
 
+	resSpec := res.Environment.PodSpec()
```
stoksc, line 49: `resSpec` comes straight out of `schemas.WithDefaults(env).(expconf.EnvironmentConfig)` for which the default value given by the jsonschema is `null`, and `expconf.PodSpec.Merge` isn't nil-safe (the k8s code inside it does an immediate deref), so wouldn't this always panic?

stoksc: Local non-k8s run gets a panic exactly here
```
INFO[2021-09-27T09:15:48-04:00] starting checkpoint garbage collection        id=experiment-8-checkpoint-gc system=master type=checkpointGCTask
ERRO[2021-09-27T09:15:48-04:00] runtime error: invalid memory address or nil pointer dereference
goroutine 5042 [running]:
runtime/debug.Stack(0xc000eb3578, 0x34f27e0, 0x5162340)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/determined-ai/determined/master/pkg/actor.(*Ref).close(0xc000fa5500)
        /Users/bradleylaney/Code/determined/master/pkg/actor/ref.go:306 +0x9a5
panic(0x34f27e0, 0x5162340)
        /usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/determined-ai/determined/master/pkg/tasks.GCCkptSpec.ToTaskSpec(0x0, 0x0, 0x0, 0x0, 0xc00013a750, 0x24, 0xc00005c640, 0x36, 0x0, 0x0, ...)
        /Users/bradleylaney/Code/determined/master/pkg/tasks/task_gc.go:50 +0x333
github.com/determined-ai/determined/master/internal.(*checkpointGCTask).Receive(0xc0004f2c00, 0xc001070e40, 0x39919ef, 0x18)
        /Users/bradleylaney/Code/determined/master/internal/checkpoint_gc.go:55 +0x808
github.com/determined-ai/determined/master/pkg/actor.(*Ref).processMessage(0xc000fa5500, 0x34ed900)
        /Users/bradleylaney/Code/determined/master/pkg/actor/ref.go:253 +0x36e
github.com/determined-ai/determined/master/pkg/actor.(*Ref).run(0xc000fa5500)
        /Users/bradleylaney/Code/determined/master/pkg/actor/ref.go:265 +0xa5
created by github.com/determined-ai/determined/master/pkg/actor.newRef
        /Users/bradleylaney/Code/determined/master/pkg/actor/ref.go:91 +0x571  id=experiment-8-checkpoint-gc system=master type=checkpointGCTask
```

stoksc: I think we'd want to merge the other way, `g.LegacyConfig.PodSpec().Merge(resSpec)` since the receiver takes higher priority in the merge, and we prob want to just make merge nil-safe.

rb-determined-ai: > and expconf.PodSpec.Merge isn't nil-safe

There's a fix for this:

> Use schemas.Merge() and it will handle these nil cases automatically

eecsliu: ahhhhhh ok I didn't properly read @rb-determined-ai's original instructions

stoksc: > schemas.Merge()

forgot this existed

stoksc: side note: this highlights we don't have a single test that makes sure GC runs OK, right? else it would've failed.

stoksc: https://determinedai.atlassian.net/browse/DET-6032

rb-determined-ai: We have an e2e test which should catch this.

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -90,6 +90,9 @@ func (e ExperimentConfig) AsLegacy() LegacyConfig {
 		envvars: schemas.Copy(
 			e.Environment().EnvironmentVariables(),
 		).(EnvironmentVariablesMap),
+		podSpec: schemas.Copy(
+			e.Environment().PodSpec(),
+		).(*PodSpec),
```
rb-determined-ai, line 95: nit: are you sure this doesn't fit on one line?

master/pkg/schemas/expconf/legacy.go
```diff
@@ -136,6 +142,31 @@ func getEnvironmentVariables(raw map[string]interface{}) (EnvironmentVariablesMa
...
+	}
+
+	evOnly := envOnly.(map[string]interface{})["pod_spec"]
+	if evOnly == nil {
+		return nil, nil
+	}
+
+	rawBytes, err := json.Marshal(evOnly)
+	if err != nil {
+		return ps, errors.Wrap(err, "unable to remarshal environment as json")
```
rb-determined-ai, line 160: probably should return `nil, error`, since it's a `*PodSpec` and not a `PodSpec` we're returning.

master/pkg/tasks/task_gc.go
stoksc: oh.. lol i see, it failed. i just totally missed it.

rb-determined-ai: > I think we'd want to merge the other way, g.LegacyConfig.PodSpec().Merge(resSpec) since the receiver takes higher priority in the merge

I think bradley's right that the order here isn't quite right.  Normally what we do for submitted experiments is:
- load user-provided values
- merge in values from templates
- merge in values from task container defaults
- call WithDefaults for any remaining values.

I think when we instantiate the `expconf.EnvironmentConfig{}` object above we should set the `g.LegacyConfig.PodSpec()` in there directly, since this is closest to the user-provided values, and therefore should have the highest priority (as the envvars do)

master/pkg/schemas/expconf/legacy.go
```diff
@@ -136,6 +142,31 @@ func getEnvironmentVariables(raw map[string]interface{}) (EnvironmentVariablesMa
...
 
+func getPodSpec(raw map[string]interface{}) (*PodSpec, error) {
+	ps := &PodSpec{}
+
+	envOnly := raw["environment"]
+	if envOnly == nil {
+		return nil, nil
+	}
+
+	evOnly := envOnly.(map[string]interface{})["pod_spec"]
```
rb-determined-ai, line 153: `evOnly`?  Prolly rename this.

master/pkg/schemas/expconf/legacy.go
```diff
@@ -136,6 +142,31 @@ func getEnvironmentVariables(raw map[string]interface{}) (EnvironmentVariablesMa
...
+		return nil, nil
+	}
+
+	rawBytes, err := json.Marshal(evOnly)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to remarshal environment as json")
+	}
+
+	if err = json.Unmarshal(rawBytes, ps); err != nil {
+		return nil, errors.Wrap(err, "unable to unmarshal environment variables bytes")
```
rb-determined-ai, line 164: ... and check for other copy/pasta errors

rb-determined-ai (commented):
other than the ordering issue, this looks good

master/pkg/schemas/expconf/legacy.go
eecsliu: ahahaha I got sloppy this morning

eecsliu: thanks

rb-determined-ai (approved):

master/pkg/schemas/expconf/environment_config.go
```diff
@@ -25,13 +25,14 @@ func (p PodSpec) Copy() interface{} {
...
-	k8sOther := k8sV1.Pod(other.(PodSpec))
+	k8sOther := k8sV1.Pod(*other.(*PodSpec))
 	// Copy the low-priority values first.
 	k8sOther.DeepCopyInto(&out)
 	// Overwrite the object with high-priority values.
 	// DeepCopyInto will only copy non-nil values, so this will effectively merge the objects.
 	k8sP.DeepCopyInto(&out)
-	return PodSpec(out)
+	output := PodSpec(out)
+	return &output
```
rb-determined-ai, line 35: Something is wrong with this.  A `func (p PodSpec) Merge` _must_ return a plain `PodSpec`, not a `*PodSpec`.  Otherwise, keeping track of what functions returned what types would be a terrible mess.

What arose that caused you to need this?

eecsliu: Was testing on a stale copy of my code. We're good

PR #3015, opened 2021-09-27
chore: upgrade to Go 1.17

stoksc:
## Description
This change updates the `determined-master` and `determined-agent` to Go 1.17.

## Test Plan
- [x] run reasonable number of e2es to test stuff (at least aws/gcp/k8s)
- [x] read release notes

## Checklist


agent/go.mod
```diff
@@ -29,6 +29,223 @@ require (
 	gotest.tools v2.2.0+incompatible
 )
 
+require (
```
mackrorysd, line 32: For posterity, anyone wondering what's up with this, from the Go 1.17 release notes:

> For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

mackrorysd (approved):

agent/go.mod
stoksc: I'd tack on https://golang.org/ref/mod is great.

PR #2992, opened 2021-09-22
chore: handle query param jwt for external auth

hkang1:
## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234

PR #3016, opened 2021-09-27
chore: add server-side portion of external session handling

mackrorysd:
## Description

Server-side portion of this work: https://github.com/determined-ai/determined/pull/2992

## Test Plan

Did an end-to-end test with an implementation of the external session functionality.


hkang1 (approved):
The WebUI and the protobuf portions looks good to me. The go portions look reasonable but may want another set of eyes for them if they are cause for concern.

master/internal/api_auth.go
```diff
@@ -29,6 +30,10 @@ func replicateClientSideSaltAndHash(password string) string {
 
 func (a *apiServer) Login(
 	_ context.Context, req *apiv1.LoginRequest) (*apiv1.LoginResponse, error) {
+	if len(os.Getenv("EXTERNAL_JWT_KEY")) > 0 {
```
mackrorysd, line 33: I wasn't a big fan of doing this via environment variables, but the config isn't easily accessible from some of the packages we need to make this decision, like early in the process of handling protobuf requests.

stoksc: I feel pretty strongly this has to go in into the config. You can even wire up config variables to be settable by environment variables if it's necessary, but I don't want to have random env vars read all over the master code, I want the source of truth to be the master config (which is already confusing enough as is, tbh). 

> but the config isn't easily accessible from some of the packages we need to make this decision, like early in the process of handling protobuf requests.

What exactly do you mean by this?

stoksc (commented):
after having read the code, I'm not convinced the things being pulled from the env are terribly hard to pack into/read from the master config. can we move them there?

master/internal/api_auth.go
mackrorysd: grpcutil is the tricky part where we're authenticating with very little context built up. Attempting to use master config from in there created a mess of circular dependencies. After chatting yesterday I tried a few times to refactor this. What I've ended up with works and avoids the use of environment variables. But it does mean the master config is split up further - moving it to pkg ends up being a massive refactor and eventually around the storage configs it starts creating even worse circular dependencies. So it's not as clean as I'd like, but it does mean we use the config, of pieces of it.

master/internal/db/postgres_users.go
```diff
@@ -2,15 +2,19 @@ package db
 
 import (
 	"crypto/ed25519"
+	"crypto/rsa"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/golang-jwt/jwt"
```
stoksc, line 11: since I had to double check this was the right library, could you add `github.com/dgrijalva/jwt-go` to the golangci-lint blacklist?

master/internal/db/postgres_users.go
```diff
@@ -72,6 +76,76 @@ WHERE user_sessions.id=$1`, &user, session.ID); errors.Cause(err) == ErrNotFound
...
+		}
+
+		if rErr := tx.Rollback(); rErr != nil {
+			log.Errorf("error during rollback: %v", rErr)
+		}
+	}()
+	if err != nil {
+		return nil, nil, errors.WithStack(err)
+	}
+	user, err := db.UserByUsername(claims.Email)
```
stoksc, line 118: when access control comes in to play, this will still happen auto-magically but the user will have nil permissions right? so it'll be inconsequential.

master/internal/user/service.go
```diff
@@ -119,6 +153,11 @@ func (s *Service) postLogout(c echo.Context) (interface{}, error) {
 }
 
 func (s *Service) postLogin(c echo.Context) (interface{}, error) {
+	if s.extConfig.JwtKey != "" {
```
stoksc, line 156: need a similar error to the this one and the one in grpc for patch user password below, i think.

master/internal/user/service.go
```diff
@@ -48,40 +48,74 @@ func (h *agentUserGroup) Validate() (*model.AgentUserGroup, error) {
...
+	} else if cookie, err := c.Cookie("auth"); err == nil {
+		return cookie.Value, nil
+	}
+	// If we found no token, then abort the request with an HTTP 401.
+	return "", echo.NewHTTPError(http.StatusUnauthorized)
+}
+
+// ProcessAuthentication is a middleware processing function that attempts
+// to authenticate incoming HTTP requests.
 func (s *Service) ProcessAuthentication(next echo.HandlerFunc) echo.HandlerFunc {
```
stoksc, line 82: since this is auth logic, i really like it to be dedup'd as much as possible to avoid careless mistakes.. we should refactor ProcessAuthentication/ProcessExternalAuthentication/grpcutil.GetUser to avoid this probably, at some point. there was a lot you had to remember to do here.

stoksc (approved):
killer, looks great. just had a few minor comments but nothing blocking.

master/internal/api_auth.go
stoksc: yeah, i think it's fine for now.. i think i can one click (with goland) refactor the entire mess to like `pkg/master_config` and never think about it again.

stoksc: maybe i'll see what https://github.com/rsc/rf can do one day

stoksc: we do have a 'master config' cleanup on the books along the lines of the expconf stuff, just never enough bandwidth

master/internal/db/postgres_users.go
mackrorysd: When access control comes into play, something else will determine is Admin / Active are true or false when the user is created / or will have already updated the database.

stoksc: cool, yeah, makes sense to me.

PR #3019, opened 2021-09-28
chore: mark open allocs as closed on restart

stoksc:
## Description
This change ensure that, upon master crashes, allocations are marked as ended, so they don't get counted ad infinitum.

## Test Plan
- [x] test the query on a real DB

## Checklist

### DIFF
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -119,3 +119,16 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 		"DELETE FROM allocation_sessions WHERE allocation_id=$1", allocationID)
 	return err
 }
+
+// CloseOpenAllocations finds all allocations that were open when the master crashed
+// and adds an end time.
+func (db *PgDB) CloseOpenAllocations() error {
+	if _, err := db.sql.Exec(`
+UPDATE allocations
+SET end_time = current_timestamp AT TIME ZONE 'UTC'
+WHERE end_time IS NULL
+`); err != nil {
+		return errors.Wrap(err, "closing old allocations")
+	}
+	return nil
+}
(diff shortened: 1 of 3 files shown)


master/internal/db/postgres_tasks.go
```diff
@@ -119,3 +119,16 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 		"DELETE FROM allocation_sessions WHERE allocation_id=$1", allocationID)
 	return err
 }
+
+// CloseOpenAllocations finds all allocations that were open when the master crashed
+// and adds an end time.
+func (db *PgDB) CloseOpenAllocations() error {
+	if _, err := db.sql.Exec(`
+UPDATE allocations
+SET end_time = current_timestamp AT TIME ZONE 'UTC'
```
dzhu, line 128: This is somewhat subjective, but my first thought would've been to set the end time to the start time. I vaguely feel like losing the allocation in the aggregations is better than potentially having it show up spuriously for the entirety of the potentially long time the master wasn't running. But maybe it's a good enough assumption that the master will be restarted soon after it crashes? Maybe just drop a quick comment to explain the reasoning.

dzhu (approved):

master/internal/db/postgres_tasks.go
stoksc: Yeah I did think about this exactly, I figured I'd rather have accurate accounting for clusters that are always up (or close to) as opposed to optimizing for clusters that have long periods of downtime. I think both have downsides, worst case with this, you could count a lot of stuff longer than you should, worst case with the other, you could count all allocations as 0 if the master is constantly restarting.

stoksc: Best case may be a `cluster_heartbeat` table and to take the end as `cluster_heartbeat.latest` on restart.

stoksc: Sorry, I forgot to drop the comment.

PR #3021, opened 2021-09-28
fix: fix nested hparams with grid [DET-6045]

rb-determined-ai:
## Description

Nested hparams weren't working properly with grid search.

## Test Plan

Added a test which failed without this fix.

### DIFF
--- a/master/pkg/searcher/grid_test.go
+++ b/master/pkg/searcher/grid_test.go
@@ -1,9 +1,11 @@
 package searcher
 
 import (
+	"encoding/json"
 	"strconv"
 	"testing"
 
+	"github.com/pkg/errors"
 	"gotest.tools/assert"
 
 	"github.com/determined-ai/determined/master/pkg/ptrs"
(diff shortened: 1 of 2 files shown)


rb-determined-ai:
@liamcli as I said before, I rewrote the grid sampling logic almost entirely.  I think it is simpler as-written, but I can go back and fix _just_ the bug if you'd like.

master/pkg/searcher/grid_test.go
```diff
@@ -44,17 +46,34 @@ func TestGridFunctionality(t *testing.T) {
 
 func TestHyperparameterGridMethod(t *testing.T) {
 	dParam := expconf.DoubleHyperparameter{RawMaxval: 2.0, RawCount: ptrs.IntPtr(5)}
-	assert.Equal(t, len(grid(expconf.Hyperparameter{RawDoubleHyperparameter: &dParam})), 5)
+	assert.Equal(t,
+		len(getGridAxes([]string{"x"}, expconf.Hyperparameter{RawDoubleHyperparameter: &dParam})),
```
liamcli, line 50: need to index 0 on the result of getGridAxes for this assert as well as others in this test.

rb-determined-ai: makes sense, thank you

liamcli (approved):
lgtm, thanks for fixing it!

PR #3022, opened 2021-09-28
fix: hide secrets when printing config in trial log  [DET-6046]

varlogtim:
Manually tested.

### DIFF
--- a/harness/determined/exec/launch_autohorovod.py
+++ b/harness/determined/exec/launch_autohorovod.py
@@ -5,11 +5,13 @@ It launches the entrypoint script under horovodrun when slots_per_trial>1, or as
 subprocess otherwise.
 """
 
+import copy
 import logging
 import socket
 import subprocess
 import sys
 import time
+from typing import Dict
 
 import simplejson
 
@@ -19,6 +21,22 @@ from determined import horovod
 from determined.common import api, constants, storage
 from determined.common.api import certs
 from determined.constants import HOROVOD_SSH_PORT
+
+
+def mask_config_dict(d: Dict) -> Dict:
+    mask = "********"
+    new_dict = copy.deepcopy(d)
+
+    # checkpoint_storage
+    hidden_checkpoint_storage_keys = ("access_key", "secret_key")
+    try:
+        for key in new_dict["checkpoint_storage"].keys():
+            if key in hidden_checkpoint_storage_keys:
+                new_dict["checkpoint_storage"][key] = mask
+    except KeyError:
+        pass
+
+    return new_dict
 
 
 def main() -> int:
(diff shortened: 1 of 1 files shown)


varlogtim:
I think I need to make this walk through lists as well.

varlogtim:
> I thought the approach we agreed on was "mask the things we _know_ are secrets and leave everything else alone"? This looks like the "try and guess what things might be secrets, including user-provided values where we have no context at all".

Sure. Done.

harness/determined/common/util.py
```diff
@@ -5,7 +5,8 @@
 import platform
 import random
 import sys
-from typing import IO, Any, Callable, Iterator, Sequence, TypeVar, Union, overload
+from copy import deepcopy
```
rb-determined-ai, line 8: can you import the module instead of the function?

This is the standard we (try to) use in new code, and has been for a couple years.  There's still lots of old that hasn't been modified during that time that doesn't follow this rule, but it's still much better if we follow it.

harness/determined/common/util.py
```diff
@@ -127,3 +128,19 @@ def get_config_path() -> pathlib.Path:
         config_path = pathlib.Path.home().joinpath(".config")
 
     return config_path.joinpath("determined")
+
+
+def mask_config_dict(d: Dict) -> Dict:
```
rb-determined-ai, line 133: This isn't really a "utility" function at all... it only works on dictionaries which _happen_ to be the experiment config, which only really occurs in one place.

What I'm getting at is: in the absence of other places that we want to use this function, can we just put the function right in `harness.py`?

rb-determined-ai:
otherwise, the change looks good

harness/determined/common/util.py
varlogtim: Sure, can do. I was curious about the motivation behind this? Is it just so that function calls in the code specify the full  module.submodule.function()?

harness/determined/common/util.py
varlogtim: Do you mean `launch_autohorovid.py` instead of `harness.py`?

harness/determined/exec/launch_autohorovod.py
```diff
@@ -21,6 +23,22 @@
...
 
+def mask_config_dict(d: Dict) -> Dict:
+    mask = "********"
+    new_dict = copy.deepcopy(d)
+
+    # checkpoint_storage
+    hidden_checkpoint_stroage_keys = ["access_key", "secret_key"]
+    try:
+        for key in new_dict["checkpoint_storage"].keys():
+            if key in hidden_checkpoint_stroage_keys:
```
rb-determined-ai, line 34: two things: one is a spelling error ("stroage") and the other is that you could really just inline `["access_key", "secret_key"]` here like `if key in ("access_key", "secret_key"):`

rb-determined-ai (approved):

PR #3025, opened 2021-09-29
feat: run db migrations in transactions [DET-5987]

ioga:
## Checklist


master/internal/db/migrations.go
```diff
@@ -1,37 +1,196 @@
...
+	}
+
+	return nil
+}
+
+// Migrate runs the migrations from the specified directory URL.
+func (db *PgDB) Migrate(migrationURL string, actions []string) error {
+	// go-pg/migrations uses go-pg/pg connection API, which is not compatible
+	// with pgx, so we use a one-off go-pg/pg connection.
+	pgOpts, err := makeGoPgOpts(db.url)
```
stoksc, line 135: seems this could be a method on `db *PgDB` and we could store more than just `url` on it to avoid all the regex'ing, etc. but not a huge deal.

stoksc (commented):
this is interesting; i'm still not completely convinced running migrations in transactions is the going to be right for us long term and this has become relatively complex, so i have mixed feelings about adding it. how do you feel about like, go-migrate + https://stackoverflow.com/questions/731851/idempotent-postgresql-ddl-scripts + no down migrations, when we start if migrate.Version() is dirty we Step(-1) down one and just re-migrate. The only issue is enforcing this, and it does make some migrations very difficult (but there are also some pg stuff that can't be done in a tx, but there are usually work arounds there too).

ioga:
>this has become relatively complex

it has become complex because of the go-pg/pg vs pgx drivers. if we didn't have to do `makeGoPgOpts`, it'd be smooth. We could write our own library in a single file, even make it compatible with the go-pg/pg metadata table structure to make the transition smooth.

>go-migrate + https://stackoverflow.com/questions/731851/idempotent-postgresql-ddl-scripts + no down migrations, when we start if migrate.Version() is dirty we Step(-1) down one and just re-migrate. The only issue is enforcing this, and it does make some migrations very difficult

With transactions, you can ensure the "outside world" (i.e. the old running master) will either see the version "before" or "after". If you do an idempotent sequence, it'll see "before", "after", or any intermediate state. You can't address that in the old code, only in the migration... and the standard way to do it in migration is transactions + optional explicit locks.

>(but there are also some pg stuff that can't be done in a tx, but there are usually work arounds there too).

I suppose the workaround for "not MVCC-safe" actions (`DROP DATABASE`, `TRUNCATE`, some `ALTER TABLE`s) is to take an exclusive lock on everything before the migration. So we'll stop the world, but we won't break it. This can be handled as a "offline upgrade" by the operators. 
Running the same commands without transactions and locks will in fact break it.

master/internal/db/migrations.go
ioga: I think *ideally* we either patch ParseURL in the upstream to make this not as bad, or make our own migrations lib where this is not necessary at all. Then having a specialized `PgOpts` construction in `*PgDB` would have to be taken out... And so I went for this more compartmentalized approach with all migrations code (and hacks) limited to this one file.

stoksc:
> With transactions, you can ensure the "outside world" (i.e. the old running master) will either see the version "before" or "after". If you do an idempotent sequence, it'll see "before", "after", or any intermediate state. You can't address that in the old code, only in the migration... and the standard way to do it in migration is transactions + optional explicit locks.

yeah, i'm aware, i guess the other part (i didn't state sorry) is you have to write migrations that are additive only, or have a second, destructive half that runs after a flip to a new version. it's fine to see "before", "after" or intermediate so long as you write migrations in a certain way.

> I suppose the workaround for "not MVCC-safe" actions (DROP DATABASE, TRUNCATE, some ALTER TABLEs) is to take an exclusive lock on everything before the migration. So we'll stop the world, but we won't break it. This can be handled as a "offline upgrade" by the operators.
Running the same commands without transactions and locks will in fact break it.

even though postgres uses mvcc, a lot of DDL's performed in transactions will lock the entire table anyway (without an explicit lock). if you insert into a table in a hot loop on postgres 10 and begin; alter table ... it will lock the table. so what i mean by "i don't see how running in transactions works long term" is exactly this. but if we're eternally ok with offline upgrades this is better, i agree.

ioga:
>yeah, i'm aware, i guess the other part (i didn't state sorry) is you have to write migrations that are additive only, or have a second, destructive half that runs after a flip to a new version. it's fine to see "before", "after" or intermediate so long as you write migrations/code to not know/care.

But down migrations are still useful, if you want to do a rollback. It has to be explicit (via `determined-master migrate down $VERSION`) because it's potentially destructive, but I don't think we can "ban" it. It's helpful in development when switching between branches, if nothing else.

>even though postgres uses mvcc, a lot of DDL's performed in transactions will lock the entire table anyway (without an explicit lock). if you insert into a table in a hot loop on postgres 10 and begin; alter table ... it will lock the table. so what i mean by "i don't see how running in transactions works long term" is exactly this. but if we're eternally ok with offline upgrades this is better, i agree.

But if it's not wrapped in an explicit transaction, the alter table will still acquire the locks to run this particular query. So for a one-line migration with a single `ALTER TABLE` statement, it'll be equivalent.
I doubt all changes can be online, but I hope a chunk of them will.

stoksc:
> But down migrations are still useful, if you want to do a rollback. It has to be explicit (via determined-master migrate down $VERSION) because it's potentially destructive, but I don't think we can "ban" it. It's helpful in development when switching between branches, if nothing else.

true

> But if it's not wrapped in an explicit transaction, the alter table will still acquire the locks to run this particular query. So for a one-line migration with a single ALTER TABLE statement, it'll be equivalent.
I doubt all changes can be online, but I hope a chunk of them will.

duh, of course, i had a brain fart.

PR #3029, opened 2021-09-30
chore: rewrite schemas package without typing internals

rb-determined-ai:


harness/determined/common/schemas/_schema_base.py
```diff
@@ -1,13 +1,45 @@
...
-import typing
-from typing import Any, List, Mapping, Optional, Sequence, Type, TypeVar, cast
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Type, TypeVar
 
 from determined.common import schemas
 from determined.common.schemas import expconf
 
 PRIMITIVE_JSON_TYPES = (numbers.Number, str, bool, type(None))
 
+# `typing` has some awkward APIs for type inspection.  In general, the only thing we can rely on
```
ioga, line 10: when can we make it simpler? after retiring 3.6, 3.7, ...?

ioga (approved):
where did duck typing

harness/determined/common/schemas/_schema_base.py
rb-determined-ai: my guess is "when we decide to remove _instance_from_annotation and come up with a whole different solution".

In 3.6 you could check if a given annotation was some sort of Union, but I can't find any API for that in 3.7+, so all I could come up with was checking equality against a list of whitelisted Union types.

rb-determined-ai:
> where did duck typing [go?]

If we subclassed all of the primitive types (`int`, `str`, `bool`, etc) then we could do an easy duckt-yping solution.  I was never willing to consider that though, even in the Go code.

PR #3032, opened 2021-10-01
feat: fastapi py client

hamidzr:
## Description

python API client code using fastapi client.  generator. compared to swaggers this results in 1/10 lines of text in py modules and provides modern type hints. 

- using determined auth
- response shape validation
- type definitions for inputs, outputs, and apis
- no third-party runtime dependencies.

- remove determined_ prefix from openapi spec operationids
- speed up harness `make check`

## Test Plan

- `cli/xperiment.py`
- `python -X importtime -c 'import determined.cli' 2>&1 > /tmp/clistartup.txt`
- `python -m cProfile -o /tmp/cligen.prof $(which det) e list`
- `snakeviz /tmp/cligen.prof`

## Commentary (optional)

the following is no longer needed:
```
optimization route 1: include only what we need and remove/comment the rest => about a 50ms startup hit
optimization route 2: remove fastapi => 100ms startup hit w/o having to do any other manual work.
 
suggested usage for minimizing startup cost:
- import the module you want eg `from .fastapi_client/api/experiments_api.py ...`
- selectively (un)comment parts in `fastapi_client/api/**` and `fastapi_client/models.py`
```

future work:
  - add ci check

TODO
- [x] make a real example
- [x] ? improve json repr
- [x] remove demo code
- [ ] address the rest of mypy complaints

## Checklist


harness/determined/cli/xperiment.py
```diff
@@ -0,0 +1,60 @@
...
+# def list(_: Namespace) -> None:
+#     # type info shows for some language servers but not all since the generated code
+#     # is using legacy docstring types
+#     api_response = swg.experiment_api.determined_get_experiments(resource_pools=["default"])
+#     print(api_response.experiments)
+
+
+@auth_required
+def list(args: Namespace) -> None:
+    response = experiments_api.determined_get_experiments(limit=5)
```
ioga, line 24: why do all these methods have `determined` in them?

harness/determined/common/api/fapi.py
```diff
@@ -0,0 +1,80 @@
...
+        url = (self.host or "") + url.format(**path_params)
+        response = do_request(method, self.host, url, auth=self.auth, **kwargs)
+        try:
+            return parse_obj_as(type_, response.json())
+        except ValidationError as e:
+            raise ResponseHandlingException(e)
+
+
+client = ApiClient(host="http://localhost:8080")
+experiments_api = SyncExperimentsApi(client)  # type: ignore
```
ioga, line 47: ok so right now this code always hits localhost, and is not tied up to MASTER_URL / certificate infrastructure?

requirements.txt
```diff
@@ -26,3 +26,4 @@ bump2version>=1.0.1
 # Earlier versions had different type annotations
 pyzmq>=22
 devcluster>=1,<2
+fastapi
```
ioga, line 29: there're fastapi and pydantic imports in the generated files, so I suspect `harness/setup.py` needs that as a dependency.

harness/determined/cli/xperiment.py
hamidzr: I think they come from operationids in the swagger spec file.
image

not sure where those come from but we should be able to modify them. they appear in the name of the operation here as well https://docs.determined.ai/latest/rest-api/#/Experiments/Determined_GetExperiments

rb-determined-ai:
I spent some time playing with the `pyinstrument` you showed me, and I saw what you had reported, which is roughly 100ms slowdown with this commit.

Furthermore, I noticed that a 25ms of that seems to be importing `jsonschema` from the validation packages.  50ms of that seems to be almost entirely from `register_known_type`, which is straight out of my schemas code.

I'd kinda rather not slow down our cli by 25% (I'm seeing just shy of 0.3s without this change, just shy of 0.4s with this change).  Can I spend the next day taking a shot at this problem?  It seems like the only weak point is where you are importing my code lol...

rb-determined-ai:
also, could you rebase this change and make sure it runs?  I had to fix a syntax error to test it.

hamidzr:
> 50ms of that seems to be almost entirely from register_known_type,

with the latest changes to the cli in master the pyinstrument output might be more readable now. could you tell if this from the added runtime call to register the newly defined models here or something else? I tried timing that section alone which happens in `fapi_client/__init__.py` and seemed much faster.

> I'd kinda rather not slow down our cli by 25% (I'm seeing just shy of 0.3s without this change, just shy of 0.4s with this change). Can I spend the next day taking a shot at this problem? It seems like the only weak point is where you are importing my code lol...

it makes sense, it used to be a much smaller % of the total time but not anymore. Yeah that'd be great!

PR #3033, opened 2021-10-01
chore: adding markdown component

EmilyBonar:
## Description
This PR doesn't use the new Markdown component anywhere yet, it just introduces it to the codebase.

## Checklist


webui/react/config-overrides.js
```diff
@@ -111,7 +111,7 @@ const webpackConfig = override(
   addWebpackPlugin(
     new MonacoWebpackPlugin({
       // available options are documented at https://github.com/Microsoft/monaco-editor-webpack-plugin#options
-      languages: ['yaml'],
+      languages: ['yaml', "markdown"],
```
hkang1, line 114: **super nit:** alphabetical

webui/react/src/components/Markdown.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+_This is italic text_
+
+~~Strikethrough~~
+
+
+## Blockquotes
+
+
+> This is a blockquote
```
hkang1, line 43: **chore:** add secondary `>>` and tertiary `>>>` block quotes, since they're cool to see!

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
+import { Tabs } from 'antd';
+import { default as MarkdownViewer } from 'markdown-to-jsx';
```
hkang1, line 2: **question:** can't this just be:
```
import MarkdownViewer from 'markdown-to-jsx';
```

webui/react/src/components/Markdown.module.scss
```diff
@@ -0,0 +1,754 @@
+.base {
```
hkang1, line 1: **chore:** I'm guessing this is sourced from something. We can have a credit link? Also if it changes down the line with improvements we will know where to go take a peek.

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import css from './Markdown.module.scss';
+import Spinner from './Spinner';
+
+const { TabPane } = Tabs;
+const MonacoEditor = React.lazy(() => import('components/MonacoEditor'));
+
+interface Props {
+  editing?: boolean;
+  height?: string;
```
hkang1, line 13: **question:** css props like these can support numbers too I think:
```
height?: number | string;
```

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+  markdown: string;
+  onChange?: (editedMarkdown: string) => void;
+}
+
+const Markdown: React.FC<Props> = ({ editing=false, height='80vh', markdown, onChange }: Props) => {
+  return (
+    editing ?<Tabs>
```
hkang1, line 20: **nit style:** for ternary JSX (not sure if there's a eslint for this) but IMO something like this a little bit easier to distinguish the true and false JSXs.
```
return editing ? (
  <Tabs>
    ...
  </Tabs>
) : (
  <div ...>
    ...
  </div>
);
```

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <TabPane key="1" tab="Edit">
```
hkang1, line 21: **chore:** this is not immediately needed but planning ahead, we might want to expose the editing state in the URL of the respective pages. To better prep for that, it'll be good to have an `enum TabType` and use that as the key. Take a look at `TrialDetails.tsx` for an example.

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <TabPane key="2" tab="Preview">
+        <div className={css.base} style={{ height }}>
+          <MarkdownViewer options={{ disableParsingRawHTML: true }}>
+            {markdown}
+          </MarkdownViewer>
+        </div>
+      </TabPane>
+    </Tabs> :
+      <div className={css.base} style={{ height, overflow: 'auto' }}>
+        <MarkdownViewer options={{ disableParsingRawHTML: true }}>
```
hkang1, line 45: nice, way to keep this HTML/JS injection safe

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+      </TabPane>
+      <TabPane key="2" tab="Preview">
+        <div className={css.base} style={{ height }}>
+          <MarkdownViewer options={{ disableParsingRawHTML: true }}>
+            {markdown}
+          </MarkdownViewer>
+        </div>
+      </TabPane>
+    </Tabs> :
+      <div className={css.base} style={{ height, overflow: 'auto' }}>
```
hkang1, line 44: **suggestion:** having written some interaction tests recently, one thing that came up that I think impacts the JSX design is that it might be better to have one parent div in the reusable component that remains consistent. Meaning, taking this setup for example, sometimes the top level is a `<Tab>`, and other times it's a generic `<div>` where the reference to the component element changes. I think that behavior is ok for pages, but for reusable components, the following pattern would be preferred (for testing purposes too).

```
return (
  <div aria-label="markdown-editor" className={css.base} style={...}>
    {editing ? (
      <Tab />
    ) : (
      <MarkdownViewer />
    )}
  </div>
);
```

The reason for why this is nice for interaction testing and accessibility is that it will always render a DOM element and when we start using more and more aria-labels for accessibility compatibility, it will be a single place to apply it along with styling. Even if the base element won't show up via CSS, the reference to the component is stable and reliable.

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
+import { Tabs } from 'antd';
+import { default as MarkdownViewer } from 'markdown-to-jsx';
+import React from 'react';
+
+import css from './Markdown.module.scss';
+import Spinner from './Spinner';
+
+const { TabPane } = Tabs;
+const MonacoEditor = React.lazy(() => import('components/MonacoEditor'));
```
hkang1, line 9: 👍

webui/react/src/components/Markdown.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface Props {
+  editing?: boolean;
+  height?: string;
+  markdown: string;
+  onChange?: (editedMarkdown: string) => void;
+}
+
+const Markdown: React.FC<Props> = ({ editing=false, height='80vh', markdown, onChange }: Props) => {
```
hkang1, line 18: **style suggestion:** To give `Markdown` a bit more flexibility, I think `height: '100%'` would be a bit better. `80vh` can potentially interfere with views where there are headers (or footers, tabs, other vertical space hogs) that are forced to being larger than `20vh` due to limited space.

hkang1 (commented):
This looks really good! IMHO, I think your code as really improved recently!

Wrote the usual. Nothing blocking, but enough items that probably warrant another review.

webui/react/src/components/Markdown.tsx
EmilyBonar: Unfortunately not, since the default export from `markdown-to-JSX` is also just named `Markdown`, it causes a name collision with the component.

hkang1 (approved):
All the changes look great!

PR #3038, opened 2021-10-04
fix: Make agent names unique for det deploy local agent-up

varlogtim:
[DET-5943]

Tested by deploying two servers in AWS, one was running the master and an agent, the other was running just an agent.

[DET-5943]: https://determinedai.atlassian.net/browse/DET-5943


ioga (approved):
lgtm

PR #3039, opened 2021-10-05
feat: job queue internal messaging

eecsliu:
## Checklist


hamidzr:
```
curl-da.sh "${DET_MASTER}/api/v1/resource-pools/queues?resource_pool=default" -s | jq '.jobs[] | .summary.jobId + .type + (.priority|tostring)'

```

```
curl-da.sh -X POST --data '{"updates": [{"job_id": "5ec94708-a54c-41ae-8391-503e7ebb45a9", "source_resource_pool": "default", "priority": 31}]}' "${DET_MASTER}/api/v1/resource-pools/queues" -s
```

found an issue with the api to submit change requests and updated it here
[eric.patch.txt](https://github.com/determined-ai/determined/files/7377163/eric.patch.txt) (needs to be applied before requests are parsed correctly)

the api call goes through but still returns an error


it changing the priority on a shell wipes it out of memory?

eecsliu:
yeah restoring from restart is coming soon

hamidzr:
> The latter will be fixed in this PR and the former will be addressed in a separate one.

could you add a ticket or an inline comment so we don't forget this?

PR #3049, opened 2021-10-06
fix: gracefully handle prestart agent failures

stoksc:
## Description
Fix a bug that could crash the resource manager when an agent disconnects before it sent the master it's `AgentStarted` message. In this case, the failure caused the master to move the agent to the draining state but, since it wasn't registered with the resource manager, this made the RM panic.

## Test Plan
- [x] retest these reconnect features
- [x] test the exact bug with a 5s sleep

## Commentary (optional)
oops

## Checklist


rb-determined-ai (approved):
Looks good.  Also I tested it, and it works.

PR #3051, opened 2021-10-06
ci: update gke version

rb-determined-ai:
1.18.* is no longer available.  Rather than doing the minimum possible
update, switch to what gke says is the default version of the "Regular
Channel".

https://cloud.google.com/kubernetes-engine/docs/release-notes


eecsliu (approved):
I like it

PR #3083, opened 2021-10-14
feat: job queue scheduler internals

eecsliu:
## Description
Blocked on #3039
This PR enables scheduler compatibility with the new job queue mechanics and brings the k8s implementation up to speed with the resource pool implementation. Changes include:
- adding internal messages and plumbing for kubernetes specific handling
- enable the priority scheduler to set initial queue positions for jobs
- allow resource pools to insert jobs in arbitrary queue positions
- allow k8s RM to change priority and queue position
- allow reading of job queue in k8s

Note: there is a known bug where changing queue position in k8s is unsuccessful.

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/kubernetes_resource_manager.go
+++ b/master/internal/resourcemanagers/kubernetes_resource_manager.go
@@ -205,9 +205,6 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error 
 	case sproto.SetGroupMaxSlots:
 		k.getOrCreateGroup(ctx, msg.Handler).maxSlots = msg.MaxSlots
 
-	case sproto.SetGroupWeight:
-		// SetGroupWeight is not supported by the Kubernetes RP.
-
 	case sproto.SetGroupPriority:
 		group := k.getOrCreateGroup(ctx, msg.Handler)
 		if msg.Priority != nil {
@@ -305,7 +302,7 @@ func (k *kubernetesResourceManager) receiveJobQueueMsg(ctx *actor.Context) error
 	//	resp := k.getJobSummary(msg.JobID)
 	//	ctx.Respond(resp)
 	case job.GetJobQStats:
-		ctx.Respond(*k.getJobQStats())
+		ctx.Respond(k.getJobQStats())
 	default:
 		return actor.ErrUnexpectedMessage(ctx)
 	}
@@ -326,6 +323,7 @@ func (k *kubernetesResourceManager) receiveJobQueueMsg(ctx *actor.Context) error
 //	return nil
 //}
 
+// TODO: consolidate these jq functions into the job queue
 func (k *kubernetesResourceManager) getJobQStats() *jobv1.QueueStats {
 	stats := &jobv1.QueueStats{}
 	reqs, _ := sortTasksWithPosition(k.reqList, k.groups)
--- a/master/internal/resourcemanagers/kubernetes_resource_manager.go
+++ b/master/internal/resourcemanagers/kubernetes_resource_manager.go
@@ -205,6 +205,9 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error 
 	case sproto.SetGroupMaxSlots:
 		k.getOrCreateGroup(ctx, msg.Handler).maxSlots = msg.MaxSlots
 
+	case sproto.SetGroupWeight:
+		// setting weights in kubernetes is not supported
+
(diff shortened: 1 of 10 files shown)


master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -253,31 +290,64 @@ func (k *kubernetesResourceManager) receiveJobQueueMsg(ctx *actor.Context) error
...
+//	jobs := mergeToJobs(requests, k.groups, kubernetesScheduler)
+//	for _, job := range jobs {
+//		if job.JobId == jobID.String() {
+//			return job.Summary
+//		}
+//	}
+//	return nil
+//}
+
+func (k *kubernetesResourceManager) getJobQStats() *jobv1.QueueStats {
```
eecsliu, line 329: inside the k8s resource manager, the functions used and arguments are a bit different, so I more or less redefined them here. I think consolidating them might be the right move, and I'll either add that in a different PR or update this PR in the next couple of days.

master/internal/sproto/kubernetes_resource_manager.go
```diff
@@ -20,6 +20,22 @@ type (
 		PodID container.ID
 	}
 
+	// PreemptTaskPod notifies the pods actor to preempt a pod.
+	PreemptTaskPod struct {
+		PodName string
+	}
+
+	// ChangePriority notifies the pods actor to preempt a pod
```
hamidzr, line 28: comment

master/internal/sproto/kubernetes_resource_manager.go
```diff
@@ -20,6 +20,22 @@ type (
...
+	ChangePriority struct {
+		PodID container.ID
```
hamidzr, line 30: question: does it need to specify a priority?

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -199,8 +205,39 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error
 	case sproto.SetGroupMaxSlots:
 		k.getOrCreateGroup(ctx, msg.Handler).maxSlots = msg.MaxSlots
 
-	case sproto.SetGroupWeight, sproto.SetGroupPriority:
-		// SetGroupWeight and SetGroupPriority are not supported by the Kubernetes RP.
+	case sproto.SetGroupWeight:
+		// SetGroupWeight is not supported by the Kubernetes RP.
```
hamidzr, line 209: is this temporary? if not we should return an error

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -253,31 +290,64 @@ func (k *kubernetesResourceManager) receiveJobQueueMsg(ctx *actor.Context) error
...
 					})
 				}
+				// do nothing if message is setting the weight
 			}
 		}
+	//case job.GetJobSummary:
+	//	resp := k.getJobSummary(msg.JobID)
+	//	ctx.Respond(resp)
+	case job.GetJobQStats:
+		ctx.Respond(*k.getJobQStats())
```
hamidzr, line 308: nit: if we're returning a ptr in the other place let's do the same here?

master/internal/resourcemanagers/kubernetes_resource_manager.go
hamidzr: maybe leave an inline todo as well so this doesn't get lost?

hamidzr (commented):
Overall looks good to me. 

- do you want to keep the commented out code in?
- The reported job order seems to flip when going from the job-queue branch to this branch for priority schedulers (agent) is this expected? The scheduler does process the jobs as advertised by the queue. 
image
- for the k8 changes I took a look but did not test it.

master/internal/sproto/kubernetes_resource_manager.go
eecsliu: nope! We just need to notify the pods actor to kill the pod. After that, we automatically re-submit under a new priority.

master/internal/resourcemanagers/kubernetes_resource_manager.go
eecsliu: returning an error causes further errors upstream - this would require a lot more changes and special handling for k8s cases.

hamidzr (commented):
This is looking good. I added some comments below and ran into a probably unrelated issue:

PR #3135, opened 2021-10-27
feat: create and persist jobs for experiments and NTbSC

hamidzr:
## Description

- add a job table
- assign job ids to experiments and NTbSC

## Test Plan

```
determined=# select job_id, id from experiments;
                job_id                | id
--------------------------------------+----
 backfilled-1                         |  1
 backfilled-2                         |  2
 518afb93-5a94-40f4-b03d-e99326e76730 |  3
(3 rows)

determined=# select * from jobs;
                job_id                |  job_type
--------------------------------------+------------
 backfilled-2                         | EXPERIMENT
 backfilled-1                         | EXPERIMENT
 518afb93-5a94-40f4-b03d-e99326e76730 | EXPERIMENT
(3 rows)

determined=#
```

## Checklist


master/internal/command/command.go
```diff
@@ -91,6 +94,8 @@ func (c *command) Receive(ctx *actor.Context) error {
 			TaskType:  c.taskType,
 			StartTime: c.registeredTime,
 		}); err != nil {
+			// TODO why do we persist the task when the underlying command isn't persisted?
```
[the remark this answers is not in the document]
hamidzr: sounds good. I'll take this comment out. let me know if you got any other feedback.

master/internal/db/postgres_experiments.go
```diff
@@ -1117,22 +1117,31 @@ FROM (
 }
 
 // AddExperiment adds the experiment to the database and sets its ID.
-func (db *PgDB) AddExperiment(experiment *model.Experiment) error {
+func (db *PgDB) AddExperiment(experiment *model.Experiment) (err error) {
```
stoksc, line 1120: nit: prefer to just not use named returns imo

master/internal/db/postgres_experiments.go
```diff
@@ -1117,22 +1117,31 @@ FROM (
...
-		:notes)
-RETURNING id`, experiment)
-	if err != nil {
-		return errors.Wrapf(err, "error inserting experiment %v", *experiment)
-	}
-	return nil
+	return db.withTransaction("add_experiment", func(tx *sqlx.Tx) error {
+		job := model.Job{
+			JobID:   experiment.JobID,
+			JobType: "EXPERIMENT",
```
stoksc, line 1127: use the constant here?

master/internal/db/postgres_experiments.go
```diff
@@ -1117,22 +1117,31 @@ FROM (
...
+		}
+		if err = addJob(tx, &job); err != nil {
+			return errors.Wrapf(err, "error inserting job %v", job)
+		}
+		err := namedGet(tx, &experiment.ID, `
+	INSERT INTO experiments
+	(state, config, model_definition, start_time, end_time, archived,
+	 git_remote, git_commit, git_committer, git_commit_date, owner_id, original_config, notes, job_id)
```
stoksc, line 1135: let's go ahead and add `job_id` to the selects too, so experiment queries don't return "" job_id's

master/internal/restore.go
```diff
@@ -170,6 +170,7 @@ func (m *Master) retrieveExperimentSnapshot(expModel *model.Experiment) ([]byte,
 	case err != nil:
 		return nil, errors.Wrap(err, "failed to retrieve experiment snapshot")
 	default:
+		// TODO snapshot restore
```
stoksc, line 173: ?

master/pkg/model/experiment.go
```diff
@@ -61,6 +63,12 @@ const (
 	TrialWorkloadSequencerType WorkloadSequencerType = "TRIAL_WORKLOAD_SEQUENCER"
 )
 
+// StateFromProto maps experimentv1.State to State.
+func StateFromProto(state experimentv1.State) State {
```
stoksc, line 67: was this used? also i prefer an exhaustive switch with panic as opposed to parsing something into state that isn't a valid enum (e.g. if proto gets STATE_WEIRD_STATE, since state is just a string we can parse WEIRD_STATE into it, which is invalid)

master/pkg/model/job.go
```diff
@@ -0,0 +1,40 @@
...
+	JobTypeTensorboard JobType = "TENSORBOARD"
+	// JobTypeExperiment is the "EXPERIMENT" job type for the enum.job_type in Postgres.
+	JobTypeExperiment JobType = "EXPERIMENT"
+)
+
+// Job is the model for a job in the database.
+type Job struct {
+	JobID   JobID   `db:"job_id"`
+	JobType JobType `db:"job_type"`
+	QPos    float64 `db:"q_position"`
```
stoksc, line 39: qpos

master/pkg/model/job.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/google/uuid"
+)
+
+// JobID is the unique ID of a job among all jobs.
+type JobID string
+
+func (id JobID) String() string {
+	return string(id) // FIXME does this add any value? is this a common interface in go?
```
stoksc, line 11: not really. `String() string` is the method to impl `Stringer`, so anything that would use Stringer if it exists (fmt.Sprintf/fmt.Printf internally) will call it, but since the type aliases string it has the same behavior anyway (but it you wanted it to print, say, "Job: :id" as it's default %s repr, then it could be useful

master/pkg/model/job.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// JobType is the type of a job.
+type JobType string
+
+// NewJobID returns a random, globally unique job ID.
+func NewJobID() JobID {
+	return JobID(uuid.New().String())
```
stoksc, line 19: note to self, new version of `github.com/google/uuid` has `NewString()`

master/static/migrations/20211028145058_jobs.up.sql
```diff
@@ -0,0 +1,31 @@
...
+    'TENSORBOARD'
+);
+
+CREATE TABLE public.jobs (
+    job_id text NOT NULL UNIQUE,
+    job_type job_type NOT NULL
+);
+
+
+BEGIN;
```
stoksc, line 15: there is no need as far as i can tell to nest a transaction here.

stoksc (commented):
one meta question, do we necessarily need a backfill of the jobs table? is it completely necessary, since these things are known to cause issues occassionally

master/internal/db/postgres_experiments.go
stoksc: specifically like ExperimentByID and friends

master/static/migrations/20211028145058_jobs.up.sql
hamidzr: is this because of the new "all migrations being treated as transactions" ? otherwise could it fails halfway? or would the go migrate take care of it somehow.

master/pkg/model/job.go
hamidzr: ops removed.

master/pkg/model/experiment.go
hamidzr: not here. and that makes sense it was just me being lazy..

hamidzr:
> one meta question, do we necessarily need a backfill of the jobs table? is it completely necessary, since these things are known to cause issues occassionally

good question.. I think we'd lose the non-null contraint on the column and would have to account for it in master for both reads and writes but otherwise it seems better to not backfill

master/internal/db/postgres_experiments.go
hamidzr: copilot in action:

image

stoksc: incredible. i'm scared, it's official

master/static/migrations/20211028145058_jobs.up.sql
stoksc: yeah, this is because of that, all migrations are transactions now, so this is just a nested transaction.

master/internal/db/postgres_experiments.go
hamidzr: I could also add it to our old endpoint db helpers eg the ones here https://github.com/determined-ai/determined/blob/41b20e6680fe7eb5fe9f5ec55cce8ee361f7f604/master/internal/core_experiment.go#L78 but do we keep updating those? or should we let them fall off

master/internal/db/postgres_experiments.go
```diff
@@ -1212,7 +1212,7 @@ func (db *PgDB) ExperimentByTrialID(id int) (*model.Experiment, error) {
 	experiment := model.Experiment{}
 	return &experiment, db.sql.QueryRowx(`
 SELECT e.id, e.state, e.config, e.model_definition, e.start_time, e.end_time,
-e.archived, e.git_remote, e.git_commit, e.git_committer, e.git_commit_date
```
stoksc, line 1206: because of the duplication in these get experiment queries, and because we had a bug essentially here, not including owner id, i think i can get down with @ioga 's thought in #infra that we should consider an orm

stoksc: using an orm would disallow this

stoksc (approved):

master/static/migrations/20211028145058_jobs.up.sql
stoksc: i approved but, i think we still need to remove this.

master/static/migrations/20211028145058_jobs.up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+INSERT INTO jobs (
+    SELECT 'backfilled-' || id as job_id, 'EXPERIMENT' as job_type FROM experiments
+);
+
+ALTER TABLE public.experiments
```
stoksc, line 21: i think this is fine, since the experiments table is probably small enough (<100k rows for sure)

master/internal/db/postgres_experiments.go
hamidzr: 👍🏼  I don't know if we have notes on this topic or not IIRC we had talked about it before. It'd help to see where that got stuck.

master/internal/db/postgres_experiments.go
hamidzr: e.name is missing from a bunch of these as well. I'll avoid adding them to avoid adding more risk.

master/static/migrations/migration-move-to-top.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
hamidzr, line 1: I ended up doing something like this a few times to keep up with changes in master. I think it might help to have it in the repo.

master/static/migrations/20211028145058_jobs.up.sql
hamidzr: removed

master/static/migrations/migration-move-to-top.sh
stoksc: fine with me

stoksc (approved):
lgtm. i did a quick grep for `FROM experiments` and found all the same places to put job id that you did.

PR #3136, opened 2021-10-27
feat: add `det deploy gke-experimental` [DET-5752]

ioga:
## Description

I've inherited #2842 and did a minimal amount of changed and testing to make it mergeable. Compared to the original PR,

- rename to `gke-experimental` to better highlight the nature of the tool.
- fixed a number of lint issues.
- added auto-naming for node pools and buckets (because providing these by hand is unnecessary and confusing).
- some extra flag validation to avoid basic failures.
- make `down` operation actually work.
- destroy GCS bucket on `down`.

## Test Plan

Basic:
1. `det deploy gke-experimental up --cluster-id <test-cluster-name>` (while in repo root, or pass `--helm-dir`)
2. kubectl get service determined-master-service-determined-gke -n default
3. export DET_MASTER=<ip address>:8080
4. Use it.
5. `det deploy gke-experimental down --cluster-id <test-cluster-name>`

There's a number of configuration flags. Use at your own risk.

NB: running `up` again on an existing cluster fails. there's no updates or upgrades.

## Checklist

### DIFF
--- a/harness/determined/deploy/gke/cli.py
+++ b/harness/determined/deploy/gke/cli.py
@@ -1,0 +1,539 @@
+import argparse
+import json
+import subprocess
+import sys
+from pathlib import Path
+from typing import Any, Dict, Optional, Union, cast
+
+from termcolor import cprint
+
+import determined
+from determined.common import yaml
+from determined.common.declarative_argparse import Arg, ArgGroup, Cmd
+from determined.common.util import safe_load_yaml_with_exceptions
+from determined.deploy.gke.constants import defaults
+
+
+def make_spec(task_container_defaults: Dict[str, Any], key: str) -> Dict[str, Union[Dict, str]]:
+    pod_spec = task_container_defaults.get(key)  # type: Optional[Dict]
+    if not pod_spec:
+        pod_spec = {"apiVersion": "v1", "kind": "Pod"}
+    pod_spec = cast(Dict[str, Union[Dict, str]], pod_spec)
+    if not pod_spec.get("spec"):
+        pod_spec["spec"] = {}
+    return pod_spec
+
+
+def validate_accelerator_type(s: str) -> None:
+    json_value = subprocess.check_output(
+        ["gcloud", "compute", "accelerator-types", "list", "--format=json(name)"]
+    )
+    json_names = json.loads(json_value)
+    valid_accelerator_types = {accelerator["name"] for accelerator in json_names}
+
+    if s not in valid_accelerator_types:
+        raise ValueError("Accelerator must be one of {}".format(valid_accelerator_types))
+
+
+def validate_location(location: str, isZone: bool = True) -> None:
+    try:
+        cmd = ["gcloud", "compute"]
+        if isZone:
+            cmd += ["zones"]
+        else:
+            cmd += ["regions"]
+        cmd += ["describe", location]
+        subprocess.check_call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+    except subprocess.CalledProcessError:
+        raise ValueError(
+            "The specified {} {} was not found".format("zone" if isZone else "region", location)
+        )
+
+
+def validate_machine_type(machine_type: str, zone: str) -> None:
+    try:
+        subprocess.check_call(
+            [
+                "gcloud",
+                "compute",
+                "machine-types",
+                "describe",
+                machine_type,
+                "--zone",
+                zone,
+            ],
+            stdout=subprocess.DEVNULL,
+        )
+    except subprocess.CalledProcessError:
+        raise ValueError(
+            "The specified machine type {} was not found in the specified zone {}.".format(
+                machine_type, zone
+            ),
+        )
+
+
+def validate_accelerator_zone(args: argparse.Namespace, zone: str) -> None:
+    try:
+        subprocess.check_call(
+            [
+                "gcloud",
+                "compute",
+                "accelerator-types",
+                "describe",
+                args.gpu_type,
+                "--zone",
+                zone,
+            ],
+            stdout=subprocess.DEVNULL,
+        )
+    except subprocess.CalledProcessError:
+        raise ValueError(
+            "The specified accelerator type {} was not found in the specified zone {}.".format(
+                args.gpu_type, zone
+            ),
+        )
+
+
+def validate_args(args: argparse.Namespace) -> None:
+    validate_location(args.zone, isZone=True)
+    validate_accelerator_type(args.gpu_type)
+
+    if args.master_machine_type != "n1-standard-16":
+        validate_machine_type(args.master_machine_type, args.zone)
+
+    if args.agent_machine_type != "n1-standard-32":
+        validate_machine_type(args.agent_machine_type, args.zone)
+
+    validate_accelerator_zone(args, args.zone)
+
+    if args.gpu_coscheduler and args.preemption:
+        raise ValueError(
+            "--gpu-coscheduler and --preemptive-scheduler are mutually exclusive and cannot both be"
+            " specified",
+        )
+
+    if args.gcs_bucket_name is None:
+        args.gcs_bucket_name = args.cluster_id + "-checkpoints"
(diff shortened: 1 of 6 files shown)


harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+            ),
+            "yellow",
+        )
+        exp_config = {
+            "environment": {
+                "podspec": {
+                    "metadata": {
+                        "labels": {
+                            "pod-group.scheduling.sigs.k8s.io/name": "<unique task name>",
+                            "pod-group.scheduling.sigs.k8s.io/min-available": "<# of Nodes required>",  # noqa: E501
```
mackrorysd, line 345: This doesn't seem right that when referencing a CPU agent pool we're talking about a number of nodes required, and we're not showing how to set the priority of a task when telling people how to use the priority scheduler. Am I missing something?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+        args.zone,
+        "--num-nodes=1",
+        "--machine-type={}".format(args.master_machine_type),
+    ]
+    subprocess.check_call(cmd, stdout=subprocess.DEVNULL)
+
+    create_nodepools(region, args)
+
+    if not args.no_manage_bucket:
+        cmd = ["gsutil", "mb", "gs://{}".format(args.gcs_bucket_name)]
```
mackrorysd, line 156: Would be nice to see if gsutil, gcloud, etc. is on the PATH and fail with a helpful message if it's missing.

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+                            help="maximum number of nodes for the CPU node group",
+                        ),
+                        Arg(
+                            "--cpu-node-pool-name",
+                            type=str,
+                            default=None,
+                            help="a unique name for the GPU node pool",
+                        ),
+                        Arg(
+                            "--multi-node-pool",
```
mackrorysd, line 509: --multi-node-pool sounds like it's referring to a pool with >1 nodes, not multiple node pools. Would suggest --multiple-node-pools. I suppose the reason to not just do both like we do for other platforms is because auto-scaling can't work with our recommended scheduler?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+                    None,
+                    [
+                        Arg(
+                            "--region",
+                            type=str,
+                            default="us-west1",
+                            help="region containing cluster to delete",
+                        ),
+                        Arg(
+                            "--no-manage-bucket",
```
mackrorysd, line 579: Should be 'no-managed-bucket' with the 'd', no?

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+
+    if args.master_machine_type != "n1-standard-16":
+        validate_machine_type(args.master_machine_type, args.zone)
+
+    if args.agent_machine_type != "n1-standard-32":
+        validate_machine_type(args.agent_machine_type, args.zone)
+
+    validate_accelerator_zone(args, args.zone)
+
+    if args.gpu_coscheduler and args.preemption:
```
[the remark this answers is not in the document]
ioga: This predates me but my understanding is that the goal was to allow configuring coscheduler on GPU pool only while having two pools.

Technically we have distributed CPU jobs, although I doubt there're any users on k8s using it.

harness/determined/deploy/gke/cli.py
```diff
@@ -0,0 +1,596 @@
...
+        raise ValueError(
+            "--gpu-coscheduler and --preemptive-scheduler are mutually exclusive and cannot both be"
+            " specified",
+        )
+
+    if args.multi_node_pool and args.cpu_coscheduler and not args.gpu_coscheduler:
```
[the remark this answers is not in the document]
ioga: IIRC if you've got 2 pools, and you want a coscheduler on CPUs, you've got to have it on GPUs as well or weird things happen.

harness/determined/deploy/gke/cli.py
ioga: I think it'd already give a somewhat understandable message: `FileNotFoundError: [Errno 2] No such file or directory: 'gsutil'`
Or do you want something better?

harness/determined/deploy/gke/cli.py
ioga: That really predates me. And honestly I haven't ever heard about CPU coscheduling setups working. I could cut this entire section if it's not useful. 
I'd say the original point of this tool was to avoid writing 8 command lines necessary to make a cluster. Not to try to make it produce all possible working configurations for everyone easily. I'd rather remove the guardrails to allow configurability.

harness/determined/deploy/gke/cli.py
mackrorysd: That works. Failing early before we've attempted ANY steps (and perhaps created a partial deployment) is a nice-to-have. It doesn't need to block the PR though.

harness/determined/deploy/gke/cli.py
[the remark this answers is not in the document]
mackrorysd: Ah - yeah I always interpret the "CPU training" work as meaning something other than what it means :smile:

harness/determined/deploy/gke/cli.py
mackrorysd: @eecsliu probably has the best context on this. Perhaps you could enlighten us on this issue and the ones I've mentioned above so we don't have to relearn anything the hard way? :smile:

eecsliu: yeah CPU training with coscheduling is broken because it requires a number of nodes, which you currently have to manually calculate. If you don't set this `pod-group.scheduling.sigs.k8s.io/min-available` field, the coscheduler will ignore it. However, it then becomes no better than the default scheduler.

harness/determined/deploy/gke/cli.py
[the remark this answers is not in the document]
eecsliu: A little late to this party, but schedulers in k8s are global, and not per resource pool. So while it is possible to run multiple schedulers in a cluster, the current state of determined (since k8s doesn't have resource pools yet) is that we can only easily use one type of scheduler at a time.

harness/determined/deploy/gke/cli.py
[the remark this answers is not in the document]
mackrorysd: The split is separate flags to enable the scheduler for the GPU / CPU resource pools.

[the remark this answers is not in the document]
eecsliu: hmm well it makes sense to check for whether the coscheduler and the preemption scheduler are both set (since only one should be set at a time), but it is not currently possible to run a CPU pool alongside a GPU pool.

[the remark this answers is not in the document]
ioga: I believe the case which is handled here is that you may have slotted jobs running on GPUs with coscheduler; and then there's CPU-only nodes without coscheduler for zero-slot tasks; and there's some amount of code to setup taints/tolerations here to ensure that stuff is getting scheduled that way.

If you also want coscheduler on CPU's, this code will also force you to have coscheduler on GPUs as well. So that case is covered.

I could rip it out if you think it doesn't make sense. 

But in this tool I'd try to lean towards being more flexible than having guardrails and only allowing safe and sound setups. This thing is not recommended for production really. I see it as a development tool. So that if you're testing something, you could setup a configuration using some combination of flags, and give that combination of flags to other people - which is better than having to do all this setup manually (where you've got to run 5+ pretty long command lines).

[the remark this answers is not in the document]
ioga: alright, I've ripped it out and rebased the whole branch. PTAL.

mackrorysd (approved):
Thanks for all the iteration on this. LGTM if we can get it to not fall over on Rehan's credit.

PR #3138, opened 2021-10-28
feat: Add DELETE to Model and Model Version API [DET-6118]

mapmeld:
Adds endpoints to delete a model or modelversion from the model registry. Confirm that this user created / owns the model before deletion.

## Checklist


proto/src/determined/api/v1/model.proto
```diff
@@ -101,8 +101,10 @@ message PostModelResponse {
 
 // Request for updating a model in the registry.
 message PatchModelRequest {
+  // The id of the model being updated.
+  int32 model_id = 1;
```
ioga, line 105: is this just to make the documentation look "good"?

proto/src/determined/api/v1/model.proto
```diff
@@ -218,8 +220,10 @@ message PostModelVersionResponse {
 message PatchModelVersionRequest {
   // The id of the model being updated.
   int32 model_id = 1;
+  // The id of the model version being updated.
+  int32 model_version_id = 2;
   // The model version being updated.
-  determined.model.v1.PatchModelVersion model_version = 2;
+  determined.model.v1.PatchModelVersion model_version = 3;
```
ioga, line 226: you should keep the proto tag number the same as before when you "remove" a field or add new ones.
e.g. `model_version_id` here could be `= 3` or `= 30`, it doesn't have to be sorted in the code, but it has to be stable.

https://developers.google.com/protocol-buffers/docs/proto3#updating

it probably won't break anything for us because these messages are only used in json-serialized form, but I think we should stick to the guidelines anyway.

proto/src/determined/model/v1/model.proto
```diff
@@ -51,22 +51,17 @@ message Model {
...
 // PatchModel is a partial update to a model with only id required
 message PatchModel {
-  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
-    json_schema: { required: [ "id" ] }
-  };
-  // The id of this model.
-  int32 id = 1;
   // An updated name for the model.
-  google.protobuf.StringValue name = 2
+  google.protobuf.StringValue name = 1
```
ioga, line 55: same comment as above applies to the changes in this file as well.

master/static/srv/delete_model.sql
```diff
@@ -1,13 +1,23 @@
-WITH m AS (
-  SELECT * FROM models
+WITH myuser AS (
```
ioga, line 1: optional: I believe `a.CurrentUser` already gives you a user instance (so you know id and `admin` on go side).
so potentially you don't need this extra `WITH clause`, and just have `WHERE (user_id = $2 OR $3 IS TRUE)`, where $2 is current_user.id and $3 is current_user.admin.

proto/src/determined/api/v1/model.proto
mapmeld: I will go un-change the proto numbering so we aren't confusing previous versions of these requests.

We do actually need model_id here and in PatchModelVersionRequest because (my current understanding) that's where we get the input from the `/models/{model_id}` url into PatchModelRequest, and not in a subfield of the Model

PR #3140, opened 2021-10-28
chore: limit concurrent ssh keygen 

stoksc:
## Description
Exactly as the msg says.

## Test Plan
- [x] test should pass

## Commentary (optional)
this limit is a little conservative, but mostly we just want to allow the master to make progress still if a large hp search generates tons of keys concurrent and the limit is high.

## Checklist

### DIFF
--- a/master/pkg/ssh/ssh.go
+++ b/master/pkg/ssh/ssh.go
@@ -5,6 +5,8 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"math"
+	"runtime"
 
 	"github.com/pkg/errors"
 	sshlib "golang.org/x/crypto/ssh"
@@ -21,8 +23,20 @@ type PrivateAndPublicKeys struct {
 	PublicKey  []byte
 }
 
+var (
+	// For each 4 cores, allow another concurrent call.
+	maxConcurrentKeyGenCalls = int(math.Ceil(float64(runtime.NumCPU()) / 4))
+	// keyGenSemaphore limits the number of concurrent calls to GenerateKey
+	// since it can be fairly costly and lock the master. Callers will see this
+	// as calls just taking longer.
+	keyGenSemaphore = make(chan struct{}, maxConcurrentKeyGenCalls)
+)
+
 // GenerateKey returns a private and public SSH key.
 func GenerateKey(passphrase *string) (PrivateAndPublicKeys, error) {
+	keyGenSemaphore <- struct{}{}
+	defer func() { <-keyGenSemaphore }()
+
 	var generatedKeys PrivateAndPublicKeys
 	privateKey, err := rsa.GenerateKey(rand.Reader, trialKeySize)
 	if err != nil {


stoksc:
so @rb-determined-ai , for that case:
* on determined agents, this would start fairly quickly
* on k8s it would take a while since all the keys would be generated but it wouldn't lock the master
essentially, this is unsolvable/very hard to solve within the actor framework; we need to thread a `context.Context` through with the `actor.Context` and have actors be aware of it/correctly decide what to do when handling a given message is preempted by a recv on `ctx.Done`.

pretty much the same as it would be sans actors but with the mental burden of thinking about what interrupting a message means semantically for each actor (unless only certain actors impl some "cancelable" interface which allows all asks/recv to take a ctx, but then you have to grok what that may mean in each switch in a receive, even those where you're not expecting... which is half the problem with actors, things that only make sense in somes contexts seemingly can happen and need to be handled in all unless you totally grok the intra-actor protocols, which are usually undocumented).

rb-determined-ai:
I don't think the right solution is that difficult.  Conceptually (actor system or not), you want a pool of workers pumping out keys at a limitable rate.  And you want the key requesters to not block (since it could be _hours_ of blocking), and you want key requests to be cancelable (again, it could be hours).

I don't see how doing this in an actor system changes any of those requirements.

Possible implementations:
* have a singleton keygen actor:
  * You can `Tell(KeyRequest), and you can `Tell(CancelKeyRequest)` to that actor.  
  * That actor will `Tell(KeyGenerated)` back to you at some point in time.
  * In the Trial actor, you would send the KeyRequest right away and you'd add a check in `maybeAllocateTask` to not allocate before you get the KeyGenerated response.  
  * You'd also send a CancelKeyRequest in PostStop if you had an outstanding KeyRequest.
* have a type KeyGen interface:
  ```go
  type KeyGen interface {
    RequestKey() -> chan Key
  }
  ```
  * you select on that channel if you want to wait for the key, or you just close the channel if you change your mind (the key generator would finish generating your key, try to write it to you, but it would see you closed the channel and would just pass the key to the next requester)
  * to interface with the actor system, the trial could start a goroutine that hijacks some the trial actor context to be able to send itself another message later (same as how the timer actor behind NotifyAfter() hijacks its prestart actor context)
  * after we grow out of our actor system, we would just remove the goroutine layer in front of the interface and use the interface directly.

But overall, the reason we're in this pickle is that we chose a quick-easy hack a long time ago and now it's coming back to bite us.  We are trying not to do that again, which is exactly why we think that limiting concurrent keygen is important _even if_ we do key-per-user.  So if we are striving for doing the right thing, I don't see why we tolerate a quick solution where there may be Asks in the system that might block for hours.

stoksc:
> But overall, the reason we're in this pickle is that we chose a quick-easy hack a long time ago and now it's coming back to bite us. 

Even if we had chosen the other solution way back, we would still be susceptible to this with k8s. But besides that

> I don't see how doing this in an actor system changes any of those requirements.

I think the 2 tells pattern vs just using a cancel-able context is exactly how the actor system changes things; suddenly it's another reentrant async block were we store instruction pointers manually. This is sort of the exact pattern that, in excess, can make the actor system hard to grok.

The 2nd way you suggested is something I've been leaning to for a while as a way to transition from actors (I mentioned it to dzhu just today). Let me get that a hack, if you insist. In the meantime, do you agree this is strictly better than locking the master up?

rb-determined-ai:
> This is the exact pattern that, in excess, can make the actor system hard to grok.

Yes, I agree.  My point was that if you want to limit this computationally expensive thing, but you don't want your event-based software to deadlock ("datalock"?) then you have introduce some complexity somewhere, and there's just no way around that.

> In the meantime, do you agree this is strictly better than locking the master up (this is still debatable.. i could see just killing the master being potentially preferable)? I think this is a case where this solution is strictly better (than the status quo) and took me 5m and the other potentially takes 2-3 days.

But the 5m solution just kicks the problem down the road, for some future Ryan and Bradley who are working on some future project and have to stop immediately and work long hours before a release because some important customer is on fire.  That is why I don't like leaving known issues in the system.  So 2-3 days now for the bulletproof solution seems pretty cheap to me.

PR #3143, opened 2021-10-29
fix: non-scalar nan handling

rb-determined-ai:
As a quick release-party fix, add a post-processing step to catch NaN's in arrays.

This means we do a pre-processing step and a post-processing step.  This is less code (better for a release party fix) but the long-term fix will involve replacing simplejson.dumps, which doesn't allow us to customize handling of python-native types.

### DIFF
--- a/harness/determined/_generic/_searcher.py
+++ b/harness/determined/_generic/_searcher.py
@@ -5,7 +5,6 @@ from typing import Iterator, Optional
 
 import determined as det
 from determined.common.experimental.session import Session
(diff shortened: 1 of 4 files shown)


mapmeld (approved):
Looks good, I think this will work well with the inf/nan values test :+1:

rb-determined-ai:
The `test_efficientdet_coco_pytorch_const` has passed, so in favor of not making people stay up all night cutting a release, I'm landing this now.

PR #3144, opened 2021-10-29
fix: fail non-restartable tasks on allocation termination [DET-6158]

ioga:
## Description

Currently, when an allocation is terminated for agent-related reasons (e.g. agent failing), it's not counted as failure, and as a result the master treats the non-trial tasks (e.g. commands) as if they exited successfully. 

- We should count these as failures.
- _Not counting_ these against trial `max_restarts` is moved to trial.
- Update devcluster to latest version, and add a new test option for a "managed devcluster", where we use devcluster API to spawn and programmatically control it to kill/restart agents as needed for the test purposes.

## Test Plan

Added managed-devcluster tests for
 - making sure commands die without reporting success on agent kill.
 - checking that max_restarts=0 experiments still restart in that setup.

## Commentary (optional)

This managed setup will get used more extensively in the work related to recovering containers after agent restarts. Also we could potentially move the entirety of tests currently using devcluster (i.e. most e2e setups) to this managed setup.

This PR includes a (thin) wrapper `ManagedCluster`. I expect it to include various common behaviors that'd otherwise be helper functions. 

Additionally I'd like to implement an ability to *not use* managed cluster for managed setup, and connect to an externally running devcluster instead via a client. The motivation for that is to allow running a test against a local development cluster you're currently running "in another terminal".

One more thing: this setup has a hard limit of one devcluster per node because of a) devcluster file lock b) binding master to local port, reusing same fluent ports, etc. When that's no longer a limit, we could have this managed devcluster look more like a true pytest fixture, allowing us to get rid of existing manual split between e2e-cpu / e2e-cpu-double / e2e-managed-devcluster, and do them all in one circleci job with parallelism.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_agent_restart.py
+++ b/e2e_tests/tests/cluster/test_agent_restart.py
@@ -1,0 +1,226 @@
+import json
+import os
+import subprocess
+import time
+from typing import Any, Dict, Iterator, List, Tuple, cast
+
+import pytest
+
+from tests import config as conf
+from tests import experiment as exp
+
+from .utils import get_command_info, run_command, wait_for_command_state
+
+DEVCLUSTER_CONFIG_PATH = conf.PROJECT_ROOT_PATH.joinpath(
+    ".circleci/devcluster/double.devcluster.yaml"
+)
+
+
+def _get_agent_data(master_url: str) -> List[Dict[str, Any]]:
+    command = ["det", "-m", master_url, "agent", "list", "--json"]
+    output = subprocess.check_output(command).decode()
+    agent_data = cast(List[Dict[str, Any]], json.loads(output))
+    return agent_data
+
+
+class ManagedCluster:
+    # This utility wrapper uses double agent yaml configurations,
+    # but provides helpers to run/kill a single agent setup.
+
+    def __init__(self) -> None:
+        # Strategically only import devcluster on demand to avoid having it as a hard dependency.
+        from devcluster import Devcluster
+
+        self.dc = Devcluster(config=str(DEVCLUSTER_CONFIG_PATH))
+        self.master_url = conf.make_master_url()
+
+    def __enter__(self) -> "ManagedCluster":
+        self.old_cd = os.getcwd()
+        os.chdir(str(conf.PROJECT_ROOT_PATH))
+        self.dc.__enter__()
+        return self
+
+    def __exit__(self, *_: Any) -> None:
+        os.chdir(self.old_cd)
+        self.dc.__exit__(*_)
+
+    def initial_startup(self) -> None:
+        self.dc.set_target("agent1", wait=True, timeout=3 * 60)
+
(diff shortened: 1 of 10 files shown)


master/internal/task/allocation.go
```diff
@@ -609,12 +609,12 @@ func (a *Allocation) terminated(ctx *actor.Context) {
 				exit.Err = err
 				return
 			case aproto.AgentError, aproto.AgentFailed:
-				// Questionable, could be considered failures, but for now we don't.
 				ctx.Log().WithError(err).Warnf("allocation exited due to agent (%s)", err.FailureType)
```
stoksc, line 612: i could appreciate condensing this into a pprintable error type

master/pkg/agent/exit.go
```diff
@@ -82,3 +82,25 @@ const (
 	// AgentError denotes that the agent failed to launch the container.
 	AgentError = FailureType("agent failed to launch the container")
 )
+
+// IsRestartableSystemError checks if the error is caused by the system and
```
stoksc, line 86: all this should eventually not really be here since k8s uses these error types too (fine for now, just a note)

stoksc (commented):
code looks great, i'll review the tests monday morning

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def restart_agent(self) -> None:
+        agent_data = _get_agent_data(self.master_url)
+        if len(agent_data) == 1 and agent_data[0]["enabled"]:
+            return
+
+        # Currently, we've got to wait for master to "forget" the agent before reconnecting.
+        WAIT_FOR_AMNESIA = 60
+        for _i in range(WAIT_FOR_AMNESIA):
+            agent_data = _get_agent_data(self.master_url)
```
stoksc, line 70: it's a bit awkward to read this when it uses double then only expects to have and lose one agent when it uses `double.devcluster.yaml`, but then i noticed we only ever go to stage `agent1`.  just a comment in init or something could help.

stoksc (approved):
killer

stoksc (approved):

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -0,0 +1,222 @@
...
+
+        exp.wait_for_experiment_state(exp_id, "COMPLETED")
+
+
+@pytest.mark.managed_devcluster
+def test_agent_restart_cmd_container_failure(managed_cluster: ManagedCluster) -> None:
+    # Launch a cmd, kill agent, wait for reconnect timeout, check it's not marked as success.
+    managed_cluster.ensure_agent_ok()
+    try:
+        command_id = run_command(60)
```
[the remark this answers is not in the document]
stoksc: maybe drop a comment to enumerate these cases? (if anything for my education)

PR #3148, opened 2021-11-01
chore: remove simplejson

rb-determined-ai:
simplejson is just externally-maintained json.  We don't have any
particular need to bring in an external dependency over what's packaged
in python's standard library, so just use the built-in json module.

Additionally, stop using json's "default" argument for encoding custom
types.  The function passed to "default" argument is not consulted when
encoding default types (especially "float") so it is impossible to catch
NaNs.  As a result, when using "default", we had to pre-and-post-process
the object during encoding.  Now, we do our own pre-processing step,
where we explicitly visit every object so we can eliminate the
post-processing step.

### DIFF
--- a/harness/determined/util.py
+++ b/harness/determined/util.py
@@ -2,7 +2,9 @@ import collections
 import datetime
 import enum
 import inspect
+import json
 import math
+import numbers
 import os
 import pathlib
 import random
@@ -10,9 +12,7 @@ import shutil
 import time
 import uuid
 import warnings
-from typing import Any, Callable, Dict, List, Optional, Set, TypeVar, cast
-
-import simplejson
+from typing import Any, Callable, Dict, List, Optional, Set, SupportsFloat, TypeVar, cast
 
 from determined import constants
 from determined.common import check, util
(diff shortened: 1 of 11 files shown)


harness/determined/util.py
```diff
@@ -129,48 +129,51 @@ def make_metrics(num_inputs: Optional[int], batch_metrics: List[Dict[str, Any]])
...
+            obj = cast(SupportsFloat, obj)
+            # float, np.float64, etc.  Serialize nan/±infinity as strings.
+            if math.isfinite(obj):
+                return float(obj)
+            if math.isnan(obj):
+                return "NaN"
+            if obj > 0:
+                return "Infinity"
+            else:
+                return "-Infinity"
```
ioga, line 155: nitpicking: I'd have "infinity" section to look like it's literally checking that the code is dealing with infinity. it's a little weird to have infinity as a last possible option.

```suggestion
            if math.isnan(obj):
                return "NaN"
            if not math.isfinite(obj):
                if obj > 0:
                    return "Infinity"
                else:
                    return "-Infinity"
            return float(obj)
```

ioga (approved):

harness/determined/util.py
rb-determined-ai: Stylistically I agree with you.  I was thinking that if this ever had to deal with large arrays, I wanted to pass through the least possible volume of python code.  But probably the difference in performance is negligible.

mapmeld (approved):

PR #3150, opened 2021-11-01
feat: Add /labels to Model Registry API [DET-6178]

mapmeld:
## Description

Adds a new endpoint `/model/labels` to the Model Registry, listing all labels used in any model, in order of popularity.

We SELECT the model id and order by COUNT(DISTINCT(id)) to prevent duplicate labels on a model counting more than once. Labels with the same number of model occurrences are secondarily ordered alphabetically. 

Adds a call to the model registry's e2e test

## Checklist

### DIFF
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -195,3 +195,11 @@ class Determined:
 
         models = r.json().get("models")
         return [model.Model.from_json(m, self._session) for m in models]
+
+    def get_model_labels(self) -> List[str]:
+        """
+        Get a list of labels used on any models, sorted from most-popular to least-popular.
+        """
+        r = self._session.get("/api/v1/model/labels")
+
+        return r.json().get("labels")
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -202,4 +202,5 @@ class Determined:
         """
         r = self._session.get("/api/v1/model/labels")
 
-        return r.json().get("labels")
+        labels = r.json().get("labels")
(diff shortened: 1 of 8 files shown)


ioga (approved):

proto/src/determined/api/v1/model.proto
```diff
@@ -76,6 +76,15 @@ message GetModelsResponse {
   Pagination pagination = 2;
 }
 
+// Get a list of model labels.
+message GetModelLabelsRequest {}
+
+// Response to GetModelLabelsRequest.
+message GetModelLabelsResponse {
+  // List of labels used across all models.
+  repeated string labels = 1;
```
ioga, line 85: The code looks good, but I am mildly concerned about extensibility of this. 
e.g. right now it's a list of plain strings, but I can see a case for returning a list of `[{'label': <label name>, 'count': <number of objects with the label>}]`, e.g. for faceted search UX. 
It'd be a breaking change if you go from list of strings -> list of objects, but won't be a problem if you start with objects.

harness/determined/common/experimental/determined.py
```diff
@@ -202,4 +202,5 @@ def get_model_labels(self) -> List[str]:
         """
         r = self._session.get("/api/v1/model/labels")
 
-        return r.json().get("labels")
+        labels = r.json().get("labels")
+        return [str(label) for label in labels]
```
ioga, line 206: since you already expect it to be a list of strings, you probably could do `return cast(List[str], labels)` which should be enough to satisfy the type checker.

mapmeld: 👍

PR #3152, opened 2021-11-02
fix: Continue using model names on the CLI [DET-6152]

mapmeld:
## Description

Restores references to model names from the command line to be compatible with our shift to models.id on the backend. This un-does our quick fix of #3134 . Restores a unique constraint so an instance will not have models with duplicate names.

DET-6152 ticket should stay active, as I understand it to mean we would add more model registry actions to the CLI.

## Checklist

### DIFF
--- a/docs/release-notes/3152-model-registry-cli.txt
+++ b/docs/release-notes/3152-model-registry-cli.txt
@@ -4,3 +4,7 @@
 
 -  Revert changes to Model Registry CLI format so models can be referred to by a unique name, and
    not only ID
+
+**Potential for Breaking Change**
+
+-  If multiple models were created with the same name in the registry, the names will change.
--- a/docs/release-notes/3152-model-registry-cli.txt
+++ b/docs/release-notes/3152-model-registry-cli.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Improvement**
+
+-  Revert changes to Model Registry CLI format so models can be referred to by a unique name, and
+   not only ID
--- a/harness/determined/cli/model.py
+++ b/harness/determined/cli/model.py
@@ -213,7 +213,7 @@ args_description = [
                     Arg(
                         "--version",
                         type=int,
-                        default=0,
+                        default=-1,
                         help="model version information to include in output",
                     ),
                 ],
--- a/harness/determined/cli/model.py
+++ b/harness/determined/cli/model.py
@@ -78,16 +78,23 @@ def list_models(args: Namespace) -> None:
 
         render.tabulate_or_csv(headers, values, False)
 
+def model_by_name(args: Namespace) -> Model:
+    models = Determined(args.master, None).get_models(name=args.name)
+    if len(models) == 0:
+        raise Error("No model was found with the given name.")
+    if len(models) > 1:
+        raise Error("Multiple models were found with the given name.")
+    return models[0]
 
 @authentication.required
 def list_versions(args: Namespace) -> None:
+    model = model_by_name(args)
     if args.json:
-        r = api.get(args.master, "models/{}/versions".format(args.id))
+        r = api.get(args.master, "models/{}/versions".format(model.model_id))
         data = r.json()
         print(json.dumps(data, indent=2))
 
     else:
-        model = Determined(args.master).get_model(args.id)
         render_model(model)
         print("\n")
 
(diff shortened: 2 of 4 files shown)


master/static/migrations/20211102110152_make_model_registry_names_unique.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE public.models ADD CONSTRAINT uni_name UNIQUE (name);
```
rb-determined-ai, line 1: This seems pretty unsafe to me now.

I'm concerned that we shipped a release where we let users violate this and now the cat is out-of-the-bag, so-to-speak.

harness/determined/cli/model.py
```diff
@@ -79,15 +79,24 @@ def list_models(args: Namespace) -> None:
         render.tabulate_or_csv(headers, values, False)
 
 
+def model_by_name(args: Namespace) -> Model:
```
rb-determined-ai, line 82: should this logic be on the master side?

master/static/migrations/20211102110152_make_model_registry_names_unique.up.sql
stoksc: it's safe because of the dml above, but i'm not sure that is OK without a much bigger warning that we may be renaming things without warning.

also, constraint names are global I think, so name this something like `models_name_unique` (the norm is `{table_name}_{column_name}_unique`.

stoksc (commented):
i'd also much prefer a query param to `(GET /api/v1/models).filter(x => x.name = queried_name)`. maybe just add `?name=:name` and pass it down to the sql, likely would be useful for webui, too.

master/static/migrations/20211102110152_make_model_registry_names_unique.up.sql
stoksc: i'm inclined to add a big warning and do the DML, though, so names as non-unique is so counter intuitive and i doubt it affects many users.

docs/release-notes/3152-model-registry-cli.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Improvement**
+
+-  Revert changes to Model Registry CLI format so models can be referred to by a unique name, and
+   not only ID
```
stoksc, line 6: i would add a warning that between release x and y (i'm not sure the versions), if duplicate named models were added that they will be renamed.

harness/determined/cli/model.py
stoksc: @mapmeld pointed out the filter is master side, so this is fine with me. `len(models) > 1` is basically enforced by the db constraint, 2nd check doesn't hurt, len(models) == 0 just means nothing matched the filter, makes sense to be client side.

stoksc (approved):
lgtm, though i would add the warning before merging

harness/determined/cli/model.py
rb-determined-ai: @stoksc What do you mean the filter is master side?  I see logic that should be done in SQL being done in python code:

* get model id from name: one round trip with sql like `select id from models where name=$name`
* describe model using model id: another round trip with sql like `select * from models where id=$1`

Wheras you could do it with a single round-trip and a single query like `select * from models where name=$name`.

mapmeld: @rb-determined-ai Is your objection more that (A) this requests the same info from the API twice, or (B) there should be an API endpoint to access model versions by name?

We are calling this only from `describe`, `list_versions`, and `register_version`
- In `describe` we make another call because `get_models` returns `num_versions` on each model but not specifics.
- In the other two, we need to call an API path which includes the model ID

rb-determined-ai: > we need to call an API path which includes the model ID

right, but we own the REST APIs.  If they are not suitable for the needs of our CLI they might not be suitable for end-users, either.

> Is your objection more that (A) this requests the same info from the API twice, or (B) there should be an API endpoint to access model versions by name?

I think (B).  I think it's an anitpattern to do some sql query, then feed output back into another sql query.  This is an antipattern even if you have direct access to the sql database (as the determined-master has) and it is a worse antipattern when you are doing it client-side.

I consider this a non-blocking issue. If you and @stoksc both understand my concern but don't share the concern, I'm fine landing this as-is.

PR #3155, opened 2021-11-03
feat: basic support for container recovery on agent restarts [DET-6061]

ioga:
## Description

Basic support for "reattaching" running containers on agent restart (or crash). 
- The full feature is somewhat dependent on #3070 (task logs); otherwise the task logs are lost. So, this PR hides the reattachment behind a feature flag: `resource_pool` configuration ffield `agent_reattach_enabled`.
- Since #2991 when an agent tried to cleanly reconnect under the disconnect timeout, it'd be rejected. With this change, they'll be "accepted" once again; same agent actor is reused. When reattachment if off, old container state is cleaned up as if they exited to an error.
- When reattachment is on, the master will send the agent a list of container it expects it to have; the agent will try to reattach these. It'll send a list of successfully reattached ones back to the master. Master will treat the containers not-successfully-reattached as if they exited to an error.
- Agent will now list and kill the containers with the same "agent_id", if they aren't supposed to be running (under reattachment).
- Allow users to configure `agent_reconnect_wait` on per-resource pool basis.

## Test Plan

`ManagedCluster` fixture is extended to account for the feature flag. A handful of tests added to check for successful recovery.

Manual test:
- Run `det cmd run sleep 120`.
- Kill agent; restart agent.
- The task will continue running and will finish successfully. (although logs are lost at the moment)

## Commentary (optional)

Future work: add tests for task logs working properly once #3070 lands.

## Checklist

### DIFF
--- a/agent/internal/containers.go
+++ b/agent/internal/containers.go
@@ -1,15 +1,19 @@
 package internal
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
 
+	"github.com/docker/docker/api/types"
 	dcontainer "github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/labstack/echo/v4"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 
 	"github.com/determined-ai/determined/master/pkg/actor"
 	"github.com/determined-ai/determined/master/pkg/aproto"
@@ -27,6 +31,8 @@ const (
 	dockerAgentLabel            = "ai.determined.container.agent"
 	dockerClusterLabel          = "ai.determined.container.cluster"
 	dockerMasterLabel           = "ai.determined.container.master"
+	dockerContainerVersionLabel = "ai.determined.container.version"
+	dockerContainerVersionValue = "0"
 )
 
 type containerManager struct {
@@ -39,6 +45,15 @@ type containerManager struct {
 	fluentPort int
 	docker     *client.Client
 }
+
+type (
+	requestReattachContainers struct {
+		ContainersToReattach []aproto.ContainerReattach
+	}
+	responseReattachContainers struct {
+		ContainersReattached []aproto.ContainerReattachAck
+	}
+)
 
 func newContainerManager(a *agent, fluentPort int) (*containerManager, error) {
 	return &containerManager{
@@ -92,6 +107,14 @@ func (c *containerManager) Receive(ctx *actor.Context) error {
 			dockerAgentLabel:         c.Options.AgentID,
 			dockerClusterLabel:       c.MasterInfo.ClusterID,
 			dockerMasterLabel:        c.MasterInfo.MasterID,
+		}
+	case requestReattachContainers:
+		reattachedContainers, err := c.reattachContainers(ctx, msg.ContainersToReattach)
+		if err != nil {
+			ctx.Log().WithError(err).Warn("failed to reattach containers")
+			ctx.Respond(responseReattachContainers{})
+		} else {
+			ctx.Respond(responseReattachContainers{ContainersReattached: reattachedContainers})
 		}
 
 	case aproto.ContainerLog, aproto.ContainerStateChanged, model.TrialLog:
@@ -160,13 +183,10 @@ func overwriteSpec(
 	for key, value := range labels {
 		spec.RunSpec.ContainerConfig.Labels[key] = value
 	}
-	spec.RunSpec.ContainerConfig.Labels[dockerContainerIDLabel] = cont.ID.String()
-	spec.RunSpec.ContainerConfig.Labels[dockerContainerParentLabel] = cont.Parent.String()
-	var slotIds []string
-	for _, d := range cont.Devices {
-		slotIds = append(slotIds, strconv.Itoa(d.ID))
+
+	for k, v := range makeContainerDockerLabels(cont) {
+		spec.RunSpec.ContainerConfig.Labels[k] = v
 	}
-	spec.RunSpec.ContainerConfig.Labels[dockerContainerDevicesLabel] = strings.Join(slotIds, ",")
 
 	spec.RunSpec.HostConfig.DeviceRequests = append(
 		spec.RunSpec.HostConfig.DeviceRequests, gpuDeviceRequests(cont)...)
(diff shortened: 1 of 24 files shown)


stoksc (commented):
mostly this makes sense to me, and looks good/is pretty logical.. this is just a partial first pass, i will finish soon (prob on ✈️). can you describe to me, briefly, the plans for the next move? now that we are recovering agents and their containers, how to we plan to transpose from this to recovering agents when the agent actor still isn't in the master? pretty much the same, but with rerunning the scheduler and recovering the allocator state, so that we can know the list of containers we expect to survive?

agent/internal/containers.go
```diff
@@ -212,3 +232,168 @@ func containerEnvVars(cont cproto.Container) []string {
...
+	ctx *actor.Context, expectedSurvivors []aproto.ContainerReattach) (
+	[]aproto.ContainerReattachAck, error) {
+	result := make([]aproto.ContainerReattachAck, 0, len(expectedSurvivors))
+
+	runningContainers, err := c.listRunningContainers(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, survivor := range expectedSurvivors {
```
stoksc, line 246: nit, it's not necessarily a survivor, but an expectedSurvivor. i had to double take reading. maybe just actual/expected? idk

agent/internal/containers.go
```diff
@@ -212,3 +232,168 @@ func containerEnvVars(cont cproto.Container) []string {
...
+	if containerCurrState.State != containerPrevState.State {
+		return nil, errors.New("container has changed state while offline")
+	}
+
+	cid := containerPrevState.ID
+	_, ok := ctx.ActorOf(cid, reattachContainerActor(*containerCurrState, c.docker))
+	if !ok {
+		errorMsg := fmt.Sprintf("failed to reattach container %s: actor already exists", cid)
+		ctx.Log().Warnf(errorMsg)
+		if killed := ctx.Kill(cid); killed {
```
stoksc, line 312: could sigkill the survivor here too, but this is already very unlikely (just dont remove it from runningContainers in the event of certain errors?)

master/internal/resourcemanagers/agent/slots.go
```diff
@@ -40,7 +40,13 @@ func (s *slots) Receive(ctx *actor.Context) error {
 			ctx.Tell(child, msg)
 		}
 	case aproto.ContainerStateChanged:
-		s.sendToSlots(ctx, msg.Container, msg)
+		if msg.Broadcast {
+			for _, child := range ctx.Children() {
```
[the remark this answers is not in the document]
stoksc: how does this come up? when would the container already be forgotten by the slot but we still attempt to recover it. i figure if the slot isn't aware of it, then it wouldn't be in the list of expected survivors?

stoksc (commented):
looks good, and after our chat it mostly master sense. my meta thoughts follow

An alternative solution to this problem is to have the agent journal its state and recover from journaled state. Then the reconnect scheme in the master would not need a change, except maybe to allow longer time outs, since the agent can recover itself. Instead we've opted to have the agent's state be redundant in the master, and to have the master reconstruct and relay it on crashes. This makes an assumption that the master (and other sources) will always have all the recoverable state required to facilitate an agent recovery. Now, we have to rely on Docker to not crash and keep some of this state in the event we crash. Does this hold true for all container runtimes, e.g. can we rely on singularity for the same? Is there any state we would need that isn't generally recoverable in this way? Should the agent recover from journaled state and not rely on the master to be able to reconstruct all its state? I think the solution is fine for now, and probably works in general (i can't really think of anything that can't be overcome, e.g. if we need more state, we can tack some recovery info union onto the container state updates we send to the master and have it give us those back. i do think, done right, the disk ft solution is probably simpler/self contained, easier to reason about or at least has a smaller blast radius when things go wrong.

agent/internal/containers.go
stoksc: oops, yep nice

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -343,3 +371,68 @@ func (a *agent) summarize(ctx *actor.Context) model.AgentSummary {
 		Draining:       a.draining,
 	}
 }
+
+func (a *agent) gatherContainersToReattach(ctx *actor.Context) []aproto.ContainerReattach {
+	result := make([]aproto.ContainerReattach, 0, len(a.containers))
+	for containerID, allocation := range a.containers {
+		resp := ctx.Ask(allocation, sproto.GetTaskContainerState{
```
stoksc, line 378: needing this ask is a bit much, side effect of the master architecture though (which, of course, i'm determined to fix). asks like this become a bit dangerous, we might should start a list somewhere, so we can statically check for deadlock potential

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -92,16 +95,31 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
+		if a.awaitingReconnect && a.agentReattachEnabled {
+			optsCopy := *a.opts
+			optsCopy.ContainersToReattach = a.gatherContainersToReattach(ctx)
+			masterSetAgentOptions = aproto.AgentMessage{MasterSetAgentOptions: &optsCopy}
+		} else {
+			// Quick reconnect after restart/crash.
+			clearAll := map[cproto.ID]aproto.ContainerReattachAck{}
+			a.clearNonReattachedContainers(ctx, clearAll)
+			masterSetAgentOptions = aproto.AgentMessage{MasterSetAgentOptions: a.opts}
+		}
```
stoksc, line 116: just a bit clearer that it's the same when it's a first run. also a bit less wasteful.
```suggestion
		var masterSetAgentOptions aproto.AgentMessage
		switch {
		case a.awaitingReconnect && a.agentReattachEnabled:
			optsCopy := *a.opts
			optsCopy.ContainersToReattach = a.gatherContainersToReattach(ctx)
			masterSetAgentOptions.MasterSetAgentOptions = &optsCopy
		case a.awaitingReconnect:
			// Quick reconnect after restart/crash.
			clearAll := map[cproto.ID]aproto.ContainerReattachAck{}
			a.clearNonReattachedContainers(ctx, clearAll)
			masterSetAgentOptions.MasterSetAgentOptions = a.opts
		default:
			masterSetAgentOptions.MasterSetAgentOptions = a.opts
		}
```

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -92,16 +95,31 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
 		lastColonIndex := strings.LastIndex(msg.Ctx.Request().RemoteAddr, ":")
 		if lastColonIndex == -1 {
 			a.address = msg.Ctx.Request().RemoteAddr
 		} else {
 			a.address = msg.Ctx.Request().RemoteAddr[0:lastColonIndex]
 		}
-		wsm := ws.WriteMessage{Message: aproto.AgentMessage{MasterSetAgentOptions: a.opts}}
+
+		var masterSetAgentOptions aproto.AgentMessage
+		if a.awaitingReconnect && a.agentReattachEnabled {
```
stoksc, line 107: trying to read this carefully and think - previously if an agent had an intermittent network failure, we would not lose the containers (due to the reconnect/replay pr i landed). now if one has a network failure with a.agentReattachEnabled == false, i think when it reconnects we will drop all containers? we only need to clear all when the agent has forgotten its containers, right?

master/internal/resourcemanagers/agent/slots.go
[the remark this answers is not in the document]
stoksc: ah i see what you mean. i think the alternative makes sense.. i mean, a failure means we're already in a bad state i think. but i think this is fine. maybe just Broadcast -> BroadcastToAllSlots and leave a comment that in the event we couldn't recover the container we may not be sure which slots it was on (without more work) so we just tell all slots that a container they may have died.

master/internal/resourcemanagers/agent/agent.go
stoksc: @ioga this is the only comment that is critical i think

master/internal/agent/agent.go
```diff
@@ -92,16 +95,31 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
 		if lastColonIndex == -1 {
 			a.address = msg.Ctx.Request().RemoteAddr
 		} else {
 			a.address = msg.Ctx.Request().RemoteAddr[0:lastColonIndex]
 		}
-		wsm := ws.WriteMessage{Message: aproto.AgentMessage{MasterSetAgentOptions: a.opts}}
+
+		var masterSetAgentOptions aproto.AgentMessage
+		if a.awaitingReconnect && a.agentReattachEnabled {
+			optsCopy := *a.opts
```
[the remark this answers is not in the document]
stoksc: speaking on perf, we're declaring two aproto.AgentMessage's every time

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -92,16 +95,35 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
+			optsCopy := *a.opts
+			optsCopy.ContainersToReattach = a.gatherContainersToReattach(ctx)
+			masterSetAgentOptions = aproto.AgentMessage{MasterSetAgentOptions: &optsCopy}
+		} else {
```
stoksc, line 111: excuse me for thinking out loud here a bit.

same comment here about enumerating cases explicitly; here it is awaitingReconnect x agentReattachEnabled x IsReconnect, which is a lot of cases, sure, but 'compressing' them requires the reader to carefully trace the code to check we handle each given case correctly. tracing agentReattachEnabled == false, IsReconnect == true, awaitingReconnect == false in isolation, i see we may have an agent reconnect (maybe with containers still active), but since we think the agent is reconnecting cleanly we don't clear its containers, but awaitingReconnect == false implies we may've forgotten about its containers. but of course IsReconnect should always imply awaitingReconnect when the actor is still alive, and if the agent actor wasn't alive, we wouldn't reconnect it improperly but kick the agent away in the agents actor while it had IsReconnect == true (and if was reconnecting while "still connected", we always panic on a second connecting), so this case is bogus (not reachable). but we handle it as if it is possible and i think that is what makes this hard to grok (that being said, if it _were_possible, killing the containers would make sense I think). i think verifying this is correct would be a lot easier with them all enumerated with comments, and we called out cases like this. of course, explicitly specifying the dependent variables and verifying invariants would help too (like having the invariant IsReconnect implies awaitingReconnect explicit, and explaining how it is upheld, in the code would be helpful).

in general i've thought about this some before, when i was trying to write the trial actor and trying to rewrite the trial in a way i knew was correct, and there are a spectrum of options between two ends: enumerating cases explicitly, entirely, or fulling decomposing the case space into regions that express like cases with if/else and boolean logic (taking advantage of implicit relationships between variables). the former is wordy, and you have to use routines to not repeat yourself, but has the advantage of allowing less mistakes since all cases are thought through by force, the latter is the norm (as a side effect how we all write code) but does allow for subtle mistakes by nature the added work of attempting to simplify/specify a function and making assumption about invariants that may not be upheld.

so i guess my takeaway here is maybe this is complicated enough and the cases are subtle enough that enumerating more of them explicitly, panicking on invalid states and leaving good comments about intended behavior could be fun. idk... but it's been fun to think about.

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -224,3 +262,149 @@ def test_agent_restart_cmd_container_failure(
...
+        managed_cluster.kill_agent()
+        time.sleep(downtime)
+        managed_cluster.restart_agent(wait_for_amnesia=False)
+
+        wait_for_command_state(command_id, "TERMINATED", 30)
+
+        # Commands fail if they have finished while the agent was off.
+        reattach_wait = managed_cluster.fetch_config_reattach_wait()
+        succeeded = "success" in get_command_info(command_id)["exitStatus"]
+        assert succeeded is (reattach_wait > downtime)
```
stoksc, line 329: why is?
```suggestion
        assert succeeded == (reattach_wait > downtime)
```

stoksc (approved):
lgtm, of course giving tests are working (at your discretion to take the remaining feedback). amazing work on this, it was a very complex pr but the approach is sound and has helped a lot in making it easy to understand. also the tests are amazing, great.

master/internal/resourcemanagers/agent/agent.go
stoksc: i explained what i meant by this more clearly further down, so resolving.

master/internal/resourcemanagers/agent/slots.go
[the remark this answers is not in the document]
stoksc: would still be nice to add this too, i think this would be a natural question to ask after reading the code (i could see a chesterton's fence situation where it gets removed by mistake)

e2e_tests/tests/cluster/test_agent_restart.py
ioga: mypy tends to recommend using `is` for boolean / None checks.

stoksc: had no idea, gtk

PR #3156, opened 2021-11-03
chore: write our own swagger bindings

rb-determined-ai:
try:

    cd bindings
    python bindings.py > b.py
    pyinstrument b.py
    mypy b.py

On my laptop, `pyinstrument b.py` shows:
 - 20ms for the python interpreter to read/compile the code
 - 10ms to actually import the module

## Commentary (optional)

Initial draft, passes mypy.  Not tested yet.

### DIFF
--- a/bindings/bindings.py
+++ b/bindings/bindings.py
@@ -1,0 +1,558 @@
+import abc
+import json
+import os
+import sys
+import typing
+
+SWAGGER = "proto/build/swagger/determined/api/v1/api.swagger.json"
+SWAGGER = os.path.join(os.path.dirname(__file__), "..", SWAGGER)
+
+Code = str
+TypeDefs = typing.Dict[str, typing.Optional["TypeDef"]]
+
+
+class TypeAnno:
+    def annotation(self, prequoted=False) -> Code:
+        raise NotImplementedError(type(self))
+
+    def need_parse(self) -> bool:
+        raise NotImplementedError(type(self))
+
+    def load(self, val: Code) -> Code:
+        raise NotImplementedError(type(self))
+
+    def dump(self, val: Code) -> Code:
+        raise NotImplementedError(type(self))
+
+    def isnone(self) -> bool:
+        # Only Refs to empty structs ever return True; we skip generating them.
+        return False
+
+
+class TypeDef:
+    def gen_def(self) -> Code:
+        raise NotImplementedError(type(self))
+
+
+class NoParse:
+    """A compositional class for things where json.loads/dumps is sufficient."""
+
+    def need_parse(self) -> bool:
+        return False
+
+    def load(self, val: Code) -> Code:
+        return val
+
+    def dump(self, val: Code) -> Code:
+        return val
+
+
+class Any(NoParse, TypeAnno):
+    def __repr__(self):
+        return "Any"
+
+    def annotation(self, prequoted=False) -> Code:
+        return "typing.Any"
+
+
+class String(NoParse, TypeAnno):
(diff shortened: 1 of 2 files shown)


bindings/bindings.py
```diff
@@ -0,0 +1,558 @@
...
+    def dump(self, val: Code) -> Code:
+        return f"{self.name}.from_json({val})"
+
+    def gen_def(self) -> Code:
+        out = [f"class {self.name}:"]
+        out += ["    def __init__("]
+        out += ["        self,"]
+        out += [f"        {k}: {v.annotation()}," for k, v in self.members.items()]
+        out += ["    ):"]
+        out += [f"        {k} = {k}" for k in self.members]
```
ioga, line 225:
```suggestion
        out += [f"        self.{k} = {k}" for k in self.members]
```

bindings/bindings.py
```diff
@@ -0,0 +1,558 @@
+import abc
```
ioga, line 1: that's not used; this file itself is not under mypy so there's a few "linter issues" here and there.

bindings/bindings.py
```diff
@@ -0,0 +1,558 @@
...
+class Enum(TypeDef):
+    def __init__(self, name, members):
+        self.name = name
+        self.members = members
+
+    def load(self, val: Code) -> Code:
+        return f"{val}.value"
+
+    def dump(self, val: Code) -> Code:
+        return f"{self.name}({val})"
```
ioga, line 260: I think these two are swapped

```suggestion
    def load(self, val: Code) -> Code:
        return f"{self.name}({val})"

    def dump(self, val: Code) -> Code:
        return f"{val}.value"
```

hamidzr:
this sounds great, I'm working on a review. Would we want to also commit the generated code in this pr?

hamidzr (commented):
I ran into some issues testing the generated code. after applying a patch and fixing a seemingly swapped load and dump

```
Traceback (most recent call last):
  File "/home/hmd/projects/da/determined/harness/determined/cli/cli.py", line 240, in main
    parsed_args.func(parsed_args)
  File "/home/hmd/projects/da/determined/harness/determined/cli/master.py", line 44, in get_master
    print(resp.to_json())
  File "/home/hmd/projects/da/determined/harness/determined/common/api/b.py", line 1523, in to_json
    "version": self.version,
AttributeError: 'v1GetMasterResponse' object has no attribute 'version'
Failed to fetch master info
```
[bindings.patch.txt](https://github.com/determined-ai/determined/files/7488572/bindings.patch.txt)

rb-determined-ai:
I'll take a look when I get a chance

hamidzr:
I found the issue, fixed it and added an example https://github.com/determined-ai/determined/compare/master...hamidzr:inhouse-bindings?expand=1

hamidzr:
something else we might want camelCase to snake_case conversion for python. maybe not worth it.

rb-determined-ai:
pushed my updates.  Looks like you have some fixes I don't.  I'm fine with you taking this over, I clearly don't have the bandwidth and you are blocked on it.

rb-determined-ai:
> something else we might want camelCase to snake_case conversion for python. maybe not worth it.

See https://github.com/determined-ai/determined/blob/master/schemas/gen.py#L15.  If you choose to use that function, please just copy/paste it.  The cross-repo import would be much messier than a copy/paste I think.

rb-determined-ai:
closing in favor of https://github.com/determined-ai/determined/pull/3361

PR #3157, opened 2021-11-03
feat: make `determined-agent [ARGS]` behave as `determined-agent run [ARGS]`.

ioga:
## Description

One can run master as `determined-master`, but `determined-agent` just prints help, and you need `determined-agent run`. This is odd. 

This PR makes it work either way for the sake of backwards compatibility and convenience.

Cobra doesn't really support that, and they simply recommend users to inject `os.Args[1]`: https://github.com/spf13/cobra/issues/725

## Test Plan

- `determined-agent [RUN SUBCOMMAND ARGS]` should work the same as `determined-agent run [RUN SUBCOMMAND ARGS]`.
- other subcommands should still work fine.

## Checklist

### DIFF
--- a/agent/cmd/determined-agent/main.go
+++ b/agent/cmd/determined-agent/main.go
@@ -1,10 +1,39 @@
 package main
 
 import (
+	"os"
+
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


stoksc (approved):
lgtm

PR #3159, opened 2021-11-04
draft: distributed data parallel with pytorch native

azhou-determined:
## Description

- Get rid of hvd_config
- Test model for native pytorch DDP
## Checklist


examples/tutorials/mnist_pytorch_dtrain/README.md
```diff
@@ -0,0 +1,29 @@
+# PyTorch MNIST CNN Tutorial
```
rb-determined-ai, line 1: is the example just copy/pasted, for testing purposes?

harness/determined/_execution.py
```diff
@@ -128,11 +128,11 @@ def _make_local_execution_env(
         test_mode=test_mode,
         on_cluster=False,
     )
-    hvd_config = horovod.HorovodContext.from_configs(env.experiment_config, env.hparams)
+    # hvd_config = horovod.HorovodContext.from_configs(env.experiment_config, env.hparams)
```
rb-determined-ai, line 131: don't forget to delete this comment

harness/determined/_generic/_distributed.py
```diff
@@ -34,6 +35,7 @@ def __init__(
         local_size: int,
         cross_rank: int,
         cross_size: int,
+            backend: Optional[str],
```
rb-determined-ai, line 38: I don't think this is the right way to pass the backend information around.  The backend is 'training loop-specific information', while the DistributedContext is more like a 'training-loop-independent way to ensure a minimum set of distributed communication primitives'

In the case of a totally custom training loop, the backend would just be some random string that the user sets and knows how to interpret, and that's a strange thing to attach to this object I think.

harness/determined/_trial_controller.py
```diff
@@ -38,20 +36,17 @@ def __init__(
...
         self.scheduling_unit = self.env.experiment_config.scheduling_unit()
 
-        if self.hvd_config.use:
-            self.is_chief = hvd.rank() == 0
-        else:
-            self.is_chief = True
+        self.is_chief = context.distributed.rank == 0
 
-        if self.hvd_config.use and not self.is_chief:
+        if context.distributed.backend == "horovod" and not self.is_chief:
```
rb-determined-ai, line 41: prolly this logging setup is a good default regardless of the backend.

harness/determined/estimator/_estimator_context.py
```diff
@@ -60,7 +68,7 @@ def wrap_optimizer(self, optimizer: Any) -> Any:
             return optimizer
 
         self.optimizer_initialized = True
-        if not self.hvd_config.use:
+        if not self.distributed.backend == "horovod":
```
rb-determined-ai, line 71: does `distributed.backend == "horovod"` indicate:
1. that we are using horovod for dtrain, and we are currently running under horovodrun, or
2. that we are using horovod for dtrain, even though we may be on a single gpu at the moment (no horovodrun)?

Also, what happens if the backend is something other than horovod, but we are in distributed training?  I'm in a rush with this review, but do you assert that EstimatorTrial only works with horovod somewhere?

rb-determined-ai (commented):
it's friday afternoon. I'm only halfway through this review, but that's all I got for this week, sorry.

Removing horovod context from everywhere looks great, I love it!

PR #3160, opened 2021-11-04
fix: fluentbit should be able to run as nonroot [DET-5949]

stoksc:
## Description
This change updates the kubernetes resource manager so that experiments launched from non-root users will also launch fluentbit as the `nonroot` user included in the fluentbit container.

## Test Plan
- [x] test that launching a non root experiment works on k8s
- [ ] try to come up with a satisfactory explanation for why other seemingly reasonable solutions do not seem to work (may be due to a bug, looking that way)

## Commentary (optional)
Note: the init container still runs as root because `tar --same-owner` requires root. Changing that will require more work.

## Checklist

### DIFF
--- a/docs/release-notes/3160-nonroot-fluentbit.txt
+++ b/docs/release-notes/3160-nonroot-fluentbit.txt
@@ -2,5 +2,5 @@
 
 **Improvements**
 
--  Security and Logging: if a non-root user is used to run jobs on Kubernetes, the Fluentbit sidecar
+-  Security and Logging: if a non-root user is used to run jobs on Kubernetes, the Fluent Bit sidecar
    will also run as non-root.
--- a/docs/release-notes/3160-nonroot-fluentbit.txt
+++ b/docs/release-notes/3160-nonroot-fluentbit.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Improvements**
+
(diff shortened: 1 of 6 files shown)


eecsliu (approved):
lgtm! 👩‍🍳😗👌

master/pkg/fluent/fluent.go
```diff
@@ -170,13 +186,18 @@ end
 
 	var config strings.Builder
 
+	bufferDir := "/var/log/flb-buffers"
+	if asNonRoot {
+		bufferDir = nonRootHome + "flb-buffers"
```
dzhu, line 191:
```suggestion
		bufferDir = filepath.Join(nonRootHome, "flb-buffers")
```

docs/release-notes/3160-nonroot-fluentbit.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Improvements**
+
+-  Security and Logging: if a non-root user is used to run jobs on Kubernetes, the Fluentbit sidecar
```
dzhu, line 5:
```suggestion
-  Security and Logging: if a non-root user is used to run jobs on Kubernetes, the Fluent Bit sidecar
```

PR #3162, opened 2021-11-04
fix: update transformers GLUE example

mapmeld:
## Description

Updates the NLP GLUE example to use latest transformers (v4.12.3) and datasets (1.15.1) to tokenize, train, download, and evaluate the main GLUE metrics.  Makes it easier to use any dataset from the datasets hub.

## Commentary

Main changes are: installing git-lfs (required to download newer models), using `AutoTokenizer` and `AutoModel` instead of model-specific imports, and downloading / processing the GLUE metrics with the `datasets` library.

TODOs: I probably should make a new notebook / ipynb.

## Checklist


liamcli:
> ## Description
> Updates the NLP GLUE example to use latest transformers (v4.12.3) and datasets (1.15.1) to tokenize, train, download, and evaluate the main GLUE metrics. Makes it easier to use any dataset from the datasets hub.
> 
> ## Commentary
> Main changes are: installing git-lfs (required to download newer models), using `AutoTokenizer` and `AutoModel` instead of model-specific imports, and downloading / processing the GLUE metrics with the `datasets` library.
> 
> TODOs: I probably should make a new notebook / ipynb.
> 
> ## Checklist
> * [x]  User-facing API changes need the "User-facing API Change" label.
> * [] Release notes should be added as a separate file under `docs/release-notes/`.
> * [x]  Licenses should be included for new code which was copied and/or modified from any external code.

Can we add verify convergence performance for `const.yaml` config as an outstanding TODO here?

liamcli (approved):
We shouldn't spend too much time on this example because it overlaps with our model hub support for transformers.  We should probably deprecate this example at some point in favor of model hub.

liamcli (changes requested):
Will approve once we verify that convergence behavior is as expected.

PR #3171, opened 2021-11-05
refactor: classmethods instead of staticmethods in trial controllers.

ioga:
## Description

Use classmethods for some trial controller internals to allow users inherit from our TrialControllers and override these w/o issues with inheritance.

Motivated by a question https://determined-community.slack.com/archives/CV3MTNZ6U/p1636077811035500 , where they wanted to setup tf logical devices in order to debug some OOM issue. That requires overriding `TFKerasTrialController#_configure_session`, but staticmethod style forces you to also override `from_trial`. 

Existing overrides using `staticmethod` shouldn't be affected and should continue working.

## Test Plan

Everything works as before.

## Checklist

### DIFF
--- a/harness/determined/_trial_controller.py
+++ b/harness/determined/_trial_controller.py
@@ -1,6 +1,6 @@
 import abc
 import logging
-from typing import Any, Optional
+from typing import Any, Optional, Type
 
 import determined as det
 from determined import horovod, profiler, workload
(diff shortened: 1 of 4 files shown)


rb-determined-ai (approved):
I like it.  I had never considered the consequences this had to subclassability before.

PR #3167, opened 2021-11-05
feat: jq init webui for job queue

hamidzr:
## Description

- adds the initial webui for job queue
- generic action dropdown interface

## Commentary (optional)

jobq

- skipping over the API calls: since the APIs could still change I haven't wrapped all the API inputs and outputs in webui falvored types and I'm planning to keep it like that until we're landing ito production
- [x] remove reviewonly comments before releasing into production

## Checklist

- [x] User-facing API changes need the "User-facing API Change" label.
- [x] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [x] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3172, opened 2021-11-05
feat: read-only model registry UI [DET-5992] [DET-5993] [DET-5994]

EmilyBonar:
## Description
This isn't quite done yet but I thought I'd put it up to allow for a head start on reviewing since it's such a big addition.

## Checklist


webui/react/src/components/EditableMetadata.tsx
```diff
@@ -0,0 +1,108 @@
+import { Button, Dropdown, Form, Input, Menu } from 'antd';
+import { FormListFieldData } from 'antd/lib/form/FormList';
+import React, { useCallback, useMemo } from 'react';
+
+import css from './EditableMetadata.module.scss';
+import Icon from './Icon';
+import InfoBox, { InfoRow } from './InfoBox';
+
+interface Props {
+  editing: boolean;
```
hkang1, line 10: **suggestion:** if editing is most likely going to be `false` and an opt in by the user, I would suggest making it optional, so that not supplying one would default to `undefined` which would be the same as an explicit `false` behavior.
```
editing?: boolean;
```

webui/react/src/components/EditableMetadata.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return <Form.Item
+    {...field}
+    name={name}
+    noStyle>
+    <Input.Group className={css.row} compact>
+      <Form.Item name={[ name, 'key' ]} noStyle>
+        <Input placeholder="Enter metadata label" />
+      </Form.Item>
+      <Form.Item name={[ name, 'value' ]} noStyle>
+        <Input placeholder="Enter metadata" />
```
mapmeld, line 90: Optional: would it be better to have this as 'Enter metadata value' ?

hkang1:
I'll list out UI observations and issues here.

Updates to the tag on the version detail page doesn't take.

2021-11-15 at 1 17 PM

When saving new `Notes` the UI immediately shows the previous note before quickly showing the updated Note. Instead the UI should block the entire Notes section until the saving takes and update the UI with the new note (without making another call to the API). Ignore the 2nd note on the screenshot :D shouldn't have to make another API call since we already have the new note string.

2021-11-15 at 1 20 PM

There is currently no feedback that the copy to clipboard worked. Maybe shoot an `antd` notification. Check out Master Logs with the same interaction.

2021-11-15 at 1 24 PM

webui/react/src/pages/ModelRegistry.tsx
```diff
@@ -0,0 +1,404 @@
...
+
+    const overflowRenderer = (_:string, record: ModelItem) => {
+      const isDeletable = user?.isAdmin || user?.username === record.username;
+      return (
+        <Dropdown
+          overlay={(
+            <Menu>
+              <Menu.Item
+                key="switch-archived"
+                onClick={() => switchArchived(record)}>
```
hkang1, line 246: **chore:** Archiving and unarchiving can use a `try/catch` block plus `async/await` to ensure the right sequence of API call + UI update.

webui/react/src/pages/ModelVersionDetails.tsx
```diff
@@ -0,0 +1,295 @@
+import { CopyOutlined, EditOutlined } from '@ant-design/icons';
```
hkang1, line 1: marking that all the same API + try/catch application will need to be applied here as well

webui/react/src/pages/ModelDetails.tsx
```diff
@@ -0,0 +1,285 @@
...
+    setIsEditingMetadata(true);
+  }, []);
+
+  const saveMetadata = useCallback(() => {
+    setIsEditingMetadata(false);
+    patchModel({
+      body: { id: parseInt(modelId), metadata: editedMetadata },
+      modelId: parseInt(modelId),
+    });
+    fetchModel();
```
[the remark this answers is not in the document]
hkang1: as for the repetitive nature of these changes (one for each different API call), `TaskActionDropdown.tsx` and also @hamidzr's new PR wrote this in a minimal way. Might be worth taking a look at [his PR](https://github.com/determined-ai/determined/pull/3167/files#diff-8dd6bea64a7a22358e61798d64cd2ca4f3640837057db78d7af2386f77b6a246R1), expecially `ActionDropdown.tsx` and `JobQueue.tsx`

hkang1 (commented):
The model registry is looking amazing! Excited to see this in the wild!

I've written a lot of little things down, but conceptually the biggest one is to put in some sort of UI blocking mechanism so that the user doesn't think the UI is unresponsive when an API for changing model / model version is being called.

webui/react/src/pages/ModelDetails.tsx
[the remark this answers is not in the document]
hamidzr: I've separated that part so we can start using it sooner https://github.com/determined-ai/determined/pull/3194/files there are some comment from caleb that I still need to apply there

webui/react/src/components/DownloadModelPopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+  { children, modelVersion }: PropsWithChildren<Props>,
+) => {
+  const [ visible, setVisible ] = useState(false);
+
+  const downloadCommand = useMemo(() => {
+    return `det checkpoint download ${modelVersion.checkpoint.uuid}`;
+  }, [ modelVersion.checkpoint.uuid ]);
+
+  const handleCopy = useCallback(async () => {
+    await navigator.clipboard.writeText(downloadCommand);
```
hkang1, line 21: **suggestion:** we have a `copyToClipboard` utility function that does this and also has some backwards compatibility. It'll be good to use this so we can centralize a single point of location to update code if we want to change the behavior of it in the future.

webui/react/src/pages/ModelDetails/ModelHeader.tsx
```diff
@@ -0,0 +1,137 @@
...
+    onSaveDescription, onUpdateTags, onSaveName,
+  }: Props,
+) => {
+  const { auth: { user } } = useStore();
+
+  const infoRows: InfoRow[] = useMemo(() => {
+    return [ {
+      content:
+      (<Space>
+        {userRenderer(model.username, model, 0)}
```
hkang1, line 39: **suggestion:** the `<xxx>Renderer` functions are generally for Table use only. How about taking `Avatar` component and adding a prop like `showName?: boolean` or something like to support showing both the user bubble + the username/name.

webui/react/src/pages/ModelDetails/ModelHeader.tsx
```diff
@@ -0,0 +1,137 @@
...
+        {model.username + ' on ' +
+      formatDatetime(model.creationTime, 'MMM D, YYYY', false)}
+      </Space>),
+      label: 'Created by',
+    },
+    { content: relativeTimeRenderer(new Date(model.lastUpdatedTime)), label: 'Updated' },
```
hkang1, line 45: we can leave this one alone. This seems common enough of a use case and I think Alex had some ideas on displaying a slightly different `TimeAgo` logic where it can be its own reusable component in the near future.

hkang1 (approved):
Looking much better! Mostly minor improvement suggestions.

2021-11-19 at 1 25 PM

PR #3179, opened 2021-11-10
feat: get jobs for fairshare scheduler

hamidzr:
## Checklist

### DIFF
--- a/master/internal/resourcemanagers/fair_share.go
+++ b/master/internal/resourcemanagers/fair_share.go
@@ -54,9 +54,22 @@ func (f *fairShare) Schedule(rp *ResourcePool) ([]*sproto.AllocateRequest, []*ac
 }
 
 func (f *fairShare) JobQInfo(rp *ResourcePool) map[model.JobID]*job.RMJobInfo {
-	// TODO report job states.
-	// compute a list of jobs.
-	return make(map[model.JobID]*job.RMJobInfo)
+	for _, resAllocated := range rp.taskList.allocations {
+		req := rp.taskList.taskByID[resAllocated.ID]
+		if req.JobID == nil {
+			continue
+		}
+		req.State = job.SchedulingStateScheduled
+	}
+	reqs := make(AllocReqs, 0)
+	for _, req := range rp.taskList.taskByID {
+		reqs = append(reqs, req)
+	}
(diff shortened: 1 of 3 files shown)


master/internal/resourcemanagers/fair_share.go
```diff
@@ -50,11 +50,41 @@ func (g groupState) String() string {
 }
 
 func (f *fairShare) Schedule(rp *ResourcePool) ([]*sproto.AllocateRequest, []*actor.Ref) {
+	defer f.updateJobs(rp) // CHECK does this work how I think it does?
```
eecsliu, line 53: are you trying to send updates after the scheduling occurs? Then yes it should work :)

eecsliu (approved):
Looks great! just a couple of clarifications

master/internal/resourcemanagers/fair_share.go
hamidzr: yep exactly and I didn't want to store it in a temp var explicitly.

master/internal/resourcemanagers/fair_share.go
```diff
@@ -50,11 +50,41 @@ func (g groupState) String() string {
...
+		}
+		req.State = job.SchedulingStateScheduled
+	}
+	reqs := make(AllocReqs, 0)
+	for _, req := range rp.taskList.taskByID {
+		reqs = append(reqs, req)
+	}
+	jobQ, jobActors := mergeToJobQInfo(reqs)
+	for _, j := range jobQ {
+		j.JobsAhead = -1 // we don't support job order for fairshare scheduler.
```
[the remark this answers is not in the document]
hamidzr: would that accurately represent the order in which the jobs get scheduled? if so we should do that otherwise I'd stick with just going off of the job id to favor correctness and clarity 🤔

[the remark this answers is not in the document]
hamidzr: > Or is this a functionality we can provide in the webui side?

due to pagination, we'd probably need to do it server-side but I think we can do it easily in a separate pr. I'd say we move ahead as is to get a baseline job representation for fairshare first.

master/internal/resourcemanagers/fair_share.go
hamidzr: thanks

PR #3175, opened 2021-11-08
chore: move config into its own package

hamidzr:
## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3181, opened 2021-11-10
chore: jq read weight, priority, and preemption from config

hamidzr:
## Checklist

### DIFF
--- a/master/internal/config/config.go
+++ b/master/internal/config/config.go
@@ -251,6 +251,10 @@ func ReadPreemptionStatus(config *Config, rpName string, jobConf interface{}) bo
 			RMPremption = *preemption
 			return jobPreemptible && RMPremption
 		}
+		if rpConfig.Provider != nil && rpConfig.Provider.GCP != nil {
+			RMPremption = rpConfig.Provider.GCP.InstanceType.Preemptible
+			return jobPreemptible && RMPremption
+		}
 		break
 	}
 
--- a/master/internal/config/config.go
+++ b/master/internal/config/config.go
@@ -264,7 +264,12 @@ func ReadPreemptionStatus(config *Config, rpName string, jobConf interface{}) bo
 		}
 	}
 
-	// TODO k8 RM?
+	if config.ResourceManager.KubernetesRM != nil {
+		if config.ResourceManager.KubernetesRM.DefaultScheduler == "preemption" {
+			RMPremption = true
+		}
+	}
+
 	return jobPreemptible && RMPremption
 }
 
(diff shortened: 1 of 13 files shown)


hamidzr:
related https://github.com/determined-ai/determined/pull/3175

master/internal/config/config.go
```diff
@@ -191,3 +191,106 @@ type InternalConfig struct {
...
+	}
+
+	RMPremption := false // wheter the RM supports preemption
+
+	schdConf := getSchedulerConfig(config, rpName)
+	if preemption := readPreemptionFromScheduler(schdConf); preemption != nil {
+		RMPremption = *preemption
+	}
+
+	// TODO k8 RM?
```
hamidzr, line 244: @eecsliu would these need to be read out of help configuration?

eecsliu: You should be able to get these from the `config.ResourceManager.KubernetesRM`. The scheduler is a string, where "preemption" indicates the preemption scheduler. There is no weight, and priority would be 50 by default. K8s doesn't support setting a default priority right now.

master/internal/config/config.go
```diff
@@ -211,3 +211,116 @@ type InternalConfig struct {
...
+
+	// if not found, fall back to resource manager config
+	schedulerConf = config.ResourceManager.AgentRM.Scheduler
+	prio = readPriorityFromScheduler(schedulerConf)
+	if prio != nil {
+		return *prio
+	}
+
+	// TODO k8 RM?
+	return resourcemanagers.DefaultSchedulingPriority
```
eecsliu, line 310: You should be able to read the priority from the kubernetes resource manager - the code will be a little different from reading it from the scheduler, but the priority should be up to date and stored in the groups. Default kubernetes priority is 50, though.

master/internal/config/config.go
eecsliu: If you want to read whether or not a job is currently being preempted, then we'd need to go to the k8s api. But for this stuff, you can infer from the config options very easily.

eecsliu:
forgot to leave a review comment: looks mostly good - few things k8s related. They should be pretty easy to fix up.

master/internal/config/config.go
hamidzr: > but the priority should be up to date and stored in the groups

we should talk about this part and whether we want it to bubble up to the config level to persist. this is currently what happens with adjusting priority I beleive.

eecsliu (approved):
looks good!

eecsliu:
I'll also talk to you about leftover k8s things

hamidzr:
rebased to resolve the merge conflicts.

PR #3183, opened 2021-11-11
docs: remove det exp from docs

philipnrmn:
## Description

[DET-6196](https://determinedai.atlassian.net/browse/DET-6196) - det exp does not work

The [CLI docs](https://docs.determined.ai/latest/interact/cli.html#usage) suggest that `det exp` can be used as a shorthand for `det experiment`, but this is not implemented. Since there are no other occurrences of this shorthand elsewhere in the docs, I simply trimmed this line. 

## Test Plan

N/A

## Checklist

### DIFF
--- a/docs/interact/cli.txt
+++ b/docs/interact/cli.txt
@@ -84,7 +84,6 @@ For example, the different commands within each of the blocks below all do the s
 
    # List all experiments.
    $ det experiment list
-   $ det exp list
    $ det e list
    $ det e
 


mackrorysd (approved):
Seems reasonable to me. We should probably only have 1 canonical shorthand per command. Doesn't look like we *try* to implement this either.

The CLA requirement should be taken care of. Since it's not managed by Circle CI, I think to reset the CI check we should just push a dummy change and then force-push it away - and then verify that it passes and that I didn't screw it up...

PR #3185, opened 2021-11-11
feat: jq api report experiment username and progress

hamidzr:
## Description

- report experiment progress for jobs
- get experiment owner's username from db and keep in memory. 

## Commentary (optional)

we're going toward relying on username instead of id as the unique key for users

## Checklist

### DIFF
--- a/master/internal/db/postgres_experiments.go
+++ b/master/internal/db/postgres_experiments.go
@@ -1226,6 +1226,7 @@ FROM experiments e, trials t  WHERE t.id = $1 AND e.id = t.experiment_id`,
 
 // NonTerminalExperiments finds all experiments in the database whose states are not terminal.
 func (db *PgDB) NonTerminalExperiments() ([]*model.Experiment, error) {
+	// TODO get user's username + whereever else owner_id is pulled
 	rows, err := db.sql.Queryx(`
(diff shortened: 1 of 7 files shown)


master/internal/db/postgres_experiments.go
```diff
@@ -1156,9 +1156,11 @@ func (db *PgDB) ExperimentByID(id int) (*model.Experiment, error) {
 	var experiment model.Experiment
 
 	if err := db.query(`
-SELECT id, state, config, model_definition, start_time, end_time, archived,
-	   git_remote, git_commit, git_committer, git_commit_date, owner_id, job_id
-FROM experiments
+SELECT e.id, state, config, model_definition, start_time, end_time, archived,
+	   git_remote, git_commit, git_committer, git_commit_date, owner_id, notes,
+		 job_id, u.username as username
```
eecsliu, line 1161: what's the purpose for the additional selection of nodes and username?

eecsliu (approved):
lgtm! just one question

master/internal/db/postgres_experiments.go
hamidzr: in general, I think wherever we're returning the experiment model we want to fetch all the attributes it defines unless otherwise is clearly requested. Here `notes` was missing and I just added `username`

PR #3191, opened 2021-11-15
feat: jq cli interface to read the queue

hamidzr:
## Test Plan

image

## Checklist


harness/determined/common/api/b.py
```diff
@@ -0,0 +1,211 @@
+import enum
```
hamidzr, line 1: this file is generated. but not really generated here. I'm working on wrapping up the codegen with @rb-determined-ai help but this should do for now and we'll be able to just replace it later.

harness/determined/cli/job.py
```diff
@@ -0,0 +1,106 @@
...
+        "manage job",
+        [
+            Cmd(
+                "list",
+                ls,
+                "list jobs",
+                [
+                    Arg(
+                        "-o",
+                        "--output",
```
hamidzr, line 95: TODO make sure thiis style is consistent with the rest of the cli. this arg should be somethin we can share across different cli commands like pagination.

harness/determined/cli/job.py
```diff
@@ -0,0 +1,109 @@
...
+    Arg(
+        "--limit",
+        type=int,
+        default=50,
+        help="Limit the returned set.",
+    ),
+    Arg(
+        "--order-by",
+        type=str,
+        default="ORDER_BY_DESC",
```
eecsliu, line 78: this might be confusing for kubernetes users, where higher priority is better. I don't think it's a major issue though and I'm happy to keep it like this for now.

harness/determined/cli/job.py
```diff
@@ -0,0 +1,109 @@
...
+output_format = Group(
+    Arg("--csv", action="store_true", help="print as CSV"),
+    Arg("--json", action="store_true", help="print as JSON"),
+    Arg("--yaml", action="store_true", help="print as JSON"),
+    Arg("--table", action="store_true", default=True, help="print as JSON"),
+)
+
+args_description = [
+    Cmd(
+        "j|ob",
```
eecsliu, line 93: is this a typo?

eecsliu (approved):
looks great! Thanks for carrying this. Minor comments below:

harness/determined/cli/job.py
hamidzr: this is just a syntax that cmd support to define both -j and --job

harness/determined/cli/job.py
hamidzr: for this one I was trying to arrive at a set of arguments that we can reuse whereever pagination is involved. maybe a more universally applicable argument could be just `--reverse` ?

PR #3198, opened 2021-11-16
fix: use ListValue to receive empty list on labels field

mapmeld:
## Description

Allow the server to tell the difference between no list and an empty list (0 items) in the PatchModel and PatchModelVersion endpoints.

## Checklist


master/internal/api_model.go
```diff
@@ -157,8 +157,9 @@ func (a *apiServer) PatchModel(
 	currLabels := strings.Join(currModel.Labels, ",")
 	if req.Model.Labels != nil {
 		var reqLabelList []string
-		for i := 0; i < len(req.Model.Labels); i++ {
-			reqLabelList = append(reqLabelList, req.Model.Labels[i].Value)
+		for _, el := range req.Model.Labels.Values {
+			sval := el.String()
+			reqLabelList = append(reqLabelList, sval[14:len(sval)-1])
```
ioga, line 162: this should be a bit more reliable and readable:

```suggestion
			reqLabelList = append(reqLabelList, el.GetStringValue())
```

also we could check if `el.GetKind()` is stringish for extra reliability

ioga (approved):

ioga:
unrelated comment: it'd be convenient if `test_model_registry.py` could work against a "dirty" database. 
e.g. I was running that test locally against my devcluster which already has some models registered in it. Having to run against a clean one each time is a bit of a bother.

master/internal/api_model.go
```diff
@@ -157,8 +157,11 @@ func (a *apiServer) PatchModel(
 	currLabels := strings.Join(currModel.Labels, ",")
 	if req.Model.Labels != nil {
 		var reqLabelList []string
-		for i := 0; i < len(req.Model.Labels); i++ {
-			reqLabelList = append(reqLabelList, req.Model.Labels[i].Value)
+		for _, el := range req.Model.Labels.Values {
+			sval := el.String()
+			if sval[:13] == "string_value:" {
+				reqLabelList = append(reqLabelList, sval[14:len(sval)-1])
+			}
```
ioga, line 164: would something like this work?

```suggestion
            if _, ok := el.GetKind().(*structpb.Value_StringValue); ok {
                reqLabelList = append(reqLabelList, el.GetStringValue());
            }
```

mapmeld: I like this change, just trying to catch up now 😅

PR #3202, opened 2021-11-16
docs: remove note about model.predict

rb-determined-ai:
Nobody is 100% confident if the old restriction still applies, but I was
able to run the old model that was provided as a reproducing example to
DET-2728, which was the ticket that caused us to add this documentation.

So it's pretty darn likely that the restriction was secretly lifted
during one of a couple of TFKerasTrial rewrites that occurred.


shiyuann (approved):
stamped

PR #3204, opened 2021-11-16
feat: detect MIG instances in agents

mackrorysd:
## Description

This is a step toward supporting MIG instances, a feature of CUDA 11 and A100's that allows a GPU to be split into smaller virtual devices that have their own compute / memory resources. For instance, a 40GB A100 can be split into 7 5GB MIG instances. This could be useful for more granular scheduling of smaller, distinct jobs on bigger GPUs. For instance, a single A100 could be shared by 7 distinct JupyterLab instances, or 7 concurrent (smaller) trials.

This support is considered experimental only and applies only to Kubernetes clusters and static agent clusters. Distributed training is not supported.

## Test Plan

See https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html for how to do the MIG portion of this. The GPU cards need to have their MIG mode enabled and the MIG instances created before starting the Determined agent (this, and the distinction between how many physical GPUs you request and how many logical devices are then presented, is what's missing from supporting dynamic agents). 

I tested on a cluster with 2 static agents. Each agent had 2 A100 split into 7 MIG instances each (profile 19). I confirmed that the MIG slots were reported and accounted for correctly (although static agents always report -1 for "Slots per node" and I have filed a separate issue to improve this). I confirmed that distributed training is not possible on PyTorch or TensorFlow because of APIs that deal exclusively with physical devices even in the presence of a MIG configuration. I don't believe distributed training is a very useful thing to use MIG instances for anyway. I have run other jobs including simple experiments and hyperparameter search jobs and confirmed the additional slots are scheduled as expected. I have run GPU notebooks and confirmed that you get the number of MIG instances passed through to the JupyterLab that you select as "GPU slots".

I have also run all the upstream GPU tests in Circle CI and done manual testing locally to confirm non-MIG configurations still behave the same. I have previously run similar tests on a GKE cluster with daemonsets to configure MIG, etc. in an identical configuration. NVIDIA's documentation of setting this up is here: https://docs.nvidia.com/datacenter/cloud-native/kubernetes/mig-k8s.html.

## Commentary

Again, this support is experimental. Not every feature is supported, and the workloads for which this is useful need to be better understood and tested. This feature will be activated if MIG is enabled and MIG instances are created, otherwise it will fall back to using physical GPUs only.

The assumption that all nodes and GPUs within a resource pool are homogeneous still applies: you cannot use MIG instances of differing sizes (e.g. the 4/3 configuration), and you cannot use MIG on only a subset of GPUs within a node or on only a subset of nodes within a resource pool.

## Checklist

### DIFF
--- a/agent/internal/detect.go
+++ b/agent/internal/detect.go
@@ -1,10 +1,12 @@
 package internal
 
 import (
+	"bufio"
 	"encoding/csv"
 	"fmt"
 	"io"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


mackrorysd:
Upstream tests: https://app.circleci.com/pipelines/github/determined-ai/determined?branch=static-mig

Actually I see there's 1 failure I'll need to go look into.

agent/internal/detect.go
```diff
@@ -71,11 +73,69 @@ func detectCPUs() ([]device.Device, error) {
...
+				continue
+			}
+			brand := matches[1]
+			uuid := matches[2]
+			devices = append(devices,
+				device.Device{ID: deviceIndex, Brand: brand, UUID: uuid, Type: device.GPU})
+			deviceIndex++
+		}
+	}
+	return devices, err
```
stoksc, line 129: i prefer explicit `return devices, nil`. save the next reader of needing to reparse to make sure it's always nil.

agent/internal/detect.go
```diff
@@ -71,11 +73,69 @@ func detectCPUs() ([]device.Device, error) {
 	}
 }
 
+var detectMIGEnabled = []string{
+							"nvidia-smi", "--query-gpu=mig.mode.current", "--format=csv,noheader"}
+var detectNvidiaDevices = []string{"nvidia-smi", "-L"} // Lists both GPUs and MIG instances
```
stoksc, line 78: this is a bit unfortunate. i saw `nvidia-smi mig` has some flags for listing but i didn't see anything that supports `--format=csv`, and i can't seem to find any evidence that you can get it from `--query-gpu`..

agent/internal/detect.go
```diff
@@ -71,11 +73,69 @@ func detectCPUs() ([]device.Device, error) {
...
+			deviceIndex++
+		}
+	}
+	return devices, err
+}
+
 // detectGPUs returns the list of available Nvidia GPUs.
 func detectGPUs(visibleGPUs string) ([]device.Device, error) {
+	devices, err := detectMigInstances(visibleGPUs)
+	if err == nil && devices != nil && len(devices) > 0 {
```
stoksc, line 135: if we err when we should truly be detecting mig devices and fallthrough and detect normal devices, what's the behavior (of running on a full device with mig enabled)? same?

stoksc (commented):
neat. i may give the scheduler a quick read to think over implications if i have time.

agent/internal/detect.go
mackrorysd: Yeah, I found a lot of inconsistencies when I was trying to see alternative ways of doing this in the CLI. The --id option doesn't get parsed consistently either: --id=[value] is used in one of our commands, but others can't read it unless there's a space intsead of an equals.

agent/internal/detect.go
mackrorysd: It depends a little bit on how far you get. I believe you can use the physical GPU like normal even if MIG is enabled. But if MIG instances exist and you don't try using them, I'm actually not sure what happens. So we should definitely at least log the error here, and as a semi-separate issue I like the idea we discussed of rejecting agents if something about what they're reporting is inconsistent with the rest of the resource pool.

stoksc: Interesting - i was testing on my local (home) with a nvidia card. I'll play with mig then. Ok well if it will still work, i just won't be exactly what the user expects, and it usually won't happen unless the format of nvidia-smi changes (i hope they know the whole world is just parsing text), it's fine.

stoksc (approved):
lgtm.

agent/internal/detect.go
stoksc: nit, `devices != nil && len(devices) > 0` => `len(devices) > 0`

PR #3210, opened 2021-11-17
chore: remove hvd_config usage [DET-6220]

azhou-determined:
## Description

Separating this change from native pytorch dtrain support.

## Checklist

### DIFF
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -1,8 +1,8 @@
-from determined._generic import DistributedContext
-from typing import Callable, Optional
+from typing import Callable
 
 import determined as det
-from determined import _data_layer, horovod
+from determined import _data_layer
+from determined._generic import DistributedContext
 
 
 class DataLayerContext:
@@ -17,7 +17,7 @@ class DataLayerContext:
             env=env,
             training=False,
             per_slot_batch_size=env.per_slot_batch_size,
-            distributed_context=distributed_context
+            distributed_context=distributed_context,
         )
 
     def cache_train_dataset(
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -2,11 +2,11 @@ from typing import Callable
 
 import determined as det
 from determined import _data_layer
-from determined._generic import DistributedContext
+import determined._generic
 
 
 class DataLayerContext:
-    def __init__(self, env: det.EnvContext, distributed_context: DistributedContext) -> None:
+    def __init__(self, env: det.EnvContext, distributed_context: determined._generic.DistributedContext) -> None:
         self._training_cacheable = _data_layer._CacheableDecorator(
             env=env,
             training=True,
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -1,3 +1,4 @@
+from determined._generic import DistributedContext
 from typing import Callable, Optional
 
 import determined as det
@@ -5,18 +6,18 @@ from determined import _data_layer, horovod
 
 
 class DataLayerContext:
-    def __init__(self, env: det.EnvContext, distributed_backend: Optional[str]) -> None:
+    def __init__(self, env: det.EnvContext, distributed_context: DistributedContext) -> None:
         self._training_cacheable = _data_layer._CacheableDecorator(
             env=env,
             training=True,
             per_slot_batch_size=env.per_slot_batch_size,
-            distributed_backend=distributed_backend,
+            distributed_context=distributed_context,
         )
         self._validation_cacheable = _data_layer._CacheableDecorator(
             env=env,
             training=False,
             per_slot_batch_size=env.per_slot_batch_size,
-            distributed_backend=distributed_backend
+            distributed_context=distributed_context
         )
 
     def cache_train_dataset(
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -1,13 +1,12 @@
 from typing import Callable
 
 import determined as det
-import determined._generic
-from determined import _data_layer
+from determined import _data_layer, _generic
 
 
 class DataLayerContext:
     def __init__(
-        self, env: det.EnvContext, distributed_context: determined._generic.DistributedContext
+        self, env: det.EnvContext, distributed_context: _generic.DistributedContext
     ) -> None:
         self._training_cacheable = _data_layer._CacheableDecorator(
             env=env,
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -1,12 +1,14 @@
 from typing import Callable
 
 import determined as det
-from determined import _data_layer
 import determined._generic
+from determined import _data_layer
 
 
 class DataLayerContext:
-    def __init__(self, env: det.EnvContext, distributed_context: determined._generic.DistributedContext) -> None:
+    def __init__(
+        self, env: det.EnvContext, distributed_context: determined._generic.DistributedContext
+    ) -> None:
         self._training_cacheable = _data_layer._CacheableDecorator(
             env=env,
             training=True,
--- a/harness/determined/_data_layer/_context.py
+++ b/harness/determined/_data_layer/_context.py
@@ -1,22 +1,22 @@
-from typing import Callable
+from typing import Callable, Optional
 
 import determined as det
 from determined import _data_layer, horovod
 
 
 class DataLayerContext:
-    def __init__(self, env: det.EnvContext, hvd_config: horovod.HorovodContext) -> None:
+    def __init__(self, env: det.EnvContext, distributed_backend: Optional[str]) -> None:
         self._training_cacheable = _data_layer._CacheableDecorator(
             env=env,
-            hvd_config=hvd_config,
             training=True,
             per_slot_batch_size=env.per_slot_batch_size,
+            distributed_backend=distributed_backend,
         )
         self._validation_cacheable = _data_layer._CacheableDecorator(
             env=env,
-            hvd_config=hvd_config,
             training=False,
             per_slot_batch_size=env.per_slot_batch_size,
+            distributed_backend=distributed_backend
         )
 
     def cache_train_dataset(
--- a/harness/determined/_data_layer/_data_layer.py
+++ b/harness/determined/_data_layer/_data_layer.py
@@ -35,12 +35,12 @@ class _CacheableDecorator:
     def __init__(
         self,
         env: det.EnvContext,
-        hvd_config: horovod.HorovodContext,
         training: bool,
         per_slot_batch_size: int,
+        distributed_backend: Optional[str]
     ) -> None:
         self._env = env
-        self._hvd_config = hvd_config
+        self._distributed_backend = distributed_backend
         self._training = training
         self._per_slot_batch_size = per_slot_batch_size
 
@@ -63,7 +63,7 @@ class _CacheableDecorator:
         self._offset = self._env.latest_batch * batch_size
 
     def _init_shard(self) -> None:
-        if not self._hvd_config.use:
+        if not self._distributed_backend == "horovod":
             return
 
         self._shard_rank = hvd.rank()
@@ -71,7 +71,7 @@ class _CacheableDecorator:
 
     def _configure_storage(self) -> None:
         session_config = None  # type: Optional[tf.compat.v1.ConfigProto]
-        if self._hvd_config.use:
+        if self._distributed_backend == "horovod":
             # For multi-GPU training, we map processes to individual GPUs. TF requires
             # that for each instantiation of `tf.Session`, the process is mapped
             # to the same GPU.
--- a/harness/determined/_data_layer/_data_layer.py
+++ b/harness/determined/_data_layer/_data_layer.py
@@ -2,6 +2,8 @@ import enum
 import functools
 import logging
 import pathlib
+
+from determined._generic import DistributedContext
 from typing import Any, Callable, Optional, cast
 from urllib import parse
 
@@ -37,10 +39,10 @@ class _CacheableDecorator:
         env: det.EnvContext,
         training: bool,
         per_slot_batch_size: int,
-        distributed_backend: Optional[str]
+        distributed_context: DistributedContext
     ) -> None:
         self._env = env
-        self._distributed_backend = distributed_backend
+        self._distributed_context = distributed_context
         self._training = training
         self._per_slot_batch_size = per_slot_batch_size
 
@@ -63,20 +65,20 @@ class _CacheableDecorator:
         self._offset = self._env.latest_batch * batch_size
 
     def _init_shard(self) -> None:
-        if not self._distributed_backend == "horovod":
+        if not self._distributed_context.size > 1:
             return
 
-        self._shard_rank = hvd.rank()
-        self._num_shards = hvd.size()
+        self._shard_rank = self._distributed_context.rank
+        self._num_shards = self._distributed_context.size
 
     def _configure_storage(self) -> None:
         session_config = None  # type: Optional[tf.compat.v1.ConfigProto]
-        if self._distributed_backend == "horovod":
+        if self._distributed_context.size > 1:
             # For multi-GPU training, we map processes to individual GPUs. TF requires
             # that for each instantiation of `tf.Session`, the process is mapped
             # to the same GPU.
             session_config = tf.compat.v1.ConfigProto()
-            session_config.gpu_options.visible_device_list = str(hvd.local_rank())
+            session_config.gpu_options.visible_device_list = str(self._distributed_context.local_rank)
 
         parsed = parse.urlparse(self._env.master_url)
         scheme = "wss" if parsed.scheme == "https" else "ws"
--- a/harness/determined/_data_layer/_data_layer.py
+++ b/harness/determined/_data_layer/_data_layer.py
@@ -2,8 +2,6 @@ import enum
 import functools
 import logging
 import pathlib
-
-from determined._generic import DistributedContext
 from typing import Any, Callable, Optional, cast
 from urllib import parse
 
@@ -12,9 +10,8 @@ import yogadl
 from yogadl import storage, tensorflow
 
 import determined as det
-from determined import horovod
+from determined._generic import DistributedContext
 from determined.common import check
-from determined.horovod import hvd
 
 
 def init_container_storage_path(configured_storage_path: Optional[str]) -> pathlib.Path:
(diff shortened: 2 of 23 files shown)


harness/determined/estimator/_estimator_context.py
```diff
@@ -40,13 +40,21 @@ def __init__(self, *arg: Any, **kwarg: Any) -> None:
 
         self.experimental = EstimatorExperimentalContext(
             env=self.env,
-            hvd_config=self.hvd_config,
             parent=self,
+            distributed_backend=self.distributed_backend
         )
 
+        if self.distributed_backend == "horovod":
```
rb-determined-ai, line 47: EstimatorTrial can't run with non-horovod distributed backend.  If you ran it with a different dtrain backend, what you have here would silently exhibit weird behavior.  The check you have here follows more closely to the pre-existing code paths, which I assume is why you implemented it this way.  But the pre-existing code paths implicitly assumed that we only ever ran horovod train.  Now that the system no longer enforces that restriction globally, what you have here doesn't make sense in isolation.

As it is right now, if I read this in isolation, I would say, "why does average_training_metrics, something that is common to all dtrain, not apply with to a tensorflow-native distributed backend?  Why only horovod?"

I of course don't want you to add non-horovod support to EstimatorTrial.  I would recommend writing sections like this as `if self.context.distributed.size > 1` (which is equivalent, since only horovod is supported) and I would simply add a check in harness.py that throws an error if somebody tries to launch EstimatorTrial or TFKerasTrial without a horovod launch layer.  With this strategy, there's no need to pass `distributed_backend` into EstimatorTrial or TFKerasTrial at all.

harness/determined/exec/harness.py
```diff
@@ -75,9 +75,10 @@ def main(chief_ip: Optional[str]) -> int:
     )
 
     multi_machine_trial = len(info.container_addrs) > 1
-    hvd_config = horovod.HorovodContext.from_configs(
-        env.experiment_config, env.hparams, multi_machine_trial
-    )
+    multi_slot_trial = (env.experiment_config.slots_per_trial() > 1
+                        and not env.experiment_config.native_parallel_enabled())
+
+    distributed_backend = "horovod" if multi_machine_trial or multi_slot_trial else None
```
rb-determined-ai, line 81: This looks like clear evidence that you could use `if context.distributed.size > 1` everywhere that `if hvd_config.use` is used

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -82,6 +82,12 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._reducers = pytorch._PyTorchReducerContext()
         self._determined_profiler = None  # type: Optional[profiler.ProfilerAgent]
 
+        optimizations_config = self.get_optimizations_config()
+        self.aggregation_frequency = cast(int, optimizations_config.get("aggregation_frequency"))
+        self.fp16_compression = cast(bool, optimizations_config.get("gradient_compression"))
+        self.average_aggregated_gradients = cast(bool, optimizations_config.get("average_aggregated_gradients"))
+        self.average_training_metrics = cast(bool, optimizations_config.get("average_training_metrics"))
```
rb-determined-ai, line 89: Do you mean for these to be user-facing APIs?  If not they need to each have an underscore prefix.

There are some old attributes that didn't get this treatment but also aren't really public APIs, like context.env and context.experiment_config, but we should be consistent with new values.

rb-determined-ai (commented):
Ultimately, this pattern of having a `distributed_backend` which is a string that everything in our code knows how to interpret is the reverse of what we want to build.

This pattern means that all of the objects in our code need to know how to deal with each value that `distributed_backend` could take.  That's a non-extensible interface, because users might be able to set a different string but they couldn't teach all of the objects in our code how to deal with it.

A better pattern is to write an interface that each backend could provide, and let all of the objects in our code accept an object that implements that interface.  That way, if a user wants to extend the system, they could implement that interface themselves, and our system would continue to operate normally.

In this case, you don't need to introduce a new interface type (please don't, in fact); I'm pretty sure you can get away with just passing the `DistributedContext` around, since that provides most of the features that we need, at least for side-features like the data layer.

This leaves one important exception though: how will you deal with PyTorchTrial, where you do need to know backend information?  Any ideas?

azhou-determined:
> A better pattern is to write an interface that each backend could provide, and let all of the objects in our code accept an object that implements that interface. That way, if a user wants to extend the system, they could implement that interface themselves, and our system would continue to operate normally.
> 

> In this case, you don't need to introduce a new interface type (please don't, in fact); I'm pretty sure you can get away with just passing the `DistributedContext` around, since that provides most of the features that we need, at least for side-features like the data layer.
> 

agreed. after some experimentation of different implementations for an interface, I've settled on the environment variable approach. 

each launch layer will set an environment variable flag (USE_HOROVOD = True/False and subsequently USE_PYTORCH_DTRAIN) indicating which distributed backend is being used, which will be read/interpreted by the harness and trial classes. 

this is nice because it logically separates out the code concerns of distributed backend and multi-GPU training: i.e. we could be running a horovod backend but only using one GPU (USE_HOROVOD = True and context.distributed.size = 1)

for features like data layer that only need to know whether multi-GPU training is being used, passing in the DistributedContext is enough. for class methods like pre_execute_hook, it only needs to know which backend(s) is being run.

this will also allow "layering" of launch layers in the future, like if you wanted to run torch distributed along with a custom profiling layer or etc, while keeping if/else statements clean-ish and to a minimum.

while setting too many environment variables can get messy, i think this is the best approach moving forward. other implementations get tangled up with various trial and distributed contexts and inevitably get glued together with scattered if/else logic.

harness/determined/_trial_controller.py
```diff
@@ -1,12 +1,17 @@
...
 from typing import Any, Optional, Type
-
+from distutils.util import strtobool
+from enum import Enum
 import determined as det
 from determined import horovod, profiler, workload
 from determined.common import check
 from determined.horovod import hvd
 
+class DistributedBackend(Enum):
```
azhou-determined, line 13: this needs a better place to live. I don't feel like it belongs in trial context or harness though

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -212,7 +212,7 @@ def _configure_session(
             gpus = tf.config.experimental.list_physical_devices("GPU")
 
             if len(gpus) > 0:
-                local_rank = hvd.local_rank() if hvd_config.use else 0
+                local_rank = hvd.local_rank() if use_horovod else 0
```
rb-determined-ai, line 215: just use `self.context.distributed.local_rank` here, it'll be easier.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -507,7 +506,7 @@ def _configure_callbacks(self, user_callbacks: Optional[List]) -> None:
         )
         callbacks = [self.multiplexer]
 
-        if self.hvd_config.use:
+        if self.context.distributed.size > 1 and self.use_horovod:
```
rb-determined-ai, line 509: as with estimator trial, I'd prefer you put an assert somewhere that only allows dtrain tfkeras when horovod is in use.

If we ever support non-horovod dtrain tfkeras then we'll already have to go through the code with a fine-tooth comb, and these if statements where we try to predict what the right behavior would be if we supported non-horovod dtrain tfkeras would be more likely to cause quiet bugs than to be helpful.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -820,7 +819,7 @@ def _control_loop(self) -> None:
         raise det.errors.WorkerFinishedGracefully()
 
     def _allreduce_logs(self, logs: Dict) -> Dict:
-        if not self.hvd_config.use:
+        if not (self.context.distributed.size > 1 and self.use_horovod):
```
rb-determined-ai, line 822: ditto

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -877,7 +876,7 @@ def _post_train_batch_end(self, num_inputs: int, logs: Dict) -> None:
                 "as this will affect Determined training behavior",
             )
 
-        if self.hvd_config.use:
+        if self.context.distributed.size > 1 and self.use_horovod:
```
rb-determined-ai, line 879: ditto

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -914,7 +913,7 @@ def _compute_validation_metrics(self) -> workload.Response:
         metrics = self._launch_evaluate()
         num_inputs = self.multiplexer.get_test_inputs()
 
-        if self.hvd_config.use:
+        if self.context.distributed.size > 1 and self.use_horovod:
```
rb-determined-ai, line 916: ditto

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -197,13 +207,12 @@ def train_batch(
                 "backward_passes_per_step for local gradient aggregation must be >= 1",
             )
 
-            if self.hvd_config.use:
-                use_compression = self.hvd_config.fp16_compression
+            if self.distributed.size > 1:
+                use_compression = self._fp16_compression
```
rb-determined-ai, line 211: this line seems pointless.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -426,17 +429,19 @@ def _train_for_step(
...
+        if (
+            self.context.distributed.size > 1
+            and self.use_horovod
+            and self.context._average_training_metrics
+        ):
             with self.prof.record_timing("average_training_metrics"):
                 per_batch_metrics = self._average_training_metrics(per_batch_metrics)
-        if self.hvd_config.use:
-            num_inputs *= hvd.size()
+        num_inputs *= self.context.distributed.get_size()
```
rb-determined-ai, line 444: just use `self.context.distributed.size` (you have quite a few of these)

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -426,17 +429,19 @@ def _train_for_step(
...
-                samples_per_second *= hvd.size()
+            samples_per_second *= self.context.distributed.get_size()
             self.prof.record_metric("samples_per_second", samples_per_second)
             per_batch_metrics.append(tr_metrics)
 
         # Aggregate and reduce training metrics from all the training processes.
-        if self.hvd_config.use and self.hvd_config.average_training_metrics:
+        if (
+            self.context.distributed.size > 1
+            and self.use_horovod
```
rb-determined-ai, line 439: Same thing about trying to distinguish horovod dtrain from non-horovod dtrain.  As evidence of my claim that "trying to predict the horovod codepaths and the non-horovod codepaths before you actually add a second dtrain backend will mostly cause bugs", you have not been consistent in these checks even in this PR.  For example, context.wrap_optimizer uses horovod but does not have protection.  Having some checks now is more of a false sense of security, IMO.

However, since you are actively working on non-horovod dtrain for pytorch trial, it's really up to you if you want to get rid of these horovod-specific checks with pytorch trial.  You do have enough background to make correct decisions here, which will be meaningful in the near future.  You can take my suggestions here or not.  With TFKeras or Estimator though, I would definitely like you to remove the horovod-specific checks and just have assertions that only the horovod backend is in use.  I really doubt that we ever need to revisit those trial classes, but if we do, it would be better to start with a blank slate, so to speak.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -268,7 +268,10 @@ def _average_training_metrics(
         self, per_batch_metrics: List[Dict[str, Any]]
     ) -> List[Dict[str, Any]]:
         """Average training metrics across GPUs"""
-        check.true(self.hvd_config.use, "Can only average training metrics in multi-GPU training.")
+        check.true(
```
rb-determined-ai, line 271: same as before on replacing `check.true` with an assertion.  If you are touching this line, please make this update as well.

rb-determined-ai (commented):
Overall, looks pretty good.  Lots of pretty small feedback items, with only one incorrect thing that I could find.

harness/determined/pytorch/_pytorch_trial.py
azhou-determined: since I want to get this change out before introducing native dtrain for pytorch, I went with your suggestion of removing horovod checks and asserting horovod must be used if distributed training.

harness/determined/keras/_tf_keras_trial.py
azhou-determined: this method is a @classmethod, can't use self... methods for class instances.

harness/determined/_trial_controller.py
```diff
@@ -1,11 +1,17 @@
 import abc
 import logging
+import os
+from distutils.util import strtobool
+from enum import Enum
```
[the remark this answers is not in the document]
azhou-determined: why?

harness/determined/estimator/_estimator_trial.py
```diff
@@ -632,20 +632,20 @@ def _init_session_config(
...
-        if not hvd_config.use:
+        if not use_horovod:
             return session_config
 
         if version.parse(tf.__version__) >= version.parse("2.5.0"):
             gpus = tf.config.experimental.list_physical_devices("GPU")
 
             if len(gpus) > 0:
-                local_rank = hvd.local_rank() if hvd_config.use else 0
+                local_rank = hvd.local_rank()
```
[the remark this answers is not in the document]
azhou-determined: can't access self instance variables on a @classmethod

harness/determined/_trial_controller.py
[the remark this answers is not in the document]
rb-determined-ai: please read the google style guide for imports, as linked before.

harness/determined/_data_layer/_context.py
```diff
@@ -1,22 +1,25 @@
 from typing import Callable
 
 import determined as det
-from determined import _data_layer, horovod
+import determined._generic
```
rb-determined-ai, line 4: from determined import _generic

Prefer to import the final module as a single name.

harness/determined/_data_layer/_data_layer.py
```diff
@@ -10,9 +10,8 @@
 from yogadl import storage, tensorflow
 
 import determined as det
-from determined import horovod
+from determined._generic import DistributedContext
```
rb-determined-ai, line 13: from determined import _generic
...
_generic.DistributedContext

Prefer to import the final module as a single name (and only import modules, not symbols)

harness/determined/_trial_controller.py
```diff
@@ -1,11 +1,16 @@
...
 from typing import Any, Optional, Type
 
 import determined as det
-from determined import horovod, profiler, workload
+from determined import profiler, workload
 from determined.common import check
-from determined.horovod import hvd
+
+
+class _DistributedBackend(enum.Enum):
```
rb-determined-ai, line 12: Please see the previous comment about this being an enum.  I marked it as unresolved.

Also, please don't mark something as resolved unless you are 100% sure you've fixed it.  IMO it's better not to mark something as resolved until the review has had a chance to see how you've fixed it, then the reviewer can mark it as resolved.  But not everybody agrees with that.

But definitely don't mark something as resolved when it's not fixed.

harness/determined/estimator/_estimator_context.py
```diff
@@ -1,11 +1,12 @@
...
 import logging
-from typing import Any, Callable, List, Tuple, Union
+from typing import Any, Callable, List, Tuple, Union, cast
 
 import tensorflow as tf
 
 import determined as det
-from determined import _data_layer, estimator, horovod, util
+from determined import _data_layer, estimator, util
+from determined._generic import DistributedContext
```
rb-determined-ai, line 9: from determined import _generic
...
_generic.DistributedContext

harness/determined/estimator/_estimator_trial.py
```diff
@@ -404,10 +409,10 @@ def __init__(
     def pre_execute_hook(
         cls: Type["EstimatorTrialController"],
         env: det.EnvContext,
-        hvd_config: horovod.HorovodContext,
+        use_horovod: bool,
```
rb-determined-ai, line 412: I can't recall anymore what your last version had here.  It seems like probably you want to use the new _DistributedBackend here maybe?  Seems to carry the information you're trying to portray here?

I'd consider this a "non-blocking" review comment.  Ignore it if you wish.

harness/determined/keras/_tf_keras_context.py
```diff
@@ -1,11 +1,12 @@
...
 import logging
-from typing import Any, Callable, Dict, List, NamedTuple, Optional, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Union, cast
 
 import tensorflow as tf
 
 import determined as det
-from determined import _data_layer, errors, horovod, keras
+from determined import _data_layer, errors, keras
+from determined._generic import DistributedContext
```
rb-determined-ai, line 9: from determined import _generic
...
_generic.DistributedContext

rb-determined-ai (approved):
Very close now.

I'll approve this since I'm on vacation, but please do make the touchups I suggested.

Also before you land this, since you are touching so much of the multi-gpu codepath you need to run the upstream tests first:
* create an upstream PR (push to master/deprecate-hvd-config).  Make sure to alter the commit hash before pushing (git commit --amend) because sometimes I've seen circleci be weird if there are two prs with identical commit hashes trying to run tests
* Enable all of the multi-gpu tests manually.  Don't bother with the nightly or convergence tests or weekly tests or whatever they're called.  Or the model hub tests.  Those are unreasonably slow for this.
* Wait for those to pass on the upstream PR, then land this PR (with identical code, of course).  That is better for the git history, since this PR will have all of the review comments.
* If you have any questions about this process, ask @ioga.

azhou-determined:
> 

Made suggested changes and deploying upstream tests. GKE is failing for unrelated reasons, but other GPU tests seem to be running fine. Going to monitor/rerun a few times, and if all looks well, I will merge.

azhou-determined:
> Very close now.
> 
> I'll approve this since I'm on vacation, but please do make the touchups I suggested.
> 
> Also before you land this, since you are touching so much of the multi-gpu codepath you need to run the upstream tests first:
> 
> * create an upstream PR (push to master/deprecate-hvd-config).  Make sure to alter the commit hash before pushing (git commit --amend) because sometimes I've seen circleci be weird if there are two prs with identical commit hashes trying to run tests
> * Enable all of the multi-gpu tests manually.  Don't bother with the nightly or convergence tests or weekly tests or whatever they're called.  Or the model hub tests.  Those are unreasonably slow for this.
> * Wait for those to pass on the upstream PR, then land this PR (with identical code, of course).  That is better for the git history, since this PR will have all of the review comments.
> * If you have any questions about this process, ask @ioga.

upstream tests are passing here (https://app.circleci.com/pipelines/github/determined-ai/determined/17468/workflows/e635e4d4-a4b9-491a-8513-c8160cc79a7c). some intermittent failures on GKE clusters spinning up, but seems like no errors related to this change. going to land this.

PR #3213, opened 2021-11-18
feat: model registry API can update name, has Notes field

mapmeld:
## Description

In this updated PR, we make it possible to edit the Name field in PATCH /models/:id.  Re-add the Notes column to Models.

## Checklist

### DIFF
--- a/master/static/srv/get_models.sql
+++ b/master/static/srv/get_models.sql
@@ -4,7 +4,7 @@ LEFT JOIN users as u ON u.id = m.user_id
 WHERE ($1 = '' OR m.archived = $1::BOOL)
 AND ($2 = '' OR (u.username IN (SELECT unnest(string_to_array($2, ',')))))
 AND ($3 = '' OR (m.labels <@ string_to_array($3, ',')))
-AND ($4 = '' OR LOWER(m.name) LIKE $4)
-AND ($5 = '' OR LOWER(m.description) LIKE $5)
+AND ($4 = '' OR LOWER(m.name) = $4)
+AND ($5 = '' OR (LOWER(m.description) LIKE CONCAT($6::text, $5, $6::text)))
 GROUP BY m.id, u.id
 ORDER BY %s;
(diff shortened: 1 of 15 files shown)


master/static/srv/get_models.sql
```diff
@@ -1,10 +1,10 @@
...
+SELECT m.id, m.name, m.description, m.notes, m.metadata, m.creation_time, m.last_updated_time, array_to_json(m.labels) AS labels, u.username, m.archived, COUNT(mv.version) as num_versions FROM models as m
 LEFT JOIN model_versions as mv ON mv.model_id = m.id
 LEFT JOIN users as u ON u.id = m.user_id
 WHERE ($1 = '' OR m.archived = $1::BOOL)
 AND ($2 = '' OR (u.username IN (SELECT unnest(string_to_array($2, ',')))))
 AND ($3 = '' OR (m.labels <@ string_to_array($3, ',')))
-AND ($4 = '' OR LOWER(m.name) LIKE $4)
-AND ($5 = '' OR LOWER(m.description) LIKE $5)
+AND ($4 = '' OR LOWER(m.name) = $4)
+AND ($5 = '' OR (LOWER(m.description) LIKE CONCAT($6::text, $5, $6::text)))
```
ioga, line 8: the only time you're using $6, it's `%`. I don't really see how $6 can be any other value in a useful way. 
could this just be hardcoded to `ILIKE ('%%' || $5 || '%%')`?

ioga: come to think of it, postgres has `ILIKE` as case-insensitive LIKE. so we could 
- skip lower here
- not call ToLower in the go code for the input parameter.

ioga (commented):
high-level question: it's weird some parameters are exact match (e.g. name) and others are substring. so there's no way to tell how it works by looking at it, one'll have to read docs or code which is confusing. perhaps we should do something about that?

Couple options:
- `GetExperiments` API just treats everything as a substring pattern match, including the experiment name. It'd be somewhat consistent to do the same here.
- We could allow users to provide a pattern in the same parameter. or have separate query parameters for equality and substr matches.

master/static/srv/get_models.sql
mapmeld: oh this was only an issue where the SQL builder didn't know what to do with '%', if I can do %% then that's better

mapmeld:
I've edited the PR to reflect a more limited change to the `models` table and the route `PATCH /models/:id`

master/static/srv/update_model.sql
```diff
@@ -1,3 +1,3 @@
-UPDATE models SET description = $2, metadata = $3, labels = string_to_array($4, ','), last_updated_time = current_timestamp
+UPDATE models SET name = $2, description = $3, notes= $4, metadata = $5, labels = string_to_array($6, ','), last_updated_time = current_timestamp
```
ioga, line 1:
```suggestion
UPDATE models SET name = $2, description = $3, notes = $4, metadata = $5, labels = string_to_array($6, ','), last_updated_time = current_timestamp
```

ioga (approved):

PR #3214, opened 2021-11-18
feat: Adding Openshift route support in Helm chart

mluyo3414:
## Description

Adding Openshift 4.X route setup for Helm chart.

## Test Plan

- Change `determined/helm/charts/determined/values.yaml` to create route:

```
route:
 enabled: true
 termination: edge
```
- Validate route was created:
```
oc get routes -n $your_namespace
```
- Verify Determined master is accessible from the route.

## Commentary 

More variables can be added if needed.

## Checklist

### DIFF
--- a/helm/charts/determined/values.yaml
+++ b/helm/charts/determined/values.yaml
@@ -19,6 +19,11 @@ masterPort: 8080
 # load-balancer.
 useNodePortForMaster: false
 
+# Enable route support for Openshift by setting enabled to true. Configure tls termination (i.e edge) if needed.
+# route:
+  # enabled: 
+  # termination: 
+
 # tlsSecret enables TLS encryption for all communication made to the Determined master (TLS
 # termination is performed in the Determined master). This includes communication between the
 # Determined master and the task containers it launches, but does not include communication between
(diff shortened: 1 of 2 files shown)


rb-determined-ai:
@cla-bot[bot] check

helm/charts/determined/values.yaml
```diff
@@ -19,6 +19,11 @@ masterPort: 8080
 # load-balancer.
 useNodePortForMaster: false
 
+# Enable route support for Openshift by setting enabled to true. Configure tls termination (i.e edge) if needed.
+# route:
```
mackrorysd, line 23: Mostly this looks good. I lack any context on exactly how we would test this, etc. but it's a no-op if you don't enable it, so it's not going to cause any regressions, etc.

I do feel like we ought to namespace this under "openshift" somehow since it's an OpenShift-specific API that it's enabling the use of. Maybe openShiftRoute: instead of route: Thoughts?

mackrorysd (changes requested):

helm/charts/determined/values.yaml
mluyo3414: that makes sense. I will change it for openshiftRoute

mackrorysd (approved):

PR #3218, opened 2021-11-19
feat: generic_context.searcher.epochs() [DET-6083]

rb-determined-ai:
If generic_context.searcher.ops() is the advanced way to use the
searcher, generic_context.searcher.epochs() is the basic way.

## Commentary

This is basically exactly what @stoksc proposed before we started implementing push architecture.  Thanks, Bradley!


rb-determined-ai:
deleting this, as @liamcli pointed out that the searcher operations really ought to be unitless.  I'll be promoting the "advanced searcher api" to being just the "only searcher api" for now.

PR #3219, opened 2021-11-19
feat: don't delete image before force_pull_image [DET-6145]

rb-determined-ai:
Tested manually.

### DIFF
--- a/docs/release-notes/3219-force-pull-image.txt
+++ b/docs/release-notes/3219-force-pull-image.txt
@@ -1,0 +1,8 @@
+:orphan:
+
+**New Features**
+
+-  The ``environment.force_pull_image: true`` behavior has improved when running on Determined
+   Agents. Previously, it would delete the image if it existed before re-pulling it. This behavior
+   was wasteful of network resources and execution time. Now, it will pull without deleting
+   anything, effectively pulling only the updated layers of the image.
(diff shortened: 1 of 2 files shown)


docs/release-notes/3219-force-pull-image.txt
```diff
@@ -0,0 +1,8 @@
+:orphan:
+
+**New Features**
+
+-  The ``environment.force_pull_image: true`` behavior has improved when running on Determined
+   Agents. Previously, it would attempt delete the image if it existed before re-pulling it. This
```
stoksc, line 6: this plus, i could take or leave the "this behavior was wasteful ..." sentence
```suggestion
   Agents. Previously, it would delete the image if it existed before re-pulling it. This
```

docs/release-notes/3219-force-pull-image.txt
```diff
@@ -0,0 +1,8 @@
...
+   behavior was wasteful of network resources and execution time. Now, it will attempt a pull
```
stoksc, line 7:
```suggestion
   behavior was wasteful of network resources and execution time. Now, it will pull
```

stoksc (approved):
lgtm. also independently verified the subtle pull vs run differences. i can't think of any case where this wouldn't be the desired behavior (the other is very costly).

PR #3222, opened 2021-11-22
feat: k8s exp status

eecsliu:
## Description
k8s previously did not update job status because we just throw pods at k8s and let it handle the entire lifecycle. This is a quality of life improvement for job queue that allows us to track whether the job is active or pending.

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/kubernetes_resource_manager.go
+++ b/master/internal/resourcemanagers/kubernetes_resource_manager.go
@@ -29,7 +29,7 @@ type kubernetesResourceManager struct {
 
 	reqList           *taskList
 	groups            map[*actor.Ref]*group
-	addrToContainerID map[*actor.Ref]cproto.ID
+	containerIDtoAddr map[string]*actor.Ref
 	slotsUsedPerGroup map[*group]int
 
 	// Represent all pods as a single agent.
@@ -53,7 +53,7 @@ func newKubernetesResourceManager(
 
 		reqList:           newTaskList(),
 		groups:            make(map[*actor.Ref]*group),
-		addrToContainerID: make(map[*actor.Ref]cproto.ID),
+		containerIDtoAddr: make(map[string]*actor.Ref),
 		slotsUsedPerGroup: make(map[*group]int),
 
 		echoRef:         echoRef,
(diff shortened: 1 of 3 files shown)


master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -214,6 +217,21 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error
 	case sproto.ResourcesReleased:
 		k.resourcesReleased(ctx, msg.TaskActor)
 
+	case sproto.UpdatePodStatus:
+		var ref *actor.Ref
+		for addr, id := range k.addrToContainerID {
```
hamidzr, line 222: suggestion: would a reverse map of `addrToContainerID map[*actor.Ref]cproto.ID` work better here?  if it's the only usecase?

master/internal/sproto/task_actor.go
```diff
@@ -45,6 +47,12 @@ type (
 		ContainerStopped *TaskContainerStopped
 	}
 
+	// UpdatePodStatus notifies the resource manager of experiment state change.
```
hamidzr, line 50: nit: "experiment"? maybe job or task would apply better?

hamidzr (commented):
lgtm some small coments

master/internal/resourcemanagers/kubernetes_resource_manager.go
eecsliu: There will be a trade-off between using a map here and using a map during deletion. I only expect each to be used once per experiment, so the runtime will be the same, so I chose this design where the code is a bit more compact. We could use two maps as an index and reverse index. Thoughts?

master/internal/sproto/task_actor.go
eecsliu: This is only applicable to kubernetes though, would it be accurate to associate it with experiments in general?

master/internal/resourcemanagers/kubernetes_resource_manager.go
hamidzr: that make sense I didn't consider deletions

master/internal/sproto/task_actor.go
hamidzr: we chatted about this offline

hamidzr (approved):
:shipit: 
(pods update comment)

master/internal/resourcemanagers/kubernetes_resource_manager.go
eecsliu: I'll actually use a map during Update Pod Status and let deletion be a linear time operation. In the future, we may add finer grained statuses on k8s, in which case the pod status could be updated multiple times per experiment.

PR #3230, opened 2021-11-24
fix: make Helm deployment work with no OpenShift config present

dzhu:
## Description

We need to account for the case where the `openshift` key is not present
at all; Helm fails to run in that case without this change.

## Test Plan

- [x] deploy a cluster on Helm


stoksc (approved):

stoksc:
tested with `det deploy gke-experimental`, works great.

mluyo3414:
Tested on OCP 4.8.11 adding:

```
openshiftRoute:
  enabled: true
  termination: edge
```

and commenting out all three lines on GKE.

mluyo3414 (approved):

PR #3240, opened 2021-11-29
fix: stop Model Registry from showing empty message while loading

EmilyBonar:
## Checklist


hkang1 (approved):
This is good for the short term fix.

The crux of the problem is that AntD `Spin` component **always** renders the `props.children` if provided, regardless of the loading state. One thought is to add a `noContentWhileLoading` prop (or some better name) to our `Spinner` wrapper component, so that we have the option to render or not render the children when the loading state is true. We may NOT want to render it if we want to avoid the children from re-rendering when loading state changes (we have a few places where this is true).

PR #3241, opened 2021-11-30
fix: Add fields to new model versions, and protect archived models

mapmeld:
## Description

Adds name and comment to insert_model_version - these attributes are already received by the API but unused. This also prevents archived models from getting updated attributes or registering new model versions.

## Checklist


ioga (approved):
I believe you'll need to force-push the commit with its title fixed to have lowercase `fix` for the CI to redo the semantic check. Just changing the PR title on github was not sufficient in the past.

PR #3243, opened 2021-11-30
fix: add Validate methods for checkpoint_storage structs [DET-6281]

varlogtim:


master/pkg/schemas/expconf/storage_config.go
```diff
@@ -179,3 +188,23 @@ func (a AzureConfigV0) Merge(other interface{}) interface{} {
...
 }
+
+// Validate implements the check.Validatable interface.
+func (c AzureConfigV0) Validate() []error {
+	var errs []error
+	if c.RawContainer == nil {
+		errs = append(errs, errors.New("'container' must not be empty"))
+	}
+	if c.RawConnectionString != nil {
+		if c.RawAccountURL != nil || c.RawCredential != nil {
```
stoksc, line 199: These checks are a teeny bit different than jsonschema land checks I think? They check `connectionString != nil != accountURL != nil` (exactly one set) and `connectionString == nil || credential == nil` (both not set). There is a case like `connectionString = nil, accountURL = {}, credential = nil` that fails here and passes here (I mean, this is all static analysis, could've read wrong). IDK enough about azure blob store to know if that case is meaningful, maybe for public buckets or something? Thoughts?

varlogtim: Ahhh, Yes. I was going with what our docs were saying. I think we need to allow `nil` in the `credential` field per this:

https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobserviceclient?view=azure-python#from-connection-string-conn-str--credential-none----kwargs-

I will make this match the jsonschema.

varlogtim: Done.

stoksc (approved):
lgtm

PR #3244, opened 2021-11-30
feat: Model Registry UX Phase 2 - Creation [DET-6257] [DET-6258]

EmilyBonar:
## Description
This took much less time than expected.

Screenshot 2021-12-06 at 15-38-15 Experiment 2353 - Determined
Screenshot 2021-12-06 at 15-38-37 Experiment 2353 - Determined
Screenshot 2021-12-06 at 15-38-56 Experiment 2353 - Determined

## Checklist


webui/react/src/components/NewModelButton.tsx
```diff
@@ -0,0 +1,27 @@
+import { Button } from 'antd';
+import React, { useCallback, useState } from 'react';
+
+import NewModelModal from './NewModelModal';
+
+const NewModelButton: React.FC = () => {
+  const [ showNewModelModal, setShowNewModelModel ] = useState(false);
```
mapmeld, line 7: setShowNewModelModel => setShowNewModelMod**al**

webui/react/src/components/RegisterModelVersionModal.tsx
```diff
@@ -0,0 +1,203 @@
...
+        modelId: selectedModelId,
+      });
+      onCloseAll?.();
+      notification.open(
+        {
+          btn: null,
+          description: (
+            <div className={css.toast}>
+              <p>{`"${versionName || `Version ${selectedModelNumVersions + 1}`}"`} registered</p>
+              <Link path={paths.modelVersionDetails(selectedModelId, response?.id ?? 0)}>
```
mapmeld, line 79: I have two suggestions on this notification - more specific text ("View Model Version") and handling a missing response.id differently (maybe as an error?)

webui/react/src/components/RegisterModelVersionModal.tsx
```diff
@@ -0,0 +1,203 @@
...
+              View Model
+              </Link>
+            </div>),
+          message: '',
+        },
+      );
+    } catch {
+      handleError({ message: 'Unable to create model.', silent: true, type: ErrorType.Api });
```
mapmeld, line 87: For more specificity, let's make the error 'Unable to create model version.'

hkang1:
Model Registry is shaping up nicely!

Some early reviews I can provide:

- There are some components that seem very specific to the model registry page. Would it make more sense to make a `page/ModelRegistry` (or something like that) to house components that really will only exist on that page/section?
- For modals it might be worth taking a look at `useCreateExperimentModal`. This could be a larger effort later on, but the idea there is modularize all the specialized modal logic into the hook so wherever the modal is reused, it's only a few lines of code to introduce it. Things like `NewModelButton` and the modal supporting code would not be needed as you can just create any button and have the click handler call the `showModal` function that the hook provides. If you have questions about this one, happy to walk through this one in detail.
- This is should have been part of the phase 1 review sorry I didn't write this then, so feel free to ignore this. For metadata, thoughts on using antd Table for it? We can get inline editing (looks like you already do this with model versions table), sorting, ability to add drag and drop in the future, overflow interactions, etc.

hkang1 (commented):
LGTM, thanks for the tweaks! A couple of minor things but non-blocking.

hkang1 (approved):

webui/react/src/components/CheckpointModal.tsx
```diff
@@ -96,7 +113,9 @@ const CheckpointModal: React.FC<Props> = (
 
   return (
     <Modal
-      footer={null}
+      footer={checkpoint.uuid ? undefined : null}
```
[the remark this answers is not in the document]
EmilyBonar: Not in this case! `undefined` renders the "default" footer (a primary and secondary button, which are given text and functions in other props), while `null` renders no footer.

PR #3245, opened 2021-11-30
docs: adding job-queue docs

eecsliu:
## Description
Adds new job queue docs and connects them to the existing scheduler docs.

## Checklist

### DIFF
--- a/docs/features/job-queue.txt
+++ b/docs/features/job-queue.txt
@@ -5,12 +5,17 @@
 ##################
 
 The Determined Queue Management system extends the functionality of scheduler options to offer users
-better functionality and control over scheduling decisions. Queue Management is currently in beta
-and is available to the priority scheduler and the Kubernetes priority scheduler. It is available in
-read-only mode, which allows users to observe all submitted jobs and their state, but does not allow
-users to modify the position or any configuration options.
+better functionality in scheduling decisions. The primary role of Queue Management is to provide
+better information about job order, which some schedulers are unable to provide. For example, the
+priority scheduler can provide job ordering information via priority, but the info can be inaccurate
+due to backfilling behavior.
 
-To view the jobs that are currently active in the cluster, use the following command in the CLI:
+Queue Management is currently in beta and is available to the priority scheduler and the Kubernetes
+priority scheduler. It is available in read-only mode, which allows users to observe all submitted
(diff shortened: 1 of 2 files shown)


docs/features/scheduling.txt
```diff
@@ -109,6 +109,9 @@ Here is an example of how the priority scheduler behaves with preemption enabled
    priority 1 distributed training experiment starts running. Once that experiment is complete,
    distributed training experiment with priority 2 restarts.
 
+The priority scheduler can be used with Determined's :ref:`job-queue`, which allows more insight and
+control over scheduling decisions.
```
hamidzr, line 113: do we leave "control" out for now?

docs/features/job-queue.txt
```diff
@@ -0,0 +1,23 @@
...
+To view the jobs that are currently active in the cluster, use the following command in the CLI:
+
+.. code:: bash
+
+   det job list
+
+This will display a table of results, ordered by scheduling order. Note that the scheduling order is
+not necessarily the priority of the job. In addition to the order of the job, users can see the
+state of the jobs and how many slots each job is using. Jobs can be in the following states: queued,
+running, backfilled, or preempted. Jobs that have completed or errored out are not counted as active
```
hamidzr, line 22: > obs can be in the following states: queued, running, backfilled, or preempted.

IMO `preempted` used here could be confusing.

wdy think about mapping these to the states that users would see in the api response or the UI? 

in those places, we report one of 3 queued, scheduled, or scheduled backfilled.

hamidzr (approved):

docs/features/scheduling.txt
hamidzr: same for the k8 version of this

docs/features/job-queue.txt
eecsliu: ahhh good catch. This was just me spilling my thoughts and forgot to go back to edit/verify.

docs/features/scheduling.txt
eecsliu: will do

hamidzr:
two other points worth mentioning I think would be 
- the job queue can be viewed on the webui
- not all schedulers provide job ordering information
- controlling and adjusting the queue would become available in an upcoming release..

PR #3246, opened 2021-11-30
fix: invalidhp handling for single trial experiments [DET-6259]

liamcli:
## Description
If we catch an InvalidHP exception for a single trial experiment, we will continue creating new trials with the same constant experiment config.  This PR fixes that.

## Test Plan
Tested manually.

### DIFF
--- a/master/pkg/searcher/random.go
+++ b/master/pkg/searcher/random.go
@@ -83,15 +83,17 @@ func (s *randomSearch) trialExitedEarly(
 	ctx context, requestID model.RequestID, exitedReason model.ExitedReason,
 ) ([]Operation, error) {
 	s.PendingTrials--
-	if exitedReason == model.InvalidHP || exitedReason == model.InitInvalidHP {
-		var ops []Operation
-		create := NewCreate(ctx.rand, sampleAll(ctx.hparams, ctx.rand), model.TrialWorkloadSequencerType)
-		ops = append(ops, create)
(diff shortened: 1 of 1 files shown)


master/pkg/searcher/random.go
```diff
@@ -83,15 +83,17 @@ func (s *randomSearch) trialExitedEarly(
...
-	if exitedReason == model.InvalidHP || exitedReason == model.InitInvalidHP {
-		var ops []Operation
-		create := NewCreate(ctx.rand, sampleAll(ctx.hparams, ctx.rand), model.TrialWorkloadSequencerType)
-		ops = append(ops, create)
-		ops = append(ops, NewValidateAfter(create.RequestID, s.MaxLength()))
-		ops = append(ops, NewClose(create.RequestID))
-		// We don't increment CreatedTrials here because this trial is replacing the invalid trial.
-		s.PendingTrials++
-		return ops, nil
+	if s.randomSearchState.SearchMethodType == RandomSearch {
```
stoksc, line 86: At this point I'm inclined to just pull single out into its own searcher. That way for features like invalid hp the dev would explicitly think through "now what should this do for single search" and bugs like this would be less likely.

But that's secondary to just fixing the bug.

stoksc (approved):
lgtm, good catch

master/pkg/searcher/random.go
stoksc:
```suggestion
	if s.SearchMethodType == RandomSearch {
```

PR #3247, opened 2021-12-01
fix: hp search progress to account for early-stopped trials [DET-6231]

liamcli:
## Description
We are not counting progress for early exit trials correctly for random and grid search.  The desired behavior is as follows:
- Random: trials with exit reason InvalidHPs should not count towards trial progress at all. All other early exit trials should count as completing the max_length when accounted for progress.
- Grid: all early exit trials including InvalidHP trials should count as completing the max_length when accounted for progress.

This PR changes progress calculations to reflect the above.

### DIFF
--- a/master/pkg/searcher/grid.go
+++ b/master/pkg/searcher/grid.go
@@ -61,13 +61,31 @@ func (s *gridSearch) initialOperations(ctx context) ([]Operation, error) {
 	return ops, nil
 }
 
-func (s *gridSearch) progress(trialProgress map[model.RequestID]model.PartialUnits) float64 {
+func (s *gridSearch) progress(
+	trialProgress map[model.RequestID]model.PartialUnits,
+	trialsClosed map[model.RequestID]bool) float64 {
 	if s.MaxConcurrentTrials() > 0 && s.PendingTrials > s.MaxConcurrentTrials() {
 		panic("pending trials is greater than max_concurrent_trials")
 	}
-	unitsCompleted := sumTrialLengths(trialProgress)
+	unitsCompleted := 0.
+	for k, v := range trialsClosed {
+		if _, ok := trialProgress[k]; !ok {
+			if v {
+				unitsCompleted += float64(s.MaxLength().Units)
+			}
+		}
+	}
+	for k, v := range trialProgress {
+		if closed, ok := trialsClosed[k]; ok {
+			if closed {
+				unitsCompleted += float64(s.MaxLength().Units)
+			}
+		} else {
+			unitsCompleted += float64(v)
+		}
+	}
 	unitsExpected := s.MaxLength().Units * s.trials
-	return float64(unitsCompleted) / float64(unitsExpected)
+	return unitsCompleted / float64(unitsExpected)
 }
 
 // trialExitedEarly does nothing since grid does not take actions based on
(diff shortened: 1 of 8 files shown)


master/pkg/searcher/grid.go
```diff
@@ -61,13 +61,31 @@ func (s *gridSearch) initialOperations(ctx context) ([]Operation, error) {
...
+	}
+	for k, v := range trialProgress {
+		if closed, ok := trialsClosed[k]; ok {
+			if closed {
+				unitsCompleted += float64(s.MaxLength().Units)
+			}
+		} else {
+			unitsCompleted += float64(v)
+		}
+	}
```
stoksc, line 86: i think this an equivalent and much easier to read version of this logic. trialsClosed.where(has no progress) union trialProgress.where(closed) is the same set as just trialsClosed.
```suggestion
	// Count each closed trial completely, no matter how it closed (invalid HP counts).
	for range trialsClosed {
		unitsCompleted += float64(s.MaxLength().Units)
	}
	// Count not closed trials for what they have completed.
	for _, v := range trialProgress {
		unitsCompleted += float64(v)
	}
```

master/pkg/searcher/random.go
```diff
@@ -68,13 +68,24 @@ func (s *randomSearch) initialOperations(ctx context) ([]Operation, error) {
...
-func (s *randomSearch) progress(trialProgress map[model.RequestID]model.PartialUnits) float64 {
+func (s *randomSearch) progress(
+	trialProgress map[model.RequestID]model.PartialUnits,
+	trialsClosed map[model.RequestID]bool) float64 {
 	if s.MaxConcurrentTrials() > 0 && s.PendingTrials > s.MaxConcurrentTrials() {
 		panic("pending trials is greater than max_concurrent_trials")
 	}
-	unitsCompleted := sumTrialLengths(trialProgress)
+	unitsCompleted := 0.
+	for k, v := range trialProgress {
```
stoksc, line 78: this (and grid) is a bit tricky to understand it works properly, since it only because we delete the progress of invalid hp trials (it's not 'locally understandable'). maybe just drop a comment noting this?

master/pkg/searcher/random.go
```diff
@@ -68,13 +68,24 @@ func (s *randomSearch) initialOperations(ctx context) ([]Operation, error) {
...
+		if closed, ok := trialsClosed[k]; ok {
+			if closed {
+				unitsCompleted += float64(s.MaxLength().Units)
+			}
```
stoksc, line 82: existence in this map indicates a trial is closed (it's just a set, but go doesn't have a set), not the bool (this is where it may be better to use `map[...]struct{}{}`. but since we always set it to true, we can use it like below.
```suggestion
		if trialsClosed[k] {
			unitsCompleted += float64(s.MaxLength().Units)
```

master/pkg/searcher/grid.go
stoksc: it might just be the set map value checks. it's 2 extra things to think about.
```suggestion
for k := range trialsClosed {
		if _, ok := trialProgress[k]; !ok {
			unitsCompleted += float64(s.MaxLength().Units)
		}
	}
	for k, v := range trialProgress {
		if trialsClosed[k] {
			unitsCompleted += float64(s.MaxLength().Units)
		} else {
			unitsCompleted += float64(v)
		}
	}
```

stoksc (approved):
lgtm, nice catch once again

PR #3248, opened 2021-12-01
test: store and report webui test results

hamidzr:
## Description

ci: 
- see which tests failed
- enable tracking of test flakiness in circleci at test level

## Checklist


hamidzr:
sample output https://app.circleci.com/pipelines/github/determined-ai/determined/17249/workflows/403b75b3-c89f-4ce3-a1b0-cf4502b9defe/jobs/526871/tests

webui/react/package.json
```diff
@@ -13,7 +13,7 @@
     "lint:js": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
     "lint:css": "stylelint --config .stylelintrc.js \"src/**/*.{css,less,scss}\"",
     "test": "craco test --config jest.config.js",
-    "test:coverage": "CI=true npm test -- --env=jsdom --coverage",
+    "test:coverage": "CI=true JEST_JUNIT_OUTPUT_DIR=coverage npm test -- --env=jsdom --coverage --reporters=default --reporters=jest-junit",
```
hamidzr, line 16: this could be added to the jest-config too I think

.circleci/config.yml
```diff
@@ -1178,6 +1178,8 @@ jobs:
       - checkout
       - react-get-deps
       - run: make -C webui/react test
+      - store_test_results:
+          path: webui/react/coverage
```
hkang1, line 1182: **suggestion:** there is an HTML report that is generated at `webui/react/coverage/lcov-report/` (including entry point of `index.html`). wdy think about adding:
```
- store_artifacts:
    path: webui/react/coverage/lcov-report
```

hkang1 (approved):
LGTM! Thanks for running the coverage for CI! Added one suggestion

webui/react/package.json
hkang1: if that's easy to do, I think that's a good idea to do

.circleci/config.yml
hamidzr: https://circleci.com/docs/2.0/code-coverage/#javascript from circleci suggests the same as well

webui/react/package.json
hkang1: [related thread](https://stackoverflow.com/questions/59372493/override-jest-junit-default-output-location)

hamidzr: https://github.com/determined-ai/determined/pull/3248/commits/70176e0780d5316b143a3c11e10ba469e5ae9840 this generates the correct config but seems to be overridden somewhere.

hkang1: **question:** idk much about the reporters. For these flags `--reporters=default --reporters=jest-junit` are they ok to have as part of `npm run test` since that also uses `jest.config.js`

hkang1: > 70176e0 this generates the correct config but seems to be overridden somewhere.

hmmm... there is a `jest` config in `package.json`:
```
  "jest": {
    "coveragePathIgnorePatterns": [
      "/src/services/api-ts-sdk/"
    ]
  }
```
could that be the culprit? should we move the package.json `jest` config into `jest.config.js` also?

hamidzr:
```
These options in your package.json Jest configuration are not currently supported by Create React App:

  • reporters

If you wish to override other Jest options, you need to eject from the default setup. You can do so by running npm run eject but remember that this is a one-way operation. You may also file an issue with Create React App to discuss supporting more options out of the box.
```

hamidzr: reporters https://jestjs.io/docs/cli#--reporters instruct which reporters to run. we only want it when running in the ci since it's not more readable than the cli output in local usecase IMO

hamidzr: > **question:** idk much about the reporters. For these flags `--reporters=default --reporters=jest-junit` are they ok to have as part of `npm run test` since that also uses `jest.config.js`

we'd only really want those options if we're looking at the junit report which AFAIK we'd only do through CI. adding it to `test` would just add additional work that `npm run test` would need to do if we run it locally

hamidzr: I simplified the package.json scripts https://github.com/determined-ai/determined/pull/3248/commits/f4a6bf35284db1a788e45330358d691aca44cb77

hamidzr:
@hkang1  https://534564-253846879-gh.circle-artifacts.com/0/webui/react/coverage/lcov-report/src/index.html shouldn't this report this tested file `src/utils/string.ts` ?

hkang1:
Love these changes, thanks for making this happen!

- [ ] lcov-report artifacts are available under CI `test-unit-react`
- [ ] `test-unit-react` reports the unit / react test results
- [ ] local `npm run test` works
- [ ] local `npm run test:coverage` works

hkang1 (approved):
Love these changes, thanks for making this happen!

- [x] lcov-report artifacts are available under CI `test-unit-react`
- [x] `test-unit-react` reports the unit / react test results
- [x] local `npm run test` works
- [x] local `npm run test:coverage` works

PR #3249, opened 2021-12-01
fix: generate SSH keys for restored trials

dzhu:
## Description

Previously, the SSH RSA key size for trials was set on a code path that
was only run for newly created experiments. A trial restored after a
master restart would therefore attempt to generate keys with a size of 0
and crash. Instead, we now set the key size in the master's base task
spec so that it applies to all tasks automatically.

## Test Plan

- [x] check that restoring a trial after a master restart fails without the fix and succeeds with it
- [x] check that shells and trials generate a key of the configured size

## Commentary

The error in key generation and, more generally, the reason an allocation was killed don't seem to be surfaced anywhere at all, as far as I can tell.


shiyuann (approved):
I encountered this error and found your PR. Nice work!

PR #3252, opened 2021-12-02
fix: det job get returns inaccurate info

eecsliu:
## Description
Fixes two bugs:
- bug where the reported priority was always the default
- bug where master panics when calling `det job list`

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/kubernetes_resource_manager.go
+++ b/master/internal/resourcemanagers/kubernetes_resource_manager.go
@@ -205,8 +205,14 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error 
 	case sproto.SetGroupMaxSlots:
 		k.getOrCreateGroup(ctx, msg.Handler).maxSlots = msg.MaxSlots
 
-	case sproto.SetGroupWeight, sproto.SetGroupPriority:
+	case sproto.SetGroupWeight:
 		// SetGroupWeight and SetGroupPriority are not supported by the Kubernetes RP.
+
+	case sproto.SetGroupPriority:
+		group := k.getOrCreateGroup(ctx, msg.Handler)
+		if msg.Priority != nil {
+			group.priority = msg.Priority
+		}
 
 	case sproto.SetTaskName:
 		k.receiveSetTaskName(ctx, msg)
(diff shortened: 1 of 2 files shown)


master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -205,9 +205,15 @@ func (k *kubernetesResourceManager) receiveRequestMsg(ctx *actor.Context) error
 	case sproto.SetGroupMaxSlots:
 		k.getOrCreateGroup(ctx, msg.Handler).maxSlots = msg.MaxSlots
 
-	case sproto.SetGroupWeight, sproto.SetGroupPriority:
+	case sproto.SetGroupWeight:
 		// SetGroupWeight and SetGroupPriority are not supported by the Kubernetes RP.
 
+	case sproto.SetGroupPriority:
```
hamidzr, line 211: isn't this a write side feature?

master/internal/resourcemanagers/priority.go
```diff
@@ -383,3 +383,39 @@ func taskFilter(label string, zeroSlots bool) func(*sproto.AllocateRequest) bool
...
+	var reqs []*sproto.AllocateRequest
+
+	for it := taskList.iterator(); it.next(); {
+		reqs = append(reqs, it.value())
+	}
+
+	sort.Slice(reqs, func(i, j int) bool {
+		p1 := *groups[reqs[i].Group].priority
+		p2 := *groups[reqs[j].Group].priority
+		if k8s { // in k8s, higher priority == more prioritized
```
hamidzr, line 401: nit I think you can flip p1 and p2 value like we do in api_job for a bit shorter syntax

master/internal/resourcemanagers/priority.go
```diff
@@ -383,3 +383,39 @@ func taskFilter(label string, zeroSlots bool) func(*sproto.AllocateRequest) bool
 		return request.Label == label && (request.SlotsNeeded == 0) == zeroSlots
 	}
 }
+
+func sortTasks(
```
hamidzr, line 387: nit: will this be used in the agent-based priority scheduler?

eecsliu: not at the moment - the agent one sets the job status, whereas in k8s we must observe the pod status

master/internal/resourcemanagers/kubernetes_resource_manager.go
eecsliu: this is necessary for determined to adjust the priority from the default to what the user sets in the config.

hamidzr (approved):

PR #3257, opened 2021-12-02
fix: update bumpversion config to account for new React config file

dzhu:
## Test Plan

- [x] run bumpversion locally


dzhu:
Quick-merging since this only affects release stuff and is currently blocking something I need to do.

PR #3258, opened 2021-12-02
feat: add prometheus endpoint for internal Determined state mappings [DET-5890]

azhou-determined:
## Description

adds an endpoint at /debug/prom/det-metrics that exposes entity mappings needed for DCGM exporter and cAdvisor.

## Test Plan

- enable prometheus config in master.yaml
- create prometheus.yml 
- deploy and hit /debug/prom/det-metrics and verify mappings

## Commentary (optional)

Open questions/concerns:
- do we ship prometheus.yml with this? i didn't include my mapping file because it doesn't seem like it belongs in the default master package that is bundled. however, it's not straightforward to rewrite some of the configs/labels that are necessary for prometheus to appropriately scrape
- only the endpoint checks if the master config has prometheus enabled. all other code that associates mappings will run no matter what. adding an if wrapper around all of these sprinkled associations is very messy, but not doing so is additional overhead. 

## Checklist


master/internal/config/config.go
```diff
@@ -88,7 +88,7 @@ type Config struct {
 	ClusterName           string                            `json:"cluster_name"`
 	Logging               model.LoggingConfig               `json:"logging"`
 	HPImportance          hpimportance.HPImportanceConfig   `json:"hyperparameter_importance"`
-
+	EnablePrometheus      bool                              `json:"enable_prometheus"`
```
stoksc, line 91: can we put this under an `observability` subsection or something (i imagine in due time we could have more related configs and I dont want to pollute the root namespace)?

master/internal/core.go
```diff
@@ -825,10 +827,12 @@ func (m *Master) Run(ctx context.Context) error {
 	m.echo.Any("/debug/pprof/symbol", echo.WrapHandler(http.HandlerFunc(pprof.Symbol)))
 	m.echo.Any("/debug/pprof/trace", echo.WrapHandler(http.HandlerFunc(pprof.Trace)))
 
-	if m.config.InternalConfig.PrometheusEnabled {
+	if m.config.EnablePrometheus {
 		p := prometheus.NewPrometheus("echo", nil)
 		p.Use(m.echo)
 		m.echo.Any("/debug/prom/metrics", echo.WrapHandler(promhttp.Handler()))
+		m.echo.Any("/debug/prom/det-metrics",
```
stoksc, line 834: personally
```suggestion
		m.echo.Any("/prom/det-state-metrics",
```

master/internal/prom/prom.go
```diff
@@ -0,0 +1,233 @@
...
+func init() { //nolint: gochecknoinits
+	DetStateMetrics.MustRegister(containerIDToAllocationID)
+	DetStateMetrics.MustRegister(containerIDToRuntimeID)
+	DetStateMetrics.MustRegister(gpuUUIDToContainerID)
+	DetStateMetrics.MustRegister(experimentIDToLabels)
+	DetStateMetrics.MustRegister(allocationIDToTaskID)
+}
+
+// AssociateAllocationContainer associates an allocation with its container ID.
+func AssociateAllocationContainer(aID string, cID string) {
```
stoksc, line 81: since it's so easy to mix these up when calling, i think it's better to drop the `string(x)`/`.String()` calls and let these take the more descriptive types, and then it can do the conversion itself.

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -310,6 +310,7 @@ func (a *agent) handleIncomingWSMessage(ctx *actor.Context, msg aproto.MasterMes
 	switch {
 	case msg.AgentStarted != nil:
 		telemetry.ReportAgentConnected(ctx.Self().System(), a.uuid, msg.AgentStarted.Devices)
+		prom.AddAgentAsTarget(ctx, a.uuid.String(), a.address, a.resourcePoolName)
```
stoksc, line 313: might not matter (prom might not do anything) but i think this should be in `if !a.agentStartedInit {`

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -358,6 +359,7 @@ func (a *agent) containerStateChanged(ctx *actor.Context, sc aproto.ContainerSta
 		rsc.ContainerStarted = &sproto.TaskContainerStarted{
 			Addresses: sc.ContainerStarted.Addresses(),
 		}
+		prom.AssociateContainerRuntimeID(sc.Container.ID.String(), sc.ContainerStarted.ContainerInfo.ID)
```
stoksc, line 362: i think this would be better in `allocation.go` because if it's here this association won't work for k8s. i think maybe `sproto.TaskContainerStarted` should have a field like `NativeReservationID` that would be a container id for agents and a pod id for k8s (and so on), then the alloc can do this for each reservation (where reservation is just our abstract name or process/pod/container/etc one clump of resources)

master/internal/task/allocation.go
```diff
@@ -165,6 +167,12 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 		if err := a.ResourcesAllocated(ctx, msg); err != nil {
 			a.Error(ctx, err)
 		}
+
+		for cID := range a.reservations {
+			prom.AssociateAllocationTask(a.req.AllocationID.String(), a.req.TaskID.String())
+			prom.AddAllocationContainer(a.reservations[cID].Summary())
```
stoksc, line 173: so re comment above, this would become `AddAllocationReservation` and know how to handle the stuff happening in agents.go too.

master/internal/prom/prom.go
```diff
@@ -0,0 +1,233 @@
...
+			detAgentIDLabel:      agentID,
+			detResourcePoolLabel: agentResourcePool,
+		},
+	}
+
+	fileSDConfigs := getFileSDConfigs()
+
+	fileSDConfigs = append(fileSDConfigs, fileSDConfig)
+
+	err := writeConfigsToTargetsFile(fileSDConfigs)
```
stoksc, line 177: This is interesting and works for now; maybe though it is better to expose an api like `/prom/det-file-sd-config` that returns the file sd config. I think that would be more flexible for users to consume.

master/internal/prom/prom.go
```diff
@@ -0,0 +1,233 @@
...
+
+	if err != nil {
+		ctx.Log().Errorf("Error updating targets file %s", err)
+	}
+}
+
+func getFileSDConfigs() []fileSDConfigEntry {
+	var fileSDConfigs []fileSDConfigEntry
+
+	fileContent, _ := ioutil.ReadFile(targetsFile)
```
stoksc, line 212: i think `fs.ReadFile` would be preferred now from `io/fs`

master/internal/prom/prom.go
stoksc: either way though, I think this file would need a mutex?

stoksc (commented):
looks pretty good, mostly my comments are:
* file sd config should be an api
* we should swap out container for reservation everywhere and support k8s since it would be really easy then

philipnrmn:
Are we expecting these metrics to be collected and stored by Prometheus? If so I have concerns about labels with unbounded dimensions like `container_id` and `allocation_id` - they will introduce a significant increase in cardinality, which will decrease the time window which can be stored in a TSDB and affect its query performance.

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -274,6 +272,7 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
 		ctx.Self().Stop()
 	case actor.PostStop:
 		ctx.Log().Infof("agent disconnected")
+		//prom.RemoveAgentAsTarget(ctx, a.uuid.String())
```
ioga, line 275: remove commented calls?

master/internal/sproto/task_actor.go
```diff
@@ -31,7 +31,8 @@ type (
 	}
 	// TaskContainerStarted contains the information needed by tasks from container started.
 	TaskContainerStarted struct {
-		Addresses []cproto.Address
+		Addresses           []cproto.Address
+		NativeReservationID string
```
ioga, line 35: could you please add a comment what this value is? I assume on agent it's a docker container id (a long hex string). is it a magic k8s-generated pod name on k8s? or also just a hex string?

ioga (approved):
got a minor comment about the code, but I am not super qualified on anything prometheus. looks like @philipnrmn had a comment, not sure if it was addressed.

philipnrmn:
> Are we expecting these metrics to be collected and stored by Prometheus? If so I have concerns about labels with unbounded dimensions like container_id and allocation_id - they will introduce a significant increase in cardinality, which will decrease the time window which can be stored in a TSDB and affect its query performance.

@stoksc and I caught up about this earlier. Basically my concern is that this could cause runaway cardinality. It sounds like we're not going to get anywhere near Prometheus' limits, and even in the worst case we could imagine we would only be looking at a few tens of thousands of new series per day, which shouldn't be a problem. 

I'd recommend we delete these series as soon as they become inactive to prevent a backup on systems with a long retention window, but I withdraw the basic objection.

azhou-determined:
> Are we expecting these metrics to be collected and stored by Prometheus? If so I have concerns about labels with unbounded dimensions like `container_id` and `allocation_id` - they will introduce a significant increase in cardinality, which will decrease the time window which can be stored in a TSDB and affect its query performance.

Good concern. These metrics will be scraped by Prometheus. And while it's true that generally unbounded labels can be dangerous, I think these are fine because they will not be active for all query time series.

master/internal/resourcemanagers/agent/agent.go
azhou-determined: this comment is no longer relevant since moving to HTTP SD configs for the targets

master/internal/config/config.go
azhou-determined: done

master/internal/core_observability.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func (m *Master) getPrometheusTargets(c echo.Context) (interface{}, error) {
+	var resp actor.Response
+
+	switch {
+	case sproto.UseAgentRM(m.system):
+		resp = m.system.AskAt(sproto.AgentsAddr, &apiv1.GetAgentsRequest{})
+	case sproto.UseK8sRM(m.system):
+		resp = m.system.AskAt(sproto.PodsAddr, &apiv1.GetAgentsRequest{})
```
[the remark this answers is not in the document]
ioga: we are planning to add another RM (Slurm) in 2022, and so I'd like us to move in a direction where the current RM is a box with a fixed and more or less defined interface for all the code outside of `resourcemanagers`.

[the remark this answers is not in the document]
azhou-determined: made the change.

PR #3266, opened 2021-12-03
feat: jq webui persist selected resource pool

hamidzr:
## Description

Remember selected RP 
Exp details to link to the job queue 
Update default job order
Show exp ids in the table 

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3267, opened 2021-12-03
fix: job cli returns different info from webui

eecsliu:
## Description
Changes the output of `det job list` to match what the webui displays.

## Checklist

### DIFF
--- a/harness/determined/cli/job.py
+++ b/harness/determined/cli/job.py
@@ -1,5 +1,6 @@
 import json
 from argparse import Namespace
+from datetime import datetime
 from typing import Any, List
 
 import yaml
@@ -18,7 +19,7 @@ def ls(args: Namespace) -> None:
         resourcePool=args.resource_pool,
         pagination_limit=args.limit,
         pagination_offset=args.offset,
-        orderBy="ORDER_BY_DESC" if not args.reverse else "ORDER_BY_ASC",
+        orderBy="ORDER_BY_ASC" if not args.reverse else "ORDER_BY_DESC",
     )
     if args.yaml:
         print(yaml.safe_dump(response.to_json(), default_flow_style=False))
@@ -26,16 +27,15 @@ def ls(args: Namespace) -> None:
         print(json.dumps(response.to_json(), indent=4, default=str))
     elif args.table or args.csv:
         headers = [
-            "Jobs Ahead",
+            "#",
             "ID",
-            "Entity ID",
-            "Status",
             "Type",
-            "Slots Acquired",
-            "Slots Requested",
-            "Name",
+            "Job Name",
+            "Priority",
+            "Submitted",
+            "Slots",
+            "Status",
             "User",
-            "Submission Time",
         ]
         values = [
             [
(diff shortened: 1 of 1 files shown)


harness/determined/cli/job.py
```diff
@@ -18,39 +19,37 @@ def ls(args: Namespace) -> None:
...
+            "#",
             "ID",
-            "Entity ID",
-            "Status",
             "Type",
-            "Slots Acquired",
-            "Slots Requested",
-            "Name",
+            "Job Name",
+            "Priority",
```
hamidzr, line 34: blocking: this should cover fairshare with weights as well

hamidzr (changes requested):
other than the issue I tagged it looks good

hamidzr:
python 3.6 doesn't seem to have `fromisoformat` I think we do still advertise py3.6 support

https://docs.python.org/3.6/library/datetime.html
AttributeError: type object 'datetime.datetime' has no attribute 'fromisoformat'

hamidzr (commented):
lgtm minus the python 3.6 support comment I added above.

image

nit: the original ask was to make it look like webui columns but I wonder whether the slots column is better of now compared to when it was reported as two columns. in the webui we get away with this since we can add on hover tool tips to explain what the numbers are. cc @shiyuann

harness/determined/cli/job.py
```diff
@@ -1,56 +1,64 @@
...
 from determined.common.api.b import get_GetJobs
 from determined.common.declarative_argparse import Arg, Cmd, Group
 
 
 @authentication.required
 def ls(args: Namespace) -> None:
+    priority = True
+    config = api.get(args.master, "config").json()
+    try:
+        if config["resource_manager"]["scheduler"]["type"] == "fair_share":
```
hamidzr, line 21: what if the resource pool config overrides it?

eecsliu: ah good point. Will fix

eecsliu:
For slots, what if the table read `Slots (acquired/needed)` ?

harness/determined/cli/job.py
```diff
@@ -1,56 +1,75 @@
...
 from determined.cli.session import setup_session
+from determined.common import api
 from determined.common.api import authentication
 from determined.common.api.b import get_GetJobs
 from determined.common.declarative_argparse import Arg, Cmd, Group
 
 
 @authentication.required
 def ls(args: Namespace) -> None:
+    priority = True
```
hamidzr, line 19: nit: would be nice to be able to tell the var type from the name.

harness/determined/cli/job.py
```diff
@@ -1,56 +1,75 @@
...
+    config = api.get(args.master, "config").json()
+    try:
```
hamidzr, line 21: suggestion: it seems safer and perhaps easier to me to go through the resource pool endpoint (the cli wouldn't need to assume this nested structure schema for the config.)

hamidzr (approved):
looks good. did you want to update the slots header? 
image (3)


following the feedback we got I updated the id column from the webui and tacked on the experiment id onto the name col. https://github.com/determined-ai/determined/pull/3266

I'm not sure if we want this for the cli but we can do that separately if we do.

PR #3269, opened 2021-12-05
refactor: refactor trial model and trial actor

shiyuann:
## Description

This changes a few things:

1. Replaced a UUID with a `trial-<trial-id>` as the segment for the trial actor.
2. A trial actor has to be associated with a trial entry in the database. We used to create the trial entry in the database after it is assigned. The creation of the trial entry is moved to the experiment logic.
3. Removed `SetTaskName` message in the resource manager communication protocol since it is used for changing the trial task name to use the trial ID.
4. Added an actor path field to all the log entries from the actors and remove the fields that indicate the actor path.
5. Added some logging code to help debug.

This helps move to the ideal world that
1. doesn't use UUIDs in the system.
2. has simply control logic that depends on well-designed DB-persistent state machines.

## Test Plan

1. Test with trial creation and restoring on (paused trials, active trials) x (0-progress trials, has-some-progress trials).
2. Pass the e2e tests.

## Checklist


stoksc:
@shiyuann what is the deal here?

stoksc:
@shiyuann i plan to move in this direction anyway, i'll keep this in mind when i change related code. if you want to rebase i will try to review, sorry.

PR #3270, opened 2021-12-06
feat: Add python method for listing trials in an experiment [DET-5088]

mpkouznetsov:
## Description
Add method to ExperimentReference to list trials within an experiment [DET-5088].

## Test Plan

Tested manually, locally.

## Commentary (optional)

There is an endpoint in the REST API to get trials within an experiment. It has a rich set of parameters: sorting, pagination, filtering by state. I decided to only expose the sorting parameters. The only similar method, `determined.experimental.client.get_models`, also does not offer pagination, which makes sense if you view serving UIs as one of responsibility of a RESTful API of which the python API is free.
I don't have as strong an argument for not offering filtering by state; it just does not feel very important.

## Checklist

[DET-5088]: https://determinedai.atlassian.net/browse/DET-5088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/harness/determined/common/experimental/experiment.py
+++ b/harness/determined/common/experimental/experiment.py
@@ -113,7 +113,7 @@ class ExperimentReference:
                 :class:`~determined.experimental.TrialOrderBy`.
         """
         r = self._session.get(f"/api/v1/experiments/{self.id}/trials")
-        trials = r.json().get("trials")
+        trials = r.json()["trials"]
         return [trial.TrialReference(t["id"], self._session) for t in trials]
 
     def kill(self) -> None:
--- a/harness/determined/common/experimental/experiment.py
+++ b/harness/determined/common/experimental/experiment.py
@@ -3,7 +3,7 @@ import sys
 import time
 from typing import Any, Dict, List, Optional, cast
 
-from determined.common.experimental import checkpoint, session
+from determined.common.experimental import checkpoint, session, trial
 
 
 class ExperimentState(enum.Enum):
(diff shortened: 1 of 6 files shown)


harness/determined/common/experimental/experiment.py
```diff
@@ -98,6 +98,24 @@ def delete(self) -> None:
...
+        Get the list of :class:`~determined.experimental.TrialReference` instances
+        representing trials for an experiment.
+
+        Arguments:
+            sort_by: Which field to sort by. See :class:`~determined.experimental.TrialSortBy`.
+            order_by: Whether to sort in ascending or descending order. See
+                :class:`~determined.experimental.TrialOrderBy`.
+        """
+        r = self._session.get(f"/api/v1/experiments/{self.id}/trials")
+        trials = r.json().get("trials")
```
ioga, line 116: maybe hard fail here if `trials` field is missing (probably meaning something is very wrong)?

```suggestion
        trials = r.json()["trials"]
```

ioga (approved):

harness/determined/common/experimental/experiment.py
mpkouznetsov: Good thing you asked; I forgot to mention it in my comment. I was following what I though was a pattern established in
https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L156
https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L165
https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L196
https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L205
(In all of the above examples, `None` is not a viable result; another exception is imminent if `get` returns `None`).
However, upon closer look, we also have it the other way in a few instances:
https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/experiment.py#L205
and 6 or 7 more.

Should we homogenize the code? (There is, of course, a risk of regressions).

ioga: I suspect that in 90% of these cases if `.get(` doesn't have `default=` parameter than it's probably "useless". Master normally returns these fields so we never run into a situation where it'd blow up.

For example in https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L156
it won't fail immediately, but then it'll fail in `from_json` on this lookup: https://github.com/determined-ai/determined/blob/9e91c01bdb8f591f4978654f1d7b1b15677156d3/harness/determined/common/experimental/model.py#L351

In https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/determined.py#L205 it'd make more sense to do `.get("labels", [])`, otherwise it's also not looking right.

PR #3271, opened 2021-12-06
chore: better telemetry 

shiyuann:
## Description

This PR does the following things:
1. normalize the harness telemetry events to avoid creating separate tables for different trials.
2. fix the experiment creation event which was previously called in creating experiment actor as opposed to creating an experiment database model. The old experiment creation event can be fixed by deduplicating the tuple of the cluster ID and the experiment ID.
3. remove the unused event of agent disconnected and agent connected.
4. remove the unused event of resource pool created.
5. add an allocation terminal event to get the slot time of other task types.
6. add more properties to the event of master tick, including master version, gpu number, etc.
7. fix a telemetry bug that causes telemetry not working in 0.17.4 and 0.17.5.
8. add the CI to use the development key for telemetry.

## Test Plan

1. Test the telemetry on dev account.
   - submit an experiment that successfully complete
   - submit an experiment that fails.
   - submit a command/notebook/shell/tensorboard.
2. CI

## Checklist

### DIFF
--- a/harness/determined/common/api/__init__.py
+++ b/harness/determined/common/api/__init__.py
@@ -1,4 +1,4 @@
-from determined.common.api import authentication, errors, metric, request
+from determined.common.api import analytics, authentication, errors, metric, request
 from determined.common.api.authentication import Authentication, Session, salt_and_hash
 from determined.common.api.experiment import (
     activate_experiment,
@@ -26,9 +26,6 @@ from determined.common.api.request import (
     post,
     put,
     ws,
-)
-from determined.common.api.analytics import (
-    send_analytics,
 )
 from determined.common.api.profiler import (
     post_trial_profiler_metrics_batches,
--- a/harness/determined/common/api/__init__.py
+++ b/harness/determined/common/api/__init__.py
@@ -1,4 +1,4 @@
-from determined.common.api import analytics, authentication, errors, metric, request
+from determined.common.api import authentication, errors, metric, request
 from determined.common.api.authentication import Authentication, Session, salt_and_hash
 from determined.common.api.experiment import (
     activate_experiment,
--- a/harness/determined/common/api/analytics.py
+++ b/harness/determined/common/api/analytics.py
@@ -1,21 +1,46 @@
+import logging
 import os
+import sys
+from typing import Any, Dict, Type
 
 import analytics
 
+enabled = os.environ.get("DET_SEGMENT_ENABLED") == "true"
 analytics.write_key = os.environ.get("DET_SEGMENT_API_KEY")
 _cluster_id = os.environ.get("DET_CLUSTER_ID")
 
-if (
-    os.environ.get("DET_SEGMENT_ENABLED") == "true"
-    and _cluster_id is not None
-    and analytics.write_key is not None
-):
 
-    def send_analytics(tracking_key: str) -> None:
-        analytics.track(_cluster_id, tracking_key)
+def get_library_version_analytics() -> Dict[str, Any]:
+    modules = [
+        "determined",
+        "model_hub",
+        "torch",
+        "pytorch_lightning",
+        "tensorflow",
+        "transformers",
+        "mmcv",
+        "mmdet",
+    ]
+    versions = {}
+    for m in sys.modules:
+        if m in modules:
+            try:
+                versions[m] = sys.modules[m].__version__  # type: ignore
+            except Exception:
+                pass
+    return versions
 
 
-else:
+def get_trial_analytics(obj: Type) -> Dict[str, Any]:
+    ancestors = ",".join(each.__name__ for each in obj.mro())
+    return {
+        "library_version": get_library_version_analytics(),
+        "trial_name": obj.__name__,
+        "trial_ancestors": ancestors,
+    }
 
-    def send_analytics(tracking_key: str) -> None:
-        pass
+
+def send_analytics(event: str, properties: Dict) -> None:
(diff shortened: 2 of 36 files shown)


harness/determined/common/api/__init__.py
```diff
@@ -1,4 +1,4 @@
-from determined.common.api import authentication, errors, metric, request
+from determined.common.api import analytics, authentication, errors, metric, request
```
rb-determined-ai, line 1: I think we should not be importing this in the top-layer namespace at all.  This would basically cause it to show up in the user's IDE when they try to tab complete after `determined.<TAB>`.

Removing it here will probably not affect any of your other code since you're following the google import guide anyway.

harness/determined/common/api/analytics.py
```diff
@@ -1,21 +1,46 @@
+import logging
 import os
+import sys
+from typing import Any, Dict, Type
 
 import analytics
 
+enabled = os.environ.get("DET_SEGMENT_ENABLED") == "true"
```
rb-determined-ai, line 8: Can I request that we start using environment variables to represent bools differently?

With the vast majority of projects I use, they say something like, 'set `ENABLE_XYZ` to any non-empy string to enable XYZ".  We instead inspect the value of the environment variable, even when the variable is just a flag value.

Can we switch to treating new environment variables as true when they are any non-empty string, when they are bools like this?

harness/determined/exec/harness.py
```diff
@@ -82,6 +82,11 @@ def main(chief_ip: Optional[str]) -> int:
         # without horovod, and we can't load the right horovod until we know which Trial class the
         # user implemented.
         trial_class, controller_class = load.get_trial_and_controller_class(env.experiment_config)
+        if info.container_rank == 0:
+            try:
+                analytics.send_analytics("trial_loaded", analytics.get_trial_analytics(trial_class))
+            except Exception as e:
+                logging.debug(f"Cannot sending analytics: {e}")
```
rb-determined-ai, line 89: Cannot send~ing~ analytics

But also, a suggestion: let's just drop this exception.  I think any user might be slightly alarmed to see it, and I doubt any user would be particularly inclined to fix it.  It's like a message that only we care about, but which only the user would see.

In the worst case, if something went really wrong with segment then all of our users might get extra messages that they can't really do anything about.

harness/determined/exec/harness.py
```diff
@@ -82,6 +82,11 @@ def main(chief_ip: Optional[str]) -> int:
         # without horovod, and we can't load the right horovod until we know which Trial class the
         # user implemented.
         trial_class, controller_class = load.get_trial_and_controller_class(env.experiment_config)
+        if info.container_rank == 0:
+            try:
+                analytics.send_analytics("trial_loaded", analytics.get_trial_analytics(trial_class))
```
rb-determined-ai, line 87: The trial class name is unlikely to be useful, since that will be a user-provided name.

It might be more useful to add a `._framework_name` or something to `PyTorchTrial` and friends and use the `trial_class._framework_name` instead, perhaps.

harness/determined/common/api/__init__.py
shiyuann: will delete it

harness/determined/common/api/analytics.py
shiyuann: Yes. And I think for such a switch we should first write it down in our coding style guideline and then use a separate PR to make the system consistent in using that style.

harness/determined/exec/harness.py
shiyuann: >In the worst case, if something went really wrong with segment then all of our users might get extra messages that they can't really do anything about.

right. so are other debug-level messages. It doesn't hurt users but they might tell us if anything goes wrong so that we know we can fix it in the next release.

harness/determined/exec/harness.py
shiyuann: An example data point is:

```
analytics.track('cfe8cddd-a184-4727-b937-27835fb6e33d', 'trial_loaded', {
    'library_version': {
        'determined': '0.17.6-dev0',
        'pytorch_lightning': '1.3.5',
        'torch': '1.9.0+cpu'
    },
    'trial_ancestors': 'GANTrial,LightningAdapter,PyTorchTrial,Trial,object',
    'trial_name': 'GANTrial'
})
```

> It might be more useful to add a ._framework_name or something to PyTorchTrial and friends and use the trial_class._framework_name instead, perhaps.

The framework is already tracked in the `trial_ancestors`. I think this is the cleanest way without bothering using `_framework_name`.

`trial_name` helps us to estimate (not exactly track) how examples are used directly.

stoksc (commented):
few comments for now, will review again in a bit once it has sunk in

master/internal/telemetry/reports.go
```diff
@@ -1,48 +1,132 @@
...
-func ReportAgentDisconnected(system *actor.System, uuid uuid.UUID) {
-	report(system, "agent_disconnected", map[string]interface{}{
-		"uuid": uuid,
-	})
-}
+// ReportAllocationTerminal reports that an allocation ends.
+func ReportAllocationTerminal(
+	system *actor.System, db db.DB, a model.Allocation, d *device.Device,
+) {
+	res, err := db.CompleteAllocationTelemetry(a.AllocationID)
```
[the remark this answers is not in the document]
stoksc: this is resolved, but i really think it would be much nicer if we could save the db call here, but non-blocking.

PR #3282, opened 2021-12-08
feat: experimental ROCm support. [DET-6285]

ioga:
## Description

Add support for ROCm.
- Harness side: rocm key under environment images/variables for ROCm-specific execution, and a ton of related schema updates. If/when we'd be adding more keys here, some refactoring would be useful to reduce the amount of boilerplate.
- Agent side: `rocm-smi` device discovery in agents.  Handling card access and isolation with docker is a bit more involved that it is for nvidia driver. 

Current setup is pytorch and tensorflow working in single-instance multi-gpu configuration. Waiting for hardware access from AMD to test multi-instance dtrain.

Framework versions: ROCm 4.2; Pytorch 1.9; TF 2.5.
Horovod: 0.23.0 + AMD's outstanding PR + our patches. Will work on upstreaming these.

Known limitations: 
- TF 2.6.2 didn't work. 
- No k8s support yet, due to a lack of appropriate hw access.
- No benchmarking (pending next gen hardware)
- No profiling (lower priority, pending benchmarking)

Bonus changes:
- Rename `gpu: ` key in experiment and master configs for environment.image and environment.environment_variables to `cuda: `.
- Rename `gpu` slot type to `cuda`. This affects agent's slot_type parameter and k8s slot_type in the chart.

All of these should be backwards-compatible.

## Test Plan

1. HW access to allow for ROCm-side testing is a challenge at the moment.
2. gpu/cuda key migration: ensure the old configs (using `gpu:` key) still work, ensure `cuda:` key works.

## Commentary (optional)
@rb-determined-ai  please see harness & schema changes
@stoksc please see agent & master changes

ROCm image is currently set to 0.17.6, while the other env images are still at 0.17.4. I didn't bump the others to keep this PR clean. I'll do the other bumpenvs in a separate PR.

## Checklist


docs/sysadmin-basics/cluster-config.txt
```diff
@@ -831,8 +832,9 @@ The master supports the following configuration settings:
    -  ``auto``: Automatically detects the slot type. The agent will detect if there are GPUs. If
       there are GPUs, it maps each GPU to one slot. Otherwise, it maps all the CPUs to a slot.
    -  ``none``: The agent will not create any slots for detected devices.
-   -  ``gpu``: The agent will map each detected GPU to a slot.
+   -  ``gpu``: The agent will map each detected NVIDIA GPU to a slot.
```
rb-determined-ai, line 835: it's _almost_ worth renaming `gpu` to be `nvidia`, but that sounds like an obnoxious migration...

master/internal/resourcemanagers/kubernetes/pod.go
```diff
@@ -428,8 +428,6 @@ func (p *pod) preparePodUpdateMessage(msgText string) string {
 			switch p.slotType {
 			case device.CPU:
 				resourceName = "CPU slots"
-			case device.GPU:
```
rb-determined-ai, line 431: rename `device.GPU` to be `device.NVIDIA`?

rb-determined-ai (commented):
I think perhaps we should automatically migrate `cpu`/`gpu` inputs to be `cpu`/`cuda`/`rocm` inputs, rather than just allow a new `rocm` field.  I feel like it would actually affect relatively few users, since I kinda doubt many users use the `cpu`/`gpu` distinction at all, but I can imagine in a year that we would be glad that we put the work in now, rather than have a schema that implicitly favors one brand of accelerators.

Otherwise the schema changes look pretty good.

ioga:
I've decided to bite the bullet and migrate `gpu` -> `cuda`, which also adds a chunk of backwards-compatibility in environment config for schema-based and for "legacy" master config parsers. 
This rename ended up bleeding into webui, so I've updated it as well to the new naming, and added ROCm slot display where appropriate.

Example cluster page for mixed CUDA/ROCM resource pool setup:
Screen Shot 2021-12-16 at 09 49 43

webui/react/src/types.ts
```diff
@@ -62,12 +62,13 @@ export interface Telemetry {
 
 export enum ResourceType {
   CPU = 'CPU',
-  GPU = 'GPU',
+  CUDA = 'CUDA',
+  ROCM = 'ROCM',
```
hkang1, line 66: **nit:** the value is used for UI display. `ROCm` instead of `ROCM`?
```
ROCM = 'ROCm',
```

ioga:
okay, I think this is in a stable state once again with the gpu->cuda migration and webui MVP added. PTAL.

agent/internal/containers.go
```diff
@@ -222,6 +242,45 @@ func gpuDeviceRequests(cont cproto.Container) []dcontainer.DeviceRequest {
...
+	mappedDevices := []string{"/dev/kfd"}
+
+	for _, uuid := range uuids {
+		rocmDevice := getRocmDeviceByUUID(uuid)
+		devPaths := []string{
+			fmt.Sprintf("/dev/dri/by-path/pci-%s-card", strings.ToLower(rocmDevice.PCIBus)),
+			fmt.Sprintf("/dev/dri/by-path/pci-%s-render", strings.ToLower(rocmDevice.PCIBus)),
+		}
+		for _, symlink := range devPaths {
+			resolved, err := filepath.EvalSymlinks(symlink)
```
stoksc, line 264: are there some tools that may look where the symlink was, not just the absolute paths (since I guess we're only mounting those)?

agent/internal/containers.go
```diff
@@ -222,6 +242,45 @@ func gpuDeviceRequests(cont cproto.Container) []dcontainer.DeviceRequest {
...
+func injectRocmDeviceRequests(cont cproto.Container, hostConfig *dcontainer.HostConfig) error {
+	uuids := cont.DeviceUUIDsByType(device.ROCM)
+
+	if len(uuids) == 0 {
+		return errors.New("no rocm device uuids")
+	}
+
+	hostConfig.SecurityOpt = append(
+		hostConfig.SecurityOpt, "seccomp=unconfined")
+	hostConfig.GroupAdd = append(hostConfig.GroupAdd, "video")
```
stoksc, line 254: why do we need to do this? can you leave comments or point me to docs so i can understand?

stoksc (commented):
code looks great, rocm is new to me so i want to take a second to read some docs

examples/graphs/proteins_pytorch_geometric/adaptive.yaml
```diff
@@ -33,4 +33,4 @@ entrypoint: model_def:GraphConvTrial
 environment:
   image:
     cpu: determinedai/environments:py-3.7-pytorch-1.9-lightning-1.3-tf-2.4-cpu-b06dafb
-    gpu: determinedai/environments:cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-b06dafb
+    cuda: determinedai/environments:cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-b06dafb
```
rb-determined-ai, line 36: Why do some examples have cpu and cuda images, but some examples you removed the environment.image entirely?

harness/determined/common/schemas/expconf/_v0.py
```diff
@@ -221,58 +221,84 @@ def __init__(
...
+            self.cuda = self.gpu
+        # Equalize values.
+        self.gpu = self.cuda
+
+
+class EnvironmentImageV0(schemas.SchemaBase, CudaCompatMixin):
     _id = "http://determined.ai/schemas/expconf/v0/environment-image.json"
     cpu: Optional[str] = None
+    cuda: Optional[str] = None
     gpu: Optional[str] = None
```
rb-determined-ai, line 238: I don't think that there should be a `gpu` field here anymore.

Instead, do the translation in `from_dict` instead, the same as the go code would do the translation while parsing.

Probably this means you can make your `_ensure_cuda_compat` into a simple function rather than a mixin.

harness/determined/common/schemas/expconf/_v0.py
```diff
@@ -221,58 +221,84 @@ def __init__(
...
-                "determinedai/environments:cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-825e8ee"
-            )
+        if self.rocm is None:
+            self.rocm = "determinedai/environments:rocm-4.2-pytorch-1.9-tf-2.5-rocm-4537b95"
+
+        self.cuda = (
+            self.gpu
+            or "determinedai/environments:cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-825e8ee"
+        )
+        self.gpu = self.cuda
```
rb-determined-ai, line 272: looks like you are still setting up the experiment config to be backwards compatible, but it's still dead code that still hasn't been given to users, so I don't think there's any need to.

master/pkg/model/environment_config.go
```diff
@@ -38,7 +38,10 @@ type Environment struct {
 
 // RuntimeItem configures the runtime image.
 type RuntimeItem struct {
-	CPU string `json:"cpu,omitempty"`
+	CPU  string `json:"cpu,omitempty"`
+	CUDA string `json:"cuda,omitempty"`
+	ROCM string `json:"rocm,omitempty"`
+	// Deprecated: use CUDA instead.
 	GPU string `json:"gpu,omitempty"`
```
rb-determined-ai, line 45: No need to have deprecated fields in go structs, since there's no user code with access to these.  Structuring it this way just leaks deprecated configs into the wider go code.

master/pkg/model/environment_config.go
```diff
@@ -47,7 +50,9 @@ func (r *RuntimeItem) UnmarshalJSON(data []byte) error {
 	var plain string
 	if err := json.Unmarshal(data, &plain); err == nil {
 		r.CPU = plain
-		r.GPU = plain
+		r.ROCM = plain
+		r.CUDA = plain
+		r.GPU = r.CUDA
 		return nil
 	}
 	type DefaultParser RuntimeItem
```
rb-determined-ai, line 58: Either have a temporary struct here that includes the deprecated GPU field or and have some logic to shim it into a modern, or just parse it into a `map[string]string`.  But this is where the `gpu` field from old configs should be stripped and should never make it further in to the go code.

(well, here and in the copy inside `pkg/schemas/expconf` and in RuntimeItems and in the copy of RuntimeItems in `pkg/schemas/expconf`)

master/pkg/schemas/expconf/legacy_test.go
```diff
@@ -214,8 +216,10 @@ func TestLegacyConfig(t *testing.T) {
 				},
 				bindMounts: BindMountsConfig{},
 				envvars: EnvironmentVariablesMap{
-					RawCPU: []string{"HOME=/where/the/heart/is"},
-					RawGPU: []string{"HOME=/where/the/cuda/is"},
+					RawCPU:  []string{"HOME=/where/the/heart/is"},
+					RawCUDA: []string{"HOME=/where/the/cuda/is"},
+					RawGPU:  []string{"HOME=/where/the/cuda/is"},
+					RawROCM: []string{},
```
rb-determined-ai, line 222: seems like if the legacy experiment config has a `gpu` field, that should populate the modern `CUDA` and `ROCM` fields, no?

schemas/test_cases/v0/unions.yaml
```diff
@@ -154,8 +157,11 @@
     image:
       cpu: "alpine"
       gpu: "alpine"
```
rb-determined-ai, line 159: this one should be `cuda: "alpine"` I think

examples/graphs/proteins_pytorch_geometric/adaptive.yaml
ioga: There was a number of PTL tests that had images explicitly specified but they in fact were the default images. This is the only experiment which is "pinned" (because of the reasons described in README).

master/pkg/model/environment_config.go
ioga: got it done through a shim with two-stage parsing of sorts.

webui/react/src/types.ts
ioga: it turned out it's also used in the code (e.g. `overview.ROCM` in `ClusterOverview.tsx`), not just UI display.

stoksc (approved):
lgtm, pending other reviewers of course

webui/react/src/types.ts
hkang1: ah gotcha

docs/sysadmin-basics/cluster-config.txt
```diff
@@ -229,13 +229,14 @@ The master supports the following configuration settings:
    -  ``image``: Defines the default docker image to use when executing the workload. If a docker
       image is specified in the :ref:`experiment config <exp-environment-image>` this default is
       overriden. This image must be accessible via ``docker pull`` to every Determined agent machine
-      in the cluster. Users can use different container images for GPU vs. CPU agents differently by
-      specifying a dict with two keys, ``cpu`` and ``gpu``. Default values:
+      in the cluster. Users can configure different container images for NVIDIA GPU tasks using
+      ``cuda`` key (``gpu`` prior to 0.17.6), CPU tasks using ``cpu`` key, and ROCm (AMD GPU) tasks
```
rb-determined-ai, line 233: non-blocking: I think we normally don't embed historical docs into our present docs, so IMO you would be fine simply saying `cuda` here and ignoring `gpu`.  But I also don't think it hurts to write it how you have it.

harness/determined/common/schemas/expconf/_v0.py
```diff
@@ -224,44 +224,56 @@ def __init__(
...
     @classmethod
     def from_dict(cls, d: Union[dict, str], prevalidated: bool = False) -> "EnvironmentImageV0":
         # Accept either a string or a map of strings to strings.
         if isinstance(d, str):
-            d = {"cpu": d, "gpu": d}
-        return super().from_dict(d, prevalidated)
+            d = {"cpu": d, "cuda": d, "rocm": d}
+        if "cuda" not in d and "gpu" in d:
+            d["cuda"] = d["gpu"]
+            del d["gpu"]
```
rb-determined-ai, line 246: Non-blocking:

Would this be a more desirable migration strategy:
```python
if "gpu" in d:
    d["cuda"] = d["gpu"]
    d["rocm"] = d["gpu"]
    del d["gpu"]
```

Pros I can think of:
- forking an existing experiment where the user original set `image: alpine` (which we converted to `cpu: alpine`/`gpu: alpine`) will still render with `rocm: alpine`
 
Cons I can think of:
- `gpu` will have a unique behavior that some users might grow to depend on, making it potentially hard to remove later.
- when forking, using an old `gpu` default image as the `rocm` image is probably unhelpful.

**Update**: I think the last con has convinced me that I like what you have best, but I'm still posting this for your consideration.

harness/determined/common/schemas/expconf/_v0.py
```diff
@@ -224,44 +224,56 @@ def __init__(
...
+
+        result = super().from_dict(d, prevalidated)
+        return result
```
rb-determined-ai, line 249: why the temporary variable: `result = ... ; return result`?

master/pkg/model/environment_config.go
rb-determined-ai: works for me

schemas/expconf/v0/environment-image.json
```diff
@@ -3,7 +3,7 @@
     "$id": "http://determined.ai/schemas/expconf/v0/environment-image.json",
     "title": "EnvironmentImage",
     "union": {
-        "defaultMessage": "is neither a string nor a map of cpu/gpu to strings",
+        "defaultMessage": "is neither a string nor a map of cpu, cuda/gpu, or rocm to strings",
```
rb-determined-ai, line 6: Can we remove the deprecated name `gpu` from the error message?  I want people who see this message to be clearly guided by it toward the modern usage.

schemas/expconf/v0/environment-variables.json
```diff
@@ -3,7 +3,7 @@
     "$id": "http://determined.ai/schemas/expconf/v0/environment-variables.json",
     "title": "EnvironmentVariables",
     "union": {
-        "defaultMessage": "is neither a list of strings nor a map of cpu/gpu to lists of strings",
+        "defaultMessage": "is neither a list of strings nor a map of cpu, cuda/gpu, or rocm to lists of strings",
```
rb-determined-ai, line 6: ditto

rb-determined-ai (approved):
experiment config stuff looks great

docs/sysadmin-basics/cluster-config.txt
ioga: yep, I believe it'd be helpful reference for existing users who'd be searching for `gpu` to see that information inline.

harness/determined/common/schemas/expconf/_v0.py
ioga: >when forking, using an old gpu default image as the rocm image is probably unhelpful.
right. I also don't see how that can ever work.

PR #3285, opened 2021-12-08
fix: Return last_updated_time on model versions

mapmeld:
## Description

This fixes a bug where the Model Version API always returns `null` for `last_updated_time`.

Reproducing the bug: if you update the description or other attribute of a model version, you'll see the updated time does not change - it is falling back to the creation time.

## Checklist


EmilyBonar (approved):
LGTM, thanks for catching this

PR #3304, opened 2021-12-10
fix: improve CPU core count parsing on agents with CPU slots.

ioga:
## Description

Agent CPU slot detection was incorrectly parsing `cpuinfo` on linux. As a result, `det slot list` will display CPU-based slot device brand as if it only uses a single core, e.g. `Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz x 1 physical cores`, even though in reality it uses all 32 cores.

This PR fixes that, and also tries to handle multi-socket systems with different CPU models, if that ever happens.

## Test Plan

Run a CPU-slotted agent on linux and confirm it reports a correct number of cores.

## Checklist

### DIFF
--- a/agent/internal/detect.go
+++ b/agent/internal/detect.go
@@ -67,8 +67,22 @@ func detectCPUs() ([]device.Device, error) {
 	case len(cpuInfo) == 0:
 		return nil, errors.New("no CPUs detected")
 	default:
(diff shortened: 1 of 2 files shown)


stoksc (approved):
lgtm

PR #3306, opened 2021-12-10
chore: implement closeBar feature for omnibar

hamidzr:
## Description

implement the missing closeBar feature. 
keep the previous command on close
improve omnibar action error handling

## Commentary (optional)

I've found myself wanting the omnibar to close after a successful command execution. 

## Checklist

### DIFF
--- a/webui/react/src/App.tsx
+++ b/webui/react/src/App.tsx
@@ -125,7 +125,7 @@ const AppView: React.FC = () => {
           <Button onClick={refreshPage}>Try Again</Button>
         </PageMessage>
       )}
-      <Omnibar visible={ui.omnibar.isShowing} />
+      <Omnibar />
     </div>
   );
 };
--- a/webui/react/src/App.tsx
+++ b/webui/react/src/App.tsx
@@ -125,7 +125,7 @@ const AppView: React.FC = () => {
           <Button onClick={refreshPage}>Try Again</Button>
         </PageMessage>
       )}
-      {ui.omnibar.isShowing && <Omnibar />}
+      <Omnibar visible={ui.omnibar.isShowing} />
     </div>
   );
 };
--- a/webui/react/src/omnibar/Omnibar.tsx
+++ b/webui/react/src/omnibar/Omnibar.tsx
@@ -2,6 +2,7 @@ import OmnibarNpm from 'omnibar';
 import React, { useCallback, useEffect } from 'react';
 
 import { StoreAction, useStoreDispatch } from 'contexts/Store';
+import handleError from 'ErrorHandler';
 import * as Tree from 'omnibar/tree-extension/index';
 import TreeNode from 'omnibar/tree-extension/TreeNode';
 import { BaseNode } from 'omnibar/tree-extension/types';
@@ -35,8 +36,12 @@ const Omnibar: React.FC<Props> = ({ visible }) => {
 
     if (!input) return;
     if (isTreeNode(item)) {
-      const close = await Tree.onAction(input, item, query);
-      if (close) hideBar();
+      try {
+        const close = await Tree.onAction(input, item, query);
+        if (close) hideBar();
+      } catch (e) {
+        handleError(e);
+      }
     }
   }, [ hideBar ]);
 
--- a/webui/react/src/omnibar/Omnibar.tsx
+++ b/webui/react/src/omnibar/Omnibar.tsx
@@ -2,11 +2,11 @@ import OmnibarNpm from 'omnibar';
 import React, { useCallback, useEffect } from 'react';
 
 import { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
-import handleError from 'ErrorHandler';
 import * as Tree from 'omnibar/tree-extension/index';
 import TreeNode from 'omnibar/tree-extension/TreeNode';
 import { BaseNode } from 'omnibar/tree-extension/types';
 import { isTreeNode } from 'omnibar/tree-extension/utils';
+import handleError from 'utils/error';
 
 import css from './Omnibar.module.scss';
 
--- a/webui/react/src/omnibar/Omnibar.tsx
+++ b/webui/react/src/omnibar/Omnibar.tsx
@@ -25,7 +25,7 @@ const Omnibar: React.FC<Props> = ({ visible }) => {
     [ storeDispatch ],
   );
 
-  const onAction = useCallback((item, query) => {
+  const onAction = useCallback(async (item, query) => {
     /*
     Ideally we wouldn't need to access the element like this.
     A potential option is use the value prop in combinatio with encoding the tree path into
@@ -35,9 +35,10 @@ const Omnibar: React.FC<Props> = ({ visible }) => {
 
     if (!input) return;
     if (isTreeNode(item)) {
-      return Tree.onAction(input, item, query);
+      const close = await Tree.onAction(input, item, query);
+      if (close) hideBar();
     }
-  }, []);
+  }, [ hideBar ]);
 
   useEffect(() => {
     if (visible) {
--- a/webui/react/src/omnibar/Omnibar.tsx
+++ b/webui/react/src/omnibar/Omnibar.tsx
@@ -37,8 +37,8 @@ const Omnibar: React.FC<Props> = ({ visible }) => {
     if (!input) return;
     if (isTreeNode(item)) {
       try {
-        const close = await Tree.onAction(input, item, query);
-        if (close) hideBar();
+        await Tree.onAction(input, item, query);
+        if (item.closeBar) hideBar();
       } catch (e) {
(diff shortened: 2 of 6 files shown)


hkang1 (commented):
The changes overall look good, but when interacting with the bar, there are some weird issues of commands not taking after a little while. Especially when you use backspace the backtrack commands.

The omnibar actually doesn't close for me upon hitting the enter key after a command such as `goto tasks` or `goto experiments`. The `Action Completed` toast pops up but the bar remains afterwards.

Here's an example of the command bar not executing at all:
Screen Shot 2021-12-15 at 12 26 44 PM

Here's an example where `goto experiment` was in the bar and I selected the `experiment <id>` option but it picked the `goto experiments` option instead:
Screen Shot 2021-12-15 at 12 28 23 PM

hamidzr:
I should have updated the title. I originally changed to to autoclose everytime but then switched to implementing a feature I had in mind before where the action determines if the bar should close or not

hamidzr:
Thanks for reviewing.

> The changes overall look good, but when interacting with the bar, there are some weird issues of commands not taking after a little while. Especially when you use backspace the backtrack commands.

Are these intorduced in this pr or general existing issues with our omnibar? I do wanna address all the issues
but if they're not introduced here I'd work on them in their own PR.

> The omnibar actually doesn't close for me upon hitting the enter key after a command such as goto tasks or goto experiments. The Action Completed toast pops up but the bar remains afterwards.

This is expected. this wouldn't make the omnibar close on all commands (that was what I was going for intially but instead I opted to finish implementing `closeBar` attribute). It would instead only close actions that have `closeBar` set to true.


> Here's an example where goto experiment was in the bar and I selected the experiment <id> option but it picked the goto experiments option instead:

We have a known issue of this happening when the cursor is over an option when the omnibar launches if this
is a separate issue, is this easily reproducable? could you add a bit more details on it?

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -1,37 +1,59 @@
...
 
+/*
+Ideally we wouldn't need to access the element like this.
+A potential option is use the value prop in combinatio with encoding the tree path into
+the options returned by the tree extension.
+*/
+const omnibarInput = () => document.querySelector(
+  '#omnibar input[type="text"]',
+) as (HTMLInputElement | null);
+
```
hkang1, line 21: **suggestion:** I think the module of `omnibar` should expose the input instead of having the consumer going the round about way to get this.

One option is to use the `forwardRef` in `OmnibarNpm` to provide a reference to the input element.

Another option (this might be better) is to provide the `input` element through a handler prop. For example, in your use case, it looks like the goal is to focus on the `input` element when the omnibar opens. So the prop would be something like:
```
interface OmnibarNpmProps {
  onShow: (input: HTMLInputElement) => void;
  ...
}
```

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -1,37 +1,59 @@
...
+      }
+    }
+  }, [ hideBar ]);
+
+  useEffect(() => {
+    const input: HTMLInputElement|null = omnibarInput();
+    if (ui.omnibar.isShowing) {
+      if (input) {
+        input.focus();
+      }
```
hkang1, line 51: **question:** shouldn't this work automatically with the `autoFocus={true}`? Also `={true}` might be unnecessary.

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -1,37 +1,59 @@
...
     }
-  }, []);
+  }, [ ui.omnibar.isShowing ]);
 
   return (
-    <div className={css.base}>
+    <div className={css.base} style={{ display: ui.omnibar.isShowing ? 'unset' : 'none' }}>
```
hkang1, line 56: **question:** I thought about this a bit more. Is there a reason why the omnibar state needs to be handled outside of the Omnibar? Can all the omnibar related state (such as `ui.omnibar.isShowing` be handled within `Omnibar.tsx` itself? Such that `App.tsx` can just simply embed `<Omnibar />` and be done.

If it's because of the App level key emitter, you can just create your own key listener inside `Omnibar`.

webui/react/src/App.tsx
```diff
@@ -124,7 +124,7 @@ const AppView: React.FC = () => {
           <Button onClick={refreshPage}>Try Again</Button>
         </PageMessage>
       )}
-      {ui.omnibar.isShowing && <Omnibar />}
+      <Omnibar />
```
hkang1, line 127: 👍

hkang1 (commented):
I've left some comments throughout. Take a look at them and let me know what you think. Happy to discuss them if you want to.

webui/react/src/omnibar/Omnibar.tsx
hamidzr: I tried it but it did not not sure why.

webui/react/src/omnibar/Omnibar.tsx
hamidzr: valid point but since we're not adding to app.tsx here I'd like to leave this off.

hamidzr:
I updated the branch. IMO the open comments are not related to the PR's goal and I'd like to land the changes proposed here without addressing the rest as we're not putting much time into this feature now

webui/react/src/omnibar/tree-extension/trees/actions.tsx
```diff
@@ -5,6 +5,7 @@ import root from 'omnibar/tree-extension/trees';
 import { FinalAction } from 'omnibar/tree-extension/types';
 import { dfsStaticRoutes } from 'omnibar/tree-extension/utils';
 import { routeToReactUrl } from 'shared/utils/routes';
+/** generates a handler that alerts when called */
```
hkang1, line 8: **super nit:** Format for one liner comments
```
// Generate a handler that alerts when called.
```

webui/react/src/omnibar/tree-extension/trees/dev.ts
```diff
@@ -42,6 +50,7 @@ const dev: TreeNode[] = [
     title: 'resetLocalStorage',
   },
   {
+    closeBar: true,
```
hkang1, line 53: **question:** The `closeBar` property is working nicely. What's your approach in terms of figuring out what commands should/shouldn't close? Like I was trying out `dev serverAddress reset` and was expecting the bar to close upon success.

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -1,37 +1,59 @@
...
+        handleError(e);
+      }
+    }
+  }, [ hideBar ]);
+
+  useEffect(() => {
+    const input: HTMLInputElement|null = omnibarInput();
+    if (ui.omnibar.isShowing) {
+      if (input) {
+        input.focus();
```
hkang1, line 50: **suggestion:** add `input.select();` below the focus? With the preservation of the previous commands, I find that majority of the time I want to do a different command from the previous so having all of it highlighted by default saves me the trouble of having to do a `ctrl/cmd + A` to highlight everything, then hit delete or enter a new command.

hkang1 (approved):
LGTM. Some minor suggestions and question about what should and shouldn't close after commands.

webui/react/src/omnibar/Omnibar.tsx
```diff
@@ -1,37 +1,59 @@
...
 import { isTreeNode } from 'omnibar/tree-extension/utils';
+import handleError from 'utils/error';
 
 import css from './Omnibar.module.scss';
 
+/*
+Ideally we wouldn't need to access the element like this.
+A potential option is use the value prop in combinatio with encoding the tree path into
+the options returned by the tree extension.
+*/
```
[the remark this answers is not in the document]
hamidzr: 👍🏼 left over from last year..

webui/react/src/omnibar/tree-extension/trees/dev.ts
hamidzr: hmm, these are open to suggestions. generally, I was thinking

- if this command succeeds am I likely to want to run a similar command immediately?

serveraddres reset seems like a good candidate to close the bar after

webui/react/src/omnibar/Omnibar.tsx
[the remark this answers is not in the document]
hamidzr: updated. although in this case I don't know if it's helpful to have this comment picked up and showed by the IDE everytime `omnibarInput` is hovered.

webui/react/src/omnibar/tree-extension/trees/dev.ts
hamidzr: I added closeBar to some more commands. happy to tune that somemore after.

PR #3310, opened 2021-12-13
chore: rename Generic API to Core API [DET-6243]

rb-determined-ai:
"Generic" was too... generic.  It doesn't communicate anything.

"Lite" was proposed, but suggests an API that is not full-featured.

"Sprinkle" was proposed, but its name is based too closely on our
history of plugin designs, which won't always be relevant.

"Core" suggests an API that is low-level and also suggests that other
higher-level APIs are likely built on top of it.


harness/determined/layers/_workload_sequencer.py
```diff
@@ -87,17 +87,17 @@ def __init__(
 
     def __init__(
         self,
-        generic_context: _generic.Context,
+        core_context: _core.Context,
         env: det.EnvContext,
     ) -> None:
-        self.generic = generic_context
+        self.core = core_context
```
ioga, line 93: maybe it'd be better for readability to do:

```suggestion
        self.core_context = core_context
```

ioga (commented):
seems like the tests also need some renaming.

ioga (approved):

PR #3313, opened 2021-12-13
fix: collect metrics from all agents [DET-6332]

stoksc:
## Description
This change corrects usage of rank/cross_rank so that the profiler correctly knows it should collect metrics when local rank is 0.

## Test Plan
- [x] test on latest master, add a test, too

## Checklist


rb-determined-ai (approved):
might be nice to rename `global_rank` to just `rank`, since that's what we call it literally everywhere else.

But certainly that's not a blocker here.

PR #3315, opened 2021-12-13
fix: handle calls to old command endpoints, make it harder to crash cmd managers [DET-6336]

stoksc:
## Description
This change updates the command manager code to not crash permanently when receiving API calls from echo routed through the actor system. The actor's handling of these calls was removed but the system still could route calls to it, so, for example, an old CLI could crash commands until a restart occurred. We still need to route to the actors below `/commands/...`, `/shells/...`, etc, for the event endpoints et al, so the routing cannot just be purged.

## Test Plan
- [x] make sure an old cli can't crash us

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -64,11 +64,12 @@ func createGenericCommandActor(
 	a, _ := ctx.ActorOf(cmd.taskID, cmd)
 	summaryFut := ctx.Ask(a, getSummary{})
 	if err := summaryFut.Error(); err != nil {
-		ctx.Respond(errors.Wrap(err, "failed to create generic command"))
-		return nil
+		return errors.Wrap(err, "failed to create generic command")
 	}
-	summary := summaryFut.Get().(summary)
-	ctx.Respond(summary.ID)
+	// Sync with the actor, but we don't really need the summary. actor.Ping works too,
+	// but this makes sure it can form some sort of useful response (ping doesn't actually
+	// hit the receive block).
+	summaryFut.Get()
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


stoksc:
gimme a bit to test this

master/internal/command/command_manager.go
```diff
@@ -33,9 +39,17 @@ func (c *commandManager) Receive(ctx *actor.Context) error {
 	case tasks.GenericCommandSpec:
 		taskID := model.NewTaskID()
 		jobID := model.NewJobID()
-		return createGenericCommandActor(
+		if err := createGenericCommandActor(
```
stoksc, line 42: `createGenericCommandActor` never returned an error, it just `ctx.Respond`'d it because an error returned would return from the actor and crash commands. Just moved the responding up a layer and let it return errors so we didn't accidentally return one one day

master/internal/command/command.go
```diff
@@ -64,11 +64,12 @@ func createGenericCommandActor(
 	a, _ := ctx.ActorOf(cmd.taskID, cmd)
 	summaryFut := ctx.Ask(a, getSummary{})
 	if err := summaryFut.Error(); err != nil {
-		ctx.Respond(errors.Wrap(err, "failed to create generic command"))
-		return nil
+		return errors.Wrap(err, "failed to create generic command")
```
eecsliu, line 67: whats the difference between ctx.Respond and returning an error here?

eecsliu (commented):
Makes sense to me! Just one clarifying question.

master/internal/command/command.go
stoksc: we just return the err and ctx.Respond in the caller. i like to keep ctx.Respond's in the top level to keep the code as non-actory as possible

eecsliu (approved):
👌👌

PR #3316, opened 2021-12-13
ci: update to python3.7.

ioga:
## Description

- switch CI to python3.7.
- the tasks running on a machine image seemed to use the system python, which used to be python3.8, and it was effectively pinned through the CI cache. when I updated the cache buster keys, it ended up using python3.9, which broke another set of stuff.
- we are now reigning this in by using a conda-installed python, similarly to the way we install it in environments repo.
- `mypy==0.790` crashes on circleci with python3.7, both their builds and conda versions. debugging went nowhere, so I had to update it to the latest `mypy==0.910`.
- `mypy` has changed the way they recursively look for files to scan in `0.800`. it turned out we've had a number of files that were not properly scanned, mostly in `tests`. because of a huge number of errors, I've opted to not address them and added `# type: ignore`. I hope when someone works with these files again, they'll remove `type: ignore` and address the issues.
- this change also includes #3320 to help with some debugging.

## Test Plan

CI is green.

## Commentary (optional)

mypy should be an integrated part of the language.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,11 @@ parameters:
   machine-image:
     type: string
     default: ubuntu-2004:202101-01
+  # Some python, go, and react dependencies are cached by circleci via `save_cache`/`restore_cache`.
+  # If the dependencies stay the same, but the circleci code that would produce them is changed,
+  # it may be necessary to invalidate the cache by incrementing this value.
+  # For example, if you change an env variable affecting a build of a python package with the same version,
+  # the old build may be cached, and you may need to invalidate it.
   cache-buster:
     type: string
     default: v1dev10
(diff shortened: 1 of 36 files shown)


.circleci/config.yml
```diff
@@ -27,6 +27,15 @@ parameters:
   det-version:
     type: string
     default: 0.17.6-dev0
+  docker-image:
+    type: string
+    default: cimg/base:2021.11
+  machine-image:
+    type: string
+    default: ubuntu-2004:202101-01
+  cache-buster:
```
rb-determined-ai, line 36: when/why would we update this cache-buster value?  Maybe that's worth a comment?

rb-determined-ai (approved):
looks tedious, but thanks for doing this.

PR #3319, opened 2021-12-14
feat: add config option for Tensorboard

varlogtim:
Removed refineArgs from the following commit:
81de46a5adcbd8c3d5aaafc63444f1de8ae65519

... because this doesn't allow arbitrary args, it tries to be smart and
mucks with what you specified. I am not able to specify:
`tensorboard_args=[-v 9]` with this function in place. We should allow
the user to specify whatever they want.

To test:

`det tensorboard start --config="tensorboard_args=[-v 9]" -t <trialId>`

... Note that tensorboard has debug logs now.


master/internal/api_tensorboard.go
```diff
@@ -259,8 +259,6 @@ func (a *apiServer) LaunchTensorboard(
 	}
 	expConf = schemas.WithDefaults(expConf).(expconf.ExperimentConfig)
 
-	refineArgs(spec.Config.TensorBoardArgs)
```
eecsliu, line 262: should we retain functionality that allows users to specify tensorboard args directly in the command?

varlogtim: Did I miss something? Doesn't `spec.Config.TesnorBoardArgs` just pull from `tensorboard_args`? I thought I was removing us not being able to specify `what ever` as a tensorboard arg. It would change it to `--what ever` ... or `["t", "foo"]` would become `--t --foo`. Is there some functionality that I am missing here?

eecsliu: oh no you're right - it pulls direct from the config. I approve 👍

eecsliu (approved):
👌

PR #3320, opened 2021-12-14
chore: unpin google-cloud dependencies.

ioga:
## Description

- Unpin google-cloud dependencies as we have no reason pinning them anymore.
- Pull in `pytz` which used to be a transient dependency but now is something we use directly.
- `mypy` doesn't quite like the new packages, so we disable it for a single method.

## Checklist

### DIFF
--- a/harness/determined/common/storage/gcs.py
+++ b/harness/determined/common/storage/gcs.py
@@ -2,7 +2,7 @@ import contextlib
 import logging
 import os
 import tempfile
-from typing import Iterator, Optional
+from typing import Iterator, Optional, no_type_check
 
 import requests.exceptions
 import urllib3.exceptions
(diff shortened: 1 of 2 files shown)


harness/determined/common/storage/gcs.py
```diff
@@ -60,6 +60,7 @@ def restore_path(self, storage_id: str) -> Iterator[str]:
         finally:
             self._remove_checkpoint_directory(storage_id)
 
+    @no_type_check
```
rb-determined-ai, line 63: can you explain what this is for?  the [typing docs](https://docs.python.org/3/library/typing.html#typing.no_type_check) say it indicates that 'type annotations are not type hints' but that doesn't make sense to me.

rb-determined-ai (approved):

PR #3321, opened 2021-12-14
chore: workaround upstream torch bug

rb-determined-ai:
setuptools 59.6.0 exposed a bug in pytorch, where pytorch is improperly
importing a particular submodule of distutils.  In old setuptools,
distutils.version was automatically imported, but in new setuptools, it
is not.  For details, see:

    https://github.com/pytorch/pytorch/issues/69894

Since forcing users to upgrade pytorch is not feasible, we intervene by
enforcing that distutils.version is imported before we try to import the
problematic torch.utils.tensorboard submodule.  This workaround will not
help users who import torch.utils.tensorboard directly, but it will
protect users who only import torch.utils.tensorboard because of us.

## Test Plan

I imported `determined.tensorboard.metric_writers.torch` with and without the fix.

## Commentary

Example test failure [here](https://app.circleci.com/pipelines/github/determined-ai/determined/17542/workflows/cd0043c8-11ed-4b6a-a63e-39d9391bc88b/jobs/536910/parallel-runs/1)

### DIFF
--- a/harness/determined/tensorboard/metric_writers/pytorch.py
+++ b/harness/determined/tensorboard/metric_writers/pytorch.py
@@ -1,9 +1,15 @@
 from typing import Any, Union
 
 import numpy as np
-from torch.utils.tensorboard import SummaryWriter
 
 from determined import tensorboard
+
+# As of torch v1.9.0, torch.utils.tensorboard has a bug that is exposed by setuptools 59.6.0.  The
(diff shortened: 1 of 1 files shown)


ioga (approved):

mpkouznetsov (approved):
Makes sense.

PR #3322, opened 2021-12-14
chore: print trial logs during test failure [DET-6347]

varlogtim:
Tested locally.

### DIFF
--- a/e2e_tests/tests/cluster/test_checkpoints.py
+++ b/e2e_tests/tests/cluster/test_checkpoints.py
@@ -1,4 +1,5 @@
 import operator
+import sys
 import tempfile
 import time
 from typing import Dict, Set
@@ -64,7 +65,7 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
                 by_state.setdefault(checkpoint["state"], set()).add(checkpoint["total_batches"])
 
             if by_state == result:
-                all_checkpoints.append((config, checkpoints))
+                all_checkpoints.append((config, checkpoints, trials[0]["id"]))
                 break
 
             if retry + 1 == retries:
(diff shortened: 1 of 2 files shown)


e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -64,7 +65,7 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
                 by_state.setdefault(checkpoint["state"], set()).add(checkpoint["total_batches"])
 
             if by_state == result:
-                all_checkpoints.append((config, checkpoints))
+                all_checkpoints.append((config, checkpoints, trials[0]["id"]))
```
rb-determined-ai, line 68: no need to put the same trial id in this data structure over and over.  There's only one trial in the experiment, just save it as a variable.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -79,7 +80,7 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
     for i in range(max_checks):
         time.sleep(1)
```
rb-determined-ai, line 81: oh snap, is this just a matter of the timeout being too short?

that comment above:

```python
    # Check that the actual checkpoint storage (for shared_fs) reflects the
    # deletions. We want to wait for the GC containers to exit, so check
    # repeatedly with a timeout.
```

would indicate that if checkpoint gc doesn't finish in 30 seconds then we would have this problem.

Can you read through the master logs and find out if this is the case?  I still think it's worth adding better logging to the test case in any case.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -91,6 +92,7 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
                     elif checkpoint["state"] == "DELETED":
                         try:
                             with storage_manager.restore_path(storage_id):
+                                exp.print_trial_logs(trial_id)
                                 raise AssertionError("checkpoint not deleted")
```
rb-determined-ai, line 96: this strategy, which is basically 'bail on the first failure' is quick to write but gives very little information when it actually fails.

Instead of 'bail on first failure', can you make this a 'gather a list of failures and raise an exception with all failures'?

varlogtim:
This one went in here: https://github.com/determined-ai/determined/pull/3353

PR #3328, opened 2021-12-15
feat: support agent on Apple Silicon without Rosetta

mpkouznetsov:
## Description

Agent exits on Apple with M1 chip because of an error when obtaining the CPU info using gopsutil ( https://github.com/shirou/gopsutil/issues/1000 ). This PR works around the error by providing placeholder values for brand and UUID.

## Test Plan

Start Agent successfully on M1 Mac.

## Commentary (optional)

We may choose to wait for the fix to gopsutil or we can merge this fix now and replace it with an upgrade of gopsutil when a fix becomes available there.

## Checklist

### DIFF
--- a/agent/internal/detect.go
+++ b/agent/internal/detect.go
@@ -21,6 +21,10 @@ import (
 	"github.com/determined-ai/determined/master/pkg/device"
 )
 
+const (
+	osDarwin = "darwin"
+)
+
 func (a *agent) detect() error {
 	switch {
 	case a.ArtificialSlots > 0:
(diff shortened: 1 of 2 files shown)


agent/internal/detect.go
```diff
@@ -63,6 +69,19 @@ func (a *agent) detect() error {
 func detectCPUs() ([]device.Device, error) {
 	switch cpuInfo, err := cpu.Info(); {
 	case err != nil:
+		// Apple M1 does not report CPU frequency,
+		// and that can result in an error here
+		if errno, ok := err.(syscall.Errno); ok {
+			switch errno {
+			case syscall.ENOENT:
+				if runtime.GOARCH == "arm64" && runtime.GOOS == osDarwin {
+					return []device.Device{{ID: 0, Brand: "Apple", UUID: "AppleM1", Type: device.CPU}}, nil
```
ioga, line 78: I think we should call it "Apple Silicon" instead of M1, because I foresee M2 next spring or next fall at the latest.

ioga (approved):
lgtm, a small comment on the future-proofed naming.
also I'd say this PR is probably a `feat: support agent on Apple Silicon without Rosetta`, not a chore.

PR #3277, opened 2021-12-07
chore: singleton config (jq)

hamidzr:
## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3330, opened 2021-12-15
fix: fix incorrect preemption status report from Kubernetes RP

hamidzr:
## Description

- dummy k8 resource pool reported inaccurate preemption status when using the `preemption` scheduler plugin
- set up scheduler and Kubernetes config structs to report their own preemption status

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -424,7 +424,7 @@ func (c *command) toV1Job() *jobv1.Job {
 		Name:           c.Config.Description,
 	}
 
-	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) && false // CHECK we can't make command preemptible can we?
+	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) && false
 	j.Priority = int32(config.ReadPriority(j.ResourcePool, &c.Config))
 	j.Weight = config.ReadWeight(j.ResourcePool, &c.Config)
 
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -424,7 +424,7 @@ func (c *command) toV1Job() *jobv1.Job {
 		Name:           c.Config.Description,
 	}
 
-	j.IsPreemptible = config.ReadPreemptionStatus(j.ResourcePool, &c.Config)
+	j.IsPreemptible = config.ReadPreemptionStatus(j.ResourcePool, false) // CHECK we can't make command preemptible can we?
 	j.Priority = int32(config.ReadPriority(j.ResourcePool, &c.Config))
 	j.Weight = config.ReadWeight(j.ResourcePool, &c.Config)
 
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -424,7 +424,7 @@ func (c *command) toV1Job() *jobv1.Job {
 		Name:           c.Config.Description,
 	}
 
-	j.IsPreemptible = config.ReadPreemptionStatus(j.ResourcePool, false) // CHECK we can't make command preemptible can we?
+	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) && false // CHECK we can't make command preemptible can we?
 	j.Priority = int32(config.ReadPriority(j.ResourcePool, &c.Config))
 	j.Weight = config.ReadWeight(j.ResourcePool, &c.Config)
 
--- a/master/internal/config/config.go
+++ b/master/internal/config/config.go
@@ -12,6 +12,7 @@ import (
 	"github.com/determined-ai/determined/master/internal/db"
 	"github.com/determined-ai/determined/master/internal/hpimportance"
 	"github.com/determined-ai/determined/master/internal/resourcemanagers"
+	"github.com/determined-ai/determined/master/internal/resourcemanagers/kubernetes"
 	"github.com/determined-ai/determined/master/pkg/logger"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/pkg/schemas/expconf"
@@ -268,7 +269,7 @@ func ReadRMPreemptionStatus(rpName string) bool {
 		}
 		return config.ResourceManager.AgentRM.Scheduler.GetPreemption()
 	case config.ResourceManager.KubernetesRM != nil:
-		return config.ResourceManager.KubernetesRM.DefaultScheduler == "preemption"
+		return config.ResourceManager.KubernetesRM.DefaultScheduler == kubernetes.PreemptionScheduler
 	default:
 		panic("unexpected resource configuration")
 	}
--- a/master/internal/config/config.go
+++ b/master/internal/config/config.go
@@ -257,9 +257,6 @@ func readRMPreemptionStatus(config *Config, rpName string) bool {
 		if rpConfig.Scheduler != nil {
 			return rpConfig.Scheduler.GetPreemption()
 		}
-		if rpConfig.Provider != nil && rpConfig.Provider.GCP != nil {
-			return rpConfig.Provider.GCP.InstanceType.Preemptible
-		}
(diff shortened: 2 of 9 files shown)


master/internal/config/config.go
```diff
@@ -236,67 +237,42 @@ type ObservabilityConfig struct {
...
 	if conf == nil || conf.Priority == nil {
 		return nil
 	}
 	return conf.Priority.DefaultPriority
 }
 
-// ReadPreemptionStatus resolves the desired preemption status for a job.
-func ReadPreemptionStatus(rpName string, jobConf interface{}) bool {
+// ReadRMPreemptionStatus resolves the preemption status for a resource manager.
+func ReadRMPreemptionStatus(rpName string) bool {
```
eecsliu, line 248: Should we call it `ReadRPPreemptionStatus` ?

eecsliu (approved):
👩‍🍳😗👌

master/internal/config/config.go
hamidzr: hmm that works too. As this is implemented in the case of k8 it doesn't really look at the RP that's why I stayed away from it 🤔

master/internal/config/config.go
```diff
@@ -236,67 +237,45 @@ type ObservabilityConfig struct {
...
-		if config.ResourceManager.KubernetesRM.DefaultScheduler == "preemption" {
-			RMPremption = true
+	switch {
+	case config.ResourceManager.AgentRM != nil:
+		if config.ResourceManager.AgentRM.Scheduler == nil {
+			panic("scheduler not configured")
 		}
+		return config.ResourceManager.AgentRM.Scheduler.GetPreemption()
+	case config.ResourceManager.KubernetesRM != nil:
+		return config.ResourceManager.KubernetesRM.DefaultScheduler == kubernetes.PreemptionScheduler
```
hamidzr, line 275: either this line or this look odd https://github.com/determined-ai/determined/blob/master/master/internal/resourcemanagers/kubernetes_resource_manager.go#L187

hamidzr: looks like this is a bug on the dummy resource pool side.

master/internal/config/config.go
```diff
@@ -236,67 +236,42 @@ type ObservabilityConfig struct {
...
-			return jobPreemptible && RMPremption
-		}
-	}
-
-	if config.ResourceManager.KubernetesRM != nil {
-		if config.ResourceManager.KubernetesRM.DefaultScheduler == "preemption" {
-			RMPremption = true
+	switch {
+	case config.ResourceManager.AgentRM != nil:
+		if config.ResourceManager.AgentRM.Scheduler == nil {
```
stoksc, line 266: https://github.com/determined-ai/determined/blob/95999f97ed3428c1315b3b621784fad4afc84768/master/internal/config/config.go#L152-L154 this should be impossible because of this

master/internal/command/command.go
```diff
@@ -424,7 +424,7 @@ func (c *command) toV1Job() *jobv1.Job {
 		Name:           c.Config.Description,
 	}
 
-	j.IsPreemptible = config.ReadPreemptionStatus(j.ResourcePool, &c.Config)
+	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) && false
```
stoksc, line 427: this is always false?

master/internal/experiment.go
```diff
@@ -581,7 +581,7 @@ func (e *experiment) toV1Job() *jobv1.Job {
 		Name:           e.Config.Name().String(),
 	}
 
-	j.IsPreemptible = config.ReadPreemptionStatus(j.ResourcePool, &e.Config)
+	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) // && true
```
stoksc, line 584: what is this comment?

master/internal/resourcemanagers/scheduler_config.go
```diff
@@ -91,6 +91,20 @@ func (s *SchedulerConfig) GetType() string {
 	}
 }
 
+// GetPreemption returns whether the scheduler is set to preempt.
+func (s *SchedulerConfig) GetPreemption() bool {
```
stoksc, line 95: `GetPreemption` -> `PreemptionEnabled` i think sounds better

master/internal/config/config.go
```diff
@@ -236,67 +236,42 @@ type ObservabilityConfig struct {
...
 		return nil
 	}
 	return conf.Priority.DefaultPriority
 }
 
-// ReadPreemptionStatus resolves the desired preemption status for a job.
-func ReadPreemptionStatus(rpName string, jobConf interface{}) bool {
+// ReadRMPreemptionStatus resolves the preemption status for a resource manager.
+func ReadRMPreemptionStatus(rpName string) bool {
 	config := GetMasterConfig()
```
stoksc, line 248: didn't get to look at the previous pr, this master config singleton thing is pretty strange to me, why did we need this? things like this have a tendency to cause weird bugs/makes the system really hard to break up / untangle later. also it seems the `sync.Once` is just so GetMasterConfig works in tests (it is always set otherwise)? they should probably just set a default config, since this could cause a really bad bug if it ever gets called wrong?

stoksc (commented):
sorry after the fact review, just happened to look when i saw the commit land

master/internal/config/config.go
```diff
@@ -236,67 +236,42 @@ type ObservabilityConfig struct {
...
 	if conf == nil || conf.Priority == nil {
 		return nil
 	}
 	return conf.Priority.DefaultPriority
 }
 
-// ReadPreemptionStatus resolves the desired preemption status for a job.
-func ReadPreemptionStatus(rpName string, jobConf interface{}) bool {
+// ReadRMPreemptionStatus resolves the preemption status for a resource manager.
+func ReadRMPreemptionStatus(rpName string) bool {
```
stoksc, line 247: I think this should be part of the RM interface, whether it supports preemption or not. I don't think callers should be digging into the config to try to figure this out. On top of making it more clear what the RP interface is within the system and encapsulating it all, it's more flexible (e.g. if preemption could be disabled dynamically on RPs or if we don't know if a RP supports preemption or not until we make some API call that would be made by the RM itself)

stoksc: I think this highlights my gripe with the singleton config too; more than likely it is not the right place to get the information; the RM should get its configuration and construct its state, and if you need to know something about that state, you should ask it, not try to infer it from its configuration.

master/internal/experiment.go
hamidzr: I was trying to signify how this is calculated. as of now experiments on their own are always marked as preemptible, hence the constant true, and commands always not preemptible and consequently [the constant false here](https://github.com/determined-ai/determined/pull/3330#discussion_r771533125) 
I thought about linking these const values in some way to account for future changes eg using a shared var or a method on experiments and commands eg `PreemptionEnabled` wdy think?

master/internal/command/command.go
hamidzr: ture. I wanted to keep this similar to how IsPreemtible is calculated on Experiment jobs but it does seem wasteful. I wrote down more on the other one. should we avoid throwing a `false` here without it being linked to anywhere else

master/internal/config/config.go
hamidzr: should we chat about this over at the [the singleton config pr](https://github.com/determined-ai/determined/pull/3277)
I might have not gotten the second part correctly but the sync.Once is to ensure the singleton isn't initialized twice if the correct interfaces are used.

master/internal/resourcemanagers/scheduler_config.go
hamidzr: that is better I can update it at once with the other comments

master/internal/config/config.go
stoksc: yeah i can move it over there

PR #3334, opened 2021-12-16
chore: More info when test failes [DET-6347]

varlogtim:


e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -75,25 +76,40 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
...
             for config, checkpoints in all_checkpoints:
                 checkpoint_config = config["checkpoint_storage"]
 
                 storage_manager = storage.build(checkpoint_config, container_path=None)
+                storage_state = {}
                 for checkpoint in checkpoints:
                     storage_id = checkpoint["uuid"]
+                    # Trying again
                     if checkpoint["state"] == "COMPLETED":
                         with storage_manager.restore_path(storage_id):
```
rb-determined-ai, line 94: There's a second type of failure possible here, which is implicit in the code.

If the checkpoint was supposed to exist but doesn't, this line will raise a CheckpointNotFound.  You'll have to catch that error and add an entry into your lists of issues as well.

Otherwise, this looks pretty good.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -75,25 +76,40 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
...
         try:
+            storage_states = []
             for config, checkpoints in all_checkpoints:
                 checkpoint_config = config["checkpoint_storage"]
 
                 storage_manager = storage.build(checkpoint_config, container_path=None)
+                storage_state = {}
                 for checkpoint in checkpoints:
                     storage_id = checkpoint["uuid"]
+                    # Trying again
```
rb-determined-ai, line 92: What does this comment mean?

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -75,25 +76,40 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
     # Check that the actual checkpoint storage (for shared_fs) reflects the
     # deletions. We want to wait for the GC containers to exit, so check
     # repeatedly with a timeout.
+
```
rb-determined-ai, line 79: can you remove this stray empty line?  Seems like an unnecessary change.

e2e_tests/tests/cluster/test_checkpoints.py
varlogtim: Not sure how this got in there, removing it.

e2e_tests/tests/cluster/test_checkpoints.py
varlogtim: Got it. Yeah, this makes more sense after I looked at it today.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -79,21 +80,44 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
...
+                                pass
                         except errors.CheckpointNotFound:
-                            pass
+                            storage_state[storage_id]["deleted"] = True
+                        storage_states.append(storage_state)
+
+            for storage_state in storage_states:
+                for state in storage_state.values():
+                    deleted = state.get("deleted", None)
+                    if deleted is not None and not deleted:
```
rb-determined-ai, line 110:
```python
if state.get("deleted") == False:
```

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -79,21 +80,44 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
...
+                        json_states = json.dumps(storage_states)
+                        raise AssertionError(
+                            f"Some checkpoints were not deleted: JSON:{json_states}"
+                        )
+                    found = state.get("found", None)
+                    if found is not None and not found:
```
rb-determined-ai, line 116:
```python
if state.get("found") == False:
```

rb-determined-ai (approved):
two nits, otherwise looks like a nice improvement.  Next time there is an issue with this test we'll have a much clearer explanation what went wrong.

rb-determined-ai (approved):
Great!

PR #3335, opened 2021-12-16
fix: move deploy dependencies to `extras_require`

schmidt-jake:
## Description
Moves "deploy" dependencies to an extras section. This helps segment dependencies that may not be needed for every use case. In particular, this will help reduce the odds of determined causing dependency conflicts with other python dependencies users need to install for their project.

## Test Plan
Update the existing tests to install the extra and run them.

## Commentary (optional)

What pieces of the documentation need to be updated?

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -202,8 +202,8 @@ commands:
           working_directory: <<parameters.package-location>>
           command: |
             make build
-            pip install --find-links dist <<parameters.package-name>>==<< pipeline.parameters.det-version >>
-            pip install --no-deps --force-reinstall --find-links dist <<parameters.package-name>>==<< pipeline.parameters.det-version >>
+            pip install --find-links dist <<parameters.package-name>>[deploy]==<< pipeline.parameters.det-version >>
+            pip install --no-deps --force-reinstall --find-links dist <<parameters.package-name>>[deploy]==<< pipeline.parameters.det-version >>
   setup-python-venv:
     description: Set up and create Python venv.
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1455,7 +1455,7 @@ jobs:
       # Windows needs special treatment of cryptography for unknown reasons.
       # LMDB also has special requirements that only apply to windows (patch-ng).
       - run: sh -c "if which conda ; then conda install cryptography patch-ng --yes ; fi"
-      - run: pip install --find-links build determined==<< pipeline.parameters.det-version >>
+      - run: pip install --find-links build determined[deploy]==<< pipeline.parameters.det-version >>
       - run: pip freeze --all
       # Allow this to fail, but it is useful for debugging.
       - run: sh -c "pip check || true"
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -42,17 +42,20 @@ setup(
         "tabulate>=0.8.3",
         # det preview-search "pretty-dumps" a sub-yaml with an API added in 0.15.29
         "ruamel.yaml>=0.15.29",
-        # Deploy
-        # "docker[ssh]>=3.7.3",
-        # "google-api-python-client>=1.12.1",
-        # "paramiko>=2.4.2",  # explicitly pull in paramiko to prevent DistributionNotFound error
-        # "docker-compose>=1.13.0",
-        # "tqdm",
-        # "appdirs",
-        # "websocket-client",
         # Telemetry
         "analytics-python",
     ],
+    extras_require={
+        "deploy": [
+            "docker[ssh]>=3.7.3",
+            "google-api-python-client>=1.12.1",
+            "paramiko>=2.4.2",  # explicitly pull in paramiko to prevent DistributionNotFound error
+            "docker-compose>=1.13.0",
+            "tqdm",
+            "appdirs",
+            "websocket-client",
+        ]
+    },
     zip_safe=False,
     entry_points={
         "console_scripts": [
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -49,9 +49,7 @@ setup(
         "docker-compose>=1.13.0",
         "tqdm",
         "appdirs",
-        # docker-compose has a requirement not properly propagated with semi-old pip installations;
-        # so we expose that requirement here.
-        "websocket-client<1",
+        "websocket-client",
         # Telemetry
         "analytics-python",
     ],
--- a/harness/setup.py
+++ b/harness/setup.py
@@ -53,7 +53,7 @@ setup(
             "docker-compose>=1.13.0",
             "tqdm",
             "appdirs",
-            "websocket-client",
+            "websocket-client<1",
         ]
     },
     zip_safe=False,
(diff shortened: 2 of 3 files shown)


rb-determined-ai:
what is the motivation for this change?

I don't think it's a good idea for us to ship the `det` command (which contains `det deploy`) with missing dependencies.  Requiring many of our users to remember to `pip install determined[deploy]` rather than just `determined` is probably not going to happen.

schmidt-jake:
> what is the motivation for this change?
> 
> I don't think it's a good idea for us to ship the `det` command (which contains `det deploy`) with missing dependencies. Requiring many of our users to remember to `pip install determined[deploy]` rather than just `determined` is probably not going to happen.

See https://github.com/determined-ai/determined/pull/3320 for a little more context — but the issue is the deploy dependencies are causing dependency conflicts in my project that I can't circumnavigate.

rb-determined-ai:
can you share more information about your dependencies? Perhaps even a requirements.txt?

Probably it'll be more likely that we can help you solve the particular problem... I don't think we can make the change you're proposing here.

schmidt-jake:
> can you share more information about your dependencies? Perhaps even a requirements.txt?
> 
> Probably it'll be more likely that we can help you solve the particular problem... I don't think we can make the change you're proposing here.

Our dependencies are organized in multiple stages / groups. We have a set of core project dependencies, dependencies for model inference which are constrained by the core dependencies, then dependencies for training that are constrained by the core and inference dependencies. `determined` is in this last group. We have a build tool resolve + lock dependencies starting with the core group and traversing the constraints, so by the time it reaches `determined` there are pins in google-cloud-storage and `websockets-client` that are unresolvable.

Having `extras` is a common pattern with clear advantages: it helps keep dependencies organized, mitigates dependency conflicts such as I'm experiencing, offers users more flexible installation options, and reduces environment size.

I'm curious how many of your users who run `det deploy` would be unfamiliar / uncomfortable with running `pip install determined[deploy]`. In my use case, only our infra team needs deployment dependencies — most of our determined users just need the SDK and therefore it seems that more people would benefit from this change than would be impacted by it.

johnurbanik:
> what is the motivation for this change?
> 
> I don't think it's a good idea for us to ship the `det` command (which contains `det deploy`) with missing dependencies. Requiring many of our users to remember to `pip install determined[deploy]` rather than just `determined` is probably not going to happen.

I definitely hear where you're coming from - breaking changes aren't to be made lightly. Is more of your hesitation about requiring people to remember the (not so common) `package[extra]` syntax? I can't say I fully understand the decision to centralize things and deprecate `determined-deploy`



Coming from the other side (especially as someone used to the `package[extra]` format)...I'd say it's pretty suboptimal that anyone using the determined api/cli needs to install docker + an old version of docker compose. Even if we were able to get around the dependency issues that @schmidt-jake is having, that is quite a bit of bloat.

rb-determined-ai:
> In my use case, only our infra team needs deployment dependencies — most of our determined users just need the SDK and therefore it seems that more people would benefit from this change than would be impacted by it.

We have a lot of users where there is no distinction between sysadmins and ML users.  I would say that your team is among the most advanced users of our system, certainly not the norm.  We gathered a lot of feedback about the installation process, and we consistently heard that users were confused by the separation of the cli, the training apis, and det-deploy.  We basically chose to optimize the onboarding experience over the technical merits of separated dependencies.  From our perspective, increasing adoption is one of our top priorities, and this decision is in-line with that.

> Having extras is a common pattern with clear advantages

To be clear, the reason that we can't ship this request _in its current form_ is not because we can't reverse the above decision or find some happy-medium, it is because this PR is using the `extras_require` incorrectly.  `docker-compose` is not an optional dependency of the code in the `determined` package.  It is required to be installed; there's no handling of import errors, there's no fallback mechanisms, you'll just get a really confusing error message if you use the wrong command.

> Even if we were able to get around the dependency issues that @schmidt-jake is having, that is quite a bit of bloat.

One of the reasons that we weren't too concerned about merging det-deploy into determined was that the bloat from these packages isn't nearly as big as some of the storage libraries, and massively smaller than any training libraries that you need to use the system.  It seems like a tiny win from an optimization perspective.

------

I understand that python packaging can be a nightmare, and we have over time put quite a bit of effort into loosening requirements and supporting as wide of a variety of package versions as possible, and even going through effort to drop dependencies that we didn't feel were necessary.  I would prefer that strategy here if possible.

If you could share with me the library versions that you are running into, I am very happy to help work around the package versions.  If the docker python client and the docker-compose really are that big of a deal, I can bring this up with the team and we can discuss separating the det deploy command in a more proper way.

johnurbanik:
Thanks for the detailed and thoughtful response, Ryan.

> To be clear, the reason that we can't ship this request _in its current form_ is not because we can't reverse the above decision or find some happy-medium, it is because this PR is using the `extras_require` incorrectly. `docker-compose` is not an optional dependency of the code in the `determined` package. It is required to be installed; there's no handling of import errors, there's no fallback mechanisms, you'll just get a really confusing error message if you use the wrong command.

This contrasts with your previous comments, namely:

> I don't think it's a good idea for us to ship the det command (which contains det deploy) with missing dependencies. Requiring many of our users to remember to pip install determined[deploy] rather than just determined is probably not going to happen.

I guess I misread you; it seemed like you were considering this PR a complete non-starter, not that you felt that better error handling is necessary in order to get the PR through. I think @schmidt-jake would be happy to raise some more informative errors in `harness/determined/deploy/*` - that seems to be the only place these deps are used, but if we're wrong that can come out through additional code review.

> If you could share with me the library versions that you are running into, I am very happy to help work around the package versions. If the docker python client and the docker-compose really are that big of a deal, I can bring this up with the team and we can discuss separating the det deploy command in a more proper way.

After some requirements wrangling, it looks like adding artificial deps for `websocket-client < 1` and `pyyaml<6` earlier in our build DAG should work as a temporary fix. There are some bugfixes in `pyyaml==6.0` that we'd like though, so some movement here would still be appreciated if you decide to decline the idea of using `extra_requirements`.

Some other options:
- Install the newer Go based version of `docker-compose` via `distutils` (though that probably means a bunch of wheels)
- Remove `docker-compose` from reqs and add some error messages to `harness/determined/deploy/local/cluster_utils.py` and `determined/tools/scripts/scan-docker-images.sh`, and have people install the newer docker compose via their package manager if they're using those functionalities.





Separately...
> We have a lot of users where there is no distinction between sysadmins and ML users. I would say that your team is among the most advanced users of our system, certainly not the norm. We gathered a lot of feedback about the installation process, and we consistently heard that users were confused by the separation of the cli, the training apis, and det-deploy. We basically chose to optimize the onboarding experience over the technical merits of separated dependencies. From our perspective, increasing adoption is one of our top priorities, and this decision is in-line with that.

If you're concerned about onboarding people, I'd question whether default `pip install determined[deploy]` would be very confusing. Personally, I'd be more concerned about existing users, and I would expect that most ML users would be fine with a different `pip install`, and most sysadmins would be putting this in a script / requirements. But ultimately, if you don't want to accept this, it's your call - we can probably move forward for now with one of the solutions outlined above.

rb-determined-ai:
@johnurbanik after talking off-line, are you ok closing this?  It sounds like you have figured out the dependency issues on your side, and on our side we need to remove the dependency on `docker-compose`, which seems to no longer be under development.

PR #3339, opened 2021-12-17
chore: AdvancedSearcher->Searcher

rb-determined-ai:
Turns out that the master doesn't even need units when configuring the
searcher.  That means that the basic searcher doesn't make much since,
since it only really worked with a large unit (epochs) rather than a
small unit (batches/records), so for now the Core API will only support
SearcherOp-based training loops.

This is a bit unforunate, since a SearcherOp-based training loop
requires an extra layer of for loop that is not totally intuitive.  We
can revisit this at a later time when it is clearer how we can reduce
user friction.  In the mean time, keeping the Core API barebones is a
good idea.

### DIFF
--- a/harness/determined/_core/_searcher.py
+++ b/harness/determined/_core/_searcher.py
@@ -92,8 +92,56 @@ class SearcherOp:
         )
 
 
-class AdvancedSearcher:
-    """A namespaced API to make it clear which searcher-related things go to which API"""
+class Searcher:
+    """
+    Searcher gives direct access to operations emitted by the search algorithm in the master.  Each
+    SearcherOp emitted has a (unitless) length that you should train for, then you complete the op
+    by reporting the validation metric you are searching over.
+
+    It is the user's responsibility to execute the required training.  Since the user configured the
+    length of the searcher in the experiment configuration, the user should know if the unitless
+    length represents epochs, batches, records, etc.
+
+    It is also the user's responsibility to evaluate the model after training and report the correct
+    metric; if you intend to search over a metric called val_accuracy, you should report
+    val_accuracy.
+
+    Lastly, it is recommended (not required) to report progress periodically, so that the webui can
+    accurately reflect current progress.  Progress is another unitless length.
+
+    Example:
+
+    .. code:: python
+
+       # We'll pretend we configured we configured the searcher
+       # in terms of batches, so we will interpet the the op.length as a
+       # count of batches.
+       # Note that you'll have to load your starting point from a
+       # checkpoint if you want to support pausing/continuing training.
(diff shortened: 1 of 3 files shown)


harness/determined/_core/_searcher.py
```diff
@@ -92,8 +92,54 @@ def complete(self, searcher_metric: float) -> None:
...
+               batches_trained += 1;
+
+               # Reporting progress every batch would be expensive:
+               if batches_trained % 1000:
+                   op.report_progress(batches_trained)
+
+           # After training the required amount, pass your searcher
+           # metric to op.complete():
+           val_metrics = my_validate()
+           op.complete(va_metrics["my_searcher_metric"])
```
ioga, line 139:
```suggestion
           op.complete(val_metrics["my_searcher_metric"])
```

harness/determined/_core/_searcher.py
```diff
@@ -92,8 +92,54 @@ def complete(self, searcher_metric: float) -> None:
...
+       batches_trained = 0
+
+       for op in generic_context.searcher.ops():
+           # Train for however long the op requires you to.
+           # Note that op.length is an absolute length, not an
+           # incremental length:
+           while batches_trained < op.length:
+               my_train_batch()
+
+               batches_trained += 1;
```
ioga, line 130:
```suggestion
               batches_trained += 1
```

harness/determined/_core/_searcher.py
```diff
@@ -92,8 +92,54 @@ def complete(self, searcher_metric: float) -> None:
...
+               # Reporting progress every batch would be expensive:
+               if batches_trained % 1000:
+                   op.report_progress(batches_trained)
+
+           # After training the required amount, pass your searcher
+           # metric to op.complete():
+           val_metrics = my_validate()
+           op.complete(va_metrics["my_searcher_metric"])
+
+    Note that reportig metrics is completely independent of the Searcher API.
```
ioga, line 141:
```suggestion
    Note that reporting metrics (e.g. `report_training_metrics`, `report_validation_metrics` calls) is completely independent of the Searcher API.
```

A person reading the docs probably won't have the knowledge and the context of the full "Core API". Including the literal API function names it is referring may be helpful.

ioga (approved):

harness/determined/_core/_searcher.py
rb-determined-ai: oops, some C leaked through

harness/determined/_core/_searcher.py
rb-determined-ai: good idea

PR #3341, opened 2021-12-17
fix: add return to SDK command

schmidt-jake:
## Description
`determined.experimenta.create` was missing a return statement.

## Test Plan
Technically there could be a unit test to check the return value, but I'm not sure how valuable this is, especially if there's a more general system of dynamic type checking, e.g. `typeguard`.

## Checklist

### DIFF
--- a/harness/determined/experimental/_native.py
+++ b/harness/determined/experimental/_native.py
@@ -182,7 +182,7 @@ def create(
     context_dir: str = "",
     command: Optional[List[str]] = None,
     master_url: Optional[str] = None,
-) -> Any:
+) -> int:
     # TODO: Add a reference to the local development tutorial.
     """
     Create an experiment.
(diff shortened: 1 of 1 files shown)


harness/determined/experimental/_native.py
```diff
@@ -235,6 +235,9 @@ def create(
             An optional string to use as the Determined master URL when
             ``local=False``. If not specified, will be inferred from the
             environment variable ``DET_MASTER``.
+
+    Returns:
+        int: the created experiment ID.
```
rb-determined-ai, line 240: Can you share a code snippet of how you are intending to use this in your scripts?

This is a change to a deprecated API.  It's not out of the question but it would help to see what you're trying to use it for.

schmidt-jake: Ahh dang, I didn't know this was deprecated! What it's replaced by?

Use case is to trigger, wait on, and get results of an experiment using the SDK as part of a) some automated workflows and b) integration tests.

rb-determined-ai:
@schmidt-jake after talking off-line, are you ok with closing this in favor of transitioning to the python sdk?  That is a much better way forward, long-term.

schmidt-jake:
Yep that's fine!

PR #3343, opened 2021-12-17
chore: unitless searcher [DET-6359]

rb-determined-ai:
Remove units from the searcher.

Let the client figure out the units from the searcher config for now.


rb-determined-ai:
@stoksc can you have a quick look at the master-side and proto-side changes?

There's also some master-side unit test work, some expconf backwards-compatibility work, and some webui work to be done.  But before I dig any further I would like to know if this approach goes too far or if it's about what you'd expect.

stoksc (commented):
what would the trainer api look like so that the wsq knows what it's actually doing?

we don't surface these units anywhere in the ui right? those just use the steps/vals table to saturate the ui. side note, would those eventually go the same way?

generally, this is fine with me, it think it drops a lot of needless complexity

rb-determined-ai:
> what would the trainer api look like so that the wsq knows what it's actually doing?

The wsq is almost pure legacy, IMO.  I don't think it should be used when we switch to the Trainer APIs.  The new training loops should be based on real epochs, or on batches or whatever.

> we don't surface these units anywhere in the ui right?

Only in config-editing things.  Those will be dealt with.

> side note, would those eventually go the same way?

Would what go away?  The table of metrics, which is currently populated by steps?  Like DET-4980, or something else?

stoksc:
> The wsq is almost pure legacy, IMO. I don't think it should be used when we switch to the Trainer APIs. The new training loops should be based on real epochs, or on batches or whatever.

maybe i'm not grokking the python side, how would the trialapi/wsq code know what to do in this world? would they not still need a unit?

> Would what go away? The table of metrics, which is currently populated by steps? Like DET-4980, or something else?

no, more like the 'arbitrary x-axis' stuff we talked about.

rb-determined-ai:
> how would the trialapi/wsq code know what to do in this world? would they not still need a unit?

Yes, there's a `core_context.searcher.get_configured_units()` here for handling those sorts of legacy cases.  Basically the client reads that straight from the config rather than passing it through the master.

> no, more like the 'arbitrary x-axis' stuff we talked about.

You mean, would we eventually allow arbitrary x-axis definitions?  Right now, we require 'batches' as a monotonically increasing X axis, but theoretically a user could treat that as epochs and only report metrics on epoch boundaries.

Good question.  I think the answer is "yes we should eventually solve this" but I also think the answer is that "this is too baked into the system to solve now, so we should punt for now and eat the cost later".  Basically, we need this monotonically increasing x axis to archive metrics, to associate checkpoints / validations / training metrics when they occur at the same point in training (but at the exact same point in time).

rb-determined-ai:
@hkang1 requesting your for the webui changes.  Based on our previous conversation, I think the only thing is to preserve the state of `max_length` in the continue-trial modal; unitless searcher inputs produce unitless searcher outputs, otherwise the old behavior is preserved.

webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
```diff
@@ -159,7 +160,7 @@ const useModalExperimentCreate = (props?: Props): ModalHooks => {
           </Form.Item>
           {!isFork && (
             <Form.Item
-              label={`Max ${getMaxLengthType(config)}`}
+              label={`Max ${getMaxLengthType(config) ?? 'length'}`}
```
hkang1, line 163: **nit:** for this one `||` might be slightly "better" than `??`. `||` will trigger if the value is an empty string or other non-truthy values such as zero, null and undefined. Where as `??` will only trigger if the value is `null` or `undefined` only.

webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
```diff
@@ -214,7 +215,12 @@ const useModalExperimentCreate = (props?: Props): ModalHooks => {
     }
     if (formValues.maxLength) {
       const maxLengthType = getMaxLengthType(newConfig);
-      newConfig.searcher.max_length = { [maxLengthType]: parseInt(formValues.maxLength) };
+      if (maxLengthType === undefined){
+        // Unitless searcher config.
+        newConfig.searcher.max_length = parseInt(formValues.maxLength);
+      }else{
```
hkang1, line 221: **super nit:** `} else {`
We probably need to add this to the linting rule.

hkang1 (approved):
The WebUI changes look good!

webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
rb-determined-ai: I would argue that the only legal values of "getMaxLengthType" used to be `{"epochs", "records", "batches"}` but has been expanded to `{"epochs", "records", "batches", undefined}`, so both cases are equally correct for handling all valid outputs.  I can make that change anyway if you prefer it though.

stoksc:
@stoksc stop being lazy (this is so i have a notification, so i review next time I do a round)

rb-determined-ai:
ping @stoksc

harness/determined/_core/_searcher.py
```diff
@@ -10,57 +10,38 @@
...
 class Unit(enum.Enum):
-    EPOCHS = "UNIT_EPOCHS"
-    RECORDS = "UNIT_RECORDS"
-    BATCHES = "UNIT_BATCHES"
+    EPOCHS = "EPOCHS"
+    RECORDS = "RECORDS"
+    BATCHES = "BATCHES"
+
+
+def _parse_searcher_units(experiment_config: dict) -> Optional[Unit]:
```
stoksc, line 18: this is all a bit unfortunate, that there isn't nicer tooling to get something so simple

harness/determined/_core/_searcher.py
```diff
@@ -10,57 +10,38 @@
...
+    searcher = experiment_config.get("searcher", {})
+    # All searchers have max_length, except pbt which has a length_per_round.
+    length_example = searcher.get("max_length") or searcher.get("length_per_round")
+    if isinstance(length_example, dict) and len(length_example) == 1:
+        key = next(iter(length_example.keys()))
+        return {"records": Unit.RECORDS, "epochs": Unit.EPOCHS, "batches": Unit.BATCHES}.get(key)
```
stoksc, line 24: lol

master/pkg/schemas/expconf/length.go
```diff
@@ -18,68 +17,50 @@ const (
 	Records     Unit = "records"
 	Batches     Unit = "batches"
 	Epochs      Unit = "epoches"
+	Unitless    Unit = "_unitless"
```
stoksc, line 20:
```suggestion
	Unitless    Unit = "unitless"
```

master/pkg/schemas/expconf/searcher_config.go
```diff
@@ -214,6 +214,7 @@ type AdaptiveConfigV0 struct {
 	RawTrainStragglers *bool         `json:"train_stragglers"`
 	RawMode            *AdaptiveMode `json:"mode"`
 	RawMaxRungs        *int          `json:"max_rungs"`
+
```
stoksc, line 217:
```suggestion
```

master/pkg/searcher/adaptive_asha.go
```diff
@@ -155,6 +155,16 @@ func max(initial int, values ...int) int {
 	return maxValue
 }
 
+func u64max(initial uint64, values ...uint64) uint64 {
```
stoksc, line 158: weird to write this way, because not this calling pattern doesn't work (`u64max(xs...)`) https://go.dev/play/p/V6PS20--qHd if this worked though, that would be neat. i like pattern matching.

stoksc (commented):
lgtm, expect maybe preview search cli needs to be updated? or did i miss it. also, curious, what's the medium to long term gain for this change (other than, the master just doesnt ened units so why does it have it)? i guess, where are we going with this again?

harness/determined/_core/_searcher.py
rb-determined-ai: I don't think it's that simple.  The concept of "which units does the searcher use" isn't configured as `searcher.units`, it's inferred from other searcher properties.  At least IMO, the issue is with the complexity of the config and not with the tooling.

rb-determined-ai:
> except maybe preview search cli needs to be updated?

good call, done.

> what's the medium to long term gain for this change (other than, the master just doesnt ened units so why does it have it)?

I mean, that's basically it.  We're simplifying the configuration of the searcher (which is good) and making the overall Searcher configuration and API more intuitive.  When @liamcli was reviewing the searcher API as it was written, the first thing he said was, "well if the client is responsible for interpreting the units and the client is responsible for configuring the units, why does the master even care about the units?"  And I think he was right.  This is a more "core"-like behavior for the searcher component to the Core API.

stoksc (approved):
lgtm (of course with tests and all)

PR #3351, opened 2021-12-20
fix: Convert all experiment and job states to labels

mapmeld:
## Description

Raw values I believe came from JobState (STATE_QUEUED, STATE_SCHEDULED, and STOPPING_CANCELED) were displayed on the experiment details header, instead of human format (Queued, Scheduled, Canceling). This patch uses the unified `stateToLabel` to guarantee whatever relevant state object is coming through, it's decoded for human user.

## Checklist


hkang1 (approved):
LGTM! thanks for fixing this

PR #3359, opened 2021-12-21
chore: fix small data race found by --race

stoksc:
## Description
This change just moves some initialization code around to clean up a data race found by `go run --race`. The race itself was that `grpcutil.RegisterHTTPProxy` tries to use the gRPC logger in a goroutine it spawns but that isn't initialized until `grpcutil.NewGRPCServer` runs, so there is a race writing/reading the global logger set by `grpclogrus.ReplaceGrpcLogger`.

## Test Plan
- [x] `--race` should no longer report this

## Checklist

### DIFF
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -403,7 +403,7 @@ func (m *Master) startServers(ctx context.Context, cert *tls.Certificate) error 
 		})
 	}
 
-	// This must be before grpcutil.RegisterHTTPProxy is called since it may use stuff setup by the
+	// This must be before grpcutil.RegisterHTTPProxy is called since it may use stuff set up by the
 	// gRPC server (logger initialization, maybe more). Found by --race.
 	gRPCServer := grpcutil.NewGRPCServer(m.db, &apiServer{m: m},
 		m.config.Observability.EnablePrometheus,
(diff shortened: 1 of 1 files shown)


master/internal/core.go
```diff
@@ -429,14 +435,11 @@ func (m *Master) startServers(ctx context.Context, cert *tls.Certificate) error
 		}()
 	}
 	start("gRPC server", func() error {
```
stoksc, line 437: any reason we don't start the cmux mux first?

dzhu: The order doesn't really matter, since they're all in goroutines anyway, right?

master/internal/core.go
```diff
@@ -403,12 +403,18 @@ func (m *Master) startServers(ctx context.Context, cert *tls.Certificate) error
 		})
 	}
 
-	// Initialize listeners and multiplexing.
+	// This must be before grpcutil.RegisterHTTPProxy is called since it may use stuff setup by the
```
dzhu, line 406:
```suggestion
	// This must be before grpcutil.RegisterHTTPProxy is called since it may use stuff set up by the
```

dzhu (approved):

master/internal/core.go
stoksc: Indeed

PR #3360, opened 2021-12-21
feat: Add /tasks/:task_id endpoint to GRPC API [DET-6354] [DET-6355]

mapmeld:
## Description

Adds an endpoint to GET status / readiness of a specific task. This approach adds columns to the allocations table, and an UPDATE query when container state changes.

An alternative would be to combine `/api/v1/tensorboards` and `/notebooks` calls, or a similar call to `a.ask(tensorboardsAddr.Child(req.TensorboardId)` format.

## Checklist

### DIFF
--- a/master/internal/api_task.go
+++ b/master/internal/api_task.go
@@ -1,0 +1,27 @@
+package internal
+
+import (
+	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/internal/db"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
+	"github.com/determined-ai/determined/proto/pkg/taskv1"
+)
+
+func (a *apiServer) GetTask(
+	_ context.Context, req *apiv1.GetTaskRequest) (resp *apiv1.GetTaskResponse, err error) {
+	t := &taskv1.Task{}
+	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); err {
+	case db.ErrNotFound:
+		return nil, status.Errorf(
+			codes.NotFound, "task %s not found", req.TaskId)
+	default:
+		return &apiv1.GetTaskResponse{Task: t},
+			errors.Wrapf(err, "error fetching task %s from database", req.TaskId)
+	}
+}
--- a/master/internal/api_task.go
+++ b/master/internal/api_task.go
@@ -16,8 +16,8 @@ import (
 func (a *apiServer) GetTask(
 	_ context.Context, req *apiv1.GetTaskRequest) (resp *apiv1.GetTaskResponse, err error) {
 	t := &taskv1.Task{}
-	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); err {
-	case db.ErrNotFound:
+	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); {
+	case errors.Is(err, db.ErrNotFound):
 		return nil, status.Errorf(
 			codes.NotFound, "task %s not found", req.TaskId)
 	default:
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -121,7 +121,7 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 }
 
 // UpdateAllocationState stores the latest task state and readiness.
-func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
+func (db *PgDB) UpdateAllocationState(a *model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
 		SET state=$2, is_ready=$3
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -121,13 +121,13 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 }
 
 // UpdateAllocationState stores the latest task state and readiness.
-func (db *PgDB) UpdateAllocationState(a model.Allocation) (model.AllocationState, error) {
+func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
 		SET state=$2, is_ready=$3
 		WHERE allocation_id=$1
 	`, a.AllocationID, a.State, a.IsReady)
-	return a.State, err
+	return err
 }
 
 // CloseOpenAllocations finds all allocations that were open when the master crashed
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -122,12 +122,12 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 
 // UpdateAllocationState stores the latest task state and readiness.
 func (db *PgDB) UpdateAllocationState(allocationID model.AllocationID,
-	state model.AllocationState, readiness bool) error {
+	state model.AllocationState, isReady bool) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
-		SET state=$2, readiness=$3
+		SET state=$2, isReady=$3
 		WHERE allocation_id=$1
-	`, allocationID, state, readiness)
+	`, allocationID, state, isReady)
 	return err
 }
 
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -125,7 +125,7 @@ func (db *PgDB) UpdateAllocationState(allocationID model.AllocationID,
 	state model.AllocationState, isReady bool) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
-		SET state=$2, isReady=$3
+		SET state=$2, is_ready=$3
 		WHERE allocation_id=$1
 	`, allocationID, state, isReady)
 	return err
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -121,7 +121,7 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 }
 
 // UpdateAllocationState stores the latest task state and readiness.
-func (db *PgDB) UpdateAllocationState(a *model.Allocation) error {
+func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
 		SET state=$2, is_ready=$3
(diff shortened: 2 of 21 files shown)


master/internal/api_task.go
```diff
@@ -0,0 +1,27 @@
...
+func (a *apiServer) GetTask(
+	_ context.Context, req *apiv1.GetTaskRequest) (resp *apiv1.GetTaskResponse, err error) {
+	t := &taskv1.Task{}
+	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); err {
+	case db.ErrNotFound:
+		return nil, status.Errorf(
+			codes.NotFound, "task %s not found", req.TaskId)
+	default:
+		return &apiv1.GetTaskResponse{Task: t},
+			errors.Wrapf(err, "error fetching task %s from database", req.TaskId)
```
hkang1, line 25: **question:** I looked at other endpoint examples that do this, but figure you can just tell me. If `err` ends being something other than `db.ErrNotFound` (such as `db. ErrUnsupported`), it'll reach the `default` case and then would it return  `errors.Wrapf(...)` instead of the `{ Task: t }` response?

master/internal/task/allocation_test.go
```diff
@@ -100,6 +100,18 @@ func TestAllocation(t *testing.T) {
...
+				db.On("UpdateAllocationState", a.req.AllocationID, model.AllocationStatePulling,
+					false).Return(nil)
+				db.On("UpdateAllocationState", a.req.AllocationID, model.AllocationStateStarting,
+					false).Return(nil)
+				db.On("UpdateAllocationState", a.req.AllocationID,
+					model.AllocationStateRunning, false).Return(nil)
+				db.On("UpdateAllocationState", a.req.AllocationID, model.AllocationStateTerminating,
+					false).Return(nil)
+				db.On("UpdateAllocationState", a.req.AllocationID, model.AllocationStateTerminated,
+					false).Return(nil)
```
hkang1, line 114: **question:** would it be valuable to have some tests that tests for db failures?

master/internal/task/allocation.go
```diff
@@ -194,6 +194,7 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 						IsReady:           a.logBasedReadinessPassed,
 						ServiceReadyEvent: &msg,
 					})
+					_ = a.db.UpdateAllocationState(a.req.AllocationID, a.state, a.logBasedReadinessPassed)
```
hkang1, line 197: 👍

master/internal/task/allocation.go
```diff
@@ -444,6 +445,8 @@ func (a *Allocation) TaskContainerStateChanged(
 			prom.RemoveAllocationReservation(a.reservations[cID].Summary())
 		}
 	}
+
+	_ = a.db.UpdateAllocationState(a.req.AllocationID, a.state, a.logBasedReadinessPassed)
```
hkang1, line 449: 👍

hkang1 (approved):
LGTM! I'd feel more comfortable if someone from Infra take a look at the go code a bit more in depth than me. I have a few minor non-blocking things sprinkled throughout.

proto/src/determined/task/v1/task.proto
```diff
@@ -23,3 +23,13 @@ enum State {
...
 }
+
+// A record of Jupyter or TensorBoard.
+message Task {
+  // Unique ID of task.
+  string task_id = 1;
+  // The current state of the task.
+  State state = 2;
+  // Whether the TensorBoard or Notebook is ready to access.
+  bool readiness = 3;
```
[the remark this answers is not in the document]
mapmeld: updated to is_ready throughout the code

proto/src/determined/api/v1/api.proto
```diff
@@ -952,6 +953,16 @@ service Determined {
     };
   }
 
+  // Check the status of a requested task.
+  rpc GetTask(GetTaskRequest) returns (GetTaskResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/tasks/{task_id}"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "Notebooks"
```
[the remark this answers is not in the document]
mapmeld: We would need to work this out later. I was able to change the tag and move things around in `api-ts-sdk/api.ts` but the binding generation script will overwrite it with the previous config.

master/internal/api_task.go
mapmeld: yes that sounds right, it's based on GetModel and similar endpoints

master/internal/api_task.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/determined-ai/determined/master/internal/db"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
+	"github.com/determined-ai/determined/proto/pkg/taskv1"
+)
+
+func (a *apiServer) GetTask(
+	_ context.Context, req *apiv1.GetTaskRequest) (resp *apiv1.GetTaskResponse, err error) {
+	t := &taskv1.Task{}
+	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); err {
+	case db.ErrNotFound:
```
stoksc, line 20: probably safer, in case someone wraps it one day
```suggestion
	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); {
	case errors.Is(err, db.ErrNotFound):
```

stoksc, line 20: probably safer, in case someone wraps it one day
```suggestion
	switch err := a.m.db.QueryProto("get_task", t, req.TaskId); {
	case errors.Is(err, db.ErrNotFound):
```

master/internal/task/allocation.go
stoksc, line 197: have to handle the error here.. at least log it, but with the way this is, if it is failed we are in trouble since the webui will never know to follow the proxy. also i'm not sure is_ready should be a bit in the db, since it only has meaning when the allocation is in memory/active. i think it probably makes more sense for us to just `ctx.Ask()` the actor if it is ready. but let me think about it some.

master/internal/task/allocation_test.go
stoksc: probably, this just mocks out the db and assumes everything goes ok. having a test that systematically fails each db call and checks the outcome would be great.

master/static/migrations/20211221113503_add-allocation-state.down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE public.allocations
+    DROP COLUMN state;
+
+ALTER TABLE public.allocations
+    DROP COLUMN isReady;
```
stoksc, line 5: if we go this route, i think `is_ready` is in line with all the rest of the columns in the db.

master/static/migrations/20211221113503_add-allocation-state.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE public.allocations
+    ADD COLUMN state INT;
+
+ALTER TABLE public.allocations
+    ADD COLUMN isReady BOOLEAN;
```
stoksc, line 5:
```suggestion
ALTER TABLE public.allocations
    ADD COLUMN state INT,
    ADD COLUMN isReady BOOLEAN;
```

proto/src/determined/task/v1/task.proto
```diff
@@ -23,3 +23,13 @@ enum State {
   // The task has begun to exit.
   STATE_TERMINATING = 7;
 }
+
+// A record of Jupyter or TensorBoard.
```
stoksc, line 27: A task is a lot more than a record of jupyter or tensorboard, and this API should probably reflect that. I think there is a lot more state we should add (we should at least return everything that is in the DB).

master/internal/task/allocation.go
stoksc, line 449: ditto, have to handle the error here, also i think it makes sense to make is_ready NULL in the db and just null it out when we're done (like we do with progress, since it's meaningless once completed)

master/internal/task/allocation.go
stoksc: let's just chat over zoom, i think this is probably fine and it doesn't rely on the actor system. but i want to at least think out loud through the other way to write it.

master/internal/task/allocation_test.go
```diff
@@ -100,6 +100,18 @@ func TestAllocation(t *testing.T) {
 						Devices: []device.Device{},
 					},
 				}
+				db.On("UpdateAllocationState", a.req.AllocationID, model.AllocationStateAssigned,
```
stoksc, line 103: can we weave these between the asks happening and make sure they get called correctly?

master/static/migrations/20211221113503_add-allocation-state.down.sql
stoksc: bump

master/internal/task/allocation.go
```diff
@@ -444,6 +448,11 @@ func (a *Allocation) TaskContainerStateChanged(
 			prom.RemoveAllocationReservation(a.reservations[cID].Summary())
 		}
 	}
+
+	err := a.db.UpdateAllocationState(a.req.AllocationID, a.state, a.logBasedReadinessPassed)
```
stoksc, line 452: i think this would can just be a `ctx.Log().Error()` and the one above it should `a.Error()` and return. wdyt?

master/internal/db/postgres_tasks.go
```diff
@@ -120,6 +120,17 @@ func (db *PgDB) DeleteAllocationSession(allocationID model.AllocationID) error {
 	return err
 }
 
+// UpdateAllocationState stores the latest task state and readiness.
+func (db *PgDB) UpdateAllocationState(allocationID model.AllocationID,
```
stoksc, line 124: after talking yesterday, I think these fields should be added to the `model.Allocation` struct, updated in the copy of it nested in the allocation and it should be received here.

mapmeld: so would this replace `UpdateAllocationState` with `UpdateAllocation`?

master/internal/task/allocation.go
```diff
@@ -197,7 +197,7 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 					})
 					a.model.State = a.state
 					a.model.IsReady = a.logBasedReadinessPassed
-					err := a.db.UpdateAllocationState(&a.model)
+					err := a.db.UpdateAllocationState(a.model)
 					if err != nil {
```
stoksc, line 201: my preference, no pressure
```suggestion
					if err := a.db.UpdateAllocationState(a.model); err != nil {
```

master/internal/api_task.go
stoksc: yeah, personally i dont love this, but it works. it relies on errors.Wrap returning nil if the err is nil and it being ok to return a response and err in the err case

proto/src/determined/task/v1/task.proto
```diff
@@ -23,3 +23,25 @@ enum State {
   // The task has begun to exit.
   STATE_TERMINATING = 7;
 }
+
+// Allocation tracks a specific instance of a Task.
+message Allocation {
+  // Unique ID of task associated with the allocation.
+  string task_id = 1;
```
stoksc, line 30: seems repetitive to have task_id here, but maybe allocation id _should_ be here

PR #3361, opened 2021-12-21
chore: write our own swagger bindings

### LINKED ISSUE #2835: build: refine where we generate and persist API bindings
https://determined-community.slack.com/archives/C025A98H82X/p1628812241031100

hamidzr:
## Description

- generate python bindings
- move where generated typescript bindings are checked

## Commentary (optional)

Taking over from @rb-determined-ai. Original PR [here](https://github.com/determined-ai/determined/pull/3156)

@rb-determined-ai let me know if you rather we use the [original PR ](https://github.com/determined-ai/determined/pull/3156) instead. 

sample usage.
```
from determined.common.api.authentication import Authentication
from determined.common.api.b import get_GetExperiments
from determined.common.experimental import session

user = "determined"
password = ""
auth = Authentication(None, requested_user=user, password=password, cert=None)
s = session.Session(None, user=user, auth=auth, cert=None)

rv = get_GetExperiments(s._do_request)
print(rv.to_json())
```

fixes #2835 
closes #3032 

## Checklist

- [x] typing as t and bindings as b ?


TODO.md
```diff
@@ -0,0 +1,7 @@
+# TODO
```
hamidzr, line 1: remove the file

bindings/Makefile
```diff
@@ -1,5 +1,7 @@
 SWAGGER_SPEC := ../proto/build/swagger/determined/api/v1/api.swagger.json
 CODEGEN_BIN := swagger-codegen-cli-2.4.14.jar
+py_bindings_dest := ../harness/determined/common/api/b.py
```
ioga, line 3: can we have a more distinct name than `b.py`? `bindings.py`? `sdk_bindings.py`? `api_sdk.py`?

bindings/bindings.py
```diff
@@ -0,0 +1,593 @@
+import abc
```
ioga, line 1: can this file be called `generate_bindings.py` or something along these lines to indicate this is the generator?

bindings/bindings.py
```diff
@@ -0,0 +1,593 @@
...
+    for ref in Ref.all_refs:
+        ref.linked = True
+        ref.defn = defs[ref.name]
+
+
+def pybindings(swagger: dict) -> str:
+    prefix = """
+import enum
+import math
+import typing as t
```
ioga, line 539: we don't really do such alias anywhere. maybe just use it as `typing`?

mapmeld (commented):
:+1: can confirm that this uses the generated Python bindings to call `get_GetMaster`.  Here are two possible improvements

- Do we have a reference doc for the command webui should use to regenerate the bindings, i.e. replacing `make -C webui/react build-bindings` with `make -C bindings build/typescript-fetch` and the new `make -C bindings build/python`

- We don't use `echo` in the other Makefiles, it seems maybe because we see both the input command and the output; could this be moved to a log / print in bindings.py?

```
echo "Generated ../harness/determined/common/api/b.py"
Generated ../harness/determined/common/api/b.py
```

harness/determined/cli/master.py
```diff
@@ -22,6 +24,17 @@ def config(args: Namespace) -> None:
         raise ValueError(f"Bad output format: {args.output}")
 
 
+def get_master(args: Namespace) -> None:
+    session = setup_session(args)
+    resp = get_GetMaster(session._do_request)
```
ioga, line 29: passing session itself may look more natural (boto at the call site
```suggestion
    session = setup_session(args)
    resp = get_GetMaster(session)
```

a class instance method style like `session.get_GetMaster` would probably be even more in line with things like boto3.

bindings/bindings.py
```diff
@@ -0,0 +1,593 @@
...
+        out += ["        None,"]
+        out += ["        None,"]
+        out += ["    )"]
+        for expect, returntype in responses.items():
+            out += [f"    if _req.status_code == {expect}:"]
+            if returntype.isnone():
+                out += ['        return']
+            else:
+                out += [f'        return {returntype.load("_req.json()")}']
+        # XXX: raise something else here
```
ioga, line 404: maybe add a small custom `class APIHttpError(Exception):` in the prefix, and throw it here?
It'd also be very helpful for debugging and logging to include the entire response object on the exception, and not just the status code

ioga (commented):
haven't tried it yet, but here's a batch of comments.

harness/determined/cli/master.py
hamidzr: The idea here was to leave the (experimental) determined package out of the generated code. This way you can pass in any request method that supports the same parameters and use the API instead of having to have determined installed.
Although if we're strictly using it as the examples here show `session` works as well

bindings/Makefile
hamidzr: the only benefit with b would be to make the name shorter when importing assuming it's usage scope is gonna be very high and everyone would know what it is. 

this remind me we should add a comment in the generated file to warn people from editing it.

bindings/bindings.py
hamidzr: I only did this since it's repeated so many times in the outputted code. does the file size affect how fast the interpreter goes through it?

harness/determined/cli/master.py
ioga: maybe we could use something like that to have typing for the `session` parameter while avoiding imports: https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols

ioga: >This way you can pass in any request method that supports the same parameters and use the API instead of having to have determined installed.

that makes sense as a feature; however passing around an underscored property `session._do_request`  seems too weird for me in our code.

hamidzr:
> Do we have a reference doc for the command webui should use to regenerate the bindings, i.e. replacing make -C webui/react build-bindings with make -C bindings build/typescript-fetch and the new make -C bindings build/python

this should already be triggered when users build determined or the bindings module directly. The only other times these need to be generated by developers is if they change the api, or the generator in which case a simple `make -C bindings build` is probably required. What are you thinking in terms of references? a read me in `bindings` module?

hamidzr:
updated sample usage

```
from determined.common.api.bindings import get_GetExperiments, get_GetExperiment
from determined.experimental.client import http_request

rv = get_GetExperiments(http_request)
# rv = get_GetExperiment(s._do_request, experimentId=23423)
print(rv.to_json())
```

bindings/bindings.py
hamidzr: sounds good this would still be dependent on the passed in http request to not raise any expectations (or do if that's what the user of the bindings wants) unless we opt to wrap and catch those as well

harness/determined/cli/master.py
hamidzr: > maybe we could use something like that to have typing for the session parameter while avoiding imports: https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols

it'd still mean that the user of the bindings would have to pass in a class that conforms to what a determined `session` looks like. how does this look? 🤔

mapmeld:
This makes sense.  I was able to figure out how remote instances work
```python
from determined.experimental.client import http_request, login
login(master=latest_master, user=user, password=password)
...
rv = get_GetExperiment(http_request, experimentId=2185)
```
It is not super-intuitive that the Determined object / singleton is remembering the connection info after `login()` for future use of the imported `http_request`; would it make sense to return callable `http_request` from login(), and/or to rename things to something like
```python
api_session = APISession() # replacing login(), defaulting to localhost
rv = get_GetExperiment(api_session, experimentId=2185)
```

bindings/bindings.py
rb-determined-ai: Not particularly, as far as I know.  You could time it.

harness/determined/cli/master.py
rb-determined-ai: I don't think we're here to write a swagger bindings generator that works for anybody but us.  That will keep this code simple and keep its surface area small.  So requiring determined to be installed makes perfect sense.

rb-determined-ai:
> would it make sense to return callable http_request from login(),

@mapmeld Just use a `Determined()` object explicitly, which takes the same args as `login`, if that is what you want.

harness/determined/cli/master.py
hamidzr: hmm true but if there is no real cost to allowing the generated bindings to be used without the determined package being installed why not do it? It'd still allow people interested in working with determined to use the bindings without depending on another package. Does it look off without passing the session? Is there a way I could set it up so that it doesn't look off?

hamidzr: > a class instance method style like session.get_GetMaster would probably be even more in line with things like boto3.

a class instance would be a different style and we can certainly consider that as well.

rb-determined-ai: > a class instance method style like session.get_GetMaster would probably be even more in line with things like boto3

I don't think "being like boto3" is actually desirable.  That library (and the botocore library behind it) accomplish very little in an absolutely enormous volume of code.  Subclassing parts of it to get custom behavior is very difficult.  See our boto3_credential_manager.py for example.> It'd still allow people interested in working with determined to use the bindings without depending on another package.

No, the swagger spec itself is what you give to these people.  If they want to use Determined without using the python package, then they should generate their own bindings from the swagger spec.  That's like the stated goal of publishing the swagger spec.

> how does this look

Also, where are these people?  I've never heard of a single user trying to use the REST API directly.  I've heard of a lot of people who want to see good python bindings.

The introduction of `client.http_request` means that in practice the bindings still depend on a Session, only instead of being an explicit Session as before, now it's an implicit singleton Session.  Also, it's the wrong Session; there are already singletons for the cli's master certs and session tokens in the code, but using the user-facing Session embedded in the client module will cause bugs.

> hmm true but if there is no real cost to allowing the generated bindings to be used without the determined package being installed why not do it?

I would argue there is a real cost to what you have.  It's another layer of indirection in how our own code works.  The Session object wraps the requests library in the cert handling and token handling that we need.  Having another function that wraps the Session object and has an API kinda like the requests library is another layer of indirection.  Our developers have to read and understand what's going on, so extra layers of indirection have a very real cost, especially for onboarding new engineers.

In 


I think it's important to remember that we are adding swagger bindings for the type-checking behavior that they give us.  They are not particularly suitable for giving to a user (a function named `get_GetMaster()`... that would never pass code review for a user-api) because they're just clunky.  Everywhere that we want to have a user-facing Python API  
- We want type checking

And why we are not adding swagger bindings:
-

rb-determined-ai: > It'd still allow people interested in working with determined to use the bindings without depending on another package.

No, the swagger spec itself is what you give to these people.  If they want to use Determined without using the python package, then they should generate their own bindings from the swagger spec.  That's like the stated goal of publishing the swagger spec.

Also, where are these people?  I've never heard of a single user trying to use the REST API directly.  I've heard of a lot of people who want to see good python bindings.

> how does this look

The introduction of `client.http_request` means that in practice the bindings still depend on a Session, only instead of being an explicit Session as before, now it's an implicit singleton Session.  This is worse because it is harder to understand.  Also, it's the wrong Session; there are already singletons for the cli's master certs and session tokens in the code, but using the user-facing Session embedded in the client module will cause bugs.

> hmm true but if there is no real cost to allowing the generated bindings to be used without the determined package being installed why not do it?

I would argue there is a real cost to what you have.  It's another layer of indirection in how our own code works.  The Session object wraps the requests library in the cert handling and token handling that we need.  Having another function that wraps the Session object and has an API kinda like the requests library is another layer of indirection.  Our developers have to read and understand what's going on, so extra layers of indirection have a very real cost, especially for onboarding new engineers.

Ultimately, I'm arguing for the absolute simplest solution that gives us type-checking.  Early generalization is a form of early optimization, and I think it just makes our lives harder without making any user lives' easier.

rb-determined-ai: > a class instance would be a different style and we can certainly consider that as well.

This I could be ok with, but if you do go this route please make that giant generated class separate from the Session itself.  Make it some SessionContainingObject that takes a Session as input (i.e. separate the generated code from the hand-written code).

Personally, I still think that all of these generated APIs are always clunky.  You can never give them to end users directly because names like `get_GetMaster()` are absolutely terrible names.  If you hand-wrote a function like that and tried to put it in a user-facing API it would never, never make it through code-review.  Therefore, since these generated APIs are too clunky to give to users, we'll always end up writing wrappers around them.  This is basically what the Deterimined() object is; a bunch of user-friendly wrappers around requests against the REST APIs.  So when we write those wrappers, wrapping a function that takes a Session vs wrapping an object method that was already given a Session seems basically equally clunky from the determined developer's perspective, and I wouldn't bother.

harness/determined/cli/master.py
```diff
@@ -7,8 +7,10 @@
 
 from determined.common import api, yaml
 from determined.common.api import authentication
+from determined.common.api.bindings import get_GetMaster
```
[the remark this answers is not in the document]
hamidzr: makes sense. in these cases I prefer to keep the name shorter. wdy think about `import bindings as b`. or do you think it's not worth the readability hit

bindings/generate_bindings.py
```diff
@@ -0,0 +1,600 @@
...
+                out += [f'        return {returntype.load("_resp.json()")}']
+        out += [f'    raise APIHttpError("{self.method}_{self.name}", _resp)']
+
+        return "\n".join(out)
+
+
+def classify_type(path: str, schema: dict) -> TypeAnno:
+    # enforce valid jsonschema:
+    assert isinstance(schema, dict), (path, schema)
+    # XXX: how to handle inlined enums?
```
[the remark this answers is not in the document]
hamidzr: this is talking about the following situation where some parameters are defined inline as the following:

```
// params:...
            {
              "name": "orderBy",
              "description": "Order commands in either ascending or descending order.\n\n - ORDER_BY_UNSPECIFIED: Returns records in no specific order.\n - ORD
  ER_BY_ASC: Returns records in ascending order.\n - ORDER_BY_DESC: Returns records in descending order.",
              "in": "query",
              "required": false,
              "type": "string",
*             "enum": [
                "ORDER_BY_UNSPECIFIED",
                "ORDER_BY_ASC",
                "ORDER_BY_DESC"
              ],
              "default": "ORDER_BY_UNSPECIFIED"
            },
```
instead of being a reference to the existing enum:

```
// definitions..
*     "v1OrderBy": {
        "type": "string",
        "enum": [
          "ORDER_BY_UNSPECIFIED",
          "ORDER_BY_ASC",
          "ORDER_BY_DESC"
        ],
        "default": "ORDER_BY_UNSPECIFIED",
        "description": "Order records in either ascending or descending order.\n\n - ORDER_BY_UNSPECIFIED: Returns records in no specific order.\n - ORDER_BY_A  SC: Returns records in ascending order.\n - ORDER_BY_DESC: Returns records in descending order.",
*       "title": "OrderBy"
      },
```

Not sure if this is a result of how we generated the openapi spec or if this is standard. there is a mention of reusable enums in OpenAPI3
which might be what we're looking for
https://swagger.io/docs/specification/data-models/enums/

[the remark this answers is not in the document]
hamidzr: without direct a reference from the swagger spec, we could attempt to link them ourselves. what else could we do? 🤔

harness/determined/cli/master.py
```diff
@@ -65,6 +77,10 @@ def process_response(response: Response, latest_log_id: int) -> int:
             Arg("-o", "--output", type=str, default="yaml",
                 help="Output format, one of json|yaml")
         ]),
+        Cmd("info", get_master, "fetch master info", [
+            Arg("-o", "--output", type=str, default="yaml",
```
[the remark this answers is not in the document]
hamidzr: would you say a two-letter flag with a single is `-` also uncommon?

bindings/Makefile
```diff
@@ -13,14 +15,39 @@ deps/$(CODEGEN_BIN):
...
-build/typescript-fetch: $(SWAGGER_SPEC) deps/$(CODEGEN_BIN)
-	mkdir -p build/typescript-fetch
+.PHONY: bindings-copy-over
+bindings-copy-over:
+	rm -rf $(ts_bindings_dest)
+	cp -r  build/typescript-fetch $(ts_bindings_dest)
+
+.PHONY: build/python
+build/python: $(SWAGGER_SPEC)
+	python generate_bindings.py > $(py_bindings_dest)
```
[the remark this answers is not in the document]
hamidzr: thats a good tip. I started with a normal target but debuggin CI without "Run with ssh" was taking too long maybe I should just add a target like this 

```
.PHONY: force-gen
force-gen:
	touch ../schemas/gen.py
```

harness/determined/cli/master.py
hamidzr: I'd say the generator itself is added value if anyone wanted to use the rest apis in python. As opposed to asking them to generate their own since if there already was a way of generating the bindings that's as fast as we wanted it we wouldn't be writing our own generator. Now if we don't consider that use case at all then the point that `client.http_request` would make our code somewhat harder to read IMO is enough to have it take a `determined session` instead.

bindings/bindings.py
hamidzr: added a todo to do this before landing.

harness/determined/cli/master.py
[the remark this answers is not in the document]
rb-determined-ai: yes

harness/determined/cli/master.py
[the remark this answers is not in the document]
rb-determined-ai: I think that short names are nice, especially when all you're working on is bindings.  But when somebody else comes along who is not focused on bindings, it won't be nearly as obvious to them what `b` is or why it needed to be so short.  Basically, I think sticking to the style guide usually has a lot of small opportunity costs, but the overall benefit of standardization in the code is well worth it.

bindings/generate_bindings.py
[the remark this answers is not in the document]
rb-determined-ai: Yeah I like how you did it there.

mapmeld (approved):
LGTM with notes - I think it'd be cleaner to generate the code in a Jinja template or library instead of adding strings and remembering indentation, but there's a lot in this PR as is, so I'd support getting this merged and leaving it to another PR.

bindings/generate_bindings.py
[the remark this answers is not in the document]
hamidzr: added a todo for the streaming portion. and addressed the one about asserting no `.` outside curly braces in `path`

harness/determined/cli/master.py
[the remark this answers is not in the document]
hamidzr: we seem to be using `--output` elsewhere in the cli as well but not for `--format`. would it be a breaking change? if so would it be better to tackle that in a separate pr

bindings/generate_bindings.py
```diff
@@ -0,0 +1,653 @@
...
+        out += [f'    raise APIHttpError("{self.method}_{self.name}", _resp)']
+
+        return "\n".join(out)
+
+
+def classify_type(enums: dict, path: str, schema: dict) -> TypeAnno:
+    # enforce valid jsonschema:
+    assert isinstance(schema, dict), (path, schema)
+    if "enum" in schema:
+        name = enums[hash(json.dumps(schema["enum"]))]
```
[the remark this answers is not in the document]
hamidzr: >  only the dictionary is able to resolve conflicts.

which conflicts are you thinking about? the possibility that the hash of two different (serialized) enums is the same? or how the keys are mapped to the underlying buckets for a dictionary

bindings/generate_bindings.py
```diff
@@ -0,0 +1,653 @@
...
+
+def link_all_refs(defs: TypeDefs) -> None:
+    for ref in Ref.all_refs:
+        ref.linked = True
+        ref.defn = defs[ref.name]
+
+
+def pybindings(swagger: dict) -> str:
+    prefix = """
+# The contents of this file are programatically generated.
```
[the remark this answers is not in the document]
hamidzr: that's a cool module.

bindings/generate_bindings.py
```diff
@@ -0,0 +1,653 @@
...
+        self.serialized_name = serialized_name
+        self.type = typ
+        self.required = required
+        self.where = where
+        # validations
+        assert where in ("query", "body", "path", "definitions"), (name, where)
+        assert where != "path" or required, name
+        if where == "path":
+            if not isinstance(typ, (String, Int, Bool)):
+                raise AssertionError(f"bad type in path parameter {name}: {typ}")
```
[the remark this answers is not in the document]
hamidzr: this was removed in `80d161fafb2b264ebb1d5331a341cb2f9eff4152`
```
* -        if where == "query":
  -            if not isinstance(typ, (String, Int, Bool)):
  -                if not isinstance(typ, Sequence) or not isinstance(typ.items, (String, Int, Bool)):
  -                    raise AssertionError(f"bad type in query parameter {name}: {typ}")
```
when I linked the enums I wanted to allow enums to be passed in as query params and didn't have a good way of knowing that a Ref is referencing an Enum before linking.  in `7d09bdd014943d80d89f92cfa337dd88b138dd04` I attach that piece of information to the Ref.

bindings/generate_bindings.py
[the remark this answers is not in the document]
hamidzr: basically allowing the hash collision resolution strategy of python's builtin dict which I don't know what it is but it has some strategy right?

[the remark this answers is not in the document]
rb-determined-ai: > for my understanding which conflict are you thinking about? the possibility that the hash of two different (serialized) enums is the same? or how the keys are mapped to the underlying buckets for a dictionary

Suppose that `hash()` has a conflict where `{1,2}` and `{3,4}` both hash to the same number, say 7.

When you store `mydict[hash({1,2})] = 1` and then you store `mydict[hash({3,4})]` = 2, then you'll have a map with only one value.  If you store `mydict[{1,2}) = 1` and then `mydict[{3,4}] = 2` then you'll have a map with two values.  There's a hash collision either way, but if you let the dictionary call `hash()` (as it will under the hood) it will know that even though the two entries would normally go into the same bucket they are not the same.  When you call `hash()` yourself and use that as the key, then the dictionary will just see two identical keys and they'll overwrite each other.

The chances that you have a collision with `hash()` is really low but it's also pointless to write extra code which introduces a possible bug.

bindings/generate_bindings.py
```diff
@@ -0,0 +1,684 @@
...
+
+    def gen_def(self) -> Code:
+        out = [f"class {self.name}:"]
+        out += ["    def __init__("]
+        out += ["        self,"]
+        required = sorted(p for p in self.params if self.params[p].required)
+        optional = sorted(p for p in self.params if not self.params[p].required)
+        for name in required + optional:
+            out += ["    " + self.params[name].gen_function_param()]
+        # out += [f"        {k}: {v.annotation()}," for k, v in self.members.items()]
```
rb-determined-ai, line 275: delete this line?

bindings/generate_bindings.py
```diff
@@ -0,0 +1,684 @@
...
+                defs[name] = None
+                continue
+        raise ValueError(f"unhandled schema: {schema} @ {path}")
+    return defs
+
+
+def is_expected_path(text: str) -> bool:
+    """
+    check if there are any dots appearing outside of curly braces if any.
+    assuming there are no nested curly braces.
```
rb-determined-ai, line 528: nit: the docstring is good but let's make it proper grammar (complete sentences and capitalizations).

harness/determined/cli/master.py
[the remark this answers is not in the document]
rb-determined-ai: Only in one command.  I'd argue that command is wrong, and copying it doesn't make sense because far more cli commands use `--json` for json formatting.  If you have to copy a style I'd prefer to copy that one.

Also I realize now that `-f` is frequently used in our cli for `--follow`, so actually I think `--json` is a fine idea here.

But, yeah our cli options are a total mess

rb-determined-ai (approved):
Content looks very good.  Great work, thank you for picking this up!

bindings/Makefile
```diff
@@ -13,14 +16,46 @@ deps/$(CODEGEN_BIN):
...
+
+.PHONY: check
+check: check/python check/typescript-fetch
+	# Checking that the (committed) generated code is up-to-date by ensuring that
+	# git reports the files as unchanged after forcibly regenerating them.
+	# WARNING this has a dependency on proto module being built recently
+	# which isn't linked here.
+
+.PHONY: check-python
+check/python: build/python
```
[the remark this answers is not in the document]
hamidzr: 👍🏼  adding force-gen to the main `check` target should ensure this for other generated bindings as well. I'll go with that

hamidzr:
I updated the command arguments added in this pr. we can update `-rp` and `--output` for the existing ones separately as they are breaking changes.

harness/determined/cli/master.py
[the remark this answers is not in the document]
hamidzr: I [updated](https://github.com/determined-ai/determined/pull/3361/commits/7f0d3de48b02c21c92bfcb1d44acd7afe6d76cfe) the command arguments added in this pr. we can update `-rp` and `--output` for the existing ones separately as they are breaking changes.

hamidzr:
thanks for the reviews

PR #3366, opened 2021-12-21
feat: add systemd socket activation support to the master

dzhu:
## Description

This change allows the master to be run using systemd's socket
activation feature. It simply changes the creation of the base socket
listener to check for one provided by systemd or fall back to a manually
created listener; everything else built on that stays the same.

We also ship a new socket unit file so that users can switch to socket
activation easily with a few `systemctl` invocations.

## Test Plan

- [x] spin up a VM and go through installing from the package and trying out the various documented `systemctl` commands

### DIFF
--- a/docs/sysadmin-deploy-on-prem/linux-packages.txt
+++ b/docs/sysadmin-deploy-on-prem/linux-packages.txt
@@ -194,6 +194,50 @@ Master and Agent
    the right-hand side of the top navigation bar, and if you select the ``Cluster`` view in the
    left-hand navigation panel.
 
+.. _socket-activation:
+
+Socket Activation
+=================
+
+The master can be configured to use `systemd socket activation
+<https://0pointer.de/blog/projects/socket-activation.html>`__, allowing it to be started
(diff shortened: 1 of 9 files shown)


docs/sysadmin-deploy-on-prem/linux-packages.txt
```diff
@@ -194,6 +194,50 @@ Master and Agent
...
+
+   sudo systemctl edit determined-master.socket
+
+which will open a text editor window. To change the listening port, insert the following text (with
+the port number substituted appropriately) into the editor and then exit the editor:
+
+.. code::
+
+   [Socket]
+   ListenStream=
```
ioga, line 226: it's odd this line is required but I've tried to remove it and it broke stuff 🤷

ioga (approved):
lgtm

docs/sysadmin-deploy-on-prem/linux-packages.txt
dzhu: There can be multiple values, so taking this out causes systemd to configure two sockets. This line tells it to clear the list so that we only get the one from the next line.

PR #3368, opened 2021-12-22
feat: accept env and envFrom in k8s

eecsliu:
## Description
This change allows determined pods in k8s to accept env vars from the `pod_spec` field in the config. 

## Checklist


master/internal/resourcemanagers/kubernetes/spec.go
```diff
@@ -365,7 +365,21 @@ func (p *pod) createPodSpec(ctx *actor.Context, scheduler string) error {
...
+	if podSpec != nil {
+		for _, container := range podSpec.Spec.Containers {
+			if container.Name == "determined-container" {
+				for _, e := range container.Env {
+					specEnvVars[e.Name] = e.Value
+				}
+				specEnvFrom = container.EnvFrom
+			}
+		}
+	}
```
stoksc, line 380: why is this not all in `configureEnvVars`? or, better even (? maybe), it seems like it could be merged in `configurePodSpec`, where stuff for the `determinedContainer` is merged into it from the default pod spec (where requests/limits are merged in). there are a few patterns here, it's a bit confusing to figure out where stuff happens and why i think. why this way?

master/internal/resourcemanagers/kubernetes/spec.go
```diff
@@ -365,7 +365,21 @@ func (p *pod) createPodSpec(ctx *actor.Context, scheduler string) error {
 		p.ports = append(p.ports, port)
 	}
 
-	envVars, err := p.configureEnvVars(spec.EnvVars(), env, deviceType)
+	specEnvVars := spec.EnvVars()
+	var specEnvFrom []k8sV1.EnvFromSource
+	podSpec := env.PodSpec()
+	if podSpec != nil {
+		for _, container := range podSpec.Spec.Containers {
+			if container.Name == "determined-container" {
```
stoksc, line 373: is this a const somewhere? seems like it should be.

master/internal/resourcemanagers/kubernetes/spec.go
eecsliu: Good points. I think my original thinking was that I should modify the env before the container gets constructed, but you're right - it does look like `configurePodSpec` is a much more natural place to put it.

master/internal/resourcemanagers/kubernetes/spec.go
eecsliu: oops - forgot to change this

master/pkg/model/environment_config.go
```diff
@@ -176,8 +176,6 @@ func validatePodSpec(podSpec *k8sV1.Pod) []error {
 						"container Ports are not configurable"),
 					check.Equal(len(container.EnvFrom), 0,
```
stoksc, line 177: should this check be here still?

stoksc, line 177: should this check be here still?

master/internal/resourcemanagers/kubernetes/spec.go
```diff
@@ -315,6 +315,13 @@ func (p *pod) configurePodSpec(
 			continue
 		}
 
+		for _, e := range container.Env {
```
stoksc, line 318: pro/con putting this in all containers vs just the determined container?

master/pkg/model/environment_config.go
eecsliu: fixed! Interestingly, this did not stop me from setting `EnvFrom`.

master/internal/resourcemanagers/kubernetes/spec.go
eecsliu: I thought about this and I didn't think there was a use case for setting env variables for the init container or fluent container. But allowing it would make it feel more true to the k8s experience as opposed to just borking it.

master/pkg/model/environment_config.go
stoksc: this validation would be what run for commands/shells/etc still

master/internal/resourcemanagers/kubernetes/spec.go
stoksc: yeh, probably never necessary, but idk.. if someone needs to drop aws creds into fluent for the connection to a elasticsearch.. lol idk, we'll get there when we get there

stoksc: probably a better way to do that anyway

stoksc (approved):
lgtm (of course, once tests pass)

master/internal/resourcemanagers/kubernetes/spec.go
eecsliu: I'll investigate and if I can support more containers easily, I'll include it

PR #3371, opened 2021-12-22
chore: combine entrypoint and launch layer

azhou-determined:
## Description

- combine entrypoint with launch layer
- separate out launch layer from native dtrain changes
## Checklist


harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -1143,11 +1143,6 @@
                 "string",
```
rb-determined-ai, line 1143: You also need to allow a list of strings

master/static/srv/entrypoint.sh
```diff
@@ -64,4 +64,4 @@ test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
 # Do rendezvous last, to ensure all launch layers start around the same time.
 "$DET_PYTHON_EXECUTABLE" -m determined.exec.prep_container --rendezvous
 
-exec "$DET_PYTHON_EXECUTABLE" -m determined.exec.launch_autohorovod "$@"
+exec "$DET_PYTHON_EXECUTABLE" -m determined.launch "$@"
```
rb-determined-ai, line 67: Let's figure out the layout of naming we want for this.

What I see is that we have plans to make `determined.launch.autohorovod` a user-facing name, but `determined.launch` an internal name (i.e., only we call it, only once, and in the entrypoint.sh).  Perhaps we leave `determined.exec` for our internal executable scripts, like `determined.exec.launch` and we use `determined.launch.*` as user-facing names?

harness/determined/launch/autohorovod.py
```diff
@@ -1,10 +1,10 @@
 """
-launch_autohorovod.py is the default launch layer for Determined.
+autohorovod.py is the default launch layer for Determined.
```
rb-determined-ai, line 2: I thought there was no default launch layer?

harness/determined/launch/__main__.py
```diff
@@ -0,0 +1,66 @@
...
+        storage.validate_config(
+            experiment_config["checkpoint_storage"],
+            container_path=constants.SHARED_FS_CONTAINER_PATH,
+        )
+    except Exception as e:
+        logging.error("Checkpoint storage validation failed: {}".format(e))
+        return 1
+
+    # Parse entrypoint command. If launch script not specified or using legacy configs,
+    # assume non-distributed training
```
rb-determined-ai, line 54: This is a breaking change (do you see why?)

azhou-determined: yeah, i'm not sure how to make this clear to users though. we probably need to update documentation. is there any other way we should announce this change to users?

harness/determined/launch/autohorovod.py
azhou-determined: bad comment.

master/static/srv/entrypoint.sh
azhou-determined: agree with this scheme. moved `launch` into determined.exec

harness/determined/common/schemas/expconf/_gen.py
azhou-determined: made the change. 👍

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -242,6 +243,28 @@ func (b BindMountsConfigV0) Merge(other interface{}) interface{} {
...
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (e *EntrypointV0) UnmarshalJSON(data []byte) error {
+	var entrypoint []string
+	if err := json.Unmarshal(data, &entrypoint); err == nil {
+		e.RawEntrypoint = []string{}
+		e.RawEntrypoint = append(e.RawEntrypoint, entrypoint...)
+		return nil
+	}
+	e.RawEntrypoint = strings.Fields(string(data))
```
rb-determined-ai, line 264: Maybe just leave the string in its original format?  I think the expected behavior here is for the single-string case to be interpreted as `sh -c $the_string`, which means that you can't really parse it here.

Also, the master doesn't really care what is inside of this struct anyway, so maybe you could get away with having it be an `interface{}`?  Not sure if that will lead to complications or not.

harness/determined/launch/__main__.py
azhou-determined: reworded the comment as we're not enforcing launch layer as of yet.

master/pkg/schemas/expconf/experiment_config.go
azhou-determined: done

harness/determined/common/schemas/expconf/_gen.py
azhou-determined: done

schemas/expconf/v0/experiment.json
```diff
@@ -71,13 +71,9 @@
         "entrypoint": {
             "type": [
                 "string",
+                "array",
```
rb-determined-ai, line 74: Oh, sorry, I left the comment about "items" in _gen.py, but you need to make the change here, not in _gen.py (which is generated from this)

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,74 @@
...
+            experiment_config["checkpoint_storage"],
+            container_path=constants.SHARED_FS_CONTAINER_PATH,
+        )
+    except Exception as e:
+        logging.error("Checkpoint storage validation failed: {}".format(e))
+        return 1
+
+    # Parse entrypoint command. If launch script is not specified, assume we are using legacy
+    # configs (model_def:Trial) and defaulting to horovod if distributed training
+    entrypoint_cmd = experiment_config.get("entrypoint")
```
rb-determined-ai, line 58: you use `.get()` but you don't handle the None case.

Actually this reminds me, there's still an old legacy case for native experiments which don't use `entrypoint`... so you actually do have to handle that case.  We have at least one large customer who depends on the old Native API still.

harness/determined/launch/autohorovod.py
```diff
@@ -81,6 +58,9 @@ def main() -> int:
     # easiest way to get the chief_ip is with container_addrs.
     chief_ip = info.container_addrs[0]
 
+    # Chief IP is set as an environment variable to support multiple launch layers
```
rb-determined-ai, line 61: nit: can you say "to support nested launch layers"?  Because supporting "multiple launch layers" could also mean autohorovod vs torch-native vs tensorflow-native launching, or something like that.

rb-determined-ai (commented):
I think this looks very good now.

Can you add some test cases to test the bounds of this new entrypoint field in `schemas/tests`?

harness/determined/exec/launch.py
azhou-determined: done

e2e_tests/tests/config.py
```diff
@@ -196,3 +196,9 @@ def set_profiling_enabled(config: Dict[Any, Any]) -> Dict[Any, Any]:
     config.setdefault("profiling", {})
     config["profiling"]["enabled"] = True
     return config
+
+
+def set_entrypoint(config: Dict[Any, Any], entrypoint: str) -> Dict[Any, Any]:
+    config = config.copy()
+    config["entrypoint"] = entrypoint
+    return config
```
rb-determined-ai, line 204: Can you configure your editor to always write the last line of the file with a terminating newline?  That is an old unix standard, and lots of other people's editors will do that automatically, so it creates annoying git churn when some files are created not like that.

schemas/expconf/v0/entrypoint.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://determined.ai/schemas/expconf/v0/entrypoint.json",
+  "title": "Entrypoint",
+  "union": {
+    "items": [
+      {
+        "unionKey": "const:type=array",
+        "$ref": "http://determined.ai/schemas/expconf/v0/entrypoint-str.json"
```
rb-determined-ai, line 9: `type=array` but `$ref=entrypoint-str.json`

schemas/expconf/v0/entrypoint.json
```diff
@@ -0,0 +1,37 @@
...
+      },
+      {
+        "unionKey": "const:type=string",
+        "$ref": "http://determined.ai/schemas/expconf/v0/entrypoint-array.json"
```
rb-determined-ai, line 13: `type=str` but `$ref=entrypoint-array.json`

schemas/expconf/v0/entrypoint.json
```diff
@@ -0,0 +1,37 @@
...
+      },
+      {
+        "unionKey": "always",
+        "type": "object",
+        "checks": {
+          "if a hyperparameter object's [\"type\"] is set, it must be one of \"int\", \"double\", \"log\", const\", or \"categorical\"": {
```
rb-determined-ai, line 19: hyperparameters?

schemas/expconf/v0/entrypoint.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://determined.ai/schemas/expconf/v0/entrypoint.json",
```
rb-determined-ai, line 3: out of curiosity, why go the Entrypoint-as-an-object route, vs the way you had it before, with the `"items": {"type": "string"}`?

Actually... is entrypoint.json actually used anywhere?

schemas/expconf/v0/experiment.json
```diff
@@ -71,11 +71,27 @@
...
+                    }
+                }
+            },
+            "items": {
+                "type": "string"
+            },
+            "if": {
+                "type": "string"
+            },
+            "then": {
```
rb-determined-ai, line 90: there's an `"if"`/`"then"` in json schema?? whaaat??

schemas/expconf/v0/experiment.json
```diff
@@ -71,11 +71,27 @@
...
+                "checks": {
+                    "entrypoint must contain a trial class of the form \"module.submodule:ClassName\"": {
+                        "pattern": "(^| )[a-zA-Z0-9_.]+:[a-zA-Z0-9_]+.*$"
+                    }
```
rb-determined-ai, line 94: What if the user has an entrypoint like:
```yaml
entrypoint: ./train.py
```
Should that be valid?

schemas/expconf/v0/experiment.json
```diff
@@ -71,11 +71,27 @@
...
             "type": [
                 "string",
+                "array",
                 "null"
             ],
             "checks": {
-                "entrypoint must be of the form \"module.submodule:ClassName\"": {
-                    "pattern": "^[a-zA-Z0-9_.]+:[a-zA-Z0-9_]+$"
+                "entrypoint must contain a trial class of the form \"module.submodule:ClassName\"": {
+                    "contains": {
```
rb-determined-ai, line 79: I'm not actually familiar with the "contains" keyword.  What is this doing?

schemas/test_cases/v0/entrypoint.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - http://determined.ai/schemas/expconf/v0/experiment.json
+  case:
+    entrypoint: ["python3", "-m", "determined.launch.autohorovod", "model_def:Trial"]
+
+- name: entrypoint check (array, invalid)
+  sanity_errors:
+    http://determined.ai/schemas/expconf/v0/experiment.json:
+      - "entrypoint must contain a trial class of the form \"module.submodule:ClassName\""
+  case:
+    entrypoint: ["python3", "-m", "determined.launch.autohorovod"]
```
rb-determined-ai, line 56: nit: another file without a terminating newline

schemas/test_cases/v0/entrypoint.yaml
```diff
@@ -0,0 +1,56 @@
+
```
rb-determined-ai, line 1: nit: delete empty line at top of file

harness/determined/load.py
```diff
@@ -150,11 +150,14 @@ def get_trial_controller_class(trial_class: Type[det.Trial]) -> Type[det.TrialCo
...
+    train_entrypoint: Optional[str],
 ) -> Tuple[Type[det.Trial], Type[det.TrialController]]:
     if experiment_config.native_enabled():
         command = experiment_config["internal"]["native"]["command"]  # type: List[str]
         trial_class = get_trial_class_from_native(command)
     else:
-        trial_class = trial_class_from_entrypoint(experiment_config["entrypoint"])
+        if not train_entrypoint:
+            raise ValueError("Entrypoint must be specified when native is disabled")
+        trial_class = trial_class_from_entrypoint(train_entrypoint)
```
rb-determined-ai, line 161: Is this helper function actually used anywhere except harness.py::main()?

It seems like the helper function has ceased to become useful.  It used to parse things from the config, but now it forces the caller to parse some things out of the config, while still requiring the config to get passed in to parse other things.  It might be cleaner to just do away with it, IMO.

Non-blocking though, feel free to disagree.

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,94 @@
...
+
+    entrypoint_cmd = experiment_config.get("entrypoint", None)
+    slots_per_trial = experiment_config.get("resources", {}).get("slots_per_trial", 1)
+
+    if not entrypoint_cmd:
+        logging.error("Entrypoint not specified in experiment config")
+        return 1
+
+    if isinstance(entrypoint_cmd, str):
+        entrypoint_cmd = entrypoint_cmd.split(" ")
```
rb-determined-ai, line 75: What if somebody submits something like this:
```
entrypoint: ./my\ train\ script\ with\ spaces.py
```

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,94 @@
...
+
+    logging.info(f"Launching: {entrypoint_cmd}")
+
+    # Legacy entrypoint (model_def:Trial) detected
+    # Default to horovod if distributed training, else continue in non-distributed training mode
```
rb-determined-ai, line 80: nit: don't put a comment that says "Legacy entrypoint detected" outside the `if` statement, because at the indentation level you have that statement, you actually haven't detected that yet.  But I like the comment, I think it is a good explanation, just not quite in the right place.

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,94 @@
...
+    if len(entrypoint_cmd) == 1:
```
rb-determined-ai, line 81: What if somebody submits something like this:
```yaml
entrypoint:
- ./train.py
```

master/pkg/model/test_utils.go
```diff
@@ -37,7 +37,7 @@ func ExperimentModel(opts ...ExperimentModelOption) *Experiment {
 				RawMaxLength: &maxLength,
 			},
 		},
-		RawEntrypoint: ptrs.StringPtr("model_def:SomeTrialClass"),
+		RawEntrypoint: &expconf.EntrypointV0{RawEntrypoint: "model_def:SomeTrialClass"},
```
rb-determined-ai, line 40: Don't use versioned symbols (`EntrypointV0`) outside of the expconf package.  There's a `latest.go` where you should create an alias.

harness/determined/experimental/_native.py
```diff
@@ -110,14 +111,25 @@ def _make_test_workloads(config: det.ExperimentConfig) -> workload.Stream:
...
 def _load_trial_for_checkpoint_export(
     context_dir: pathlib.Path,
     managed_training: bool,
     config: Dict[str, Any],
     hparams: Dict[str, Any],
 ) -> Tuple[Type[det.Trial], det.TrialContext]:
     with det._local_execution_manager(context_dir):
-        trial_class = load.trial_class_from_entrypoint(config["entrypoint"])
+        trial_class_entrypoint = _parse_trial_class_from_entrypoint(config["entrypoint"])
+        assert trial_class_entrypoint is not None
```
rb-determined-ai, line 131: If there is a chance that a users hits a "silent" assertion error (one without an explanation afterwards) during `Checkpoint.load()` then we need to either eliminate that chance or write an error message.

In this case, what will this do when you load a checkpoint that doesn't have the legacy entrypoint field?  I think writing an error message would be an insufficient solution; I think there are cases here that are just broken.

schemas/expconf/v0/experiment.json
azhou-determined: it's a new-ish keyword (added in json schema draft-06 i believe) that checks if one or more of the items in an array follows specification (http://json-schema.org/draft-06/json-schema-validation.html#rfc.section.6.14)

schemas/expconf/v0/experiment.json
azhou-determined: this is necessary because the array/ "contains" check will fail when it's a string type

schemas/expconf/v0/entrypoint.json
azhou-determined: my bad, i had originally had this "Entrypoint-as-an-object route" implementation, but in the end went with the "items" as before. so this object was not being used. i've deleted it.

schemas/expconf/v0/experiment.json
azhou-determined: no. until we have the trainer API available, harness needs module:ClassName specification to load context and instantiate user code.

harness/determined/load.py
azhou-determined: yeah, I thought about this too. but i think the helper function does more than parse things from the config. i think it's mostly useful right now as differentiating between which type of trial class (native or not) to load.

rb-determined-ai: But actually the caller of this helper function has always already checked that case before in `launch.py`, where you have a codepath that checks for `native_enabled`.

harness/determined/exec/launch.py
azhou-determined: changed the subprocess to open a shell to avoid shell command parsing.

harness/determined/experimental/_native.py
azhou-determined: as discussed, I'm going to save the trial class entrypoint specification in checkpoint metadata. once your checkpoint API changes are merged, I can make the changes. for now, launch layer works only with legacy model_def:Trial format.

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,98 @@
...
+        from determined.exec import harness
+
+        return harness.main(train_entrypoint=None)
+
+    entrypoint = experiment_config.get_entrypoint()
+    slots_per_trial = experiment_config.slots_per_trial()
+
+    if not entrypoint:
+        logging.error("Entrypoint not specified in experiment config")
+        return 1
```
rb-determined-ai, line 33: maybe raise an error?  This technically works but is a little unpythonic.

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if isinstance(entrypoint, list):
+        entrypoint = " ".join(entrypoint)
```
rb-determined-ai, line 36: How do you distinguish between:
```yaml
entrypoint: "./my file with spaces"
```
and
```yaml
entrypoint: ["./my", "file", "with", "args"]
```
?

(there's a fundamental problem with your strategy here, can you name it?)

harness/determined/exec/launch.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if match_legacy_trial_class(entrypoint):
+        # Legacy entrypoint ("model_def:Trial" or ["model_def:Trial"]) detected
```
rb-determined-ai, line 39: wait, why would we let somebody use the new format (`["model_def:Trial"]`) with the legacy behavior?

azhou-determined: well, i guess no one would/should do this. it seemed like a natural case to support since entrypoint can be a list or string type now, but it doesn't make sense to explicitly allow it for the legacy use case. i've removed it.

harness/determined/exec/launch.py
azhou-determined: ah, didn't account for spaces in commands. yeah, passing a string into shell makes it ambiguous as to whether you're looking for an executable with spaces in it or an executable with args passed in.

 i've made changes to pass in string or list as is to shell instead of doing any kind of parsing on commands.

rb-determined-ai (commented):
great work!

rb-determined-ai (approved):

PR #3375, opened 2022-01-03
ci: check webui in circleci

rb-determined-ai:
## Commentary

I can't actually tell definitively if this is being checked at any point in the automated tests already or not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1195,6 +1195,7 @@ jobs:
     steps:
       - checkout
       - react-get-deps
+      - run: make -C webui check
       - run: make -C webui/react check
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1195,6 +1195,7 @@ jobs:
     steps:
       - checkout
       - react-get-deps
+      - run: make -C webui check
```
hamidzr, line 1198: as is these are synonymous. is the idea that it'll be better if we just check the higher-level module and let that trickle down to `react` ?

rb-determined-ai: no, it's that I didn't realize they were synonymous.

PR #3376, opened 2022-01-03
fix: Fix sphinx-build parsing bug [DET-6369]

varlogtim:
The output from Sphinx wasn't displayed well and made it difficult to
determine what the actual error from sphinx-build was.

To test:

Add a link to a file that doesn't exist. Example here:
docs/sysadmin-deploy-on-k8s/install-on-kubernetes.txt

1) Make sure that the output with the wrapper looks the same as the
   output without the wrapper.
2) That the script exits non-zero and mentioned the missing file.


eecsliu (approved):
lgtm!

PR #3377, opened 2022-01-03
ci: stop testing EOL python

rb-determined-ai:
Python 3.6 became EOL on 23 Dec 2021:

    www.python.org/dev/peps/pep-0494/#lifespan

The Python 3.6 CLI test started failing because psutil no longer ships
wheels for Python 3.6 for the latest post-3.6-eof version, which means
our CI would need gcc to build that package.  Instead, just delete the
test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,9 +10,6 @@ orbs:
   queue: eddiewebb/queue@volatile
 
 executors:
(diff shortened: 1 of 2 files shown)


mackrorysd (approved):
LGTM. We should probably add a release note about dropping Python 3.6, though?

PR #3378, opened 2022-01-03
fix: stop re-rendering loops and throw the appropriate errors for continue trial modal [DET-6368]

hkang1:
## Description

There are couple of issues identified related to this issue.

1) The modal component function props has a dependency on the modal state. However, the way it is written the modal props constantly changes, causing the `openOrUpdate` to constantly be called, which causes a rendering loop.
2) The experiment create submission was not throwing the errors properly to trigger a failure when there is a form error or when the submission endpoint call returns with a failure.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/useModal.tsx
+++ b/webui/react/src/hooks/useModal/useModal.tsx
@@ -73,7 +73,7 @@ const useModal = (
             resolve(thenArgs);
             modalClose(reason);
           })
-          .catch(reject);
+          .catch(() => reject());
       }) as unknown as AntModalPromise;
     } else {
       return async (...args: any[]) => {
--- a/webui/react/src/hooks/useModal/useModal.tsx
+++ b/webui/react/src/hooks/useModal/useModal.tsx
@@ -73,7 +73,7 @@ const useModal = (
             resolve(thenArgs);
             modalClose(reason);
           })
-          .catch(() => reject());
+          .catch((e: unknown) => reject(e));
       }) as unknown as AntModalPromise;
     } else {
       return async (...args: any[]) => {
--- a/webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
+++ b/webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
@@ -1,6 +1,6 @@
 import { Alert, Form, FormInstance, Input, ModalFuncProps } from 'antd';
 import yaml from 'js-yaml';
-import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 
 import Icon from 'components/Icon';
 import Spinner from 'components/Spinner';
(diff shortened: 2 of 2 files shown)


webui/react/src/hooks/useModal/useModal.tsx
```diff
@@ -73,7 +73,7 @@ const useModal = (
             resolve(thenArgs);
             modalClose(reason);
           })
-          .catch(reject);
+          .catch(() => reject());
```
hkang1, line 76: this was required for some reason. The behavior is completely different. As `.catch(reject)` when errors are thrown, the `reject` function does NOT get called properly. 🤷

hkang1: I also noticed this behavior for `useEffect` return callbacks.

webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
```diff
@@ -330,10 +335,11 @@ const useModalExperimentCreate = (props?: Props): ModalHooks => {
...
    */
   useEffect(() => {
-    if (modalRef.current || (!prevModalProps?.visible && modalProps?.visible)) {
-      openOrUpdate(modalProps);
-    }
-  }, [ modalProps, modalRef, openOrUpdate, prevModalProps ]);
+    if (modalState === prevModalState || !modalState.visible) return;
+
+    const modalProps = getModalProps(modalState);
+    openOrUpdate(modalProps);
```
hkang1, line 341: The main thing this change does is, instead of relying on changes in the `modalProps` to trigger the `openOrUpdate`, it purely relies on `modalState`. This was necessary because doing a `isEqual` comparison on `modalProps` causes major issues. `JSON.stringify` on `modalProps.content` causes a bad complaint about circular dependency. This is probably best anyways to avoid having to compare such large objects.

webui/react/src/hooks/useModal/useModal.tsx
hamidzr: hmm yeah [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) doesn't show any new `.catch` call signature I'm not aware of either.
should we still pass in the error reason when we call reject?

webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
```diff
@@ -246,26 +198,78 @@ const useModalExperimentCreate = (props?: Props): ModalHooks => {
       }
 
       setModalState(prev => ({ ...prev, error: errorMessage }));
+
+      // We throw an error to prevent the modal from closing.
+      throw new Error();
```
hamidzr, line 203: suggestion: this seems to get swallowed by antd modal and not show up anywhere but in anycase it might help if we still include the original error or a message for it.

hamidzr (approved):
LGTM. thanks for fixing this

PR #3379, opened 2022-01-03
chore: upgrade golangci-lint to latest

stoksc:
## Description
This change upgrades `golangci-lint` to the latest version to use the `exhaustivestruct` linter on the expconf package. There was recently a bug where a mapping was incomplete and some fields from `model.Enviroment` weren't passed to `expconf.EnvironmentConfigV0`, resulting in a user being unable to use capabilities. In order to avoid other, similar bugs, I've enabled `exhaustivestruct` for every struct in the package and made sure they all conform or that it is OK when they don't.

## Test Plan
- [x] linters

## Commentary (optional)
I initally looked at this linter but decided instead to write a test with reflect, something like new(model.Environment).ToExpconf() and then making sure all the fields are set, but then i needed to write a lot of tests for each object and go get something to generate a saturated fake version to ensure there is something to check the existence of. After that I played with using the ast/parser packages and marking something like `exhaustivemapping` but then you need to dynamically import the type that is marked to see if it is exhaustive, which was too much of a pain in go. When I considered just marking the struct and parsing for the matching tags, i realized i'd mostly reinvented this linter. Though, the number of false positives make this linter fairly annoying, so I may revisit it soon.

## Checklist


stoksc:
forgot the agent..

stoksc:
blocked on #3381

master/Makefile
```diff
@@ -46,13 +46,13 @@ check-gen: force-gen gen
...
 .PHONY: get-deps
 get-deps:
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint
-	go install golang.org/x/tools/cmd/goimports
-	go install github.com/goreleaser/goreleaser
-	# For hyperparameter importance computation
-	go get github.com/ryanbressler/CloudForest
-	go install github.com/ryanbressler/CloudForest/growforest
-	go get github.com/vektra/mockery/v2/.../@v2.9.2
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
```
mackrorysd, line 49: I'm surprised that we didn't already move to the latest when we weren't specifying a version. Am I missing something?

master/.golangci.yml
```diff
@@ -85,3 +88,38 @@ linters:
     - testpackage
     - unparam
     - wsl
+
+    # Below here are new linters from upgrading to 1.43.0. Since we enable all and disable
+    # selectively, when we upgrade we get a ton of new linters. For convenience next upgrade,
+    # golangci-lint can tell you which linters are enabled:
+    #   golangci-lint linters | sed -n '/Enabled/,/Disabled/p'
+    # Only in the new list were disabled with this change:
```
mackrorysd, line 96: Don't understand this line. Missing something from the comment?

agent/tools.go
```diff
@@ -1,10 +0,0 @@
-//go:build tools
```
mackrorysd, line 1: Again, educate me: what was the point of this before?

master/internal/trial_test.go
```diff
@@ -1,3 +1,4 @@
+//nolint:exhaustivestruct
```
mackrorysd, line 1: Just curious how much due diligence was done on these. Are these all experiment configs like the one above, or are these ignored just when they're in tests because that's less critical?

agent/tools.go
mackrorysd: Ahh - this is explained here: https://github.com/determined-ai/determined/pull/3383

master/Makefile
stoksc: yeah we were, it was pinned in the `go.mod` and imported into `tools.go` so it was "used"

master/.golangci.yml
stoksc: tried to make it clearer

master/internal/trial_test.go
stoksc: so this is a test, so I wasn't super concerned, tests make weird partial versions of stuff all the time (but i did look at all of them). any time it didn't appear in a test i read carefully and left a comment about why it was ok.

master/pkg/schemas/expconf/storage_config.go
```diff
@@ -174,19 +174,20 @@ type AzureConfigV0 struct {
...
 		credSource = c
 	} else {
 		credSource = otherConfig
 	}
-	out.RawConnectionString = schemas.Copy(credSource.RawConnectionString).(*string)
-	out.RawAccountURL = schemas.Copy(credSource.RawAccountURL).(*string)
-	out.RawCredential = schemas.Copy(credSource.RawCredential).(*string)
-	return out
+
+	return AzureConfigV0{
```
stoksc, line 185: prob could do more like this, in a way, writing the merge like this with this linter makes sure it stays complete.

master/.golangci.yml
```diff
@@ -85,3 +88,38 @@ linters:
     - testpackage
     - unparam
     - wsl
+
+    # Below here are new linters from upgrading to 1.43.0. Since we enable all and disable
+    # selectively, when we upgrade we get a ton of new linters. For convenience next upgrade,
+    # golangci-lint can tell you which linters are enabled:
+    #   golangci-lint linters | sed -n '/Enabled/,/Disabled/p'
+    # To maintain the same set us linters, disable those in the new set that are not in the old:
```
mackrorysd, line 96:
```suggestion
    # To maintain the same set of linters, disable those in the new set that are not in the old:
```

mackrorysd (approved):

PR #3380, opened 2022-01-04
chore: make the workload compatible with the last incomplete training batch

slyviacassell:
Make the workload more compatible with the last incomplete training batch.

## Description

For instance, given a training dataloader with 795 samples whose global batch size is set to 8, the num of the last incomplete batch is 3.
If we set `min_validation_period: epochs: 10`, the workload would validate at the `(epochs * self.records_per_epoch) // self.global_batch_size=(10*795)//8=993`-th batch before patching which is not the end of a training epoch.
After patching, it would validate at the end of the training epoch.

However, this change is not suitable for `drop_last=True`, and maybe there is a more elegant way to pass `drop_last` into this class.

## Checklist

### DIFF
--- a/harness/determined/layers/_workload_sequencer.py
+++ b/harness/determined/layers/_workload_sequencer.py
@@ -151,12 +151,20 @@ class WorkloadSequencer(workload.Source):
             return batches
         if records is not None:
             check.gt(self.global_batch_size, 0, "global_batch_size must be positive")
-            return max(records // self.global_batch_size, 1)
+            # todo: make this compatible for determined.pytorch.DataLoader with drop_last=True (the last imcomplete batch)
+            # usually we set the drop_last to False, but some user would define it to True
+            # that meas we should pass this parameter setting to this class
+            # we're finding a more elegent way to implement it
+            return max(math.ceil(records // self.global_batch_size), 1)
         if epochs is not None:
             check.is_instance(self.records_per_epoch, int, "length must be an integer")
             assert self.records_per_epoch is not None
             check.gt(self.global_batch_size, 0, "global_batch_size must be positive")
-            return max((epochs * self.records_per_epoch) // self.global_batch_size, 1)
+            # todo: make this compatible for determined.pytorch.DataLoader with drop_last=True (the last imcomplete batch)
+            # usually we set the drop_last to False, but some user would define it to True
+            # that meas we should pass this parameter setting to this class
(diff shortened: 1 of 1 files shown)


rb-determined-ai:
Hi @slyviacassell, and thank you for digging into our code and offering a contribution!  Honestly, you dug into some of the ugly parts, I'm impressed.

Unfortunately, we are probably not going to land this change.  The problem you are addressing is real, but this isn't really a full solution.  We also have users whose training dataset changes in length over time.  Because of issues like this, there's a year-long rearchitecture work underway which will eliminate `records_per_epoch` entirely.  One of the outcomes of this effort will be the Core API [proposed here](https://github.com/rb-determined-ai/determined/blob/sprinkle-spec/sprinkle/user-facing/low-level.md).

The fundamental problem with the WorkloadSequencer is that it tries to dictate how long epochs must be, rather than allowing dataloaders to define the length of epochs according to the actual data.  The necessity of `records_per_epoch` is a side-effect of that.

We're working on newer trainer APIs (proposed [here](https://github.com/rb-determined-ai/determined/blob/sprinkle-spec/sprinkle/user-facing/keras.md) and [here](https://github.com/rb-determined-ai/determined/blob/sprinkle-spec/sprinkle/user-facing/pytorch_trial.md)) which will hopefully use exclusively the actual dataloaders to define how long training goes for.  When those APIs are ready, then only legacy user code will need to specify `records_per_epoch` t all.

As it happens this WorkloadSequencer was migrated from the determined-master (in golang) to here (in python) only in 0.17.0.  We'll probably leave it as-is for a long while while users slowly migrate to the new APIs, then delete it.

rb-determined-ai:
@slyviacassell are you in [our slack](https://join.slack.com/t/determined-community/shared_invite/zt-cnj7802v-KcVbaUrIzQOwmkmY7gP0Ew), by chance?  I'd love to know more how you are using Determined in your work.

slyviacassell:
> @slyviacassell are you in [our slack](https://join.slack.com/t/determined-community/shared_invite/zt-cnj7802v-KcVbaUrIzQOwmkmY7gP0Ew), by chance? I'd love to know more how you are using Determined in your work.

Thank you for your patient reply. I will join the slack.

slyviacassell:
As the official developer suggested, another workflow is trying to handle this.

slyviacassell:
As the official developer suggested, another workflow is trying to handle this.

PR #3381, opened 2022-01-04
chore: force github.com/containerd/containerd upgrade

stoksc:
## Description
This should hit the last dependabot bot alert. I thought the docker bump would hit this but since it doesn't seem to require a higher version, the go mod tool didn't bump it.

## Test Plan
- [x] tests should pass

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3383, opened 2022-01-04
chore: remove tools.go/go.mod dep tracking pattern, just go install versions of tools

stoksc:
## Description
We used to track versions of tool dependencies in the `go.mod` files (as was pretty normal for go). We would `go get` it to pin a version in the `go.mod`, add a line to `tools.go` so it was "used" and then go install it (this is what cloudforest did, some cases were slightly different). It's really nice, though, to separate tool dependencies from code we compile into the master. This means the `go.mod` is much more straightforward when it comes to questions like 'how did this get here?', and in general there should be fewer funky deps and spurious security alerts.

There are some downsides; dependabot may not alert us if there are security issues in tools now, but for tools, we can probably pin to latest mostly, anyway. Also, there are some times we use a tool and code from the same repo and they need to be kept in sync (so upgraders will need to catch both places). But this is an OK trade to me, all things considered.

## Test Plan
- [ ] tests should pass

## Commentary (optional)
this is not critical but i think it makes things a lot clearer (and is kinda the normal pattern for this afaict, from reading the installation guides for all these tools as they exist now).

## Checklist


stoksc:
blocked on #3381

master/Makefile
```diff
@@ -46,13 +46,13 @@ check-gen: force-gen gen
...
-	go install github.com/goreleaser/goreleaser
-	# For hyperparameter importance computation
-	go get github.com/ryanbressler/CloudForest
-	go install github.com/ryanbressler/CloudForest/growforest
-	go get github.com/vektra/mockery/v2/.../@v2.9.2
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.28.3
+	go install github.com/bufbuild/buf/cmd/buf@v0.42.1
+	go install golang.org/x/tools/cmd/goimports@v0.1.5
+	go install github.com/goreleaser/goreleaser@v1.1.0
+	go install github.com/ryanbressler/CloudForest/growforest@v0.0.0-20161201194407-d014dc32840a
```
mackrorysd, line 53: You mentioned pinning to latest to mitigate the problem with not having dependabot alerts. CloudForest should possibly be pinned to avoid unexpected changes since we bundle it into the product. Why not pin other build tools to latest?

stoksc: Just because drift may break CI without warning (golangci-lint definitely would). There may be a way to rework things such that pinning to latest is easier but it always could happen that tools make breaking changes and we need to update config files/scripts for them. I wanted to avoid that.

mackrorysd (approved):
LGTM, modulo the commits and the blocking PR. This is very likely to cause a downstream merge conflict after merging, so let's get all related changes in and then go fix that.

PR #3388, opened 2022-01-06
feat: Add text augmentation example

mapmeld:
## Description

Adds a subclass of the BERT / GLUE task using the TextAttack library to extend training data. The config file has a custom list of Transformations from the library:

```
data:
  augmentations:
    - WordSwapRandomCharacterDeletion
    - WordSwapHomoglyphSwap
```

Potential todos:
- support other types of augmentation (not only Transformations)
- refactor `preprocess_function` so the subclass needs less code

## Checklist


model_hub/examples/huggingface/text-classification/startup-hook.sh
```diff
@@ -0,0 +1,2 @@
+pip install attrdict model-hub tensorflow-text nltk==3.6.5 textattack==0.3.4 datasets transformers
```
liamcli, line 1: attrdict, model-hub, datasets, and transformers are installed automatically in the model-hub-transformers docker image so we shouldn't need to install them as part of startup.

liamcli:
In my run, the experiment without data augmentation reached 57.3 matt corr on COLA while with data augmentation reached 59.3.  Let's update the readme for this example to include a section on data augmentation and the new results that we get.  

Also, please add the data augmentation experiment to our e2e tests at determined/e2e_tests/tests/model_hub/test_transformers.py.  

Nice work!

liamcli (changes requested):

model_hub/examples/huggingface/text-classification/startup-hook.sh
mapmeld: :+1:

liamcli (changes requested):
Looking good.  Let's update the result charts and README to show the new numbers (~57 for no aug and 59 for with aug).  Matt corr 57 on COLA is closer to the numbers they report in the upstream repo as well: https://github.com/huggingface/transformers/tree/master/examples/pytorch/text-classification

PR #3391, opened 2022-01-06
fix: set gc-policy broken [DET-6373]

varlogtim:
To test fix:

1. Run an experiment which creates a checkpoint.
2. Run the following command:
det experiment set gc-policy \
  --save-experiment-best 0 \
  --save-trial-best 0 \
  --save-trial-latest 0 <exp_id>

^ this should not error and remove all the checkpoints.

Tested manually.

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -593,7 +593,7 @@ def set_gc_policy(args: Namespace) -> None:
                 c["uuid"],
                 render.format_resources(c["resources"]),
             ]
-            for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["step_id"]))
+            for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["step"]["id"]))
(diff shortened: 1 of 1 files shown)


harness/determined/cli/experiment.py
```diff
@@ -593,7 +593,7 @@ def set_gc_policy(args: Namespace) -> None:
                 c["uuid"],
                 render.format_resources(c["resources"]),
             ]
-            for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["step_id"]))
+            for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["step"]["id"]))
```
stoksc, line 596: i think we just shouldn't be using step id anymore (checkpoints soon won't be tied to a step, so this will change completely)
```suggestion
            for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["end_time"]))
```

stoksc (approved):
lgtm (as long as its tested, of course)

PR #3393, opened 2022-01-06
fix: avoid Can't pickle local object in TestPIDServer

mpkouznetsov:
## Description

On MacOS, with python 3.9 (which is required to install tensorflow on Mac with Apple Silicon),
I see TestPIDServer tests fail, with similar stack traces:

```
            for p in procs:
>               p.start()

tests/test_ipc.py:332:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../miniforge3/lib/python3.9/multiprocessing/process.py:121: in start
    self._popen = self._Popen(self)
../../../miniforge3/lib/python3.9/multiprocessing/context.py:224: in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
../../../miniforge3/lib/python3.9/multiprocessing/context.py:284: in _Popen
    return Popen(process_obj)
../../../miniforge3/lib/python3.9/multiprocessing/popen_spawn_posix.py:32: in __init__
    super().__init__(process_obj)
../../../miniforge3/lib/python3.9/multiprocessing/popen_fork.py:19: in __init__
    self._launch(process_obj)
../../../miniforge3/lib/python3.9/multiprocessing/popen_spawn_posix.py:47: in _launch
    reduction.dump(process_obj, fp)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = <Process name='Process-8' parent=47213 initial>, file = <_io.BytesIO object at 0x107928400>, protocol = None

    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       AttributeError: Can't pickle local object 'TestPIDServer.test_health_check_pre_connect.<locals>.worker_proc'

../../../miniforge3/lib/python3.9/multiprocessing/reduction.py:60: AttributeError
```

According to the chatter here https://github.com/spack/spack/issues/14102, python 3.8 was the first version to outlaw pickling local objects (explicitly raising an exception), which is why these tests pass with python<=3.7 .

I am not sure whether these tests fail on Linux with python>=3.8 or this is Mac-specific.

## Checklist

### DIFF
--- a/harness/tests/test_ipc.py
+++ b/harness/tests/test_ipc.py
@@ -237,20 +237,34 @@ def test_distributed_context(cross_size: int, local_size: int, force_tcp: bool) 
 
 
 class TestPIDServer:
+    @staticmethod
+    def _worker_proc(
+        addr: int,
+        keep_alive: bool = False,
+        sleep_time: float = 10,
+        repeat: int = 1,
+        crash: bool = False,
+    ) -> None:
+        with ipc.PIDClient(addr) as pid_client:
+            for _ in range(repeat):
+                if keep_alive:
+                    pid_client.keep_alive()
+                time.sleep(sleep_time)
+            if crash:
+                raise ValueError("Crashing...")
+
     def test_normal_execution(self) -> None:
         with ipc.PIDServer(addr=0, num_clients=2) as pid_server:
             assert pid_server.listener
             _, port = pid_server.listener.getsockname()
 
-            def worker_proc() -> None:
-                with ipc.PIDClient(port) as pid_client:
-                    for _ in range(5):
-                        pid_client.keep_alive()
-                        time.sleep(0.1)
-
             procs = [
-                multiprocessing.Process(target=worker_proc),
-                multiprocessing.Process(target=worker_proc),
+                multiprocessing.Process(
+                    target=TestPIDServer._worker_proc, args=(port, True, 0.1, 5)
+                ),
+                multiprocessing.Process(
+                    target=TestPIDServer._worker_proc, args=(port, True, 0.1, 5)
+                ),
             ]
 
             for p in procs:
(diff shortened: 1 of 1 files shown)


harness/tests/test_ipc.py
```diff
@@ -237,20 +237,35 @@ def make_distributed_context(rank: int, cross_rank: int, local_rank: int) -> Any
...
             assert pid_server.listener
             _, port = pid_server.listener.getsockname()
 
-            def worker_proc() -> None:
-                with ipc.PIDClient(port) as pid_client:
-                    for _ in range(5):
-                        pid_client.keep_alive()
-                        time.sleep(0.1)
-
+            # starting with python 3.8, an error is raised when process target is a local function
```
rb-determined-ai, line 261: Probably explain this in the git commit message, but not in a comment.  My thinking is that what you are saying applies to the transition from old code to new code more than it applies in particular to the new code.  If I saw this comment in the future and hadn't known what the test looked like before, I'd be pretty confused by it.

harness/tests/test_ipc.py
```diff
@@ -237,20 +237,35 @@ def make_distributed_context(rank: int, cross_rank: int, local_rank: int) -> Any
 
 
 class TestPIDServer:
+    @staticmethod
+    def _worker_proc(
+        addr: Union[str, int, Tuple[str, int]],
```
rb-determined-ai, line 242: might as well make the `addr` just a single port, since that's all we invoke it with

rb-determined-ai (approved):
looks good

PR #3395, opened 2022-01-06
fix: add k8s permissions to fix patch bug

eecsliu:
## Description
We were encountering an issue where k8s would not be able to update pod labels - this ended up not being necessary to solve the problem, but I think the patch pod permission should still be granted to the cluster.

## Checklist

### DIFF
--- a/helm/charts/determined/templates/master-permissions.yaml
+++ b/helm/charts/determined/templates/master-permissions.yaml
@@ -25,7 +25,7 @@ rules:
     verbs: ["get", "list"]
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["watch"]
+    verbs: ["watch", "patch"]
   - apiGroups: [""]
     resources: ["nodes", "events"]
     verbs: ["list", "watch"]


master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -374,10 +374,6 @@ func (k *kubernetesResourceManager) assignResources(
 		//	k.addrToContainerID[req.TaskActor] = []cproto.ID{}
 		//}
 		k.addrToContainerID[req.TaskActor] = container.id
-		ctx.Tell(k.agent.handler, kubernetes.SetPodOrder{
```
hamidzr, line 377: is this intentional?

hamidzr (approved):

master/internal/resourcemanagers/kubernetes_resource_manager.go
eecsliu: yes - the pod label doesn't allow the "-" character, so this code is pointless.

dannysauer:
We gonna wait until Friday to merge this, so it can be a full year to get a one-word fix in? :stuck_out_tongue:

eecsliu:
absolutely we'll wait until Friday 😂

PR #3397, opened 2022-01-07
feat: add v2 of raw allocation endpoint

stoksc:
## Description
This change updates the raw allocation endpoint to use the allocations table. Because this table wasn't backfilled, before a certain time its results will need to be merged with results based on aggregating the steps table, or we at least need to note that in release notes.

## Test Plan
- [ ] add a test (probably over the entire thing, before landing)

## Commentary (optional)
I just realized Core API is going to completely bust v1 of this endpoint and we need to go ahead with this (cc @rb-determined-ai, just fyi). The old query uses the steps table, it really just has to go.

## Checklist

### DIFF
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -185,7 +185,7 @@ func (m *Master) getMasterLogs(c echo.Context) (interface{}, error) {
 	return entries, nil
 }
 
-// @Summary Get a detailed view of resource allocation during the given time period (CSV).
+// @Summary Deprecated, use /v2/allocation/raw for correct results after Determined version 0.17.8.
 // @Tags Cluster
 // @ID get-raw-resource-allocation-csv
 // @Accept  json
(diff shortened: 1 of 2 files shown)


master/internal/core.go
```diff
@@ -256,6 +256,84 @@ func (m *Master) getRawResourceAllocation(c echo.Context) error {
...
+	}
+	end, err := time.Parse("2006-01-02T15:04:05Z", args.End)
+	if err != nil {
+		return errors.Wrap(err, "invalid end time")
+	}
+	if start.After(end) {
+		return errors.New("start time cannot be after end time")
+	}
+
+	resp := &apiv1.ResourceAllocationRawResponse{}
```
stoksc, line 294: todo: add this to new api too, dont reuse this (even though it works)

stoksc:
this actually isn't that bad yet, and we're busy

PR #3399, opened 2022-01-07
chore: get gov images in refresh-ubuntu-amis.py

rb-determined-ai:
Unfortunately, this uses a non-public endpoint (the endpoint backing
the webui at https://cloud-images.ubuntu.com/locator/ec2), but that
seems to be the only automated way to get the gov-region amis.

### DIFF
--- a/tools/scripts/README.md
+++ b/tools/scripts/README.md
@@ -17,6 +17,14 @@ be registered in the determined repo. Here is the process:
 1. Land the desired change in the environments repo. Remember the full commit
    hash (we'll call it `THECOMMIT` in these steps).
 
+   Note, you might as well include an update to the base AMIs if you're going
+   to rebuild them.  From the `tools/scripts` directory of the determined repo,
+   run:
+
+   ```bash
+   ./refresh-ubuntu-amis.py --packer-json /path/to/environments-packer.json
+   ```
+
 2. Wait for the post-merge-to-master CircleCI jobs on the environments repo to
    finish. These will publish the relevant Docker/AWS/GCP images and create
    machine-readable artifacts containing the image tags.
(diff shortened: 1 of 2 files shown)


tools/scripts/README.md
```diff
@@ -31,14 +39,14 @@ be registered in the determined repo. Here is the process:
...
-   base images in environments/cloud/environments-packer.json also need to be
-   manually updated periodically. AWS GovCloud images need to be updated
-   manually as they are (at the time of this writing) missing or not current
-   in Ubuntu's AMI locator.
+5. Run `./refresh-ubuntu-amis.py --bumpenvs-yaml bumpenvs.yaml`.  This will
+   fetch the up-to-date Ubuntu AMIs for each region for each of the
+   `*_master_ami` and `*_bastion_ami` image tags in bumpenvs.yaml.  This isn't
+   strictly necessary; we just need to run it periodically, and now is a fine
+   time. The base images in environments/cloud/environments-packer.json also
+   need to be manually updated periodically. AWS GovCloud images need to be
```
mackrorysd, line 47: Nit: refresh-ubuntu-amis.py will now do it, they just have to go push those changes too.

tools/scripts/refresh-ubuntu-amis.py
```diff
@@ -65,30 +64,69 @@ def update_tag_for_image_type(subconf: Dict[str, str], new_tag: str) -> bool:
...
+    parser.add_argument("--bumpenvs-yaml", metavar="/PATH/TO/BUMPENVS.YAML")
+    parser.add_argument("--packer-json", metavar="/PATH/TO/ENIRONMENTS-PACKER.JSON")
+    args = parser.parse_args()
+
+    if not args.bumpenvs_yaml and not args.packer_json:
+        parser.print_help(sys.stderr)
+        sys.exit(1)
+
+    # Do one fetch for all the ubuntu amis.
+    # Note that this is not a public endpoint, but it's the only one that has the gov amis.
```
mackrorysd, line 121:
```suggestion
    # Note that this is not intended as a machine-readable endpoint, but it's the only one that has the gov amis.
```

tools/scripts/README.md
mackrorysd: Actually there's another circular dependency here: we refresh the base AMIs after rebuilding the AMIs for the upcoming release... I volunteer to fix this if I merge environments back into Determined next week, FWIW.

mackrorysd (approved):
Some minor comments and a suggestion, but please push when you've seen them.

tools/scripts/refresh-ubuntu-amis.py
rb-determined-ai: I mean... it is machine readable (it's ready by a javascript page) but that's different than a public api, which you could expect to be more stable over time.

PR #3402, opened 2022-01-07
test: Create Avatar component test [DET-6404]

EmilyBonar:
## Checklist

### DIFF
--- a/webui/react/src/components/Avatar.test.tsx
+++ b/webui/react/src/components/Avatar.test.tsx
@@ -12,7 +12,13 @@ jest.mock('antd', () => {
    * sets the DOM container and if this prop is set, the popup div may not be available in the body
    */
   const Tooltip = (props) => {
-    return <antd.Tooltip {...props} getPopupContainer={null} mouseEnterDelay={0} />;
+    return (
+      <antd.Tooltip
+        {...props}
+        getPopupContainer={(trigger: HTMLElement) => trigger}
+        mouseEnterDelay={0}
+      />
+    );
   };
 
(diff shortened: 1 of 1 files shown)


webui/react/src/components/Avatar.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+const initName = { initials: 'BB', name: 'Bugs Bunny' };
+
+jest.mock('antd', () => {
+  const antd = jest.requireActual('antd');
+
+  /** We need to mock Tooltip in order to override getPopupContainer to null. getPopupContainer
+   * sets the DOM container and if this prop is set, the popup div may not be available in the body
+   */
+  const Tooltip = (props) => {
+    return <antd.Tooltip {...props} getPopupContainer={null} mouseEnterDelay={0} />;
```
hkang1, line 15: **question:** looking at the TypeScript definition for `ToolTip` for prop `getPopupContainer`:
```
getPopupContainer?: (triggerNode: HTMLElement) => HTMLElement;
```
`null` is technically not a valid entry. Would it work to have it set to the following?
```
<ToolTip getPopupContainer={trigger => trigger} />
```
Where it gets added to the triggering element?

hkang1 (approved):
Nice work with this test! One question about Tooltip mock.

webui/react/src/components/Avatar.test.tsx
EmilyBonar: Yeah, that works too. I'll change it to that so it's a little more explicit.

PR #3403, opened 2022-01-07
ci: make checkpoint gc tests actually wait for gc

rb-determined-ai:
When the checkpoint gc tests were written, checkpoint gc ran directly on
the same compute node as the experiment ran, so waiting for 35 seconds
for checkpoint gc to finish was a reasonable design.

Now that we run checkpoint gc in a separate resource pool which might be
empty, we need to actually wait for the checkpoint gc to run.  There's
still a hard-coded 5-minute timeout, but cpu instances usually don't
take too long to spin up, so that should be reasonable.

### DIFF
--- a/e2e_tests/tests/cluster/test_checkpoints.py
+++ b/e2e_tests/tests/cluster/test_checkpoints.py
@@ -1,5 +1,6 @@
 import json
 import operator
+import sys
 import tempfile
 import time
 from typing import Any, Dict, Set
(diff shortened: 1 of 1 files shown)


stoksc (approved):
5m may not be long enough in crazy cases, i think (technically it could need to acquire an instance, right?). idk. otherwise lgtm

rb-determined-ai:
> technically it could need to acquire an instance
yes, but I'm operating under the assumption that the cpu-only instances are far more available than the gpu instances

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -8,11 +9,32 @@
...
+    for _ in range(600):
+        r = api.get(conf.make_master_url(), "tasks").json()
+        names = [task["name"] for task in r.values()]
+        gc_name = f"Checkpoint GC (Experiment {experiment_id})"
+        if gc_name in names:
+            saw_gc = True
+        elif saw_gc:
+            # We previously saw checkpoint gc but now we don't, so it must have finished.
+            return
+        time.sleep(0.5)
```
stoksc, line 32: eh, i mean, too, we could miss seeing it, right? but then hopefully the test would just pass and we'd have teh little error message. this is definitely far from perfect, hopefully the task apis will help.

rb-determined-ai: We would have to expose the checkpoint gc state for an experiment to make this absolutely not flaky.

PR #3405, opened 2022-01-08
feat: pass metrics to simple reducer in original order

mpkouznetsov:
## Description
Pass metrics to simple reducer in original order
[DET-5726]

The way that we shard changes the order of the observations (metric values captured by calls to `update`) when we later gather observations for custom reducers.

At a small extra cost we put the observations back in the original order before passing them to the reducer.

Note: 
1. This is only implemented for Pytorch and only for the "simple" metric reducer, i.e. one that is a function as opposed to a `MetricReducer`.
2. If the user calls `update` a variable number of times per batch (whether in training or validation) we cannot guarantee the order.

## Test Plan

A new unit test creates a bare-bones Trial-like object and distributes calls to `update` in a way consistent with how `DistributedBatchSampler` does it.
It then confirms that the reducer receives values in the original order.

## Commentary
Restoring the original order relies on two things: 1) allgather returning collected data in the order of worker rank (as opposed to a presumably random order gather_with_polling could return it) and 2) `_SimpleReducer.cross_slot_reduce` performing a zip-like "sweep" over the gathered metrics. The standard zip does not work because the number of batches/metrics handled by different workers may be off by one if the total number of batches is not divisible by the number of workers. Also, although it is impossible to support restoring the original order in case the user calls `update` a random number of times per batch, we take care to still submit _all_ the gathered metrics to the reducer. This makes for a slightly more complex implementation of `cross_slot_reduce`.

## Checklist

[DET-5726]: https://determinedai.atlassian.net/browse/DET-5726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/harness/determined/_core/_distributed.py
+++ b/harness/determined/_core/_distributed.py
@@ -256,11 +256,13 @@ class DistributedContext:
             return [stuff]
         logging.debug(f"Worker {self.get_rank()} beginning zmq allgather.")
         if self._is_chief:
-            worker_stuff, _ = self._chief_zmq.gather_with_polling(lambda: None)
+            worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
+            worker_stuff_ranked.sort(key=lambda x: x[1])
+            worker_stuff = [value for value, index in worker_stuff_ranked]
             all_stuff = [stuff, *worker_stuff]
             self._chief_zmq.broadcast(all_stuff)
         else:
-            self._worker_zmq.send(stuff)
+            self._worker_zmq.send((stuff, self.get_rank()))
             all_stuff = self._worker_zmq.recv()
         logging.debug(f"Worker {self.get_rank()} finished zmq allgather.")
         return all_stuff
--- a/harness/determined/_core/_distributed.py
+++ b/harness/determined/_core/_distributed.py
@@ -9,10 +9,6 @@ from typing import Any, Callable, List, Optional, Tuple, cast
 from determined import constants, ipc
 
 
-def _tuple_key_function(v: Any) -> int:
-    return cast(Tuple[Any, int], v)[1]
-
-
 class DistributedContext:
     """
     DistributedContext provides useful methods for effective distributed training.
@@ -220,12 +216,11 @@ class DistributedContext:
         logging.debug(f"Worker {self.get_rank()} beginning zmq gather.")
         if self._is_chief:
             worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
-            worker_stuff_ranked.sort(key=_tuple_key_function)
-            worker_stuff = [value for value, _ in worker_stuff_ranked]
+            worker_stuff = [value for _, value in sorted(worker_stuff_ranked)]
             self._chief_zmq.broadcast(None)
             out = [stuff, *worker_stuff]  # type: Optional[List]
         else:
-            self._worker_zmq.send((stuff, self.get_rank()))
+            self._worker_zmq.send((self.get_rank(), stuff))
             # Synchronize with the chief so that there is no risk of accidentally calling send()
             # for a future gather before all workers have called send() on this gather.
             _ = self._worker_zmq.recv()
@@ -243,12 +238,11 @@ class DistributedContext:
         logging.debug(f"Worker {self.get_rank()} beginning zmq gather local.")
         if self._is_local_chief:
             worker_stuff_ranked, _ = self._local_chief_zmq.gather_with_polling(lambda: None)
-            worker_stuff_ranked.sort(key=_tuple_key_function)
-            worker_stuff = [value for value, _ in worker_stuff_ranked]
+            worker_stuff = [value for _, value in sorted(worker_stuff_ranked)]
             self._local_chief_zmq.broadcast(None)
             out = [stuff, *worker_stuff]  # type: Optional[List]
         else:
-            self._local_worker_zmq.send((stuff, self.get_local_rank()))
+            self._local_worker_zmq.send((self.get_local_rank(), stuff))
             # Synchronize with the chief so that there is no risk of accidentally calling send()
             # for a future gather before all workers have called send() on this gather.
             _ = self._local_worker_zmq.recv()
@@ -265,12 +259,11 @@ class DistributedContext:
         logging.debug(f"Worker {self.get_rank()} beginning zmq allgather.")
         if self._is_chief:
             worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
-            worker_stuff_ranked.sort(key=_tuple_key_function)
-            worker_stuff = [value for value, _ in worker_stuff_ranked]
+            worker_stuff = [value for _, value in sorted(worker_stuff_ranked)]
             all_stuff = [stuff, *worker_stuff]
             self._chief_zmq.broadcast(all_stuff)
         else:
-            self._worker_zmq.send((stuff, self.get_rank()))
+            self._worker_zmq.send((self.get_rank(), stuff))
             all_stuff = self._worker_zmq.recv()
         logging.debug(f"Worker {self.get_rank()} finished zmq allgather.")
         return all_stuff
@@ -284,12 +277,11 @@ class DistributedContext:
         logging.debug(f"Worker {self.get_rank()} beginning zmq local allgather.")
         if self._is_local_chief:
             worker_stuff_ranked, _ = self._local_chief_zmq.gather_with_polling(lambda: None)
-            worker_stuff_ranked.sort(key=_tuple_key_function)
-            worker_stuff = [value for value, _ in worker_stuff_ranked]
+            worker_stuff = [value for _, value in sorted(worker_stuff_ranked)]
             all_stuff = [stuff, *worker_stuff]
             self._local_chief_zmq.broadcast(all_stuff)
         else:
-            self._local_worker_zmq.send((stuff, self.get_local_rank()))
+            self._local_worker_zmq.send((self.get_local_rank(), stuff))
             all_stuff = self._local_worker_zmq.recv()
         logging.debug(f"Worker {self.get_rank()} finished zmq local allgather.")
         return all_stuff
--- a/harness/determined/_core/_distributed.py
+++ b/harness/determined/_core/_distributed.py
@@ -4,7 +4,7 @@ import logging
 import shutil
 import socket
 import tempfile
-from typing import Any, Callable, List, Optional, Tuple, cast
+from typing import Any, Callable, List, Optional
 
 from determined import constants, ipc
 
--- a/harness/determined/_core/_distributed.py
+++ b/harness/determined/_core/_distributed.py
@@ -4,9 +4,13 @@ import logging
 import shutil
 import socket
 import tempfile
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, List, Optional, Tuple, cast
 
 from determined import constants, ipc
+
+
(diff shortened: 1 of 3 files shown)


harness/determined/pytorch/_reducer.py
```diff
@@ -184,7 +185,21 @@ def per_slot_reduce(self) -> Any:
         return self.values
 
     def cross_slot_reduce(self, per_slot_metrics: List) -> Any:
-        flat_metrics = [item for sublist in per_slot_metrics for item in sublist]
+
+        # restore the initial (pre-sharding) order of metrics
```
rb-determined-ai, line 189: This is a perfect place to put the explanation of assumptions and limitations that you have in the PR comment, I think.

Also, I have never used it, but I checked if `itertools` has a tool for this style of interleaved iteration and it turns that out [it does](https://docs.python.org/3/library/itertools.html#itertools.zip_longest)

harness/tests/experiment/pytorch/test_reducer.py
```diff
@@ -12,3 +23,96 @@ def test_reducer() -> None:
...
+            def _fn(rank: int, cross_rank: int, local_rank: int) -> None:
+                try:
+                    results[rank] = fn(rank, cross_rank, local_rank)
+                except Exception:
+                    errors[rank] = traceback.format_exc()
+                    raise
+
+            threads.append(threading.Thread(target=_fn, args=(rank, cross_rank, local_rank)))
+
+        for thread in reversed(threads):
```
rb-determined-ai, line 64: why reversed()?

harness/tests/experiment/pytorch/test_reducer.py
```diff
@@ -12,3 +23,96 @@ def test_reducer() -> None:
 
     batches_per_process = [1, 2, 5, 4, 5, 6]
     assert np.around(_reduce_metrics(Reducer.AVG, metrics, batches_per_process), decimals=2) == 6.43
+
+
+DummyTrial = namedtuple("DummyTrial", "distributed_context reducer_context wrapped_reducer")
```
rb-determined-ai, line 28: Since Trial is of particular meaning, perhaps rename this to something without "Trial"?  Don't waste too much time on it though.

harness/determined/_core/_distributed.py
```diff
@@ -256,11 +256,13 @@ def _zmq_allgather(self, stuff: Any) -> List:
             return [stuff]
         logging.debug(f"Worker {self.get_rank()} beginning zmq allgather.")
         if self._is_chief:
-            worker_stuff, _ = self._chief_zmq.gather_with_polling(lambda: None)
+            worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
```
rb-determined-ai, line 259: This looks good, but perhaps we should either:
* apply this rank-sort to all variations of gather/allgather in this file, or
* apply the rank-sort at a different layer

Waht do you think?

harness/determined/pytorch/_reducer.py
```diff
@@ -409,11 +424,15 @@ def reduce_metrics(self, for_training: bool) -> Dict[str, Any]:
...
         metrics = {}
 
+        # gatherables = enumerate(wrapped.per_slot_reduce() for wrapped in reducables)
+        # print(f"gatherables={gatherables}")
         gatherables = [wrapped.per_slot_reduce() for wrapped in reducables]
 
         # Do one allgather for all metrics to improve .
         gathered = self._allgather_fn(gatherables)
 
+        logging.debug(f"len(gathered)={len(gathered)}")
```
rb-determined-ai, line 434: what is the plan for this debug statement in?  Get left in?  Get cleaned up?

harness/tests/experiment/pytorch/test_reducer.py
```diff
@@ -12,3 +23,96 @@ def test_reducer() -> None:
...
+            trial.wrapped_reducer.update(batch)
+
+    results = do_parallel(lambda rank, _, __: trials[rank].reducer_context.reduce_metrics(False))
+    logger.debug(results)
+
+    # Close all distributed contexts
+    for trial in trials:
+        trial.distributed_context.close()
+
+    assert results[0]["sum"] == dataset_size * (dataset_size - 1) // 2
```
rb-determined-ai, line 115: maybe test all of the results?  Not sure it's critical though.

rb-determined-ai (commented):
Nice test!

harness/determined/_core/_distributed.py
mpkouznetsov: I don't see another layer where the rank-sort would be both meaningful and possible so I will go with the first option. (rank is not available outside `DistributedContext`, and we need both workers and the chief to be aware of rank being packaged inside the message).

harness/tests/experiment/pytorch/test_reducer.py
mpkouznetsov: I want the test to catch a possible incorrect implementation of `_zmq_allgather`.
 
If all operations are started in order, even on separate threads, there is a higher chance that the results of `gather_with_polling` will arrive in order as well. (The order is random but often correct). Then it may _appear_ that we can do without sorting the content reported by workers in `zmq_allgather`. By starting threads in reverse I am making it very unlikely that "worker stuff" will be gathered in order, thereby making it much more likely that the test will fail of the implementation is incorrect.
In other words, if someone screws up the implementation of `_zmq_allgather` by removing the sorting, the test will start failing intermittently, with or without `reverse`, but with `reverse` it will be failing most of the time.

I understand this is a mouthful, and maybe this `reversed` introduces more confusion than benefit.

harness/tests/experiment/pytorch/test_reducer.py
mpkouznetsov: Not sure we care. I mean, this would mostly be testing pre-existing functionality (the fact that chief and all workers see the same thing after `_zmq_allgather`). I guess it's not bad to have net new coverage.

harness/tests/experiment/pytorch/test_reducer.py
rb-determined-ai: Oh, of course. No this makes perfect sense.

Maybe a comment that says "encourage the allgather to occur in not-the-correct order to test the reordering" or something.

harness/determined/_core/_distributed.py
```diff
@@ -215,11 +219,13 @@ def _zmq_gather(self, stuff: Any) -> Optional[List]:
...
+            worker_stuff_ranked.sort(key=_tuple_key_function)
+            worker_stuff = [value for value, _ in worker_stuff_ranked]
             self._chief_zmq.broadcast(None)
             out = [stuff, *worker_stuff]  # type: Optional[List]
         else:
-            self._worker_zmq.send(stuff)
+            self._worker_zmq.send((stuff, self.get_rank()))
```
rb-determined-ai, line 228: You could reorder this tuple so you don't have to provide a key function, I believe.

I'm pretty sure that the ordering of tuples by default is based on the default sorting of their first elements.

harness/determined/_core/_distributed.py
```diff
@@ -215,11 +219,13 @@ def _zmq_gather(self, stuff: Any) -> Optional[List]:
             return [stuff]
         logging.debug(f"Worker {self.get_rank()} beginning zmq gather.")
         if self._is_chief:
-            worker_stuff, _ = self._chief_zmq.gather_with_polling(lambda: None)
+            worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
+            worker_stuff_ranked.sort(key=_tuple_key_function)
+            worker_stuff = [value for value, _ in worker_stuff_ranked]
```
rb-determined-ai, line 224: If you reorder the worker tuple, this could be:
```python
worker_stuff_ranked, _ = self._chief_zmq.gather_with_polling(lambda: None)
worker_stuff = [x[1] for x in sorted(worker_stuff_ranked)]
```

The way you have it ok though, so please consider this non-blocking.

rb-determined-ai (approved):
Nice work!

harness/determined/pytorch/_reducer.py
```diff
@@ -184,7 +185,17 @@ def per_slot_reduce(self) -> Any:
...
 
     def cross_slot_reduce(self, per_slot_metrics: List) -> Any:
-        flat_metrics = [item for sublist in per_slot_metrics for item in sublist]
+        """
+        Call the reducer function on metrics in their original order.
+
+        Interleave metrics from different slots as we flatten
+        the list of metrics to undo the effect of sharding.
+        Note: this will only reconstruct the original order if the user
+        calls ``update()`` the same number of times per batch
```
rb-determined-ai, line 194: wait, this will only reconstruct the original order if the user calls `update()` exactly 1 time per batch, right?  If they call it 2 times every batch, the order would be... well I don't know that they would care about ordering in that case.

harness/determined/pytorch/_reducer.py
```diff
@@ -184,7 +185,17 @@ def per_slot_reduce(self) -> Any:
...
+        """
+        interleaved = itertools.zip_longest(*per_slot_metrics)
+        flat_metrics = [m for sublist in interleaved for m in sublist if m is not None]
```
rb-determined-ai, line 197: Oh, I think we're silently preventing users from passing a None value.  Can we use some sort of sentinel that is not None?

```python
class NonMetric:
   pass

nonmetric = NonMetric()

interleaved = itertools.zip_longest(*per_slot_metrics, fillvalue=not_metric)
flat_metrics = [m for sublist in interleaved for m in sublist if m is not nonmetric]
```

rb-determined-ai (changes requested):
oops, I approved earlier but this is a subtle bug I noticed afterwards

harness/determined/pytorch/_reducer.py
mpkouznetsov: wow, good catch.

harness/determined/_core/_distributed.py
mpkouznetsov: I don't want to pass on an opportunity to simplify the code.

harness/determined/pytorch/_reducer.py
mpkouznetsov: lmk if the scoping I chose for `NotMetric` is ok.

harness/determined/pytorch/_reducer.py
```diff
@@ -183,8 +184,25 @@ def update(self, value: Any) -> None:
...
+        pass
+
+    NOT_METRIC = _NotMetric()
+
     def cross_slot_reduce(self, per_slot_metrics: List) -> Any:
-        flat_metrics = [item for sublist in per_slot_metrics for item in sublist]
+        """
+        Call the reducer function on metrics in their original order.
+
+        Interleave metrics from different slots as we flatten
```
rb-determined-ai, line 196: No need to reformat, but fwiw I don't think we normally wrap docstring lines this narrowly.

If there is good reason to start, let me know though.

rb-determined-ai (approved):

harness/determined/pytorch/_reducer.py
mpkouznetsov: No, there is no good reason to go down from 100; we have ample real estate nowadays.

The first sentence is the "header"; it is followed by an empty line.
The second sentence does not fit into one line, and rather than split it closer to the 100 character limit I prefer to split it in the middle.
The third sentence starts with "Note:" and therefore deserves to be on a separate line IMHO. And again it does not fit on a single line.

PR #3406, opened 2022-01-10
ci: increase resource_class for Docker image scan

mackrorysd:
## Description

The scan is timing out. I suspect (and am confirming) that a larger resource class would address this.


mackrorysd:
According to myself, on Slack the day I submitted this (very reliable source!), with the larger executor size it did finish but it took a couple of hours still. We can't go any higher on our current CircleCI plan.

ioga (approved):

ioga:
plan b: circleci parallelism and a run per image?

PR #3408, opened 2022-01-10
chore: make mypy happy with requests wrapper

rb-determined-ai:
The latest types-request==2.27.1 combined with our requests wrapper
module is confusing mypy, which we fix with a type cast.

### DIFF
--- a/harness/determined/common/requests.py
+++ b/harness/determined/common/requests.py
@@ -30,4 +30,5 @@ class Session(requests.sessions.Session):
 def request(method: str, url: str, **kwargs: Any) -> requests.Response:
     server_hostname = kwargs.pop("server_hostname", None)
     with Session(server_hostname) as session:
-        return session.request(method=method, url=url, **kwargs)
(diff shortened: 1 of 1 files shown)


harness/determined/common/requests.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, server_hostname: Optional[str]) -> None:
         self.mount("https://", HTTPAdapter(server_hostname))
 
 
-def request(method: str, url: str, **kwargs: Any) -> requests.Response:
+def request(method: str, url: str, **kwargs: Any) -> Any:
```
ioga, line 30: types-requests source code suggests it still should be a response: https://github.com/python/typeshed/blob/master/stubs/requests/requests/sessions.pyi#L107
is this mypy misbehaving again?..

rb-determined-ai: hm... not sure.  I'll just fix it with a type-cast.

ioga (approved):

PR #3417, opened 2022-01-11
chore: improve rate of false OK sends on websockets

stoksc:
## Description
This PR comprises two small changes. First, we try to return errors from writes on websockets to the caller more promptly by returning the error from `s.conn.WriteMessage` rather than returning success always. Second, we buffer failed sends when the error was not ours (validation), assuming the socket will be closed immediately and the recovery routine will kick off. When finished, we resend the buffered messages.

## Test Plan
- [ ] run all upstream CI a few times, try to write a test.

## Checklist

### DIFF
--- a/agent/internal/agent.go
+++ b/agent/internal/agent.go
@@ -52,7 +52,8 @@ type agent struct {
 	masterProto  string
 	masterClient *http.Client
 
-	reconnecting bool
+	reconnecting         bool
+	reconnectSendBacklog []interface{}
 }
 
 func newAgent(version string, options Options) *agent {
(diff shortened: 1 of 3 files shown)


agent/internal/agent.go
```diff
@@ -52,7 +52,8 @@ type agent struct {
 	masterProto  string
 	masterClient *http.Client
 
-	reconnecting bool
+	reconnecting         bool
+	reconnectSendBacklog []interface{}
```
ioga, line 56: maybe `[]api.WriteMessage`?

ioga, line 56: I wish we could have a "reliability" wrapper around the sockets so all the backlog retry logic was hidden away in there, and not be an intrinsic part of `agent` actor here.

master/pkg/actor/api/ws.go
```diff
@@ -32,6 +33,16 @@ const (
 	MaxWebsocketMessageSize = 128 * 1024 * 1024
 )
 
+// MaxMaxWebsocketMessageSizeExceededError indicates to the call that the message exceeded the
+// maximum websocket size.
+type MaxMaxWebsocketMessageSizeExceededError struct {
```
ioga, line 38:
```suggestion
type MaxWebsocketMessageSizeExceededError struct {
```

agent/internal/agent.go
stoksc: that is my aim in a follow up pr, but this one needs to make release it seems

agent/internal/agent.go
```diff
@@ -351,6 +377,13 @@ func (a *agent) attemptReconnect(ctx *actor.Context) bool {
 	for i := 0; i < aproto.AgentReconnectAttempts; i++ {
 		switch err := a.connect(ctx); {
 		case err == nil:
+			for msg := range a.reconnectSendBacklog {
+				if err = ctx.Ask(a.socket, msg).Error(); err != nil {
```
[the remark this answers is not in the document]
stoksc: we try to retry once, once the connection has been reestablished. if we fail again immediately, we just call it quits and fail.

agent/internal/agent.go
```diff
@@ -89,14 +90,34 @@ func (a *agent) Receive(ctx *actor.Context) error {
...
+			switch err := ctx.Ask(a.socket, wsm).Error(); err.(type) {
+			case nil:
+				return nil
+			case api.MaxMaxWebsocketMessageSizeExceededError:
+				ctx.Log().WithError(err).Error("container log too large")
+			default:
+				ctx.Log().WithError(err).Warnf("writing container log %v", msg)
+				a.bufferSendForRecovery(ctx, wsm)
+			}
+		} else {
```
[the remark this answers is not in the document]
stoksc: there is a way to run an agent standalone for testing, i think that's all this is.

stoksc:
@ioga going to just close this

PR #3418, opened 2022-01-11
chore: Move parts of Experiment CLI to GRPC API

mapmeld:
## Description

Related to DET-6353, update Experiment CLI features to use our newer GRPC API and Python bindings, instead of `core_experiment`. This cannot neatly replace all experiment-related features because: creating experiment expects additional params, reading trial is expected to return steps/metrics, core_experiment supports patch-ing resources and checkpoints which are not implemented in GRPC, downloading experiment not implemented.

This does change: activate, archive and unarchive, cancel, delete, describe, config, model-def, kill, list, list-trials, pause, name and description, labels; most of these share `patchExperiment` which we can trim down and replace with the specific GRPC endpoints.

todo: e2e_tests uses the old system to change experiment state; before merging I should update those and remove state from `patchExperiment` too.

## Checklist


hamidzr:
another potential approach for patch endpoints: https://determined-community.slack.com/archives/CV3MVAQJU/p1642034920030100?thread_ts=1641856211.020200&cid=CV3MVAQJU

master/internal/api_experiment.go
```diff
@@ -595,10 +595,6 @@ func (a *apiServer) PatchExperiment(
 				req.Experiment.Labels)
 		case path == "description":
 			exp.Description = req.Experiment.Description
-		case !strings.HasPrefix(path, "update_mask"):
```
hamidzr, line 598: reading [this](https://determined-community.slack.com/archives/CV3MVAQJU/p1642614972035700?thread_ts=1641856211.020200&cid=CV3MVAQJU) did you mean to leave this out?
I know it'll limit the changes you can make to the cli and prevent removing parts of the old patch endpoint. This works given/if we have some future plans to do option 1 or 2 or add new endpoints for what patch experiment does.

If you go forward with this could you update the endpoint comment in proto to include what's updatable and maybe that trying to patch other attributes won't fail as an extra warning?

harness/determined/common/api/experiment.py
```diff
@@ -27,7 +27,9 @@ def patch_experiment(master_url: str, exp_id: int, patch_doc: Dict[str, Any]) ->
 
 
 def activate_experiment(master_url: str, exp_id: int) -> None:
```
hamidzr, line 29: question: do we still need this with this change you've done https://github.com/determined-ai/determined/pull/3418/files?show-viewed-files=true&file-filters%5B%5D=#diff-46809115da97aa7d7f43e797b3b733525ec317cd5535494d6022d733acf06079R40

harness/determined/cli/experiment.py
```diff
@@ -416,33 +413,33 @@ def wait(args: Namespace) -> None:
...
         if args.all:
-            result.append(e["archived"])
+            result.append(e.archived)
         return result
 
     headers = [
         "ID",
         "Owner",
         "Name",
+        "Parent ID",
```
hamidzr, line 442: 👍🏼

harness/determined/common/api/experiment.py
hamidzr: I see in in create_experiment 👍🏼

hamidzr (approved):
great work.

master/internal/api_experiment.go
mapmeld: If I leave this part alone, sending the rest of the Experiment object in the request will cause an error here.

Is your suggestion a change to PatchExperimentRequest in proto/src/determined/api/v1/experiment.proto , we don't really say anything there about what fields to send

I think the other option would be changing `bindings.v1Experiment.to_json` to have a `to_patch_json` and only include patch-able fields?

hamidzr: > If I leave this part alone, sending the rest of the Experiment object in the request will cause an error here.

true. Maybe I could have put it better but my original assumption was that by option0 you're gonna leave out any changes to this but then considering that
> I know it'll limit the changes you can make to the cli and prevent removing parts of the old patch endpoint.

I agreed with going forward with it : )
> This works for me given/if we have some future plans to do option 1 or 2 or add new endpoints for what patch experiment does.

> Is your suggestion a change to PatchExperimentRequest in proto/src/determined/api/v1/experiment.proto , we don't really say anything there about what fields to send

I'm just suggesting adding more details to the comment here. if you go with the current approach you're taking clarifying a 200 doesn't mean their request was necessarily fulfilled.
```
  // Patch an experiment's fields.
  rpc PatchExperiment(PatchExperimentRequest)
```

> I think the other option would be changing bindings.v1Experiment.to_json to have a to_patch_json and only include patch-able fields?

that could be a good option but did you wanna leave discussing and addressing that for another pr? [ref](https://determined-community.slack.com/archives/CV3MVAQJU/p1642631609038600?thread_ts=1641856211.020200&cid=CV3MVAQJU)

PR #3425, opened 2022-01-11
feat: Record parent_id from WebUI, return forkedFrom from API

mapmeld:
## Description

- Fixes an issue where experiments forked and continued from the Web UI were stored with parent_id = NULL
- Returns parent_id in the experiments API

## Checklist

### DIFF
--- a/proto/src/determined/experiment/v1/experiment.proto
+++ b/proto/src/determined/experiment/v1/experiment.proto
@@ -4,6 +4,7 @@ package determined.experiment.v1;
 option go_package = "github.com/determined-ai/determined/proto/pkg/experimentv1";
 
 import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
 import "protoc-gen-swagger/options/annotations.proto";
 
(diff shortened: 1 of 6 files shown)


proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -85,6 +86,8 @@ message Experiment {
   string notes = 14;
   // Associated job's id.
   string job_id = 15;
+  // Original id of a forked or continued experiment.
+  google.protobuf.Int32Value forked_from = 16;
```
EmilyBonar, line 90: question: why does this need to be a `google.protobuf.Int32Value` and not just an `int32`?

EmilyBonar (approved):
Overall looks good to me!

proto/src/determined/experiment/v1/experiment.proto
mapmeld: I prefer the `google.protobuf.` type because it can be `null` if it's empty/blank, if it's int32 it would send a 0

PR #3427, opened 2022-01-11
fix: Add allocation state to db test object

mapmeld:
## Description

Quick fix for master branch, adding new fields to allocation into the db test

## Checklist

### DIFF
--- a/master/internal/db/postgres_tasks_intg_test.go
+++ b/master/internal/db/postgres_tasks_intg_test.go
@@ -85,6 +85,8 @@ func TestJobTaskAndAllocationAPI(t *testing.T) {
 		AgentLabel:   "something",
 		ResourcePool: "somethingelse",
 		StartTime:    time.Now().UTC().Truncate(time.Millisecond),
+		State:        1,
+		IsReady:      false,
 	}
 	err = db.AddAllocation(aIn)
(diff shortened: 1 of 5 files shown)


stoksc (commented):
It is probably more correct to call add allocation then update allocation before the get allocation db call in the test rather than changing add allocation to make the test pass

master/internal/db/postgres_tasks_intg_test.go
```diff
@@ -85,6 +85,8 @@ func TestJobTaskAndAllocationAPI(t *testing.T) {
 		AgentLabel:   "something",
 		ResourcePool: "somethingelse",
 		StartTime:    time.Now().UTC().Truncate(time.Millisecond),
+		State:        1,
+		IsReady:      false,
 	}
 	err = db.AddAllocation(aIn)
```
stoksc, line 91: what i meant was, we should call update allocation right below here, not change `AddAllocation`'s code add all and `model.Allocation.State` and `model.Allocation.IsReady` both need to be pointers since the column is null. sorry i didn't catch this in review.

stoksc (approved):

stoksc:
there is still some feedback from the original pr but we can land that separately if you can still do it.

PR #3428, opened 2022-01-11
fix: fix priority scheduler preemption and prioritization issues

hamidzr:
## Description

Fixes two separate issues when priority between tasks/jobs are equal in the priority scheduler and it has fall back on its secondary sorting logic. Assuming that we want the scheduler to prioritize the oldest user-submitted ENTbCS (aka jobs)

1. the sorting was based on the Allocation actor's registered time which is unstable, changed during different cluster events, and would easily go out of sync with "user submission time"
2. the scheduler picked the oldest user-submitted job to preempt when needed instead of the most recent submitted one

## Commentary (optional)

issue discussion, demo, and reproduction steps: https://determined-community.slack.com/archives/CV3MVAQJU/p1639766427055600

## Checklist

### DIFF
--- a/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
+++ b/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
@@ -2,4 +2,5 @@
 
 **Bug Fixes**
 
--  Master: Fix a bug in the priority scheduler where jobs with equal priority would be scheduled or preempted in an order not correctly respecting job submission time.
+-  Master: Fix a bug in the priority scheduler where jobs with equal priority would be scheduled or
+   preempted in an order not correctly respecting job submission time.
--- a/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
+++ b/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
@@ -2,5 +2,4 @@
 
 **Bug Fixes**
 
--  Master: Fixes two separate issues when priority between tasks/jobs is equal in priority scheduler
-   and it falls back on its secondary sorting logic.
+-  Master: Fix a bug in the priority scheduler where jobs with equal priority would be scheduled or preempted in an order not correctly respecting job submission time.
--- a/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
+++ b/docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  Master: Fixes two separate issues when priority between tasks/jobs is equal in priority scheduler
+   and it falls back on its secondary sorting logic.
--- a/master/internal/checkpoint_gc.go
+++ b/master/internal/checkpoint_gc.go
@@ -14,6 +14,7 @@ import (
 	"github.com/determined-ai/determined/master/pkg/tasks"
 )
 
+// TODO these are job related tasks as well.
 type checkpointGCTask struct {
 	rm *actor.Ref
 	db *db.PgDB
--- a/master/internal/checkpoint_gc.go
+++ b/master/internal/checkpoint_gc.go
@@ -14,7 +14,6 @@ import (
 	"github.com/determined-ai/determined/master/pkg/tasks"
 )
 
-// TODO these are job related tasks as well.
 type checkpointGCTask struct {
 	rm *actor.Ref
 	db *db.PgDB
--- a/master/internal/db/postgres_tasks_intg_test.go
+++ b/master/internal/db/postgres_tasks_intg_test.go
@@ -10,13 +10,12 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
-	"github.com/google/uuid"
-	"github.com/stretchr/testify/require"
-	"gopkg.in/guregu/null.v3"
-
 	"github.com/determined-ai/determined/master/pkg/etc"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/pkg/ptrs"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/guregu/null.v3"
 )
 
 // TestJobTaskAndAllocationAPI, in lieu of an ORM, ensures that the mappings into and out of the
--- a/master/internal/db/postgres_tasks_intg_test.go
+++ b/master/internal/db/postgres_tasks_intg_test.go
@@ -10,12 +10,13 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
-	"github.com/determined-ai/determined/master/pkg/etc"
-	"github.com/determined-ai/determined/master/pkg/model"
-	"github.com/determined-ai/determined/master/pkg/ptrs"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/guregu/null.v3"
+
+	"github.com/determined-ai/determined/master/pkg/etc"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/master/pkg/ptrs"
 )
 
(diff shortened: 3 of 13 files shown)


master/internal/db/postgres_tasks_intg_test.go
```diff
@@ -10,12 +10,13 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
```
hamidzr, line 12: revert this

master/internal/checkpoint_gc.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/determined-ai/determined/master/pkg/tasks"
 )
 
+// TODO these are job related tasks as well.
 type checkpointGCTask struct {
```
hamidzr, line 18: how do these get scheduled. what priority/weight do they take on? if I"m not mistaken they're 0 slot so they get scheduled separately

azhou-determined:
👍 tested this fix locally by submitting 4 experiments with priority 20, verified they are being run in the order of submission.

azhou-determined:
also modified some of the unit tests in `priority_test.go` here: https://github.com/azhou-determined/determined/tree/unit-tests-priority-schedule-job-time
can merge if it looks good

master/internal/resourcemanagers/priority.go
```diff
@@ -185,7 +185,8 @@ func trySchedulingTaskViaPreemption(
 	log.Debugf("trying to schedule task %s by preempting other tasks", allocationRequest.Name)
 
 	for priority := model.MaxUserSchedulingPriority; priority > allocationPriority; priority-- {
-		for _, preemptionCandidate := range priorityToScheduledTaskMap[priority] {
+		for i := len(priorityToScheduledTaskMap[priority]) - 1; i >= 0; i-- {
```
stoksc, line 188: idea here is we should preempt newest jobs first when trying to make room for high priority jobs?

master/internal/resourcemanagers/priority.go
```diff
@@ -244,8 +245,8 @@ func sortTasksByPriorityAndTimestamp(
 	priorityToPendingTasksMap := make(map[int][]*sproto.AllocateRequest)
 	priorityToScheduledTaskMap := make(map[int][]*sproto.AllocateRequest)
 
-	for it := taskList.iterator(); it.next(); {
-		req := it.value()
+	reqs := sortTasks(taskList, groups, false)
```
stoksc, line 248: i dont think we need this change if we also change the comparator passed to treeset.NewWith, so that the iterator is still in the correct order? if so, why this change? after a quick glance, that looks more correct for elsewhere the iterator is used (where it matters, and order is important).

master/internal/resourcemanagers/priority.go
hamidzr: yes an extension of "we prioritize jobs that have been submitted the longest" which I think was also the original intent?

master/internal/resourcemanagers/priority.go
hamidzr: yeah we shouldn't need to but I wanted to make sure there is less chance of deviation on how the jobs are sorted but I can take it out for this particular fix.

hamidzr: we could pass in the filter to  sortTasks too if that's a concern.

master/internal/experiment.go
```diff
@@ -444,8 +444,8 @@ func (e *experiment) processOperations(
 			state := trialSearcherState{Create: op, Complete: true}
 			e.TrialSearcherState[op.RequestID] = state
 			ctx.ActorOf(op.RequestID, newTrial(
-				trialTaskID(e.ID, op.RequestID), e.JobID, e.ID, e.State, state, e.rm, e.trialLogger, e.db,
-				config, checkpoint, e.taskSpec,
+				trialTaskID(e.ID, op.RequestID), e.JobID, e.StartTime, e.ID, e.State, state, e.rm,
```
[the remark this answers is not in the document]
hamidzr: what kind of arguments are you thinking it would encompass? would the trial keep it in memory that way too?

hamidzr:
> also modified some of the unit tests in `priority_test.go` here: https://github.com/azhou-determined/determined/tree/unit-tests-priority-schedule-job-time can merge if it looks good

thank you. I also cherry-picked the tests here

master/internal/resourcemanagers/scheduler_test.go
```diff
@@ -435,6 +438,18 @@ func assertEqualToAllocate(
 		"actual tasks and expected tasks must have the same length")
 }
 
+func assertEqualToAllocateOrdered(
```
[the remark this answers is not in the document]
hamidzr: I didn't write this but It seems to go off of `assertEqualToAllocate`. any name suggestions? these are the other ones we have. could probably be added as a parameter to `assertEqualToAllocate` too.

```
master/internal/resourcemanagers/scheduler_test.go:424:func assertEqualToAllocate(
master/internal/resourcemanagers/scheduler_test.go:441:func assertEqualToAllocateOrdered(
master/internal/resourcemanagers/scheduler_test.go:453:func assertEqualToRelease(
master/internal/resourcemanagers/provisioner/scale_decider_test.go:13:func assertEqualInstancesMarked(t
master/pkg/schemas/merge.go:46:func assertTypeMatch(obj
master/test/integration/api/api_trials_test.go:605:func assertStringContains(t

```

master/internal/experiment.go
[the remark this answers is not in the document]
stoksc: this was just me pointing out this function has 12 params. let's handle it later.

master/internal/resourcemanagers/priority.go
stoksc: makes sense.

master/internal/sproto/task.go
```diff
@@ -15,13 +15,15 @@ type (
...
-		TaskID       model.TaskID
-		JobID        *model.JobID
-		State        job.SchedulingState
-		Name         string
-		TaskActor    *actor.Ref
-		Group        *actor.Ref
+		AllocationID      model.AllocationID
+		TaskID            model.TaskID
+		JobID             *model.JobID
+		JobSubmissionTime *time.Time // TODO or a job info struct
```
stoksc, line 21: i think we should either cleanup these todos or make a ticket for them

docs/release-notes/3428-fix-priority-scheduler-timelogic.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  Master: Fixes two separate issues when priority between tasks/jobs is equal in priority scheduler
+   and it falls back on its secondary sorting logic.
```
stoksc, line 6:
```suggestion
-  Master: Fix a bug in the priority scheduler where jobs with equal priority would be scheduled or preempted in an order not correctly respecting job submission time.
```

master/internal/resourcemanagers/priority.go
stoksc, line 188: nit golang, why no slice.Reverse...

stoksc (approved):
this by itself is fine but i still think we might should change `master/internal/resourcemanagers/task_list.go`'s `taskComparator` to use `compareByRegisteredTime` rather than also still using registered time.

master/internal/sproto/task.go
hamidzr: since I didn't have clear stories for addressing them I just cleaned them up

master/internal/resourcemanagers/priority.go
hamidzr: sort.Reverse? I didn't know about it. and that seems to not modify the array in place which we can't have.

hamidzr:
> this by itself is fine but i still think we might should change `master/internal/resourcemanagers/task_list.go`'s `taskComparator` to use `compareByRegisteredTime` rather than also still using registered time.

makes sense. I'll update that after this lands

master/internal/resourcemanagers/priority.go
stoksc: Oh I mean nit the go language lol

PR #3433, opened 2022-01-12
ci: update gauge version [DET-6432]

hkang1:
## Description

UPDATE: on the 9th run of the test, it failed too. :(

Problem of webui-e2e-tests flaking around the Authentication section. Only recently observed...
```
# Authentication
  ## Sign out	 ✔
  ## Sign in as user without password	Error: TextBox[id="login_username"]is not visible
at /Users/user/Projects/determined/webui/tests/node_modules/taiko/lib/actions/pageActionChecks.js:189:11
at waitAndGetActionableElement (/Users/user/Projects/determined/webui/tests/node_modules/taiko/lib/actions/pageActionChecks.js:156:3)
at focus (/Users/user/Projects/determined/webui/tests/node_modules/taiko/lib/actions/focus.js:12:12)
at Object.module.exports.focus (/Users/user/Projects/determined/webui/tests/node_modules/taiko/lib/taiko.js:1038:16)
at Object.module.exports.<computed> [as focus] (/Users/user/Projects/determined/webui/tests/node_modules/taiko/lib/taiko.js:2612:14)

        Error Message: Runner is not Alive
        Stacktrace: 
        
        Failed Step: Sign in as "user-wo-pw" without password
        Specification: specs/02-authentication.spec:12
        Error Message: transport is closing
        Stacktrace: 
        
    Error Message: Runner is not Alive
    Stacktrace: 
    
  Error Message: Runner is not Alive
  Stacktrace: 
  
  Error Message: Runner is not Alive
  Stacktrace: 
```

Solution:
We were only able to narrow it down to possibly taiko's `textBox` API causing the issue. Replaced it with a more generic selector.

## Checklist


hamidzr:
makes sense to update gauge

hamidzr (approved):
should we change the pr to "update gauge" ? 
if the textBox change hasn't resolved the flake do you still want to include it?

hkang1:
> should we change the pr to "update gauge" ? if the textBox change hasn't resolved the flake do you still want to include it?

good call, I've changed the PR title.

So the weird thing is (and this might just be in my head) but with the `$(#id)` approach, seems to have reduced the frequency of the flake (1/9) vs (3/6 original). But it's probably just by chance.

hamidzr:
`$()` probably has less library code compared to `textBox` so that could have an effect. if you wanna keep it in that works too

PR #3439, opened 2022-01-12
fix: update to show full version when nav sidebar is not collapsed [DET-6264]

gt2345:
## Description

Update nav sidebar to display full version when not collapsed

## Test Plan

Manual tested with long and short version


hkang1 (approved):
This looks great! Thanks for fixing this.

PR #3448, opened 2022-01-13
chore: update checkpoint export apis [DET-6037]

rb-determined-ai:
## Description

* Deprecate some old accidentally-public Checkpoint methods.
* Introduce framework-specific alternatives to Checkpoint.load(), and deprecate Checkpoint.load()

## Test Plan

* I still need unit tests, but presently the data model doesn't really lend itself to unit tests, since `metadata.json` is required for loading a checkpoint, but the file is only created by fetching metadata from the master in `Checkpoint.download()`


harness/determined/_core/_checkpointing.py
```diff
@@ -60,6 +62,9 @@ def store_path(self, metadata: Optional[Dict[str, Any]] = None) -> Iterator[Tupl
 
         with self._storage_manager.store_path() as (uuid, path):
             yield uuid, path
+            # XXX: Do we break the format of metadata.json now?  Or later?
```
ioga, line 65: is this still breaking anything?..

ioga (approved):

ioga:
does this API change need a release note? or maybe an entry in the docs?

rb-determined-ai:
@ioga yes it sure does.  I totally forgot to check on the docs situation.

rb-determined-ai:
@ioga yes it sure does.  I totally forgot to check on the docs situation.

harness/determined/_core/_checkpointing.py
rb-determined-ai: So, right now the format of `metadata.json` is something like:
```json
{
  "determined-version": "0.17.6",
  "framework": "tensorflow",
  "format": "saved-weights",
  "metadata": {
    "user-key": "user-value"
  }
}
```

But in the future I think we need to lose the distinction between users's metadata and the "system" metadata (since the Core API is slightly altering what counts as part of the "system"):
```json
{
  "determined-version": "0.17.6",
  "framework": "tensorflow",
  "format": "saved-weights",
  "user-key": "user-value"
}
```

And since `metadata.json::metadata` is purely user-values, I have to assume that this counts as a breaking change.

I think what I settled on was that we can introduce `load_data.json` now and break the `metadata.json` later; I don't see that those changes would need to happen in the same version.  But I wanted to call it out as it is something we had not previously discussed in slack.

gh-determined-ai (approved):
Not currently reviewing codeblocks or docstrings.
Need a Release Notes template to review release notes. In the meantime, good as-is.

PR #3325, opened 2021-12-15
fix: stop profiler spinner when terminal [DET-6326]

hkang1:
## Description

When a trial with profiler enabled reaches terminal state, the profiler continues to show a spinner. Update to stop the spinner when the trial reaches a terminal state.

The single trial experiment also shows and flickers `no data available` incorrectly when loading the page tabs, including the profiler tab. Update the `empty` and `loading` logic for the profiler.

UPDATE: added specialized skeletons and corresponding storybook stories

Screen Shot 2021-12-21 at 6 55 03 PM
Screen Shot 2021-12-21 at 6 55 07 PM
Screen Shot 2021-12-21 at 6 55 09 PM

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3453, opened 2022-01-18
fix: avoid terminating profiler streaming [DET-6459]

hkang1:
## Description

[Relevant PR that introduced the error](https://github.com/determined-ai/determined/pull/3325)

The abort controller was created as a state and once canceled the all future profiler calls don't happen. This included changes to the filters which needs to trigger a new call with different filtering parameters.

Instead, when a trial reaches a terminal state, change the `follow` (indicator of whether to stream or just GET once) to `false`.

## Checklist


EmilyBonar (approved):
LGTM

PR #3454, opened 2022-01-18
fix: Update sample models code for Determined client

mapmeld:
## Description

- Adds `client.get_model_by_name("ABCDE")` to Determined client
- Fix non-working code sample on `/det/models/{model_id}/versions/{version_id}/overview`

## Checklist


EmilyBonar (approved):
LGTM!

PR #3461, opened 2022-01-19
fix: jq error changing ntbcs priority

eecsliu:
## Description
Because notebooks, tensorboards, commands, and shells are not persisted, we'll prevent users from modifying them for now and instead ask them to terminate the instances if they wish to modify the priority.

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -243,7 +243,7 @@ func (c *command) Receive(ctx *actor.Context) error {
 		ctx.Respond(&apiv1.KillNotebookResponse{Notebook: c.toNotebook(ctx)})
 		c.clearJobInfo()
 	case *apiv1.SetNotebookPriorityRequest:
-		c.setPriority(ctx, int(msg.Priority))
+		_ = c.setPriority(ctx, int(msg.Priority))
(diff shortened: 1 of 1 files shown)


eecsliu:
There are other priority change messages being addressed here that are not affected by the job queue. Should I catch those as well?

eecsliu:
This is the cleanest way of erroring out on submission because any deeper and we lose information on whether or not the job is an experiment or command.

hamidzr (changes requested):
Thanks for putting this up. I added some comments below

- messaging: "not supported" instead of unable to? and a smaller comment on whether we want to include instructions on what they should do as an api response.
- it'd be more changes but we might want to return an error and leave responding to the caller as it would probably double respond this way
```
	case *apiv1.SetNotebookPriorityRequest:
		c.setPriority(ctx, int(msg.Priority))
		ctx.Respond(&apiv1.SetNotebookPriorityResponse{Notebook: c.toNotebook(ctx)})
```
- as is it'll still go ahead and set the priority, notify the rm, etc

eecsliu:
my bad, didn't test after the refactor. Refactoring again

hamidzr (approved):

PR #3462, opened 2022-01-19
fix: use different tf on macos

mpkouznetsov:
## Description

Running `make all` on Mac results in
```
ERROR: Could not find a version that satisfies the requirement tensorflow==2.4.3 (from versions: none)
ERROR: No matching distribution found for tensorflow==2.4.3
make[1]: *** [get-deps-pip] Error 1
make: *** [all] Error 2
```

The problem is tensorflow-macos does not exist for 2.4.3; it only begins with 2.5.0.

## Test Plan

Have new users ( @johnkim-det ) try it :)

## Checklist

### DIFF
--- a/e2e_tests/tests/requirements.txt
+++ b/e2e_tests/tests/requirements.txt
@@ -5,8 +5,8 @@ pytest-timeout
 pexpect
 torch==1.9.0
 torchvision==0.10.0
-tensorflow==2.4.3; sys_platform != 'darwin'
-tensorflow-macos==2.7.0; sys_platform == 'darwin'
+tensorflow==2.4.3; sys_platform != 'darwin' or platform_machine != 'arm64'
+tensorflow-macos==2.7.0; sys_platform == 'darwin' and platform_machine == 'arm64'
 pandas
 pyyaml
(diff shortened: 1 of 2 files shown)


e2e_tests/tests/requirements.txt
```diff
@@ -5,7 +5,8 @@ pytest-timeout
 pexpect
 torch==1.9.0
 torchvision==0.10.0
-tensorflow==2.4.3
+tensorflow==2.4.3; sys_platform != 'darwin'
+tensorflow-macos==2.7.0; sys_platform == 'darwin'
```
ioga, line 9: This is cool! I didn't know `requirements.txt` could do that.

I was trying it out and it seems `tensorflow-macos` for x86_64 has only a build for py38. Should we perhaps do this for arm64 only, i.e.:
```suggestion
tensorflow==2.4.3; sys_platform != 'darwin' or platform_machine != 'arm64'
tensorflow-macos==2.7.0; sys_platform == 'darwin' and platform_machine == 'arm64'
```

ioga (approved):

PR #3464, opened 2022-01-19
chore: associate gc tasks with their job

hamidzr:
## Description

based on the idea that all our current tasks are associated with a user-submitted job.

## Test Plan

check that jobid and jobsubmission are set wherever we create an allocate request outside of tests
```
hrz ➜ determined git:(master) grep --exclude=\*_test.go 'allocaterequest{' -R . 2>/dev/null
./master/internal/command/command.go:		allocation := task.NewAllocation(sproto.AllocateRequest{
./master/internal/checkpoint_gc.go:		t.task = &sproto.AllocateRequest{
./master/internal/resourcemanagers/priority.go:	for _, tasksMap := range []map[int][]*sproto.AllocateRequest{
./master/internal/trial.go:	t.allocation, _ = ctx.ActorOf(t.runID, taskAllocator(sproto.AllocateRequest{
```

run the clutster with an assertion https://github.com/determined-ai/determined/pull/3464/commits/593f97f38105e45bd4abe05114879bc2cbf8b012

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/resource_pool.go
+++ b/master/internal/resourcemanagers/resource_pool.go
@@ -90,10 +90,7 @@ func (rp *ResourcePool) addTask(ctx *actor.Context, msg sproto.AllocateRequest) 
 	if len(msg.AllocationID) == 0 {
 		msg.AllocationID = model.AllocationID(uuid.New().String())
 	}
-	if msg.Group == nil {
-		msg.Group = msg.TaskActor
-	}
-	rp.getOrCreateGroup(ctx, msg.Group)
+	rp.getOrCreateGroup(ctx, msg.GroupID())
 	if len(msg.Name) == 0 {
 		msg.Name = "Unnamed Task"
 	}
--- a/master/internal/resourcemanagers/resource_pool.go
+++ b/master/internal/resourcemanagers/resource_pool.go
@@ -90,7 +90,10 @@ func (rp *ResourcePool) addTask(ctx *actor.Context, msg sproto.AllocateRequest) 
 	if len(msg.AllocationID) == 0 {
 		msg.AllocationID = model.AllocationID(uuid.New().String())
 	}
-	rp.getOrCreateGroup(ctx, msg.GroupID())
+	if msg.Group == nil {
+		msg.Group = msg.TaskActor
+	}
+	rp.getOrCreateGroup(ctx, msg.Group)
 	if len(msg.Name) == 0 {
 		msg.Name = "Unnamed Task"
 	}
--- a/master/internal/resourcemanagers/scheduler_test.go
+++ b/master/internal/resourcemanagers/scheduler_test.go
@@ -35,7 +35,7 @@ func (g *mockGroup) Receive(ctx *actor.Context) error {
 	switch ctx.Message().(type) {
 	case actor.PreStart:
 	case actor.PostStop:
-	case job.RMJobInfo:
+	case *job.RMJobInfo:
 	default:
 		return actor.ErrUnexpectedMessage(ctx)
 	}
@@ -80,9 +80,7 @@ func (t *mockTask) Receive(ctx *actor.Context) error {
 			ResourcePool: t.resourcePool,
 			TaskActor:    ctx.Self(),
 		}
-		if t.group == nil {
-			task.Group = ctx.Self()
-		} else {
+		if t.group != nil {
 			task.Group = ctx.Self().System().Get(actor.Addr(t.group.id))
 		}
 		if ctx.ExpectingResponse() {
@@ -273,7 +271,6 @@ func forceAddTask(
 	req := &sproto.AllocateRequest{
 		AllocationID: model.AllocationID(taskID),
 		TaskActor:    ref,
-		Group:        ref,
 		SlotsNeeded:  slotsNeeded,
 	}
 	taskList.AddTask(req)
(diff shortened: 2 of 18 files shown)


master/internal/sproto/task.go
```diff
@@ -166,3 +167,19 @@ type Event struct {
...
+		return r.Group
+	}
+	return r.TaskActor
+}
+
+// UserVisible determines whether the AllocateRequest should
+// be considered in user-visible reports.
+// AKA HasUserVisibleActor.
+func (r AllocateRequest) UserVisible() bool {
+	return r.Group != nil // CHECK we might want to define a new property for this
```
stoksc, line 183: yeah i may just make this its own thing. like, a checkpoint gc task for an experiment seems like it should be in the experiment group but not user visible? right? idk, could go either and not make a huge difference but that seems more consistent

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -90,10 +90,7 @@ func (rp *ResourcePool) addTask(ctx *actor.Context, msg sproto.AllocateRequest)
 	if len(msg.AllocationID) == 0 {
 		msg.AllocationID = model.AllocationID(uuid.New().String())
 	}
-	if msg.Group == nil {
-		msg.Group = msg.TaskActor
-	}
-	rp.getOrCreateGroup(ctx, msg.Group)
+	rp.getOrCreateGroup(ctx, msg.GroupID())
```
stoksc, line 93: i would just make user visible its own thing rather than trying to have cases where `msg.Group == nil` encode special meaning. the surface area/risk of this change is pretty large from this, since now you have to check `msg.GroupID()` is used or `nil` is OK in every place it was used.

stoksc: i think it makes more sense for group to always be non-nil too. nothing should have a nil group.

stoksc: seems like the other way this change is +/- 10 or 20 lines at most

master/internal/resourcemanagers/scheduler_test.go
```diff
@@ -354,15 +355,16 @@ func setupSchedulerStates(
...
 		if mockTask.jobID != "" {
 			jid := model.JobID(mockTask.jobID)
-			jobID = &jid
+			jobID = jid
 		}
 
 		req := &sproto.AllocateRequest{
 			AllocationID:      mockTask.id,
 			JobID:             jobID,
+			IsUserVisible:     true,
```
hamidzr, line 367: I have a small refactor in another pr to reuse how we go from mockTask to AllocateRequest

master/internal/sproto/task.go
hamidzr: I agree I went ahead and updated it.

master/internal/resourcemanagers/resource_pool.go
hamidzr: makes sense. 
as is it's used as two things 1. an identifier for finding the group (doesn't necessarily need to be actor ref) 2. (new use case with the addition of jobs)  used to contact the job actor
I probably want to do some work later to change number 2. we're probably better off finding/messaging the job actor through `job/job.go`

master/internal/resourcemanagers/task_list.go
```diff
@@ -122,19 +122,10 @@ func registerTimeComparator(t1 *sproto.AllocateRequest, t2 *sproto.AllocateReque
...
-		return -1
-	}
-	if a.JobSubmissionTime.Equal(*b.JobSubmissionTime) {
-		return registerTimeComparator(a, b)
-	}
-	if a.JobSubmissionTime.Before(*b.JobSubmissionTime) {
+	if a.JobSubmissionTime.Before(b.JobSubmissionTime) {
 		return -1
 	}
 	return 1
```
stoksc, line 131: random: at this point, why would this not be 0?

master/internal/sproto/task.go
```diff
@@ -17,10 +17,13 @@ type (
...
-		JobID             *model.JobID
-		JobSubmissionTime *time.Time
-		State             job.SchedulingState
-		Name              string
+		JobID             model.JobID
+		JobSubmissionTime time.Time
+		// IsUserVisible determines whether the AllocateRequest should
+		// be considered in user-visible reports.
+		IsUserVisible bool
+		State         job.SchedulingState
```
stoksc, line 25: note: i didn't notice until now this state was added. why is it on the allocation request, shouldn't the request be immutable basically? looking through the code, this seems wrong. i think taskList needs some repr of it's side of things rather than updating this request object. also, to not it there is a lot of overlap between job.SchedulingState and model.AllocationState and we should figure out where we're going with that

stoksc (approved):
pr lgtm, but i left some meta comments we should dig in to

master/internal/sproto/task.go
hamidzr: there's probably a better way of tracking that. we look at the resource assignment info when an assignment happens to set this. (minus backfilling)  we could have tracked this separately on a job_list struct or by looking up the assignment on each read. I'd need to think if we can track backfill status with the latter.

hamidzr: > also, to not it there is a lot of overlap between job.SchedulingState and model.AllocationState and we should figure out where we're going with that.

👍🏼  I'd definitely chat more before making changes to the schedulingstate to see what you think. I see scheduling as a higher level state but helps to see where they overlap and where they don't

PR #3465, opened 2022-01-19
fix: Allow allocations array to be empty without null value

mapmeld:
Fixes an issue in the new GRPC Tasks API, so that before a task gets an allocation, the Task can return an empty array  of Allocations instead of a null (causes a SQL/proto error)

### DIFF
--- a/master/static/srv/get_task.sql
+++ b/master/static/srv/get_task.sql
@@ -1,11 +1,8 @@
-WITH allo AS (
+SELECT tasks.task_id, (SELECT coalesce(jsonb_agg(allo), '[]'::jsonb) FROM (
   SELECT allocation_id, task_id, state, is_ready, start_time, end_time
   FROM allocations
-  WHERE task_id = $1
+  WHERE task_id = tasks.task_id
   ORDER BY end_time DESC NULLS FIRST
-)
-SELECT tasks.task_id, json_build_array(allo) AS allocations
+) allo) AS allocations
 FROM tasks
-LEFT JOIN allo ON tasks.task_id = allo.task_id
-WHERE tasks.task_id = $1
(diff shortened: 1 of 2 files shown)


master/static/srv/get_task.sql
```diff
@@ -1,11 +1,8 @@
-WITH allo AS (
+SELECT tasks.task_id, (SELECT coalesce(jsonb_agg(allo), '[]'::jsonb) FROM (
   SELECT allocation_id, task_id, state, is_ready, start_time, end_time
   FROM allocations
-  WHERE task_id = $1
+  WHERE task_id = tasks.task_id
   ORDER BY end_time DESC NULLS FIRST
```
dzhu, line 5: The ordering should be inside the `jsonb_agg` call.

master/static/srv/get_task.sql
```diff
@@ -1,11 +1,8 @@
-WITH allo AS (
+SELECT tasks.task_id, (SELECT coalesce(jsonb_agg(allo), '[]'::jsonb) FROM (
   SELECT allocation_id, task_id, state, is_ready, start_time, end_time
   FROM allocations
-  WHERE task_id = $1
+  WHERE task_id = tasks.task_id
```
dzhu, line 4: Perhaps
```suggestion
  WHERE allocations.task_id = tasks.task_id
```
just to be explicit.

dzhu (approved):

master/static/srv/get_task.sql
mapmeld: 👍

master/static/srv/get_task.sql
mapmeld: 👍

PR #3466, opened 2022-01-19
chore: improve fairshare scheduler for heterogeneous agents, improve tests.

ioga:
## Description

- Restructure `fairshareScheduler` to closer match the `priorityScheduler`.
- Maintain internal agent state to make it aware of its own scheduling fits to avoid the cases where there're enough slots for multiple tasks, but no fits for all of them. That may happen with heterogeneous agents. See `TestFairShare884435Sched88` test case.
- Enable scheduler tests to have tasks scheduled across multiple agents.
- Improve assertion messages for some scheduling checks to make it more evident which particular test or part of a test failed.

## Test Plan

No changes should be observed for the standard homogeneous agents setup.

## Commentary (optional)

Initially, I've set out to fix a problem I've seen with fairshare scheduling of (2 slot, 2 slot, 1 slot) tasks sequentially on (2 slot, 1 slot) agents that I've seen in December, but when I tried to repro it now or make a test for it, I could not do that, mysteriously. 
Perhaps something more complex has been happening there, but I can't spend to much time on that. So at least lets improve testing a bit.

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/fair_share.go
+++ b/master/internal/resourcemanagers/fair_share.go
@@ -11,13 +11,13 @@ import (
 	"github.com/determined-ai/determined/master/pkg/model"
 )
 
-type fairShare struct{}
+type fairShareScheduler struct{}
 
 // NewFairShareScheduler creates a new scheduler that schedules tasks according to the max-min
 // fairness of groups. For groups that are above their fair share, the scheduler requests
 // them to terminate their idle tasks until they have achieved their fair share.
 func NewFairShareScheduler() Scheduler {
-	return &fairShare{}
+	return &fairShareScheduler{}
 }
 
 type groupState struct {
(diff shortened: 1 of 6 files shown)


master/internal/resourcemanagers/fair_share.go
```diff
@@ -344,9 +251,11 @@ func assignTasks(
 			state.offered -= state.activeSlots
 			for _, req := range state.pendingReqs {
 				if req.SlotsNeeded <= state.offered {
-					if fits := findFits(req, agents, fittingMethod); len(fits) == 0 {
+					fits := findFits(req, agents, fittingMethod)
+					if len(fits) == 0 {
 						continue
 					}
+					addTaskToAgents(fits)
```
stoksc, line 258: so, this is our copy of agent state and we track our fits as we go, so we are aware of previous fits when allocating slot offers? as i understand it, `TestFairShare884435Sched88`'s only issue before was that fairshare made slot offers it couldn't possible fulfill, but that should've been fine because they're just slot offers and the fitting phase in `ResourcePool.allocateResources` would've fit them the same as this, right? what the added benefit is here?

hamidzr (commented):
test changes lgtm. I have yet to read through the scheduler changes.

hamidzr:
I'm sorry don't think I'll find time to review in more detail before my OOO, (i should try to wrap up something else). lmk if there is a section you think I need to see. (or if I should make time)

ioga:
to be revisited.

PR #3468, opened 2022-01-19
build: fix make -C harness clean

rb-determined-ai:
This also fixes make devcluster, which invokes:

    make -C harness clean build

as one of its startup steps.

### DIFF
--- a/harness/Makefile
+++ b/harness/Makefile
@@ -1,7 +1,7 @@
 py_bindings_dest=determined/common/api/bindings.py
 
 .PHONY: build
-build: gen $(py_bindings_dest)
+build: gen
 	python -W ignore:Normalizing:UserWarning:setuptools.dist setup.py -q bdist_wheel
 
 .PHONY: publish
@@ -43,7 +43,6 @@ clean: ungen
 	rm -rf pip-wheel-metadata/
 	rm -rf dist/
 	rm -rf build/
-	rm $(py_bindings_dest)
 	find . \( -name __pycache__ -o -name \*.pyc \) -delete
 
 AWS_TEMPLATES_PATH := determined/deploy/aws/templates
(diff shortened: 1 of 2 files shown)


harness/Makefile
```diff
@@ -1,7 +1,10 @@
-py_bindings_dest=determined/common/api/bindings.py
+PY_BINDINGS_DEST=determined/common/api/bindings.py
+PY_SCHEMAS_DEST=determined/common/schemas/expconf/_gen.py
+SCHEMA_INPUTS = ../schemas/gen.py $(shell find ../schemas/expconf -name '*.json')
+GENERATED_COMMITTED := $(PY_SCHEMAS_DEST) $(PY_BINDINGS_DEST)
 
 .PHONY: build
-build: gen $(py_bindings_dest)
+build: gen
```
hamidzr, line 7: question: is this removed dependency captured elsewhere? (other than maybe the higher level make file)

rb-determined-ai: `gen` depends on `GENERATED_COMMITTED`, which should by definition include the generated python bindings

rb-determined-ai:
shoot, with this change, `make -C harness check` effectively depends on `make -C bindings build/python` working, which depends on `make -C proto build` working, which I think it won't in our current circleci configuration.  This might take more work/thought than I realized going in.

hamidzr:
> shoot, with this change, `make -C harness check` effectively depends on `make -C bindings build/python` working, which depends on `make -C proto build` working, which I think it won't in our current circleci configuration. This might take more work/thought than I realized going in.

yeah that goes into go land but we can probably use the circleci `persist_path`  and `attach` directives for this if needed

rb-determined-ai:
Looks like we already have a bindings-specific check.  Now I let the `bindings` module that does the building and the checking also do the cleaning.  That works well enough for now.

hamidzr (approved):

PR #3470, opened 2022-01-19
feat: jq allow exp change RPs

eecsliu:
## Description
Support experiments switching resource pools.

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -623,7 +623,7 @@ func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
 	}
 
 	if _, err := sproto.GetResourcePool(ctx.Self().System(), msg.ResourcePool, 0, false); err != nil {
-		return fmt.Errorf("invalid resource pool name")
+		return fmt.Errorf("invalid resource pool name %s", msg.ResourcePool)
 	}
 
 	resources := e.Config.Resources()
@@ -631,11 +631,10 @@ func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
 	e.Config.SetResources(resources)
 
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
-		return errors.Wrapf(err, "setting experiment %s RP to %s", e.ID, msg.ResourcePool)
+		return errors.Wrapf(err, "setting experiment %d RP to %s", e.ID, msg.ResourcePool)
 	}
-	for _, ref := range ctx.Children() {
-		ctx.Tell(ref, sproto.ChangeRP{ResourcePool: msg.ResourcePool})
-	}
+	ctx.TellAll(sproto.ChangeRP{ResourcePool: msg.ResourcePool}, ctx.Children()...)
+
 	return nil
 }
 
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -277,6 +277,13 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 	case job.GetJobSummary:
 		ctx.Respond(e.toV1Job().Summary)
 
+	case job.SetRP:
+		resources := e.Config.Resources()
+		resources.SetResourcePool(msg.ResourcePool)
+		// TODO: check if the RP exists
+		e.Config.SetResources(resources)
+		ctx.Tell(e.rm, job.SetRP{ResourcePool: msg.ResourcePool, Handler: ctx.Self()})
+
 	// Experiment shutdown logic.
 	case actor.PostStop:
 		if err := e.db.SaveExperimentProgress(e.ID, nil); err != nil {
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -278,14 +278,9 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 		ctx.Respond(e.toV1Job().Summary)
 
 	case job.SetRP:
-		resources := e.Config.Resources()
-		resources.SetResourcePool(msg.ResourcePool)
-		// TODO: check if the RP exists
-		e.Config.SetResources(resources)
-		for _, ref := range ctx.Children() {
-			ctx.Tell(ref, msg)
+		if err := e.setRP(ctx, msg); err != nil {
+			return err
 		}
-		ctx.Tell(e.rm, job.SetRP{ResourcePool: msg.ResourcePool, Handler: ctx.Self()})
 
 	// Experiment shutdown logic.
 	case actor.PostStop:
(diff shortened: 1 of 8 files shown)


hamidzr:
would it make sense to have the pr target be the master branch?

master/internal/job/job.go
```diff
@@ -60,6 +60,11 @@ type (
 		ResourcePool string
 		Handler      *actor.Ref
 	}
+	// ChangeRP switches the resource pool that the job belongs to
```
hamidzr, line 63: nit full ResourcePool?

master/internal/experiment.go
```diff
@@ -280,6 +281,11 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 	case job.GetJobSummary:
 		ctx.Respond(e.toV1Job().Summary)
 
+	case job.SetRP:
```
hamidzr, line 284: ditto

master/internal/experiment.go
```diff
@@ -280,6 +281,11 @@ func (e *experiment) Receive(ctx *actor.Context) error {
...
+		if err := e.setRP(ctx, msg); err != nil {
+			return err
```
hamidzr, line 286: returning an error I think will kill the actor. do we want this? when setting priority and the like we don't do this.

master/internal/experiment.go
```diff
@@ -633,6 +639,29 @@ func (e *experiment) setOrder(ctx *actor.Context, queuePosition float64) error {
 	return nil
 }
 
+func (e *experiment) setRP(ctx *actor.Context, msg job.SetRP) error {
+	if sproto.UseK8sRM(ctx.Self().System()) {
+		ctx.Respond(fmt.Errorf("kubernetes does not support setting resource pools"))
+	}
```
hamidzr, line 645: break the execution flow?
also return the error and leave responding to the `Receive` method? if that's how we do it elsewhere.

hamidzr: in this method sometimes we respond and sometimes return an error we'd probably want one.

master/internal/experiment.go
```diff
@@ -633,6 +639,29 @@ func (e *experiment) setOrder(ctx *actor.Context, queuePosition float64) error {
...
+
+	resources := e.Config.Resources()
+	resources.SetResourcePool(msg.ResourcePool)
+	e.Config.SetResources(resources)
+
+	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
+		return errors.Wrapf(err, "setting experiment %s RP to %s", e.ID, msg.ResourcePool)
+	}
+	for _, ref := range ctx.Children() {
```
hamidzr, line 659: I think we have an actor method for sending the same message to a list of refs. that might help us wait for the responses on all of them as well before returning

master/internal/trial.go
```diff
@@ -206,6 +210,7 @@ func (t *trial) maybeAllocateTask(ctx *actor.Context) error {
 	}
 
 	ctx.Log().Info("decided to allocate trial")
+	t.runID += 1
```
hamidzr, line 213: question: is this a separate bugfix?

master/internal/experiment.go
eecsliu: we should only return error if there is a database issue - otherwise we do a ctx.Response from inside the setRP function

master/internal/experiment.go
eecsliu: will fix

master/internal/trial.go
eecsliu: yes, but this PR is dependent on it

master/internal/experiment.go
eecsliu: fixed!

master/internal/experiment.go
eecsliu: somewhat fixed - I think each error caught here is worthy of being stopped for on its own.

hamidzr: hmm I'm not sure I understand. what sort of error you're seeing this catch?
I was thinking about `askAll` or `getAll` I'm not sure how they handled error responses

master/internal/trial.go
hamidzr: great. if it's user-facing we might want a small mention release notes.

master/internal/experiment.go
```diff
@@ -612,6 +617,28 @@ func (e *experiment) setWeight(ctx *actor.Context, weight float64) error {
 	return nil
 }
 
+func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
+	if sproto.UseK8sRM(ctx.Self().System()) {
+		return fmt.Errorf("kubernetes does not support setting resource pools")
+	}
+
+	if _, err := sproto.GetResourcePool(ctx.Self().System(), msg.ResourcePool, 0, false); err != nil {
+		return fmt.Errorf("invalid resource pool name")
```
hamidzr, line 626: nit: include the invalid resource pool name, ideally in quotes

hamidzr (commented):
it looks good to me just some final small comments

I tried it out through the cli on local deployment

image


do we want release notes with this one or should we leave it for when we land the webui for it?

hamidzr (approved):
great work

master/internal/task/allocation.go
```diff
@@ -172,7 +172,7 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 		} else {
 			ctx.Respond(errors.New(fmt.Sprintf("unknown container %s", msg.ContainerID)))
 		}
-	case sproto.ReleaseResources:
+	case sproto.ReleaseResources, sproto.ChangeRP:
```
stoksc, line 175: nit: semantically, the allocation isn't really changing RPs. I think it trial.go should just send it a `task.Terminate` signal.

master/internal/experiment.go
```diff
@@ -612,6 +617,27 @@ func (e *experiment) setWeight(ctx *actor.Context, weight float64) error {
...
+		return fmt.Errorf("invalid resource pool name %s", msg.ResourcePool)
+	}
+
+	resources := e.Config.Resources()
+	resources.SetResourcePool(msg.ResourcePool)
+	e.Config.SetResources(resources)
+
+	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
+		return errors.Wrapf(err, "setting experiment %d RP to %s", e.ID, msg.ResourcePool)
+	}
```
stoksc, line 635: in this case, the db errs and 
* we respond to the user with the error
* we don't log the error
* we leave the RP change as is in memory
* we don't tell the change RP message to trials

if asha comes along and decides to launch a new trial, it will get launched into the RP indicate by memory, which is different than the config that will appear to users at that point.

maybe we should have the db errs and
* we respond to the user with the error
* we log the error
* we rollback the RP change in memory
* we don't tell the change RP message to trials

or we could just tells the current trials and log that we were unable to persist the change to the DB and it may cause issues in the event of a master crash. but that seems more complicated than worth for the rate of this error. but at least, as is, it's inconsistent to me that future trials would reflect the config in memory but we bail out before telling current trials.

PR #3471, opened 2022-01-19
fix: update task list Resource Pool sorter to alphaNumericSorter [DET-6434]

gt2345:
## Description 
Update task list "Resource Pool" column sorting to sort by alphabetical order.

Before:

https://user-images.githubusercontent.com/40620519/150424829-4b798328-d27e-4df6-80e3-05f089f1ba9b.mov

After:

https://user-images.githubusercontent.com/40620519/150424869-6f5ae74b-7284-4be9-88cd-b5f557c9b94f.mov

## Test Plan
- [ ] Navigate to "Tasks", click on "Resource Pool" header to verify sorting works.

### DIFF
--- a/webui/react/src/pages/TaskList.tsx
+++ b/webui/react/src/pages/TaskList.tsx
@@ -334,7 +334,7 @@ const TaskList: React.FC = () => {
       {
         dataIndex: 'resourcePool',
         key: 'resourcePool',
-        sorter: true,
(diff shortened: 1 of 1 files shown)


hkang1:
@gt2345 would you mind applying the same logic for `Experiment List (resource pool column)` as well? I just noticed that it's also set to `true` and it doesn't behave properly either.

PR #3472, opened 2022-01-19
ci: allow gcloud * create to take 30m

rb-determined-ai:
Creating resources in gcloud sometimes takes longer than 10m.  Wait up
to 30m to make our CI a little more robust.


rb-determined-ai:
looks like @ioga beat me to it

PR #3473, opened 2022-01-20
chore: add the basics for alternative job ordering attribute

hamidzr:
## Description

shifts the responsibility from individual jobs to the RM.

## Commentary (optional)

This doesn't fully implement all the necessary methods but puts down the skeleton. I'm looking to get this in early to agree on the structure and avoid other project work from diverging.

## Checklist


master/internal/resourcemanagers/priority.go
```diff
@@ -392,18 +333,13 @@ func comparePositions(a, b *sproto.AllocateRequest, groups map[*actor.Ref]*group
 func sortTasksWithPosition(
```
eecsliu, line 333: non-blocking: we can do this in a later commit, but it might be good to combine this fn with sortTasksWithPositionAndTimestamp

master/internal/resourcemanagers/group.go
```diff
@@ -12,10 +12,8 @@ type groupActorStopped struct {
...
-	handler *actor.Ref
-	// job      *sproto.JobSummary
-	maxSlots  *int
-	weight    float64 // QUESTION why is prio a ptr but weight is not.
-	priority  *int
-	qPosition float64
+	handler  *actor.Ref
+	maxSlots *int
+	weight   float64 // QUESTION why is prio a ptr but weight is not.
+	priority *int
```
eecsliu, line 18: question - I get that moving qPosition makes it easier to work with, but I'm wondering if we should keep it here as it is a "scheduling restriction". We're also forwarding the move job message to the RP anyways, and we would keep the scheduler code centralized around groups.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -60,16 +65,22 @@ func NewResourcePool(
...
 		fittingMethod: fittingMethod,
 
-		agents:      make(map[*actor.Ref]*agentState),
-		taskList:    newTaskList(),
-		groups:      make(map[*actor.Ref]*group),
-		scalingInfo: &sproto.ScalingInfo{},
+		agents:         make(map[*actor.Ref]*agentState),
+		taskList:       newTaskList(),
+		groups:         make(map[*actor.Ref]*group),
+		queuePositions: make(jobSortState),
```
eecsliu, line 71: I'm not sure I understand the point of having a queuePositions map here. What does this bring that groups doesn't already have? This also seems like it requires us to create the groupActorToID map just to translate from actor.Ref and access queuePositions.

eecsliu (commented):
looks good, barring some of the changes in resource pools.

master/internal/job/job.go
```diff
@@ -54,11 +54,11 @@ type (
...
 	}
-	// SetGroupOrder sets the order of the group in the priority scheduler.
-	SetGroupOrder struct {
-		QPosition    float64
-		ResourcePool string
-		Handler      *actor.Ref
+	// MoveJob requests the job to be moved within a priority queue relative to another job.
+	MoveJob struct {
+		ID      model.JobID
+		Anchor  model.JobID
```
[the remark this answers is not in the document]
hamidzr: yeah I thought it might be simpler? the alternative is to have two messages.

master/internal/resourcemanagers/group.go
hamidzr: there is a clear 1 to one relationship between job and qPosition but there is no such strict direct relationship between group and job.

master/internal/resourcemanagers/resource_pool.go
hamidzr: hmm it might be easier to do a call but some of my thinking is:
1. we want RM to persist and restore this scheduling state but nothing else from the group and having them as a map together makes it simpler
2. the other is that direct missing jobid to group mapping I mentioned in the other comment

master/internal/resourcemanagers/group.go
hamidzr: tasks when created sometimes have their `.Group` attribute pointed to the job actor but other times it'd be to the allocate actor and that group wouldn't deal with qPosition ever.

eecsliu (approved):
Talked about it and my concerns are addressed.

PR #3478, opened 2022-01-20
test: see if circleCI test freezes [DET-6474]

### LINKED ISSUE #3477: 'Kill all table rows' webui e2e test in 'Kill all experiments' step in 'Clean Up' spec times out when additional table columns are added
The webui-e2e test times out on CircleCI at the step listed in the title but passes locally.

EmilyBonar:
## Description
Resolves #3477

## Checklist

### DIFF
--- a/webui/tests/tests/StepImplementation.ts
+++ b/webui/tests/tests/StepImplementation.ts
@@ -410,13 +410,13 @@ export default class StepImplementation {
 
   @Step('<action> experiment row <row>')
   public async modifyExperiment(action: string, row: string) {
-    await t.click(t.tableCell({ row: parseInt(row) + 1, col: 13 }));
+    await t.click(t.tableCell({ row: parseInt(row) + 1, col: 16 }));
     await t.click(t.text(action, t.within(t.$('.ant-dropdown'))));
   }
 
   @Step('Open TensorBoard from experiment row <row>')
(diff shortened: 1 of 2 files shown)


webui/tests/tests/StepImplementation.ts
```diff
@@ -410,13 +410,15 @@ export default class StepImplementation {
 
   @Step('<action> experiment row <row>')
   public async modifyExperiment(action: string, row: string) {
-    await t.click(t.tableCell({ row: parseInt(row) + 1, col: 13 }));
+    const lastCol = (await t.$('th').elements()).length;
+    await t.click(t.tableCell({ row: parseInt(row) + 1, col: lastCol }));
```
hamidzr, line 414: if the test only failed after this change (even in another spot): could we avoid addressing this by counting cols? it should work but seems more brittle than it needs to and relies on taiko. wdy about using a selector based on the dropdown component name or some other attribute inherent to the dropdown

EmilyBonar: I like the idea, but afaik that would mean not using the `tableCell` function, and that brings up the question of how to control  which row is being selected

hamidzr: ah you're right

PR #3479, opened 2022-01-20
feat: additional pytorch callbacks [DET-5180]

mpkouznetsov:
## Description
DET-5180 additional pytorch callbacks
https://determinedai.atlassian.net/browse/DET-5180

Added
- `on_training_epoch_end(self, epoch_idx: int)`

Modified
- `on_training_epoch_start(self, epoch_idx: int)`  (legacy callbacks will log a warning `[2022-01-25 00:09:56] 9ef58fcb [rank=1] || 2022-01-25 00:09:56,151:WARNING [81]: on_training_epoch_start() without parameters is deprecated in 0.17.7. Please add epoch_idx parameter
`)

## Test Plan

Enhanced unit test for pytorch callbacks

## Commentary (optional)

1. The way I read other implementations of callbacks, `on_training_epoch_end` should be called _after_ processing the last batch in an epoch.
2. As I mentioned earlier, it is possible that `on_training_epoch_end` will not get invoked after the latest batch in an epoch if `is_epoch_end()` returns `False`.
3. Upon discussion with @liamcli we decided not to add `on_training_batch_(start|end)` for now since we are not sure of the performance hit of aggregating batch metrics if we were to add `on_training_batch_end` that would have a metrics argument. Here is his take on it:

> To answer this question, i think it helps to think about what the user would put in an on_train_batch_end callback vs just in train_batch at the end.  the two things that come to mind that are better in on_train_batch_end are things that we do not want to impact the profiler and things that depend on aggregated metrics.  otherwise, i don't see a reason to not put it in train_batch.  If we believe this, then i think we should take out on_train_batch until we can safely pass in aggregated global batch metrics.

## Checklist

### DIFF
--- a/docs/release-notes/3479-new-pytorch-callback-method.txt
+++ b/docs/release-notes/3479-new-pytorch-callback-method.txt
@@ -1,0 +1,5 @@
+:orphan:
+
+**New Features**
+
+- API: Add ``on_training_epoch_end`` to PytorchCallback. Add ``epoch_idx`` argument to ``on_training_epoch_start``.
--- a/docs/release-notes/3479-new-pytorch-callback-method.txt
+++ b/docs/release-notes/3479-new-pytorch-callback-method.txt
@@ -2,4 +2,5 @@
 
 **New Features**
 
-- API: Add ``on_training_epoch_end`` to PytorchCallback. Add ``epoch_idx`` argument to ``on_training_epoch_start``.
+-  API: Add ``on_training_epoch_end`` to PytorchCallback. Add ``epoch_idx`` argument to
+   ``on_training_epoch_start``.
--- a/docs/release-notes/3479-new-pytorch-callback-method.txt
+++ b/docs/release-notes/3479-new-pytorch-callback-method.txt
@@ -3,4 +3,6 @@
 **New Features**
 
 -  API: Add ``on_training_epoch_end`` to PytorchCallback. Add ``epoch_idx`` argument to
-   ``on_training_epoch_start``.
+   ``on_training_epoch_start``. Overriding ``on_training_epoch_start`` without ``epoch_idx``
+   argument is still supported for backwards compatibility but using this outdated signature is
+   discouraged.
--- a/harness/determined/pytorch/_callback.py
+++ b/harness/determined/pytorch/_callback.py
@@ -91,7 +91,7 @@ class PyTorchCallback:
         """
         pass
 
-    def on_training_batch_start(self) -> None:
+    def on_training_batch_start(self, batch_idx: int) -> None:
         """
         Run on start of a new training batch
         """
@@ -103,7 +103,7 @@ class PyTorchCallback:
         """
         pass
 
-    def on_training_epoch_start(self) -> None:
+    def on_training_epoch_start(self, epoch_idx: int) -> None:
         """
         Run on start of a new training epoch
         """
--- a/harness/determined/pytorch/_callback.py
+++ b/harness/determined/pytorch/_callback.py
@@ -91,6 +91,18 @@ class PyTorchCallback:
         """
         pass
 
+    def on_training_batch_start(self) -> None:
+        """
+        Run on start of a new training batch
+        """
+        pass
+
+    def on_training_batch_end(self, batch_idx: int) -> None:
+        """
+        Run on end of a training batch
+        """
+        pass
+
     def on_training_epoch_start(self) -> None:
         """
         Run on start of a new training epoch
--- a/harness/determined/pytorch/_callback.py
+++ b/harness/determined/pytorch/_callback.py
@@ -91,19 +91,22 @@ class PyTorchCallback:
         """
         pass
 
-    def on_training_batch_start(self, epoch_idx, batch_idx: int) -> None:
+    def on_training_batch_start(self, epoch_idx: int, batch_idx: int) -> None:
         """
         Run on start of a new training batch
         """
         pass
 
-    def on_training_batch_end(self, epoch_idx: int, batch_idx: int) -> None:
+    def on_training_batch_end(
+        self, epoch_idx: int, batch_idx: int, batch_metrics: Dict[str, Any]
+    ) -> None:
         """
         Run on end of a training batch
 
         Arguments:
             epoch_idx (int): index of the current epoch
             batch_idx (int): global batch index (does not get reset between epochs)
+            batch_metrics (Dict[str, Any]): metrics for this batch
         """
         pass
 
--- a/harness/determined/pytorch/_callback.py
+++ b/harness/determined/pytorch/_callback.py
@@ -97,6 +97,12 @@ class PyTorchCallback:
         """
         pass
 
+    def on_training_epoch_end(self, epoch_idx: int) -> None:
+        """
+        Run on end of a training epoch
+        """
+        pass
+
     def on_validation_epoch_start(self) -> None:
         """
         Run on start of a new validation epoch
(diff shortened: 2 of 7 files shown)


harness/determined/pytorch/_callback.py
```diff
@@ -91,12 +91,30 @@ def on_training_start(self) -> None:
...
+        """
+        pass
+
+    def on_training_epoch_start(self, epoch_idx: int) -> None:
         """
         Run on start of a new training epoch
         """
         pass
 
+    def on_training_epoch_end(self, epoch_idx: int) -> None:
```
rb-determined-ai, line 112: usually `on_thing_end()` would include the metrics for that thing.  We can't actually deliver end-of-epoch metrics, due to how training windows over which metrics are calculated in our system do not align with epoch boundaries.

I'd like to fix that in the future, with the Trainer API.  But it's going to be tricky since it will involve reconsidering the sharding model we currently have.  My first thought was to add a `metrics={}` argument here that will always be empty until we support real epoch-based training.  That's probably a terrible idea though... probably another inspection hack in the future is better.  What do you think @liamcli?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -383,12 +384,27 @@ def _train_for_step(
 
             self.context._current_batch_idx = batch_idx
             if self.context.is_epoch_start():
+                epoch_idx = self.get_epoch_idx(batch_idx)
                 for callback in self.callbacks.values():
                     with self.prof.record_timing(
                         f"callbacks.{callback.__class__.__name__}.on_training_epoch_start"
                     ):
-                        callback.on_training_epoch_start()
+                        sig = signature(callback.on_training_epoch_start)
```
rb-determined-ai, line 392: I approve of this hack.  Can you put a comment here that indicates in which determined version the old form was deprecated?  That way, when it is ancient we'll have a clear reminder of just how ancient it is, and it will be more likely we remember to actually remove the deprecated form.

harness/determined/pytorch/_callback.py
```diff
@@ -91,12 +91,30 @@ def on_training_start(self) -> None:
...
         pass
 
-    def on_training_epoch_start(self) -> None:
+    def on_training_batch_start(self, batch_idx: int) -> None:
+        """
+        Run on start of a new training batch
+        """
+        pass
+
+    def on_training_batch_end(self, batch_idx: int) -> None:
```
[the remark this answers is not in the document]
mpkouznetsov: it rises monotonically (within a given workload but across epochs, i. e. it does _not_ reset at the end of an epoch). I am not sure what our users expect / the right thing to do is.

harness/determined/pytorch/_callback.py
liamcli: For this particular case, I would either not pass in metrics or pass in metrics aggregated over batches in an epoch (with the understanding that the last batch will not align with epoch boundary but it would at least match what we print for that training workload).  I prefer the latter since a common use case will probably be early stopping decisions based on metrics.

harness/determined/pytorch/_pytorch_trial.py
mpkouznetsov: I changed the warning message to include the next version.

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
rb-determined-ai: What you say makes sense.  I mistakenly had thought that `train_batch()` took `batch_idx` that was not monotonically increasing.  Looking at the code, that's not the case though.

Perhaps we pass in `epoch_idx` too, the way we do to `train_batch()`?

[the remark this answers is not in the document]
rb-determined-ai: Also, what about passing the training metrics for this batch?

harness/determined/pytorch/_callback.py
mpkouznetsov: the way that loop is structured, at the time when we detect an end of epoch (by seeing that we processed the number of batches that's a factor of the epoch length) we have a list of per-batch metrics for all the batches encountered so far _on this worker_. Today, we don't build an aggregate here. We wait until the end of the workload, and then aggregate metrics across batches and workers. I am guessing that is what "_way_ easier" is referring to?

liamcli: given the complications, i'm okay with just passing `epoch_idx` to this callback and pointing users to something like `on_training_workload_end` for early stopping decisions.

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
liamcli: i'm in favor of passing `epoch_idx` to both batch start and batch end and passing batch metrics to batch end.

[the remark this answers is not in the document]
rb-determined-ai: is it ok if batch metrics are different on each worker?  If we had to sync batch metrics every batch I think it would become very expensive.

harness/determined/pytorch/_callback.py
rb-determined-ai: > at the time when we detect an end of epoch ... we have a list of per-batch metrics for all the batches encountered so far _on this worker_

I mean, we only have the metrics we've gathered _on this worker_ since the last training workload ended, right?  I think we flush the batch metrics after every workload.

mpkouznetsov: correct: https://github.com/determined-ai/determined/blob/36bb4d9beac348e32b42cecd9da7b26956f889c1/harness/determined/pytorch/_pytorch_trial.py#L272

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
mpkouznetsov: > i'm in favor of passing `epoch_idx` to both batch start and batch end and passing batch metrics to batch end.

I was interpreting "batch metrics" as metrics for the _current batch only_. Was your interpretation different, @liamcli ? Did you mean all per batch metrics seen on this worker in this workload so far?

harness/tests/experiment/fixtures/pytorch_xor_model.py
```diff
@@ -296,6 +297,9 @@ def __init__(self) -> None:
         self.validation_steps_started = 0
```
rb-determined-ai, line 297: What about adding another PyTorchCallback with the legacy signature for on_training_epoch_end, to make sure the backwards compatibility is not broken?

mpkouznetsov: Wait, _this callback_ is using the legacy signature for `on_training_epoch_start` (without the epoch argument). `on_training_epoch_end` is a net new method and therefore has no legacy signature :)

Would you like me to add another callback that uses _the new signature_ ?

docs/release-notes/3479-new-pytorch-callback-method.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**New Features**
+
+-  API: Add ``on_training_epoch_end`` to PytorchCallback. Add ``epoch_idx`` argument to
+   ``on_training_epoch_start``.
```
rb-determined-ai, line 6: Can you add to the note that we have backwards compatibility for existing code, so this is not a breaking change?

rb-determined-ai (approved):

docs/release-notes/3479-new-pytorch-callback-method.txt
mpkouznetsov: 👍

PR #3481, opened 2022-01-22
fix: resolve previous `/agents` fetch before polling for a new one [DET-6490]

hkang1:
## Description

There seems to be multiple problems within the reported issue with over polling `/agents`. The known and addressable issue is to ensure that the WebUI doesn't poll the agents endpoint while an existing one is still in progress. Update the polling policy for `agents` to only continue polling if an existing agents poll is not currently in progress.

## Test Plan

A couple of ways to test this...

Way to test without editing code:
- [ ] Open 6 or more tabs and watch for `/agents` endpoint to eventually enter a `pending` state (force saturation). When this occurs, the `usePolling` shouldn't continue to make additional `/agents` calls if it is behaving properly.

Ways to test via editing code:
- [ ] Modify an existing polling async endpoint call to take more than 5 seconds (default polling interval is 5, make it take 10 seconds via setTimeout) and check to make sure that the endpoint only gets called at a 10 sec interval instead.

## Checklist

### DIFF
--- a/webui/react/src/hooks/usePolling.ts
+++ b/webui/react/src/hooks/usePolling.ts
@@ -28,6 +28,12 @@ const DEFAULT_OPTIONS: PollingOptions = {
   runImmediately: true,
 };
 
+/**
+ * Polling hook that polls a given polling function.
+ * @param pollingFn
+ *    The function to poll. Pass an async/await function to ensure
+ *    the previous poll resolves first before making a new call.
+ */
 const usePolling = (pollingFn: PollingFn, options: PollingOptions = {}): PollingHooks => {
   const savedPollingFn = useRef<PollingFn>(pollingFn);
   const pollingOptions = useRef<PollingOptions>({ ...DEFAULT_OPTIONS, ...options });
--- a/webui/react/src/components/Navigation.tsx
+++ b/webui/react/src/components/Navigation.tsx
@@ -21,8 +21,8 @@ const Navigation: React.FC<Props> = ({ children }) => {
   const fetchAgents = useFetchAgents(canceler);
   const fetchResourcePools = useFetchResourcePools(canceler);
 
-  const fetchAuthOnly = useCallback(() => {
-    if (auth.isAuthenticated) fetchAgents();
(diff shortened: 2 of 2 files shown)


webui/react/src/hooks/usePolling.ts
```diff
@@ -28,6 +28,12 @@ const DEFAULT_OPTIONS: PollingOptions = {
   runImmediately: true,
 };
 
+/**
+ * Polling hook that polls a given polling function.
+ * @param pollingFn
+ *    The function to poll. Pass an async/await function to ensure
```
hamidzr, line 34: suggestion: if we only ever usePolling for async functions, or want to enforce, that wdy think about adding a runtime check that the passed in function actually does return a promise?

typescript types can probably also check for the declared return type of the `pollingFn`
`type PollingFn = (() => Promise<void>) | (() => void);`

hamidzr (approved):
lgtm. I  know we've tested manually before that usePolling does wait for the previous call to finish but since this is relying on that could you add your testing plan to the pr description?

hamidzr:
also I can't see the original ticket description so I'm just going based off of the code change

hkang1:
> also I can't see the original ticket description so I'm just going based off of the code change

Ah yes, so this was to address an observed issue where once enough HTTP calls are pending and outstanding, the polling of `/agents` compounds the issue by adding more calls that enter the `pending` state and never resolving. This PR attempts to avoid making more `/agents` calls if a previous already exists and has not resolved.

webui/react/src/hooks/usePolling.ts
hkang1: This is a good suggestion. I thought about this but decided to hold off on this because there are some polling calls that are non-async based still. But it's possible that they go away when another big PR (task logs) lands. Then at that point, we can enforce the polling function to be async functions.

varlogtim (approved):
I tested with this change and while it doesn't fully resolve the issue it prevents the response times from "running" away. Nice catch!

hkang1:
@cla-bot[bot] check

PR #3482, opened 2022-01-22
chore: wui error handling; api error handling

hamidzr:
## Description

- call handleError as late as possible and as a stop to bubbling up an exception
- avoid handling the same error twice.

## Commentary (optional)

some thoughts I'll organize this later

when the error is originated at the bottom a call stack it'll have some general information on what has happened.

- We have to avoid stopping it too early because we want the final caller to interrupt its execution
- As the exception bubbles up we have more and more opportunities and information to supplement the error

we use the DetErrror class to define all expected errors

no error should break the whole application
no exception should reach the highest level and crash the application. if an exception reaches the highest level either it'll be DetError which handleError knows perfectly well how to handle or we catch this in some other way.

should we 
1. lean toward never swalloing the exceptions and only building up on them (as DetError) unless we clearly know why we don't want it to interrup higher callers
2. have a catch at the highest level with a single handleError that hanles any non DetError as well

could we use https://reactjs.org/docs/error-boundaries.html ? 

## Checklist

- [ ] remove review inline comments

### DIFF
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -139,11 +139,7 @@ const ExperimentList: React.FC = () => {
         return response.experiments;
       });
     } catch (e) {
-      handleError(e, {
-        publicSubject: 'Unable to fetch experiments.',
-        silent: false, // CHECK do we want this to be silent?
-        // type: ErrorType.Api, // FIXME server? server vs api is a bit confusing to me.
-      });
+      handleError(e, { publicSubject: 'Unable to fetch experiments.' });
     } finally {
       setIsLoading(false);
     }
@@ -290,11 +286,11 @@ const ExperimentList: React.FC = () => {
         experimentId: id,
       });
     } catch (e) {
-      handleError({
-        message: 'Unable to save experiment description.',
-        silent: false, // CHECK example of why we want this. this was silent probably because a
-        // lower level handleError was covering it but that lower level error did not have the
-        // same information to provide as good a DetError.
+      handleError(e, {
+        isUserTriggered: true, // IMO we want to show errors for user triggered failures unless we
+        // have a better way of communicating this to the user.
+        publicMessage: 'Unable to save experiment description.',
+        silent: true,
       });
       setIsLoading(false);
     }
@@ -524,11 +520,11 @@ const ExperimentList: React.FC = () => {
         'Unable to View TensorBoard for Selected Experiments' :
         `Unable to ${action} Selected Experiments`;
       handleError(e, {
+        isUserTriggered: true,
         level: ErrorLevel.Error,
         publicMessage: 'Please try again later.',
         publicSubject,
         silent: false,
-        // type: ErrorType.Server,
       });
     }
   }, [ fetchExperiments, sendBatchActions, updateSettings ]);
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -292,8 +292,9 @@ const ExperimentList: React.FC = () => {
     } catch (e) {
       handleError({
         message: 'Unable to save experiment description.',
-        silent: true,
-        type: ErrorType.Api,
+        silent: false, // CHECK example of why we want this. this was silent probably because a
+        // lower level handleError was covering it but that lower level error did not have the
+        // same information to provide as good a DetError.
       });
       setIsLoading(false);
     }
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -141,7 +141,7 @@ const ExperimentList: React.FC = () => {
     } catch (e) {
       handleError(e, {
         publicSubject: 'Unable to fetch experiments.',
-        silent: true, // CHECK do we want this to be silent?
+        silent: false, // CHECK do we want this to be silent?
         // type: ErrorType.Api, // FIXME server? server vs api is a bit confusing to me.
       });
     } finally {
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -287,8 +287,7 @@ const ExperimentList: React.FC = () => {
       });
     } catch (e) {
       handleError(e, {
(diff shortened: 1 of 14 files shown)


webui/react/src/services/utils.ts
```diff
@@ -72,7 +79,13 @@ export const processApiError = (name: string, e: unknown): void => {
     options.publicSubject = `Failed in decoding ${name} API response.`;
   }
 
-  handleError(e, options);
+  let msg: string | undefined;
+  if (isApiResponse(e)) {
+    msg = e.statusText;
```
hamidzr, line 84: or await for response.text.

webui/react/src/utils/error.ts
```diff
@@ -102,14 +106,23 @@ const log = (e: DetError) => {
...
+  if (!e) {
+    throw new Error(`Unexpected error encountered: ${error}`);
+  }
   // Ignore request cancellation errors.
   if (isAborted(e)) return;
+  if (e.isHandled) {
+    return;
+  } else {
+    e.isHandled = true;
+  }
```
hkang1, line 125: **nit:**
```
// Ignore previously handled errors.
if (e.isHandled) return;
e.isHandled = true;
```

webui/react/src/services/utils.ts
```diff
@@ -53,7 +57,10 @@ export const isAborted = (e: any): boolean => {
 
 export const http = axios.create({ responseType: 'json', withCredentials: false });
 
-export const processApiError = (name: string, e: unknown): void => {
+/*
+Checks the api error turns in into a DetError and decides if it needs to be bubbled up.
```
hamidzr, line 61: reword

webui/react/src/services/types.ts
```diff
@@ -233,6 +233,8 @@ export interface TaskLogsParams extends LogsParams {
 /* eslint-disable-next-line @typescript-eslint/ban-types */
 export type EmptyParams = {}
 
+// FIXME we shouldn't need to include FetchOptions here. The api generator should already handle it.
```
hamidzr, line 236: probably better left for another pr.

hamidzr:
brought up as part of https://github.com/determined-ai/determined/projects/2#card-76051389

hkang1 (commented):
Looks good! General comments and a question.

webui/react/src/utils/error.ts
```diff
@@ -76,6 +79,7 @@ export class DetError extends Error {
     this.publicSubject = options.publicSubject || detError?.publicSubject || undefined;
     this.silent = options.silent || detError?.silent || false;
     this.type = options.type || detError?.type || ErrorType.Unknown;
+    this.isHandled = false;
```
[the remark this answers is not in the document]
hamidzr: the high-level idea is that each error needs to be handled only once and it should happen closest to where it has the most information on the cause.

eg (this was happening prior to this)

low in the callstack, eg processApiError, the error is handled and rethrown. I'm thinking probably a usecase we want to avoid. either handle or just create/augment a DetError and throw.
then at a higher level, eg where use had clicked forkExperiment button. they want to provide an error notification and break the flow of a successful fork then if they use handleerror it would do a double notification. 

this helps avoid getting into the double handling situation but it might be better if it didnt' silently ignore it 🤔 

also, another scenario is if we have a catch-all wrapper. it can know whether a DetError should be ignored or needs to be handled. it feels like it has some flaws

webui/react/src/utils/error.ts
```diff
@@ -102,14 +106,23 @@ const log = (e: DetError) => {
...
-
-  // Normalize error as DetError.
-  const e = isError(error) ? new DetError(error, options) : error;
-
+  let e: DetError | undefined;
+  if (isDetError(error)) e = error;
+  if (isError(error)) e = new DetError(error, options);
+  if (!e) {
+    throw new Error(`Unexpected error encountered: ${error}`);
+  }
```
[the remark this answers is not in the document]
hamidzr: I gotta see how this affects production but what I was trying to address was that some errors would just disappear if someone called `handleError` and passed in something handleError didn't expect.

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -290,10 +286,11 @@ const ExperimentList: React.FC = () => {
         experimentId: id,
       });
     } catch (e) {
-      handleError({
-        message: 'Unable to save experiment description.',
+      handleError(e, {
+        isUserTriggered: true, // IMO we want to show errors for user triggered failures unless we
```
hamidzr, line 290: remove/update

webui/react/src/services/utils.ts
```diff
@@ -72,7 +76,15 @@ export const processApiError = (name: string, e: unknown): void => {
     options.publicSubject = `Failed in decoding ${name} API response.`;
   }
 
-  handleError(e, options);
+  if (isApiResponse(e)) {
+    try {
+      const response = await e.json();
+      options.publicMessage = response.message || response.error?.error;
+    } catch (err) {
+      // Ignore the error.
```
hkang1, line 84: **question:** does the comment make eslint treat it as non-empty?

hkang1 (approved):
This is great work, thanks for improving error handling!

webui/react/src/services/utils.ts
hamidzr: I should probably make the comment more useful though. why we ignore it

hamidzr: maybe we shouldn't ignore and somehow encode it in the returned DetError

webui/react/src/utils/error.ts
```diff
@@ -102,21 +106,31 @@ const log = (e: DetError) => {
...
+const handleError = (error: DetError | unknown, options?: DetErrorOptions): void => {
   // Ignore request cancellation errors.
-  if (isAborted(e)) return;
+  if (isAborted(error)) return;
+  let e: DetError | undefined;
+  if (isDetError(error)) {
+    e = error;
+  } else {
+    e = new DetError(error, options);
+  }
```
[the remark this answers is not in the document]
hamidzr: 👍🏼 

todo: add the eslint rule https://eslint.org/docs/rules/no-throw-literal

hamidzr:
thanks for the review

PR #3483, opened 2022-01-22
chore: Remove deprecated routes from old API

mapmeld:
## Description

Removes deprecated code from core_experiment.go (old API)

Specifically: experiment-list, experiment-summaries, and kill

Work in progress

## Checklist

### DIFF
--- a/docs/release-notes/3483-remove-deprecated-from-experiment-api.txt
+++ b/docs/release-notes/3483-remove-deprecated-from-experiment-api.txt
@@ -2,6 +2,6 @@
 
 **Removed Features**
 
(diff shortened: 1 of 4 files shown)


hamidzr (commented):
lgtm so far.
I don't if we want to include a bit in the release notes for redirecting potential users to the alternatives

hamidzr (approved):
I did a search and looks like we still use experiment-list in one location. other than that it lgtm

```
harness/determined/cli/experiment.py
661:    r = api.get(parsed_args.master, "experiment-list", params=params)
```

PR #3484, opened 2022-01-24
fix: running experiment log misbehave [DET-6339]

gt2345:
## Description

For running experiment, there is a chance that when first navigate to logs tab, (especially when logs are long and appearing quickly), logs are not automatically tailing. 

## Test Plan

Navigate to a running experiments with continuous long log 10+ times (or just refresh the logs page), verify that logs are tailing. 
Also verify that scrolling up/down, scroll to oldest, tailing enable are all functional. 

## Commentary (optional)

Upon observation of `addLogs` and `handleItemsRendered`, I noticed that when the page loaded, `addLogs` is triggered several times with different length of `newLogs` argument, and I noticed that logs only misbehave when `handleItemsRendered` is triggered (with `isScrollReady` being true) before `addLogs` triggered with `newLogs.length` equal to 1 (after initialization), so I add an indicator for this event to `handleEnableTailing` manually. 

Before:

https://user-images.githubusercontent.com/40620519/150706180-e6aad909-f171-44ac-bc6e-f162d4675dd8.mov

After:

https://user-images.githubusercontent.com/40620519/150706184-f41f6881-dc23-4253-b336-23e4864106d2.mov

### DIFF
--- a/webui/react/src/components/LogViewerCore.tsx
+++ b/webui/react/src/components/LogViewerCore.tsx
@@ -148,14 +148,14 @@ const LogViewerCore: React.FC<Props> = ({
 
   const addLogs = useCallback((newLogs: ViewerLog[], prepend = false): void => {
     if (newLogs.length === 0) return;
-    setLogs(prevLogs =>  {
+    setLogs(prevLogs => {
       // Indicator for running experments when log stream is stable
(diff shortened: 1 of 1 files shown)


EmilyBonar (approved):
Seems to work, and good job documenting your PR!

PR #3486, opened 2022-01-24
chore: Use model's name as its identifier

mapmeld:
## Description

Use the name of a model as the identifier, changing the REST API, Web UI, and Python client for the model registry. 

If a model with the id '300' had the name 'traffic-cone' then `/api/v1/models/300` would become `/api/v1/models/traffic-cone`; `/det/models/300/*` (i.e. including model version pages) becomes `/det/models/traffic-cone/*`, and `client.get_model(300)` becomes `client.get_model('traffic-cone')`.

Python users can still use the ID with `client.get_model_by_id`

TODOs: Proto API did not route models with a '/' in the name, even after converting to %2F. We ought to migrate these and block future model names from using it.  We also should discourage future model names which are too long, case-insensitive matches, or use all numbers (I think this can be UI and not DB-level constraints)

## Checklist


harness/determined/common/experimental/determined.py
```diff
@@ -155,32 +155,36 @@ def create_model(
...
         """
-        r = self._session.get("/api/v1/models?name={}".format(name))
+        r = self._session.get("/api/v1/models?id={}".format(model_id))
         models = r.json().get("models")
         assert len(models) > 0
         return model.Model.from_json(models[0], self._session)
 
+    def get_model_by_name(self, name: str) -> model.Model:
+        return self.get_model(name)
+
```
rb-determined-ai, line 180: I'd prefer we not introduce alias functions unnecessarily.  I think `get_model(name)` is intuitive enough we can skip this.

harness/determined/common/experimental/determined.py
```diff
@@ -155,32 +155,36 @@ def create_model(
...
     def get_models(
         self,
         sort_by: model.ModelSortBy = model.ModelSortBy.NAME,
         order_by: model.ModelOrderBy = model.ModelOrderBy.ASCENDING,
         name: str = "",
         description: str = "",
+        model_id: int = 0,
```
rb-determined-ai, line 187: do you have a guarantee that `model_id`s are always non-zero?

rb-determined-ai: Also... there should never be multiple models with the same model_id, so why isn't `get_model_by_id()` sufficient?

(also, I have the same question for `name`)

harness/determined/experimental/client.py
```diff
@@ -209,17 +209,17 @@ def create_model(
...
     Arguments:
-        model_id (int): The unique id of the model.
+        name (str): The unique name of the model.
     """
     assert _determined is not None
-    return _determined.get_model(model_id)
+    return _determined.get_model(name)
 
 
 @_require_singleton
```
rb-determined-ai, line 225: Can you remove this `get_model_by_name` too? (same reason, to avoid aliases)

(if I am correct that this has not been released yet...)

harness/determined/experimental/client.py
```diff
@@ -233,7 +233,21 @@ def get_model_by_name(name: str) -> Model:
...
+    with the provided numeric id. If no model with that id is found in the registry,
+    an exception is raised.
+
+    Arguments:
+        model_id (int): The unique id of the model.
+    """
+    assert _determined is not None
+    matching_models = _determined.get_models(model_id=model_id)
     assert len(matching_models) > 0
     return matching_models[0]
```
rb-determined-ai, line 252: this logic seems suspect.  What if there are multiple matching models?

master/static/srv/get_models.sql
```diff
@@ -1,10 +1,11 @@
...
-WHERE ($1 = '' OR m.archived = $1::BOOL)
-AND ($2 = '' OR (u.username IN (SELECT unnest(string_to_array($2, ',')))))
-AND ($3 = '' OR (m.labels <@ string_to_array($3, ',')))
-AND ($4 = '' OR m.name ILIKE $4)
-AND ($5 = '' OR m.description ILIKE $5)
+WHERE ($1 = 0 OR m.id = $1)
+AND ($2 = '' OR m.archived = $2::BOOL)
+AND ($3 = '' OR (u.username IN (SELECT unnest(string_to_array($3, ',')))))
+AND ($4 = '' OR (m.labels <@ string_to_array($4, ',')))
+AND ($5 = '' OR m.name ILIKE $5)
```
rb-determined-ai, line 8: Question: What is `ILIKE`? why not require exact match between name and model?

rb-determined-ai (commented):
I am not qualified to review either the webui or the master-side changes here, but I'm happy to review the python-side stuff.  You'll need to find other reviewers for those parts and get their approvals in addition to mine.

Question: are there cli changes necessary with this change as well?

webui/react/src/pages/ModelVersionDetails.tsx
```diff
@@ -219,17 +222,17 @@ const ModelVersionDetails: React.FC = () => {
     }));
   }, [ modelVersion?.checkpoint ]);
 
-  if (isNaN(parseInt(modelId))) {
-    return <Message title={`Invalid Model ID ${modelId}`} />;
+  if (modelName === '') {
+    return <Message title={`Invalid Model Name ${modelName}`} />;
```
EmilyBonar, line 226: Same thing about the error message here

webui/react/src/pages/ModelDetails.tsx
```diff
@@ -32,13 +32,13 @@ import settingsConfig, { Settings } from './ModelDetails/ModelDetails.settings';
...
 interface Params {
-  modelId: string;
+  modelName: string;
 }
 
 const ModelDetails: React.FC = () => {
   const { auth: { user } } = useStore();
   const [ model, setModel ] = useState<ModelVersions>();
-  const { modelId } = useParams<Params>();
+  const modelName = decodeURIComponent(useParams<Params>().modelName);
```
EmilyBonar, line 41: question: why is decodeURIComponent necessary?

webui/react/src/pages/ModelDetails.tsx
```diff
@@ -125,9 +123,9 @@ const ModelDetails: React.FC = () => {
     useCallback(async (editedDescription: string, versionId: number) => {
       try {
         await patchModelVersion({
-          body: { comment: editedDescription, id },
-          modelId: id,
-          versionId: versionId,
+          body: { comment: editedDescription, id: versionId },
```
EmilyBonar, line 126: Unless you changed this on the backend, the `id` field in body in `patchModelVersion` is expecting the model ID, not the version ID.

EmilyBonar (commented):
The webUI portion looks pretty good, but I can't speak to the rest of it.

webui/react/src/pages/ModelVersionDetails.tsx
mapmeld: ok I am going to replace these both with
```javascript
  if (!modelName) {
    return <Message title={`Model name is empty`} />;
  }
```

webui/react/src/pages/ModelDetails.tsx
mapmeld: A few characters (I know &, =, and /, probably ? and #) are not decoded in the param import. I get this error on the model page:

Screen Shot 2022-01-25 at 1 25 26 PM

I'm banning / from model names because (even escaped) it messes up `/models/:name/versions`, but I don't want to find and ban all of the characters in this set.

webui/react/src/pages/ModelDetails.tsx
mapmeld: Thanks!  I got turned around on this. This should be fixed in any patchModelVersion calls with the latest commit

docs/release-notes/3486-use-model-name-as-identifier.txt
```diff
@@ -0,0 +1,20 @@
...
+   special characters in a name must be URI-encoded. You can get a model by ID with
+   `/api/v1/models?id=`.
+
+-  API: On the list of models (`/api/v1/models`) the optional name parameter is now a case-sensitive
+   match, unless you add the parameter `name_case_insensitive=true`.
+
+-  Client: `get_model(int: id)` is deprecated - use `get_model(name: string)` or
+   `get_model_by_id(int: id)`.
+
+-  Client: `get_model_by_name` is deprecated - use `get_model(name: string)`.
```
ioga, line 15: to me, "deprecated" means its use is discouraged but it's still available. maybe this should read "removed" instead?

ioga (approved):
lgtm

docs/release-notes/3486-use-model-name-as-identifier.txt
mapmeld: good idea, will do

harness/determined/common/experimental/determined.py
mapmeld: I've removed `get_model_by_name`

harness/determined/experimental/client.py
mapmeld: `get_model_by_name` is now removed

harness/determined/common/experimental/determined.py
mapmeld: My experience is that model_id in Postgres starts at 1. A 0 would cause problems elsewhere in our code (proto returns a 0 for an empty int).

I wrote this for `get_model_by_id` to piggy-back onto `get_models`, if they are separated then we need a new API route to accommodate it.

The name param will be included here for backwards compatibility / case-insensitive queries.

harness/determined/experimental/client.py
rb-determined-ai: maybe `assert len(matching_models) == 1` then?

docs/release-notes/3486-use-model-name-as-identifier.txt
rb-determined-ai: wait, was `get_model_by_name()` released?  I thought it never made it into a release (in which case it doesn't need a release note?)

rb-determined-ai (approved):
python side looks good

docs/release-notes/3486-use-model-name-as-identifier.txt
mapmeld: I think you're right `get_model_by_name()` was 6 days ago, it's not in the release but it is [in determined-ee](https://github.com/determined-ai/determined-ee/commit/057a25773e1c7976af2df330980dd419cdfe71d1#diff-2c4cec18a64aa21942e745435466ff4499ee1e9e797990a7ade4461d688a8869) -- I'll remove it from the release notes on Tuesday

harness/determined/experimental/client.py
mapmeld: will do ✅

PR #3487, opened 2022-01-24
feat: Tensorboard use local storage [DET-6029]

varlogtim:
Features:
 - Tensorboard rewrite to sync to local storage.
 - Start of named logger use
 - Added debug flag for task specs. Only used by Tensorboard.
 - Tensorboard now starts for any new files found, i.e., profiler fix

Tensorboard Fetchers:
 - GCS Fetcher
 - S3 Fetcher
 - Azure Fetcher
 - SharedFS

Bug Fixes:
 - Empty directories in checkpoints did not work for Azure
 - Upload for azure_client should overwrite to match S3 behavior.

Unit Tests added:
 - Azure live storage manager test.
 - S3 Tensorboard Fetcher
 - GCS Tensorboard Fetcher
 - Azure Tensorboard Fetcher
 - SharedFS Tensorboard Fetcher

### DIFF
--- a/harness/determined/common/storage/azure.py
+++ b/harness/determined/common/storage/azure.py
@@ -62,13 +62,19 @@ class AzureStorageManager(StorageManager):
     def upload(self, storage_id: str, storage_dir: str) -> None:
         storage_prefix = storage_id
         for rel_path in sorted(self._list_directory(storage_dir)):
-            if rel_path.endswith("/"):
-                continue
             # Use posixpath so that we always use forward slashes, even on Windows.
             container_blob = posixpath.join(self.container, storage_prefix, rel_path)
-            blob_dir, blob_base = posixpath.split(container_blob)
-            abs_path = os.path.join(storage_dir, rel_path)
-            logging.debug(f"Uploading blob {blob_base} to container {blob_dir}.")
+
+            if rel_path.endswith("/"):
+                blob_dir, blob_base = posixpath.split(container_blob.rstrip("/"))
+                blob_base = f"{blob_base}/"
+                abs_path = "/dev/null"
+                logging.debug(f"Uploading blob empty {blob_base} to container {blob_dir}.")
+            else:
+                blob_dir, blob_base = posixpath.split(container_blob)
+                abs_path = os.path.join(storage_dir, rel_path)
+                logging.debug(f"Uploading blob {blob_base} to container {blob_dir}.")
+
             self.client.put(blob_dir, blob_base, abs_path)
 
     @util.preserve_random_state
@@ -79,11 +85,11 @@ class AzureStorageManager(StorageManager):
             found = True
             dst = os.path.join(storage_dir, os.path.relpath(blob, storage_prefix))
             dst_dir = os.path.dirname(dst)
-            if not os.path.exists(dst_dir):
-                os.makedirs(dst_dir, exist_ok=True)
+            os.makedirs(dst_dir, exist_ok=True)
 
             # Only create empty directory for keys that end with "/".
             if blob.endswith("/"):
+                os.makedirs(dst, exist_ok=True)
                 continue
 
             # Use posixpath so that we always use forward slashes, even on Windows.
--- a/harness/determined/common/storage/azure_client.py
+++ b/harness/determined/common/storage/azure_client.py
@@ -25,6 +25,8 @@ class AzureStorageClient(object):
             self.client = blob.BlobServiceClient.from_connection_string(connection_string)
         elif account_url:
             self.client = blob.BlobServiceClient(account_url, credential)
+        else:
+            raise ValueError("Either 'connection_string' or 'account_url' must be specified.")
 
         logging.info(f"Trying to create Azure Blob Storage Container: {container}.")
         try:
@@ -50,7 +52,7 @@ class AzureStorageClient(object):
     def put(self, container_name: str, blob_name: str, filename: Union[str, Path]) -> None:
         """Upload a file to the specified blob in the specified container."""
         with open(filename, "rb") as file:
-            self.client.get_blob_client(container_name, blob_name).upload_blob(file)
+            self.client.get_blob_client(container_name, blob_name).upload_blob(file, overwrite=True)
 
     @util.preserve_random_state
     def get(self, container_name: str, blob_name: str, filename: str) -> None:
--- a/harness/determined/exec/tensorboard.py
+++ b/harness/determined/exec/tensorboard.py
@@ -1,11 +1,25 @@
+import json
+import logging
 import os
 import subprocess
 import sys
+import tempfile
 import time
-from typing import Callable, List, Tuple
+from typing import Any, Dict, List
 
 import boto3
 import requests
+from packaging import version
+
+import determined.common
+from determined.tensorboard import fetchers
+
+TENSORBOARD_TRIGGER_READY_MSG = "TensorBoard contains metrics"
+FETCH_INTERVAL = 1
+MAX_WAIT_TIME = 600
+
+
+logger = logging.getLogger("determined.exec.tensorboard")
 
 
 def set_s3_region() -> None:
@@ -13,11 +27,11 @@ def set_s3_region() -> None:
     if bucket is None:
         return
 
-    endpoint_url = os.environ.get("DET_S3_ENDPOINT_URL", None)
+    endpoint_url = os.environ.get("DET_S3_ENDPOINT_URL")
     client = boto3.client("s3", endpoint_url=endpoint_url)
-    bucketLocation = client.get_bucket_location(Bucket=bucket)
+    bucket_location = client.get_bucket_location(Bucket=bucket)
 
-    region = bucketLocation["LocationConstraint"]
+    region = bucket_location["LocationConstraint"]
 
     if region is not None:
         # We have observed that in US-EAST-1 the region comes back as None
@@ -26,121 +40,123 @@ def set_s3_region() -> None:
         os.environ["AWS_REGION"] = str(region)
 
 
-def wait_for_tensorboard(max_seconds: float, url: str, still_alive_fn: Callable[[], bool]) -> bool:
-    """Return True if the process successfully comes up before a deadline."""
+def get_tensorboard_args(tb_version: str, tfevents_dir: str, add_args: List[str]) -> List[str]:
+    """Build tensorboard startup args.
 
-    deadline = time.time() + max_seconds
-
-    while True:
-        if time.time() > deadline:
-            print(f"TensorBoard did not find metrics within {max_seconds} seconds", file=sys.stderr)
-            return False
-
-        if not still_alive_fn():
-            print("TensorBoard process died before reporting metrics", file=sys.stderr)
-            return False
-
-        time.sleep(1)
-
-        try:
-            res = requests.get(url)
-            res.raise_for_status()
-        except (requests.exceptions.ConnectionError, requests.exceptions.HTTPError):
-            continue
-
-        try:
-            tags = res.json()
-        except ValueError:
-            continue
-
-        # TensorBoard will return { trial/<id> : { tag: value } } when data is present.
-        if len(tags) == 0:
-            print("TensorBoard is awaiting metrics...")
-            continue
-
-        for val in tags.values():
-            if len(val):
-                print("TensorBoard contains metrics")
-                return True
-
-
-def get_tensorboard_version(version: str) -> Tuple[str, str]:
-    """
-    Gets the version of the tensorboard package currently installed. Used
-    by downstream processes to determine args passed in.
-    :return: version in the form of (major, minor) tuple
-    """
-
-    major, minor, _ = version.split(".")
-
-    return major, minor
-
-
-def get_tensorboard_args(args: List[str]) -> List[str]:
-    """
-    Builds tensorboard startup args from args passed in from tensorboard-entrypoint.sh
     Args are added and deprecated at the mercy of tensorboard; all of the below are necessary to
     support versions 1.14, 2.4, and 2.5
 
-    - If multiple directories are specified and the tensorboard version is > 1,
-    use legacy logdir_spec behavior
-
     - Tensorboard 2+ no longer exposes all ports. Must pass in "--bind_all" to expose localhost
-
     - Tensorboard 2.5.0 introduces an experimental feature (default load_fast=true)
     which prevents multiple plugins from loading correctly.
     """
     task_id = os.environ["DET_TASK_ID"]
     port = os.environ["TENSORBOARD_PORT"]
 
-    version = args.pop(0)
-
-    # logdir is the second argument passed in from tensorboard_manager.go. If multiple directories
-    # are specified and the tensorboard version is > 1, use legacy logdir_spec behavior. NOTE:
-    # legacy logdir_spec behavior is not supported by many tensorboard plugins
-    logdir = args.pop(0)
-
     tensorboard_args = [
         "tensorboard",
         f"--port={port}",
         f"--path_prefix=/proxy/{task_id}",
-        *args,
+        *add_args,
     ]
 
-    major, minor = get_tensorboard_version(version)
-
-    if major == "2":
+    # Version dependant args
+    if version.parse(tb_version) >= version.parse("2"):
         tensorboard_args.append("--bind_all")
-        if minor >= "5":
-            tensorboard_args.append("--load_fast=false")
-        if len(logdir.split(",")) > 1:
-            tensorboard_args.append(f"--logdir_spec={logdir}")
-            return tensorboard_args
+    if version.parse(tb_version) >= version.parse("2.5"):
+        tensorboard_args.append("--load_fast=false")
 
-    tensorboard_args.append(f"--logdir={logdir}")
+    tensorboard_args.append(f"--logdir={tfevents_dir}")
+
     return tensorboard_args
 
 
-def main(args: List[str]) -> int:
-    set_s3_region()
-
+def get_tensorboard_url() -> str:
     task_id = os.environ["DET_TASK_ID"]
     port = os.environ["TENSORBOARD_PORT"]
     tensorboard_addr = f"http://localhost:{port}/proxy/{task_id}"
-    url = f"{tensorboard_addr}/data/plugin/scalars/tags"
-    tensorboard_args = get_tensorboard_args(args)
+    return tensorboard_addr
 
-    print(f"Running: {tensorboard_args}")
-    p = subprocess.Popen(tensorboard_args)
 
-    def still_alive() -> bool:
-        return p.poll() is None
+def check_tensorboard_responsive() -> bool:
+    # Ensure Tensorboard is responding to HTTP request to prevent 502 from master.
+    tensorboard_url = get_tensorboard_url()
+    try:
+        # Attempt HTTP request to Tensorboard.
+        res = requests.get(tensorboard_url)
+        res.raise_for_status()
+        return True
 
-    if not wait_for_tensorboard(600, url, still_alive):
-        p.kill()
+    except (requests.exceptions.ConnectionError, requests.exceptions.HTTPError, ValueError) as exp:
+        logger.warning(f"Tensorboard not responding to HTTP: {exp}")
 
-    return p.wait()
+    return False
+
+
+def start_tensorboard(
+    config: Dict[str, Any],
+    tb_version: str,
+    storage_paths: List[str],
+    add_tb_args: List[str],
+) -> int:
+    """Start Tensorboard and look for new files."""
+
+    stop_time = time.time() + MAX_WAIT_TIME
+    triggered = False
+    responsive = False
+
+    with tempfile.TemporaryDirectory() as local_dir:
+
+        # Get fetcher and perform initial fetch
+        fetcher = fetchers.build(config, storage_paths, local_dir)
+        num_fetched_files = fetcher.fetch_new()
+
+        # Build Tensorboard args and launch process.
+        tb_args = get_tensorboard_args(tb_version, local_dir, add_tb_args)
+        logger.debug(f"tensorboard args: {tb_args}")
+        tensorboard_process = subprocess.Popen(tb_args)
+
+        try:
+            while True:
+                ret_code = tensorboard_process.poll()
+                if ret_code is not None:
+                    raise RuntimeError(f"Tensorboard process died, exit code({ret_code}).")
+
+                # Check if we have reached a timeout without receiving metrics
+                if num_fetched_files == 0 and time.time() > stop_time:
+                    raise RuntimeError("No new files were fetched before the timeout.")
+
+                if not responsive:
+                    if time.time() > stop_time:
+                        raise RuntimeError("Tensorboard wasn't responsive before the timeout.")
+                    responsive = check_tensorboard_responsive()
+
+                if responsive and not triggered and num_fetched_files > 0:
+                    print(TENSORBOARD_TRIGGER_READY_MSG)
+                    triggered = True
+
+                time.sleep(FETCH_INTERVAL)
+                num_fetched_files += fetcher.fetch_new()
+
+        finally:
+            if tensorboard_process.poll() is None:
+                logger.info("Killing tensorboard process")
+                tensorboard_process.kill()
+
+        return tensorboard_process.wait()
 
 
 if __name__ == "__main__":
-    sys.exit(main(sys.argv[1:]))
+    tb_version = sys.argv[1]
+    config_path = sys.argv[2]
+    storage_paths = sys.argv[3].split(",")
+    additional_tb_args = sys.argv[4:]
+
+    config = {}
+    with open(config_path) as config_file:
+        config = json.load(config_file)
+
+    determined.common.set_logger(determined.common.util.debug_mode())
+
+    ret = start_tensorboard(config, tb_version, storage_paths, additional_tb_args)
+    sys.exit(ret)
--- a/harness/determined/tensorboard/fetchers/__init__.py
+++ b/harness/determined/tensorboard/fetchers/__init__.py
@@ -1,0 +1,33 @@
+from typing import Any, Dict, List, Optional, Type
+
+from .azure import AzureFetcher
+from .base import Fetcher
+from .gcs import GCSFetcher
+from .s3 import S3Fetcher
+from .shared import SharedFSFetcher
+
+__all__ = [
+    "S3Fetcher",
+    "GCSFetcher",
+    "AzureFetcher",
+    "SharedFSFetcher",
+]
+
+_FETCHERS = {
+    "s3": S3Fetcher,
+    "gcs": GCSFetcher,
+    "azure": AzureFetcher,
+    "shared_fs": SharedFSFetcher,
+}  # type: Dict[str, Type[Fetcher]]
+
+
+def build(config: Dict[str, Any], paths: List[str], local_dir: str) -> Fetcher:
+    storage_config = config.get("checkpoint_storage")
+    if storage_config is None:
+        raise ValueError("config does not contain a 'checkpoint_storage' key")
+
+    storage_type = storage_config.get("type")
+    if storage_type not in _FETCHERS:
+        raise ValueError(f"checkpoint_storage type '{storage_type}' is not supported")
+
+    return _FETCHERS[storage_type](storage_config, paths, local_dir)
--- a/harness/determined/tensorboard/fetchers/azure.py
+++ b/harness/determined/tensorboard/fetchers/azure.py
@@ -1,0 +1,73 @@
+import datetime
+import logging
+import os
+import urllib
+from typing import Any, Dict, Generator, List, Tuple
+
+from .base import Fetcher
+
+logger = logging.getLogger(__name__)
+
+
+class AzureFetcher(Fetcher):
+    def __init__(self, storage_config: Dict[str, Any], storage_paths: List[str], local_dir: str):
+        from azure.storage import blob
+
+        connection_string = storage_config.get("connection_string")
+        container = storage_config.get("container")
+        account_url = storage_config.get("account_url")
+        credential = storage_config.get("credential")
+
+        if storage_config.get("connection_string"):
+            self.client = blob.BlobServiceClient.from_connection_string(connection_string)
+        elif account_url:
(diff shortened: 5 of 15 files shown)


harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +38,140 @@ def set_s3_region() -> None:
...
+def get_tensorboard_url() -> str:
     task_id = os.environ["DET_TASK_ID"]
     port = os.environ["TENSORBOARD_PORT"]
     tensorboard_addr = f"http://localhost:{port}/proxy/{task_id}"
-    url = f"{tensorboard_addr}/data/plugin/scalars/tags"
-    tensorboard_args = get_tensorboard_args(args)
+    return f"{tensorboard_addr}/data/plugin/scalars/tags"
+
+
+def check_for_metrics() -> bool:
```
rb-determined-ai, line 83: Why wait for tensorboard to show metrics, when we now know perfectly well when metrics are ready?

harness/determined/tensorboard/fetchers/s3.py
```diff
@@ -0,0 +1,88 @@
...
+        prefix = urllib.parse.urlparse(prefix).path.lstrip("/")
+
+        list_objs_args = {}  # type: Dict[str, Any]
+        while True:
+            list_objs_ret = self.client.list_objects_v2(
+                Bucket=self.bucket_name,
+                Prefix=prefix,
+                **list_objs_args,
+            )
+            logger.debug(f"list_objects_v2 response dict: {list_objs_ret}")
```
rb-determined-ai, line 47: Is this going to become very expensive on large experiments?

I wonder if we could save tensorflow files with timestamps in the filename and ignore anything that was too old using the `start_from` parameter or whatever it's called.

Just a though, not in scope for this PR I think.

harness/tests/storage/util.py
```diff
@@ -64,3 +66,94 @@ def run_storage_lifecycle_test(
...
+        timed_backoff_sync_check(second_files)
+
+        # (Additional Sync) Upload a new set of files, sync, ensure new files downloaded.
+        third_files = get_filepath_dict([("baz", b"baz"), ("qux", b"qux")])
+        put_files(third_files)
+        files_to_remove.extend(third_files.keys())
+
+        timed_backoff_sync_check({**second_files, **third_files})
+    finally:
+        rm_files(files_to_remove)
```
rb-determined-ai, line 159: This is a bit of a strange cleanup pattern, and it's not clear that it's bulletproof.

Why not just create a temporary directory, and put all the files you are going to clean up in there, then delete the whole thing in the `finally` block?

rb-determined-ai (commented):
Looks pretty good.  I like the tests a lot.

Regarding your request for having a StorageInterface: I still think that the "right" abstraction are the ones we already have.  You want a Fetcher?  We have a Fetcher interface.  You want a StorageManager?  We have a storage manager.  The interface you outlined to me in DM looks tailor-made for blob storage, but it looks like the shared-fs storage implementations would get worse, not better.

What would you say to leaving the existing interfaces (Fetcher, StorageManager), but implementing Azure/GCS/S3 as a BlobFetcher or BlobStorageManager?  Each of those could take a BlobBackend, with methods as you describe in your StorageInterface, but since the new interface would be specific to blob storage, I think you'd end up with a lot cleaner code.

master/pkg/model/storage_config.go
```diff
@@ -0,0 +1,168 @@
+package model
```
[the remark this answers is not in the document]
varlogtim: I actually don't know where this came from at all. I must have written it at some point to add master config validation for storage related components during master startup.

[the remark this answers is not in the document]
rb-determined-ai: probably a rebasing issue... I've deleted this file since the very first pr you posted a long while ago.

harness/tests/storage/util.py
```diff
@@ -64,3 +66,94 @@ def run_storage_lifecycle_test(
...
+        local_file_list = list_files(local_sync_dir)
+        assert len(local_file_list) == 0
+
+        # (New Sync) Upload a set of files, sync, ensure files downloaded.
+        first_files = get_filepath_dict([("foo", b"foo"), ("bar", b"bar")])
+        put_files(first_files)
+        files_to_remove.extend(first_files.keys())
+
+        timed_backoff_sync_check(first_files)
+        time.sleep(1)  # racy storage propegation with Azure
```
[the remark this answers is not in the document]
varlogtim: done!

harness/determined/common/storage/azure.py
```diff
@@ -62,13 +62,19 @@ def restore_path(self, storage_id: str) -> Iterator[str]:
     def upload(self, storage_id: str, storage_dir: str) -> None:
         storage_prefix = storage_id
         for rel_path in sorted(self._list_directory(storage_dir)):
-            if rel_path.endswith("/"):
-                continue
-            # Use posixpath so that we always use forward slashes, even on Windows.
```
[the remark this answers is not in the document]
varlogtim: I don't recall doing so. Must have been an accident.

harness/determined/common/storage/azure_client.py
```diff
@@ -25,6 +25,11 @@ def __init__(
             self.client = blob.BlobServiceClient.from_connection_string(connection_string)
         elif account_url:
             self.client = blob.BlobServiceClient(account_url, credential)
+        else:
+            raise ValueError(
+                "Either 'connection_string' or 'account_url' and, "
+                "optionally, 'credential' must be specified."
```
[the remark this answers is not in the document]
varlogtim: Yeah, I teetered on a few different options here. Updated to only specify the requirements, which is the condition.

harness/determined/exec/tensorboard.py
varlogtim: I am not sure what you mean here, could you add a little clarity?

You could me that we know if we have downloaded files or not, so why wait to see if Tensorboard is responding with metrics?

The short answer is, this was how it was working.

harness/determined/tensorboard/fetchers/s3.py
varlogtim: > Is this going to become very expensive on large experiments?
I am not sure what you mean by expensive. Do you mean for situation where there are ~10K files, we will do 10 requests per second, to list files?

So, you are thinking that at some point after the initial listing of files we then start giving a time offset to the `list_objects_v2()` call so that we only see recently updated files? This is a super good idea actually!

I am curious if Tensorboard does this already.

> not in scope for this PR I think
Cool. Will do it at some point in the future.

harness/tests/storage/util.py
varlogtim: > This is a bit of a strange cleanup pattern, and it's not clear that it's bulletproof.

This will cause the test to fail if we fail to remove one of the files. Thoughts on this behavior?

I basically need to delete a dynamic list of files when I exit the test context for any reason. I thought about using a context manager(s) here, but it felt strange because the list is updating and several files get replaced, so don't need to be removed. 

(I probably should have used a dict instead of a list though in the event that the save file is appended to the list twice)


> Why not just create a temporary directory, and put all the files you are going to clean up in there, then delete the whole thing in the finally block?

These are files that are present inside an unknown storage backend. I needed to pass in the storage manager specific `put()` and `rm()` so that I didn't have to copy/paste the test for-each storage backend we have and change 2 lines, like the Fetchers.

varlogtim:
To be clear, I think this is super unimportant and doesn't matter much except in certain situations. You can completely ignore this comment.

> Regarding your request for having a StorageInterface: I still think that the "right" abstraction are the ones we already have. You want a Fetcher? We have a Fetcher interface. You want a StorageManager? We have a storage manager.

I may have used the wrong words the first time to explain what I was thinking so I wanted to give it a second shot. (Or maybe you understand my position but I still don't understand fully understand yours and instead of asking for clarity, I am restating my own position... ewww... If this is the case, sorry.)

When we interact with persistent storage we do so as if it was a hierarchical file system, which I will refer to as "HFS abstraction." I am proposing explicitly defining this HFS abstraction. Due to the current structure, we are forced to re-implement the HFS abstraction both when we go to add a new storage back end and when we add new functionality which uses our storage back ends. In the case of adding a new storage back end, we are required to implement specially named `*StorageManager`, `*TensorbordManager` and `*Fetcher` classes, in which the only thing that is different from their respective counterparts, is the HFS abstraction components. When new functionality is added, we must copy and paste the new functionality code for each of the storage back ends, and again just change the HFS abstractions. What we end up with is the exact same HFS abstaction code existing in multiple different function specific places, and the exact same *Functionality code existing in multiple places.

I am not proposing to remove *StorageManager, *TensorboardManager, and/or *Fetcher abstractions; I am just proposing a separation of the HFS abstraction and the functionality of each *Function. This will allow these *Functions, as well as our tests, and any future *Fuctions, can more cleanly and clearly interact along the HFS assumptions that these *Functions are already making internally.

> The interface you outlined to me in DM looks tailor-made for blob storage, but it looks like the shared-fs storage implementations would get worse, not better.

It isn't tailor-made for blob storage, it is made to make any persistent storage look like a hierarchical filesystem and we happen to only have implementations for blob storage at the moment. My intention is to make interacting with any persistent storage, database, blob storage, REST service, etc.. as easy as it is with SharedFS. It will be much easier for us, or users, to add storage back ends, without the need to implement a, potentially ever growing, set of the *Function classes. It also paves the way for an endpoint like, `add_storage_backend(name, code)`, which would create a storage backend without bringing down the cluster and rebuilding.
 
> What would you say to leaving the existing interfaces (Fetcher, StorageManager), but implementing Azure/GCS/S3 as a BlobFetcher or BlobStorageManager? Each of those could take a BlobBackend, with methods as you describe in your StorageInterface, but since the new interface would be specific to blob storage, I think you'd end up with a lot cleaner code.

The seems like a good idea and there are likely Blob specific components to this.

harness/determined/exec/tensorboard.py
rb-determined-ai: > You could me that we know if we have downloaded files or not, so why wait to see if Tensorboard is responding with metrics?

Yes.  I know how it was working but there's a better way now, right?

harness/determined/tensorboard/fetchers/azure.py
```diff
@@ -0,0 +1,78 @@
...
+                self._file_records[filepath] = mtime
+
+        # Download the new or updated files.
+        for filepath in new_files:
+            local_path = os.path.join(self.local_dir, self.container_name, filepath)
+
+            dir_path = os.path.dirname(local_path)
+            os.makedirs(dir_path, exist_ok=True)
+
+            with open(local_path, "wb+") as local_file:
```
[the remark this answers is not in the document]
rb-determined-ai: wait... no `+` doesn't indicate that you are appending, it indicates that you are reading and writing.  Still seems odd here.

harness/determined/tensorboard/fetchers/s3.py
rb-determined-ai: > Do you mean for situation where there are ~10K files, we will do 10 requests per second, to list files?

Yes.  If each file were 50 bytes long (a low estimate given that there's a uuid in the mix) then that would be 1MB/s (8Mbps), which is probably on the order of streaming an hd movie, just to sit there doing nothing.  It's clearly an extreme example, but the point is that this strategy does not scale well.

harness/tests/storage/util.py
rb-determined-ai: > I basically need to remove a dynamic list of files from an unknown storage when I exit the test context for any reason

Not if you run the test in a temporary directory you create at the top of the test.  Then you just need to `shutil.rmtree()` a single directory.

harness/determined/exec/tensorboard.py
varlogtim: What would be the better way? Just launch it? Say the word and I will. It solves another issue which is that I think it hangs when someone is trying to use the profiler.

harness/determined/tensorboard/fetchers/s3.py
varlogtim: Do you want me to change and test this before this goes in? Or can I commit to do it after this lands? I might need to change the tests.

harness/tests/storage/util.py
varlogtim: I don't think you are understanding that these `files_to_remove` exist in S3, Azure or GCS. Would you rather me not test against S3, Azure, and GCS? Or maybe I am not understanding what you mean?

harness/determined/common/storage/azure.py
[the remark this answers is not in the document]
varlogtim: Fixed.

master/pkg/model/storage_config.go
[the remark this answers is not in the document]
varlogtim: Removed the file. Even though I am not sure why this file isn't in the source already.

harness/determined/tensorboard/fetchers/s3.py
rb-determined-ai: Um, I'm actually not 100% certain how you would fix it.  Maybe backoff on the frequency of updates as they start to grow very large?

Anyway, no I think it's probably ok to land as-is.  But I don't feel totally confident... perhaps get another opinion, from @stoksc or @ioga?

harness/determined/tensorboard/fetchers/s3.py
```diff
@@ -0,0 +1,87 @@
...
+        self.storage_paths = storage_paths
+        self._file_records = {}  # type: Dict[str, datetime.datetime]
+
+    def _find_keys(self, prefix: str) -> Generator[Tuple[str, datetime.datetime], None, None]:
+        logger.debug(f"Listing keys in bucket '{self.bucket_name}' with prefix '{prefix}'")
+
+        prefix = urllib.parse.urlparse(prefix).path.lstrip("/")
+
+        list_objs_args = {}  # type: Dict[str, Any]
+        while True:
```
ioga, line 40: could you please use boto paginator machinery instead of doing it manually: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +38,140 @@ def set_s3_region() -> None:
...
+    add_tb_args: List[str],
+) -> int:
+    """Start Tensorboard and look for new files."""
 
-    def still_alive() -> bool:
-        return p.poll() is None
+    # Create local temporary directory
+    script_dir = os.path.dirname(__file__)
+    local_dir = os.path.join(script_dir, "tf_events")
+    os.makedirs(local_dir, mode=0o777, exist_ok=True)
```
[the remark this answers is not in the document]
varlogtim: Talked about this and will write to temp.

harness/determined/tensorboard/fetchers/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    "shared_fs": SharedFSFetcher,
+}  # type: Dict[str, Type[Fetcher]]
+
+
+def build(config: Dict[str, Any], paths: List[str], local_dir: str) -> Fetcher:
+    storage_config = config.get("checkpoint_storage", None)
+    if storage_config is None:
+        raise ValueError("config does not contain a 'checkpoint_stroage' key")
+
+    # XXX Fix me. Read the other code related to this.
```
rb-determined-ai, line 29: not sure what this means, but please resolve it

harness/determined/tensorboard/fetchers/s3.py
ioga: it's hard to judge without knowing the actual usage patterns. 
- ship it today but add more logging to catch the "expensive" cases? i.e. can we add `info` level logging for "large" responses here so we could catch it when it happens?
- strategy-wise - maybe we could backoff and poll more rarely when there're no changes to reduce the idle traffic?
- would more advanced solutions around s3 events -> SNS/SQS be viable?

harness/tests/storage/util.py
rb-determined-ai: talked on zoom, `rm_files` is about storage backend not local files, and this seems reasonable to me

harness/determined/exec/tensorboard.py
varlogtim: So, as soon as one download happens, I will trigger the ready message.

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +41,137 @@ def set_s3_region() -> None:
...
+        logger.warning(f"OK: Tensorboard not responding to HTTP: {exp}")
+
+    except ValueError as exp:
+        logger.warning(f"OK: {exp}")
+
+    except json.JSONDecodeError as exp:
+        logger.warning(f"OK: Could not JSONDecode Tensorboard HTTP response: {exp}")
+
+    if len(tags) != 0 and any([len(v) for v in tags.values()]):
+        logger.info("We are ready to go")
```
rb-determined-ai, line 104: seems like the TRIGGER_READY_MSG is probably sufficient already

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +41,137 @@ def set_s3_region() -> None:
...
+
+def check_for_metrics() -> bool:
+    tensorboard_url = get_tensorboard_url()
+    tags = {}
+    try:
+        # Attempt to retrieve metrics from tensorboard.
+        res = requests.get(tensorboard_url)
+        res.raise_for_status()
+        logger.debug(
+            f"requests.get({tensorboard_url}) -> " "Response.content: {!r}".format(res.content)
```
rb-determined-ai, line 90: Why not combine these strings?  Or can you put a `+` in between the strings?

harness/determined/tensorboard/fetchers/s3.py
varlogtim: I am not sure how/where we would get the info logs for this?

I will commit to looking into investigating if:
1. Tensorboard files are only written to once and not updated (this isn't clear to me right now).
2. Dig into source for `tensorboard/backend/event_processing/data_ingester.py` and see what they do there.
3. Assuming only new files are ever written, that we rename the files so that they are sortable for use with the `StartAfter` flag.
4. Look for corresponding options in the other storage platforms we support.

Also, @rb-determined-ai, this is another example where a generic storage interface would suffer in performance :)

harness/determined/exec/tensorboard.py
varlogtim: This was an annoying MyPy formatting thing. It doesn't like it when I printf a string and bytes, even though I intent to print bytes. Are you cool with that?

varlogtim: Done. This was a weird MyPy issue along with auto-formatting strangeness.

harness/determined/tensorboard/fetchers/s3.py
varlogtim: Oh nice! I don't think this existed the last time I wrote code for boto3.

harness/determined/tensorboard/fetchers/__init__.py
varlogtim: I think I resolved but didn't remove message.

harness/determined/common/_logging.py
```diff
@@ -14,3 +15,16 @@ def set_logger(debug_enabled: bool) -> None:
     formatter = logging.Formatter("%(levelname)s: %(message)s")
     handler.setFormatter(formatter)
     root.addHandler(handler)
+
+
+def get_named_logger(name: str, level: Optional[int] = None) -> logging.Logger:
```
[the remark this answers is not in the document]
varlogtim: done.

harness/determined/exec/tensorboard.py
```diff
@@ -1,12 +1,24 @@
...
+from typing import Any, Dict, List
 
 import boto3
 import requests
 
+import determined.tensorboard.fetchers
+from determined.common._logging import get_named_logger
+
+TENSORBOARD_TRIGGER_READY_MSG = "TensorBoard contains metrics"
+CONFIG_PATH = "/run/determined/tensorboard/experiment_config.json"
```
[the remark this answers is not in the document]
varlogtim: done.

harness/determined/tensorboard/fetchers/azure.py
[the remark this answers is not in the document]
varlogtim: Resolved.

harness/determined/tensorboard/fetchers/s3.py
rb-determined-ai: I'd say:

2. Good idea, but IMO only useful if you find out that they're doing something better than us.  If what they do is as naive as us, then that isn't really good enough, since we are embarking on this mission because we're sick of dealing with tensorboard not behaving very well in corner case (that is, we really need to be better)
3. Not a good idea.  Tensorboard fetchres should be read-only.

That SNS/SQS that Ilia mentioned would be a sweet addition.  I spent some time and arrived poking around [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html#supported-notification-event-types) documentation, and I don't think that would really be meaningful for us to support any time soon.

Also, @varlogtim, why did you mark this comment thread as "resolved" when it is clearly still active?

> I am not sure how/where we would get the info logs for this?

No, he means you should be logging fetches which become very large at the info level, so it's more obvious if we have a problem.   Maybe just log any time we do >1 fetch per second?

master/pkg/model/storage_config.go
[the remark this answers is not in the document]
rb-determined-ai: I removed it, when I removed the old ExperimentConfig object that was hardly in use.  The new one in `pkg/schemas/expconf` has several better properties.

harness/determined/common/_logging.py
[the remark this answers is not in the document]
rb-determined-ai: you say "done" and you marked this as resolved, but you didn't actually handle my biggest complaint, which is that:

> It's configuring formatters and stdout handlers, neither of which should be configured by the library

harness/determined/tensorboard/fetchers/azure.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class AzureFetcher(Fetcher):
+    def __init__(self, storage_config: Dict[str, Any], storage_paths: List[str], local_dir: str):
+        from azure.storage import blob
+
+        connection_string = storage_config.get("connection_string", None)
+        container = storage_config.get("container", None)
+        account_url = storage_config.get("account_url", None)
+        credential = storage_config.get("credential", None)
```
rb-determined-ai, line 19: remove all of these `, None`s

harness/determined/tensorboard/fetchers/azure.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if storage_config.get("connection_string", None):
```
rb-determined-ai, line 21: ditto.  I'm going to stop marking them, please find all of them.

harness/determined/tensorboard/fetchers/azure.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Look at all files in our storage location.
+        for storage_path in self.storage_paths:
+            logger.debug(f"Looking at path: {storage_path}")
+
+            for filepath, mtime in self._list(storage_path):
+                prev_mtime = self._file_records.get(filepath, None)
+
+                if prev_mtime is not None and prev_mtime >= mtime:
+                    logger.debug(f"Blob not new: '{filepath}'")
```
rb-determined-ai, line 56: most of the `logger.debug()` statements in these `fetchers/*` files strike me as "useful while developing the feature but would probably just be noise if I turned logging to debug in real life".

You do more support than I do, so if you feel these should stay I won't press the issue.  But I want you to consider them carefully, because unhelpful logging is not without cost.  This one in particular seems like it would be in a pretty hot loop, but would be particularly useless.

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +40,108 @@ def set_s3_region() -> None:
...
+
+            if not triggered and num_fetched_files > 0:
+                print(TENSORBOARD_TRIGGER_READY_MSG)
+                triggered = True
+
+            num_fetched_files += fetcher.fetch_new()
+            time.sleep(FETCH_INTERVAL)
+
+    except Exception as exp:
+        logger.error(str(exp))
```
rb-determined-ai, line 123: Why even catch the error?  When there is an error you print the error text (but not the stack trace), then you kill the tensorboard and exit with the exit code of the tensorboard process.  If you didn't catch the error, you would instead exit printing an entire stack trace, which seems preferable to me.

I'd much rather see a whole stacktrace than a `NoneType error` or some unhelpful string that could have originated in a variety of different files.

varlogtim: I think it is left over code, which doesn't make sense, from the review process. Removed.

harness/determined/tensorboard/fetchers/azure.py
varlogtim: I agree. Removed.

harness/determined/tensorboard/fetchers/azure.py
varlogtim: Yup, I suppose this is a bad habit.

harness/determined/tensorboard/fetchers/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+}  # type: Dict[str, Type[Fetcher]]
+
+
+def build(config: Dict[str, Any], paths: List[str], local_dir: str) -> Fetcher:
+    storage_config = config.get("checkpoint_storage", None)
+    if storage_config is None:
+        raise ValueError("config does not contain a 'checkpoint_stroage' key")
+
+    storage_type = storage_config.get("type", None)
+    if storage_type is None or storage_type not in _FETCHERS:
```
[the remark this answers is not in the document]
varlogtim: :+1:

harness/determined/tensorboard/fetchers/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+    "gcs": GCSFetcher,
+    "azure": AzureFetcher,
+    "shared_fs": SharedFSFetcher,
+}  # type: Dict[str, Type[Fetcher]]
+
+
+def build(config: Dict[str, Any], paths: List[str], local_dir: str) -> Fetcher:
+    storage_config = config.get("checkpoint_storage", None)
+    if storage_config is None:
+        raise ValueError("config does not contain a 'checkpoint_stroage' key")
```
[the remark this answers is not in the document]
varlogtim: done.

harness/determined/exec/tensorboard.py
```diff
@@ -1,11 +1,25 @@
...
+import tempfile
 import time
-from typing import Callable, List, Tuple
+from typing import Any, Dict, List
 
 import boto3
-import requests
+from packaging import version
+
+from determined.common import _logging, util
```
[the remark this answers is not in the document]
varlogtim: Oh, nice!

harness/determined/common/_logging.py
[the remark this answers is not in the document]
varlogtim: I think your comments more after we discussed that you don't actually want people to know which module/file is doing this logging. I will just use the existing `set_logger()`.

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +40,119 @@ def set_s3_region() -> None:
...
+                # Check if we have reached a timeout without receiving metrics
+                if num_fetched_files == 0 and time.time() > stop_time:
+                    raise RuntimeError("Tensorboard wasn't responsive before the timeout.")
+
+                if responsive and not triggered and num_fetched_files > 0:
+                    print(TENSORBOARD_TRIGGER_READY_MSG)
+                    triggered = True
+
+                time.sleep(FETCH_INTERVAL)
+                responsive = check_tensorboard_responsive()
```
rb-determined-ai, line 134: you probably need `if not responsive: ...` here.  No point in checking tensorboard over and over I think.

harness/determined/exec/tensorboard.py
```diff
@@ -26,121 +40,119 @@ def set_s3_region() -> None:
...
+
+        try:
+            while True:
+                ret_code = tensorboard_process.poll()
+                if ret_code is not None:
+                    raise RuntimeError(f"Tensorboard process died, exit code({ret_code}).")
+
+                # Check if we have reached a timeout without receiving metrics
+                if num_fetched_files == 0 and time.time() > stop_time:
+                    raise RuntimeError("Tensorboard wasn't responsive before the timeout.")
```
rb-determined-ai, line 127: Your if statement and the message you give to the user do not match.

Probably just update the message?

harness/determined/tensorboard/fetchers/shared.py
```diff
@@ -0,0 +1,55 @@
...
+import posixpath
+import shutil
+from typing import Any, Dict, Generator, List, Tuple
+
+from .base import Fetcher
+
+logger = logging.getLogger(__name__)
+
+
+class SharedFSFetcher(Fetcher):
```
rb-determined-ai, line 13: Just had a thought: Why even have a SharedFS Fetcher?  Why not just use the files right in the shared_fs?

harness/determined/exec/tensorboard.py
varlogtim: Ahh, yeah, I reworked this and that slipped through... it was there before somewhere.

harness/determined/tensorboard/fetchers/shared.py
varlogtim: Good-ish reasons:
Consistency.
I can't think of a good reason to directly access `shared_fs`.
I do think that there will be an odd exception in the tensorboard code for shared_fs in which it needs to worry about the storage type.

Not-so-good-reasons:
I would like this to go in by Wednesday. Can we make that change at a later point?

If required for your approval, please tell me and I will make the change.

harness/determined/exec/tensorboard.py
varlogtim: Yup.

harness/determined/tensorboard/fetchers/shared.py
rb-determined-ai: > I can't think of a good reason to directly access shared_fs.
Less code, and less disk io?

> Can we make that change at a later point?
Yes, I think that's fine.

rb-determined-ai (approved):
great work!  especially thank you for writing tests!

PR #3489, opened 2022-01-24
fix: gcp vm image with google guest agent working [DET-6489]

ioga:
## Description

- Use new gcp vm image that is based on a newer base image from google. google guest agent seems to be running fine on it from the first boot, enabling `gcloud compute ssh` usage.

## Test Plan

1. Deploy a new cluster via `det deploy gcp ...`
2. Run a job on it to have a dynamic agent spun up (e.g. `det shell start --config resources.slots=1`
3. Use `det agent list` to get agent name list.
4. Use `gcloud compute ssh AGENT_NAME` to ssh into the node.
5. It should work okay (and not throw `publickey` error).

## Commentary (optional)

For mysterious reasons, systemctl on the previous image won't launch `google-guest-agent.service` on first startup; worked okay on reboot (or if you explicitly start it up). Updating the base image also helped.

## Checklist


coreystatendet (approved):
Ran through test plan successfully.

PR #3492, opened 2022-01-25
feat: add further webui support for adjusting jobs

hamidzr:
## Description

- moving jobs by position in the queue
- moving to another resource pool

## Test Plan

- unit tests
- connect the UI from here to the server built on #3525 
  - move jobs around using queue position option in the UI

## Commentary (optional)

I split part of this out to https://github.com/determined-ai/determined/pull/3482

## Checklist

### DIFF
--- a/webui/react/src/pages/JobQueue/ManageJob.tsx
+++ b/webui/react/src/pages/JobQueue/ManageJob.tsx
@@ -1,7 +1,8 @@
-import { Form, FormInstance, Input, List, Modal, Typography } from 'antd';
+import { Form, FormInstance, Input, List, Modal, Select, Typography } from 'antd';
 import React, { ReactNode, useCallback, useMemo, useRef } from 'react';
 
 import Badge, { BadgeType } from 'components/Badge';
+import { useStore } from 'contexts/Store';
 import { columns } from 'pages/JobQueue/JobQueue.table';
 import { updateJobQueue } from 'services/api';
 import * as api from 'services/api-ts-sdk';
@@ -11,6 +12,7 @@ import { moveJobToPositionUpdate, orderedSchedulers } from 'utils/job';
 import { floatToPercent, truncate } from 'utils/string';
 
 import css from './ManageJob.module.scss';
+const { Option } = Select;
 
 interface Props {
   job: Job;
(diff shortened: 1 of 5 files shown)


hamidzr:
brought in a fix from the target branch for linit-go

webui/react/src/utils/job.test.ts
```diff
@@ -1,4 +1,5 @@
-import { CommandType, JobType } from 'types';
+import * as Api from 'services/api-ts-sdk';
```
hkang1, line 1: sweet sweet unit tests 👍 👍 👍

webui/react/src/utils/job.ts
```diff
@@ -43,15 +43,64 @@ export const orderedSchedulers = new Set(
   [ Api.V1SchedulerType.PRIORITY, Api.V1SchedulerType.KUBERNETES ],
 );
 
-export const moveJobToPositionUpdate = (jobId: string, position: number): Api.V1QueueControl => {
+export const moveJobToPositionUpdate = (
+  jobs: Job[],
+  jobId: string,
+  position: number,
+): Api.V1QueueControl | undefined => {
```
hkang1, line 50: **question:** Idk a good answer to this, but this might come down to a design decision. I'm not sure if our core utility functions should be able to throw errors, especially when they are used in the React settings. The risk is that the users use them outside of a try/catch block and causes an uncaught error, but they may not encounter that because they didn't force the situation to cause the runtime error.

Unfortunately, Typescript currently doesn't have a type to indicate that the function can throw errors.

wdy think about using comments above functions to indicate that the function can throw errors and hopefully the dev can see the function hinting to see that it can throw errors. https://stackoverflow.com/a/66249151/5402432 see example below:
Screen Shot 2022-01-31 at 11 09 57 AM

hkang1 (approved):
PR looks good! Are there screenshots that are specific to this PR that can be added to help the reviewer to quickly catch on to the feature introduced? A couple of questions and suggestions, all non-blocking.

webui/react/src/utils/job.ts
hamidzr: we gotta check for invalid inputs, it'd be good if we could prevent users from passing invalid values in.. maybe if we exposed a separate validator? but then how do we guarantee that the utility func isn't used without it.

we could always do it like how c or go does which is to return either a return code or a `(value, error)` return type. where the user needs to check if the error is set or not. that can be checked and enforced by typescript. 
as an alternative, the comments sound good.

webui/react/src/pages/JobQueue/ManageJob.tsx
```diff
@@ -21,15 +23,47 @@ interface Props {
...
+  position: string;
   priority?: string;
-  resourcePool?: string;
+  resourcePool: string;
   weight?: string;
 }
 
-const ManageJob: React.FC<Props> = ({ onFinish, selectedRPStats, job, schedulerType }) => {
+const formValuesToUpdate = (
+  values: FormValues,
```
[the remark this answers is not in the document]
hamidzr: yeah I think that makes sense. do we expect antd to parse them just off of a type hint?

webui/react/src/pages/JobQueue/ManageJob.tsx
```diff
@@ -21,15 +23,47 @@ interface Props {
 }
 
 interface FormValues {
-  position?: string;
+  position: string;
```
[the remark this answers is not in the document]
hamidzr: I guess the Job type would be a good place to describe some of these. we already get 

```
    /**
     * The job priority in priority scheduler.
     * @type {number}
     * @memberof V1Job
     */
    priority?: number;
    /**
     * The job weight in fairshare scheduler.
     * @type {number}
     * @memberof V1Job
     */
    weight?: number;
 ```

I can add some more description here. for the other values at least

File: webui/react/src/pages/JobQueue/ManageJob.tsx (hunk 3 above)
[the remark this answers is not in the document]
Comment (hamidzr) [reply webui/react/src/pages/JobQueue/ManageJob.tsx:33]: I think I gave that a try. let me double check

File: webui/react/src/utils/job.ts (hunk 2 above)
Comment (hamidzr) [reply webui/react/src/utils/job.ts:50]: It doesn't show up in my ide when I add it like this

```
diff --git a/webui/react/src/utils/job.ts b/webui/react/src/utils/job.ts
index 9a9bc5569..adb143ecd 100644
--- a/webui/react/src/utils/job.ts
+++ b/webui/react/src/utils/job.ts
@@ -43,6 +43,10 @@ export const orderedSchedulers = new Set(
   [ Api.V1SchedulerType.PRIORITY, Api.V1SchedulerType.KUBERNETES ],
 );
 
+/*
+Create the update request based on a given position for a job.
+@throws {DetError}
+*/
 export const moveJobToPositionUpdate = (
   jobs: Job[],
   jobId: string,

```

ah I wasn't following the correct format. 
https://jsdoc.app/tags-throws.html

I bet there is an eslint rule for that

hkang1: > we could always do it like how `C` or `Go` does which is to return either a return code or a `(value, error)` return type. where the user needs to check if the error is set or not. that can be checked and enforced by typescript. as an alternative, the comments sound good.
>
This sounds like a pretty promising approach `(value, error)` tuple.

PR #3493, opened 2022-01-25
chore: add telemetry for data layer usage [DET-5338]

azhou-determined:
## Description

adds a metric for recording data layer usage. used for determining whether or not we should deprecate yogadl dependency

## Checklist

### DIFF
--- a/harness/determined/_data_layer/_data_layer.py
+++ b/harness/determined/_data_layer/_data_layer.py
@@ -12,6 +12,7 @@ from yogadl import storage, tensorflow
 import determined as det
 from determined import _core
 from determined.common import check
+from determined.common.api import analytics
 
 
 def init_container_storage_path(configured_storage_path: Optional[str]) -> pathlib.Path:
@@ -174,6 +175,8 @@ class _CacheableDecorator:
             )
         self._decorator_used = True
         dataset_version += "_train" if self._training else "_val"
+
+        analytics.send_analytics("data_layer_cache_dataset", {})
 
         def _wrap(make_dataset_fn: Callable) -> Callable:
             @functools.wraps(make_dataset_fn)


harness/determined/_data_layer/_data_layer.py
```diff
@@ -175,6 +176,13 @@ def cache_dataset(
         self._decorator_used = True
         dataset_version += "_train" if self._training else "_val"
 
+        # Track usage
+        # Temporary measure to determine whether to deprecate data layer/yogadl dependency
+        analytics.send_analytics(
+            "data_layer_cache_dataset",
+            {"dataset_version": dataset_version, "dataset_id": dataset_id},
```
azhou-determined, line 183: this additional data might not be necessary. is a binary used/not used metric enough for determining whether or not to deprecate yogadl/data layer?

rb-determined-ai: Personally I'd rather we not collect these bits of info as I think it's slightly invasive but without any clear benefit to our organization.

harness/determined/_data_layer/_data_layer.py
```diff
@@ -175,6 +176,13 @@ def cache_dataset(
         self._decorator_used = True
         dataset_version += "_train" if self._training else "_val"
 
+        # Track usage
+        # Temporary measure to determine whether to deprecate data layer/yogadl dependency
```
rb-determined-ai, line 180: I'd recommend avoiding placing temporal comments in the code, because they have a habit of becoming out of date.  Temporal things make more sense in the commit message, which is more point-in-time by nature.

harness/determined/_data_layer/_data_layer.py
azhou-determined: makes sense. removed.

harness/determined/_data_layer/_data_layer.py
azhou-determined: removed this comment and edited the commit message.

rb-determined-ai (approved):

rb-determined-ai:
nit: can you use proper grammar in the body of the commit message? (not the title, which we decided on for some reason)

Also, if you can wrap the commit message body 72 width text, that would be ideal.  We don't have that written as a standard anywhere but it is pretty common to see in other repos.

PR #3494, opened 2022-01-26
fix: increment runID during allocation

eecsliu:
## Description
Previously, the runID of the trial was incremented in the `buildTaskSpec` function. This does not cover the case where we switch resource pools and need to restart the task on a different pool. In this case, we don't enter the function or increment the ID, leading to an error. Moving the increment to the task allocation function ensures that we increment the ID every time we allocate a task, which is the more correct behavior.

## Checklist


stoksc (approved):
lgtm

stoksc:
In the commit message pls describe the bug a bit for posterity

PR #3495, opened 2022-01-26
feat: create a new cluster page [DET-6403]

gt2345:
## Description

Create a new cluster page under `/clusters`
Move `Cluster Logs` to a tab on this page

## Test Plan

Navigate to `/clusters` to verify `Master Log` tab work.

Demo:
Screen Shot 2022-01-30 at 11 55 18 AM


webui/react/src/routes/index.ts
```diff
@@ -23,6 +24,7 @@ import { RouteConfig } from './types';
 const routeComponentMap: Record<string, FC> = {
   cluster: Cluster,
   clusterLogs: ClusterLogs,
+  clusters: ClusterOverview,
```
hamidzr, line 27: nit: matching comp name with the entry name

webui/react/src/pages/Clusters/ClusterLogs.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { useCallback } from 'react';
```
hamidzr, line 1: question: is there an ask for recreating this specific page? would just importing the existing page do the trick? It might help us avoid these two getting diverged in the meantime with bug fixes, new features, other changes, etc

hamidzr: We might want to copy this as it's done here if we have immediate plans to change this logs page for the new design

hamidzr (approved):
I have a question on whether we need to recreate the master logs or an import would suffice otherwise it looks good to me.

hamidzr (approved):
lgtm

hamidzr:
thanks for the quick update.

PR #3497, opened 2022-01-26
chore: Use nullable types in PatchExperiment [DET-6486]

mapmeld:
## Description

Replace the field masking / multiple queries format of `PATCH /api/v1/experiment/:id` and instead, following the Models API, accept a `PatchExperiment` body with nullable types. This allows us to send a partial update `{ "id": 101, "name": "new_name" }` and the other fields will be unchanged.

## Checklist


harness/determined/common/api/bindings.py
```diff
@@ -3069,6 +3051,44 @@ def to_json(self) -> typing.Any:
...
+        self.forkedFrom = forkedFrom
+
+    @classmethod
+    def from_json(cls, obj: Json) -> "v1PatchExperiment":
+        return cls(
+            id=obj["id"],
+            description=obj["description"] if obj.get("description", None) is not None else None,
+            labels=obj["labels"] if obj.get("labels", None) is not None else None,
+            name=obj["name"] if obj.get("name", None) is not None else None,
+            notes=obj["notes"] if obj.get("notes", None) is not None else None,
```
hamidzr, line 3078: unrelated couldn't `obj["notes"] if obj.get("notes", None) is not None else None,` be just `obj.get("notes", None)` did I mess this up in the genrator?

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -90,6 +91,25 @@ message Experiment {
   google.protobuf.Int32Value forked_from = 16;
 }
 
+// PatchModel is a partial update to an experiment with only id required.
+message PatchExperiment {
```
hamidzr, line 95: 👍🏼

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -90,6 +91,25 @@ message Experiment {
...
+  // The description of the experiment.
+  google.protobuf.StringValue description = 2;
+  // Labels attached to the experiment.
+  google.protobuf.ListValue labels = 3;
+  // The experiment name.
+  google.protobuf.StringValue name = 4;
+  // The experiment notes.
+  google.protobuf.StringValue notes = 5;
+  // Original id of a forked or continued experiment.
+  google.protobuf.Int32Value forked_from = 6;
```
hamidzr, line 110: do we want this field?

hamidzr (commented):
the approach lgtm. I'll check it out more and get back to you in a bit

harness/determined/common/api/bindings.py
mapmeld: `None) is not None else None,` appears in a bunch of places in the bindings file, agree it would be cleaner if it didn't generate this way

proto/src/determined/experiment/v1/experiment.proto
mapmeld: thanks, removing it ✅

hamidzr (approved):
already we get some validation which is great. It'd be great if it also prevented users from sending unacceptable patches and not accepting inaccurate values:

could we add this back?
image

mapmeld:
Proto (as of v3) drops any unknown fields before passing it to our Go code ( https://github.com/protocolbuffers/protobuf/issues/272 , https://github.com/golang/protobuf/issues/1390 ).  The best way that someone can know that their unexpected change didn't happen, at the moment, is to look at the Experiment which we returned and that the field isn't added or updated.

mapmeld:
oh I should look into this https://developers.google.com/protocol-buffers/docs/proto3#unknowns

>  in version 3.5 we reintroduced the preservation of unknown fields to match the proto2 behavior. In versions 3.5 and later, unknown fields are retained during parsing and included in the serialized output

hamidzr:
aah 34

> oh I should look into this https://developers.google.com/protocol-buffers/docs/proto3#unknowns
> 
> > in version 3.5 we reintroduced the preservation of unknown fields to match the proto2 behavior. In versions 3.5 and later, unknown fields are retained during parsing and included in the serialized output

is that how fieldmask retained unknown paths? last image here https://github.com/determined-ai/determined/pull/3497#pullrequestreview-865452597
?

looks like all the generated structs in go have an unknownFileds attached to them but that's internal

PR #3499, opened 2022-01-26
chore: update polling functions to be async where applicable [DET-6601]

hkang1:
## Description

This is a continuation of [this PR](https://github.com/determined-ai/determined/pull/3481). The previous PR addresses the polling of `agents`. This PR focuses on applying the same work for the rest of the WebUI where applicable. The one call out area where this has NOT been applied is the `Task Logs` page, because there already is a PR to migrate that page over to a new pattern which will resolve it there.

PROBLEM
There are a few places through the WebUI where the polling is occurring without the blocking behavior, where the polling mechanism waits for the previous call to resolve before making another one. Here is an example of what the network calls look like:
Screen Shot 2022-01-26 at 10 47 58

SOLUTION:
To fix this, ensure the polling function being passed into the polling hook is an async function.

## Test Plan

To test a particular page whether the blocking behavior is proper or not:
- [ ] Load a page that has polling (listed below)
- [ ] open up browser console and navigate to the `Network` tab
- [ ] change the throttle from `No throttling` to `Custom` and define a new throttle profile as such: 
Screen Shot 2022-01-26 at 12 21 02 PM
- [ ] look to ensure that the polling endpoints don't get called again if there are previous pending calls to them

List of polling pages:
- [ ] Dashboard
- [ ] Experiment List
- [ ] Task List
- [ ] Model Registry

## Checklist

### DIFF
--- a/webui/react/src/pages/Dashboard.tsx
+++ b/webui/react/src/pages/Dashboard.tsx
@@ -136,11 +136,13 @@ const Dashboard: React.FC = () => {
     }
   }, [ canceler ]);
 
-  const fetchAll = useCallback(() => {
-    fetchUsers();
-    fetchExperiments();
-    fetchActiveExperiments();
-    fetchTasks();
+  const fetchAll = useCallback(async () => {
+    await Promise.allSettled([
+      fetchUsers(),
+      fetchExperiments(),
+      fetchActiveExperiments(),
+      fetchTasks(),
+    ]);
   }, [ fetchUsers, fetchExperiments, fetchActiveExperiments, fetchTasks ]);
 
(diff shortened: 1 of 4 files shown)


hamidzr (approved):
lgtm thanks for fixing these

separate concern: seeing these I think we can augment the error that gets thrown through a polling iteration failing. maybe those are the ones we want to always make sure are set to slient and are not `userTriggered`

hamidzr:
I was just thinking how can we find out if there are any that were missed here and I think in use polling we can time how long the one iteration took and if it's below an arbitrary set network limit we alert ourselves.

PR #3500, opened 2022-01-26
feat: stopping and starting polling based on page visibility [DET-6604]

hkang1:
## Description

PROBLEM
Polling can potentially be an expensive operation, using up power and bandwidth.

SOLUTION
Use the Page Visibility API that is supported by most modern browsers to store the state of whether the current page/tab is visible (aka user is focused on the tab/page or not). Using that state, to stop polling when the page is hidden and start polling if the page is visible and previously polling (meaning if it was stopped prior to the page becoming hidden, then when the page becomes visible, don't attempt to reactivate the polling).

## Test Plan

- [ ] on the WebUI, as long as the navigation is visible, it polls against the `/agents` endpoint. Simply monitor the `Network` tab for `/agents` endpoint to ensure it is polling. Count how many `agents` calls there are, then quickly change to a different tab for a while (agents polls every 5 seconds, so maybe come back in 30 seconds). Then go back to the original page and check to make sure there are no new `agents` calls (or possibly just one more than what you previously counted because it might have called it before you were able to check it).

## Checklist

### DIFF
--- a/docs/release-notes/3500-page-visibility-polling.txt
+++ b/docs/release-notes/3500-page-visibility-polling.txt
@@ -1,0 +1,8 @@
+:orphan:
+
+**New Features**
+
+-  Web: Polling can potentially be an expensive operation. To reduce unnecessary polling, we use
+   modern browser Page Visiblity API to detect changes in page visibility, which maps to whether the
+   user is activately focused on the page/tab or not. If the page is hidden, ongoing polling are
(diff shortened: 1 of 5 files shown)


gt2345 (approved):
Verified working with Chrome and Firefox

hkang1:
> Verified working with Chrome and Firefox

Thanks for verifying on multiple browsers @gt2345!

PR #3502, opened 2022-01-26
fix: Progress on completed experiment to be reported as 1

mapmeld:
## Description

This fixes an issue with the API and CLI reporting '0%' complete on experiments (including completed ones) where we intend for this output to be blank/null.

## Checklist


master/internal/api_experiment.go
```diff
@@ -245,7 +245,7 @@ func (a *apiServer) GetExperiments(
 		apiv1.GetExperimentsRequest_SORT_BY_END_TIME:    "end_time",
 		apiv1.GetExperimentsRequest_SORT_BY_STATE:       "state",
 		apiv1.GetExperimentsRequest_SORT_BY_NUM_TRIALS:  "num_trials",
-		apiv1.GetExperimentsRequest_SORT_BY_PROGRESS:    "progress",
+		apiv1.GetExperimentsRequest_SORT_BY_PROGRESS:    "COALESCE(progress, 0)",
```
stoksc, line 248: wouldnt we want coalesce(progress, 1)? seems like nulls is closer to 1 than 0

stoksc (commented):
this lgtm, but thoughts on my suggestion in slack to just stop nulling out progress when experiments complete?

master/internal/api_experiment.go
mapmeld: We are now setting progress to 1 on completed experiments so I think this is now a moot point
but at the time, I believe the both unstarted and completed experiments could have progress=NULL and wanted them to follow the existing sort behavior

master/internal/experiment.go
```diff
@@ -279,8 +279,15 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 
 	// Experiment shutdown logic.
 	case actor.PostStop:
-		if err := e.db.SaveExperimentProgress(e.ID, nil); err != nil {
-			ctx.Log().Error(err)
+		if e.State != model.CompletedState {
```
stoksc, line 282: if thought we would want to set this to 1.0 if it _was_ completed. and that otherwise we would not set it at all (with the thought that, if an experiment fails at 12% progress, it is much more useful to see 12% progress than nil).

stoksc, line 282: if thought we would want to set this to 1.0 if it _was_ completed. and that otherwise we would not set it at all (with the thought that, if an experiment fails at 12% progress, it is much more useful to see 12% progress than nil).

stoksc: crazy github glitch here

mapmeld: oh, you're right, I wrote this one backwards, will fix and test both cases to figure out what happened here

master/internal/api_experiment.go
stoksc: ah, yeah. i would also call that very strange, that we create them with NULL not 0 progress. idk.

master/internal/experiment.go
mapmeld: ok in my latest commit I have all experiments created with `parseCreateExperiment` start at progress = 0

and we accept Completed and StoppingCompleted here

master/internal/db/postgres_experiments.go
```diff
@@ -1132,9 +1132,9 @@ func (db *PgDB) AddExperiment(experiment *model.Experiment) (err error) {
 		}
 		err := namedGet(tx, &experiment.ID, `
 	INSERT INTO experiments
-	(state, config, model_definition, start_time, end_time, archived, parent_id,
+	(state, config, model_definition, start_time, end_time, archived, parent_id, progress,
```
stoksc, line 1135: Not your PR's issue, but it's really strange progress isn't just in the model to me.

master/internal/experiment.go
```diff
@@ -284,8 +284,15 @@ func (e *experiment) Receive(ctx *actor.Context) error {
...
 	case actor.PostStop:
-		if err := e.db.SaveExperimentProgress(e.ID, nil); err != nil {
-			ctx.Log().Error(err)
+		if e.State == model.CompletedState || e.State == model.StoppingCompletedState {
+			setProgress := 1.0
+			if err := e.db.SaveExperimentProgress(e.ID, &setProgress); err != nil {
+				ctx.Log().Error(err)
+			}
+		} else {
+			if err := e.db.SaveExperimentProgress(e.ID, nil); err != nil {
```
stoksc, line 293: why not just do nothing here? i think i mentioned this, it seems really useful to see an experiment failed at 38% not nil. if the UI doesn't want to expose that, it's fine to me, but I see no reason to erase useful info.

stoksc (commented):
have some minor comments about the decision to null out progress, is all

master/internal/experiment.go
mapmeld: That does make sense. In my latest commit I remove this call and update the release note.

master/internal/experiment.go
```diff
@@ -284,8 +284,11 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 
 	// Experiment shutdown logic.
 	case actor.PostStop:
-		if err := e.db.SaveExperimentProgress(e.ID, nil); err != nil {
-			ctx.Log().Error(err)
+		if e.State == model.CompletedState || e.State == model.StoppingCompletedState {
+			setProgress := 1.0
+			if err := e.db.SaveExperimentProgress(e.ID, &setProgress); err != nil {
```
stoksc, line 289: fyi there is a `ptrs` package in determined for this quirk. so you can do `ptrs.FloatPtr(1.0)`.

mapmeld: ✅  oh! ok I'll use that here and in the future

PR #3505, opened 2022-01-27
chore: print trial logs if trial ends with 0 steps

mpkouznetsov:
## Description
In e2e distributed tests, print trial logs when a trial is completed but has 0 steps.
While trying to detect the cause of flakiness of test_unets_tf_keras_distributed ( https://determinedai.atlassian.net/browse/DET-6598 ) I realized we were not retaining (printing) the trial logs when the trial ends in a COMPLETED state but something goes wrong and it has 0 steps. This was the case in the 3 failures of this test: the assertion that then number of steps is positive failed but pytest logs did not have sufficient information to understand what went wrong.

## Commentary (optional)

A more general solution would be to modify our CI code to retain trial logs as build artifacts.

## Checklist


stoksc (approved):
lgtm

PR #3508, opened 2022-01-27
feat: create user settings modal

johnkim-det:
## Description

New modal that displays the user's avatar and username.

## Test Plan

- [ ] Navigation SideBar's dropdown menu displays an additional option called "Settings"
Screen Shot 2022-01-27 at 10 05 13 AM

- [ ] Clicking "Settings" opens a modal that displays the username and avatar that was displayed in the Navigation SideBar.
usersettingsmodal

## Commentary (optional)

For now, this modal only displays the user's avatar and username, but the intention is to add functionality to this modal to allow users to update user-specific settings (avatar, username, password, etc.)

## Checklist


webui/react/src/components/UserSettings.tsx
```diff
@@ -0,0 +1,28 @@
...
+      </div>
+      <div className={css.field}>
+        <span className={css.label}>Username</span>
+        <span className={css.value}>{username}</span>
+      </div>
+    </div>
+  );
+};
+
+export default UserSettings;
```
hkang1, line 28: **suggestion:** What do you think about moving this and the module SCSS files over to `/src/hooks/useModal/useModalUserSettings/` just because it won't really be a shared component and really only used by the modal hook itself.

hkang1 (approved):
LGTM! One suggestion on proposing moving the JSX that is used within the modal hook.

PR #3511, opened 2022-01-27
chore: remove webui tests [DET-6610]

hkang1:
## Description

Once CI is stable without `test-e2e-webui`, drop `webui/tests`.

## Test Plan

- [ ] check CI all functions properly
- [ ] ensure webui makefile commands all still work
- [ ] check that there are no references to old webui/tests anywhere (docs included)

## Checklist


webui/tests/bin/try-for-flakes.sh
```diff
@@ -1,26 +0,0 @@
-#!/bin/bash -e
```
hamidzr, line 1: keep this one? maybe move it to react dir. it should mostly apply to react unit tests as well for reproducing flakes.

hkang1: yeah that's a good call. I've updated the command to the unit tests so a simple `./bin/try-for-flakes.sh` should run just fine under `react`

hkang1:
preserved the `try-for-flake.sh` script for react unit testing with [this PR](https://github.com/determined-ai/determined/pull/3560)

hkang1:
@hamidzr NOT in a hurry with this PR, I don't plan to land this within the next 2 weeks (at least)

hkang1:
> @hamidzr NOT in a hurry with this PR, I don't plan to land this within the next 2 weeks (at least)

Actually we might want to land it sooner (maybe in a week, when I get back) just because there are a lot of dependencies under `webui/tests` that are high security alerts, etc.

hamidzr (approved):
looks straight forward. we need to also remove mentions of it from the circle ci? I'm assuming we've already done that otherwise the tests would be failing without these.

I don't think we need to wait and see how the ci does without this directory. it should be fine. If now react is the only dir we have under webui we might want to remove the extra level unless we think we might add something there in the future.

hkang1:
> looks straight forward. we need to also remove mentions of it from the circle ci? I'm assuming we've already done that otherwise the tests would be failing without these.
> 
👍  yeah, we dropped it from CI already.

> I don't think we need to wait and see how the ci does without this directory. it should be fine. If now react is the only dir we have under webui we might want to remove the extra level unless we think we might add something there in the future.
>
I am seriously considering that. I was in the same boat as you, trying to think if something other than react will go under `webui` but couldn't come up with anything...

hamidzr:
> I am seriously considering that. I was in the same boat as you, trying to think if something other than react will go under `webui` but couldn't come up with anything...

it would affect git history for every single line of webui code. IMO let's give it some time if there is no rush.

hamidzr:
> > I am seriously considering that. I was in the same boat as you, trying to think if something other than react will go under `webui` but couldn't come up with anything...
> 
> it would affect git history for every single line of webui code. IMO let's give it some time if there is no rush.

looking back at it I'm not sure if I was correct but we had other prs after this look into this change.

PR #3513, opened 2022-01-28
feat: add BYOL example

coreystatendet:
## Description

Add BYOL training example.  [DET-6450]

## Test Plan

Run example configurations and ensure convergence.  (In progress for ImageNet).

## Checklist

[DET-6450]: https://determinedai.atlassian.net/browse/DET-6450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/computer_vision/byol_pytorch/byol_pytorch.py
```diff
@@ -0,0 +1,294 @@
+"""
+Adapted from https://github.com/lucidrains/byol-pytorch/blob/master/byol_pytorch/byol_pytorch.py
```
liamcli, line 2: if this file is the same as the source, i prefer we git clone and checkout a specific commit into the working directory as part of the startup-hook so we don't have to maintain this portion.

examples/computer_vision/byol_pytorch/lars.py
```diff
@@ -0,0 +1,174 @@
+"""
+Layer-wise adaptive rate scaling for SGD in PyTorch.
+Adapted from https://github.com/untitled-ai/self_supervised/blob/master/lars.py
```
liamcli, line 3: same here.  if this file is the same as the source, i prefer we git clone and checkout a specific commit into the working directory as part of the startup-hook so we don't have to maintain this portion.

examples/computer_vision/byol_pytorch/config/distributed-imagenet.yaml
```diff
@@ -0,0 +1,92 @@
+name: imagenet_byol_distributed
+entrypoint: model_def:BYOLTrial
+records_per_epoch: 1271158
+resources:
+  slots_per_trial: 32
+min_validation_period:
+  epochs: 100
+
+profiling:
```
liamcli, line 9: glad you're already using the profiler!

examples/computer_vision/byol_pytorch/config/const-cifar10.yaml
```diff
@@ -0,0 +1,78 @@
...
+    gaussian_blur_prob: 0.1
+    gaussian_blur_kernel_size: 3
+    gaussian_blur_min_std: 0.1
+    gaussian_blur_max_std: 2.0
+    solarization_prob: 0.2
+
+hyperparameters:
+  backbone_name: resnet18
+  global_batch_size: 256
+  classifier:
```
liamcli, line 52: glad to see you using nested hps!

examples/computer_vision/byol_pytorch/data.py
```diff
@@ -0,0 +1,301 @@
...
+import os
+import random
+from typing import Callable, List, Tuple
+
+import torch
+import torch.nn as nn
+from torchvision.datasets import CIFAR10, ImageFolder, STL10
+import torchvision.transforms as T
+from torch.utils.data import Dataset
+
```
liamcli, line 13: i like what you did here.  let's keep this in mind for the future if we work on utility libraries.

examples/computer_vision/byol_pytorch/model_def.py
```diff
@@ -0,0 +1,367 @@
...
+class BYOLTrial(PyTorchTrial):
+    def __init__(self, context: PyTorchTrialContext) -> None:
+        self.context = context
+        self.hparams = AttrDict(self.context.get_hparams())
+        self.data_config = AttrDict(self.context.get_data_config())
+        self.hparams.total_epochs = self.context.get_experiment_config()["searcher"][
+            "max_length"
+        ]["epochs"]
+        self.rank = self.context.distributed.get_rank()
+        if self.data_config.use_rank_for_download_dir:
```
liamcli, line 110: i like using a filelock more since it will only download once per instance instead of once per slot:
https://github.com/determined-ai/determined/blob/master/examples/computer_vision/detr_coco_pytorch/model_def.py#L98

examples/computer_vision/byol_pytorch/model_def.py
```diff
@@ -0,0 +1,367 @@
...
+        """
+        Create classifier networks and optimizers, one for each candidate LR.
+        """
+        backbone_metadata = BACKBONE_METADATA_BY_NAME[self.hparams.backbone_name]
+        dataset_metadata = DATASET_METADATA_BY_NAME[self.data_config.dataset_name]
+        # Ensure learning rates are unique.
+        assert len(self.hparams.classifier.learning_rates) == len(
+            set(self.hparams.classifier.learning_rates)
+        )
+        self.cls_models = [
```
liamcli, line 184: later on, we loop through the train classifier dataloader and then loop through these models for each batch.  i wonder if it's better to loop through the lrs and then the dataloader to train one model at a time in which case we can just create a single classifier and model here and reset between lrs.

examples/computer_vision/byol_pytorch/model_def.py
```diff
@@ -0,0 +1,367 @@
...
+        # We combine these two datasets, and then use a custom reducer to calculate the final result.
+        lr_val_dataset = build_dataset(
+            self.data_config,
+            self.download_dir,
+            DatasetSplit.CLS_VALIDATION,
+        )
+        test_dataset = build_dataset(
+            self.data_config, self.download_dir, DatasetSplit.TEST
+        )
+        combined = JointDataset([lr_val_dataset, test_dataset], ["lr_val", "test"])
```
liamcli, line 272: are the lr_val_dataset and test_dataset always the same size?

examples/computer_vision/byol_pytorch/README.md
```diff
@@ -0,0 +1,14 @@
+# Bootstrap Your Own Latent
```
liamcli, line 1: let's follow the README format in other examples to outline the file structure of this example, instructions on how to run, and expected results.

liamcli (changes requested):
Let me know if you questions about any comments.  Will take a closer look after this round.  Looking really good!

neilconway:
[clabot:check]

examples/computer_vision/byol_pytorch/byol_pytorch.py
coreystatendet: There were changes, but I was able to add some workarounds for compatibility with the `pip install` version.

examples/computer_vision/byol_pytorch/lars.py
coreystatendet: Done.

examples/computer_vision/byol_pytorch/model_def.py
coreystatendet: This is awesome, wasn't aware of this library!  Done.

examples/computer_vision/byol_pytorch/model_def.py
coreystatendet: Unfortunately, I can't see how to make this work.  All five trained classifiers are currently used in `evaluate_batch` after being trained in `on_validation_epoch_start`.  To keep only one model in memory, we'd have to be able to do the `on_validation_epoch_start` loop and the main `evaluate_batch` loop in between each reset.  Open to recommendations.

examples/computer_vision/byol_pytorch/model_def.py
coreystatendet: No, in general they are different sizes.

examples/computer_vision/byol_pytorch/README.md
coreystatendet: Mostly done.  Still working on ImageNet results.

coreystatendet:
Remaining:
- [ ] Imagenet results.
- [ ] Add example tests.
- [ ] Add release note.

examples/computer_vision/byol_pytorch/utils.py
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Callable, Dict, TypeVar
+
+import torch.nn as nn
+
+A = TypeVar("A")
```
liamcli, line 5: This is cool for generic types.

examples/computer_vision/byol_pytorch/model_def.py
liamcli: I see, yeah what I suggested wouldn't work since we want to evaluate all 5 classifiers on the test set in the evaluate_batch loop.

examples/computer_vision/byol_pytorch/model_def.py
```diff
@@ -0,0 +1,423 @@
...
+    def _train_classifier_batch(
+        self, batch: TorchData, epoch_idx: int, batch_idx: int
+    ) -> Dict[str, torch.Tensor]:
+        imgs, labels = batch
+        embeddings = self.byol_model.forward(
+            imgs, return_embedding=True, return_projection=False
+        ).detach()
+        # May be called from validation, so need to enable gradients.
+        losses = {}
+        with torch.enable_grad():
```
liamcli, line 328: i'm not sure if this would have the same behavior as setting module to train for something like dropout.

examples/computer_vision/byol_pytorch/reducers.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        test_correct_by_param = reduce(
+            lambda x, y: merge_dicts(x, y, lambda a, b: a + b), test_correct_by_param
+        )
+        test_ct = sum(test_ct)
+        max_val_param = max(val_correct_by_param, key=val_correct_by_param.get)
+        return test_correct_by_param[max_val_param] / test_ct
+
+
+class AvgReducer(MetricReducer):
```
liamcli, line 54: we also allow passing a function to `wrap_reducer` which can replace this average custom reducer:
`self.context.wrap_reducer(lambda x: sum(x)/len(x))`

liamcli (approved):

liamcli:
You'll need to rebase off of master to fix the failing unit test.

examples/computer_vision/byol_pytorch/reducers.py
coreystatendet: Awesome, switched to this.

PR #3515, opened 2022-01-28
chore: fix spacing in ui update message

hamidzr:
## Test Plan

before 
image
image

## Checklist


johnkim-det:
looks good. minor suggestion is to prefer to not use &nbsp; (since we don't care about non-breaking behavior) and instead either use {" "} or have some of the preceding text and the <Link> tag on the same line: `our <Link>`

johnkim-det (approved):
prefer not using &nbsp;

hamidzr:
I hadn't thought of that.. I'll get this updated in another pr. (or feel free to do so if you see it :) )

PR #3521, opened 2022-01-28
chore: convert antd notification to console.error for dev error handling (error boundary) [DET-6618]

hkang1:
## Description

Make the DEV "silent" notifications less in-your-face and move them into the browser console as `console.error`.

## Checklist

### DIFF
--- a/webui/react/src/App.tsx
+++ b/webui/react/src/App.tsx
@@ -8,6 +8,7 @@ import PageMessage from 'components/PageMessage';
 import Router from 'components/Router';
 import Spinner from 'components/Spinner';
 import StoreProvider, { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
+import useErrorBoundary from 'hooks/useErrorBoundary';
 import { useFetchInfo } from 'hooks/useFetch';
 import useKeyTracker, { KeyCode, keyEmitter, KeyEvent } from 'hooks/useKeyTracker';
 import usePageVisibility from 'hooks/usePageVisibility';
@@ -40,6 +41,7 @@ const AppView: React.FC = () => {
   useKeyTracker();
   usePageVisibility();
   useRouteTracker();
+  useErrorBoundary();
 
   // Poll every 10 minutes
   usePolling(fetchInfo, { interval: 600000 });
--- a/webui/react/src/hooks/useErrorBoundary.test.tsx
+++ b/webui/react/src/hooks/useErrorBoundary.test.tsx
@@ -1,0 +1,36 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import useErrorBoundary from './useErrorBoundary';
+
+const handleGlobalError = jest.fn();
+const WrapperComponent = () => {
+  useErrorBoundary(handleGlobalError);
+  return <div />;
+};
+
+describe('useErrorBoundary', () => {
+  it('should add global erro handler upon mounting', () => {
+    render(<WrapperComponent />);
+    expect(window.onerror).toBeDefined();
+  });
+
+  it('should catch uncaught error', () => {
+    const ERROR_MESSAGE = 'my-oh-my';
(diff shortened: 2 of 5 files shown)


hamidzr:
I'll need to do a review but at a glance aren't antd notification used for nonsilent reports as well? it doesn't seem like we want to totally remove them.

hkang1:
> I'll need to do a review but at a glance aren't antd notification used for nonsilent reports as well? it doesn't seem like we want to totally remove them.
> 
oops, yeah for sure, I'll fix that.

> also if https://determined-community.slack.com/archives/C025A98H82X/p1643394532839659?thread_ts=1643394219.828549&cid=C025A98H82X is accurate. i'm surprised why `isAborted` didn't ignore it. maybe it doesn't work after I changed the error to DetError
>
Oh hmmm good point, I'll take a look at that too.

hkang1:
> also if https://determined-community.slack.com/archives/C025A98H82X/p1643394532839659?thread_ts=1643394219.828549&cid=C025A98H82X is accurate. i'm surprised why `isAborted` didn't ignore it. maybe it doesn't work after I changed the error to DetError
>
the `name` property is getting dropped between the translation from `Error` to DetError`. With the aborted errors, `name` gets set to `AbortError`.

hamidzr:
> > also if https://determined-community.slack.com/archives/C025A98H82X/p1643394532839659?thread_ts=1643394219.828549&cid=C025A98H82X is accurate. i'm surprised why `isAborted` didn't ignore it. maybe it doesn't work after I changed the error to DetError
> 
> the `name` property is getting dropped between the translation from `Error` to DetError`. With the aborted errors, `name`gets set to`AbortError`.

we probably should create a ticket for this. I can take it. we'd probably check this in the `processApiError` and set the options accordingly (maybe just bail out?)

webui/react/src/App.tsx
```diff
@@ -40,6 +41,7 @@ const AppView: React.FC = () => {
   useKeyTracker();
   usePageVisibility();
   useRouteTracker();
+  useErrorBoundary();
```
hamidzr, line 44: where you thinking of adding a default handler here or leaving that for another pr? if we do add something I guess we can look at the error and do nothing if (isDetError && isHandled)

hamidzr (commented):
this is looking great. the error boundary would be a nice addition as well.
could you add your test plan? 


I think this had an unwanted side effect. 

before


after





it'd be great if we could have a test for showing that abortErrors are correctly handled and you could kill many birds with one stone (pr) :)
I was thinking something like
1. create an abortError
2. check that it's considered abortError when passed through processApiError and without it.

i was tinkering with it like this but it probably needs a more realistic error
```
diff --git a/webui/react/src/utils/error.test.ts b/webui/react/src/utils/error.test.ts
index 32c52147b..9099531bc 100644
--- a/webui/react/src/utils/error.test.ts
+++ b/webui/react/src/utils/error.test.ts
@@ -1,3 +1,5 @@
+import { isAborted } from 'services/utils';
+
 import * as utils from './error';
 
 describe('Error Handling Utilities', () => {
@@ -19,6 +21,15 @@ describe('Error Handling Utilities', () => {
     });
   });
 
+  describe('isAbortError', () => {
+    it('should detect request aborted errors', () => {
+      // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort TODO
+      const err = { name: 'AbortError' };
+      const error = new utils.DetError(err);
+      expect(isAborted(error)).toStrictEqual(true);
+    });
+  });
+
   describe('isDetError', () => {
     it('should detect DetError objects', () => {
       const error = new utils.DetError('This is a crazy error!');
```

hkang1:
outdated, only concept that might be worth preserving in the future is the react Error Boundary

hamidzr:
I've already pulled it into a branch of mine. (mostly the error boundary) I'll wrap that up probably after the shared-web subtree talk

PR #3525, opened 2022-01-29
feat: put jobs behind or ahead of one another

hamidzr:
## Description

- switch priority lane if needed
- calculate new q position
- add guiding head and tail nodes
- avoid passing SetGroupPriority w/ null priorities
  - note to self: what was their origin?
- reinitialize the q position when moving to another lane
- update how we initialize float64 from time.Time
- delegate sorting jobs to jobs actor
- 
notable missing pieces (future work)

- persistence
- rebalancing calculation
- k8

## Test Plan

- unit tests
- through webui and cli: create jobs and move them around across priorities and resource pools

## Commentary (optional)

there are a good number of TODOs being added in the code but those are good candidates for tasks to be worked on in parallel.

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -603,7 +603,7 @@ func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
 	})
 	err := resp.Error()
 	if err != nil {
-		resources.SetPriority(&oldPriority)
+		resources.SetPriority(oldPriorityPtr)
 		e.Config.SetResources(resources)
 		err = errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -584,24 +584,28 @@ func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
 		return nil
 	}
 	oldPriority := resourcemanagers.DefaultSchedulingPriority
-	var oldPriorityPtr *int
 	resources := e.Config.Resources()
 	if resources.Priority() != nil {
 		oldPriority = *resources.Priority()
-		oldPriorityPtr = &oldPriority
 	}
 	resources.SetPriority(priority)
 	e.Config.SetResources(resources)
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
-		resources.SetPriority(oldPriorityPtr)
+		resources.SetPriority(&oldPriority)
 		e.Config.SetResources(resources)
 		return errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
-	ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
+	resp := ctx.Ask(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
 		Priority: *priority,
 		Handler:  ctx.Self(),
 	})
-	return nil
+	err := resp.Error()
+	if err != nil {
+		resources.SetPriority(&oldPriority)
+		e.Config.SetResources(resources)
+		err = errors.Wrapf(err, "setting experiment %d priority", e.ID)
+	}
+	return err
 }
 
 func (e *experiment) setWeight(ctx *actor.Context, weight float64) error {
@@ -614,11 +618,17 @@ func (e *experiment) setWeight(ctx *actor.Context, weight float64) error {
 		e.Config.SetResources(resources)
 		return errors.Wrapf(err, "setting experiment %d weight", e.ID)
 	}
-	ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupWeight{
+	resp := ctx.Ask(sproto.GetRM(ctx.Self().System()), job.SetGroupWeight{
 		Weight:  weight,
 		Handler: ctx.Self(),
 	})
-	return nil
+	err := resp.Error()
+	if err != nil {
+		resources.SetWeight(oldWeight)
+		e.Config.SetResources(resources)
+		err = errors.Wrapf(err, "setting experiment %d weight", e.ID)
+	}
+	return err
 }
 
 func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
@@ -637,6 +647,8 @@ func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
 		return errors.Wrapf(err, "setting experiment %d RP to %s", e.ID, msg.ResourcePool)
 	}
+	// TODO revert the change like the other setters
+	// also change to ask all?
 	ctx.TellAll(sproto.ChangeRP{ResourcePool: msg.ResourcePool}, ctx.Children()...)
 
 	return nil
@@ -654,7 +666,7 @@ func (e *experiment) toV1Job() *jobv1.Job {
 		Name:           e.Config.Name().String(),
 	}
 
-	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool) // && true
+	j.IsPreemptible = config.ReadRMPreemptionStatus(j.ResourcePool)
 	j.Priority = int32(config.ReadPriority(j.ResourcePool, &e.Config))
 	j.Weight = config.ReadWeight(j.ResourcePool, &e.Config)
 	job.UpdateJobQInfo(&j, e.rmJobInfo)
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -584,14 +584,16 @@ func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
 		return nil
 	}
 	oldPriority := resourcemanagers.DefaultSchedulingPriority
+	var oldPriorityPtr *int
 	resources := e.Config.Resources()
 	if resources.Priority() != nil {
 		oldPriority = *resources.Priority()
+		oldPriorityPtr = &oldPriority
 	}
 	resources.SetPriority(priority)
 	e.Config.SetResources(resources)
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
-		resources.SetPriority(&oldPriority)
+		resources.SetPriority(oldPriorityPtr)
 		e.Config.SetResources(resources)
 		return errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
--- a/master/internal/job/job.go
+++ b/master/internal/job/job.go
@@ -2,9 +2,13 @@ package job
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
+	"github.com/pkg/errors"
+
 	"github.com/determined-ai/determined/master/pkg/actor"
+	"github.com/determined-ai/determined/master/pkg/check"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/proto/pkg/apiv1"
 	"github.com/determined-ai/determined/proto/pkg/jobv1"
@@ -13,6 +17,10 @@ import (
 var (
 	// JobsActorAddr is the address of the jobs actor.
 	JobsActorAddr = actor.Addr("jobs")
+	// HeadAnchor is an internal anchor for the head of the job queue.
+	HeadAnchor = model.JobID("INTERNAL-head")
+	// TailAnchor is an internal anchor for the tail of the job queue.
+	TailAnchor = model.JobID("INTERNAL-tail")
 )
 
 // RMJobInfo packs information available only to the RM that updates frequently.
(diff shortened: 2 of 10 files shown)


master/internal/job/job.go
```diff
@@ -133,6 +141,213 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
+	}
+	// validate anchor and target
+	if anchor == nil {
+		return nil, nil, fmt.Errorf("missing anchor job")
+	}
+	if target == nil {
+		return nil, nil, fmt.Errorf("missing target job")
+	}
+	if target.JobId == anchor.JobId {
+		return nil, nil, nil
```
eecsliu, line 165: maybe raise error? Target job cannot be the same as the anchor? Or target cannot refer to itself?

master/internal/job/job.go
```diff
@@ -133,6 +141,213 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
+			anchorIdx = idx
+			break
+		}
+	}
+	if anchorJob == nil || anchorIdx == -1 {
+		return errJobNotFound(anchorID)
+	}
+
+	if targetJob.Priority != anchorJob.Priority {
+		err = j.setJobPriority(ctx, model.JobID(targetJob.JobId), int(anchorJob.Priority))
```
eecsliu, line 260: Isn't this being done on line 275?

master/internal/job/job.go
```diff
@@ -133,6 +141,213 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
+		priority := int(anchor.Priority)
+		priorityMsg = &SetGroupPriority{
+			Priority: &priority,
+		}
+	}
+
+	var moveMsg *MoveJob
+	// find the next or previous job based on aheadOf in the same priority lane
+	var secondAnchor model.JobID
+	if aheadOf {
```
eecsliu, line 186: Should we do a check for if the job is trying to be inserted in the same position? In which case it would be performing an unnecessary operation.

eecsliu (commented):
Looks mostly good - there are a couple of areas to address and a few questions I had

master/internal/resourcemanagers/job.go
```diff
@@ -126,28 +140,34 @@ func computeNewJobPos(msg job.MoveJob, jobQ job.AQueue, qPositions jobSortState)
...
+	qPos2, ok := qPositions[msg.Anchor2]
+	if !ok {
+		return 0, false, fmt.Errorf("could not find anchor job %s", msg.Anchor2)
+	}
 
-	// qPos2 := rp.queuePositions[nextOrBefore]
+	newPos := (qPos1 + qPos2) / 2
 
-	// find the mid point and update the id's qPosition.
+	return newPos, false, nil
```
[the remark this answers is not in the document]
hamidzr: that's not implemented. but I was assuming we can detect when a rebalancing is needed here. by checking if the delta is below our precision threshold?

master/internal/job/job.go
```diff
@@ -61,8 +69,8 @@ type (
 	// MoveJob requests the job to be moved within a priority queue relative to another job.
 	MoveJob struct {
 		ID      model.JobID
-		Anchor  model.JobID
-		AheadOf bool
+		Anchor1 model.JobID
```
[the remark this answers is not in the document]
hamidzr:
```
aheadOf: true
anchor1: current front most item
anchor2: headAnchor
```

master/internal/job/job.go
hamidzr: that I believe is being checked after this and I think I wrote a unit test for it as well

master/internal/job/job.go
```diff
@@ -133,6 +141,213 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
 }
 
+// figure out what changes are needed to move the job to the desired position.
+// Later we can support cross RP moving this way as well.
+func moveJobMessages(
+	jobs []*jobv1.Job,
+	target *jobv1.Job,
+	anchor *jobv1.Job,
+	anchorIdx int,
+	aheadOf bool,
```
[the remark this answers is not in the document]
hamidzr: that'd be possible. I wanted to keep this closer to the user input for simpler unit testing. otherwise I think it'd work to swap the anchor with head or tail and remove the need for aheadOf

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -377,15 +378,19 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
 
 func (rp *ResourcePool) moveJob(msg job.MoveJob) error {
 	// find out what is the job before or after the anchor
-	queueInfo := rp.scheduler.JobQInfo(rp)
+	// queueInfo := rp.scheduler.JobQInfo(rp)
 
-	newPos, err := computeNewJobPos(msg, queueInfo, rp.queuePositions)
+	fmt.Println(rp.config.PoolName, "moveJob: ", msg)
+	rp.queuePositions[job.TailAnchor] = initalizeQueuePosition(time.Now()) // x
+	newPos, rebalance, err := computeNewJobPos(msg, rp.queuePositions)
```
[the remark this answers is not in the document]
hamidzr: > Is there a way we could check for precision?

I've kinda left this as a future problem since the pr was already getting big. we should be able to come up with the correct answer.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -377,15 +378,19 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
 
 func (rp *ResourcePool) moveJob(msg job.MoveJob) error {
 	// find out what is the job before or after the anchor
-	queueInfo := rp.scheduler.JobQInfo(rp)
+	// queueInfo := rp.scheduler.JobQInfo(rp)
 
-	newPos, err := computeNewJobPos(msg, queueInfo, rp.queuePositions)
+	fmt.Println(rp.config.PoolName, "moveJob: ", msg)
```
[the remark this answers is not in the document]
hamidzr: these would be deleted before we release them. It's been helping with debugging. logs would work too, maybe better

master/internal/experiment.go
```diff
@@ -579,26 +586,30 @@ func checkpointFromTrialIDOrUUID(
...
 	if resources.Priority() != nil {
 		oldPriority = *resources.Priority()
-		oldPriorityPtr = &oldPriority
 	}
-	resources.SetPriority(priority)
+	resources.SetPriority(&priority)
 	e.Config.SetResources(resources)
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
-		resources.SetPriority(oldPriorityPtr)
+		resources.SetPriority(&oldPriority)
```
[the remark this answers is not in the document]
hamidzr: you're right

master/internal/experiment.go
```diff
@@ -579,26 +586,30 @@ func checkpointFromTrialIDOrUUID(
...
 		e.Config.SetResources(resources)
 		return errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
-	ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
-		Priority: priority,
+	resp := ctx.Ask(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
```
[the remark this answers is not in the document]
hamidzr: I changed this. in all cases we want this to be an Ask so we can wait for a setPriority to successfully be applied before moving on.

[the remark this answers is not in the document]
hamidzr: checking for nil here is a good idea. your change brought that in

master/internal/experiment.go
```diff
@@ -611,11 +622,17 @@ func (e *experiment) setWeight(ctx *actor.Context, weight float64) error {
...
 		return errors.Wrapf(err, "setting experiment %d weight", e.ID)
 	}
-	ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupWeight{
+	resp := ctx.Ask(sproto.GetRM(ctx.Self().System()), job.SetGroupWeight{
 		Weight:  weight,
 		Handler: ctx.Self(),
 	})
-	return nil
+	err := resp.Error()
+	if err != nil {
```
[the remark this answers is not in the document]
hamidzr: hmm maybe not now but we still should check and act accordingly. we don't know how the actor would respond with in the future or behave.

master/internal/job/job.go
```diff
@@ -13,6 +17,10 @@ import (
 var (
 	// JobsActorAddr is the address of the jobs actor.
 	JobsActorAddr = actor.Addr("jobs")
+	// HeadAnchor is an internal anchor for the head of the job queue.
```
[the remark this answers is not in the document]
hamidzr: chatted offline about this

master/internal/job/job.go
```diff
@@ -44,13 +52,13 @@ type (
...
-		ResourcePool string
+		ResourcePool string // TODO are we using this?
 		Handler      *actor.Ref
 	}
 	// SetGroupPriority sets the priority of the group in the priority scheduler.
 	SetGroupPriority struct {
-		Priority     *int
-		ResourcePool string
+		Priority     *int   // TODO make this an int?
+		ResourcePool string // TODO are we using this?
```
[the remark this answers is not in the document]
hamidzr: I'd leave changing these for another pr

master/internal/job/job.go
hamidzr: I was gonna say it could be fine without an error but the fact that the user has specified either an ahead of or a behind of makes it so that it has to be an error.

master/internal/job/job.go
```diff
@@ -133,6 +141,212 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
+		return nil, nil, fmt.Errorf("missing anchor job")
+	}
+	if target == nil {
+		return nil, nil, fmt.Errorf("missing target job")
+	}
+	if target.JobId == anchor.JobId {
+		return nil, nil, nil
+	}
+
+	// sanity check
```
hamidzr, line 168: probably remove this later.

master/internal/job/job.go
hamidzr: yep. leftover.

master/internal/resourcemanagers/resource_pool.go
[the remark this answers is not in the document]
hamidzr: actually, we don't have access to ctx.Log or our master logger here unless we add it to the fn signature. since this is temporary wdy think about just keeping it as is?

master/internal/experiment.go
```diff
@@ -637,6 +649,8 @@ func (e *experiment) setRP(ctx *actor.Context, msg job.SetResourcePool) error {
 	if err := e.db.SaveExperimentConfig(e.Experiment); err != nil {
 		return errors.Wrapf(err, "setting experiment %d RP to %s", e.ID, msg.ResourcePool)
 	}
+	// TODO revert the change like the other setters
+	// also change to ask all?
```
hamidzr, line 653: I think @eecsliu has a related pr here

master/internal/job/job.go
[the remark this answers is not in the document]
eecsliu: I'll investigate this after the PR lands and see if a refactor would be beneficial.

master/internal/job/job.go
```diff
@@ -133,6 +141,206 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (AQueue, er
...
+			Priority: int(anchor.Priority),
+		}
+	}
+
+	var moveMsg *MoveJob
+	// find the next or previous job based on aheadOf in the same priority lane
+	var secondAnchor model.JobID
+	if aheadOf {
+		for idx := anchorIdx - 1; idx >= 0; idx-- {
+			if jobs[idx].Priority == anchor.Priority {
```
eecsliu, line 187: can you clarify how this would work? What if there are multiple jobs within the priority? Would we just pick the last one/first one in the range and use that as the secondAnchor? Would this be a problem?

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -422,6 +437,7 @@ func (rp *ResourcePool) receiveJobQueueMsg(ctx *actor.Context) error {
 }
 
 func (rp *ResourcePool) persist() error {
+	// exlude head and tail?
 	fmt.Println("TODO persist rp", rp.config.PoolName)
```
eecsliu, line 441: can these print statements be removed in favor of a todo comment?

eecsliu (approved):
Discussed offline and my concerns are addressed. I do have a couple of comments, but they aren't blocking.

master/internal/job/job.go
hamidzr: the idea is to pick the first one right after, or right behind, in the same priority lane, and if there are none we use head or tail anchors to avoid making those edge cases.

hamidzr: I might be missing cases where this fails. let me know if you see some

master/internal/resourcemanagers/resource_pool.go
hamidzr: I wanted to make it obvious how many times we're getting here as we look through master logs in different cases to see how frequently we're doing this to see if anything would be off. code comment wouldn't expose it in that way but that works too.

master/internal/job/job.go
eecsliu: So the specific question I'm wondering here is why we're searching for the next priority level instead of searching for the job right before or right after the anchor? 

an example of what I'm concerned about is if there are more than 2 jobs within a priority and you're trying to move within that priority. So if you have job A, which is priority 1, and jobs B, C, D, E and F which are priority 2. What happens if you try to promote job F above job D?

PR #3528, opened 2022-01-31
fix: Patch issue with registering a checkpoint to a model version [DET-6622]

mapmeld:
## Description

One-line patch, when I have an experiment checkpoint and register it to a model > model version, it should use the updated models/:name/versions endpoint, instead of sending the ID.

## Checklist


mapmeld:
Here's my test of the Python code, which appears to be working

```python
from determined.experimental import Determined, Model
model_name = "transformer"
checkpoint_uuid = "0fed0062-0da6-4a50-be0e-8f54d8819f25"

model_entry = client.get_model(model_name)
model_entry.register_version(checkpoint_uuid)
model_entry.versions
```

EmilyBonar (approved):
Works on Netlify, LGTM

PR #3530, opened 2022-01-31
fix: correct suggested command to install growforest

stoksc:
## Description
The suggestion to the user to install growforest is without a version, which fails unless you are in a directory with a `go.mod` that specifies it. This change pins the version manually so that the command would succeed in general.

## Test Plan
- [x] fixes it for user that hit the issue

## Checklist


mackrorysd:
+1 (huh - I can't seem to find any option to actually approve from the mobile app.

I could've sworn CloudForest had some versioned releases, but I guess not!

mackrorysd (approved):

stoksc:
Yeah, no releases or tags as far as I see, so I just pinned to what our Makefile does

PR #3531, opened 2022-01-31
feat: return actual resource pool info for K8s [DET-6411, DET-6383]

gt2345:
## Description

For K8s resource pools, `numAgents` return the actual number of agents instead of 1. 
Also add `numSlots` field to sum up slots under each node.

## Test Plan

Navigate to Clusters page with K8s resource pool, verify the `/agents` call and the `/resource-pools` call matches.

### DIFF
--- a/master/internal/resourcemanagers/kubernetes/pods.go
+++ b/master/internal/resourcemanagers/kubernetes/pods.go
@@ -86,8 +86,8 @@ type PodsInfo struct {
 	NumSlots  int
 }
 
-// ResourceSummarize summerize pods resource.
-type ResourceSummarize struct{}
+// SummarizeResources summerize pods resource.
+type SummarizeResources struct{}
 
 // Initialize creates a new global agent actor.
 func Initialize(
@@ -183,7 +183,7 @@ func (p *pods) Receive(ctx *actor.Context) error {
 	case KillTaskPod:
 		p.receiveKillPod(ctx, msg)
 
-	case ResourceSummarize:
+	case SummarizeResources:
 		p.receiveResourceSummarize(ctx, msg)
 
 	case resourceDeletionFailed:
@@ -412,7 +412,7 @@ func (p *pods) receivePodEventUpdate(ctx *actor.Context, msg podEventUpdate) {
 	ctx.Tell(ref, msg)
 }
 
-func (p *pods) receiveResourceSummarize(ctx *actor.Context, msg ResourceSummarize) {
+func (p *pods) receiveResourceSummarize(ctx *actor.Context, msg SummarizeResources) {
 	summary := p.summarize(ctx)
 	slots := 0
 	for _, node := range summary {
--- a/master/internal/resourcemanagers/kubernetes/pods.go
+++ b/master/internal/resourcemanagers/kubernetes/pods.go
@@ -80,11 +80,13 @@ type pods struct {
 	configMapInterface typedV1.ConfigMapInterface
 }
 
+// PodsInfo contains information for pods
 type PodsInfo struct {
 	NumAgents int
 	NumSlots  int
 }
 
+// ResourceSummarize summerize pods resource
 type ResourceSummarize struct{}
 
 // Initialize creates a new global agent actor.
--- a/master/internal/resourcemanagers/kubernetes/pods.go
+++ b/master/internal/resourcemanagers/kubernetes/pods.go
@@ -82,8 +82,8 @@ type pods struct {
 
 // PodsInfo contains information for pods.
 type PodsInfo struct {
-	NumAgents int
-	NumSlots  int
+	NumAgents      int
+	SlotsAvailable int
 }
 
 // SummarizeResources summerize pods resource.
@@ -418,7 +418,7 @@ func (p *pods) receiveResourceSummarize(ctx *actor.Context, msg SummarizeResourc
 	for _, node := range summary {
 		slots += len(node.Slots)
 	}
-	ctx.Respond(&PodsInfo{NumAgents: len(summary), NumSlots: slots})
+	ctx.Respond(&PodsInfo{NumAgents: len(summary), SlotsAvailable: slots})
 }
 
 func (p *pods) receivePodPreemption(ctx *actor.Context, msg PreemptTaskPod) {
--- a/master/internal/resourcemanagers/kubernetes/pods.go
+++ b/master/internal/resourcemanagers/kubernetes/pods.go
@@ -80,13 +80,13 @@ type pods struct {
 	configMapInterface typedV1.ConfigMapInterface
 }
 
-// PodsInfo contains information for pods
+// PodsInfo contains information for pods.
 type PodsInfo struct {
 	NumAgents int
 	NumSlots  int
 }
 
-// ResourceSummarize summerize pods resource
+// ResourceSummarize summerize pods resource.
 type ResourceSummarize struct{}
 
 // Initialize creates a new global agent actor.
--- a/master/internal/resourcemanagers/kubernetes/pods.go
+++ b/master/internal/resourcemanagers/kubernetes/pods.go
@@ -80,6 +80,13 @@ type pods struct {
 	configMapInterface typedV1.ConfigMapInterface
 }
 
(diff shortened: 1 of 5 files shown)


master/internal/resourcemanagers/kubernetes/pods.go
```diff
@@ -80,6 +80,15 @@ type pods struct {
...
 }
 
+// PodsInfo contains information for pods.
+type PodsInfo struct {
+	NumAgents int
+	NumSlots  int
+}
+
+// ResourceSummarize summerize pods resource.
+type ResourceSummarize struct{}
```
stoksc, line 90: nit, to match other messages
```suggestion
type SummarizeResources struct{}
```

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -176,11 +176,14 @@ func (k *kubernetesResourceManager) summarizeDummyResourcePool(
 		slotsUsed += slotsUsedByGroup
 	}
 
+	resp := ctx.Ask(k.agent.handler, kubernetes.ResourceSummarize{}).Get().(*kubernetes.PodsInfo)
```
stoksc, line 179: `ctx.Ask()` has a method `.Error()`. Right now, how we handle can't explicitly error but could we handle it here anyway? If it error were allowed to error, this wouldn't cause a compile time error, just an interface type assertion would fail and the k8s RM would explode.

stoksc: i see code around this not setting a great example. feel free to change any of it you see, too. or just this is fine.

stoksc (commented):
just a minor preventative measure, otherwise the code looks great

proto/src/determined/resourcepool/v1/resourcepool.proto
```diff
@@ -98,6 +98,8 @@ message ResourcePool {
   ResourcePoolType type = 3;
   // The number of agents running in the resource pool
   int32 num_agents = 4;
+  // The number of slots in the resource pool
+  int32 num_slots = 33;
```
ioga, line 102: isn't this the same as `slots_available`? it stands for the total number of slots for agent RM and used this way in the web UI unless I am mistaken.

it seems to be we should be just filling in that field instead of creating a new one.

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -133,7 +136,10 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
 		ctx.Respond(getTaskSummaries(k.reqList, k.groups, kubernetesScheduler))
 
 	case *apiv1.GetResourcePoolsRequest:
-		resourcePoolSummary := k.summarizeDummyResourcePool(ctx)
+		resourcePoolSummary, err := k.summarizeDummyResourcePool(ctx)
+		if err != nil {
+			return err
```
stoksc, line 141: this will crash the resource manager on error - we'd rather not do that. the actor system has lots of pitfalls like this. instead `ctx.Respond(err)`.

proto/src/determined/resourcepool/v1/resourcepool.proto
stoksc: seems right to me, but i glanced the code and im unsure how `k.agent.numSlots()` is right in k8s. i think calling summarize may still be right, but yeah use `slots_available` not add `num_slots`?

ioga: yep. on a related note, I may rip out `k.agent` out of here because it does not bring any value, and make `agentState` private to `internal/resourcemanagers/agent`. are you ok with that or do you think it'd be useful in some way for Slurm?

stoksc: very ok with it, i was not planning on having an equivalent for slurm.

stoksc: thanks for pointing this all out.

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -170,19 +176,28 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
...
 	return &resourcepoolv1.ResourcePool{
 		Name:                         kubernetesDummyResourcePool,
 		Description:                  "Kubernetes-managed pool of resources",
 		Type:                         resourcepoolv1.ResourcePoolType_RESOURCE_POOL_TYPE_K8S,
-		NumAgents:                    1,
+		NumAgents:                    int32(pods.NumAgents),
 		SlotType:                     k.config.SlotType.Proto(),
-		SlotsAvailable:               int32(k.agent.numSlots()),
+		SlotsAvailable:               int32(pods.NumSlots),
 		SlotsUsed:                    int32(k.agent.numUsedSlots()),
```
stoksc, line 201: this and a lot of the other fields that rely on `k.agent` are just borked. maybe we can follow up to this pr and fix them all?

stoksc: i guess @ioga will fix this by ripping `k.agent` out?

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -170,19 +176,28 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
...
 		slotsUsed += slotsUsedByGroup
 	}
 
+	resp := ctx.Ask(k.agent.handler, kubernetes.SummarizeResources{})
+	if err := resp.Error(); err != nil {
+		return nil, err
+	}
+	pods, ok := resp.Get().(*kubernetes.PodsInfo)
+	if !ok {
+		return nil, status.Error(codes.Internal, "unexpected response from actor")
```
stoksc, line 191: `actor.ErrUnexpectedMessage(ctx)` exists

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -170,19 +176,28 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
...
-) *resourcepoolv1.ResourcePool {
+) (*resourcepoolv1.ResourcePool, error) {
 	slotsUsed := 0
 	for _, slotsUsedByGroup := range k.slotsUsedPerGroup {
 		slotsUsed += slotsUsedByGroup
 	}
 
+	resp := ctx.Ask(k.agent.handler, kubernetes.SummarizeResources{})
+	if err := resp.Error(); err != nil {
+		return nil, err
```
stoksc, line 187: could you wrap this?

stoksc (approved):
lgtm, modulo comments about errors. let @ioga check too first, though

master/internal/resourcemanagers/kubernetes/pods.go
```diff
@@ -80,6 +80,15 @@ type pods struct {
 	configMapInterface typedV1.ConfigMapInterface
 }
 
+// PodsInfo contains information for pods.
+type PodsInfo struct {
+	NumAgents int
+	NumSlots  int
```
ioga, line 86: I wonder if it'll look cleaner to call it `SlotsAvailable` to align it with the field name in the api structure.

ioga (approved):

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -217,6 +210,20 @@ func (k *kubernetesResourceManager) summarizeDummyResourcePool(
 	}, nil
 }
 
+func (k *kubernetesResourceManager) summerizePods(
```
ioga, line 213:
```suggestion
func (k *kubernetesResourceManager) summarizePods(
```

master/internal/resourcemanagers/kubernetes_resource_manager.go
ioga: well I was thinking when I rip them out, I'll just replace them with zeros. but yes it's be a good form to have such follow-up.

PR #3532, opened 2022-02-01
chore: modify internal priority message

eecsliu:
## Description
Change priority message to be an int instead of an int pointer.

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -119,7 +119,7 @@ func (c *command) Receive(ctx *actor.Context) error {
 		c.eventStream, _ = ctx.ActorOf("events", newEventManager(c.Config.Description))
 
 		ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
-			Priority: c.Config.Resources.Priority,
+			Priority: *c.Config.Resources.Priority,
 			Handler:  ctx.Self(),
 		})
 
@@ -313,11 +313,9 @@ func (c *command) Receive(ctx *actor.Context) error {
 		}
 
 	case job.SetGroupPriority:
-		if msg.Priority != nil {
-			err := c.setPriority(ctx, *msg.Priority)
-			if err != nil {
-				ctx.Respond(err)
-			}
+		err := c.setPriority(ctx, msg.Priority)
+		if err != nil {
+			ctx.Respond(err)
 		}
 
 	default:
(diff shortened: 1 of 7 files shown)


master/internal/experiment.go
```diff
@@ -594,7 +588,7 @@ func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
 		return errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
 	ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
-		Priority: priority,
+		Priority: *priority,
```
hamidzr, line 591: should we also change this function to also not take a pointer?

master/internal/experiment.go
```diff
@@ -165,12 +165,6 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			MaxSlots: e.Config.Resources().MaxSlots(),
 			Handler:  ctx.Self(),
 		})
-		if err := e.setWeight(ctx, e.Config.Resources().Weight()); err != nil {
```
hamidzr, line 168: are these totally not needed? would we still get the user requested priority and weight if they set it at submission time in the config?

master/internal/experiment.go
eecsliu: This one was hard to do because he priority being modified is in the config resources, which I'm reluctant to change yet. The code actually gets more complex if we don't pass in the pointer.

master/internal/experiment.go
eecsliu: it didn't seem to affect my ability to set priority or weight

master/internal/command/command.go
```diff
@@ -313,11 +313,9 @@ func (c *command) Receive(ctx *actor.Context) error {
 		}
 
 	case job.SetGroupPriority:
-		if msg.Priority != nil {
-			err := c.setPriority(ctx, *msg.Priority)
-			if err != nil {
-				ctx.Respond(err)
-			}
+		err := c.setPriority(ctx, msg.Priority)
```
hamidzr, line 316: nit:  `ctx.Respond(c.setPriority...)`

hamidzr (approved):
LGTM minus the one blocking comment.

hamidzr:
could you please add your test plan to the pr descriptoin

master/internal/command/command.go
```diff
@@ -119,7 +119,7 @@ func (c *command) Receive(ctx *actor.Context) error {
 		c.eventStream, _ = ctx.ActorOf("events", newEventManager(c.Config.Description))
 
 		ctx.Tell(sproto.GetRM(ctx.Self().System()), job.SetGroupPriority{
-			Priority: c.Config.Resources.Priority,
+			Priority: *c.Config.Resources.Priority,
```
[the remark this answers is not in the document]
eecsliu: ah yeah my bad, I missed this. Set priority isn't friendly to using pointers, so I just wrapped it in a check.

hamidzr (approved):

PR #3537, opened 2022-02-01
feat: create Change Password modal

johnkim-det:
## Description

Create a modal that allows the user to change their password.

## Test Plan

Screen Shot 2022-02-01 at 10 03 15 AM

- [ ] User settings modal should display a "Change Password" button.

Screen Shot 2022-02-01 at 10 02 46 AM

- [ ] Clicking this button should open the "Change Password" modal.
- [ ] Form validation should check
  - Old password match
  - New password length (8 characters)
  - New password contains uppercase character
  - New password contains lowercase character
  - "Confirm password" value matches "New password" value
- [ ] Submitting a validated form should update the password

## Checklist


webui/react/src/services/types.ts
```diff
@@ -265,3 +265,8 @@ export interface GetJobsResponse extends Api.V1GetJobsResponse {
 export interface GetJobQStatsParams extends FetchOptions {
   resourcePools?: string[];
 }
+
+export interface SetUserPasswordParams {
+  password: string,
+  username: string
```
EmilyBonar, line 271: nit: should probably end these lines with semicolons instead of commas (though both are valid)

webui/react/src/hooks/useModal/useModalUserSettings.tsx
```diff
@@ -24,6 +28,12 @@ const useModalUserSettings = (): ModalHooks => {
           <span className={css.label}>Username</span>
           <span className={css.value}>{username}</span>
         </div>
+        <div className={css.field}>
+          <span className={css.label}>Password</span>
+          <span className={css.value}>
+            <Button onClick={() => openChangePasswordModal()}>Change password</Button>
```
EmilyBonar, line 34: nit: you should be able to pass the onClick without the callback function, e.g. `onClick={openChangePasswordModal}`

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const [ form ] = Form.useForm();
+
+  const getModalContent = () => {
+    const submitForm = async (): Promise<void> => {
+      await setUserPassword({
+        password: form.getFieldValue('newPassword'),
+        username,
+      });
+      modalClose();
+    };
```
EmilyBonar, line 23: a couple of things:
1. why is `submitForm` defined inside `getModalContent` instead of just being a top level `useCallback` function inside of `useModalChangePassword`?
2. `await setUserPassword` should have a try/catch around it to handle any errors

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+              lowercase.
+            </span>
+          </Form.Item>
+          <Form.Item>
+            <div className={css.buttons}>
+              <Button onClick={() => onCancel()}>Cancel</Button>
+              <Button htmlType="submit" type="primary">
+                Change Password
+              </Button>
+            </div>
```
EmilyBonar, line 96: question: why are you defining your own buttons instead of just using the ones that should come with the modal?
also, nit: that onClick can just be `onClick={onCancel}`

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const modalOpen = () => {
+    openOrUpdate({
+      className: css.noFooter,
+      closable: true,
+      content: getModalContent(),
+      icon: null,
+      onCancel,
+      title: 'Change Password',
+    });
+  };
```
EmilyBonar, line 117: nit: these should both be useCallback functions unless there's a reason they can't be
question: I see that you're removing the footer here in CSS, but why, when you're just re-implementing it yourself?

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+            label="Old Password"
+            name="oldPassword"
+            required
+            rules={[
+              {
+                message: 'Incorrect password',
+                validator: async (rule, value) => {
+                  return await login({
+                    password: value || '',
+                    username: username || '',
```
EmilyBonar, line 38: nit: above you're setting `username = auth.user?.username || 'Anonymous';`, so why do you need the `|| ''` here?

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    : Promise.reject();
+                },
+              },
+            ]}>
+            <Input.Password />
+          </Form.Item>
+          <Form.Item>
+            <span>
+              Password must be at least 8 charactes and contain one uppercase,
+              lowercase.
```
EmilyBonar, line 87: nit: maybe reword this, potentially `Password must be at least 8 characters and contain at least one uppercase and one lowercase character.`

EmilyBonar (changes requested):
A good start, but I think a bit more polish will make it even better. I'd also recommend asking @alexjdetermined about the design of the settings modal, since there's a lot of blank space on the right side that I think could be improved.

webui/react/src/hooks/useModal/useModalUserSettings.tsx
johnkim-det: i think type check requires a callback but changing it to a named function wrapped in useCallback

webui/react/src/hooks/useModal/useModalChangePassword.module.scss
```diff
@@ -0,0 +1,13 @@
...
+  }
+}
+.buttons {
+  display: flex;
+  justify-content: flex-end;
+
+  button + button {
+    margin-left: 8px;
+  }
+}
```
EmilyBonar, line 13: nit: this css isn't used anymore

webui/react/src/hooks/useModal/useModalChangePassword.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    : Promise.reject();
+                },
+              },
+            ]}>
+            <Input.Password />
+          </Form.Item>
+          <Form.Item>
+            <span>
+              Password must be at least 8 characters
+              and contain at least one uppercase and one lowercase character.
```
EmilyBonar, line 97: fix: Need to add that passwords are required to contain a number

EmilyBonar (approved):
LGTM! Just a couple of small changes

PR #3538, opened 2022-02-01
chore: WebUI interaction test on MetricBadgeTag [DET-6424]

mapmeld:
## Description

Based on the test for Avatar, verify that this component on the Experiment Detail page displays the accuracy validation metric as "[ V ] accuracy" and has a "validation" tooltip.

Screen Shot 2022-02-01 at 12 30 51 PM

## Test Plan

`make test` inside of `webui/react`

## Checklist

### DIFF
--- a/webui/react/src/components/MetricBadgeTag.test.tsx
+++ b/webui/react/src/components/MetricBadgeTag.test.tsx
@@ -1,0 +1,52 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+
+import MetricBadgeTag from './MetricBadgeTag';
+
+jest.mock('antd', () => {
+  const antd = jest.requireActual('antd');
+
+  /** mocking Tooltip based on Avatar test */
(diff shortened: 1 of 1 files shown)


johnkim-det (approved):
looks good. one suggestion: maybe instead of the duplicate jest.mock calls, we can put that in a separate file that's imported into MetricBadgeTag.test, Avatar.test, or any test file for a component that uses tooltips? assuming they all would need the same fix.

mapmeld:
I attempted to move the mock function into its own shared file, but I get a mix of errors such as ```The module factory of `jest.mock()` is not allowed to reference any out-of-scope variable```.  There are suggestions [in this thread](https://stackoverflow.com/questions/62695605/how-can-i-share-jest-mocks-between-multiple-tests) to place mocks in a __mocks__ directory, but after a few tries at debugging, I don't want to change too much just to share this function. Agree this should be easier but merging it without changes for now.

PR #3539, opened 2022-02-01
chore: Return trial workloads and metrics in place of ExperimentRaw [DET-6485]

mapmeld:
## Description

The CLI `experiment describe` responds ends with a table of steps with an experiment's checkpoint and validation information. If the user adds the `--metrics` flag we should see more columns with any training and validation metrics.

This helps us remove the old GetExperiment and related ExperimentRaw, ExperimentListRaw etc code

## Test Plan

**Updated 3/15** The majority of changes (by lines) are to e2e_tests.  When testing this PR, focus should be on the `det experiment` CLI and outputs , which should look the same as the current CLI.

- config
- create [should create an experiment and follow the log output]
- describe [including `--metrics`]
- download-model-def
- list-checkpoints [including `--best`, `sortBy=SORT_BY_SEARCHER_METRIC`]
- list-trials

We should also ensure that no other parts of the code call the old API endpoints for: `/:experiment_id` (+ /checkpoints, /config, /model_def, /summary, /metrics/summary).  For example we found that `/experiments/:experiment_id/preview_gc` is still call-able by the API, so it should stay in.

## Checklist


docs/release-notes/3539-replace-core-experiment-with-new-api.txt
```diff
@@ -0,0 +1,12 @@
+:orphan:
+
+**Breaking Changes**
+
+-  Master: removes endpoints (/:experiment_id, /:experiment_id/checkpoints, /:experiment_id/config,
+   /:experiment_id/model_def, /:experiment_id/preview_gc, /:experiment_id/summary, and
+   /:experiment_id/metrics/summary) from the legacy API. This data is available in the GRPC API
```
hamidzr, line 7: nit we could say GRPC and grpc-gateway (or just grpc? 🤔 ) backed rest api. although I'm not sure if we're exposing ports for direct grpc communication or if we wan't to advertise it yet.

master/internal/core_trial.go
```diff
@@ -2,24 +2,12 @@ package internal
...
 }
 
-func (m *Master) getTrialDetails(c echo.Context) (interface{}, error) {
-	args := struct {
-		TrialID int `path:"trial_id"`
-	}{}
-	if err := api.BindArgs(&args, c); err != nil {
-		return nil, err
-	}
-	return m.db.TrialDetailsRaw(args.TrialID)
```
hamidzr, line 20: is this in use anywhere else?

harness/determined/cli/experiment.py
```diff
@@ -260,18 +264,24 @@ def describe(args: Namespace) -> None:
     render.tabulate_or_csv(headers, values, args.csv, outfile)
 
     # Display trial-related information.
+    trials_for_experiment = {}
```
hamidzr, line 267: nit: type info

harness/determined/cli/experiment.py
```diff
@@ -304,66 +320,99 @@ def describe(args: Namespace) -> None:
...
+                    validation_state = step_detail.state.value
+                    validation_end_time = step_detail.endTime
+                    for name in v_metrics_names:
+                        if step_detail.metrics and (name in step_detail.metrics):
+                            v_metrics_fields.append(step_detail.metrics[name])
+                        else:
+                            v_metrics_fields.append(None)
                 else:
-                    validation_state = None
+                    validation_state = ""
```
hamidzr, line 363: do the state enums come with an unspecified option? we could use that.

harness/determined/cli/experiment.py
```diff
@@ -304,66 +320,99 @@ def describe(args: Namespace) -> None:
...
+                        # condense training, checkpoints, validation workloads into one 'step' row
+                        # for compatibility with previous versions of describe
+                        merge_row = step_output[step_detail.totalBatches]
+                        merge_row[3] = max(merge_row[3], render.format_time(step_detail.endTime))
+                        for idx, tfield in enumerate(t_metrics_fields):
+                            if tfield and merge_row[4 + idx] is None:
+                                merge_row[4 + idx] = tfield
+                        start_checkpoint = 4 + len(t_metrics_fields)
+                        if checkpoint_state:
+                            merge_row[start_checkpoint] = checkpoint_state.replace("STATE_", "")
```
hamidzr, line 378: if the state is coming from our generated code as an enum you can do `.value` to get it without the prefix IIRC

harness/determined/cli/experiment.py
```diff
@@ -304,66 +320,99 @@ def describe(args: Namespace) -> None:
...
+                            + t_metrics_fields
+                            + [
+                                checkpoint_state,
+                                render.format_time(checkpoint_end_time),
+                                validation_state,
+                                render.format_time(validation_end_time),
+                            ]
+                            + v_metrics_fields
+                        )
+                        step_output[step_detail.totalBatches] = row
```
hamidzr, line 408: if our tests allow it we might want to go away from displaying the information in a step package all together. It'd be a breaking change but it's a direction we want to go in and could reduce the changes needed here. (If you wanna go that route I'd check with the team first on this)

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -81,14 +87,15 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
...
-                (step["checkpoint"] for step in trials[0]["steps"]),
-                key=operator.itemgetter("total_batches"),
+                (workload["checkpoint"] for workload in checkpoints),
+                key=operator.itemgetter("totalBatches"),
             )
             assert len(checkpoints) == 10
             by_state = {}  # type: Dict[str, Set[int]]
             for checkpoint in checkpoints:
-                by_state.setdefault(checkpoint["state"], set()).add(checkpoint["total_batches"])
+                by_state.setdefault(checkpoint["state"], set()).add(checkpoint["totalBatches"])
```
hamidzr, line 98: question: aren't these coming from our generated code?  why do we access them this way and not by the dot notation?

hamidzr (commented):
I'm happy we're finally removing those endpoints. I haven't done a complete review and I haven't checked it out locally here are some comments I ran into so far.

docs/release-notes/3539-replace-core-experiment-with-new-api.txt
mapmeld: :+1: will change to REST API

master/internal/core_trial.go
mapmeld: nope, this part is good to delete; this whole file should go away in a future PR once we update the other checks in e2e_tests

e2e_tests/tests/experiment/experiment.py
```diff
@@ -214,26 +222,62 @@ def experiment_has_completed_workload(experiment_id: int) -> bool:
...
+def experiment_config_json(experiment_id: int) -> Dict[str, Any]:
+    murl = conf.make_master_url()
+    certs.cli_cert = certs.default_load(murl)
+    authentication.cli_auth = authentication.Authentication(murl, try_reauth=True)
+    sess = session.Session(murl, "determined", authentication.cli_auth, certs.cli_cert)
+    r = bindings.get_GetExperiment(sess, experimentId=experiment_id)
+    return r.config
 
 
 def experiment_json(experiment_id: int) -> Dict[str, Any]:
```
stoksc, line 243: i think we should replace these (this func and all similar onces around) with something like `experiment() -> v1Experiment` or something. to me this is the point of having bindings

stoksc (commented):
this is great, i've wanted to see these changes for a long time. i think if you're going to dig in and change all these, we should properly use the bindings, so that api changes have a chance at breaking cli lint in specific places.

e2e_tests/tests/experiment/test_core.py
```diff
@@ -65,11 +66,11 @@ def test_metric_gathering() -> None:
         batches_trained = step["total_batches"]
 
     # Check validation metrics.
-    for step in trials[0]["steps"]:
-        validation = step["validation"]
-        metrics = validation["metrics"]
-        actual = metrics["validation_metrics"]
-        batches_trained = step["total_batches"]
+    validation_workloads = list(filter(lambda w: w.validation, trials[0].workloads))
```
stoksc, line 69: now that i've read more, these functions seem pretty generally useful. can we put them somewhere else?

e2e_tests/tests/experiment/test_core.py
```diff
@@ -65,11 +66,11 @@ def test_metric_gathering() -> None:
...
+    for step in validation_workloads:
```
stoksc, line 70: step is an antiquated word for a determined training workload
```suggestion
    for val in validation_workloads:
```

e2e_tests/tests/experiment/test_core.py
```diff
@@ -151,19 +152,20 @@ def test_experiment_archive_unarchive() -> None:
     )
 
     # Check that we can archive and unarchive the experiment and see the expected effects.
+    sleep(0.25)
```
stoksc, line 155: what's going on here?

e2e_tests/tests/experiment/test_noop.py
```diff
@@ -124,19 +124,18 @@ def test_noop_single_warm_start() -> None:
     assert len(trials) == 1
 
     second_trial = trials[0]
-    assert len(second_trial["steps"]) == 30
+    assert len(second_trial.workloads) == 90
 
     # Second trial should have a warm start checkpoint id.
-    assert second_trial["warm_start_checkpoint_id"] == last_checkpoint_id
+    # assert second_trial["warm_start_checkpoint_id"] == last_checkpoint_id
```
stoksc, line 130: what's up here?

e2e_tests/tests/experiment/test_noop.py
```diff
@@ -208,7 +205,7 @@ def test_cancel_one_paused_experiment() -> None:
         conf.fixtures_path("no_op"),
         ["--paused"],
     )
-
+    time.sleep(0.25)
```
stoksc, line 208: ?

stoksc (commented):
there are a bunch of places with time.sleeps and stuff commented out that is new. what's up with all that?

e2e_tests/tests/experiment/test_noop.py
mapmeld: ok it looks like I can safely remove the added `time.sleep` calls now that I handle a race condition in `wait_for_experiment_state` https://github.com/determined-ai/determined/pull/3539/files#diff-66c234e5c973d0546d1f6713c5b496deca22124eea269b1f8c5a3fb0c1193c81R116

e2e_tests/tests/experiment/experiment.py
```diff
@@ -214,74 +199,89 @@ def experiment_has_completed_workload(experiment_id: int) -> bool:
...
     r = api.get(conf.make_master_url(), "trials/{}/metrics".format(trial_id))
-    assert r.status_code == requests.codes.ok, r.text
     json = r.json()  # type: Dict[str, Any]
     return json
 
 
 def get_flat_metrics(trial_id: int, metric: str) -> List:
     full_trial_metrics = trial_metrics(trial_id)
-    metrics = [m for step in full_trial_metrics["steps"] for m in step["metrics"]["batch_metrics"]]
+    metrics = [
```
stoksc, line 282: this doesn't looks right to me, since `trial_metrics` should still be returning steps. but it looks like `get_flat_metrics` is unused so it doesn't matter?

e2e_tests/tests/experiment/experiment.py
```diff
@@ -440,15 +475,15 @@ def run_list_cli_tests(experiment_id: int) -> None:
...
-    active = sum(1 for t in trials if t["state"] == "ACTIVE")
-    paused = sum(1 for t in trials if t["state"] == "PAUSED")
-    stopping_completed = sum(1 for t in trials if t["state"] == "STOPPING_COMPLETED")
-    stopping_canceled = sum(1 for t in trials if t["state"] == "STOPPING_CANCELED")
-    stopping_error = sum(1 for t in trials if t["state"] == "STOPPING_ERROR")
-    completed = sum(1 for t in trials if t["state"] == "COMPLETED")
-    canceled = sum(1 for t in trials if t["state"] == "CANCELED")
-    errored = sum(1 for t in trials if t["state"] == "ERROR")
-    stopping_killed = sum(1 for t in trials if t["state"] == "STOPPING_KILLED")
+    active = sum(1 for t in trials if t.trial.state.value == "STATE_ACTIVE")
```
stoksc, line 478: are there constants for these enums?

e2e_tests/tests/experiment/experiment.py
```diff
@@ -615,15 +644,16 @@ def __str__(self) -> str:
 
 
 def get_experiment_durations(experiment_id: int, trial_idx: int) -> ExperimentDurations:
-    experiment_metadata = experiment_json(experiment_id)
+    r = bindings.get_GetExperiment(test_session(), experimentId=experiment_id)
```
stoksc, line 647: why not use the bindings for this one too?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -299,21 +299,22 @@ def _filter_named_parameters(self, optimizer: torch.optim.Optimizer) -> List:
         opt_params = {p for group in optimizer.param_groups for p in group.get("params", [])}
         return [(name, p) for name, p in self._main_model.named_parameters() if p in opt_params]
 
-    def _init_device(self) -> None:
+    def _init_device(self) -> torch.device:
```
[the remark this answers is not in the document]
mapmeld: This is the lint error that I mentioned last week (it was worried self.device does not have a set type). I have no idea why lint raises it in this PR and not the master branch

harness/determined/cli/experiment.py
```diff
@@ -620,7 +676,7 @@ def set_gc_policy(args: Namespace) -> None:
                 render.format_resources(c["resources"]),
             ]
             for c in sorted(checkpoints, key=lambda c: (c["trial_id"], c["end_time"]))
-            if "step" in c and c["step"].get("validation") is not None
+            if "step" in c and c["step"].get("validation")
```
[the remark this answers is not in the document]
mapmeld: I was removing a bunch of `is not None` code elsewhere and this got caught up in it. Mostly I'm not touching this code since it is still using the old trial / step JSON

proto/src/determined/trial/v1/trial.proto
```diff
@@ -25,6 +25,8 @@ message CheckpointWorkload {
   map<string, int64> resources = 5;
   // Total number of batches as of this workload's completion.
   int32 total_batches = 8;
+  // Id of the checkpoint in the database.
+  int32 id = 9;
```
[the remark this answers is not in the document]
mapmeld: It turns out that warmStartCheckpointId is the db ID and not the UUID 😬 
so when we compare is this first checkpoint based on the warmStartCheckpointId as expected, we want it available

harness/determined/pytorch/_pytorch_context.py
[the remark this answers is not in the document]
stoksc: fwiw, i undid all the changes on this file and lint passed on your branch for me?

proto/src/determined/trial/v1/trial.proto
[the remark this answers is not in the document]
stoksc: yah that's unfortunate but can't we just make this a uuid and checkpoint warm start checkpoint id on the trials api to a uuid, since we just added it, instead of releasing an api just to break eventually? then `proto_get_trials_plus.sql` can just shim the id to a uuid pretty easily.

harness/determined/pytorch/_pytorch_context.py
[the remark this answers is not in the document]
stoksc: but if half undid the changes, i got the error you showed me that day.

e2e_tests/tests/experiment/experiment.py
mapmeld: This is never called. I am going to remove this method and the ExperimentDurations class

proto/src/determined/trial/v1/trial.proto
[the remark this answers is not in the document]
mapmeld: ok I am going to remove this field, and replace `warmStartCheckpointId` with `warmStartCheckpointUuid` since this name can distinguish it from previous use in the API

harness/determined/pytorch/_pytorch_context.py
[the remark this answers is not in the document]
mapmeld: ok, I will remove changes in a future commit and see how it goes on lint. Might have been related to what tests were and weren't accessible?

e2e_tests/tests/experiment/experiment.py
```diff
@@ -214,74 +199,89 @@ def experiment_has_completed_workload(experiment_id: int) -> bool:
...
-        trial = trials[0]
-        assert trial["state"] == "CANCELED", trial["state"]
+        trial = trials[0].trial
+        assert trial.state == bindings.determinedexperimentv1State.STATE_CANCELED
 
 
 def is_terminal_state(state: str) -> bool:
-    return state in ("CANCELED", "COMPLETED", "ERROR")
+    return state in (
+        bindings.determinedexperimentv1State.STATE_CANCELED.value,
```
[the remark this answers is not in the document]
mapmeld: ok I am going to use the bindings here in is_terminal_state , wait_for_experiment_state (the only place which calls this), and the tests which use wait_for_experiment_state

e2e_tests/tests/experiment/experiment.py
mapmeld: I'll use the constants here too

e2e_tests/tests/experiment/experiment.py
mapmeld: this is no longer used so I will remove it

harness/determined/pytorch/_pytorch_context.py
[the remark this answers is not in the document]
mapmeld: restores this file back to no changes, and I get the lint issue again?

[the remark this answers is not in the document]
rb-determined-ai: I've seen this locally for a while, but now that it's in circleci we should fix it.

I think this fix looks good pretty good to me.

If you make this little fix into a separate PR I'll stamp it.

stoksc:
i have a comment i forgot to leave, `det e` and `det e --metrics` are very complicated now. I would be in favor of just rewriting them rather than trying to repair them into the right direction. but as long as you tested it a lot we can land it in this state for now.

gh-determined-ai (changes requested):
In the release notes, instead of the construct "Master: blah blah blah" provide a more narrative description, such as:
For the master (if that's even needed), deprecated the following endpoints:
- /endpointA
- /endpointB
- /endpointC
and the functionality is now available (something like that) in /endpointX.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -113,16 +144,18 @@ def run_gc_checkpoints_test(checkpoint_storage: Dict[str, str]) -> None:
                 storage_manager = storage.build(checkpoint_config, container_path=None)
                 storage_state = {}  # type: Dict[str, Any]
                 for checkpoint in checkpoints:
-                    storage_id = checkpoint["uuid"]
+                    if not checkpoint.uuid:
```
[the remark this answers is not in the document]
mapmeld: for some reason it believes the type could be None (`error: Invalid index type "Optional[str]" for "Dict[str, Any]"; expected type "str"`), so I want to filter out this possibility.  I think this could be due to no typing on `all_checkpoints = []`

harness/determined/cli/experiment.py
```diff
@@ -232,13 +232,17 @@ def delete_experiment(args: Namespace) -> None:
...
 @authentication.required
 def describe(args: Namespace) -> None:
+    session = setup_session(args)
     docs = []
     for experiment_id in args.experiment_ids.split(","):
         if args.metrics:
-            r = api.get(args.master, "experiments/{}/metrics/summary".format(experiment_id))
+            r = bindings.get_GetExperiment(session, experimentId=experiment_id)
+        else:
+            r = bindings.get_GetExperiment(session, experimentId=experiment_id)
```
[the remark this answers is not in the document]
mapmeld: ✅

harness/determined/cli/experiment.py
```diff
@@ -301,16 +311,22 @@ def describe(args: Namespace) -> None:
...
     # Display step-related information.
+    t_metrics_headers: List[str] = []
+    t_metrics_names: List[str] = []
+    v_metrics_headers: List[str] = []
+    v_metrics_names: List[str] = []
     if args.metrics:
         # Accumulate the scalar training and validation metric names from all provided experiments.
-        t_metrics_names = sorted({n for doc in docs for n in scalar_training_metrics_names(doc)})
+        for doc in docs:
+            # if len(trials_for_experiment[doc.id]) == 0:
```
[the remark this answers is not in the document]
mapmeld: ✅ 
on the previous point, I'll keep checkpoint_state as our object up until the end where we output the row to CLI

harness/determined/cli/experiment.py
```diff
@@ -324,66 +340,99 @@ def describe(args: Namespace) -> None:
...
+                            t_metrics_fields.append(wl_detail.metrics[name])
                         else:
                             t_metrics_fields.append(None)
+                else:
+                    t_metrics_fields = [None for name in t_metrics_names]
+
+                if workload.checkpoint:
+                    ckpt = workload.checkpoint.to_json()
+                    ckpt["numInputs"] = 0
+                    wl_detail = bindings.v1MetricsWorkload.from_json(ckpt)
```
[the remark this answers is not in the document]
mapmeld: What's going on here is `v1MetricsWorkload` and `v1CheckpointWorkload` are very similar and normally would have a shared superclass. It looks like I can do `wl_detail: bindings.v1MetricsWorkload|bindings.v1CheckpointWorkload = None` earlier.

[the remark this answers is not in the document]
stoksc: yeah, i get what's going on, just seems subject to breakages really easily, as opposed to just making some shared repr class. but it's fine for now, i suppose.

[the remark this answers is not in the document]
stoksc: idk, checkpoints and metrics won't look that similar for much longer, maybe not

e2e_tests/tests/cluster/test_checkpoints.py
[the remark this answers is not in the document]
mapmeld: I'm going to add typing earlier, and replace this if/continue with an `assert checkpoint.uuid`

harness/determined/cli/checkpoint.py
```diff
@@ -62,16 +63,24 @@ def render_checkpoint(checkpoint: experimental.Checkpoint, path: Optional[str] =
...
-    if args.best is not None:
-        if args.best < 0:
-            raise AssertionError("--best must be a non-negative integer")
-        params["best"] = args.best
-
-    r = api.get(
-        args.master, "experiments/{}/checkpoints".format(args.experiment_id), params=params
-    ).json()
-    searcher_metric = r["metric_name"]
+    r = bindings.get_GetExperimentCheckpoints(
```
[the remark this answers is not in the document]
mapmeld: updated to:
```python
    r = bindings.get_GetExperimentCheckpoints(
        setup_session(args),
        id=args.experiment_id,
        limit=args.best,
        sortBy=bindings.v1GetExperimentCheckpointsRequestSortBy.SORT_BY_SEARCHER_METRIC,
    )
```

harness/determined/cli/experiment.py
```diff
@@ -232,13 +232,17 @@ def delete_experiment(args: Namespace) -> None:
 
 @authentication.required
 def describe(args: Namespace) -> None:
+    session = setup_session(args)
     docs = []
```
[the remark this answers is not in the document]
mapmeld: this is now changed ✅

harness/determined/cli/experiment.py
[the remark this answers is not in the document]
mapmeld: this is now `bindings.v1MetricsWorkload | bindings.v1CheckpointWorkload | None` type -- hopefully the future changes don't affect the few attributes which we read from them within this test

mapmeld:
@gh-determined-ai I've updated the release note

gh-determined-ai (approved):

harness/determined/cli/experiment.py
mapmeld: In the bindings, the state object's value returns a string but it still has the STATE_ prefix.

I've tried a few ways of doing this where we store the `determinedexperimentv1State` instead of a string, then handle `.value` and `STATE_` in the output, but it runs into two main problems:
- Some rows have no checkpoint or validation state. We want to keep these blank and not use `STATE_UNSPECIFIED`
- These columns are passed to `render.tabulate_or_csv`, which is expecting strings and will render a state object as `determinedexperimentv1State.STATE_COMPLETED`

docs/release-notes/3539-replace-core-experiment-with-new-api.txt
```diff
@@ -0,0 +1,23 @@
+:orphan:
```
stoksc, line 1: these look much better than our prior release notes. thanks @gh-determined-ai for the guidance.

harness/determined/cli/checkpoint.py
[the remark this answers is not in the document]
stoksc: should we still check `args.best` and do by searcher metric if best, else by end time?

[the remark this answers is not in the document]
mapmeld: sounds good, just added in a commit

stoksc:
@mapmeld `det e create const.yaml . -f` is still broken because the `-f` part calls `/experiments` to look for the experiment to have trials.

in general, what is the test plan for this, to make sure bugs like this don't slip through?

mapmeld:
@stoksc that's a fair question. I've just run `git grep experiments` and `git grep trials` to search for where else we use these endpoints. Our search area would be in `harness` or a not-running part of `e2e_tests`

- the case you've mentioned, `follow_experiment_logs`, now fixed. We had a similar issue with follow_test_experiment_logs
- common/api/experiment uses patch_experiment and there are a few places (tests/cli/test_cli) where we post to create experiments, both routes were kept in for this reason
- in e2e_tests num_experiments - this function was no longer called; removed
- in e2e_tests and harness calling `trials/{}` and `trials/{}/metrics` - these routes are still supported

From the search, seeing these appear in the same files as our other changes (i.e. not bringing in other cli commands), and the `-f` option being something I should have noticed uses different code between test and prod, I feel more confident about having caught these

e2e_tests/tests/experiment/experiment.py
```diff
@@ -214,91 +199,100 @@ def experiment_has_completed_workload(experiment_id: int) -> bool:
...
+                and s.validation.state == determinedexperimentv1State.STATE_COMPLETED
+            ):
+                return True
+    return False
 
 
-def experiment_state(experiment_id: int) -> str:
-    state = experiment_json(experiment_id)["state"]  # type: str
-    return state
+def test_session() -> session.Session:
```
stoksc, line 215: still need a better way long term to do this.

e2e_tests/tests/experiment/experiment.py
```diff
@@ -214,91 +199,100 @@ def experiment_has_completed_workload(experiment_id: int) -> bool:
...
+    return r.config
 
 
-def num_experiments() -> int:
-    certs.cli_cert = certs.default_load(conf.make_master_url())
-    authentication.cli_auth = authentication.Authentication(conf.make_master_url(), try_reauth=True)
-    r = api.get(conf.make_master_url(), "experiments")
-    assert r.status_code == requests.codes.ok, r.text
-    return len(r.json())
+def experiment_trials_json(experiment_id: int) -> Sequence[bindings.trialv1Trial]:
```
stoksc, line 227: nit, name no longer makes sense

harness/determined/common/api/experiment.py
```diff
@@ -24,23 +19,20 @@ def patch_experiment(master_url: str, exp_id: int, patch_doc: Dict[str, Any]) ->
...
-    headers = {"Content-Type": "application/merge-patch+json"}
-    req.post(master_url, path, headers=headers)
-
-
 def follow_experiment_logs(master_url: str, exp_id: int) -> None:
     # Get the ID of this experiment's first trial (i.e., the one with the lowest ID).
     print("Waiting for first trial to begin...")
+    certs.cli_cert = certs.default_load(master_url)
+    authentication.cli_auth = authentication.Authentication(master_url, try_reauth=True)
+    sess = session.Session(master_url, "determined", authentication.cli_auth, certs.cli_cert)
```
stoksc, line 27: @rb-determined-ai is there a ticket or anything for unifying how we're using/passing these sessions

stoksc (approved):
lgtm, i've tried to review everything as carefully as i can but with the size of course stuff may slip, so please write a large test plan for release party and run through it yourself before merging if you can? (basically cover every CLI endpoint you could've possibly broken, i just tried to do the same and it looks good)

gh-determined-ai (approved):

gh-determined-ai (approved):

PR #3387, opened 2022-01-06
chore: change the output format args for master config cli

hamidzr:
## Description
match the output format style of other det cli commands

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3540, opened 2022-02-01
build: switch old -o json usage to --json

hamidzr:
followup to https://github.com/determined-ai/determined/pull/3387

## Test Plan

`ag det -C 10 | grep config -C 3 | ag -C 2 -- '(-o|--output|--json)' | less`
- run post merge tests

## Checklist

### DIFF
--- a/e2e_tests/tests/conftest.py
+++ b/e2e_tests/tests/conftest.py
@@ -143,8 +143,7 @@ def using_k8s(request: SubRequest) -> bool:
         config.make_master_url(),
         "master",
         "config",
-        "-o",
-        "json",
+        "--json",
     ]
 
     output = subprocess.check_output(command, universal_newlines=True, stderr=subprocess.PIPE)


hamidzr:
@ioga do you know of any spots I might have missed? I'll request all the tests to run

ioga (approved):
no, I am not aware of any

hamidzr:
partly related: potentially something we want to do later https://github.com/determined-ai/determined/pull/3480 to move the tests to not go through cli

ioga:
why?

we could test it "directly" through the API, but we don't have any users for the API (while we do have them for CLI).
so, from a certain perspective, using CLI in tests instead of API allows us to test it alongside the backend.

hamidzr:
all tests passed except the deploy test. it seems like that's due to the [s3 bucket name size limit ](https://determined-ai.slack.com/archives/C9LFPNA3Y/p1643756294331409?thread_ts=1643744755.627349&cid=C9LFPNA3Y)

hamidzr:
I haven't given it too much thoughts but testing through the cli seems better than no cli test all but separate cli test seems easier to diagnose, run, etc

hamidzr:
> why?
> 
> we could test it "directly" through the API, but we don't have any users for the API (while we do have them for CLI). so, from a certain perspective, using CLI in tests instead of API allows us to test it alongside the backend.

I haven't given it too much thoughts but testing through the cli seems better than no cli test all but separate cli test seems easier to diagnose, run, etc

PR #3541, opened 2022-02-01
fix: capture sys.exit code from launch

azhou-determined:
## Description

fixes a bug where container exits with a failure code but experiment is marked as completed because launch layer is not capturing the sys.exit of the launch call.

- actual code change is in launch.py
- moved launch layer specific e2e tests into its own file
- added return code for run_failure_experiment to capture trial logs

## Checklist

### DIFF
--- a/e2e_tests/tests/experiment/experiment.py
+++ b/e2e_tests/tests/experiment/experiment.py
@@ -668,8 +668,7 @@ def run_failure_test_with_temp_config(
     with tempfile.NamedTemporaryFile() as tf:
         with open(tf.name, "w") as f:
             yaml.dump(config, f)
-        experiment_id = run_failure_test(tf.name, model_def_path, error_str=error_str)
-    return experiment_id
+        return run_failure_test(tf.name, model_def_path, error_str=error_str)
 
 
 def shared_fs_checkpoint_config() -> Dict[str, str]:
--- a/e2e_tests/tests/experiment/experiment.py
+++ b/e2e_tests/tests/experiment/experiment.py
@@ -552,9 +552,7 @@ def verify_completed_experiment_metadata(
 
 
 # Use Determined to run an experiment that we expect to fail.
-def run_failure_test(
-    config_file: str, model_def_file: str, error_str: Optional[str] = None
-) -> None:
+def run_failure_test(config_file: str, model_def_file: str, error_str: Optional[str] = None) -> int:
     experiment_id = create_experiment(config_file, model_def_file)
 
     wait_for_experiment_state(experiment_id, "ERROR")
@@ -582,6 +580,8 @@ def run_failure_test(
         logs = trial_logs(trial_id)
         if error_str is not None:
             assert any(error_str in line for line in logs)
+
+    return experiment_id
 
 
 def get_validation_metric_from_last_step(
@@ -664,11 +664,12 @@ def run_failure_test_with_temp_config(
     config: Dict[Any, Any],
     model_def_path: str,
     error_str: Optional[str] = None,
-) -> None:
+) -> int:
     with tempfile.NamedTemporaryFile() as tf:
         with open(tf.name, "w") as f:
             yaml.dump(config, f)
-        run_failure_test(tf.name, model_def_path, error_str=error_str)
+        experiment_id = run_failure_test(tf.name, model_def_path, error_str=error_str)
+    return experiment_id
 
 
 def shared_fs_checkpoint_config() -> Dict[str, str]:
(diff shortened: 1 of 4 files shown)


e2e_tests/tests/experiment/experiment.py
```diff
@@ -664,11 +664,12 @@ def run_failure_test_with_temp_config(
     config: Dict[Any, Any],
     model_def_path: str,
     error_str: Optional[str] = None,
-) -> None:
+) -> int:
     with tempfile.NamedTemporaryFile() as tf:
         with open(tf.name, "w") as f:
             yaml.dump(config, f)
-        run_failure_test(tf.name, model_def_path, error_str=error_str)
+        experiment_id = run_failure_test(tf.name, model_def_path, error_str=error_str)
```
rb-determined-ai, line 671: you can also just `return run_failure_test(...)`.  The context managers are still respected before returning control to the caller

e2e_tests/tests/experiment/test_launch.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    collect_trial_profiles(trials[0]["id"])
+
+    assert exp.check_if_string_present_in_trial_logs(
+        trials[0]["id"], "container exited successfully with a zero exit code"
+    )
+
+
+@pytest.mark.parallel
```
rb-determined-ai, line 38: why make it a parallel test?  What codepaths are covered by making it parallel vs making it non-parallel?

e2e_tests/tests/experiment/test_launch.py
```diff
@@ -0,0 +1,58 @@
...
+    config = conf.set_profiling_enabled(config)
+    config = conf.set_entrypoint(
+        config, "python3 -m nonexistent_launch_module model_def:CIFARTrial"
+    )
+
+    experiment_id = exp.run_failure_test_with_temp_config(
+        config, conf.cv_examples_path("cifar10_pytorch")
+    )
+    trials = exp.experiment_trials(experiment_id)
+    (Determined(conf.make_master_url()).get_trial(trials[0]["id"]))
```
rb-determined-ai, line 52: what are these extra parens for?

e2e_tests/tests/experiment/test_launch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    collect_trial_profiles(trials[0]["id"])
```
rb-determined-ai, line 54: what is this testing?

e2e_tests/tests/experiment/test_launch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    assert exp.check_if_string_present_in_trial_logs(
+        trials[0]["id"], "container failed with non-zero exit code: 1"
+    )
```
rb-determined-ai, line 58: is this going to annoyingly say "assertion failed: False" if it fails?

e2e_tests/tests/experiment/test_launch.py
azhou-determined: this is a pytest hook that profiles the test for our internal Determined CI server.

e2e_tests/tests/experiment/test_launch.py
azhou-determined: nope, because i added a print statement inside of check_if_string_present_in_trial_logs a while back that dumps the trial logs and the string that was not found.

e2e_tests/tests/experiment/test_launch.py
azhou-determined: removed

e2e_tests/tests/experiment/test_launch.py
rb-determined-ai: ah.  Someday we should maybe make it `exp.assert_string_present_in_trial_logs()`.  But no need to do that today.

e2e_tests/tests/experiment/test_launch.py
azhou-determined: parallel isn't necessary. i changed it to just e2e_cpu

e2e_tests/tests/experiment/test_launch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    assert exp.check_if_string_present_in_trial_logs(
+        trials[0]["id"], "container exited successfully with a zero exit code"
+    )
+
+
+@pytest.mark.e2e_cpu
+def test_launch_layer_exit(collect_trial_profiles: Callable[[int], None]) -> None:
+    config = conf.load_config(conf.cv_examples_path("cifar10_pytorch/const.yaml"))
+    config = conf.set_max_length(config, {"batches": 200})
```
rb-determined-ai, line 41: are all of these `conf.set*` lines necessary, given that we are not even launching a real entrypoint?

azhou-determined: removed

rb-determined-ai (approved):

PR #3542, opened 2022-02-01
fix: correct tqdm log display and fix the flickering issue [DET-6470]

hkang1:
## Description

Two problems to address.
1. The tqdm logs uses a `\r` carriage return for progress bars because its an ANSI method of updating the previously printed line (and overwriting with a new string). Unfortunately we don't support the complex ANSI cursor movement behavior, especially via our streaming backend. So instead we treat them as new lines.
2. The flickering of the logs are caused by the log message being so large and tall that it exceeds the visible window range. The logic is broken for such large messages, causing the large log messages to disappear. Update the logic to address this.

NOTE: [this PR](https://github.com/determined-ai/determined/pull/3327) updates the task logs to use the new log viewer, will not have some of these visibility issues. Task logs is the last piece that will be on this old log viewer.

BEFORE:
https://user-images.githubusercontent.com/220971/152058402-ad09117f-a3a4-4a59-b0df-d12c8df5efc4.mov

AFTER:
https://user-images.githubusercontent.com/220971/152058381-e981a823-635b-4223-977b-a86883c92a7c.mov

## Test Plan

- [ ] run from the CLI `det cmd run "python -c 'from tqdm import tqdm ; from time import sleep ; [sleep(1) for _ in tqdm(range(200))] ; print("text")'"` and wait for the tqdm logs to get generated. Ensure that the rendering looks good.
- [ ] Also check that the scrolling of the tqdm logs don't cause any flickering.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer.tsx
+++ b/webui/react/src/components/LogViewer.tsx
@@ -269,7 +269,9 @@ const LogViewer: React.FC<Props> = forwardRef((
       if (!size) return false;
       const top = size.top;
       const bottom = size.top + size.height;
-      return (top > viewTop && top < viewBottom) || (bottom > viewTop && bottom < viewBottom);
+      return (top > viewTop && top < viewBottom)
+        || (bottom > viewTop && bottom < viewBottom)
+        || (top < viewTop && bottom > viewBottom);
     });
   }, [ config, logs, scroll ]);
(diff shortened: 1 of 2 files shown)


gt2345 (approved):
LGTM

PR #3544, opened 2022-02-02
fix: deflake unets-tf-keras test by avoiding redundant downloads

mpkouznetsov:
## Description
https://determinedai.atlassian.net/browse/DET-6598
The test is flaky because occasionally the dataset download times out. We run 8 workers, and sometimes one or two of them time out in the download; the end result is a COMPLETED experiment with a single trial that has 0 steps.

*Note*: A preprocessed version of the dataset has been uploaded to S3 and is available at https://s3-us-west-2.amazonaws.com/determined-ai-datasets/oxford_iiit_pet/oxford_iiit_pet.tar.gz

## Test Plan

Run the e2e test against a cluster.

## Commentary (optional)

There are 2 things we consider doing:
1. Download dataset to the same nominal location `/tmp/data` instead of distinct locations by rank.
2. For CI purposes only:
  - download and preprocess the dataset into S3
  - amend the test code (the example itself as stored on github is unchanged for this purpose) to change `startup-hook.sh` on the fly to download the dataset from S3 into `/tmp/data` so that `tfds.load()` becomes a noop.

We are certain that we want to do (1). (2) may be controversial. I am doing those two things separately in this PR so that we can make separate decisions on them.

## Checklist

### DIFF
--- a/examples/computer_vision/unets_tf_keras/model_def.py
+++ b/examples/computer_vision/unets_tf_keras/model_def.py
@@ -5,6 +5,7 @@ regarding the optional flags view the original script linked below.
 This implementation is based on:
 https://github.com/tensorflow/docs/blob/master/site/en/tutorials/images/segmentation.ipynb
 """
+import filelock
 import tensorflow as tf
 from tensorflow_examples.models.pix2pix import pix2pix
 import tensorflow_datasets as tfds
@@ -23,7 +24,7 @@ class UNetsTrial(TFKerasTrial):
 
         # Create a unique download directory for each rank so they don't overwrite each
         # other when doing distributed training.
-        self.download_directory = f"/tmp/data-rank{self.context.distributed.get_rank()}"
+        self.download_directory = "/tmp/data"
         self.data_downloaded = False
 
     def normalize(self, input_image, input_mask):
@@ -62,7 +63,8 @@ class UNetsTrial(TFKerasTrial):
         mobilenet_link = 'https://storage.googleapis.com/tensorflow/keras-applications/mobilenet_v2/' + data_file
         os.mkdir(weights_dir)
 
-        urllib.request.urlretrieve(mobilenet_link,weights_dir + data_file)
+        with filelock.FileLock(os.path.join(weights_dir, "download.lock")):
+            urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
         return weights_dir + data_file
 
     def build_model(self):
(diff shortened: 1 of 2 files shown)


examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -23,7 +24,7 @@ def __init__(self, context):
 
         # Create a unique download directory for each rank so they don't overwrite each
         # other when doing distributed training.
```
rb-determined-ai, line 26: there's a comment here that needs updating

examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -60,9 +61,10 @@ def download_weights(self):
         weights_dir = self.download_directory + '/weights/'
         data_file = self.context.get_data_config()['data_file']
         mobilenet_link = 'https://storage.googleapis.com/tensorflow/keras-applications/mobilenet_v2/' + data_file
-        os.mkdir(weights_dir)
+        os.makedirs(weights_dir, exist_ok=True)
 
-        urllib.request.urlretrieve(mobilenet_link,weights_dir + data_file)
+        with filelock.FileLock(os.path.join(weights_dir, "download.lock")):
```
rb-determined-ai, line 66: I think in the example I linked you, there was a brief comment above the filelock usage:

```python
    # Use a file lock so only one worker on each node does the download.
```

Can you put that here too?  I think one of the AMLEs requested that we keep an explanation in a comment of what the filelock was doing.  The assumption is that, while filelocks may be the right tool for the job, they're not common tools in the everyday ML developer's toolbox.

examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -109,13 +111,15 @@ def build_model(self):
...
     def build_training_data_loader(self):
-        dataset = tfds.load(
-            'oxford_iiit_pet:3.*.*',
-            split="train",
-            with_info=False,
-            data_dir=self.download_directory,
-            download=not self.data_downloaded,
-        )
+        os.makedirs(self.download_directory, exist_ok=True)
+        with filelock.FileLock(os.path.join(self.download_directory, "download.lock")):
```
rb-determined-ai, line 115: ditto on the comment

examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -109,13 +111,15 @@ def build_model(self):
...
+            dataset = tfds.load(
+                'oxford_iiit_pet:3.*.*',
+                split="train",
+                with_info=False,
+                data_dir=self.download_directory,
+                download=not self.data_downloaded,
```
rb-determined-ai, line 121: what's this `download` arg for?  I think you can mark it as true or maybe omit it now.  Hopefully `self.data_downloaded` can be done away with too

rb-determined-ai (approved):
I really like the pattern we arrived at here.  I like that we can reasonably expect our test to be rock-solid, while letting the example be highly vanilla.

I think we should remember this pattern when other similar flakes show up.

mpkouznetsov:
> I really like the pattern we arrived at here. I like that we can reasonably expect our test to be rock-solid, while letting the example be highly vanilla.
> 
> I think we should remember this pattern when other similar flakes show up.

Yes. And in addition to being sensible it is more efficient. The 3 clusters of durations in the picture below, 5'16", 6'30", and 7'30"+ correspond to
- the final implementation with 1 download per worker from S3, 
- the implementation with 1 download per worker from the official dataset hosting site
- the previous implementation to rank-specific location from the official hosting site

image

PR #3547, opened 2022-02-02
chore: stop streaming when page loses focus and restart them when they are [DET-6629]

hkang1:
## Description

PROBLEM
gRPC streaming are HTTP2 based connections so a typical browser limits them to about 6 per host. Which means when the user is opening multiple tabs, it can quickly saturate the browser's ability to make new connections. New connections are made but end up in a `pending` state until the previous ones resolves, but with streaming endpoints they generally do not resolve unless the backend closes the connection or if it errors out.

SOLUTION
We previously [applied the logic](https://github.com/determined-ai/determined/pull/3500) of canceling any ongoing polling endpoint calls when the user switches to another window or tab, and reactivating them when they come back to the page/window/tab. Apply the same logic for streaming endpoints.

## Test Plan

- [ ] open a page with streaming endpoints (such as a trial detail page for an active trial with the profiler enabled). Check to see that the streaming endpoints are open and have not reached terminal state (under Console > Network), and then click on a different tab. Come back to the tab you were on and check to see that the previous streaming endpoints have terminated and new ones are started.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
@@ -6,6 +6,7 @@ import Link from 'components/Link';
 import Message, { MessageType } from 'components/Message';
 import Spinner from 'components/Spinner';
 import { terminalRunStates } from 'constants/states';
+import { useStore } from 'contexts/Store';
 import useStorage from 'hooks/useStorage';
 import { paths } from 'routes/utils';
 import {
@@ -73,6 +74,7 @@ const ExperimentVisualization: React.FC<Props> = ({
   experiment,
   type,
 }: Props) => {
+  const { ui } = useStore();
   const history = useHistory();
   const location = useLocation();
   const storage = useStorage(`${STORAGE_PATH}/${experiment.id}`);
@@ -154,7 +156,7 @@ const ExperimentVisualization: React.FC<Props> = ({
 
   // Stream available metrics.
   useEffect(() => {
-    if (!isSupported) return;
+    if (!isSupported || ui.isPageHidden) return;
 
     const canceler = new AbortController();
     const trainingMetricsMap: Record<string, boolean> = {};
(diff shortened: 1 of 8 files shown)


EmilyBonar (approved):
Seems to work well, although the `Waiting for profiler data...` step takes a fairly long time. It would be good if we could somehow speed that up in the future.

hkang1:
> Seems to work well, although the `Waiting for profiler data...` step takes a fairly long time. It would be good if we could somehow speed that up in the future.

Do you know if the delay was caused by resources ramping up or trial getting scheduled? I think the delay of sending profiler data could have been caused by profiler date not being available yet. We have some planned worked around better visibility for experiments and trials so the user can see that the delay of data is due to them not being available. Everything currently just looks like a giant waiting view. Not easy to tell what is going on unless they look at the logs...

EmilyBonar:
> Do you know if the delay was caused by resources ramping up or trial getting scheduled? I think the delay of sending profiler data could have been caused by profiler date not being available yet. We have some planned worked around better visibility for experiments and trials so the user can see that the delay of data is due to them not being available. Everything currently just looks like a giant waiting view. Not easy to tell what is going on unless they look at the logs...

This was after some profiler data was already available and showing, then refreshing, so it shouldn't be due to data not being available. It wasn't terrible, just like 10-20 seconds.

hkang1:
> > Do you know if the delay was caused by resources ramping up or trial getting scheduled? I think the delay of sending profiler data could have been caused by profiler date not being available yet. We have some planned worked around better visibility for experiments and trials so the user can see that the delay of data is due to them not being available. Everything currently just looks like a giant waiting view. Not easy to tell what is going on unless they look at the logs...
> 
> This was after some profiler data was already available and showing, then refreshing, so it shouldn't be due to data not being available. It wasn't terrible, just like 10-20 seconds.

Ah ok. I'll create a ticket for this just so we can keep tabs on this.

PR #3550, opened 2022-02-02
chore: add tensborboard timeout logging [DET-6317] 

varlogtim:
Test by launching a Tensorboard with these options:

`det tensorboard start -t <trial_id> --config=debug=true` and note that we get INFO level logging about the IdleTimeoutTicker state.

### DIFF
--- a/master/internal/task/idle_timeout.go
+++ b/master/internal/task/idle_timeout.go
@@ -3,6 +3,8 @@ package task
 import (
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
 	"github.com/determined-ai/determined/master/internal/sproto"
 
 	"github.com/determined-ai/determined/master/internal/proxy"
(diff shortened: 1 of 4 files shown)


master/internal/task/idle_timeout.go
```diff
@@ -75,6 +77,12 @@ func (p *IdleTimeoutWatcher) ReceiveMsg(ctx *actor.Context) error {
 			}
 		}
 
+		if p.Debug {
+			ctx.Log().Infof("IdleTimeoutWatcherTick: lastActivity=%s timeout=%s",
+				lastActivity.Format(time.RFC3339), lastActivity.Add(p.Timeout).Format(time.RFC3339),
```
stoksc, line 82: not critical, just more FYI it exists
```suggestion
			ctx.Log().WithFields(
				"lastActivity", lastActivity.Format(time.RFC3339),
				"timeout", lastActivity.Add(p.Timeout).Format(time.RFC3339),
			).Infof("idle timeout watcher ticked"),
```

stoksc (approved):
lgtm for now. i think the key is the config is good long term, but we maybe could be more elegant with how we enable debug logs.

master/internal/task/idle_timeout.go
stoksc: sorry i didnt actually make sure the suggestion was right (has extra paren)

PR #3552, opened 2022-02-02
refactor: agent RM state consolidation. [DET-6638]

ioga:
## Description

Refactor agent RM:
- Move agentState from resource pool into agent itself. Have RP request it on scheduling. Remove a number of duplicated fields such as enabled, disabled, label.
- Move slot states into agent itself. Have slot(s) actors only serve the legacy APIs. The structures itself are the same, but in the future it'd be good to simplify and combine `agentState.slots` and `agentState.devices` together.
- don't do double `clearNonReattachedContainers` for agents reconnecting after crash.
- get rid of `k.agent` in kubernetes RM and use `podsActor` directly.
- `agent.agentStartedInit` is no longer necessary as it's effectively the same as `agent.started`.

## Test Plan

Keep an eye on stability, race conditions, errors in logs, etc.

## Checklist


master/internal/resourcemanagers/agent/agent.go
```diff
@@ -250,34 +257,70 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
 		if a.awaitingReconnect {
 			return errors.New("agent failed to reconnect by deadline")
 		}
+	case GetAgentState:
+		if !a.started {
+			ctx.Respond(errors.New("agent state is not available: agent not started"))
+			return nil
+		}
+
+		ctx.Respond(a.agentState)
```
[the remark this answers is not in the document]
stoksc: e.g. i could see a situation where the scheduler decides to schedule on an agent and calls AllocateFreeDevices but the slot is concurrently disabled and we hit the "panic not enough slots" case.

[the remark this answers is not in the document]
stoksc: i see some deep copying in the scheduler but it doesn't look like enough, and i think it should happen here

master/internal/resourcemanagers/agent/slot.go
```diff
@@ -10,134 +10,82 @@ import (
...
-
-func (s *slot) summarize(ctx *actor.Context) model.SlotSummary {
-	return model.SlotSummary{
-		ID:        ctx.Self().Address().Local(),
-		Device:    s.device,
-		Enabled:   s.enabled.Enabled(),
-		Container: s.container,
-		Draining:  s.enabled.draining,
+	default:
+		ctx.Respond(echo.ErrMethodNotAllowed)
```
stoksc, line 89: check `ctx.IsExpectingResponse` else I think this can panic.

master/internal/resourcemanagers/kubernetes_resource_manager.go
```diff
@@ -90,15 +88,9 @@ func (k *kubernetesResourceManager) Receive(ctx *actor.Context) error {
...
-		k.agent = &agentState{
-			handler:            podsActor,
-			devices:            make(map[device.Device]*cproto.ID),
-			zeroSlotContainers: make(map[cproto.ID]bool),
-			// Expose a fake number here just to signal to the UI
-			// that this RP does support the aux containers.
-			maxZeroSlotContainers: 1,
-			enabled:               true,
-		}
+		// Expose a fake number of zero slots here just to signal to the UI
```
stoksc, line 91: comment is nonsense now

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -398,11 +373,28 @@ func (rp *ResourcePool) receiveRequestMsg(ctx *actor.Context) error {
...
 }
 
+func (rp *ResourcePool) fetchAgentStates(ctx *actor.Context) map[*actor.Ref]*agent.AgentState {
+	agents := make([]*actor.Ref, 0, len(rp.agents))
+
+	for k := range rp.agents {
+		agents = append(agents, k)
+	}
+
+	responses := ctx.AskAll(agent.GetAgentState{}, agents...).GetAll()
```
stoksc, line 383: can we check the responses for errors so an single agent crash doesn't crash the RP when it does the type assertion below?

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -398,11 +373,28 @@ func (rp *ResourcePool) receiveRequestMsg(ctx *actor.Context) error {
 	return nil
 }
 
+func (rp *ResourcePool) fetchAgentStates(ctx *actor.Context) map[*actor.Ref]*agent.AgentState {
+	agents := make([]*actor.Ref, 0, len(rp.agents))
```
stoksc, line 377: this is a bit unfortunate

master/pkg/device/device.go
```diff
@@ -36,9 +36,12 @@ func (t Type) Proto() devicev1.Type {
 	}
 }
 
+// ID the type of Device.ID.
+type ID = int
```
stoksc, line 40:
```suggestion
type ID int
```

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -264,17 +261,27 @@ func (rp *ResourcePool) Receive(ctx *actor.Context) error {
...
-			AgentReconnectWait:   rp.config.AgentReconnectWait,
-			AgentReattachEnabled: rp.config.AgentReattachEnabled,
+			AgentReconnectWait:    rp.config.AgentReconnectWait,
+			AgentReattachEnabled:  rp.config.AgentReattachEnabled,
+			MaxZeroSlotContainers: rp.config.MaxAuxContainersPerAgent,
 		})
 
 	case schedulerTick:
 		if rp.reschedule {
+			rp.agentStatesCache = rp.fetchAgentStates(ctx)
```
stoksc, line 280: thoughts on letting agents push updates instead of this?

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -250,34 +257,70 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
...
+	case PatchSlotState:
+		if !a.started {
+			ctx.Respond(errors.New("can't patch slot state: agent not started"))
+			return nil
+		}
+
+		result, err := a.agentState.patchSlotState(ctx, msg)
+		if err != nil {
+			ctx.Respond(err)
+			ctx.Log().WithError(err)
```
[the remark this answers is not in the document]
ioga: fair - I think I may have been carrying that over from elsewhere

master/internal/resourcemanagers/kubernetes_resource_manager.go
ioga: moved elsewhere. 👍

master/internal/resourcemanagers/resource_pool.go
ioga: it's the lack of `.keys()` method

master/internal/resourcemanagers/resource_pool.go
ioga: that's a good Q. it'd mean we'd have to sprinkle something like `agent.notifyResourcePool` all around the code to tell it what's up. which feels like it'd be easy to miss. 
I can't think of a way to automate it in go without having explicit getters/setters for all fields on the state.

master/pkg/device/device.go
ioga: that has added a few unfortunate casts here and there but overall I think it looks nicer

master/internal/resourcemanagers/resource_pool.go
stoksc: yeah, good point. maybe we can make scheduler tick configurable as an escape hatch..

master/pkg/device/device.go
stoksc: i think it's strictly better usually, e.g., https://go.dev/play/p/ftp_m7N-wzm

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -123,12 +125,30 @@ func (rp *ResourcePool) allocateResources(ctx *actor.Context, req *sproto.Alloca
...
-			container: container,
-			devices:   fit.Agent.allocateFreeDevices(fit.Slots, container.id),
-		})
+		resp := ctx.Ask(fit.Agent.Handler, agent.AllocateFreeDevices{
+			Slots:       fit.Slots,
+			ContainerID: container.id,
+		}).Get()
+		switch resp := resp.(type) {
+		case agent.AllocateFreeDevicesResponse:
+			devices := resp.Devices
```
stoksc, line 134: what is this trying to do?

master/internal/resourcemanagers/agent/agent_state.go
```diff
@@ -0,0 +1,375 @@
...
+
+	return slotState.summarize()
+}
+
+func (a *AgentState) patchAllSlotsState(
+	ctx *actor.Context, msg PatchAllSlotsState) model.SlotsSummary {
+	result := model.SlotsSummary{}
+	for _, slotState := range a.slotStates {
+		summary := a.patchSlotStateInner(
+			ctx, PatchSlotState{ID: -1, Enabled: msg.Enabled, Drain: msg.Drain}, slotState)
```
stoksc, line 361: i get the -1 isn't used by it's still really confusing to read. why not just pass `slotState.device.id`

master/internal/resourcemanagers/agent/agent_state.go
```diff
@@ -0,0 +1,375 @@
...
+func (a *AgentState) containerStateChanged(ctx *actor.Context, msg aproto.ContainerStateChanged) {
+	for _, d := range msg.Container.Devices {
+		s, ok := a.slotStates[d.ID]
+		if !ok {
+			ctx.Log().Warnf("bad containerStateChanged on device: %d (%s)", d.ID, a.string())
+			continue
+		}
+
+		s.container = &msg.Container
+		if msg.Container.State == cproto.Terminated {
```
stoksc, line 266: should we also, finally, `delete(a.containers, msg.Container.ID)`. i dont see that map being deleted from

master/internal/resourcemanagers/agent/agent_state.go
```diff
@@ -0,0 +1,375 @@
...
+		Device:    s.device,
+		Enabled:   s.enabled.Enabled(),
+		Container: s.container,
+		Draining:  s.enabled.draining,
+	}
+}
+
+// AgentState holds the scheduler state for an agent. The implementation of agent-related operations
+// (e.g., socket I/O) is deferred to the actor.
+type AgentState struct {
```
stoksc, line 47: seems like we may as well write some simple unit tests over this.

stoksc (approved):
lgtm, somethings we can discuss but nothing blocking. sorry for the nits, just know it means i read the code really carefully, heh.

master/internal/resourcemanagers/resource_pool.go
stoksc: yeah

master/internal/resourcemanagers/agent/agent_state.go
ioga: that happens in `agent.containerStateChanged` (which calls this method).

master/internal/resourcemanagers/priority.go
```diff
@@ -279,38 +280,40 @@ func sortTasksByPriorityAndTimestamp(
...
+		copiedAgents[key] = agent.DeepCopy()
 	}
 	return copiedAgents
 }
 
 func addTaskToAgents(fits []*fittingState) {
 	for _, fit := range fits {
-		fit.Agent.allocateFreeDevices(fit.Slots, cproto.NewID())
+		if _, err := fit.Agent.AllocateFreeDevices(fit.Slots, cproto.NewID()); err != nil {
+			panic(err)
```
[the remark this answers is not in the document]
ioga: `check.Panic` just calls panic; `check.True/False` here don't make sense; so I'll just `panic(errors.Wrap(...`.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -123,12 +125,30 @@ func (rp *ResourcePool) allocateResources(ctx *actor.Context, req *sproto.Alloca
...
+			allocations = append(allocations, &containerReservation{
+				req:       req,
+				agent:     fit.Agent,
+				container: container,
+				devices:   devices,
+			})
+		case error:
+			// Rollback previous allocations.
```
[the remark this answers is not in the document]
ioga: we are going through the agents trying to allocate things; if we fail we've got to clear out already "allocated" fits, so we do rollback. then it'll be retried on the next scheduler cycle, when we'll refetch all agent states and see if we can schedule it again.

I also now realize this was missing updating our cached agent state after allocating successfully.

PR #3556, opened 2022-02-03
chore: remove container_id from ClusterInfo

rb-determined-ai:
Even though agents and k8s let us see the container ids before we
actually start the container, for other runtimes we plan to support
(like slurm) are likely not to support such a workflow.  A simpler, more
universal workflow is to ask some external resource manager, "please
give us N containers with this environment" and we get N containers with
identical environments.

This means that we should not be relying on container_id anywhere in the
harness (except perhaps Rendezvous API which can just use a different
implementation at some point when we have to support non-agent, non-k8s
rendezvous).  But the ClusterInfo API, which will become a public API
soon, should certainly not expose a container_id.

## Commentary (optional)

I noticed that there are a few API routes that require a container_id, which need to be modified.  I'm ok landing this with the hacks in place, but I'd like to make the `# HACK` comments into `# TODO (DET-NNNN)` comments (I mean, I'd like to discuss what it would take to change them, and to make a ticket for that work)

### DIFF
--- a/harness/determined/_env_context.py
+++ b/harness/determined/_env_context.py
@@ -11,7 +11,6 @@ class EnvContext:
         master_url: str,
         master_cert_file: Optional[str],
         master_cert_name: Optional[str],
-        container_id: str,
         experiment_config: Dict[str, Any],
         hparams: Dict[str, Any],
         latest_checkpoint: Optional[str],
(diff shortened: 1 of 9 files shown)


stoksc (approved):
thanks

rb-determined-ai:
@bradley what about the `#HACK` comments I mentioned?  Those are going to require us to stop taking `container_id`s as parameters in the REST API.  Is that going to be feasible?

rb-determined-ai:
after talking in slack, it seems like we need to know if daemon mode would be supported in slurm.

If it is, then we could have a new rendezvous-like api that we call to populate container_id in tasks where it isn't normally easy to populate, and in that case we would probably just close this PR.

rb-determined-ai:
After more talking, this time in zoom, we decided that `container_id` is more likely to be `container_rank` in the future.

In any case, removing it from the ClusterInfo API while ClusterInfo is still a non-public API seems like a good idea.

PR #3559, opened 2022-02-03
fix: Handle model name changes [DET-6648]

mapmeld:
## Description

When a model successfully changes name, the browser should be redirected to the new URL (/det/models/:new_name)

When a model fails to change name, we were already resetting InlineEditor to the previous name, but this should handle it with a returned error instead of my previous `throw Error()` idea

## Test Plan

1. Change the name of a model "test" -> "test2"
2. Change the name of a model to something not allowed "hello" -> "hello/world"

## Checklist

### DIFF
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -62,14 +62,12 @@ const InlineEditor: React.FC<Props> = ({
 
   const save = useCallback(async (newValue: string) => {
     if (onSave) {
-      try {
-        setIsSaving(true);
-        await onSave(newValue);
-      } catch (e) {
+      setIsSaving(true);
+      const err = await onSave(newValue);
+      if (err !== null) {
         updateEditorValue(value);
-      } finally {
-        setIsSaving(false);
       }
+      setIsSaving(false);
     }
   }, [ onSave, updateEditorValue, value ]);
 
--- a/webui/react/src/pages/ModelDetails.tsx
+++ b/webui/react/src/pages/ModelDetails.tsx
@@ -19,6 +19,7 @@ import TagList from 'components/TagList';
 import { useStore } from 'contexts/Store';
 import usePolling from 'hooks/usePolling';
 import useSettings from 'hooks/useSettings';
+import { paths } from 'routes/utils';
 import { archiveModel, deleteModel, deleteModelVersion, getModelDetails, patchModel,
   patchModelVersion, unarchiveModel } from 'services/api';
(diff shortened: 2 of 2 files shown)


webui/react/src/pages/ModelDetails.tsx
```diff
@@ -292,14 +293,16 @@ const ModelDetails: React.FC = () => {
         body: { name: editedName },
         modelName,
       });
+      window.location.href = `/det${paths.modelDetails(editedName)}`;
```
EmilyBonar, line 296: This line doesn't work (at least on Netlify). You should be able to do `routeToReactUrl(paths.modelDetails(editedName))` instead.

webui/react/src/pages/ModelDetails.tsx
```diff
@@ -292,14 +293,16 @@ const ModelDetails: React.FC = () => {
...
     } catch (e) {
       handleError(e, {
         publicSubject: 'Unable to save name.',
         silent: true,
         type: ErrorType.Api,
       });
-      throw Error('Revert name change');
+      return e;
     }
+    return null;
```
EmilyBonar, line 305: Is this `return null` necessary?

webui/react/src/components/InlineEditor.tsx
```diff
@@ -62,14 +62,12 @@ const InlineEditor: React.FC<Props> = ({
 
   const save = useCallback(async (newValue: string) => {
     if (onSave) {
-      try {
-        setIsSaving(true);
-        await onSave(newValue);
-      } catch (e) {
+      setIsSaving(true);
+      const err = await onSave(newValue);
+      if (err !== null) {
```
EmilyBonar, line 67: You should be able to just use `if (err)`

EmilyBonar (changes requested):
Looks good, but the redirect needs to work

webui/react/src/components/InlineEditor.tsx
mapmeld: I get this TypeScript error `An expression of type 'void' cannot be tested for truthiness.  TS1345`
I'm making the other changes but `onSave` has to have a consistent Promise<object> type or keep this if err !== null

EmilyBonar (approved):
LGTM! The only comment I have is that it fails silently, should there be a message telling users that updating the name didn't work?

mapmeld:
I agree that there could be a message (above or below the input? using the planned universal notifications button?).
I'd also like to use catch save errors like this in other places where we use InlineEditor.
I'm going to make that into a new ticket.

PR #3560, opened 2022-02-03
chore: preserve try-for-flake script for react unit tests from e2e

hkang1:
## Description

Preserve the script that re-runs tests multiple times to attempt to trigger flaky test errors.

## Test Plan

- [ ] try running `./bin/try-for-flakes.sh true "npm run test -- --watchAll=false src/components/Timeago.test.tsx"` and see that the `Timeago` unit tests run over and over again.

## Checklist

### DIFF
--- a/webui/react/bin/try-for-flakes.sh
+++ b/webui/react/bin/try-for-flakes.sh
@@ -1,24 +1,14 @@
-# eg ./bin/try-for-flakes.sh true "npm run test -- --watchAll=false src/components/Timeago.test.tsx"
+# eg ./bin/try-for-flakes.sh "npm run test -- --watchAll=false src/components/Timeago.test.tsx"
 
-# if tests are re-runnable
-re_runnable=${1:-false}
-test_cmd=${2:-"npm run test -- --watchAll=false"}
+test_cmd=${1:-"npm run test -- --watchAll=false"}
 
 c=0
 
-if $re_runnable; then
-  while true; do
-    c=$((c+1))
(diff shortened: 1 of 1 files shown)


hkang1:
I separated out this PR because I think we can take advantage of this script now (instead of waiting for the removal of web e2e-tests PR to land).

webui/react/bin/try-for-flakes.sh
```diff
@@ -0,0 +1,24 @@
+# eg ./bin/try-for-flakes.sh true "npm run test -- --watchAll=false src/components/Timeago.test.tsx"
+
+# if tests are re-runnable
+re_runnable=${1:-false}
```
hamidzr, line 4: we probably don't need to differentiate between rerunnable and not anymore but this works as is. We'll know what we need from this next time we're looking for a flake

hamidzr (approved):

webui/react/bin/try-for-flakes.sh
hkang1: something like this look ok?
```
# eg ./bin/try-for-flakes.sh "npm run test -- --watchAll=false src/components/Timeago.test.tsx"

test_cmd=${1:-"npm run test -- --watchAll=false"}

c=0

while true; do
  c=$((c+1))
  echo "run #$c"
  ${test_cmd}
done

echo "result: test failure at run #$c of $(git rev-parse --short HEAD)"
# TODO use trap to show successful result upon SIGTERM SIGINT
```

PR #3561, opened 2022-02-03
feat: customize experiment columns [DET-6398] [DET-6401]

EmilyBonar:
## Description
Adds a column picker modal to the Experiment List page. Also turns Archived filtering into a switch at the top of the page.

Screenshot 2022-02-07 at 15-22-21 Experiments - Determined
Screenshot 2022-02-07 at 15-22-50 Experiments - Determined

## Test Plan
I have done manual testing on all features, including performance testing (column picker is performant at ~50,000 elements). There are also a good number of interaction tests.

## Checklist


docs/release-notes/3561-customize-experiment-columns.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**New Features**
+
+-  Web: Adds a column picker to the Experiment List page to allow users to choose which table
+   columns to display.
```
hkang1, line 6: **suggestion:** add some screens and/or screencast to demo this awesome new feature!

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  showModal: (props: ShowModalProps) => void;
+}
+
+const useCustomizeColumnsModal = (): ModalHooks => {
+  const modalRef = useRef<ReturnType<ModalFunc>>();
+  const [ modalState, setModalState ] = useState<ModalState>({
+    columns: [],
+    defaultVisibleColumns: [],
+    initialVisibleColumns: [],
+    visible: false,
```
hkang1, line 36: **suggestion:** for `columns` and `defaultVisibleColumns`, they might be better as arguments passed into `useCustomizeColumnModal` if they are not change between each `modalOpen` call.

```
const useCustomizeColumnsModal = ({ columns: string[], defaultVisibleColumns: string[] }): ModalHooks => {
```

We should be able to drop `visible` if we use `useModal` as mentioned in the comment above.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  }, [ closeModal ]);
+
+  const handleSave = useCallback((state: ModalState) => {
+    if (!modalRef.current) return;
+    state.onSave?.(visibleColumns);
+    closeModal();
+  }, [ closeModal, visibleColumns ]);
+
+  const handleSearch = useCallback((e) => {
+    setSearchTerm(e.target.value.toLowerCase());
```
hkang1, line 66: **nit:** looking at how the search term is used, it seems it's used via regex pretty consistently with ignore case, so the `toLowerCase()` is probably unnecessary. I was initially going to suggest this to avoid mutating the search term of what the user types (ie. as the user type, changing their input on the fly to lower case) but that doesn't seem to be the case (the case is preserved) so it's not a big deal either way.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  }, [ generateModalProps, modalState ]);
+
+  // When the component using the hook unmounts, remove the modal automatically.
+  useEffect(() => {
+    return () => {
+      if (!modalRef.current) return;
+      modalRef.current.destroy();
+      modalRef.current = undefined;
+    };
+  }, []);
```
hkang1, line 229: **suggestion:** I think this will be good to have inside `useModal`. I'm not sure what kind of impact it has inside there but might be worth moving there and trying it out. I think in general should be a good behavior for all modals. Also within `useModal` could just call `modalClose` instead of lines `225 ~ 227`.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  defaultVisibleColumns: string[];
+  initialVisibleColumns: string[];
+  onSave?: (columns: string[]) => void;
+}
+
+interface ModalHooks {
+  showModal: (props: ShowModalProps) => void;
+}
+
+const useCustomizeColumnsModal = (): ModalHooks => {
```
hkang1, line 30: **nit:** I know it's a bit wonky but would `useModalCustomizeColumns` be a deal breaker? The idea is to keep all the modal hooks listed near each other in the directory.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+export interface ShowModalProps {
+  columns: string[];
+  defaultVisibleColumns: string[];
+  initialVisibleColumns: string[];
+  onSave?: (columns: string[]) => void;
+}
+
+interface ModalHooks {
+  showModal: (props: ShowModalProps) => void;
+}
```
hkang1, line 28: **suggestion:** inherit the `useModal ModalHooks` definition. Something like:
```
interface ModalHooks extends Omit<Hooks, 'modalOpen'> {
  modalOpen: (props: ShowModalProps) => void;
}
```

goal of providing a consistent modal hook interface, even though the props between the functions might vary. So ensuring that the `modalClose` and `modalRef` are returned in case they are needed.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+  // Detect modal state change and update.
+  useEffect(() => {
+    if (!modalState.visible) return;
+    const modalProps = generateModalProps(modalState);
+    if (modalRef.current) {
+      modalRef.current.update(prev => ({ ...prev, ...modalProps }));
+    } else {
+      modalRef.current = Modal.confirm(modalProps);
+    }
+  }, [ generateModalProps, modalState ]);
```
hkang1, line 220: **suggestion:** this is a continuation of the `useModal` suggestion above, but this is another good example of what you don't need to reinvent.

webui/react/src/hooks/useCustomizeColumnsModal.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  const filteredVisibleColumns = useMemo(() => {
+    const regex = RegExp(searchTerm, 'i');
+    return visibleColumns.filter(column => regex.test(camelCaseToSentence(column)));
+  }, [ visibleColumns, searchTerm ]);
+
+  const makeHidden = useCallback((transfer: string | string[]) => {
+    if (Array.isArray(transfer)) {
+      const transferSet = new Set(transfer);
+      setVisibleColumns(prev => prev.filter(column => !transferSet.has(column)));
```
hkang1, line 87: Nice! pretty nifty optimization, how did you come up with it?

webui/react/src/utils/string.test.ts
```diff
@@ -9,6 +9,14 @@ describe('String Utilities', () => {
...
   });
 
+  describe('sentenceToCamelCase', () => {
+    it('should convert camel case to a sentence', () => {
+      expect(utils.sentenceToCamelCase('Hello')).toBe('hello');
+      expect(utils.sentenceToCamelCase('Camel Case')).toBe('camelCase');
+      expect(utils.sentenceToCamelCase(' Car Jump Start ')).toBe('carJumpStart');
+    });
+  });
+
```
hkang1, line 19: muchas gracias 🙏

hkang1 (commented):
This PR looks really really good. As far as functionality it's super performant and works smoothly. Not to mention awesome tests. The hook test is particularly awesome!

The one primary suggestion is to compose your hook with the lower level `useModal` hook to avoid having to rewrite some of the props and functions you wrote that duplicate the work there. This will also better ensure the behavior of the core modal will behave consistently across all `useModal` compositions.

UI worked nicely. One small recommendation:
2022-02-04 at 1 21 PM

webui/react/src/hooks/useCustomizeColumnsModal.tsx
EmilyBonar: Searching stack overflow, haha! Plus remembering that Set.has() is an O(1) operation.

webui/react/src/hooks/useModal/useModalCustomizeColumns.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+    await view.findByText('Customize Columns');
+
+    const lists = view.getAllByRole('list');
+
+    userEvent.type(view.getByRole('textbox'), searchTerm);
+    expect(view.getByRole('textbox')).toHaveValue(searchTerm);
+
+    await waitFor(() => {
+      const visibleList = within(lists[1]).getAllByRole('listitem');
+      expect(visibleList[0].textContent).toEqual(camelCaseToListItem(searchTerm));
```
hamidzr, line 99: 👍🏼

hamidzr (approved):
LGTM.

couple of non-blocking notes about the reset text here
Screen Shot 2022-02-07 at 15 03 34

- might work well as a modal button too
- UX: I expected it to reset it to what the column lists looked liked when I opened the modal (my last saved set of columns) but it resets to a set of default columns. Which one do we want? It'd be good to clarify that either by a tooltip or a another way

EmilyBonar:
Yeah, I had the same question about the behavior of the Reset button, but after talking with Alex we decided to leave it like this.

PR #3565, opened 2022-02-04
Minor edits to conform to style guide

gh-determined-ai:
## Checklist

### DIFF
--- a/docs/features/notebooks.txt
+++ b/docs/features/notebooks.txt
@@ -161,8 +161,8 @@ option:
 
 See :ref:`command-notebook-configuration` for details on the supported configuration options.
 
-Finally, to configure notebooks to run a predefined set of commands at startup, you can include
-a :ref:`startup hook <startup-hooks>` in a directory specified with the ``--context`` option:
+Finally, to configure notebooks to run a predefined set of commands at startup, you can include a
+:ref:`startup hook <startup-hooks>` in a directory specified with the ``--context`` option:
 
 .. code::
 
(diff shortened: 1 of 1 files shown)


gh-determined-ai:
Just experimenting with the process.

docs/features/notebooks.txt
```diff
@@ -161,8 +161,8 @@ In addition to the ``--config`` flag, configuration may also be supplied via a Y
 
 See :ref:`command-notebook-configuration` for details on the supported configuration options.
 
-Finally, to configure notebooks to run a predefined set of commands at startup, you can use a
-:ref:`startup hook <startup-hooks>` along with the ``--context`` option:
+Finally, to configure notebooks to run a predefined set of commands at startup, you can specify a
+:ref:`startup hook <startup-hooks>` with the ``--context`` option:
```
dzhu, line 165: I can see that the original phrasing was a little clumsy, though I think this new version has the issue that, in isolation, it makes it sound like `--context` is specifically for specifying startup hooks. (Which a reader will know is not the case if they also read the part about `--context` above, but that's kind of far away.) Maybe something like this?
```
you can include a :ref:`startup hook <startup-hooks>` in a directory specified with the ``--context`` option
```

dzhu (approved):

docs/features/notebooks.txt
gh-determined-ai: This is only to exercise the process but I'll make you're suggested change.

PR #3570, opened 2022-02-04
fix: actually send the context directory for notebooks [DET-6634]

dzhu:
## Description

When the CLI started switching over to the new binding code, handling of
the context argument got left out of the notebook launch function, so
here we put that back in. Just need to do a bit of translation to the
new type for a context file.

## Test Plan

- [x] run `det notebook start -c ctx`, where `ctx` contains a startup hook, a `.detignore`, another file, another file to ignore, and some nested directories with files in them; check that the expected contents get through and the startup hook runs

## Commentary (optional)

Should be pretty straightforward, but I haven't investigated to see if there are any subtle mismatches in semantics between the corresponding fields of `ContextItem` and `v1File`. One visible difference is that `content` is `bytes` in one and `str` in the other, hence the `.decode()`, but the value is always a Base64-encoded string, so that shouldn't be a problem.

### DIFF
--- a/harness/determined/cli/notebook.py
+++ b/harness/determined/cli/notebook.py
@@ -9,6 +9,7 @@ from determined.cli.session import setup_session
 from determined.common import api
 from determined.common.api import authentication, bindings
 from determined.common.check import check_eq
+from determined.common.context import Context
 from determined.common.declarative_argparse import Arg, Cmd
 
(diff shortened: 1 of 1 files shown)


hamidzr (approved):
lgtm. thanks for the fix. we might want a shared way of going from Context to a list of v1Fil as some of our other endpoints use them as well.

https://github.com/determined-ai/api-bindings/blob/main/examples/python/main.py#L35 do we want to specify what the encoding is when we decode? 

also the e2e test failures seem relevant.

PR #3571, opened 2022-02-04
chore: handle setResourcePool message in commands

eecsliu:
## Description
Change the command behavior so instead of returning an error for unrecognized message, we will respond with an unsupported action error.

## Checklist


master/internal/command/command.go
```diff
@@ -321,12 +321,22 @@ func (c *command) Receive(ctx *actor.Context) error {
 			ctx.Respond(err)
 		}
 
+	case job.SetResourcePool:
+		err := c.setResourcePool(ctx, msg)
```
hamidzr, line 325: nit: could be one line.

master/internal/command/command.go
```diff
@@ -321,12 +321,22 @@ func (c *command) Receive(ctx *actor.Context) error {
...
+			ctx.Respond(err)
+		}
+
 	default:
 		return actor.ErrUnexpectedMessage(ctx)
 	}
 	return nil
 }
 
+func (c *command) setResourcePool(ctx *actor.Context, msg job.SetResourcePool) error {
```
hamidzr, line 336: if we're going with a sep fn like this. how about having the arg be the resourcepool name?

master/internal/command/command.go
```diff
@@ -321,12 +321,22 @@ func (c *command) Receive(ctx *actor.Context) error {
...
+	return fmt.Errorf("setting resource pool for job type %s is not supported", c.jobType)
```
hamidzr, line 337: IMO it'd be nice to include the why in a comment in code.

hamidzr (approved):

master/internal/command/command.go
eecsliu: sure that makes sense. This function is mostly to set up future implementation.

PR #3577, opened 2022-02-07
chore: unequate reservation existence and the fact they've been started

stoksc:
## Description
This change adds a new bool to the allocation actor to track when reservations have been started. `test_cancel_ten_experiment` was flaking occasionally, and it was caused by falsely equating `len(reservations) > 0` to "there are containers we should try to kill, and we should wait until we receive receipt of them being killed". This wasn't true when they are set at the top of `ResourcesAllocated` and we exited due to a DB failure, the experiment already being canceled, etc, before starting them. This caused allocations to hang waiting for `ContainerStopped` messages, which agents never send if you request a non-existent container to stop. This should make this happen less but still isn't perfect; in the event a container fails to launch this may still happen (since we would think we started it, but didn't). But, when starting reservations is truly an ask that can return all errors from the agent, this works (everything being completely async is slightly painful). This work is happening separately.

## Test Plan
- [x] run tests. since this is an inter-component bug that only happens in certain order of ops, very hard to write a test for, without forcing it to an extent it is unhelpful.

alternative solution is to always send back an `InvalidContainerRequest` (just sending a fake `ContainerStopped` is annoying because everywhere between the container and allocation actors try to read something off it) and something in the allocation actor handle it accordingly, but it'd be sweet if FT/reliability could _not_ be pushed up into lots of application layer specific logic.

relevant log snippet:
```
master: INFO[2022-02-04T17:39:19Z] experiment state changed to STOPPING_CANCELED  id=20 system=master type=experiment
master: INFO[2022-02-04T17:39:19Z] starting checkpoint garbage collection        id=experiment-18-checkpoint-gc system=master type=checkpointGCTask
master: DEBU[2022-02-04T17:39:19Z] patching state after request (STOPPING_CANCELED)  experiment-id=20 id=e01ed4fe-7a6a-4d89-ba95-46d0f9587713 system=master type=trial
master: INFO[2022-02-04T17:39:19Z] trial changed from state ACTIVE to STOPPING_CANCELED  experiment-id=20 id=e01ed4fe-7a6a-4d89-ba95-46d0f9587713 system=master type=trial
master: INFO[2022-02-04T17:39:19Z] decided to terminate trial                    experiment-id=20 id=e01ed4fe-7a6a-4d89-ba95-46d0f9587713 system=master type=trial
master: ERRO[2022-02-04T17:39:19Z] allocation encountered fatal error            error="could not get task spec: the allocation was canceled while this message was waiting" id=1 system=master type=Allocation
master: INFO[2022-02-04T17:39:19Z] decided to kill allocation                    id=1 system=master type=Allocation
master: DEBU[2022-02-04T17:39:19Z] still inside of kill cooldown                 id=1 system=master type=Allocation
master: INFO[2022-02-04T17:39:19Z] killing container id: 06fa4da6-e102-4f34-a0d7-8a8c6bb6d822  id=agent1 system=master type=agent
agent1: WARN[2022-02-04T17:39:19Z] error signaling container with killed, container not found: 06fa4da6-e102-4f34-a0d7-8a8c6bb6d822  id=containers system=agent1 type=containerManager
```

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -39,8 +39,10 @@ type (
 		// The state of the allocation, just informational.
 		state model.AllocationState
 
-		// State of all our reservations
+		// State of all our reservations.
 		reservations reservations
+		// Separates the existence of reservations from us having started them.
+		reservationsStarted bool
 		// Tracks the initial container exit, unless we caused the failure by killed the trial.
 		exitReason error
 		// Marks that we intentionally killed the allocation so we can know to
@@ -353,6 +355,7 @@ func (a *Allocation) ResourcesAllocated(ctx *actor.Context, msg sproto.Resources
 			IsMultiAgent: len(a.reservations) > 1,
 		})
 	}
+	a.reservationsStarted = true
 	if a.req.StreamEvents != nil {
 		ctx.Tell(a.req.StreamEvents.To, sproto.Event{
 			State:         a.state.String(),
@@ -467,6 +470,9 @@ func (a *Allocation) TaskContainerStateChanged(
 // Exit attempts to exit an allocation while not killing or preempting it.
 func (a *Allocation) Exit(ctx *actor.Context) (exited bool) {
 	switch {
+	case !a.reservationsStarted:
+		a.terminated(ctx)
+		return true
 	case len(a.reservations) == len(a.reservations.exited()):
 		a.terminated(ctx)
 		return true


stoksc:
open to discuss alternatives here, lmk what you think @ioga

master/internal/task/allocation.go
```diff
@@ -353,6 +355,7 @@ func (a *Allocation) ResourcesAllocated(ctx *actor.Context, msg sproto.Resources
 			IsMultiAgent: len(a.reservations) > 1,
 		})
 	}
+	a.reservationsStarted = true
```
ioga, line 358: Let's say
1. in line 352 here,`r.Start`, tells agent actor to start a task
2. we set reservationStarted here.
3. agent network connection breaks and it cannot write message # 1

then I suspect in the current code we'll *probably* receive `ContainerStopped` when agent actor will die one way or another?... is that right?

ioga (approved):

master/internal/task/allocation.go
stoksc: yeah, that is basically my thought exactly

stoksc: so long as we dont fail without handling it correctly, other downstream failures should cause us to handle this correctly

PR #3584, opened 2022-02-07
chore: add additional flags to pid_server

liamcli:
## Description
Add grace period and option to return error code on pid_client error. 

We can start a pid_server with a cmd via `run_subprocess`.  This is how we launch horovodrun, which in turn will spawn pid_client worker processes that report to the server.  In the case of deepspeed, the launch is handled through pdsh, which ssh's into each node to launch the harness.  pdsh will return 0 even if the ssh processes it spawned errored out so we need to have the error surface through to the pid_server to error correctly.   

## Checklist

### DIFF
--- a/harness/determined/ipc.py
+++ b/harness/determined/ipc.py
@@ -609,6 +609,7 @@ class PIDServer:
                 return
             elif data[-1:] == b"q":
                 # Graceful shutdown code.
+                logging.debug(f"pid {pid} shutdown gracefully")
                 self.graceful_shutdowns.append(pid)
             else:
                 raise ValueError("invalid message from pid_client:", data)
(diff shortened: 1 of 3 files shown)


harness/determined/ipc.py
```diff
@@ -713,7 +719,9 @@ def health_check() -> None:
                 except subprocess.TimeoutExpired:
                     logging.error(f"killing worker which didn't exit after {on_exit.name}")
                     p.send_signal(signal.SIGKILL)
-        return p.wait()
+        exit_code = p.wait()
+        logging.debug(f"exit_code: {exit_code}")
```
rb-determined-ai, line 723: if you leave this in, you'll have to provide more information on where this log line is coming from.

PR #3585, opened 2022-02-08
fix: remove icon for LOG_LEVEL_UNSPECIFIED log [DET-6436]

gt2345:
## Description

Remove icon for log with level `LOG_LEVEL_UNSPECIFIED`

## Test Plan

Navigate to log tab and verify log with level `LOG_LEVEL_UNSPECIFIED` not have an icon in front of the row.

Before:
Screen Shot 2022-02-07 at 4 20 03 PM

After:
Screen Shot 2022-02-07 at 4 19 48 PM


EmilyBonar (approved):
LGTM

PR #3586, opened 2022-02-08
fix: check if data file exists before downloading

mpkouznetsov:
## Description

The previous fix to test_unets_tf_keras_distributed protected model data file loading with filelock but did not forego loading file if it was already present. As a result, other workers started downloading the file as soon as they got the lock and the file appeared as truncated to some workers. The solution is to check if file exists inside the lock and not download it again if it does.

## Checklist

### DIFF
--- a/examples/computer_vision/unets_tf_keras/model_def.py
+++ b/examples/computer_vision/unets_tf_keras/model_def.py
@@ -61,7 +61,8 @@ class UNetsTrial(TFKerasTrial):
 
         # Use a file lock so only one worker on each node does the download
         with filelock.FileLock(os.path.join(weights_dir, "download.lock")):
-            urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
+            if not os.path.exists(weights_dir + data_file):
+                urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
         return weights_dir + data_file
 
     def build_model(self):
--- a/examples/computer_vision/unets_tf_keras/model_def.py
+++ b/examples/computer_vision/unets_tf_keras/model_def.py
@@ -5,8 +5,6 @@ regarding the optional flags view the original script linked below.
 This implementation is based on:
 https://github.com/tensorflow/docs/blob/master/site/en/tutorials/images/segmentation.ipynb
 """
-import tempfile
-
 import filelock
 import tensorflow as tf
 from tensorflow_examples.models.pix2pix import pix2pix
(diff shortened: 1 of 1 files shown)


examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -61,7 +61,8 @@ def download_weights(self):
 
         # Use a file lock so only one worker on each node does the download
         with filelock.FileLock(os.path.join(weights_dir, "download.lock")):
-            urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
+            if not os.path.exists(weights_dir + data_file):
+                urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
```
rb-determined-ai, line 65: What do you think about doing some sort of protection to ensure we recover well if the download fails halfway through?  Maybe just use a `try/except` block to delete the file if `urlretrieve()` fails?

Or does `urlretrieve` ensure that partial downloads are deleted?  It doesn't document any behavior like that.

mpkouznetsov: I think if `urlretrieve` raises an exception we are in bad shape, anyway.

rb-determined-ai: I guess I'm thinking of cases where every execution is not done in a clean docker filesystem.  If we had local training mode, or if the user bind-mounted  /tmp/data into the container to save time on consecutive trial executions, then we would want subsequent runs to be able to complete successfully in spite of the initial failure.

mpkouznetsov: as we discussed offline, rather than catching the exception and deleting the file, I am downloading into a temporary file, then copying to desired destination.

examples/computer_vision/unets_tf_keras/model_def.py
```diff
@@ -61,8 +63,12 @@ def download_weights(self):
 
         # Use a file lock so only one worker on each node does the download
         with filelock.FileLock(os.path.join(weights_dir, "download.lock")):
-            urllib.request.urlretrieve(mobilenet_link, weights_dir + data_file)
-        return weights_dir + data_file
+            full_weights_path = weights_dir + data_file
+            if not os.path.exists(full_weights_path):
+                with tempfile.NamedTemporaryFile(delete=False) as ntf:
+                    urllib.request.urlretrieve(mobilenet_link, ntf.name)
+                    os.rename(ntf.name, full_weights_path)
```
rb-determined-ai, line 70: Since you can't rename across filesystem boundaries, I don't think this is a safe approach.

But a simpler approach which is extremely unlikely to hit a filesystem boundary issue would be:
```python
if not os.path.exists(full_weights_path):
  urllib.request.urlretrieve(mobilenet_link, full_weights_path + ".part")
  os.rename(full_weights_path + ".part", full_weights_path)
```

mpkouznetsov: ok makes sense

rb-determined-ai (approved):

PR #3589, opened 2022-02-08
fix: use google.com instead of example.com as trusted domain

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/DET-6687

Occasionally https://example.com/ responds with 404. This causes `test_custom_tls_certs` to fail. Use https://google.com instead.

## Checklist


mapmeld (approved):
:+1: thanks for doing this

PR #3591, opened 2022-02-08
chore: upgrade black

neilconway:
Black now has a stable release, so pin our dependency on that to avoid future
spurious reformatting changes. Reformat a few files to make the new version of
Black happy.

Switch from using "--exclude" to "--extend-exclude", because the former disables
the built-in exclusion rules and that is not what we want. Add "build" to the
exclusion list for harness; this _should_ be part of the default exclusion list,
but it does not appear to be.

## Test Plan

"make check" failed before; it passes now.

## Checklist

### DIFF
--- a/harness/pyproject.toml
+++ b/harness/pyproject.toml
@@ -1,4 +1,4 @@
 [tool.black]
 line-length = 100
-exclude = '(_gen.py|determined/_swagger/client/*)'
+extend-exclude = '(_gen.py|determined/_swagger/client/*|determined/common/api/bindings.py)'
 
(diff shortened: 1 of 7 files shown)


harness/pyproject.toml
```diff
@@ -1,4 +1,4 @@
 [tool.black]
 line-length = 100
-exclude = '(_gen.py|determined/_swagger/client/*)'
+extend-exclude = '(_gen.py|determined/_swagger/client/*)'
```
dzhu, line 3: I guess this is no worse than before, but `--extend-exclude` completely overrides this, so currently this line would never actually be used. I figure we can either change this to `exclude` and add `build` explicitly or just remove this entirely. Or forget my other suggestion and put it all in one place.

dzhu (approved):

harness/pyproject.toml
neilconway: Good catch!

PR #3592, opened 2022-02-08
chore: handle possible errors from docker daemon call /containers/cID/wait

stoksc:
## Description
It is possible for this Docker API call to fail with an error the body, but our code ignores the error and would always return container exited with 0, causing upstream consumers to mistake it for a user requested stop (real exit 0).

## Test Plan
- [x] test no harm (exit code 1 still looks the same)

## Commentary (optional)
From reading https://github.com/moby/moby/blob/a5c757555091df6bd1e9c60c175d167dcdfde83c/container/state.go#L173-L236 i'm not super convinced we could ever see then but still trying to read code.

## Checklist

### DIFF
--- a/agent/internal/docker.go
+++ b/agent/internal/docker.go
@@ -244,6 +244,10 @@ func (d *dockerActor) runContainer(ctx *actor.Context, msg cproto.RunSpec) {
 	case err = <-eerr:
 		sendErr(ctx, errors.Wrap(err, "error while waiting for container to exit"))
 	case exit := <-exit:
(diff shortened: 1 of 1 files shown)


stoksc:
gimme a sec, reading the source to understand the meaning of this err field

ioga (approved):
what's the context here, where did that happen, what issue is being addressed?..

stoksc:
i can't see how the daemon could return this from reading the code (it's possible an older version did or something..), but it's totally no harm so i will just land.

PR #3593, opened 2022-02-08
feat: log records/second and batches/second in trials [DPS-1]

rb-determined-ai:
For now, cumulative times are not recorded or shown.  Cumulative times
wouldn't be hard to add but they would introduce a variety of failure
cases where we could show something but we wouldn't know if it were
correct.

### DIFF
--- a/harness/determined/keras/_tf_keras_trial.py
+++ b/harness/determined/keras/_tf_keras_trial.py
@@ -5,8 +5,9 @@ import pathlib
 import pickle
 import random
 import sys
+import time
 from abc import abstractmethod
-from typing import Any, Dict, List, Optional, Type, cast
+from typing import Any, Dict, List, Optional, Tuple, Type, cast
 
 import h5py
 import numpy as np
@@ -109,6 +110,7 @@ class TrialControllerMultiplexer(keras.callbacks._MultiplexerBase):
         super().__init__(*arg, **kwarg)
         self.trial_controller = trial_controller
         self.test_inputs = 0
+        self.test_batches = 0
 
     def on_train_begin(self, logs: Optional[Dict] = None) -> None:
         super().on_train_begin()
@@ -129,18 +131,20 @@ class TrialControllerMultiplexer(keras.callbacks._MultiplexerBase):
     def on_test_begin(self, logs: Optional[Dict] = None) -> None:
         super().on_test_begin(logs)
         self.test_inputs = 0
+        self.test_batches = 0
 
     def on_test_batch_end(self, batch: int, logs: Optional[Dict] = None) -> None:
         super().on_test_batch_end(batch, logs)
         assert isinstance(logs, dict)
         self.test_inputs += logs.get("size", self.batch_size)
+        self.test_batches += 1
 
     def _corrected_test_end(self, logs: Dict) -> None:
         super()._corrected_test_end(logs)
         self.trial_controller._stop_training_check()
 
-    def get_test_inputs(self) -> int:
-        return self.test_inputs
+    def get_test_inputs(self) -> Tuple[int, int]:
+        return self.test_inputs, self.test_batches
 
     def _corrected_epoch_end(self, epoch: int, logs: Dict) -> None:
         super()._corrected_epoch_end(epoch, logs)
@@ -800,6 +804,7 @@ class TFKerasTrialController(det.TrialController):
                     # Configure the state for a training step.
                     self.train_response_func = response_func
                     self.train_workload_batches = 0
+                    self.train_workload_inputs = 0
                     self.train_workload_metrics = []
                     self.train_workload_len = wkld.num_batches
                     self.multiplexer.set_batches_requested(wkld.num_batches)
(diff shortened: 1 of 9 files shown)


rb-determined-ai:
This needs an integration test, especially to make sure that all of the values are correct in dtrain situations.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -595,12 +598,16 @@ def _compute_validation_metrics(self) -> workload.Response:
                 )
                 callback.on_validation_step_end(metrics)
 
+        step_duration = time.time() - step_start_time
```
liamcli, line 601: i'm not sure whether it's better to have duration end here or after the `on_validation_end` callbacks.  i'm leaning towards after the callbacks since that is technically when the validation workload completes.  what do you think?

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -946,15 +963,21 @@ def _compute_validation_metrics(self) -> workload.Response:
             if isinstance(num_inputs, EagerTensor):
                 # Horovod will promote an int to a tensor in eager mode.
                 num_inputs = num_inputs.numpy()
+            num_batches = hvd.allreduce(num_batches, average=False, name="validation_num_batches")
+            if isinstance(num_batches, EagerTensor):
+                num_batches = num_batches.numpy()
 
         metrics = self._allreduce_logs(metrics)
         check.gt(len(metrics), 0)
+        step_duration = time.time() - validation_start_time
```
liamcli, line 977: same here. should we consider moving the duration to after _test_end.

liamcli (approved):

rb-determined-ai:
upstream tests running at #3597

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: Made the change to include the pre- and post-workload callbacks in the timings.  Also fixed another bug or two.  Please re-review.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -556,7 +559,12 @@ def _compute_validation_metrics(self) -> workload.Response:
                 metrics_reducers=self._prepare_metrics_reducers(keys=keys),
             )
 
-            num_inputs *= self.context.distributed.size
+            # Gather a list of per-worker (num_inputs, num_batches) tuples.
+            input_counts = self.context.distributed._zmq_gather((num_inputs, idx + 1))
```
liamcli, line 563: note: i think we'll actually run into a hang if dataloader lengths are not the same across slots so idx should be the same.

harness/determined/keras/_tf_keras_trial.py
```diff
@@ -900,8 +905,10 @@ def _post_train_batch_end(self, num_inputs: int, logs: Dict) -> None:
             )
 
         if self.context.distributed.size > 1:
-            num_inputs = self._hvd_allreduce(num_inputs, average=False, name="train_num_inputs")
-            num_inputs = self._convert_possible_tensor(num_inputs)
+            self.train_workload_inputs = self._hvd_allreduce(
```
liamcli, line 908: if we start supporting native dtrain backend, should we replace this call with torch.distributed primitive?  or should we add a zmq allreduce?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -569,7 +577,8 @@ def _compute_validation_metrics(self) -> workload.Response:
                 )
 
                 metrics = self._convert_metrics_to_numpy(metrics)
-                num_inputs = self.context.get_per_slot_batch_size() * len(self.validation_loader)
+                num_batches = len(self.validation_loader)
+                num_inputs = self.context.get_per_slot_batch_size() * num_batches
```
liamcli, line 581: this seems like a strong assumption that the dataloader used for evaluate_full_dataset returns the per slot batch.  what do you think about creating an iterator over dataloader to get a batch and use that for batch size?

harness/determined/keras/_tf_keras_trial.py
rb-determined-ai: this is ... keras?  I don't expect TFKerasTrial to work with non-horovod ever

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: I don't think that's the case, I think validation is all done asynchronously on each shard and metrics are reduced after all are finished.

liamcli: > I don't think that's the case, I think validation is all done asynchronously on each shard and metrics are reduced after all are finished.

Yep, I see.  It is okay if model forward calls have no synchronized distributed communication but blocking if there is something like sync batch norm.

harness/determined/keras/_tf_keras_trial.py
liamcli: i thought there was something similar in pytorch but all i see is
`hvd.broadcast_object(metrics...)`
which we can replace with zmq as part of native dtrain probably.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: what would you say about just _not_ logging this data for `evaluate_full_dataset`?

I'm thinking this is an infrequently used feature, and the only real reason to use it is if you are doing something pretty custom (multiple passes through the dataset or something?).  I'm also thinking they can basically gather the same metric on their own in this case.

liamcli: excluding this for `evaluate_full_dataset` make sense to me.

PR #3595, opened 2022-02-09
chore: containers should always be killed by their parent, remove incorrect exits with 0

stoksc:
## Description
Previously if a container was killed through the agent/slot disable code path, it would kill containers directly rather than telling the allocation actor to kill itself. In all cases, a spurious container death will cause the allocation to die anyway and `det a --drain` is available when killing the allocation is not optimal (despite its issues, this will be true eventually). Moreover, a spurious container exit while the container is pulling causes the allocation actor to just receive an empty `containerStopped{}` which is indistinguishable from an exit 0. The allocation actor would interpret that exit incorrectly as a container deciding to exit early. 

This change makes kills always go through the allocation actor, so logic like counting towards max restarts is taken into account, and also fixes this bug since the allocation knows it was killed. Also a new container exit 'ContainerAborted' is added for cases where a container is killed before the agent can start it, rather than just returning 'container exited 0' since this is just wrong and could easily other issues.

## Test Plan
- [x] run an experiment and disable agents. it should reschedule itself and re-enabling the agent it should allow the job to run to completion. 

## Commentary (optional)
Separately the new exit failure type and killing through the allocation actor both fix the bug that was seen. But also separately, they're both more 'correct' than they were previously, and it's possible the same error manifested through their 'incorrectness' in ways I haven't seen through static analysis. E.G. returning 'exited with 0' ever when we haven't is nonsense, and could cause other bugs on its own.

## Checklist

### DIFF
--- a/agent/internal/container.go
+++ b/agent/internal/container.go
@@ -122,14 +122,15 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 		case cproto.Assigned, cproto.Pulling:
 			switch msg.Signal {
 			case syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL:
-				msg := "attempting to stop container while in [%s] state"
-				ctx.Log().Infof(msg, c.State)
+				msg := "attempting to stop container %s while in [%s] state"
+				ctx.Log().Infof(msg, c.ID, c.State)
 				ctx.Self().Stop()
 				c.containerStopped(ctx,
-					aproto.ContainerError(aproto.ContainerAborted, fmt.Errorf(msg, c.State)))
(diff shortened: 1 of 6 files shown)


agent/internal/container.go
```diff
@@ -122,9 +122,11 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 		case cproto.Assigned, cproto.Pulling:
 			switch msg.Signal {
 			case syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL:
-				ctx.Log().Infof("attempting to stop container while in [%s] state", c.State)
+				msg := "attempting to stop container while in [%s] state"
+				ctx.Log().Infof(msg, c.State)
```
ioga, line 126: should we add a container id?

ioga (approved):

PR #3597, opened 2022-02-09
(upstream testing branch) feat: log records/second and batches/second in trials

rb-determined-ai:
this branch only for running upstream tests.


rb-determined-ai:
tests look good

PR #3601, opened 2022-02-09
fix: add v4 shim due to experiment snapshot change

stoksc:
## Description
The format of the experiment snapshot was changed without a shim, causing paused experiments to fail on upgrade. This change adds a shim for v4 to v5 snapshots, and bumps the current version to v5.

## Test Plan
- [x] wrote a test
- [x] start an experiment on 0.17.7, pause it, upgrade. it should live. 

## Checklist


master/internal/restore_test.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestShimExperimentSnapshotV4(t *testing.T) {
```
rb-determined-ai, line 9: This looks good.

Now that we have a test with an embedded snapshot, can we write a test with a v5 snapshot, and ensure that it is loadable into the current system?  I feel like this sort of bug ought to be preventable.

master/internal/restore.go
```diff
@@ -340,3 +341,62 @@ func shimExperimentSnapshotV2(snapshot []byte) ([]byte, error) {
 
 	return json.Marshal(experimentSnapshotV2)
 }
+
+// Version 3 => 4 had no shim, it was breaking.
+
+// Version 4 => 5 shims
```
rb-determined-ai, line 347: stray comment?  Or what is this saying?

rb-determined-ai (approved):

master/internal/restore_test.go
stoksc: Yeah, good idea, let me take a shot at that.

master/internal/restore.go
stoksc: eh, i'll remove these if they're confusing at all

PR #3602, opened 2022-02-09
chore: test gradient aggregation DET-793

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/DET-793

A test for correctness of gradient computation with gradient aggregation enabled. When aggregation frequency is set to n, only one batch out of n results in actually updating the model weights. As the n-1 batches are processed, the gradients are being aggregated, and after the n-th batch is processed, the model weights are updated with the average gradient (if `average_aggregated_gradients == True`).

This test computes the expected weights, taking advantage of the simplicity of the dataset and the model: the dataset is y = x, with x = 1.0, 1.1, 1.2, ... , and the model is a single linear layer of size 1. The expected weights are then compared with the actual weights of the model that are emitted as a validation metric.

## Commentary (optional)

I am borrowing logic from pytorch_onevar_model.py with a couple of differences:
- the dataset represents an identity function (y = x) but the values of x are 1.0 + i*0.1.
- the weight check is performed in the test driver and not in the model code. The model is just responsible for emitting its actual weights as part of `evaluate_batch`.

## Checklist


e2e_tests/tests/fixtures/pytorch_identity/model_def.py
```diff
@@ -0,0 +1,64 @@
...
+
+class IdentityDataset(torch.utils.data.Dataset):
+    def __init__(self, initial_value: int = 0):
+        self.initial_value = initial_value
+
+    def __len__(self) -> int:
+        return 64
+
+    def __getitem__(self, index: int) -> Tuple:
+        v = 1  # float(self.initial_value + index)
```
rb-determined-ai, line 16: I'm thinking that this test does not inspire enough confidence given that every batch is going to have the same data.

Can you make it a triangle dataset, like what you have in the comment?  Probably the slope of the dataset needs to be less than one or your SGD will explode (meaning, `self.initial_value + 0.001 * index` or something)

e2e_tests/tests/experiment/test_pytorch.py
```diff
@@ -165,6 +166,37 @@ def test_pytorch_native_api() -> None:
...
+
+    config_with_grad_agg = copy.deepcopy(base_config)
+    config_with_grad_agg["hyperparameters"]["global_batch_size"] = 4
+    config_with_grad_agg["optimizations"]["aggregation_frequency"] = 2
+    loss_with_aggregation = run_and_check(config_with_grad_agg, 80)
+
+    assert loss_with_aggregation[-1] == pytest.approx(
+        loss_without_aggregation[-1], 1e-4
+    ), f"{loss_with_aggregation}!={loss_without_aggregation}"
+    assert loss_with_aggregation[-1] == pytest.approx(0.852, 1e-4)
```
rb-determined-ai, line 193: I would personally find the test more convincing if you were able to hand-calculate what the gradient updates should be and check the weights do what we think they ought to.  That would also make the test run 2x as fast since you'd only need to launch one experiment instead of two.

And lastly, I would think on the cpu we can come closer than 1e-4.  I would think we'd almost be bit-perfect.

rb-determined-ai (commented):
I'm happy to help you come up with the hand-calculated weight values so we can turn the code-vs-code comparison into a math-vs-code comparison.

e2e_tests/tests/experiment/test_pytorch.py
```diff
@@ -165,6 +165,61 @@ def test_pytorch_native_api() -> None:
...
 
+@pytest.mark.parallel
+def test_pytorch_gradient_aggregation() -> None:
+    config = conf.load_config(conf.fixtures_path("pytorch_identity/distributed.yaml"))
+
+    exp_id = exp.run_basic_test_with_temp_config(config, conf.fixtures_path("pytorch_identity"), 1)
+    trials = exp.experiment_trials(exp_id)
+    assert len(trials) == 1
+    steps = trials[0]["steps"]
+    print(steps[0])
```
rb-determined-ai, line 176: delete the extra `print()`

PR #3604, opened 2022-02-10
chore: adding a limit to task endpoints by type

hkang1:
## Description

Limit each 4 task type endpoints from loading everything. Set a hard cap to something reasonable like 1000 per type.

## Test Plan

- [ ] load tasks list page and check that each network call for tasks is limited to 1000

## Checklist


johnkim-det (approved):
looks good. one thought: might make sense to set this in apiConfig, unless they'd be called with different limits from different components.

hkang1:
> 

Good idea, we can enforce a guard rail there so you can't over-request any of these task types.

PR #3605, opened 2022-02-10
fix: log viewer bug

johnkim-det:
## Description

Fix for bug where log viewer was flickering (scrolling to top) and some log entries would disappear. Flickering/scrolling issue was more noticeable in Safari but both issues appeared in both Safari and Chrome.

https://user-images.githubusercontent.com/97752292/153456995-9b812e8d-114c-4624-a507-838d12137815.mov

## Test Plan

- [ ] Run experiments and make sure that the logs display without flickering or disappearing issues. 

## Commentary (optional)

Seemed to be an issue where the update to `logs.length` made the `handleItemsRendered` callback run before `visibleStopIndex` updated. There may be another way to make sure these values match, but checking that `visibleStopIndex` matches `logs.length - 2` instead of `logs.length - 1` seems to fix the problem.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewerCore.tsx
+++ b/webui/react/src/components/LogViewerCore.tsx
@@ -188,7 +188,6 @@ const LogViewerCore: React.FC<Props> = ({
     if (!local.current.isScrollReady) return;
 
     local.current.isOnTop = visibleStartIndex === 0;
-    // mismatch between visibleStopIndex and logs updates causes flickering and disappearing logs:
     local.current.isOnBottom = visibleStopIndex === logs.length - 1;
 
     setIsTailing(local.current.isOnBottom && isNewestFirst);
--- a/webui/react/src/components/LogViewerCore.tsx
+++ b/webui/react/src/components/LogViewerCore.tsx
@@ -188,7 +188,8 @@ const LogViewerCore: React.FC<Props> = ({
     if (!local.current.isScrollReady) return;
 
     local.current.isOnTop = visibleStartIndex === 0;
-    local.current.isOnBottom = visibleStopIndex === logs.length - 1;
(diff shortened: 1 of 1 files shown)


webui/react/src/components/LogViewerCore.tsx
```diff
@@ -188,7 +188,8 @@ const LogViewerCore: React.FC<Props> = ({
     if (!local.current.isScrollReady) return;
 
     local.current.isOnTop = visibleStartIndex === 0;
-    local.current.isOnBottom = visibleStopIndex === logs.length - 1;
+    // mismatch between visibleStopIndex and logs updates causes flickering and disappearing logs:
+    local.current.isOnBottom = visibleStopIndex + 1 === logs.length - 1;
```
hkang1, line 192: **question:** I did some console logging around here:
```
console.log('isOnBottom', local.current.isOnBottom);
console.log('start', visibleStartIndex, 'stop', visibleStopIndex, 'logs.length', logs.length);
```
And it looks like majority of the time, `isOnBottom` is false even on the tail end because `visibleStopIndex` is usually equal to `logs.length - 1`. So maybe it would be better to use `>=` instead of `===` for comparison?

hkang1: Screen Shot 2022-02-15 at 11 22 12 AM

Here's the example of where the tailing is enabled button doesn't show because `visibleStopIndex` is exactly `logs.length - 1`

hkang1 (approved):
After looking at this quite a bit I think there are some other weird nuances, but this PR does seem to address the flickering and scrolling to top issue. I'll keep looking at the other issues. Thanks for investing this and address the issue!

PR #3607, opened 2022-02-10
chore: add additional logging for creds helper [DET-6710]

varlogtim:
How to test:

Launch an experiment and use a container on a repo which doesn't allow anonymous credential without the domain specified in your `~/.docker/config.json`

Exp config:
environment:
  image: us-east1-docker.pkg.dev/determined-ai/msuarez-test-f/cuda-11.1-pytorch-1.9-lightning-1.3-tf-2.4-gpu-9f0cb26

Make sure you don't have, `us-east1-docker.pkg.dev` in your `~/.docker/config.json` `credHelpers`.

Note that it fails with a message saying that we couldn't find the domain.

----

Update your `~/.docker/config.json` to have this:

```
{
  "auths": {...}
  },
  "credHelpers": {
    ...
    "us-east1-docker.pkg.dev": "gcloud"
  }
}
```

Run the same experiment and verify that there is a log line indicating we are using the credential helper for the domain.

### DIFF
--- a/agent/internal/docker.go
+++ b/agent/internal/docker.go
@@ -131,17 +131,25 @@ func (d *dockerActor) pullImage(ctx *actor.Context, msg pullImage) {
 			sendErr(ctx, errors.Wrap(err, "error encoding registry credentials"))
 			return
 		}
-	} else if store, ok := d.credentialStores[reference.Domain(ref)]; ok {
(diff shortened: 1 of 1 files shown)


stoksc (approved):
lgtm

PR #3614, opened 2022-02-11
fix: dtrain + native api

rb-determined-ai:
Previously, Native API-launched experiments always took the non-dtrain
codepath.

### DIFF
--- a/harness/determined/exec/launch.py
+++ b/harness/determined/exec/launch.py
@@ -20,14 +20,17 @@ def match_legacy_trial_class(arg: str) -> bool:
 
 
 def launch(experiment_config: det.ExperimentConfig) -> int:
+    slots_per_trial = experiment_config.slots_per_trial()
+    entrypoint = experiment_config.get_entrypoint()
+
     # If native is enabled, harness will load from native entrypoint command
     if experiment_config.native_enabled():
-        from determined.exec import harness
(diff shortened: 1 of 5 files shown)


harness/determined/exec/launch.py
```diff
@@ -20,14 +20,17 @@ def match_legacy_trial_class(arg: str) -> bool:
...
-
-        return harness.main(train_entrypoint=None)
+        if slots_per_trial < 2:
+            from determined.exec import harness
 
-    entrypoint = experiment_config.get_entrypoint()
-    slots_per_trial = experiment_config.slots_per_trial()
+            return harness.main(train_entrypoint=None)
+        else:
+            entrypoint = ["python3", "-m", "determined.launch.autohorovod", "__NATIVE__"]
```
azhou-determined, line 33: what if we pass this in as a flag? --native true or something?

rb-determined-ai: basically because we're landing it at the last minute, and this is the smallest change to fix the bug, which decreases the surface area of other new bugs we might introduce while we are hotfixing.

In general I would agree with your approach but I think we only have one Native API user in the whole world, and it's been deprecated for 1.5 years.

azhou-determined (approved):

PR #3615, opened 2022-02-11
feat: support gcs backend for terraform state in `det deploy gcp`.

ioga:
## Description

Add an option to store terraform state of `det deploy gcp` in a provided GCS bucket. This enables users to update or delete a cluster even if they lose the local terraform state files, and it enables us to create GCP clusters via CI, and be able to clean them up later on crashes, e.g. in terminator. 

## Test Plan

0. Make an initial working directory for a cluster (`mkdir -p ~/dev/cluster/gcp/lost-1 && cd !$`)
1. Create a cluster: `det deploy gcp up --cluster-id ilia-test-ci-4 --project-id determined-ai-ci --max-dynamic-agents 2 --tf-state-gcs-bucket-name determined-ai-ci-det-deploy-terraform`
2. "Lose" the initial working directory and make a new one(`mkdir -p ~/dev/cluster/gcp/lost-2 && cd !$`)
3. Dry-running with the same command will retried the state from GCS and recover what you need `det deploy gcp up --cluster-id ilia-test-ci-4 --project-id determined-ai-ci --max-dynamic-agents 2 --tf-state-gcs-bucket-name determined-ai-ci-det-deploy-terraform --dry-run`
4. `det deploy gcp down` will now work and blow away the stack.

## Commentary (optional)

GCP labels turned out to be limited: you can't label *everything* but only a subset of resources. Therefore storing TF state in GCS seems like the better solution for cleanup/recovery of lost stacks.

## Checklist

### DIFF
--- a/harness/determined/deploy/gcp/cli.py
+++ b/harness/determined/deploy/gcp/cli.py
@@ -48,7 +48,11 @@ def deploy_gcp(command: str, args: argparse.Namespace) -> None:
     if not os.path.exists(args.local_state_path):
         os.makedirs(args.local_state_path)
     os.chdir(args.local_state_path)
-    print("Using local state path:", args.local_state_path)
+
+    if args.tf_state_gcs_bucket_name:
+        print("Using GCS bucket for state:", args.tf_state_gcs_bucket_name)
+    else:
+        print("Using local state path:", args.local_state_path)
 
     # Set the TF_DATA_DIR where Terraform will store its supporting files
     env = os.environ.copy()
@@ -96,6 +100,7 @@ def deploy_gcp(command: str, args: argparse.Namespace) -> None:
         "no_wait_for_master",
         "no_prompt",
         "master_config_template_path",
+        "tf_state_gcs_bucket_name",
         "func",
         "_command",
         "_subcommand",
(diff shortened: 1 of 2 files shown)


harness/determined/deploy/gcp/gcp.py
```diff
@@ -128,6 +143,16 @@ def terraform_apply(configs: Dict, env: Dict, variables_to_exclude: List) -> Non
 
 
 def terraform_output(configs: Dict, env: Dict, variable_name: str) -> Any:
+    if configs.get("tf_state_gcs_bucket_name"):
```
mackrorysd, line 146: I'm not clear on why this is necessary. The outputs still get printed at the end when you're using a remote state store. Isn't that how we're grabbing them in the default case?

mackrorysd (approved):

harness/determined/deploy/gcp/gcp.py
ioga: we need to grab them to do a health check, i.e. check if master is up by doing a http request.

harness/determined/deploy/gcp/cli.py
```diff
@@ -48,7 +48,11 @@ def deploy_gcp(command: str, args: argparse.Namespace) -> None:
     if not os.path.exists(args.local_state_path):
         os.makedirs(args.local_state_path)
     os.chdir(args.local_state_path)
-    print("Using local state path:", args.local_state_path)
+
+    if args.tf_state_gcs_bucket_name:
```
liamcli, line 52: `det deploy gcp down` is erroring out on this line due to no arg being parsed in the down case.

PR #3618, opened 2022-02-14
fix: make SelectFilter input wide enough to see what I type in MetricSelectFilter [DPS-30] [DET-6642]

mapmeld:
## Description

On the Experiment Detail page / TrialDetails chart, in the dropdown where you select Metrics for the chart, this fixes an issue where you couldn't see what you were typing ( https://determinedai.atlassian.net/browse/DET-6642 )

The only Ant Select component / prop-level issue I could see is this happens when we set  mode='multiple', which we need.

## Test Plan

Select a metric (such as loss) by typing a metric such as 'loss' in the MetricSelectFilter dropdown of the TrialDetails chart.

## Checklist


gt2345 (approved):
Verified working as expected for Chrome, Safari, Firefox

PR #3621, opened 2022-02-14
chore: interactive test for Section [DET-6664]

gt2345:
## Description

Unit test for Section component

## Test Plan

Navigate to `webui/react` and `make test`

### DIFF
--- a/webui/react/src/components/Section.test.tsx
+++ b/webui/react/src/components/Section.test.tsx
@@ -1,0 +1,70 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+
+import Section from './Section';
+
+const setup = ({ 
+    title='', 
+    loading=false,
+    hideTitle=false, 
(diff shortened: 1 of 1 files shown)


webui/react/src/components/Section.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  it('Section hide title', () => {
+    setup({ hideTitle: true, title: 'title of section' });
+    expect(screen.queryAllByText('title of section')).toHaveLength(0);
+  });
+
+  it('Section in loading state', () => {
+    setup({ loading: true, title: 'section-title' });
+    expect(screen.getByText('section-title')).toBeInTheDocument();
+  });
```
mapmeld, line 49: Could this test instead have two expects: checking the Spinner component prop spinning=true, and check that section-filters are hidden?

mapmeld (approved):

PR #3622, opened 2022-02-15
fix: allow list hparams in local mode

coreystatendet:
## Description

Allow list hparams in --local mode.

## Test Plan

Run an experiment that uses a list in hparams and ensure it does not error out during loading.  A good example is `cd examples/computer_vision/byol_pytorch && det experiment create --local --test const-cifar10.yaml .`

## Commentary

List hparams work just fine normally, but the function `generate_random_hparam_values` (accidentally?) excluded them.

## Checklist


coreystatendet:
@rb-determined-ai Do minor fixes like this need a Release Note?

rb-determined-ai (approved):

rb-determined-ai:
> @rb-determined-ai Do minor fixes like this need a Release Note?

Not if a user didn't bring it up.  If a single user brought it up, I would personally just DM that user instead of writing a release note.

PR #3623, opened 2022-02-15
fix: bypass backoff if download_gcs_blob_with_backoff called with None

coreystatendet:
## Description

Adds a type check to download_gcs_blob_with_backoff.  This fixes an issue where the following code will hang for ~15 minutes if `filepath` is invalid:

```
blob = bucket.get_blob(filepath)
content = download_gcs_blob_with_backoff(blob)
```

This is because `get_blob` returns `None` if filepath is invalid.

## Test Plan

Check that `download_gcs_blob_with_backoff` raises an exception when passed `None` and works normally otherwise.  The following snippets should work without credentials:

```
from determined.utils import download_gcs_blob_with_backoff
# Raises an exception immediately.
download_gcs_blob_with_backoff(None)

# Downloads content successfully.
from google.cloud.storage import Client
client = Client()
bucket = client.get_bucket('determined-ai-coco-dataset')
blob = bucket.get_blob('annotations/instances_minival2014.json')
download_gcs_blob_with_backoff(blob)
```

## Checklist


harness/determined/util.py
```diff
@@ -20,6 +20,11 @@
 
 @util.preserve_random_state
 def download_gcs_blob_with_backoff(blob: Any, n_retries: int = 32, max_backoff: int = 32) -> Any:
+    from google.cloud.storage import Blob
```
rb-determined-ai, line 23: We try to stick to the [Google Python Style Guide for imports](https://google.github.io/styleguide/pyguide.html#22-imports) in all new code.  Basically: import the module for a symbol rather than the symbol itself.

So this should be:
```python
from google.cloud import storage
...
assert isinstance(blob, storage.Blob), ...
```

harness/determined/util.py
```diff
@@ -20,6 +20,11 @@
...
+
+    assert isinstance(
+        blob, Blob
+    ), f"Called download_gcs_blob_with_backoff with object of type {type(blob)}"
```
rb-determined-ai, line 27: nit: can you use `type(blob).__name__`?

rb-determined-ai (approved):

harness/determined/util.py
coreystatendet: Can you say more about why that's preferable?

harness/determined/util.py
```diff
@@ -20,6 +20,11 @@
 
 @util.preserve_random_state
 def download_gcs_blob_with_backoff(blob: Any, n_retries: int = 32, max_backoff: int = 32) -> Any:
+    from google.cloud import storage
+
+    assert isinstance(
```
rb-determined-ai, line 25: Ok, sorry I didn't think about this on the first time through.

Two things.  First: let's use an `if` here instead of an `assert`.  Python (inexplicably) has a performance mode where asserts don't run.  Since ultimately this is validation results from outside the system (user input of a filename, or gcp server's response) let's avoid `assert` in this case.

Second, while this will fix the 15-minute hang, which is clearly an improvement, can we catch the error somewhere more useful?  I'm thinking if the filepath provided was invalid, we should detect the failure right after `bucket.get_blob()` so that we can return an error with the filename, rather than saying "Called download_gcs_blob_with_backoff with object of type None", which would be rather opaque for the user.

Also, I'm not opposed to leaving the None check here too, as a safeguard.

coreystatendet: Was not familiar with no-asserts mode!  Will switch to an if -> raise.

Agree on catching the error somewhere more useful as well.  All users of `bucket.get_blob` should be checking this -- but as I ran into this on two separate occasions (one in my code, one in someone else's) it's clear that often doesn't happen.  I think users generally expect that get_blob will raise an exception if it fails, or return an object that will later cause an exception to be raised.  I also have it on my TODO to fix both locations I'm aware of -- this was just a quicker fix to test.

rb-determined-ai (approved):
oh, we only use this in example code?  Wow, we need to fix that... it's not good for our library to have random utilities to 3rd party libraries.

In the mean time, this is a decent fix.

PR #3626, opened 2022-02-17
fix: Model/Experiment API and error codes [DET-6711]

mapmeld:
## Description

This is a WIP of returning 400 error codes when the server or DB reports an input-related error, instead of a 500.  Right now it is returning a 400 when we discussed using a 409.  If this code makes sense then I can generalize the error-capture code to other PG functions and Model/Experiment API endpoints.

## Test Plan

- When creating an invalid model name (such as "a/b") or changing the name (to an invalid or duplicate name), the server should return the error with an InvalidArgument code and not a 500
- When the Postgres Database returns an error code which we recognize (such as uniqueness or bad param), that code should be turned to a message and passed up with an InvalidArgument code. I'm not sure an example how to test this (int/string API error?) it needs to be a constraint where we don't already block it in the protocol or server code

## Checklist

### DIFF
--- a/master/internal/api_model.go
+++ b/master/internal/api_model.go
@@ -131,7 +131,8 @@ func (a *apiServer) clearModelName(ctx context.Context, modelName string) error 
 func (a *apiServer) PostModel(
 	ctx context.Context, req *apiv1.PostModelRequest) (*apiv1.PostModelResponse, error) {
 	if err := a.clearModelName(ctx, req.Name); err != nil {
-		return nil, errors.Wrap(err, "error setting model name")
+		return nil, status.Errorf(codes.AlreadyExists,
+			"Error renaming model: %s", err.Error())
 	}
 
 	b, err := protojson.Marshal(req.Metadata)
@@ -173,7 +174,8 @@ func (a *apiServer) PatchModel(
 		log.Infof("model (%d) name changing from \"%s\" to \"%s\"",
 			currModel.Id, currModel.Name, req.Model.Name.Value)
 		if err = a.clearModelName(ctx, req.Model.Name.Value); err != nil {
-			return nil, errors.Wrap(err, "error renaming model")
+			return nil, status.Errorf(codes.AlreadyExists,
+				"Error renaming model: %s", err.Error())
 		}
 		madeChanges = true
 		currModel.Name = req.Model.Name.Value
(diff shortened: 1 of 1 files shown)


master/internal/api_model.go
```diff
@@ -131,7 +131,8 @@ func (a *apiServer) clearModelName(ctx context.Context, modelName string) error
 func (a *apiServer) PostModel(
 	ctx context.Context, req *apiv1.PostModelRequest) (*apiv1.PostModelResponse, error) {
 	if err := a.clearModelName(ctx, req.Name); err != nil {
-		return nil, errors.Wrap(err, "error setting model name")
+		return nil, status.Errorf(codes.AlreadyExists,
```
mpkouznetsov, line 134: I suggest `codes.InvalidArgument` here so it is translated to 400 Bad request.

mpkouznetsov (commented):
Looks good; one suggestion.

master/internal/api_model.go
```diff
@@ -173,7 +174,8 @@ func (a *apiServer) PatchModel(
 		log.Infof("model (%d) name changing from \"%s\" to \"%s\"",
 			currModel.Id, currModel.Name, req.Model.Name.Value)
 		if err = a.clearModelName(ctx, req.Model.Name.Value); err != nil {
-			return nil, errors.Wrap(err, "error renaming model")
+			return nil, status.Errorf(codes.AlreadyExists,
```
mpkouznetsov, line 177: I guess here it is the same case as above. `codes.InvalidArgument` is more appropriate

master/internal/api_model.go
```diff
@@ -105,33 +105,33 @@ func (a *apiServer) GetModelLabels(
...
+		return status.Errorf(codes.InvalidArgument, "model names cannot be only numbers")
 	}
 	getResp, err := a.GetModels(ctx,
 		&apiv1.GetModelsRequest{Name: modelName, NameCaseInsensitive: true})
 	if err != nil {
 		return err
 	}
 	if len(getResp.Models) > 0 {
-		return errors.Errorf("avoid using model names with case-sensitive similar names")
+		return status.Errorf(codes.AlreadyExists, "avoid names similar too other models (case-insensitive)")
```
mpkouznetsov, line 126:
```suggestion
		return status.Errorf(codes.AlreadyExists, "avoid names equal to other models (case-insensitive)")
```

mpkouznetsov (approved):
LGTM, except the wording/spelling of the error message.

PR #3629, opened 2022-02-18
chore: CLI resource-setting replaced with new API routes [DET-6487]

mapmeld:
## Description

Move resource allocation settings (max_slots, weight, priority) from old core_experiment to the new REST API. We only call this from the CLI, so this updates that as well.

When combined with #3539 (which removes State) then core_experiment PATCH will only be used for gc-policy

## Test Plan

Confirm CLI `det experiment set weight id 1` still changes values of weight in `/api/v1/experiments/id`
Same for `max-slots` and `priority`.

## Checklist

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -603,19 +603,34 @@ def remove_label(args: Namespace) -> None:
 
 @authentication.required
 def set_max_slots(args: Namespace) -> None:
-    patch_experiment(args, "change `max_slots` of", {"resources": {"max_slots": args.max_slots}})
+    session = setup_session(args)
+    r = bindings.get_GetExperiment(session, experimentId=args.experiment_id).experiment
+    exp = r.to_json()
+    exp["maxSlots"] = args.max_slots
+    exp_patch = bindings.v1PatchExperiment.from_json(exp)
+    bindings.patch_PatchExperiment(session, body=exp_patch, experiment_id=args.experiment_id)
     print("Set `max_slots` of experiment {} to {}".format(args.experiment_id, args.max_slots))
 
 
 @authentication.required
 def set_weight(args: Namespace) -> None:
-    patch_experiment(args, "change `weight` of", {"resources": {"weight": args.weight}})
+    session = setup_session(args)
+    r = bindings.get_GetExperiment(session, experimentId=args.experiment_id).experiment
+    exp = r.to_json()
+    exp["weight"] = args.weight
+    exp_patch = bindings.v1PatchExperiment.from_json(exp)
+    bindings.patch_PatchExperiment(session, body=exp_patch, experiment_id=args.experiment_id)
     print("Set `weight` of experiment {} to {}".format(args.experiment_id, args.weight))
 
 
 @authentication.required
 def set_priority(args: Namespace) -> None:
-    patch_experiment(args, "change `priority` of", {"resources": {"priority": args.priority}})
+    session = setup_session(args)
+    r = bindings.get_GetExperiment(session, experimentId=args.experiment_id).experiment
+    exp = r.to_json()
+    exp["priority"] = args.priority
+    exp_patch = bindings.v1PatchExperiment.from_json(exp)
+    bindings.patch_PatchExperiment(session, body=exp_patch, experiment_id=args.experiment_id)
     print("Set `priority` of experiment {} to {}".format(args.experiment_id, args.priority))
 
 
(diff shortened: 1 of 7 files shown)


gh-determined-ai:
@mapmeld  We're trying to improve our release note descriptions so I've provided a template:
[](https://determinedai.atlassian.net/wiki/spaces/DOC/pages/1101594663/Documentation+Process#Release-Note-Template)
Please, follow the template format and include the applicable information, depending on the scope and scale of the change. Let me know if I can help. Since this is a new effort, it would be helpful to get your feedback on this release note process.

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -106,6 +106,12 @@ message PatchExperiment {
   google.protobuf.StringValue name = 4;
   // The experiment notes.
   google.protobuf.StringValue notes = 5;
+  // The maximum resources in slots.
+  google.protobuf.Int32Value max_slots = 6;
```
stoksc, line 110: this using nullable fields for patches thing is mostly nice, but what if we need to set a field to nil? why did we move to this from field masks (i may've just forgotten)?

master/internal/core_experiment.go
```diff
@@ -250,21 +231,6 @@ func (m *Master) patchExperiment(c echo.Context) (interface{}, error) {
 		m.system.TellAt(actor.Addr("experiments", args.ExperimentID), *patch.State)
 	}
 
-	if patch.Resources != nil {
```
stoksc, line 253: are there not an places in the CLI that use this stuff?

harness/determined/cli/experiment.py
```diff
@@ -570,19 +570,34 @@ def remove_label(args: Namespace) -> None:
 
 @authentication.required
 def set_max_slots(args: Namespace) -> None:
-    patch_experiment(args, "change `max_slots` of", {"resources": {"max_slots": args.max_slots}})
+    session = setup_session(args)
+    r = bindings.get_GetExperiment(session, experimentId=args.experiment_id).experiment
+    exp = r.to_json()
+    exp["maxSlots"] = args.max_slots
+    exp_patch = bindings.v1PatchExperiment.from_json(exp)
+    bindings.patch_PatchExperiment(session, body=exp_patch, experiment_id=args.experiment_id)
```
stoksc, line 578: this looks like me like we're using a patch to do a put. why would we bind the entire exp json rather than just constructing a patch experiment object? it seems this would pretty easily if they ever fell out of sync (patchExperiment vs experiment) by accident?

master/internal/api_experiment.go
```diff
@@ -580,6 +582,35 @@ func (a *apiServer) PatchExperiment(
...
+				sproto.SetGroupMaxSlots{MaxSlots: ptrs.IntPtr(int(req.Experiment.MaxSlots.Value))})
+		}
+		if req.Experiment.Weight != nil {
+			resources.SetWeight(float64(req.Experiment.Weight.Value))
+			a.m.system.TellAt(actor.Addr("experiments", req.Experiment.Id),
+				job.SetGroupWeight{Weight: float64(req.Experiment.Weight.Value)})
+		}
+		if req.Experiment.Priority != nil {
+			resources.SetPriority(ptrs.IntPtr(int(req.Experiment.Priority.Value)))
+			a.m.system.TellAt(actor.Addr("experiments", req.Experiment.Id),
```
stoksc, line 605: can we make a helper func to get this actor.Addr and try to make everything use it? it would be really useful when if we try to refactor to not accidentally break stuff.

master/internal/api_experiment.go
```diff
@@ -580,6 +582,35 @@ func (a *apiServer) PatchExperiment(
...
+				job.SetGroupPriority{Priority: int(req.Experiment.Priority.Value)})
+		}
+		dbExp.Config.SetResources(resources)
```
stoksc, line 608: should we do the save first, so we're more confident mem and db are in sync?

gh-determined-ai (changes requested):
Management would like to see more descriptive information in the release note regarding the change. See the release note template for some ideas of what to put in a release note: https://determinedai.atlassian.net/wiki/spaces/DOC/pages/1101594663/Documentation+Process#Release-Note-Template

harness/determined/cli/experiment.py
mapmeld: I can make a smaller PatchExperiment object here

mapmeld:
Closing... issues with this approach: doesn't add new functionality, would be difficult to set a resource value to null, needs connections to stuff outside of `api_experiment.go`

PR #3630, opened 2022-02-18
feat: non-training-specific checkpoints

rb-determined-ai:
## Description

Checkpoints in Determined were originally designed exclusively for
training.  We would like in the future to support checkpointing for all
tasks.  To get there, we need to change what data we keep about
checkpoints.  This change focuses on migrating existing
training-specific data into checkpoint metadata, modifying how
checkpoints are uploaded so that the Core API checkpointing does not
feel training-specific, and updating the WebUI to reflect the new
schema.

## Commentary (optional)

This upstream feature branch will include backend work, webui client-side work, and python client-side work before it is complete.

### DIFF
--- a/master/internal/api_trials.go
+++ b/master/internal/api_trials.go
@@ -26,7 +26,6 @@ import (
 	"github.com/determined-ai/determined/master/pkg/protoutils"
 	"github.com/determined-ai/determined/master/pkg/searcher"
 	"github.com/determined-ai/determined/proto/pkg/apiv1"
-	"github.com/determined-ai/determined/proto/pkg/checkpointv1"
 	"github.com/determined-ai/determined/proto/pkg/experimentv1"
 	"github.com/determined-ai/determined/proto/pkg/trialv1"
 )
@@ -295,64 +294,6 @@ func (a *apiServer) TrialLogsFields(
 			Sources:      setString(append(r1.Sources, r2.Sources...)...),
 		})
 	})
-}
-
-func (a *apiServer) GetTrialCheckpoints(
-	_ context.Context, req *apiv1.GetTrialCheckpointsRequest,
-) (*apiv1.GetTrialCheckpointsResponse, error) {
-	switch exists, err := a.m.db.CheckTrialExists(int(req.Id)); {
-	case err != nil:
-		return nil, err
-	case !exists:
-		return nil, status.Error(codes.NotFound, "trial not found")
-	}
-
-	resp := &apiv1.GetTrialCheckpointsResponse{}
-	resp.Checkpoints = []*checkpointv1.Checkpoint{}
-
-	switch err := a.m.db.QueryProto("get_checkpoints_for_trial", &resp.Checkpoints, req.Id); {
-	case err == db.ErrNotFound:
-		return nil, status.Errorf(
-			codes.NotFound, "no checkpoints found for trial %d", req.Id)
-	case err != nil:
-		return nil,
-			errors.Wrapf(err, "error fetching checkpoints for trial %d from database", req.Id)
-	}
-
-	a.filter(&resp.Checkpoints, func(i int) bool {
-		v := resp.Checkpoints[i]
-
-		found := false
-		for _, state := range req.States {
-			if state == v.State {
-				found = true
-				break
-			}
-		}
-
-		if len(req.States) != 0 && !found {
-			return false
-		}
-
-		found = false
-		for _, state := range req.ValidationStates {
-			if state == v.ValidationState {
-				found = true
-				break
-			}
-		}
-
-		if len(req.ValidationStates) != 0 && !found {
-			return false
-		}
-
-		return true
-	})
-
-	a.sort(
-		resp.Checkpoints, req.OrderBy, req.SortBy, apiv1.GetTrialCheckpointsRequest_SORT_BY_BATCH_NUMBER)
-
-	return resp, a.paginate(&resp.Pagination, &resp.Checkpoints, req.Offset, req.Limit)
 }
 
 func (a *apiServer) KillTrial(
--- a/master/internal/ckpts/api.go
+++ b/master/internal/ckpts/api.go
@@ -1,0 +1,103 @@
+package ckpts
+
+import (
+	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/determined-ai/determined/master/pkg/protoutils"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
+	"github.com/determined-ai/determined/proto/pkg/checkpointv1"
+)
+
+type CkptsAPI struct{}
+
+func (s *CkptsAPI) GetCheckpoint(
+	ctx context.Context, req *apiv1.GetCheckpointRequest) (*apiv1.GetCheckpointResponse, error) {
+	ckpt, err := ByUUIDExpanded(ctx, req.CheckpointUuid)
+	if err != nil {
+		return nil, err
+	}
+
+	pc := protoutils.ProtoConverter{}
+	protoCkpt := ckpt.ToProto(&pc)
+	resp := &apiv1.GetCheckpointResponse{Checkpoint: &protoCkpt}
+	return resp, pc.Error()
+}
+
+func (a *CkptsAPI) GetTrialCheckpoints(
+	ctx context.Context, req *apiv1.GetTrialCheckpointsRequest,
+) (*apiv1.GetTrialCheckpointsResponse, error) {
+	// // XXX: check trial exists
+	// switch exists, err := a.m.db.CheckTrialExists(int(req.Id)); {
+	// case err != nil:
+	// 	return nil, err
+	// case !exists:
+	// 	return nil, status.Error(codes.NotFound, "trial not found")
+	// }
+
+	var validationStates []string
+	for _, vstate := range req.ValidationStates {
+		switch vstate {
+		case checkpointv1.State_STATE_ACTIVE:
+			validationStates = append(validationStates, "'ACTIVE'")
+		case checkpointv1.State_STATE_COMPLETED:
+			validationStates = append(validationStates, "'COMPLETED'")
+		case checkpointv1.State_STATE_ERROR:
+			validationStates = append(validationStates, "'ERROR'")
+		case checkpointv1.State_STATE_DELETED:
+			validationStates = append(validationStates, "'DELETED'")
+		default:
+			// XXX: throw an error to the user
+		}
+	}
+
+	// choose an ordering
+	asc := req.OrderBy != apiv1.OrderBy_ORDER_BY_DESC
+	var orderBy string
+	switch req.SortBy {
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_UNSPECIFIED:
+		// noop, use the default
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_BATCH_NUMBER:
+		// noop, this is the default
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_START_TIME:
+		// noop, this field doesn't exist anymore
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_END_TIME:
+		orderBy = "report_time"
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_VALIDATION_STATE:
+		orderBy = "validation_metrics != 'null'::jsonb"
+	case apiv1.GetTrialCheckpointsRequest_SORT_BY_STATE:
+		orderBy = "state"
+	}
+
+	ckpts, total, err := GetTrialCheckpointsExpanded(
+		ctx, int(req.Id), validationStates, orderBy, asc, int(req.Limit), int(req.Offset),
+	)
+	if err != nil {
+		return nil, err
+	}
+	if len(ckpts) == 0 {
+		// XXX: Why not return an empty array?
+		return nil, status.Errorf(codes.NotFound, "no checkpoints found for trial %d", req.Id)
+	}
+
+	pc := protoutils.ProtoConverter{}
+	resp := &apiv1.GetTrialCheckpointsResponse{}
+	resp.Checkpoints = make([]*checkpointv1.Checkpoint, len(ckpts))
+	for i, ckpt := range ckpts {
+		protoCkpt := ckpt.ToProto(&pc)
+		// XXX taking address of protoCkpt looks buggy to me
+		resp.Checkpoints[i] = &protoCkpt
+	}
+	resp.Pagination = &apiv1.Pagination{
+		Offset: req.Offset,
+		Limit:  req.Limit,
+		// XXX  Why have both offset and start_idx?
+		StartIndex: req.Offset,
+		EndIndex:   pc.ToInt32(int(req.Offset) + len(ckpts)),
+		Total:      pc.ToInt32(total),
+	}
+
+	return resp, pc.Error()
+}
--- a/master/internal/models/models.go
+++ b/master/internal/models/models.go
@@ -1,0 +1,53 @@
+package models
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/internal/db"
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+func ByID(ctx context.Context, id int) (model.Model, error) {
+	var out model.Model
+	err := db.Bun().NewSelect().
+		Model(&out).
+		// XXX: a better way to do a count query as a subquery?
+		ColumnExpr(`(select count (*) from model_versions
+					where model_versions.model_id = models.id) as num_versions`).
+		Where("id = ?", id).
+		Scan(ctx)
+	if err != nil {
+		return model.Model{}, errors.Wrapf(err, "error selecting model(%v)", id)
+	}
+	return out, err
+}
+
+func ByName(ctx context.Context, name string) (model.Model, error) {
+	var out model.Model
+	err := db.Bun().NewSelect().
+		Model(&out).
+		// XXX: a better way to do a count query as a subquery?
+		ColumnExpr(`(select count (*) from model_versions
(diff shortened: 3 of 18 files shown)


master/internal/db/bun.go
```diff
@@ -0,0 +1,115 @@
...
+func (db *RootBunDB) RunInTx(ctx context.Context, fn func(context.Context, BunDB) error) error {
+	wrapperFn := func(ctx_ context.Context, tx bun.Tx) error {
+		return fn(ctx_, dbImpl{tx})
+	}
+	return db.DB.RunInTx(ctx, nil, wrapperFn)
+}
+
+// QueryExt is a helper struct to define custom .Where, .Order, and .Limit without a ton of inlined
+// functions.
+type QueryExt struct {
```
ioga, line 58: I am not super excited about this solution. asymptotically it's a custom proxy that'll need to handle the entirety of bun options.

Couple thoughts/options:
1. Functions like `CheckpointsList` that use this doesn't add much before/after the query: basically only `db.NewSelect().Model(...)` call before and error handling after. is that worth the extra complexity? both can be done inline in the "view" methods.
2. We could have `filterFn` approach similar to bun's `RunInTx` interface. i.e. 
signature
```
func CheckpointsList(ctx context.Context, filterFn func(bun.SelectQuery) (bun.SelectQuery, error))
```
and then at call-site
```
# req is grpc request thing
ckpts, err := CheckpointsList(ctx, func(ctx context.Context, query bun.SelectQuery) {
  	query = query.Where("trial_id = ?", req.Id)
    ...
    return query, nil
})

Comment (rb-determined-ai) [reply master/internal/db/bun.go:58]: I experimented with the `filterFn` but I really disliked it, because there are cases where you hit errors _creating_ the filter (not running the filter), but when when you do `ckpts, err := ckpts.List(ctx, func(*bun.SelectQuery)(*bun.SelectQuery, error))`, you don't know if your `err` came from user error (defining a nonsensical error) or from the database call itself.  This would be the difference between an invalid request error and a 500 error, so the calling code does need to distinguish.

I toyed with nesting closures, but it got pretty unreadable pretty quick.

File: master/internal/api_trials.go [hunk 2]
```diff
@@ -308,52 +310,95 @@ func (a *apiServer) GetTrialCheckpoints(
...
-		if len(req.States) != 0 && !found {
-			return false
+		// apply pagination
+		ext = ext.Limit(int(req.Limit))
+		ext = ext.Offset(int(req.Offset))
+
+		ckpts, err := ckpts.ListTx(ctx, tx, ext)
+		if err != nil {
+			return errors.Wrapf(err, "error fetching checkpoints for trial %d from database", req.Id)
+		}
```
rb-determined-ai, line 377: @ioga this is the code in question, where it's really nice to have this QueryExt object.

If we decide that separate CountTx and ListTx calls are not acceptable anyway, then we should just rm -rf the QueryExt.

ioga: if we were to follow 1), i.e. simply inline it, then we'd be building a `bun.SelectQuery` from `db.NewSelect()` instead of `QueryExt`, call `query.Count(ctx)` to get the count, and then proceed with adding `Limit` and `Offset` to the same query.

ioga: on a separate note, I think this transaction block doesn't do anything. You get "read committed" level by default, and these selects can get different committed data. 
you've got to set transaction to repeatable read if you want to get count & list results matching.

rb-determined-ai: > I think this transaction block doesn't do anything
Interesting.  So I'm fine with any of the following, in order I would prefer them:
- don't bother with transactions on pagination data, it seems racy anyway.  Pagination data will change from call to call so, eh.
- increase transaction level
- switch to one megaquery, as with get_experiments.  I see this as an optimization that complicates the code to require much more SQL skill (more than I have at least, and I'm having to write it).  IMO that's worth it only when we've identified this code as a bottleneck, otherwise I think it's a premature optimization.

master/internal/db/bun.go
stoksc: >  This would be the difference between an invalid request error and a 500 error, so the calling code does need to distinguish.

At what point should we be building an invalid query because of user input?

stoksc: If for some reason we are, then we could define the signature as like
```
func CheckpointsList(ctx context.Context, filterFn func(bun.SelectQuery) (bun.SelectQuery, errInvalidQuery))
```
and use `errors.Is`.

rb-determined-ai:
Ok, @stoksc and @ioga, I would like to highlight three patterns I would like feedback on before going any further:

1. basic queries in `internal/ckpts/ckpts.go` and `internal/models/models.go`.  I don't think these are terribly controversial, but getting settled on these is probably enough to get @mapmeld taken care of.
2. pagination queries like `internal/api_trials.go::GetTrialCheckpoints()`.  See thread above to continue discussion.
3. compound queries, like `internal/api_models.go::GetModelVersion()`.  I found that `static/srv/get_model_versions.sql` was both insanely complicated and silly in its implementation; it's secretly just three separate, non-interacting queries (get model, get checkpoint, get model version) that are running as one query and using json to combine their output into a REST-API-compatible form.  Also, one of those queries (get model) is run before `get_model_versions.sql`, then the model_id is fed back into `get_model_versions.sql` so the model can be fetched again.  Instead, I wrote it as three completely separate queries in a transaction (probably has the same isolation level issue @ioga mentioned already).  But it is _way_ simpler like this.

stoksc:
totally pointless comment: I think `internal/ckpts/ckpts.go` should be `internal/db/ckpts/ckpts.go`

master/internal/api_trials.go
ioga: I'd vote for not bothering here.

ioga:
>     3. compound queries, like `internal/api_models.go::GetModelVersion()`.  I found that `static/srv/get_model_versions.sql` was both insanely complicated and silly in its implementation; it's secretly just three separate, non-interacting queries (get model, get checkpoint, get model version) that are running as one query and using json to combine their output into a REST-API-compatible form.  Also, one of those queries (get model) is run before `get_model_versions.sql`, then the model_id is fed back into `get_model_versions.sql` so the model can be fetched again.  Instead, I wrote it as three completely separate queries in a transaction (probably has the same isolation level issue @ioga mentioned already).  But it is _way_ simpler like this.

I am on board. Let's have these as separate queries, and not bother with transactions and their isolation levels. The fallout from a possible "wrong" data here seems minimal. We only try to optimize only the expensive ones.

stoksc:
personally:
1. basic queries :heavy_check_mark:  i think what's here looks great
2. for paginated/parameterized queries i much prefer using bun directly and accepting closures to extend queries. if we want different behavior on different errors, just use different error types.
3. for compound queries, i think we should have a compound/composite layer, like a service layer, that encapsulates that logic. user.Service is basically the exact pattern i think we should follow (maybe start to stash them all in `internal/service/thing`). so 3 separate queries calling into the `internal/db/things_you_need` layers is great to me.

master/internal/api_trials.go
rb-determined-ai: @stoksc So you like the closures and you don't like QueryExt (totally fine), but what about the megaquery vs two separate queries?

rb-determined-ai:
> for compound queries, i think we should have a compound/composite layer, like a service layer, that encapsulates that logic. user.Service is basically the exact pattern i think we should follow (maybe start to stash them all in internal/service/thing). so 3 separate queries calling into the internal/db/things_you_need layers is great to me.

I'm not sure I understand.  user.Service looks to me like a series of REST API endpoints (`patchUser`, `postLogin`, etc).  I don't understand the connection to multiple queries vs a megaquery.

stoksc:
> I'm not sure I understand. user.Service looks to me like a series of REST API endpoints (patchUser, postLogin, etc). I don't understand the connection to multiple queries vs a megaquery.

model api need to return the results of many queries. write a model service, and put the combining logic in there so it can be reused without calling `a.GetModel()` from inside `a.UpdateModel` later.

master/internal/api_trials.go
stoksc: two separate queries is fine.

stoksc: better, even.

rb-determined-ai:
> model api need to return the results of many queries. write a model service, and put the combining logic in there so it can be reused without calling a.GetModel() from inside a.UpdateModel later

I definitely agree with avoiding calling `a.GetModel()` from `a.UpdateModel()`, really under any circumstance.  But what you call a "service layer"... do you just mean "some helper functions"?  Or do you mean "an actor that you pass messages to and get messages back from"?

stoksc:
i mean a struct called `ModelService` that has helper functions that encapsulate shared logic around models. Stuff like "validate this model before I insert it" or "composite these 3 queries into some other object" rather than that being duplicated or forgotten at every call site, or every set of endpoints using a different pattern for how to share/reuse this type of logic.

stoksc:
in determined, a lot right now, this layer is just done by an actor (think agents actor, jobs actor now, etc)

rb-determined-ai:
> i mean a struct called ModelService that has helper functions that encapsulate shared logic around models

But isn't this what the `internal/db/models` module would be for?  What would it need to be a stateful object?

stoksc:
> But isn't this what the internal/db/models module would be for?

i thought internal/db/models was just for accessing the db, service layer would be for composite queries. if ckpts is in internal/db/ckpts and models is in internal/db/models, the model service that composites them would just be in internal/service/models so that it can call db/ckpts+models without ever hitting circular imports.

> What would it need to be a stateful object?

it doesn't have to be. it's just helper methods in a special place. (previously it may've just had a config struct or a db, but the db is a singleton at least)

ioga:
We are inching towards a model-view-controller here. "Service layer" sounds like a controller.

Similarly to the shortcuts code, I think the code should be structured not by layer but by "entity". i.e. `internal/workflows/` directory will have: 
 - `api.go` for "view" part: serialization between protobufs and internal data types.
 - `models.go` for "model" part: database-related code, basic business logic such as helpers or methods on model instances.
 - `controller.go` for "controller": composite business logic, stateful objects, watchers, polling, heartbeats, etc. in practice, simple APIs don't need that and it can be folded either into models or into view.

stoksc:
this is exactly what i'm pushing for, just pushed around some (sorry, now it is obvious i used to write a lot of vanilla .net core apis, but the pattern works "just fine"). this pattern is fairly straightforward. the users code is fairly close to this already https://github.com/determined-ai/determined/tree/master/master/internal/user

stoksc:
@ioga the only issue is defining the apis in a different package than the apiServer is basically impossible.

rb-determined-ai:
> I think the code should be structured not by layer but by "entity". i.e. internal/workflows/ directory will have:

This doesn't seem like it would solve the circular dependency problem in the way that @stoksc's suggestion would.  Where would compound queries go?  Or would we still have `internal/db/*` in addition to `internal/workflows` and `internal/ckpts`, etc? 

> @ioga the only issue is defining the apis in a different package than the apiServer is basically impossible.

I don't see why this wouldn't work:

```go
type struct apiServer {
    models.Server
    workflows.Server
    users.Server
}

**Comment by stoksc** [comment]
obviously i dont read well, let me come back to this when i have more than 5m straight (~30m)

**Comment by ioga** [comment]
> This doesn't seem like it would solve the circular dependency problem in the way that @stoksc's suggestion would. Where would compound queries go? Or would we still have `internal/db/*` in addition to `internal/workflows` and `internal/ckpts`, etc?

The "entities" should form a dependency tree. 3 options:
- workflows depend on checkpoints and import them.
- checkpoints depend on workflows and import them.
- if they're tightly intertwined and depend on each other, and separate compound entity "checkpointed_workflows" — it would probably not have `models.go`, only api and controller.

**Comment by rb-determined-ai** [comment]
So then what about leaving `api.go` in the `internal/workflows` and use:

```go
type struct apiServer {
    m *Master
    workflows.Server
}
```
?

I tried using composition like this to define the apiServer and it still compiles.

stoksc:
are we all in agreement now? seems like it?

master/internal/models/models.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func VersionTx(
+	ctx context.Context, idb bun.IDB, modelId int, version int,
+) (model.RBModelVersion, error) {
+	var out model.RBModelVersion
+	err := idb.NewSelect().
+		Model(&out).
+		Where("modelId = ?", modelId).
+		Where("version = ?", version).
```
mapmeld, line 58: This is different from what we currently do in WebUI. In our current setup `/models/Tester/versions/41/overview` , links to `model_versions.id = 41`, not the auto-incremented `model_versions.version`.

master/pkg/model/model.go
```diff
@@ -32,3 +39,116 @@ type ModelVersion struct {
...
+	return modelv1.ModelVersion{
+		Model: &m,
+		Checkpoint: &c,
+		Version: pc.ToInt32(v.Version),
+		CreationTime: pc.ToTimestamp(v.CreationTime),
+		// XXX: this seems very confusing to me, that we expose a meaningless internal ID.
+		Id: pc.ToInt32(v.ID),
+		// XXX: why on earth does ModelVersion have it's own name field?? Shouldn't we use the
+		// model's name field?
+		Name: model.Name, // Name: v.Name,
```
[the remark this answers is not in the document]
rb-determined-ai: Does the version have an editable name in the webui currently?

master/internal/models/models.go
rb-determined-ai: good catch.  I think I messed up my interpretation of the original sql.

rb-determined-ai: So it should be:
```sql
	err := idb.NewSelect().
		Model(&out).
		Where("model_id = ?", modelId).
		Where("id = ?", version).
		Scan(ctx)
```
right?

rb-determined-ai: Wait... `model_version.version` is not auto-incremented, but `id` is:
```determined=# \d model_versions
                                          Table "public.model_versions"
      Column       |           Type           | Collation | Nullable |                  Default
-------------------+--------------------------+-----------+----------+--------------------------------------------
 version           | integer                  |           | not null |
 checkpoint_uuid   | uuid                     |           | not null |
 creation_time     | timestamp with time zone |           | not null |
 last_updated_time | timestamp with time zone |           |          |
 metadata          | jsonb                    |           |          |
 model_id          | integer                  |           | not null |
 id                | integer                  |           | not null | nextval('model_versions_id_seq'::regclass)
 name              | text                     |           |          |
 comment           | text                     |           |          |
 user_id           | integer                  |           |          |
 labels            | text[]                   |           |          |
 notes             | text                     |           |          |
```

mapmeld: since ID is unique I think it can safely be just `Where("id = ?", version).`

rb-determined-ai: The old code looks buggy to me.

The go code was:
```go
QueryProto("get_model_version", resp.ModelVersion, parentModel.Model.Id, req.ModelVersion)
```

The sql was:
```sql
WITH mv AS (
  SELECT version, checkpoint_uuid, model_versions.id, creation_time, name, comment, metadata, labels, notes, username, last_updated_time
    FROM model_versions
    LEFT JOIN users ON users.id = model_versions.user_id
    WHERE model_id = $1 AND model_versions.id = $2
),
```

It looks like we were previously matching `model_versions.id` against `req.ModelVersion`.

proto/src/determined/checkpoint/v1/checkpoint.proto
```diff
@@ -30,45 +22,49 @@ enum State {
...
+  // The configuration of the experiment that created this checkpoint.
+  google.protobuf.Struct experiment_config = 3;
+  // Hyperparameter values for the trial that created this checkpoint.
+  google.protobuf.Struct hparams = 4;
+  // Training metrics reported at the same latest_batch as the checkpoint.
+  google.protobuf.Struct training_metrics = 5;
+  // Validation metrics reported at the same latest_batch as the checkpoint.
+  google.protobuf.Struct validation_metrics = 6;
+  // TODO: is this actually a good idea?
+  google.protobuf.DoubleValue searcher_metric = 17;
```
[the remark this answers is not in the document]
rb-determined-ai: > On a related note, does ProtoConverter handle these nullable types (google.protobuf.DoubleValue, google.protobuf.StringValue) well?

If there's a type it doesn't handle yet... just add it.  Really the ProtoConverter just makes it easier to copy from normal go types (`int`, `time.Time`, or `map[string]interface{}`) to protobuf equivalents without a zillion error checks and intermediate values.

master/internal/models/models.go
mapmeld: the increment is being done in `insert_model_version` but maybe should be defined by the table
`(SELECT COALESCE(max(version), 0) + 1 FROM model_versions WHERE model_id = $1)`

I'd say `req.ModelVersion` is named poorly rather than it being a bug, because in the Model page we link to ModelVersions by the id field: https://github.com/determined-ai/determined/blob/master/webui/react/src/components/Table.tsx#L167-L182 - it's a long time ago but I think I had a convo about which one we wanted to use

rb-determined-ai: wait... isn't the `id` field already an autoincrement field?  I thought that's what the `nextval('model_versions_id_seq'::regclass)` thing meant.  Why are we doing a manual "auto" increment on insert?

rb-determined-ai: > I'd say req.ModelVersion is named poorly rather than it being a bug

ok.

master/internal/ckpts/api.go
```diff
@@ -0,0 +1,110 @@
...
+	// 	return nil, err
+	// case !exists:
+	// 	return nil, status.Error(codes.NotFound, "trial not found")
+	// }
+
+	var validationStates []string
+	for _, vstate := range req.ValidationStates {
+		switch vstate {
+		case checkpointv1.State_STATE_ACTIVE:
+			validationStates = append(validationStates, "'ACTIVE'")
```
rb-determined-ai, line 51: remove extra layers of quotes

rb-determined-ai:
closing in favor of #3859.

PR #3632, opened 2022-02-18
chore: remove extra logging for credHelpers [DET-6710]

varlogtim:
Reverts determined-ai/determined#3607

Will make another PR with this change which includes the debug flag required as I think it should be approved/denied in its full context.


ioga:
I thought it was just some extra logging? why does it need a flag?..

varlogtim:
> I thought it was just some extra logging? why does it need a flag?..

@rb-determined-ai stated that he felt this was a "message that seems a lot like log-vomit" and that it would cause users to think that something was wrong.

rb-determined-ai:
@varlogtim fwiw I like the other logs you added, maybe just delete the one log line that prints for 99% of users, rather than revert the whole change?

rb-determined-ai (approved):

PR #3634, opened 2022-02-18
feat: create PATCH users endpoint 

johnkim-det:
## Description

API endpoint that allows user fields to be edited. Currently only supports displayName.

## Test Plan

Accessible in Swagger docs at `/docs/rest-api/#/Users/PatchUser`.

Send requests to `PATCH users/{username}` and verify that 
- [ ] a request to modify the displayName returns a user object with updated displayName
- [ ] subsequent `GET users/{username}` requests show the updated displayName
- [ ] changes are not allowed for a username other than the logged in user's

## Checklist

### DIFF
--- a/master/pkg/model/user.go
+++ b/master/pkg/model/user.go
@@ -21,6 +21,7 @@ type User struct {
 	ID           UserID      `db:"id" json:"id"`
 	Username     string      `db:"username" json:"username"`
 	PasswordHash null.String `db:"password_hash" json:"-"`
+	DisplayName  string 		 `db:"display_name" json:"display_name"`
 	Admin        bool        `db:"admin" json:"admin"`
 	Active       bool        `db:"active" json:"active"`
 }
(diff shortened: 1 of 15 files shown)


master/pkg/model/user.go
```diff
@@ -21,6 +21,7 @@ type User struct {
 	ID           UserID      `db:"id" json:"id"`
 	Username     string      `db:"username" json:"username"`
 	PasswordHash null.String `db:"password_hash" json:"-"`
+	DisplayName  null.String `db:"display_name" json:"-"`
```
mapmeld, line 24: password_hash has the `json:"-"` thing just to make sure we don't share it. This should have display_name

proto/src/determined/user/v1/user.proto
```diff
@@ -20,6 +20,14 @@ message User {
   bool active = 4;
   // The user and group on the agent host machine.
   AgentUserGroup agent_user_group = 5;
+  string display_name = 6;
```
mapmeld, line 23: For `proto` to generate complete docs: this, PatchUser, and the PatchUser display_name should each have a comment above them (starting with // and ending with a period)

mapmeld (approved):

PR #3635, opened 2022-02-18
feat: break workload info from trial endpoint into a new endpoint [DET-6729]

dzhu:
## Description

When a trial has run many workloads, the response to the trial details
endpoint for it can become very large and unwieldy. Since we don't
always need the full set of workloads, we move those into a new endpoint
and just have some useful workload summary information in the original
one.

## Test Plan

- [x] run a no-op experiment, check that the total checkpoint size comes through
- [x] run a no-op experiment, check the workloads endpoint with various ordering/offset/limit options
- [ ] add an E2E test (or extend an existing one)

## Commentary

First two bullet points from the ticket done, third still to do.

The total checkpoint size is represented in the protobuf as uint64, since a trial with large or many checkpoints could easily overflow 32 bits. That ends up being translated into JSON as a string [because JavaScript can't natively represent all 64-bit integers](https://github.com/grpc-ecosystem/grpc-gateway/issues/219#issuecomment-251250029). The only alternative appears to be switching the type to float, which would feel weird to me.


ioga:
cc @stoksc because I have less context on this

proto/src/determined/api/v1/api.proto
```diff
@@ -446,6 +446,16 @@ service Determined {
       tags: [ "Trials", "Experiments" ]
     };
   }
+  // Get the list of workloads for a trial.
+  rpc GetTrialWorkloads(GetTrialWorkloadsRequest)
+      returns (GetTrialWorkloadsResponse) {
+    option (google.api.http) = {
+      get: "/api/v1/trial/{trial_id}/workloads"
```
stoksc, line 453: post discussion in backend eng, should we just make the the `/metrics` endpoint?

hkang1:
@dzhu this step probably comes at the end, but should the changes generated by the following get checked in also?
```
  make -C proto build
  make -C bindings build
  make -C webui/react build-bindings
```

proto/src/determined/trial/v1/trial.proto
```diff
@@ -84,6 +84,8 @@ message Trial {
   // The wall clock time is all active time of the cluster for the trial,
   // inclusive of everything (restarts, initiailization, etc), in seconds.
   double wall_clock_time = 12;
+  // The sum of sizes of all resources in all checkpoints for the trial.
+  uint64 total_checkpoint_size = 13;
```
hkang1, line 88: **question:** I see this defined here as `uint64` but the endpoint returns this as a string for some reason:
Screen Shot 2022-03-01 at 9 11 55 PM

stoksc (approved):
lgtm

PR #3637, opened 2022-02-19
refactor: centralize configuration; align user get methods

philipnrmn:
## Description

[DET-6691] - Push user model changes upstream

This PR moves several configuration structs out from their associated packages (`db`, `hpimportance`, `resourcemanager`) and into the main configuration file. This avoids circular dependencies, eg the `db` package attempting to examine the configuration, which in turn must import `db` in order to parse the database config. 

This allowed the actual meat of the ticket, which is to abstract `UserByExternalToken` away; instead all parts of the code now call `UserByToken` and the implementation decides whether or not to use the external token based on whether or not it is available in the configuration. 

## Test Plan

Create a new user; login as this new user. If successful, `UserByToken` is still working correctly.

## Checklist

[DET-6691]: https://determinedai.atlassian.net/browse/DET-6691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/internal/config/config.go
```diff
@@ -29,19 +26,52 @@ var (
 var once sync.Once
 var masterConfig *Config
 
+// KubernetesDefaultPriority is the default K8 resource manager priority.
+// TODO: remove duplicate definition from resourcemanagers.
```
ioga, line 30: that should be trivial to do, no?

master/internal/config/resource_config.go
```diff
@@ -1,4 +1,4 @@
-package resourcemanagers
```
mackrorysd, line 1: Bradley made a suggesting that config should perhaps live in master/pkg instead of master/internal, but I don't think that's a change we need to make now. Just an FYI.

master/internal/db/postgres_users.go
```diff
@@ -43,7 +43,12 @@ func (db *PgDB) StartUserSession(user *model.User) (string, error) {
 }
 
 // UserByToken returns a user session given an authentication token.
-func (db *PgDB) UserByToken(token string) (*model.User, *model.UserSession, error) {
+func (db *PgDB) UserByToken(token string, ext *model.ExternalSessions) (
```
mackrorysd, line 46: Meta comment: it would be great if this could be served by a generic plugin, in which case we should pass the whole config. But that was impractical for other reason and this is sufficiently extensible for now.

master/internal/resourcemanagers/resource_managers.go
```diff
@@ -3,13 +3,18 @@ package resourcemanagers
...
 	"github.com/determined-ai/determined/master/pkg/actor"
 )
 
 const (
-	actionCoolDown = 500 * time.Millisecond
+	// DefaultSchedulingPriority is the default resource manager priority.
+	DefaultSchedulingPriority = 42
+
+	actionCoolDown          = 500 * time.Millisecond
+	defaultResourcePoolName = "default"
```
mackrorysd, line 17: Wait - I don't see where DefaultSchedulingPriority or defaultResourcePoolName came from before. Missing something?

master/internal/resourcemanagers/scheduler.go
```diff
@@ -3,12 +3,19 @@ package resourcemanagers
...
 
+	"github.com/determined-ai/determined/master/internal/config"
 	"github.com/determined-ai/determined/master/internal/job"
 	"github.com/determined-ai/determined/master/internal/sproto"
 	"github.com/determined-ai/determined/master/pkg/actor"
 	"github.com/determined-ai/determined/master/pkg/model"
 )
 
+const (
+	fairShareScheduling  = "fair_share"
```
mackrorysd, line 14: Again - don't see where this was before. Suggests I'm missing something...

master/internal/config/resource_config.go
philipnrmn: That's a really good point, I'll make a note to do this in the future.

master/internal/resourcemanagers/resource_managers.go
philipnrmn: `scheduler_config.go` moved into the config package; so I copied some consts back to `resource_managers.go` 

Here's the original block on master:
https://github.com/determined-ai/determined/blob/master/master/internal/resourcemanagers/scheduler_config.go#L15

I thought this was conceptually cleaner than having the exported from config, but maybe it's better overall not to maintain duplicate constants. I'll make this change inline with the fix Ilia suggested above.

master/internal/config/config.go
philipnrmn: Done!

mackrorysd (approved):

PR #3641, opened 2022-02-22
fix: Use pagination in CLI to return all experiments, not the default limit

mapmeld:
## Description

Fixes the CLI's `det experiment list` and `det experiment list-trials` to use API pagination to return all items (when >100 results). This expects 200 items / request until there are no more pages.

An alternative would be one request with a huge limit parameter (such as limit=100000); first commit shows this approach

## Test Plan

Run `det experiment list` on a server with more than 200 experiments.
- All experiments should be returned.
- Order should be first to last.
- There should be no seams between pages.

## Checklist

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -426,7 +426,7 @@ def list_experiments(args: Namespace) -> None:
     if not args.all:
         users = [authentication.must_cli_auth().get_session_user()]
 
-    r = bindings.get_GetExperiments(setup_session(args), users=users)
+    r = bindings.get_GetExperiments(setup_session(args), users=users, limit=100_000)
 
     def format_experiment(e: Any) -> List[Any]:
         result = [
@@ -499,7 +499,9 @@ def scalar_validation_metrics_names(exp: Dict[str, Any]) -> Set[str]:
 
 @authentication.required
 def list_trials(args: Namespace) -> None:
-    r = bindings.get_GetExperimentTrials(setup_session(args), experimentId=args.experiment_id)
+    r = bindings.get_GetExperimentTrials(
+        setup_session(args), experimentId=args.experiment_id, limit=100_000
+    )
     trials = r.trials
 
     headers = ["Trial ID", "State", "H-Params", "Start Time", "End Time", "# of Batches"]
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -196,16 +196,22 @@ def create(args: Namespace) -> None:
 
 
 def limit_offset_paginator(
-    method: Callable, agg_field: str, sess: session.Session, **kwargs: Any
+    method: Callable,
+    agg_field: str,
+    sess: session.Session,
+    page_limit: Optional[int] = 200,
+    offset: Optional[int] = None,
+    **kwargs: Any
 ) -> List[Any]:
     all_objects: List[Any] = []
-    offset = 0
+    limit = page_limit or 200
+    internal_offset = offset or 0
     while True:
-        r = method(sess, limit=200, offset=offset, **kwargs)
+        r = method(sess, limit=limit, offset=internal_offset, **kwargs)
         page_objects = getattr(r, agg_field)
         all_objects += page_objects
-        offset += len(page_objects)
-        if len(page_objects) < 200:
+        internal_offset += len(page_objects)
+        if offset or len(page_objects) < limit:
             break
     return all_objects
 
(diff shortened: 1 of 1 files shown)


ioga (approved):
lgtm, altough it'd be nice to have a reusable paginator helper like

```
def limit_offset_paginator(method, agg_field, *args, **kwargs):
    aggregated = []
    offset = 0
    while True:
        r = method(*args, **kwargs, limit=200, offset=offset
        aggregated += getattr(r, agg_field)
        ....
    return aggregated
```

and then 
```
limit_offset_paginator(bindings.get_GetExperiments, "experiments", setup_session(args), users=users)
...
limit_offset_paginator(bindings.get_GetExperimentTrials, "trials", setup_session(args), exerimentID=args.experiment_id)
```

harness/determined/cli/experiment.py
```diff
@@ -426,7 +426,16 @@ def list_experiments(args: Namespace) -> None:
     if not args.all:
         users = [authentication.must_cli_auth().get_session_user()]
 
-    r = bindings.get_GetExperiments(setup_session(args), users=users)
+    all_experiments = []
```
ioga, line 429: we've got to test this out with a big database of experiments. e.g. 50k experiments
if it has any negative effects then we may need to add the explicit user flags to control the pagination from the CLI

mapmeld:
I've made the suggested shared function for pagination. I don't have a good way to test on the 50k-scale

ioga:
> I've made the suggested shared function for pagination. I don't have a good way to test on the 50k-scale

A python API may be helpful. This takes about 50ms per experiment on my laptop.
```
# while in e2e_tests/tests/fixtures/no_op
from determined.experimental import client
for i in range(N):
    client.create_experiment(config="single-medium-train-step.yaml", model_dir=".")
```

harness/determined/cli/experiment.py
```diff
@@ -195,6 +195,21 @@ def create(args: Namespace) -> None:
         submit_experiment(args)
 
 
+def limit_offset_paginator(
+    method: Callable, agg_field: str, connection_args: Namespace, **kwargs: Any
+) -> List[Union[bindings.v1Experiment, bindings.trialv1Trial]]:
+    all_objects: List[Union[bindings.v1Experiment, bindings.trialv1Trial]] = []
+    offset = 0
+    while True:
+        r = method(setup_session(connection_args), limit=200, offset=offset, **kwargs)
```
ioga, line 204: I don't think we should setup session for every page separately. 
in my snippet I was just passing `setup_session(...)` as an opaque `*args` parameter. this also has a benefit of taking a session from everywhere, e.g. in python sdk where you don't have `connection_args` from argparse.

harness/determined/cli/experiment.py
```diff
@@ -195,6 +195,21 @@ def create(args: Namespace) -> None:
         submit_experiment(args)
 
 
+def limit_offset_paginator(
+    method: Callable, agg_field: str, connection_args: Namespace, **kwargs: Any
+) -> List[Union[bindings.v1Experiment, bindings.trialv1Trial]]:
```
ioga, line 200: can it just be a `List`? we may have many paged API types, and explicitly specifying it here would be a chore.

mapmeld: ok I figured out what I want to do is make this List[Any] and use the places where we call this function to expect a typed list

mapmeld:
The 200-per-page all-print behavior will be the default for `det experiment list` and `det experiment list-trials`. I've added pagination arguments `--limit` and `--offset` if the user wants more control / to get individual pages.

I just noticed the pagination_args in `cli/util.py`. They default to 50/page and include `--reverse` so I was unsure, but I could reuse those if that looks more helpful.

harness/determined/cli/experiment.py
```diff
@@ -195,6 +195,27 @@ def create(args: Namespace) -> None:
...
+) -> List[Any]:
+    all_objects: List[Any] = []
+    limit = page_limit or 200
+    internal_offset = offset or 0
+    while True:
+        r = method(sess, limit=limit, offset=internal_offset, **kwargs)
+        page_objects = getattr(r, agg_field)
+        all_objects += page_objects
+        internal_offset += len(page_objects)
+        if offset or len(page_objects) < limit:
```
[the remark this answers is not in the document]
mapmeld: I think I can fix this with offset is not None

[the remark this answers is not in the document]
ioga: 👍 makes sense to me

varlogtim:
I kind of wish we generated pages instead of waiting to get all of the pages then printing.

PR #3642, opened 2022-02-22
fix: allow modal width to shrink on mobile [DET-6443]

EmilyBonar:
## Description
Changes the default minWidth in useModal to 300px.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/useModal.tsx
+++ b/webui/react/src/hooks/useModal/useModal.tsx
@@ -34,7 +34,7 @@ export enum ModalCloseReason {
 
 const DEFAULT_MODAL_PROPS: Partial<ModalFuncProps> = {
   maskClosable: true,
-  style: { minWidth: 300 },
+  style: { minWidth: 280 },
   visible: true,
 };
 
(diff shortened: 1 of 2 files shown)


webui/react/src/hooks/useModal/useModal.tsx
```diff
@@ -34,7 +34,7 @@ export enum ModalCloseReason {
 
 const DEFAULT_MODAL_PROPS: Partial<ModalFuncProps> = {
   maskClosable: true,
-  style: { minWidth: 600 },
+  style: { minWidth: 300 },
```
hkang1, line 37: **nit:** On chrome, for the device/responsive mode, the smallest screen size support is [Galaxy Fold](https://www.samsung.com/global/galaxy/galaxy-fold/) and it's `280`. If that breaks and makes some modal screens look bad it's fine. I was thinking also, 280 might be easier to find in the code than a "clean and even" 300 if we ever need to reference it.

hkang1 (approved):
LGTM! Thanks for fixing!

PR #3643, opened 2022-02-22
feat: add on_trial_{startup,shutdown} hooks to PyTorchCallback

rb-determined-ai:
## Commentary

I have argued against this feature for a _long_ time because of how users frequently ask for it looking for an end-of-trial hook, rather than an end-of-allocation hook.  We can't deliver the first one.  But I've finally assented. I think this is a pretty basic feature that we should just support.

### DIFF
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -1,3 +1,4 @@
+import contextlib
 import json
 import logging
 import pathlib
@@ -6,7 +7,7 @@ import random
 import time
 from abc import abstractmethod
 from inspect import signature
-from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
+from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union, cast
 
 import numpy as np
 import torch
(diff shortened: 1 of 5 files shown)


harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -186,14 +185,38 @@ def _dataset_repro_warning(fn: str, data_obj: Any) -> str:
...
+            finally:
+                fn()
+
+        with contextlib.ExitStack() as exit_stack:
+            for callback in self.callbacks.values():
+                with self.prof.record_timing(
+                    f"callbacks.{callback.__class__.__name__}.on_trial_startup"
+                ):
+                    callback.on_trial_startup(self.latest_batch, self.env.latest_checkpoint)
+                    exit_stack.enter_context(defer(callback.on_trial_shutdown))
```
ioga, line 201: this is cool

mpkouznetsov: indeed.

mpkouznetsov: One question, though. Are you deliberately _not_ timing the `on_trial_shutdown` callback? I see how it could be complex/expensive (you would probably create a new `Callable` that wraps `callback.on_trial_shutdown` into yet another `with self.prof.record_timing`.

rb-determined-ai:
> caveat about the non-timing of the on_shutdown callback
good catch, that was an accident.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: @mpkouznetsov no, it was an accident.

Shouldn't be expensive enough to matter.  Though the `enter_context(defer(callback.on_trial_shutdown))` shouldn't even be inside the initial `record_timing`.

gh-determined-ai (approved):

PR #3644, opened 2022-02-22
added interaction test for inline editor, fixed component to preserve…

trentwatt:
Created tests for InlineEditor.tsx

Noticed this behavior, which I targeted with test called "preserves input when focus leaves"

editor


trentwatt:
Whups, looks like I included changes from my formatter as well as commits i Pulled

PR #3645, opened 2022-02-22
test: add interaction test for inline editor [DET-6667]

trentwatt:
Added interaction tests to InlineEditor

specifically targeted the observed behavior below: value being reset onBlur 

editor

implemented simple fix, but there are other considerations, see [3559](https://github.com/determined-ai/determined/pull/3559) and [3612](https://github.com/determined-ai/determined/pull/3612). In particular, if the save fails, we want to display a message to the user

### DIFF
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -1,9 +1,15 @@
 import React, {
-  ChangeEvent, HTMLAttributes, KeyboardEvent, useCallback, useEffect, useRef, useState,
-} from 'react';
+  ChangeEvent,
+  HTMLAttributes,
+  KeyboardEvent,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
 
-import css from './InlineEditor.module.scss';
-import Spinner from './Spinner';
+import css from "./InlineEditor.module.scss";
+import Spinner from "./Spinner";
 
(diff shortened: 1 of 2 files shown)


webui/react/src/components/InlineEditor.tsx
```diff
@@ -63,8 +63,11 @@ const InlineEditor: React.FC<Props> = ({
   const save = useCallback(async (newValue: string) => {
     if (onSave) {
       setIsSaving(true);
-      const err = await onSave(newValue);
-      if (err !== null) {
```
mapmeld, line 67: After discussion with Emily, I support making this `if (err != null) {` and no other changes to this InlineEditor file

mapmeld:
Can we add a test where onSave returns an error object?

webui/react/src/components/InlineEditor.tsx
trentwatt: Ok great! With that change, the test "preserves input when focus leaves" passes, whereas before it failed.

trentwatt:
> Can we add a test where onSave returns an error object?

Ok, done.

webui/react/src/components/InlineEditor.tsx
mapmeld: great!  It turns out I was wrong about no other changes in the file
as you pointed out, the `Promise<void>` type should also be updated here.

mapmeld (approved):

PR #3646, opened 2022-02-22
feat: create modal to edit displayName in UI (DET-6484, DET-6581)

johnkim-det:
## Description

Add a Display Name field and a "Change name" button to the User Settings modal. Clicking the button opens a modal that allows the user to set a new display name, which will be displayed instead of the user's username.

## Test Plan

- [ ] use modal to set displayName
- [ ] make sure validation prevents displayNames over 80 characters
- [ ] confirm that displayName and Avatar are updated everywhere they should be (especially Experiments page, including filters for the users column)

## Commentary (optional)

Setting the displayName itself was relatively straightforward, but I'm less confident about DET-6581 changes, particularly in areas of the app that I'm less familiar with. For review, please double-check that I didn't miss anything or change anything incorrectly.

I only changed `username` to `displayName || username` in places where it looked like username was used as a label, not as an identifier. Additional changes were needed in places that displayed a username from somewhere else other than the `auth` user (ModelHeader.tsx, ModelVersionHeader.tsx, Table.tsx). There may be additional changes needed here later, as part of https://determinedai.atlassian.net/browse/DET-6736.

## Checklist


webui/react/src/hooks/useModal/UserSettings/useModalChangeName.tsx
```diff
@@ -0,0 +1,86 @@
+import { Button, Form, Input, message } from 'antd';
```
hkang1, line 1: **chore:** interaction test please 🙏

hkang1:
I've updated my display name to `hoyoul`. Is the expected behavior that going forward it should show my updated `( H )` avatar after that update? The screen here shows the avatar on the upper left is the updated one, but the experiment list avatars (still me) are showing the old ones.

Screen Shot 2022-02-25 at 1 29 19 PM

hkang1:
The password change takes about 2~4 seconds for it to "take" and dismiss the modal. Is it possible to make the experience consistent with the display name change? (can't remember if there was a separate ticket for this)

Screen Shot 2022-02-25 at 2 01 55 PM

hkang1 (commented):
The code looks good, with minor comments. The biggest thing is the reported issue where when you change your display name it didn't seem to update the avatars on experiments (even newly created experiments).

hkang1 (commented):
The code mostly looks good, but there might be issues (possibly on my end).

So I rebuilt my cluster just in case with the master API changes. But the changes to the `displayName` is working with my specific user. But the full list of users endpoint is NOT reflecting my updated `displayName`, where it still shows it as an empty string. So as a result this happens:

https://user-images.githubusercontent.com/220971/156608734-b754a87b-13e1-470b-9d85-06456d5b7791.mov

hkang1:
> The code mostly looks good, but there might be issues (possibly on my end).
> 
> So I rebuilt my cluster just in case with the master API changes. But the changes to the `displayName` is working with my specific user. But the full list of users endpoint is NOT reflecting my updated `displayName`, where it still shows it as an empty string. So as a result this happens:
> 
>  Screen.Recording.2022-03-03.at.9.32.33.AM.mov 
> So I think the `GET /api/v1/users` endpoint needs updating to pull the display names in maybe?

My fault, after a full API rebuild the `/users` endpoint returned the display name as expected is working great!

hkang1 (approved):
LGTM, nice work with this PR. It was an interesting challenge to bring in context into something dynamically created. Thanks for establishing this pattern!

PR #3649, opened 2022-02-23
chore: Send password reset through PATCH-ing user [DET-6705]

mapmeld:
## Description

We currently have a few API routes to change user attributes. We'd like to use only PATCH user.
- WebUI had used `setUserPassword` endpoint; now uses `patchUser`
- The CLI used `internal/user/service.go` and now uses `patchUser`
- postUser is already receiving a hashed password. We do server-side hashing only if requested by login or `setUserPassword`

**Work in progress while I check a few things with backend team**

- Can we reuse `PasswordCanBeModifiedBy` in user changes?
- Do we have alternate login pages?
- Do we have alternate logins (SCIM) using `setUserPassword`, or can this be removed from proto / bindings / etc?
- Is the CLI the only use of `internal/user/service.go` ?  Would we want CLI to set `active`, `agent_user_group`, and `username` through the new endpoint?

## Test Plan

- Change password through web menu
- Change password with `det user change-password`

## Checklist


mapmeld:
Simplified our `SetUserPassword` endpoint by having it go through `PatchUser`

After discussing with backend team:

- Alternate logins / SCIM => we should keep old login and password-change endpoints. I'm considering removing password-update from `service.go`, need to verify.
- `User.PasswordCanBeModifiedBy` function cannot be used here because we have a `userv1.User` object, not a `model.User`. If we remove password update from `service.go`, this also can be removed.
- No need to migrate CLI's other user changes, because we are keeping the endpoint

gh-determined-ai (approved):
In "The `password` parameter expects ...", is it really a parameter? If this applies to all 3 interfaces, the term "field" or something more general might be more appropriate. "Parameter" implies a native API, only, and you don't show a syntax to which 'password' might apply.
Also, aren't we using double backticks for code/interface/filename syntax highlighting?

gh-determined-ai:
Yeah, there isn't much about reST conventions documented other than rstfmt.py. You almost have to grep the doc source to see we use ``.
________________________________

master/internal/api_user.go
```diff
@@ -158,17 +149,20 @@ func (a *apiServer) PatchUser(
...
+		if err = updateUser.UpdatePasswordHash(req.User.Password.Value); err != nil {
 			return nil, err
 		}
 	}
-	fullUser, err := getUser(a.m.db, req.Username)
-	return &apiv1.PatchUserResponse{User: fullUser}, err
+	// return the finalUser object so we can be assured only public fields are passed on
+	finalUser := &userv1.User{}
+	err = a.m.db.QueryProto(
+		"update_user", finalUser, updateUser.Username, updateUser.DisplayName, updateUser.PasswordHash)
```
stoksc, line 165: if this CLI is going to use this, doesn't either this or the CLI need to add the client side hash? if the webui is still sending it hashed, i think this would break.

stoksc (commented):
i didn't review when display name went in so sorry i'm a bit late to the party. i think we should write this so that `user.Service.patchUser` is public and  the `user.Service` singleton lives on the API server / master. Then `apiServer.SetUserPassword` and `apiServer.PatchUser` call `user.Service.PatchUser`, rather than fabricating api requests and having the api server call itself. this is usually more flexible long term

proto/src/determined/user/v1/user.proto
```diff
@@ -28,6 +29,8 @@ message User {
 message PatchUser {
   // Name to display in the web UI.
   string display_name = 2;
+  // Updated hashed and salted password.
+  google.protobuf.StringValue password = 3;
```
stoksc, line 33: i see this is indeed a password hash, so can we call the variable `password_hash`

stoksc (approved):
lgtm, but i _really_ think we should push shared logic to `user.Service` rather than putting so much in the apiServer and having it call itself.

PR #3653, opened 2022-02-23
fix: stop experiment modal closing unexpectedly [DET-6746]

EmilyBonar:
## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/useModal.tsx
+++ b/webui/react/src/hooks/useModal/useModal.tsx
@@ -45,6 +45,7 @@ const useModal = (
   options?: ModalOptions,
 ): ModalHooks => {
   const modalRef = useRef<ReturnType<ModalFunc>>();
+  const componentUnmounting = useRef(false);
   const [ modalProps, setModalProps ] = useState<ModalFuncProps>();
   const prevModalProps = usePrevious(modalProps, undefined);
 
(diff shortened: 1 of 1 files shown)


trentwatt:
this resolves the issue for me on Safari, Firefox, and Chrome

webui/react/src/hooks/useModal/useModal.tsx
```diff
@@ -106,10 +107,19 @@ const useModal = (
     }
   }, [ extendEventHandler, modalProps, options, prevModalProps ]);
 
+  // Sets componentUnmounting to true only when the parent component is unmounting so that the next
+  // useEffect only runs modalClose on unmount, rather than every time modalClose updates.
+  // The order of these two useEffects matters, this one has to be first.
```
hkang1, line 112: great explanation for this!

hkang1: **nit:** for multi-line comments, suggest the use of `/** */` format:
```
/**
 * Sets componentUnmounting to true only when the parent component is unmounting so that the next
 * useEffect only runs modalClose on unmount, rather than every time modalClose updates.
 * The order of these two useEffects matters, this one has to be first.
 */
```

hkang1 (approved):
This is a very simple effective way to add to handle the said problem!

PR #3656, opened 2022-02-23
chore: upgrade go-pg [DET-6363]

dzhu:
## Description

PostgreSQL 11+ can be configured to use SCRAM-SHA-256-PLUS, a new authentication
method that would cause the older version of go-pg to raise an error on
connection. The new version accepts it and doesn't raise an error (though it
still falls back to SCRAM-SHA-256 for the actual connection).

Changes generated by running `go get github.com/go-pg/pg/v10@latest`.

## Test Plan

- [x] configure database to produce the original error, upgrade, confirm that connection now succeeds
- [x] normal CI
- [x] new devcluster tests for running some basic tasks against different Postgres versions with SSL enabled (I [confirmed](https://app.circleci.com/pipelines/github/determined-ai/determined?branch=wip-postgres-ssl-tests) that the connection fails on Postgres 14 without this code change)


ioga:
huh, I thought jackc/pgx doesn't have `PLUS` support. at least there's no mention of it anywhere in the code: https://github.com/jackc/pgconn/blob/master/auth_scram.go#L41 
does it "just work"?

dzhu:
The issue and fix involve go-pg, not pgx; I'm not sure what the division of responsibilities between them is supposed to be, exactly, but in any case it did switch from not connecting to connecting after the upgrade.

ioga:
> The issue and fix involve go-pg, not pgx; I'm not sure what the division of responsibilities between them is supposed to be, exactly, but in any case it did switch from not connecting to connecting after the upgrade.

master uses go-pg for migrations and pgx for all runtime work, so we need them both to work with SCRAM. if it connects on startup, and then any runtime psql query also works fine, then we should be good.

dzhu:
Got it, thanks. Yeah, it does work with both migrations and later queries.

ioga:
is it possible to have a devcluster config and a test for this in circleci?

ioga (approved):

PR #3657, opened 2022-02-23
fix: [DET 6509] when cluster fails ensure end time of open allocation is set to the last cluster heartbeat

nrajanee:
## Description

fix: [DET 6509] when cluster fails ensure end time of open allocation is set to the last cluster heartbeat

## Test Plan

Added an integration test which tests to ensure that the cluster heartbeat is added to the cluster id table accurately. It also checks whether open allocations have the correct cluster heartbeat as their endTime. The integration test covers the new functions added to postgres_cluster.go and the change made in CloseOpenAllocaitons in postgres_tasks.go. Manually check and ensure that the cluster heartbeat is getting updated every 10 minutes. This manual check will ensure that the go routine added to core.go works correctly. 

## Checklist


master/static/migrations/20220215164216_add-cluser-id-cluser-heartbeat.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE public.cluster_id ADD COLUMN cluster_heartbeat timestamp not null DEFAULT (DATE_TRUNC('millisecond', now() at time zone 'utc')::timestamp);
```
stoksc, line 1: can you add new line endings to the sql files and format them a bit?

master/internal/db/postgres_tasks.go
```diff
@@ -192,11 +192,17 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 // CloseOpenAllocations finds all allocations that were open when the master crashed
 // and adds an end time.
 func (db *PgDB) CloseOpenAllocations() error {
+	cluster_heartbeat, err := db.GetClusterHeartBeat()
+
+	if err != nil {
+		return err
+	}
```
stoksc, line 199:
```suggestion
	cluster_heartbeat, err := db.GetClusterHeartBeat()
	if err != nil {
		return err
	}
```

master/internal/db/postgres_tasks.go
```diff
@@ -192,11 +192,17 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
...
+
 	if _, err := db.sql.Exec(`
 UPDATE allocations
-SET end_time = current_timestamp AT TIME ZONE 'UTC'
+SET end_time = $1
 WHERE end_time IS NULL
-`); err != nil {
+`, cluster_heartbeat); err != nil {
```
stoksc, line 205: I would prefer something like
```
UPDATE allocations SET end_time = cluster_heartbeat FROM cluster_id;
```
and dropping the code above this, with the advantages mainly being `GetClusterHeartBeat` becomes unused and less code is always better (and this becomes much less code), this is 1 db call instead of 2 (but this is not a big deal since it's every 10m, but applies elsewhere so it makes sense to try to be consistent).

master/internal/db/postgres_cluster.go
```diff
@@ -32,6 +32,38 @@ WHERE NOT EXISTS ( SELECT * FROM cluster_id );
...
+func (db *PgDB) UpdateClusterHeartBeat(current_cluster_heartbeat time.Time) error {
+	var uuidVal []string
+	if err := db.sql.Select(&uuidVal, `SELECT cluster_id FROM cluster_id`); err != nil {
+		return errors.Wrapf(err, "error reading cluster_id from cluster_id table")
+	}
+	if len(uuidVal) != 1 {
+		return errors.Errorf(
+			"expecting exactly one cluster_id from cluster_id table, %d values found", len(uuidVal),
+		)
+	}
```
stoksc, line 44: this is all basically unused, I think we can just get rid of this. if it is to check the 1 row invariant is upheld, then i would say the check should happen elsewhere because there is nothing we can do about it here, we may as well still do what we need to do (set cluster heartbeat).

master/internal/db/postgres_cluster.go
```diff
@@ -32,6 +32,38 @@ WHERE NOT EXISTS ( SELECT * FROM cluster_id );
...
+
+	_, err := db.sql.Exec(`UPDATE cluster_id SET cluster_heartbeat = $1`, current_cluster_heartbeat) //now() returns UTC time
+
+	return err
```
stoksc, line 48:
```suggestion
	return errors.Wrap(err, "updating cluster heartbeat")
```

master/internal/db/postgres_cluster.go
```diff
@@ -32,6 +32,38 @@ WHERE NOT EXISTS ( SELECT * FROM cluster_id );
...
+	}
+	if len(row_heartbeat) != 1 {
+		return time.Time{}, errors.Errorf(
+			"expecting exactly one cluster_heartbeat from cluster_id table, %d values found", len(row_heartbeat),
+		)
+	}
+
+	var layout = "2006-01-02T15:04:05.000Z"
+	cluster_heartbeat, err := time.Parse(layout, row_heartbeat[0])
+	return cluster_heartbeat, err
```
stoksc, line 64: i commented elsewhere i dont think we need this func, but i'm also curious, i would think this should just work if you use time.Time rather than string + time.Parse. is that not the case?

master/internal/core.go
```diff
@@ -613,6 +613,19 @@ func (m *Master) rwCoordinatorWebSocket(socket *websocket.Conn, c echo.Context)
...
+	t := time.NewTicker(10 * time.Minute)
+	defer t.Stop()
+	current_cluster_heartbeat := time.Now().UTC().Truncate(time.Millisecond)
+	for range t.C {
+		err := db.UpdateClusterHeartBeat(current_cluster_heartbeat)
+		if err != nil {
+			log.Error(err.Error())
+		}
+
+	}
```
stoksc, line 626: to get the first tick immediately
```suggestion
	t := time.NewTicker(10 * time.Minute)
	defer t.Stop()
	current_cluster_heartbeat := time.Now().UTC().Truncate(time.Millisecond)
	for {
		err := db.UpdateClusterHeartBeat(current_cluster_heartbeat)
		if err != nil {
			log.Error(err.Error())
		}
		<-t.C
	}
```

stoksc: also just because making sure we can gracefully shutdown is generally useful, can you go ahead pass a `context.Context` to this?

master/internal/db/postgres_cluster_intg_test.go
```diff
@@ -0,0 +1,88 @@
...
+	MustMigrateTestPostgres(t, db, migrationsFromDB)
+
+	_, err := db.GetOrCreateClusterID()
+
+	require.NoError(t, err, "failed to get or create cluster id")
+
+	current_time := time.Now().UTC().Truncate(time.Millisecond)
+	db.UpdateClusterHeartBeat(current_time)
+
+	cluster_heartbeat, err := db.GetClusterHeartBeat()
```
stoksc, line 30: require.NoError after this

master/internal/db/postgres_tasks.go
nrajanee: Yes! I'm not sure why I did it this way in the first place. Getting directly from the cluster_id makes a lot more sense.

master/internal/db/postgres_cluster_intg_test.go
```diff
@@ -0,0 +1,88 @@
...
+	//Don't complete the above allocation and call CloseOpenAllocations
+
+	db.CloseOpenAllocations()
+
+	//Retrieve the open allocation and check if end time is set to cluster_heartbeat
+
+	aOut, err := db.AllocationByID(aIn.AllocationID)
+
+	if *aOut.EndTime != cluster_heartbeat {
+		errors.Errorf("Expected end time of open allocation is = %q but it is = %q instead", cluster_heartbeat.String(), (*aOut.EndTime).String())
```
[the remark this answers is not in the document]
nrajanee: Oops yes thank you. I missed this one!

master/static/migrations/20220215164216_add-cluser-id-cluser-heartbeat.up.sql
nrajanee: done

master/internal/db/postgres_tasks.go
nrajanee: I changed the sql query to this: UPDATE allocations
SET end_time = cluster_heartbeat FROM cluster_id
WHERE end_time IS NULL. We need the end_time is NULL bit to ensure that only the open allocations end_time is set to cluster heartbeat, correct?

master/internal/db/postgres_cluster.go
nrajanee: I agree. done!

master/internal/db/postgres_cluster.go
nrajanee: Do you mean something like this: 

var rowHeartbeat time.Time 
	if err := db.sql.Select(&rowHeartbeat, `SELECT cluster_heartbeat FROM cluster_id`); err != nil {
		return time.Time{}, errors.Wrapf(err, "error reading cluster_heartbeat from cluster_id table")
	}

Yes, I guess that could work. I'm not sure I'll have to try that.

master/internal/db/postgres_tasks.go
nrajanee: And I got rid fo the GetClusterHeartBeat method from postgres_cluster.go

master/internal/db/postgres_cluster_intg_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestClusterAPI(t *testing.T) {
+	etc.SetRootPath(rootFromDB)
+
+	db := MustResolveTestPostgres(t)
+	MustMigrateTestPostgres(t, db, migrationsFromDB)
+
+	_, err := db.GetOrCreateClusterID()
+
+	require.NoError(t, err, "failed to get or create cluster id")
```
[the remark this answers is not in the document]
nrajanee: Yes it does make it easier to read. done!

master/internal/db/postgres_cluster_intg_test.go
nrajanee: done!

master/internal/core.go
nrajanee: done!

stoksc:
@nrajanee when you are done addressing feedback, could you reassign me so i know to re-review (or request a re-review and re-assign, too)

stoksc:
@cla-bot[bot] check

stoksc:
@nrajanee you're added to contributors but cla-bot is still failing so you'll need to follow the instructions it commented likely. fyi, make sure to sign your commits with your @hpe email.

nrajanee:
@cla-bot[bot] check

nrajanee:
@cla-bot[bot] check

nrajanee:
@cla-bot[bot] check

nrajanee:
@cla-bot[bot] check

master/internal/db/postgres_cluster_intg_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Slots:        8,
+		AgentLabel:   "something",
+		ResourcePool: "somethingelse",
+		StartTime:    time.Now().UTC().Truncate(time.Millisecond),
+	}
+
+	err = db.AddAllocation(aIn)
+	require.NoError(t, err, "failed to add allocation")
+
+	//Don't complete the above allocation and call CloseOpenAllocations
```
stoksc, line 73: nit here and elsewhere
```suggestion
	// Don't complete the above allocation and call CloseOpenAllocations
```

master/internal/db/postgres_cluster_intg_test.go
```diff
@@ -0,0 +1,80 @@
...
+	db.CloseOpenAllocations()
+
+	//Retrieve the open allocation and check if end time is set to cluster_heartbeat
+	aOut, err := db.AllocationByID(aIn.AllocationID)
+	require.Equal(t, *aOut.EndTime, clusterHeartbeat, "Expected end time of open allocation is = %q but it is = %q instead", clusterHeartbeat.String(), (*aOut.EndTime).String())
```
stoksc, line 78: require.NotNil(aOut) and aOut.Endtime first so we get an error message in tests not a large stack trace for a nil deref

master/static/migrations/20220215164216_add-cluser-id-cluser-heartbeat.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE public.cluster_id DROP COLUMN IF EXISTS cluster_heartbeat;
```
stoksc, line 1: new lines

stoksc (commented):
lgtm, just a few final comments

nrajanee:
@cla-bot[bot] check

master/internal/db/postgres_cluster_intg_test.go
nrajanee: done!

master/internal/db/postgres_cluster_intg_test.go
nrajanee: done!

master/static/migrations/20220215164216_add-cluser-id-cluser-heartbeat.down.sql
nrajanee: done

stoksc (approved):
lgtm. Great work!

PR #3659, opened 2022-02-24
feat: add `accelerator` field for resource pool [DET-6755]

gt2345:
## Description

Add GPU accelerator field for resource pool.
For AWS instance, accelerator is implied from this article: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/accelerated-computing-instances.html 

## Test Plan

Ping `/resource-pools` endpoint to verify new field `accelerator` is added

### DIFF
--- a/master/internal/resourcemanagers/agent_resource_manager.go
+++ b/master/internal/resourcemanagers/agent_resource_manager.go
@@ -294,6 +294,7 @@ func (a *agentResourceManager) createResourcePoolSummary(
 	instanceType := ""
 	slotsPerAgent := -1
 	slotType := device.ZeroSlot
+	accelerator := "-"
 
 	if pool.Provider != nil {
 		if pool.Provider.AWS != nil {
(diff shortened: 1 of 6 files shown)


master/internal/resourcemanagers/provisioner/aws_config.go
```diff
@@ -233,6 +238,39 @@ func (t ec2InstanceType) Slots() int {
...
+	if strings.HasPrefix(instanceType, "g5") {
+		return "NVIDIA A10G"
+	}
+	if strings.HasPrefix(instanceType, "g4ad") {
+		return "AMD Radeon Pro V520"
+	}
+	if strings.HasPrefix(instanceType, "g4dn") {
+		return "NVIDIA T4 Tensor Core"
+	}
+	return "-"
```
ioga, line 271: see earlier comment about "-" vs "" for empty accelerator response.

master/internal/resourcemanagers/provisioner/aws_config.go
```diff
@@ -233,6 +238,39 @@ func (t ec2InstanceType) Slots() int {
 	return 0
 }
 
+// source: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/accelerated-computing-instances.html
```
ioga, line 241: is there any way to discover that over some API?

master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -294,6 +294,7 @@ func (a *agentResourceManager) createResourcePoolSummary(
 	instanceType := ""
 	slotsPerAgent := -1
 	slotType := device.ZeroSlot
+	accelerator := "-"
```
ioga, line 297: If I am not mistaken, `gceInstanceType.GPUType` will return an empty string when there's not accelerator. I'd suggest using empty string to denote an "empty" value here and in `AWS.Accelerator()` method. If webui needs to display "-" or "N/A" or other special visual treatment, then it can be done on client side.

ioga (approved):

PR #3661, opened 2022-02-24
fix: det user whoami to return current user in notebooks

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/DET-3851

`det user whoami` always returns `determined` inside shells, notebooks, commands, etc. This happens because the authentication and current user information is not carried over from the client launching the shell (or notebook or command).

A proposed solution is to communicate a user and the corresponding user token via environment variables. We already have a precedent for injecting the gRPC token via environment variable.

## Commentary (optional)

- Following @rb-determined-ai 's suggestion I made this work for experiments as well.
- Following another suggestion of his I create a new user token for use in the started container instead of stealing the one passed by the client requesting the creation of the task/experiment.
- An automated test confirms the non-default user starting a shell will recognize themselves in the answer to `det user whoami` inside the shell

### Notes on the approach taken
We already create a variable `DET_SESSION_TOKEN` containing a gRPC token. It is not useful for our purpose here, in my opinion, as it does not pertain to a specific user. I am adding two new variables holding the current determined user and the corresponding user token that the internal client will add to the token store.

## Checklist

### DIFF
--- a/harness/determined/common/api/authentication.py
+++ b/harness/determined/common/api/authentication.py
@@ -70,7 +70,13 @@ class Authentication:
             self.token_store.drop_user(session_user)
             token = None
 
-        if token is None and util.get_container_user_name() is not None:
+        # util.get_container_user_name() and util.get_container_user_token()
+        # are either both None or both strings
+        if (
+            token is None
+            and util.get_container_user_name() is not None
+            and util.get_container_user_token() is not None
+        ):
             session_user, token = util.get_container_user_name(), util.get_container_user_token()
 
         if token is not None:
--- a/harness/determined/common/api/authentication.py
+++ b/harness/determined/common/api/authentication.py
@@ -62,10 +62,11 @@ class Authentication:
         session_user = (
             requested_user
             or self.token_store.get_active_user()
+            or util.get_initial_user()
             or constants.DEFAULT_DETERMINED_USER
         )
 
-        token = self.token_store.get_token(session_user)
+        token = self.token_store.get_token(session_user) or util.get_user_token()
         if token is not None and not _is_token_valid(self.master_address, token, cert):
             self.token_store.drop_user(session_user)
             token = None
--- a/harness/determined/common/api/authentication.py
+++ b/harness/determined/common/api/authentication.py
@@ -77,7 +77,8 @@ class Authentication:
             and util.get_container_user_name() is not None
             and util.get_container_user_token() is not None
         ):
-            session_user, token = util.get_container_user_name(), util.get_container_user_token()
+            session_user = util.get_container_user_name()  # type: ignore
+            token = util.get_container_user_token()
 
         if token is not None:
             return Session(session_user, token)
--- a/harness/determined/common/api/authentication.py
+++ b/harness/determined/common/api/authentication.py
@@ -62,14 +62,17 @@ class Authentication:
         session_user = (
             requested_user
             or self.token_store.get_active_user()
-            or util.get_initial_user()
+            or util.get_container_user()
             or constants.DEFAULT_DETERMINED_USER
         )
 
-        token = self.token_store.get_token(session_user) or util.get_user_token()
+        token = self.token_store.get_token(session_user)
         if token is not None and not _is_token_valid(self.master_address, token, cert):
             self.token_store.drop_user(session_user)
             token = None
+
+        if token is None:
+            token = util.get_user_token()
 
         if token is not None:
             return Session(session_user, token)
--- a/harness/determined/common/api/authentication.py
+++ b/harness/determined/common/api/authentication.py
@@ -62,7 +62,6 @@ class Authentication:
         session_user = (
             requested_user
             or self.token_store.get_active_user()
-            or util.get_container_user()
             or constants.DEFAULT_DETERMINED_USER
         )
 
@@ -71,8 +70,8 @@ class Authentication:
             self.token_store.drop_user(session_user)
             token = None
 
-        if token is None:
-            token = util.get_user_token()
+        if token is None and util.get_container_user_name() is not None:
+            session_user, token = util.get_container_user_name(), util.get_container_user_token()
 
         if token is not None:
             return Session(session_user, token)
--- a/harness/determined/common/util.py
+++ b/harness/determined/common/util.py
@@ -5,7 +5,7 @@ import pathlib
 import platform
 import random
 import sys
-from typing import IO, Any, Callable, Iterator, Sequence, TypeVar, Union, overload
+from typing import IO, Any, Callable, Iterator, Optional, Sequence, TypeVar, Union, overload
 
 from determined.common import yaml
 
@@ -46,6 +46,14 @@ def sizeof_fmt(val: Union[int, float]) -> str:
 
 def get_default_master_address() -> str:
     return os.environ.get("DET_MASTER", os.environ.get("DET_MASTER_ADDR", "localhost:8080"))
+
+
+def get_initial_user() -> Optional[str]:
+    return os.environ.get("DET_USER")
+
+
+def get_user_token() -> Optional[str]:
+    return os.environ.get("DET_USER_TOKEN")
 
 
 def debug_mode() -> bool:
--- a/harness/determined/common/util.py
+++ b/harness/determined/common/util.py
@@ -48,7 +48,7 @@ def get_default_master_address() -> str:
     return os.environ.get("DET_MASTER", os.environ.get("DET_MASTER_ADDR", "localhost:8080"))
 
 
-def get_initial_user() -> Optional[str]:
+def get_container_user() -> Optional[str]:
     return os.environ.get("DET_USER")
 
 
--- a/harness/determined/common/util.py
+++ b/harness/determined/common/util.py
@@ -48,11 +48,11 @@ def get_default_master_address() -> str:
     return os.environ.get("DET_MASTER", os.environ.get("DET_MASTER_ADDR", "localhost:8080"))
 
 
-def get_container_user() -> Optional[str]:
+def get_container_user_name() -> Optional[str]:
     return os.environ.get("DET_USER")
 
 
-def get_user_token() -> Optional[str]:
+def get_container_user_token() -> Optional[str]:
     return os.environ.get("DET_USER_TOKEN")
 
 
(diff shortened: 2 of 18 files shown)


harness/determined/common/util.py
```diff
@@ -48,6 +48,14 @@ def get_default_master_address() -> str:
     return os.environ.get("DET_MASTER", os.environ.get("DET_MASTER_ADDR", "localhost:8080"))
 
 
+def get_initial_user() -> Optional[str]:
```
rb-determined-ai, line 51: can we call this `get_container_user()` or something?  `initial` doesn't really explain that it's taking the username of who launched the container.

master/internal/api_command.go
```diff
@@ -100,6 +100,10 @@ func (a *apiServer) getCommandLaunchParams(ctx context.Context, req *protoComman
 	taskSpec.TaskContainerDefaults = taskContainerDefaults
 	taskSpec.AgentUserGroup = agentUserGroup
 	taskSpec.Owner = user
+	taskSpec.ExtraEnvVars = map[string]string{
```
rb-determined-ai, line 103: is there any way we can set these for all containers in `pkg/task/task.go::TaskSpec.EnvVars()`?

We definitely need these set for trial containers as well as for notebooks.

master/internal/grpcutil/auth.go
```diff
@@ -100,13 +100,13 @@ func GetUser(ctx context.Context, d *db.PgDB, extConfig *model.ExternalSessions)
 	switch err {
 	case nil:
 		if !user.Active {
-			return nil, nil, ErrPermissionDenied
+			return nil, nil, "", ErrPermissionDenied
 		}
-		return user, session, nil
+		return user, session, token, nil
```
rb-determined-ai, line 105: question: do tokens have expiration times?  As in, should a fresh container token be created, maybe with a different expiration time?  I'm thinking that reusing the actual token seems a bit fishy to me, but I can't pretend to know our master-side use system so I'm not sure.

harness/determined/common/api/authentication.py
```diff
@@ -62,10 +62,11 @@ def _init_session(
...
             or self.token_store.get_active_user()
+            or util.get_initial_user()
             or constants.DEFAULT_DETERMINED_USER
         )
 
-        token = self.token_store.get_token(session_user)
+        token = self.token_store.get_token(session_user) or util.get_user_token()
         if token is not None and not _is_token_valid(self.master_address, token, cert):
             self.token_store.drop_user(session_user)
             token = None
```
rb-determined-ai, line 72: I think this line:

```python
if token is not None and not _is_token_valid(self.master_address, token, cert):
```

should not apply to the session container token.  Maybe:

```python
        token = self.token_store.get_token(session_user)
        if token is not None and not _is_token_valid(self.master_address, token, cert):
            self.token_store.drop_user(session_user)
            token = None
        # do it here instead:
        if token is None:
            token = util.get_user_token()
```

rb-determined-ai (commented):
Honestly you made this look really easy.  I've been telling people it would be really hard for a long time.  I've taken at least two stabs at it too, it's not even like I was dismissing it without trying.

I'm very impressed!

harness/determined/common/api/authentication.py
```diff
@@ -70,6 +71,9 @@ def _init_session(
...
```
rb-determined-ai, line 72: it seems like at this point in the code, the `session_user` we set before will not longer be valid.

harness/determined/common/util.py
```diff
@@ -48,6 +48,14 @@ def get_default_master_address() -> str:
     return os.environ.get("DET_MASTER", os.environ.get("DET_MASTER_ADDR", "localhost:8080"))
 
 
+def get_container_user() -> Optional[str]:
+    return os.environ.get("DET_USER")
+
+
+def get_user_token() -> Optional[str]:
```
rb-determined-ai, line 55: These names are better than before.

Nit: maybe `get_container_user_name` and `get_container_user_token`?  To show that they are collecting information from the same source, so to speak.

master/internal/grpcutil/auth.go
```diff
@@ -116,7 +116,6 @@ func auth(ctx context.Context, db *db.PgDB, fullMethod string,
 	if unauthenticatedMethods[fullMethod] {
 		return nil
 	}
-
```
rb-determined-ai, line 119: extraneous diff, restore it to how it was

master/pkg/tasks/task.go
```diff
@@ -115,6 +117,12 @@ func (t TaskSpec) EnvVars() map[string]string {
 		"DET_ALLOCATION_ID": t.AllocationID,
 		"DET_SESSION_TOKEN": t.AllocationSessionToken,
 	}
+	if t.UserSessionToken != "" {
```
rb-determined-ai, line 120: question: when would we ever launch a container which didn't have a user and a token?  Shouldn't that always be the case?

mpkouznetsov: I learned it the hard way :) checkpoint gc is the answer.

mpkouznetsov:
Besides an added unit test, I manually confirmed that the environment-variable-based user specification can be overridden inside a notebook by logging in as a different user:

```
root@c626fbd9ad5b:/run/determined/workdir# det user whoami
You are logged in as user 'maksimka'
root@c626fbd9ad5b:/run/determined/workdir# det user login admin
Password for user 'admin': 
root@c626fbd9ad5b:/run/determined/workdir# det user whoami
You are logged in as user 'admin'
root@c626fbd9ad5b:/run/determined/workdir# det user logout
root@c626fbd9ad5b:/run/determined/workdir# det user whoami
You are logged in as user 'maksimka'
root@c626fbd9ad5b:/run/determined/workdir# 
```

master/pkg/tasks/task.go
rb-determined-ai: Can we give them the same user/token as an experiment?  That seems cleaner to me since checkpoint gc always belongs to one experiment which always belongs to one user.  If not, at least include a comment here explaining when these values are empty.

rb-determined-ai (commented):
Can you add an e2e test for this manual test you ran:

> Besides an added unit test, I manually confirmed that the environment-variable-based user specification can be overridden inside a notebook by logging in as a different user:

You are right that notebooks don't normally have `det` installed, but you can write a script that mimics our `master/static/srv/notebook-entrypoint.sh`:

```sh
#!/bin/bash

set -e
set -x

source /run/determined/task-logging-setup.sh

export PATH="/run/determined/pythonuserbase/bin:$PATH"

python3 -m pip install -q --user /opt/determined/wheels/determined*.whl

python3 -m determined.exec.prep_container --resources

det user whoami
echo PASSWORD | det user login admin
det user whoami
det user logout
det user whoami
```

Then you can use the same logic as the `e2e_tests/tests/experiment/experiment::assert_patterns_in_trial_logs()` method to ensure you find each of the `You are logged in as` lines that you expect, in order.

rb-determined-ai:
Also, can you make sure somebody with more user-system backend knowledge reviews this PR?  In particular, I want to make sure that we are not going to explode the user system or the master's cpu if somebody submits e.g. 10k experiments and we create 10k new user sessions.

ioga:
1. I agree, creating a new session for each task seems like a lot. Should we try to reuse a recent session instead (e.g. generated with an hour), if it's available? And create a new one otherwise.
2. User session today reflects a user signing in; will it be valuable for security/audit/debugging reasons to add a column on `user_sessions` to distinguish user-originated sessions vs auto-generated sessions? if yes, then I believe the "recent session" lookup from #1 will need to only take auto-generated sessions into account.

rb-determined-ai:
Another question, @ioga, can we just remove the GPRC auth token thing?  Why do we even need it now?  My understand was that we injected that to allow certain connections from training to the master, but this feels like a more general and more correct solution.

master/pkg/tasks/task.go
mpkouznetsov: After trying it for hours I could not reproduce the situation where the owner is not being set, even for a checkpoint gc. I am going to rely on it being there; if we run into problems downstream I will deal with them.

ioga:
> Another question, @ioga, can we just remove the GPRC auth token thing? Why do we even need it now? My understand was that we injected that to allow certain connections from training to the master, but this feels like a more general and more correct solution.

I don't know this. @stoksc ?

stoksc:
session per task, even if it is a million rows seems fine for a table with just (user_id, expiry). seems much more straightforward than devising some scheme to reuse sessions or anything. +1 keeping a flag indicating if it is user-originated is a good idea.

sessions only last 1 week. do we have a plan for refresh tokens or anything for an active task?\

> Another question, @ioga, can we just remove the GPRC auth token thing?

what do you mean gRPC auth token thing? do you mean the `Grpc-Metadata-x-allocation-token` header?

rb-determined-ai:
> what do you mean gRPC auth token thing? do you mean the Grpc-Metadata-x-allocation-token header? if yes, depends on where you're putting the new tokens.

I think yes, but @mpkouznetsov can clarify.  My understanding is that the code now injects either the `Authentication: Bearer` header or the `Grpc-Metadata-x-allocation-token` header, but not both.  So I'm assuming that if we always have an `Authentication: Bearer` hearder (which I think we will now) that the `Grpc-Metadata-x-allocation-token` becomes totally useless.

Or are there places where only the `Grpc-Metadata-x-allocation-token` is valid?

stoksc:
That is correct, it's one or the other.

stoksc:
I just did a quick review, looks like yeah, we can remove that, `DET_SESSION_TOKEN`, all the logic in the harness to wire them together and the logic in the master to consume it.

rb-determined-ai (approved):
My approval applies to the python code, but I'm not sure I'm the best candidate to review the master code (though the master code I looked at looks good to me).

master/internal/core_experiment.go
```diff
@@ -221,6 +221,11 @@ func (m *Master) patchExperiment(c echo.Context) (interface{}, error) {
 		agentUserGroup = &m.config.Security.DefaultTask
 	}
 
+	ownerFullUser, err := m.db.UserByID(*dbExp.OwnerID)
```
stoksc, line 224: just a note, i have no idea how this could be nil.. maybe in an ancient schema

stoksc (approved):
go side lgtm

PR #3663, opened 2022-02-25
refactor: pull out intermediary resources representation for RM

stoksc:
## Description
This change introduces an intermediary representation for container, pods or jobs called "Resources" which are an arbitrary-ish clumps of resources from a resource manager. This is the first change in a series to allow us to integrate with slurm more naturally. This reasons for this change follow.
1. When we converse with resource managers that do not speak terms of containers the rename portion of this change is required else the code reads like nonsense.
2. It provides an easier path to LCD shared resource manager features that does not involve refactor the internals of Kubernetes or agents resource managers.
3. It provides a path to put non-shared features behind clean feature flags.

This approach takes our current interface and makes it flexible enough to be generalized but as of now is mostly a noop, so to enable can work with more resource managers without too much fuss. There are a few other approaches that were not chosen. One would be to force each resource manager we support to be compatible with our current interface through a serious of shims/scripts on the resource manager side of the interface. This isn't really ideal, mostly because it isn't clear it's even possible, much less sustainable with all resource managers we may support. Another would be to move the allocation actor into resource managers and draw the line where the RM interface starts at a different place. This is promising, but then existing features have to be refactor/re-implemented at a different layer and is generally a lot more work. Also the current approach is still useful with that approach and may transpose to it later if it still seems appealing.

## Test Plan
- [x] run existing tests, don't break the RMs, this should be a straight refactor
- [x] run one of each job type while: cancelling, killing, pausing it, disabling/draining/enabling agents out from under it.

## Checklist


master/internal/prom/prom.go
```diff
@@ -104,21 +104,33 @@ func AssociateContainerRuntimeID(cID cproto.ID, dcID string) {
...
-func AddAllocationReservation(summary sproto.ContainerSummary,
-	containerStarted *sproto.TaskContainerStarted) {
-	AssociateAllocationContainer(summary.AllocationID, summary.ID)
-	AssociateContainerRuntimeID(summary.ID, containerStarted.NativeReservationID)
-	for _, d := range summary.Devices {
-		AssociateContainerGPU(summary.ID, d)
+// AddAllocationResources associates allocation and container and container and GPUs.
+func AddAllocationResources(summary sproto.ResourcesSummary,
+	containerStarted *sproto.ResourcesStarted) {
+	if summary.ContainerID == nil {
```
stoksc, line 110: this is mostly a todo to figure out for slurm, but will just be a noop until it is thought out.

master/internal/resourcemanagers/kubernetes/spec.go
```diff
@@ -92,6 +92,7 @@ func (p *pod) configureEnvVars(
 	envVarsMap["DET_MASTER_ADDR"] = p.masterIP
 	envVarsMap["DET_MASTER_PORT"] = fmt.Sprintf("%d", p.masterPort)
 	envVarsMap["DET_AGENT_ID"] = "k8agent"
+	envVarsMap["DET_RESOURCES_ID"] = p.taskSpec.ContainerID
 	envVarsMap["DET_CONTAINER_ID"] = p.taskSpec.ContainerID
```
stoksc, line 96: these are just the same, until one goes away, shortly

harness/determined/exec/prep_container.py
```diff
@@ -55,14 +55,14 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
 
 
 def do_rendezvous(info: det.ClusterInfo, cert: certs.Cert) -> None:
-    # Even though container_id is not part of the ClusterInfo API, we still depend on it in all
+    # Even though resources_id is not part of the ClusterInfo API, we still depend on it in all
```
rcorujo, line 58: Was this the method where we were supposed to send back the IP address of compute node's high speed interconnect to the master?  If so, wasn't the "get" going to be changed to a "post" or did I misunderstand?

stoksc: new API to use is in this PR https://github.com/determined-ai/determined/pull/3694

harness/determined/cli/task.py
```diff
@@ -10,13 +10,14 @@
...
 def render_tasks(args: Namespace, tasks: Dict[str, Dict[str, Any]]) -> None:
     def agent_info(t: Dict[str, Any]) -> Union[str, List[str]]:
-        containers = t.get("containers", [])
-        if not containers:
+        resources = t.get("resources", [])
+        if not resources:
             return "unassigned"
-        if len(containers) == 1:
-            agent = containers[0]["agent"]  # type: str
+        agents = [a for r in resources for a in r["agent_devices"].keys()]
```
ioga, line 16:
```suggestion
        agents = [a for r in resources for a in r["agent_devices"]]
```

master/internal/sproto/task.go
```diff
@@ -121,25 +123,45 @@ type (
...
+)
+
+// ResourcesSummary provides a summary of the resources comprising what we know at the time the
+// allocation is granted, but for k8s it is granted before being scheduled so it isn't really much
+// and `agent_devices` are missing for k8s.
+type ResourcesSummary struct {
+	ResourcesID   ResourcesID                `json:"resources_id"`
+	ResourcesType ResourcesType              `json:"resources_type"`
+	AllocationID  model.AllocationID         `json:"allocation_id"`
+	AgentDevices  map[string][]device.Device `json:"agent_devices"`
```
ioga, line 154: perhaps `type AgentID string` and
```suggestion
	AgentDevices  map[AgentID][]device.Device `json:"agent_devices"`
```

master/internal/task/allocation_test.go
```diff
@@ -83,45 +82,41 @@ func TestAllocation(t *testing.T) {
...
+				containerStateChanged.ResourcesState = sproto.Starting
 				require.NoError(t, system.Ask(self, containerStateChanged).Error())
-				containerStateChanged.Container.State = cproto.Running
-				containerStateChanged.ContainerStarted = &sproto.TaskContainerStarted{
+				containerStateChanged.ResourcesState = sproto.Running
+				containerStateChanged.ResourcesStarted = &sproto.ResourcesStarted{
 					Addresses: []cproto.Address{
 						{
-							ContainerIP:   "0.0.0.0",
+							ContainerIP:   "172.0.0.3",
```
ioga, line 109: why `.3`?

ioga (approved):
this is a big PR so can't say I fully comprehend all of it but looks nice 👍

master/internal/task/allocation_test.go
stoksc: :shrug: just the first number i typed, is a test, idk why i changed this

PR #3664, opened 2022-02-25
chore: realign task log timestamps

stoksc:
## Description
Previously task log timestamps were aligned as a side effect of being
printed in seconds.  Now they have more precision but the omission of
trailing zeroes causes them to be misaligned and hard to read.  This
change that ensures trailing zeroes are printed.

## Test Plan
- [x] run an experiment, observe well-aligned logs

## Checklist


dzhu (approved):
Thanks for taking care of that rather silly papercut!

PR #3668, opened 2022-02-28
feat: Job Queue persist position

eecsliu:
## Description
Changes the job queue position to use a decimal instead of a float, greatly increasing the number of insertions. This allows us to skip rebalancing and save job positions from the experiment or command actor instead of from the resource pool.

## Checklist


master/internal/resourcemanagers/resource_pool.go
```diff
@@ -410,7 +415,15 @@ func (rp *ResourcePool) receiveJobQueueMsg(ctx *actor.Context) error {
...
 	case job.MoveJob:
-		ctx.Respond(rp.moveJob(msg))
+		response, err := rp.moveJob(msg)
+		ctx.Respond(err)
+		if err == nil {
+			addr, ok := rp.IDToGroupActor[response.JobID]
+			if ok {
+				ctx.Tell(addr, response)
+			}
+		}
```
stoksc, line 425: nit https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
```suggestion
		if err != nil {
			ctx.Respond(err)
			return nil
		}

		addr, ok := rp.IDToGroupActor[response.JobID]
		if !ok {
			return nil
		}
			
		ctx.Tell(addr, response)
```

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -410,7 +415,15 @@ func (rp *ResourcePool) receiveJobQueueMsg(ctx *actor.Context) error {
 	case job.GetJobQ:
 		ctx.Respond(rp.scheduler.JobQInfo(rp))
 	case job.MoveJob:
-		ctx.Respond(rp.moveJob(msg))
+		response, err := rp.moveJob(msg)
+		ctx.Respond(err)
```
stoksc, line 419: if the ask'er ever didn't expect a response, this panics (why `ctx.ExpectingResponse()` exists)

master/internal/resourcemanagers/resource_pool.go
stoksc: if the ask'er is expecting a response `ctx.Respond(nil)` is implicit.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,24 +380,30 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
-	fmt.Println(rp.config.PoolName, "moveJob: ", msg)
+func (rp *ResourcePool) moveJob(msg job.MoveJob) (job.RegisterJobPosition, error) {
 	rp.queuePositions[job.TailAnchor] = initalizeQueuePosition(time.Now())
-	newPos, rebalance, err := computeNewJobPos(msg, rp.queuePositions)
+	newPos, err := computeNewJobPos(msg, rp.queuePositions)
 	if err != nil {
-		return err
+		return job.RegisterJobPosition{}, err
 	}
 	rp.queuePositions[msg.ID] = newPos
```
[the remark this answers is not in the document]
stoksc: same for the recover flow

master/internal/resourcemanagers/job.go
```diff
@@ -69,102 +70,57 @@ func assignmentIsScheduled(allocatedResources *sproto.ResourcesAllocated) bool {
...
+	bigPos := decimal.Max(qPos1, qPos2)
+	if qPos3.GreaterThan(smallPos) && qPos3.LessThan(bigPos) {
+		return decimal.NewFromInt(0), nil // no op. Job is already in the correct position.
 	}
 
-	newPos := (qPos1 + qPos2) / 2
+	newPos := decimal.Avg(qPos1, qPos2)
+
+	if newPos.Equal(qPos1) || newPos.Equal(qPos2) {
+		return decimal.NewFromInt(0), fmt.Errorf("unable to compute a new job position for job %s",
```
stoksc, line 114: what's the official math of what it would take to hit this?

master/internal/resourcemanagers/job.go
```diff
@@ -69,102 +70,57 @@ func assignmentIsScheduled(allocatedResources *sproto.ResourcesAllocated) bool {
...
 
 	// check if qPos3 is between qPos1 and qPos2
-	smallPos := math.Min(qPos1, qPos2)
-	bigPos := math.Max(qPos1, qPos2)
-	if qPos3 > smallPos && qPos3 < bigPos {
-		return 0, false, nil // no op. Job is already in the correct position.
+	smallPos := decimal.Min(qPos1, qPos2)
+	bigPos := decimal.Max(qPos1, qPos2)
+	if qPos3.GreaterThan(smallPos) && qPos3.LessThan(bigPos) {
+		return decimal.NewFromInt(0), nil // no op. Job is already in the correct position.
```
stoksc, line 108: shouldn't this return `qPos3` back. if there is no error it looks like we set it to pos `decimal.NewFromInt(0)`

stoksc (commented):
overall design for this pr looks good, just minor comments.

master/internal/resourcemanagers/job.go
```diff
@@ -69,102 +70,57 @@ func assignmentIsScheduled(allocatedResources *sproto.ResourcesAllocated) bool {
...
-	return newPos, false, nil
+	return newPos, nil
 }
 
-func initalizeQueuePosition(aTime time.Time) float64 {
-	// we could shift this back and forth to give us more more.
-	return float64(aTime.UnixMicro())
+func initalizeQueuePosition(aTime time.Time) decimal.Decimal {
+	// we could add exponent to give us more insertions if needed.
+	return decimal.New(aTime.UnixMicro(), job.DecimalExp)
```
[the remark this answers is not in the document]
eecsliu: It was an arbitrary choice. The way to add digits is with the exponent, and the number is limited to something like 16 decimal points. I picked 1000 because I did a bit of testing and it seemed to provide something like 3600 insertions before loss of precision. Also I didn't want to get too large of an exponent because it takes longer to calculate things.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,24 +380,30 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
-	// condiontally check if rebalancing is needed.
-	if rebalance {
-		adjustPriorities(&rp.queuePositions)
-	}
 
-	if err := rp.persist(); err != nil {
+	// actors.NotifyAfter(ctx, 0, schedulerTick{}) ?
```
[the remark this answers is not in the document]
eecsliu: I think this was Hamid's comment (I forked off of his initial work). I forgot to ask about this, but it might not be a great idea, especially if we're handling bulk operations.

master/internal/resourcemanagers/job.go
eecsliu: I think the formula is log<sub>2</sub>10<sup>(# of digits)</sup> since we're just taking the average of two numbers repeatedly. Not sure if job timestamps can be less than several milliseconds apart - if it's valid to assume they are at least 1 ms apart, then the difference between two positions is at least 10<sup>1016</sup>. 

That should work out to log<sub>2</sub>10<sup>1016</sup> = 3375 insertions before losing precision.

master/internal/resourcemanagers/job.go
eecsliu: ah great catch

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,36 +381,136 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
-	fmt.Println(rp.config.PoolName, "moveJob: ", msg)
-	rp.queuePositions[job.TailAnchor] = initalizeQueuePosition(time.Now())
-	newPos, rebalance, err := computeNewJobPos(msg, rp.queuePositions)
-	if err != nil {
-		return err
+func (rp *ResourcePool) moveJob(
+	ctx *actor.Context, jobID model.JobID, anchorID model.JobID, aheadOf bool,
+) error {
+	if rp.config.Scheduler.GetType() != priorityScheduling {
+		return nil
```
stoksc, line 388: should this return something like `ErrUnsupported` or something, rather than just doing nothing?

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,36 +381,136 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
-	if rebalance {
-		adjustPriorities(&rp.queuePositions)
+
+	if _, ok := rp.queuePositions[anchorID]; !ok {
+		return job.ErrJobNotFound(anchorID)
+	}
+
+	prioChange, secondAnchor, anchorPriority := rp.findAnchor(jobID, anchorID, aheadOf)
+
+	check.Panic(check.True(secondAnchor != ""))
```
stoksc, line 410: just return an error? panicking an RM is very severe

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,36 +381,136 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
+			rp.queuePositions[secondAnchor],
+			aheadOf,
+		) {
+			return nil
+		}
 	}
 
-	if err := rp.persist(); err != nil {
+	msg, err := rp.queuePositions.SetJobPosition(jobID, anchorID, secondAnchor)
+
```
stoksc, line 432: nit newline

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,36 +381,136 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
+	prioChange, secondAnchor, anchorPriority := rp.findAnchor(jobID, anchorID, aheadOf)
+
+	check.Panic(check.True(secondAnchor != ""))
+
+	if secondAnchor == jobID {
+		return nil
+	}
+
+	if prioChange {
+		resp := ctx.Self().System().AskAt(job.JobsActorAddr, job.SetJobPriority{ID: jobID, Priority: anchorPriority})
```
stoksc, line 417: bit strange this goes through the jobs actor rather than just directly to the job, but i get why, but i also think it could not

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -381,36 +381,136 @@ func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
...
+	}
+
+	if aheadOf {
+		if anchorIdx == 0 {
+			secondAnchor = job.HeadAnchor
+		} else {
+			secondAnchor = sortedReqs[anchorIdx-1].JobID
+			if rp.queuePositions[secondAnchor].GreaterThanOrEqual(anchorPos) {
+				secondAnchor = job.HeadAnchor
+			}
```
stoksc, line 468: maybe i'm just dense, i dont get how this case could ever happen if we're itering sortedReqs and successfully pulling the req prior to the anchor. same for the other case below. what am i missing?

master/internal/resourcemanagers/resource_pool.go
eecsliu: hmm Ill look into it, but yeah the jobs actor is supposed to act almost like an RM for jobs

master/internal/resourcemanagers/resource_pool.go
eecsliu: This is supposed to cover the case where the priority of the anchors might be different in addition to the queue positions. For example, if the first anchor has priority 42 & position 10, but the second anchor has priority 41 & position 20. Then the second anchor would be sorted before the first anchor, even though it's position is higher (smaller is better).

master/internal/resourcemanagers/job.go
```diff
@@ -69,102 +70,83 @@ func assignmentIsScheduled(allocatedResources *sproto.ResourcesAllocated) bool {
...
+}
+
+func (j jobSortState) RecoverJobPosition(jobID model.JobID, positionStr string) error {
+	if positionStr == "" {
+		return nil
+	}
+	position, err := decimal.NewFromString(positionStr)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
stoksc: i guess what i'm saying is `type Job struct` should have a decimal not string.

master/internal/resourcemanagers/resource_pool.go
stoksc: this is more just commentary, for now. but i think the rm could just ask the job directly, personally

stoksc: like, less complexity is always better, and i think the jobs actor could be basically just a thin proxy to help apis. if you think of the tell/ask graph, this is a totally new line whereas we could just use old paths.

master/internal/resourcemanagers/resource_pool.go
stoksc: oh. why not just if `rp.groups[secondAnchorReq.Group].priority < &anchorPriority` then. what if the first anchor is priority 42 and position 20 but the second is priority 41 and position 10. seems we should still use the head anchor not the second anchor we've found?

stoksc:
prob needs automated tests. wydt @eecsliu ?

master/internal/resourcemanagers/resource_pool.go
stoksc: eh, i think this is blocking for me. why ask job.JobsActorAddr SetJobPriority so it can ask jobActor SetGroupPriority rather than just asking groupAddr SetGroupPriority?

stoksc: and then SetJobPriority can be removed entirely

master/internal/resourcemanagers/resource_pool.go
eecsliu: Hmm I think using decimal comparisons is the least ugly way to do it. Since the list was already sorted, we know that the second anchor is right next to the first anchor. Since we promote or demote the priority of the target job to be the same as the anchor, we just need a queue position relative to the first anchor too. Ill think about this a bit more

eecsliu:
Yeah I think that's a good idea for some sanity checks. I have some unit tests incoming, and I can see about adding some e2e tests

master/internal/resourcemanagers/resource_pool.go
eecsliu: The asker (job actor) should be performing an ask and checks for an error

master/internal/resourcemanagers/resource_pool.go
stoksc: >  what if the first anchor is priority 42 and position 20 but the second is priority 41 and position 10. seems we should still use the head anchor not the second anchor we've found?

sorry for being dense. what about this case then? is there some invariant upheld elsewhere that prevents this?

e2e_tests/tests/cluster/test_job_queue.py
```diff
@@ -0,0 +1,84 @@
...
+    new_weight = jobs.get_job_weight(ordered_ids[1])
+    assert new_weight == "10"
+
+    return
+
+
+def get_raw_data() -> Tuple[List[Dict[str, str]], List[str]]:
+    data = []
+    ordered_ids = []
+    output = subprocess.check_output(["det", "job", "list"]).decode("utf-8")
```
stoksc, line 44: nit, would say just to use python bindings for the api instead, but it doesn't need to block this.

master/internal/command/command.go
```diff
@@ -327,20 +327,29 @@ func (c *command) Receive(ctx *actor.Context) error {
...
-func (c *command) setPriority(ctx *actor.Context, priority int) error {
+func (c *command) setPriority(ctx *actor.Context, priority int, handler *actor.Ref) error {
 	if sproto.UseK8sRM(ctx.Self().System()) {
 		return fmt.Errorf("setting priority for job type %s in kubernetes is not supported",
 			c.jobType)
 	}
 	c.Config.Resources.Priority = &priority
+	if handler != nil {
+		return nil
+	}
```
stoksc, line 352: this is very confusing to me. if the handler is not nil, we return and don't actually set prio? but then it seems like we don't use handler at all to send the message (and i dont understand why we dont send it in all cases)

master/internal/db/postgres_jobs.go
```diff
@@ -35,14 +38,14 @@ VALUES (:job_id, :job_type, :owner_id, :q_position)
 	return nil
 }
 
-// UpdateJob propagates the new queue position to the job.
-func (db *PgDB) UpdateJob(job *model.Job) error {
-	if job.JobID.String() == "" {
+// UpdateJobPosition propagates the new queue position to the job.
+func (db *PgDB) UpdateJobPosition(jobID fmt.Stringer, position decimal.Decimal) error {
```
stoksc, line 42: `fmt.Stringer` -> `model.JobID` no matter what `interfacer` said lol (maybe you called `.String()` but that wouldn't be needed)

master/internal/experiment.go
```diff
@@ -606,6 +625,9 @@ func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
 		e.Config.SetResources(resources)
 		return errors.Wrapf(err, "setting experiment %d priority", e.ID)
 	}
+	if handler != nil {
```
stoksc, line 628: i also dont understand this. maybe it should be like `informRMOfChange` rather than a nil-able handler that it unused?

stoksc (commented):
meta note that i message you, why not use `math/big`'s `big.Rat` rather than `decimal.Decimal`?

master/internal/command/command.go
eecsliu: mb this was a late night debugging hotfix that I forgot to mark. The problem is just that you can enter a loop and hang in certain cases. Most of the time, it's the group handler sending this message to the RP, but in the case of moving a job, it becomes RP -> group handler -> RP. So the fix was just to check for a handler. You're right that it really should be a check for what the handler is and where it comes from.

stoksc:
@eecsliu no reason to stress about it, was just a thought

eecsliu:
> meta note that i message you, why not use `math/big`'s `big.Rat` rather than `decimal.Decimal`?

Did a bit of investigation - it is a potential replacement, but it is less friendly to use and would take additional work to write utility functions and verify its behavior. I probably won't change it from decimal just yet.

master/internal/command/command.go
stoksc: this is still funny, were we going to change this @eecsliu ? just checking.

stoksc: why are commands not handling `NotifyRMPriorityChange`?

master/internal/experiment.go
```diff
@@ -588,7 +610,7 @@ func checkpointFromTrialIDOrUUID(
 	return checkpoint, nil
 }
 
-func (e *experiment) setPriority(ctx *actor.Context, priority *int) error {
+func (e *experiment) setPriority(ctx *actor.Context, priority *int, forward bool) error {
```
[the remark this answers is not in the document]
eecsliu: yeah exactly - I thought this was a bit cleaner than having two separate handler.

master/internal/resourcemanagers/priority.go
```diff
@@ -315,12 +317,12 @@ func comparePositions(a, b *sproto.AllocateRequest, jobPositions jobSortState) i
 	switch {
 	case aPosition == bPosition:
 		return aReqComparator(a, b) * -1 // CHECK
```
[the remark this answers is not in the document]
eecsliu: I'll double check with Hamid. I think this is an older comment, but this works.

master/internal/command/command.go
```diff
@@ -126,7 +126,7 @@ func (c *command) Receive(ctx *actor.Context) error {
 
 		priority := c.Config.Resources.Priority
 		if priority != nil {
-			if err := c.setPriority(ctx, *priority); err != nil {
+			if err := c.setPriority(ctx, *priority, true); err != nil {
```
stoksc, line 129: erroring the actor here is right?

stoksc (approved):
lgtm, great work. (one last final comment, just make sure that's the intended behavior)

master/internal/command/command.go
eecsliu: yeah there is an error that should be raised in the event the RM is kubernetes

stoksc: chatted offline, i think this would mean you couldn't launch any commands with priority, so maybe we tweak this a bit to allow at least the initial set priority.

PR #3670, opened 2022-03-01
Use modal context

hkang1:
## Description

PROBLEM:
The context that is provided through a Provider that is wrapped at the App level is not accessible via dynamically created modals in Antd. The modal is created via `ReactDom.render`, as such causing the context to be different.

SOLUTION:
Antd provides a `contextHolder` via `Modal.useModal()` that allows you place a React element into the component or hook you are looking to pin to to make the context of that "location" accessible.

## Checklist

### DIFF
--- a/webui/react/src/components/Avatar.tsx
+++ b/webui/react/src/components/Avatar.tsx
@@ -1,6 +1,7 @@
 import { Tooltip } from 'antd';
 import React from 'react';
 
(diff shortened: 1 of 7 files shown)


hkang1:
Closing, as this PR was serving purely as an example of how to bring context into a `useModal` hook.

PR #3672, opened 2022-03-01
test: add tests for get experiment

stoksc:
## Description
In the spirit of reliability, this change adds tests to `get_experiments.sql` before I change it drastically, for the sake of performance.

## Test Plan
- [x] is a test

## Checklist

### DIFF
--- a/master/internal/db/postgres_experiments_intg_test.go
+++ b/master/internal/db/postgres_experiments_intg_test.go
@@ -6,15 +6,270 @@ package db
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/determined-ai/determined/master/pkg/etc"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/master/pkg/ptrs"
+	"github.com/determined-ai/determined/master/pkg/schemas"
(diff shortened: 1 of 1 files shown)


master/internal/db/postgres_experiments_intg_test.go
```diff
@@ -6,16 +6,326 @@ package db
...
+				return exps
+			}(),
+		},
+		{
+			name: "fitler by state",
+			n:    3,
+			exps: func() []model.Experiment {
+				var exps []model.Experiment
+				for i := 0; i < 3; i++ {
+					exps = append(exps, model.Experiment{
```
dzhu, line 84: I think the tests would be more convincing if they created both experiments that did match the filter and ones that didn't.

master/internal/db/postgres_experiments_intg_test.go
```diff
@@ -6,16 +6,326 @@ package db
...
+						StartTime:            time.Now().Add(-time.Hour),
+						OwnerID:              &user.ID,
+						Username:             user.Username,
+					})
+				}
+				return exps
+			}(),
+		},
+		{
+			name: "fitler by state",
```
dzhu, line 79: Don't forget to fix the spelling before landing. :wink:

dzhu (approved):

master/internal/db/postgres_experiments_intg_test.go
stoksc: technically it does this already, that's why it starts with unfiltered. then the all filter by the same fake user for all the tests.

PR #3677, opened 2022-03-01
fix: GET /api/v1/trial/:id/logs/fields should return all filters

stoksc:
## Description
Task logs now get written to a new table, and the old trial logs endpoint merges the results of reading both tables, but this merging was missed for `GET /api/v1/trial/logs/fields`. This change adds this functionality.

## Test Plan
- [x] figure out why the automated test didn't fail (i'm pretty sure it should have)
- [x] add new assert so tests would fail in this case

## Checklist


gt2345 (approved):
LGTM

PR #3681, opened 2022-03-02
fix: select filter input width [DET-6777]

hkang1:
## Description

Taking this [previous work](https://github.com/determined-ai/determined/pull/3618/files) and apply it on all `SelectFilter`. The problem of the input filter not expanding all the way. Taking one step further on the previous work, the expanded input field would spill out of the container. Apply overflow appropriately so the text is contained behind the dropdown arrow.

BEFORE:
Screen Shot 2022-03-01 at 8 42 25 PM

AFTER:
Screen Shot 2022-03-01 at 8 42 39 PM

Screen Shot 2022-03-01 at 8 50 03 PM

## Test Plan

Test the select filter input expands the full dropdown container width instead of the previous 4px width. Some common places to look for:
- [x] metric filter in trial detail
- [x] hyper parameters and metric column picker within the trial comparison modal
- [x] task and trial log filters

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer/LogViewerFilters.tsx
+++ b/webui/react/src/components/LogViewer/LogViewerFilters.tsx
@@ -50,7 +50,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
     };
   }, [ options ]);
 
-  const show = useMemo(() => {
+  const moreThanOne = useMemo(() => {
     return Object.keys(selectOptions).reduce((acc, key) => {
       const filterKey = key as keyof Filters;
       const options = selectOptions[filterKey];
@@ -84,7 +84,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
   return (
     <>
       <Space>
-        {show.allocationIds && (
+        {moreThanOne.allocationIds && (
           <MultiSelect
             itemName={LABELS.allocationIds}
             value={values.allocationIds}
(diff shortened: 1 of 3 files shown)


webui/react/src/components/LogViewer/LogViewerFilters.tsx
```diff
@@ -24,48 +25,72 @@ export interface Filters {
...
+    onChange?.({ ...values, [key]: (value as Array<string>).map(item => caster(item)) });
   }, [ onChange, values ]);
 
   const handleReset = useCallback(() => onReset?.(), [ onReset ]);
 
   return (
     <>
       <Space>
-        {selectOptions?.allocationIds?.length !== 0 && (
+        {show.allocationIds && (
```
trentwatt, line 77: would it make sense to do:
image

instead for these? or maybe rename `show` to `moreThanOne` ?

trentwatt: That way we can get rid of show. Or are there rendering considerations??

trentwatt (approved):
Looks good

webui/react/src/components/LogViewer/LogViewerFilters.tsx
hkang1: 👍 

re-rendering wise it's the same because they both depend on `selectOptions` changes.

I think your suggestion makes sense, and it'll reduce the line count by a little.

hkang1: actually thought about it a bit more and the current method centralizes the show logic into one spot, so I'll follow your second suggestion instead ;)

PR #3683, opened 2022-03-02
feat: context menu row background

trentwatt:
## Description

feat: context menu for experiment list [DET-6439]

Added context menu for experiment list, that allows users to perform experiment actions by right clicking.

Context Menu can be activated when row is hovered but a link or text cell is not (decided on a per-column basis). 

Context Menu is also disabled when rows are selected with the checkboxes, since the assumption then is that the user would use the batch actions instead. Might consider future possibility to have batch actions on selected rows dispatched from context menu.

Individual row takes on a blue color when the context menu is either able to be opened, or currently open. 

## Test Plan

Terminology: 
- "context menu is active" means "the row is blue and right clicking brings up the custom context menu"
- "context menu is inactive" means "the row is grey and right clicking brings up the default browser context menu"

Basic:
- context menu is active when hovering a row
- context menu is inactive when hovering over id, name, description, etc
- archiving experiment from context menu results in the experiment disappearing (there is a separate issue with deleting experiments)

With rows selected:
- checkbox some rows
- those rows will become blue
- hover over a row that is not selected -> context menu is inactive

## Commentary (optional)

Tried in general to make experiment specific changes in `ExperimentList.tsx` and non-experiment specific changes in `TaskActionDropdown.tsx` and `ResponsiveTable.tsx`, and make it so where it would be straightforward to add context menu functionality to other tables as well.

- had to (?) `useContext` to pass state from row to cell since antd `Table` only provides separate access to rows and cells.
- passed `DropdownComponent` (a curried version of `TaskActionDropdown`) as a prop to avoid having to pass experiment specific arguments to `ResponsiveTable`...  is there a better way?

## Checklist

[DET-6439]: https://determinedai.atlassian.net/browse/DET-6439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/ResponsiveTable.module.scss
+++ b/webui/react/src/components/ResponsiveTable.module.scss
@@ -3,3 +3,7 @@
     color: var(--theme-colors-action-normal);
   }
 }
+
+.rightClickableCellWrapper {
+  display: inline-block;
+}
--- a/webui/react/src/components/ResponsiveTable.module.scss
+++ b/webui/react/src/components/ResponsiveTable.module.scss
@@ -4,5 +4,5 @@
   }
 }
 .rightClickableCellWrapper {
-  display: inline-block;
+  display: inline;
 }
--- a/webui/react/src/components/ResponsiveTable.module.scss
+++ b/webui/react/src/components/ResponsiveTable.module.scss
@@ -3,7 +3,6 @@
     color: var(--theme-colors-action-normal);
   }
 }
-
 .rightClickableCellWrapper {
   display: inline-block;
 }
--- a/webui/react/src/components/ResponsiveTable.tsx
+++ b/webui/react/src/components/ResponsiveTable.tsx
@@ -83,7 +83,17 @@ const ResponsiveTable: ResponsiveTable = ({ loading, scroll, ...props }) => {
   return (
     <div ref={tableRef}>
       <Spinner spinning={spinning}>
-        <Table scroll={tableScroll} tableLayout="auto" {...props} />
+        <Table
+          scroll={tableScroll}
+          tableLayout="auto"
+          onRow={(record, rowIndex) => ({
+            onContextMenu: e => {
+              e.preventDefault()
+              console.log("row right click")
+              // toggle visibility of custom action menu
+            }
+            })}
+          {...props} />
       </Spinner>
     </div>
   );
--- a/webui/react/src/components/ResponsiveTable.tsx
+++ b/webui/react/src/components/ResponsiveTable.tsx
@@ -8,6 +8,8 @@ import useResize from 'hooks/useResize';
 
 import Spinner from './Spinner';
 
+import css from './ResponsiveTable.module.scss'
+
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
 type Comparable = any;
 
@@ -95,7 +97,7 @@ const Cell = ({ children, isCellRightClickable, ...props }: CellProps) => {
   if (!isCellRightClickable) return <td {...props}>{children}</td>;
   return (
     <td {...props}>
-      <div {...rightClickableCellProps}>
+      <div className={css.rightClickableCellWrapper} {...rightClickableCellProps}>
         {children}
       </div>
     </td>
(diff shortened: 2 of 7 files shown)


webui/react/src/components/ResponsiveTable.module.scss
```diff
@@ -3,3 +3,6 @@
     color: var(--theme-colors-action-normal);
   }
 }
+.rightClickableCellWrapper {
+  display: inline-block;
```
EmilyBonar, line 7: chore: I think this should be `display: inline`. inline-block makes it so even if you're not hovering over text, you sometimes can't open the context menu, while inline fits the amount of text present.

webui/react/src/components/ResponsiveTable.tsx
```diff
@@ -18,8 +19,89 @@ interface Settings {
...
+    return <tr className={className} {...props}>{children}</tr>;
+  }
+
+  const rightClickableCellProps = {
+    onContextMenu: (e : React.MouseEvent) => e.stopPropagation(),
+    onMouseEnter: () => setRightClickableCellHovered(true),
+    onMouseLeave: () => setRightClickableCellHovered(false),
+  };
+
+  const rowContextMenuActive =
```
EmilyBonar, line 74: nit: maybe it's just me, but this seems like a misleading variable name. `rowContextMenuActive` = true sounds like the context menu is currently open, not that it's openable.

webui/react/src/components/TaskActionDropdown.tsx
```diff
@@ -25,14 +25,22 @@ import css from './ActionDropdown.module.scss';
 import Link from './Link';
 
 interface Props {
+  children?: React.ReactNode;
```
EmilyBonar, line 28: nit: This works, but you could instead do `PropsWithChildren<Props>` instead of including children in Props

webui/react/src/components/TaskActionDropdown.tsx
```diff
@@ -159,7 +167,9 @@ const TaskActionDropdown: React.FC<Props> = ({ task, onComplete, curUser }: Prop
   }
 
   if (menuItems.length === 0) {
-    return (
+    return children ? (
+      <>{children}</>
```
EmilyBonar, line 171: nit: i don't think wrapping children in a fragment here is necessary

EmilyBonar (approved):
Overall this looks good, minus a few nits!

webui/react/src/components/TaskActionDropdown.tsx
trentwatt: The fragment-wrapping was to placate `tsc`,  which before was saying:
```
Type '({ task, onComplete, curUser, onVisibleChange, children, }: PropsWithChildren<Props>) => {}' 
is not assignable to type 'FC<Props>'.
  Type '{}' is missing the following properties from type 'ReactElement<any, any>': 
type, props, key  TS2322
```

Looks like the below is also allowed. Is this better?
```typescript
return (children as JSX.Element) ??
  <div className={css.base} title="No actions available" onClick={stopPropagation}>
   <button disabled>
      <Icon name="overflow-vertical" />
    </button>
  </div>
```

webui/react/src/components/TaskActionDropdown.tsx
trentwatt: will do!

webui/react/src/components/ResponsiveTable.tsx
trentwatt: how about `rowContextMenuTriggerableOrOpen` ? Not too unwieldy?

webui/react/src/components/ResponsiveTable.module.scss
trentwatt: Great thanks, will make that fix!

webui/react/src/components/TaskActionDropdown.tsx
EmilyBonar: Yeah, I think I like that way better.

webui/react/src/components/ResponsiveTable.tsx
EmilyBonar: That is pretty long, but should be fine.

PR #3686, opened 2022-03-03
feat: job queue write batch cli

eecsliu:
## Description
Allows for job queue modification in batches using `det job update-batch`. Also refactors code related to the detection of scheduler type.

## Checklist

### DIFF
--- a/harness/determined/cli/job.py
+++ b/harness/determined/cli/job.py
@@ -1,5 +1,5 @@
 import json
-from argparse import Namespace
+from argparse import Namespace, ONE_OR_MORE
 from datetime import datetime
 from typing import Any, List
 
@@ -99,6 +99,43 @@ def update(args: Namespace) -> None:
         setup_session(args), body=bindings.v1UpdateJobQueueRequest([update])
     )
 
+def _single_update(job_id, session, priority=None, weight=None, resource_pool=None, behind_of=None, ahead_of=None) -> None:
+    update = bindings.v1QueueControl(
+        jobId=job_id,
+        priority=priority,
+        weight=weight,
+        resourcePool=resource_pool,
+        behindOf=behind_of,
+        aheadOf=ahead_of,
+    )
+    bindings.post_UpdateJobQueue(
+        session, body=bindings.v1UpdateJobQueueRequest([update])
+    )
+    print("job_id:", job_id)
+    print("priority:", priority)
+    print("weight:", weight)
+    print("resource_pool:", resource_pool)
+    print("behind_of:", behind_of)
+    print("ahead_of:", ahead_of)
+    return
+
+
+@authentication.required
+def process_updates(args: Namespace) -> None:
+    session = setup_session(args)
+    for arg in args.operation:
+        inputs = {"session": session}
+        values = arg.split(".")
+        if len(values) != 2:
+            raise ValueError("Incorrect format")
+        inputs["job_id"] = values[0]
+        operation = values[1].split("=")
+        if len(operation) != 2:
+            raise ValueError("Incorrect format")
+        inputs[operation[0]] = operation[1]
+        _single_update(**inputs)
+
+
 
 args_description = [
     Cmd(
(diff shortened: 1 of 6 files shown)


harness/determined/cli/job.py
```diff
@@ -100,6 +87,83 @@ def update(args: Namespace) -> None:
...
+                f"Please ensure they are formatted as <jobID>.<operation>=<value>."
+            )
+        inputs["job_id"] = values[0]
+        operation = values[1].split("=")
+        if len(operation) != 2:
+            raise ValueError(
+                f"Operation for job {values[0]} has invalid format. "
+                f"Please ensure it is formatted as <jobID>.<operation>=<value>."
+            )
+        inputs[operation[0]] = operation[1]
```
ioga, line 108: will it be nicer to parser this using a regexp?
i.e. `r'(?P<job_id>[0-9]+)\.(?P<operation>priority|weight|resource-pool|ahead-of|behind-of)=(?P<value>[\w]+)'`
(I have not tested this regexp)

ioga: on 2nd thought, with regexp you won't be to provide "nice" error messages as easy.

harness/determined/cli/job.py
```diff
@@ -100,6 +87,83 @@ def update(args: Namespace) -> None:
...
+            if pool["pool_name"] == resource_pool and pool["scheduler"]["type"] == "fair_share":
+                return False
+        return is_priority_rm(config)
+
+    except (KeyError, TypeError):
+        return is_priority_rm(config)
+    return True
+
+
+def replace_keys(operations: dict) -> None:
```
ioga, line 155: maybe
```suggestion
def validate_operation_args(operations: dict) -> None:
```

and have this function additionally check for validity of operation, printing a human-readable error when it's not right?

harness/determined/cli/job.py
```diff
@@ -167,6 +227,21 @@ def update(args: Namespace) -> None:
...
                 ],
             ),
+            Cmd(
+                "update-batch",
+                process_updates,
+                "batch update jobs",
+                [
+                    Arg(
+                        "operation",
+                        nargs=ONE_OR_MORE,
```
ioga, line 237: cool! I did not know about this. argparse docs don't even mention `ONE_OR_MORE` — they use `nargs="+"` directly.

harness/determined/cli/job.py
```diff
@@ -152,17 +137,32 @@ def check_is_priority(config: dict, resource_pool: str) -> bool:
...
+    operation = values[1].split("=")
+    if len(operation) != 2:
+        raise ValueError(
+            f"The operation for job {values[0]} has invalid format. "
+            f"Please ensure it is formatted as <operation>=<value>."
+        )
+    if operation[0] in replacements:
+        args[replacements[operation[0]]] = operation[1]
+    else:
+        args[operation[0]] = operation[1]
```
ioga, line 163:
```suggestion
    args[replacements.get(operation[0], operation[0])] = operation[1]
```

harness/determined/cli/job.py
```diff
@@ -152,17 +137,32 @@ def check_is_priority(config: dict, resource_pool: str) -> bool:
...
     }
-    for k, v in replacements.items():
-        if k in operations:
-            operations[v] = operations[k]
-            del operations[k]
+    args = {}
+    values = operation.split(".")
+    if len(values) != 2:
+        raise ValueError(
+            f"Invalid operation specified for job {values[0]}. "
```
ioga, line 150: this says invalid operation but it doesn't actually check for the validity of the operation name?

ioga (approved):

harness/determined/cli/job.py
```diff
@@ -100,6 +87,83 @@ def update(args: Namespace) -> None:
...
+        behindOf=behind_of,
+        aheadOf=ahead_of,
+    )
+    bindings.post_UpdateJobQueue(session, body=bindings.v1UpdateJobQueueRequest([update]))
+    return
+
+
+def is_priority_rm(config: dict) -> bool:
+    try:
+        if config["resource_manager"]["scheduler"]["type"] == "fair_share":
```
ioga, line 121: we have other schedulers besides `fair_share` e.g. round robin. so maybe it's better to check for priority specifically?

harness/determined/cli/job.py
```diff
@@ -100,6 +87,89 @@ def update(args: Namespace) -> None:
...
+    update = bindings.v1QueueControl(
+        jobId=job_id,
+        priority=int(priority) if priority != "" else None,
+        weight=int(weight) if weight != "" else None,
+        resourcePool=resource_pool if resource_pool != "" else None,
+        behindOf=behind_of if behind_of != "" else None,
+        aheadOf=ahead_of if ahead_of != "" else None,
+    )
+    bindings.post_UpdateJobQueue(session, body=bindings.v1UpdateJobQueueRequest([update]))
+    return
```
ioga, line 116: redundant return

harness/determined/cli/job.py
```diff
@@ -100,6 +87,89 @@ def update(args: Namespace) -> None:
...
+    session: Session,
+    priority: str = "",
+    weight: str = "",
+    resource_pool: str = "",
+    behind_of: str = "",
+    ahead_of: str = "",
+) -> None:
+    update = bindings.v1QueueControl(
+        jobId=job_id,
+        priority=int(priority) if priority != "" else None,
```
ioga, line 109: hmm perhaps just convert `""` -> `None` for all `operand[1]` in input validation instead of repeating it for all 5 arguments here?

harness/determined/cli/job.py
eecsliu: ....sorry I don't know how I didn't see this until AFTER merging. I'll fix this in a follow up PR.

harness/determined/cli/job.py
eecsliu: yeah I did a bunch of things to try and satisfy the checker. Another problem that comes up is that ``priority`` and ``weight`` needed to be converted to integers , but the linter complains if you try to assign ``None`` to a declared ``int`` type. So the most elegant way to do this for ``priority`` and ``weight`` is to clean up the args here. At that point, I thought maybe I should just keep it all consistent.

PR #3687, opened 2022-03-03
docs: restructure training debugging doc

gh-determined-ai:
## Checklist

### DIFF
--- a/docs/training-debug/index.txt
+++ b/docs/training-debug/index.txt
@@ -370,12 +370,12 @@ step only applies if you have multiple GPUs and want to use distributed training
 
       -  If you have four Determined agents running with four GPUs each, your ``slots_per_trial``
          could be ``1``, ``2``, ``3``, or ``4``, which fits on a single machine.
-      -  A ``slots_per_trial`` value of ``8``, ``12``, or ``16`` completely utilizes a number
-         of agent machines.
+      -  A ``slots_per_trial`` value of ``8``, ``12``, or ``16`` completely utilizes a number of
+         agent machines.
       -  A ``slots_per_trial`` value of ``5`` implies more than one agent but it is not a multiple
          of agent size so this is an invalid case.
-      -  A ``slots_per_trial`` value of ``32`` is too large for the cluster and is also an
-         invalid case.
+      -  A ``slots_per_trial`` value of ``32`` is too large for the cluster and is also an invalid
+         case.
 
       Ensure that there are no other notebooks, shells, or experiments on the cluster that might
       consume too many resources and prevent the experiment from starting.
--- a/docs/training-debug/index.txt
+++ b/docs/training-debug/index.txt
@@ -19,7 +19,7 @@ Determined model. Follow the nine steps as applicable to your environment:
 
 -  Model-related Issues
       -  `Step 1 - Verify that your code runs locally`_
-      -  `Step 2 - Verify that each Trial subclass method work locally`_
+      -  `Step 2 - Verify that each Trial subclass method works locally`_
       -  `Step 3 - Verify local test mode`_
 
 -  Docker- or Cluster-related Issues
@@ -53,7 +53,7 @@ Confirm that your code works as expected before continuing.
 .. _step2:
 
 **************************************************************
- Step 2 - Verify that each Trial subclass method work locally
+ Step 2 - Verify that each Trial subclass method works locally
 **************************************************************
 
 This step assumes you have a working local environment for training. If you typically run your code
--- a/docs/training-debug/index.txt
+++ b/docs/training-debug/index.txt
@@ -1,343 +1,395 @@
 .. _model-debug:
 
-###########
- Debugging
-###########
+##############
+ Debug Models
+##############
 
-This document aims to provide useful guidelines for debugging models with Determined. Hopefully, it
-will help you become a power user of Determined, but please don't hesitate to contact us on `Slack
-<https://join.slack.com/t/determined-community/shared_invite/zt-cnj7802v-KcVbaUrIzQOwmkmY7gP0Ew>`__
-if you get stuck!
+This document recommends how to use Determined to debug models depending on your environment.
 
-This document focuses on model debugging, not cluster debugging, so it is assumed that you have
-already successfully :ref:`installed Determined <install-cluster>`.
+Your code on a Determined cluster differs from typical training scripts in the following ways:
 
-Successfully running code on a Determined cluster differs from normal training scripts in the
-following ways:
+-  The code conforms to the Trial APIs as a subclass of the Determined ``Trial`` class, indirectly,
+   by using one of the concrete subclasses, such as :class:`~determined.pytorch.PyTorchTrial`.
+-  The code runs in a Docker container on another machine.
+-  Your model can run many times in a hyperparameter search.
+-  Your model can be distributed across multiple GPUs or machines.
 
--  Your code will conform to the Trial APIs by being organized into a subclass of Determined's
-   ``Trial`` class (indirectly through one of its concrete subclasses, such as
-   :class:`~determined.pytorch.PyTorchTrial`).
+These debugging steps introduce code changes incrementally, working toward a fully functioning
+Determined model. Follow the nine steps as applicable to your environment:
 
--  The code will run in a Docker container on another machine.
+-  Model-related Issues
+      -  `Step 1 - Verify that your code runs locally`_
+      -  `Step 2 - Verify that each Trial subclass method work locally`_
+      -  `Step 3 - Verify local test mode`_
 
--  Your model may be run many times in a hyperparameter search.
+-  Docker- or Cluster-related Issues
+      -  `Step 4 - Verify that the original code runs in a notebook or shell`_
+      -  `Step 5 - Verify that each Trial subclass method works in a notebook or shell`_
+      -  `Step 6 - Verify that local test mode works in a notebook or shell`_
 
--  Your model may be run distributed across multiple GPUs or machines.
+-  Higher-level Issues
+      -  `Step 7 - Verify that cluster test mode works with slots_per_trial set to 1`_
+      -  `Step 8 - Verify that a single-GPU experiment works`_
+      -  `Step 9 - Verify that a multi-GPU experiment works`_
 
-This guide will introduce each change incrementally as we work towards achieving a fully functioning
-model in Determined.
+**************
+ Prerequisite
+**************
 
-The basic steps for debugging are:
+Successful cluster installation as described in :ref:`Install the Determined cluster
+<install-cluster>`.
 
-**Model-related issues:**
+.. _step1:
 
-#. Does the original code run locally?
-#. Does each method of your ``Trial`` subclass work locally?
-#. Does local test mode work?
+*********************************************
+ Step 1 - Verify that your code runs locally
+*********************************************
 
-**Docker- or cluster-related issues:**
+This step assumes you have ported your model from code outside of Determined. Otherwise, skip to
+:ref:`Step 2 <step2>`.
 
-#. Does the original code run in a notebook or shell?
-#. Does each method of your ``Trial`` subclass work in a notebook or shell?
-#. Does local test mode work in a notebook or shell?
+Confirm that your code works as expected before continuing.
 
-**Higher-level issues:**
+.. _step2:
 
-#. Does cluster test mode work with ``slots_per_trial`` set to ``1``?
-#. Does a single-GPU experiment work?
-#. Does a multi-GPU experiment work?
+**************************************************************
+ Step 2 - Verify that each Trial subclass method work locally
+**************************************************************
 
-**********************
- Model-related Issues
-**********************
+This step assumes you have a working local environment for training. If you typically run your code
+in a Docker environment, skip to :ref:`Step 4 <step4>`. This step also ensures that your class
+performs as expected by calling its methods and verifying the output.
 
-1. Does the original code run locally?
-======================================
+#. Create simple tests to verify each ``Trial`` subclass method.
 
-This step assumes you have have ported your model from some code outside of Determined. If your
-model is not based on any such code, skip to Step 2.
+   Examples of what these tests might look like for :class:`~determined.pytorch.PyTorchTrial` and
+   :class:`~determined.keras.TFKerasTrial` can be found in the
+   :meth:`determined.TrialContext.from_config` documentation, but only you can verify what is
+   reasonable for your test.
 
-Probably you already know your code works, but this is a reminder to double-check before proceeding.
+#. Perform failure diagnostics:
 
-2. Does each method of your ``Trial`` subclass work locally?
-============================================================
+   If you experience issues running the ``Trial`` subclass methods locally, it is likely there are
+   errors are in your trial class or the ``hyperparameters`` section of your configuration file.
+   Ideally, method-by-method evaluation makes it easier to find and solve issues.
 
-This step assumes you have a working local environment for training. If you normally run your code
-in a Docker environment instead, skip to Step 4.
+.. _step3:
 
-The goal of this step is to ensure that your class is behaving the way you think it should by
-calling its methods yourself and verifying the output.
+*********************************
+ Step 3 - Verify local test mode
+*********************************
 
-**How to test:** You should create some simple tests to verify that each method of your ``Trial``
-subclass is doing what you think it ought to be doing. There are some simple and practical examples
-of what these tests might look like for :class:`~determined.pytorch.PyTorchTrial` and
-:class:`~determined.keras.TFKerasTrial` in the :meth:`determined.TrialContext.from_config`
-documentation, but ultimately only you know what a reasonable test for your trial needs to look
-like.
+:ref:`Step 2 <step2>`, validated that your Trial API calls work as expected. This step uses your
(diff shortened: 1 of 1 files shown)


docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
-Then create an experiment with the ``--test`` flag (but not the ``--local`` flag):
+#. Launch a notebook or shell as described in :ref:`Step 4 <step4>`.
 
-.. code:: bash
+   On the cluster, testing is the same as :ref:`Step 3 <step3>`, except that the second, model
+   definition argument of the ``det experiment create`` command should be
+   ``/run/determined/workdir``, or ``.``, the current directory if you have not changed the working
+   directory after connecting to the cluster. This is because the ``--context`` specified when
+   creating the shell or notebook is copied to the ``/run/determined/workdir`` directory inside the
+   container, the same as the model definition argument is copied to ``det experiment create``.
```
rb-determined-ai, line 219: The commas here seem awkward.

> except that the second, model definition argument of the ``det experiment create`` command

Here "second, model definition argument" seems grammatically correct ("second" and "model definition" are two adjectives that both apply) but it feels less clear than just saying the "model definition argument".  I say this because from a CLI perspective, both `experiment` and `create` are technically arguments to the `det` command, but functionally `experiment create` represents a subcommand and a subsubcommand, with arguments after it, so numbering arguments is not crystal clear in this situation.

> should be ``/run/determined/workdir``, or ``.``

No comma after workdir: "should be X or Y".

> or ``.``, the current directory if you

"the current directory" is an appositive, and so needs a comma after it.  But in this case, I would guess it's widely enough known that `.` represents the current directory, and I would just remove the `, the current directory,` appositive entirely.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
--  The experiment config is fully validated for cluster-managed experiments, more strictly than it
-   is for ``--local --test`` mode. If you get errors related to ``unmarshaling JSON`` when trying to
-   submit the experiment to the cluster, that is an indication that the experiment config has
-   errors. Please review the :ref:`experiment configuration <experiment-configuration>`.
+   .. code:: yaml
 
-Again, if you are unable to identify the root cause of the issue yourself, please do not hesitate to
-contact Determined through our `community support
+      resources:
+      slots_per_trial: 1
```
rb-determined-ai, line 249: the `slots_per_trial` line needs an indent to be valid yaml

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+      into a custom Docker image or included in the ``startup-hook.sh`` file in the model definition
+      directory root. See :ref:`startup-hooks` for more information.
+
+   -  Custom bind mounts required in the interactive environment are specified in the experiment
+      configuration.
+
+   -  Environment variables are correctly set in the experiment configuration.
+
+   If no customizations are missing, the following, new layers introduced with a cluster-managed
+   experiment could be the problem cause:
```
rb-determined-ai, line 275: "could be the cause of the problem" feels far more idiomatic to me than "the problem cause"

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+   -  For a cluster-based experiment, configured ``checkpoint_storage`` settings are validated
+      before training starts. The message ``Checkpoint storage validation failed``, indicates that
+      you should review the ``checkpoint_storage`` setting values.
+
+   -  The experiment configuration is more strictly validated for cluster-managed experiments than
+      for ``--local --test`` mode. Errors related to ``unmarshaling JSON`` when attempting to submit
+      the experiment to the cluster indicate that the experiment configuration has errors. Review
+      the :ref:`experiment configuration <experiment-configuration>`.
+
+If you are unable to identify the root cause of the issue, contact Determined `community support
```
rb-determined-ai, line 295: suggest reprasing "the root cause of the issue" to match the phrasing X used before in "If no customizations are missing, the following, new layers introduced with a cluster-managed experiment could be the X"

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+
+      -  :ref:`master-configuration`
+      -  :ref:`experiment-configuration`
+
+   -  For a cluster-based experiment, configured ``checkpoint_storage`` settings are validated
+      before training starts. The message ``Checkpoint storage validation failed``, indicates that
+      you should review the ``checkpoint_storage`` setting values.
+
+   -  The experiment configuration is more strictly validated for cluster-managed experiments than
+      for ``--local --test`` mode. Errors related to ``unmarshaling JSON`` when attempting to submit
```
rb-determined-ai, line 291: Oh!  This error message is no longer what we give to users (thankfully, since it was a terrible message we used to show users).

Instead of ``` ``unmarshalling JSON`` ``` we can say ``` ``invalid experiment configuration`` ``` which is the new pattern that would show up in the command output.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+   it is set to ``1``. For example:
+
+   .. code:: yaml
+
+      resources:
+        slots_per_trial: 1
+
+#. Create an experiment without the ``--test`` or ``--local`` arguments:
+
+   You might find the ``--follow``, or ``-f``, arguments helpful:
```
rb-determined-ai, line 319: I do think `flag` is a more correct word than `argument`.  "Flags" generally don't have parameters (like `--verbose`), while "options" usually do (like `--output dump.txt`).  I think of "arguments" as either being positional arguments (neither flags nor options) or as a general term for tokens on the command line (where flags and options and positional arguments are all types of arguments).

In either case, "arguments" should not be plural because it's the same flag, just in a long form and a short form.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
 
-**How to diagnose failures:** If you are using the ``determined`` library APIs correctly, then
-theoretically distributed training should "just work". However, you should be aware of some common
-pitfalls:
+      -  If you have four Determined agents running with four GPUs each, your ``slots_per_trial``
+         could be ``1``, ``2``, ``3``, or ``4``, which would fit on a single machine.
+      -  A ``slots_per_trial`` value of ``8``, ``12``, or ``16`` would completely utilize a number
+         of agent machines.
+      -  A ``slots_per_trial`` value of ``5``, implies more than one agent but is not a multiple of
+         agent size.
```
rb-determined-ai, line 375: this is an invalid case, but that is not explained in the text of the bullet

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
 .. _model-debug:
 
-###########
- Debugging
-###########
+##############
+ Debug Models
```
rb-determined-ai, line 4: Why not "Debugging Models"?  My thinking is that "Debug Models" does not make it clear that "Debug" is a verb; it could be an adjective the way that it is when you talk about a "debug build" of some C or C++ code.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
 
--  Your code will conform to the Trial APIs by being organized into a subclass of Determined's
-   ``Trial`` class (indirectly through one of its concrete subclasses, such as
-   :class:`~determined.pytorch.PyTorchTrial`).
+These debugging steps introduce code changes incrementally, working toward a fully functioning
+Determined model. Follow the nine steps as applicable to your environment:
 
--  The code will run in a Docker container on another machine.
+-  Model-related Issues
+      -  `Step 1 - Verify that your code runs locally`_
```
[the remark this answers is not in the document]
gh-determined-ai: Referencing the header thusly: href="#step-1-verify-that-your-code-runs-locally"
Pretty standard link encoding.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+#. Launch a notebook or shell on the cluster:
 
-**How to test:** First launch a notebook or shell on the cluster, passing the root directory
-containing your model and training scripts to the ``--context`` option on the command line.
+   Pass the root directory containing your model and training scripts in the ``--context`` argument:
 
-Note that changes made to the ``--context`` directory while inside the notebook or shell will not
-affect the original files outside of the notebook or shell; see :ref:`notebook-state` for more
-details.
+   If you are using a Jupyter notebook, enter:
```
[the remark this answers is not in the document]
gh-determined-ai: Fixed as requested.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
+ Step 3 - Verify local test mode
+*********************************
 
-**How to test:** You should create some simple tests to verify that each method of your ``Trial``
-subclass is doing what you think it ought to be doing. There are some simple and practical examples
-of what these tests might look like for :class:`~determined.pytorch.PyTorchTrial` and
-:class:`~determined.keras.TFKerasTrial` in the :meth:`determined.TrialContext.from_config`
-documentation, but ultimately only you know what a reasonable test for your trial needs to look
-like.
+:ref:`Step 2 <step2>`, validated that your Trial API calls work as expected. This step uses your
```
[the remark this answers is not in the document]
gh-determined-ai: Fixed as requested.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
-that it is meeting Determined's requirements.
+      det experiment create myconfig.yaml my_model_dir --local --test
 
-**How to test:** Simply create an experiment, passing ``--local --test`` on the command line:
+   The ``--local`` argument specifies that training occurs where you launched the experiment instead
+   of occurring on a cluster. The ``--test`` argument runs abbreviated workloads to try to detect
+   bugs sooner and exits immediately.
 
-.. code:: bash
+#. The test is considered to have passed if the command completes successfully.
```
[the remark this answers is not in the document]
gh-determined-ai: You're right. It belongs as the last sentence of the previous step.

docs/training-debug/index.txt
gh-determined-ai: Changed according to what I think you meant.

docs/training-debug/index.txt
gh-determined-ai: Fixed.

docs/training-debug/index.txt
gh-determined-ai: I'm still leaning toward 'argument' (singular), as in Python's argparse(). 'Option' could also be ok for a CLI. If it was a program, it would be 'parameter'. I think 'flag' is the more general of all of those but usually refers to a programming variable, unofficially.

Let's have one go-around and pick one. They'll know what we mean.

docs/training-debug/index.txt
gh-determined-ai: Clarified as requested, along with the last bullet for the '32' case.

docs/training-debug/index.txt
gh-determined-ai: Agree. But headings should not use a gerund (except for Getting Started because it's so ubiquitous). The right title would be 'How to Debug Models' (infinitive form).

docs/training-debug/index.txt
```diff
@@ -1,343 +1,395 @@
...
 
-   det experiment create myconfig.yaml my_model_dir --local --test
+#. Perform failure diagnostics:
```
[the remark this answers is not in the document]
gh-determined-ai: Fixed as requested.

docs/training-debug/index.txt
gh-determined-ai: Fixed. All good suggestions.

docs/training-debug/index.txt
gh-determined-ai: fixed.

docs/training-debug/index.txt
gh-determined-ai: Already fixed.

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
-If you prefer to interact with your model via a Jupyter notebook, try:
+   .. code:: bash
 
-.. code:: bash
+      det notebook start --context my_model_dir
+      # Your browser should automatically open the notebook.
 
-   det notebook start --context my_model_dir
-   # Your browser should automatically open the notebook.
+   If you are using SSH to interact with your model, enter:
```
rb-determined-ai, line 137: "if you prefer"...

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
+This step is the same as :ref:`Step 2 <step2>`, except the original code runs on the Determined
+cluster instead of locally.
 
-**How to diagnose failures:** Failure diagnosis is a combination of the failure diagnosis for Step 2
-and for Step 4.
+#. Launch a notebook or shell:
 
-6. Does local test mode work in a notebook or shell?
-====================================================
+   If you are using a Jupyter notebook, enter:
```
rb-determined-ai, line 183: "if you prefer"

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
 
-**How to test:** Launch a notebook or shell as in Step 4.
+      det notebook start --context my_model_dir
+      # Your browser should automatically open the notebook.
 
-Once you are on the cluster, testing is identical to Step 3, with the important caveat that the
-model definition argument to ``det experiment create`` (the second positional argument) should
-always be ``/run/determined/workdir`` (or ``.`` if you have not changed the working directory from
-when you first connected to the cluster).
+   If you are using SSH to interact with your model, enter:
```
rb-determined-ai, line 190: "if you prefer"

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
-**How to diagnose failures:**
+   After you are on the cluster, testing is the same as :ref:`Step 1 <step1>`.
 
--  If you are unable to start the container with a message about the context directory exceeding the
-   maximum allowed size, it is because the ``--context`` directory cannot be more than 95MB in size.
-   If you need files larger than that as part of your model definition, consider setting up a bind
-   mount via the ``bind_mounts`` field of the :ref:`task configuration
-   <command-notebook-configuration>`. The :ref:`prepare-data` document lists some additional
-   strategies for accessing files inside a containerized environment that may also be relevant.
+#. Perform the following failure diagnostic steps:
```
rb-determined-ai, line 152: "Diagnose failures", maybe?

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
-.. code:: bash
+   On the cluster, testing is the same as :ref:`Step 3 <step3>`, except that the second model
+   definition argument of the ``det experiment create`` command should be
+   ``/run/determined/workdir`` or ``.`` if you have not changed the working directory after
+   connecting to the cluster. This is because the ``--context`` specified when creating the shell or
+   notebook is copied to the ``/run/determined/workdir`` directory inside the container, the same as
+   the model definition argument is copied to ``det experiment create``.
 
-   det experiment create myconfig.yaml my_model_dir --test
+#. Perform failure diagnosis:
```
rb-determined-ai, line 221: "Diagnose failures"

docs/training-debug/index.txt
```diff
@@ -1,343 +1,397 @@
...
 
-**How to diagnose failures:** If you were able to run local test mode inside a notebook or shell,
-but you are unable to successfully submit an experiment, you should focus on making sure that any
-customizations you made to get it to work in the notebook or shell have been properly replicated in
-your :ref:`experiment config <experiment-configuration>`:
+   Follow the combined failure diagnostic steps used for :ref:`Step 3 <step3>` and :ref:`Step 4
```
rb-determined-ai, line 223: suggested reword: "Follow the combined failure diagnoses from..."

Similar motivation as "Diagnose failures" vs "Perform failure diagnosis".  I don't feel strongly about this reword at all.

docs/training-debug/index.txt
rb-determined-ai: What I mean is: a couple of sentences earlier you say 

> If no customizations are missing, the following new layers introduced with a cluster-managed experiment could be the cause of the problem:"

I'm suggesting changing what you had before: 

> If you are unable to identify ~the root cause of the issue~, ...

to

> If you are unable to identify **the cause of the problem**, ...

specifically to match the phrasing you used just before.

I'm not a big fan of the newest phrasing:

> If no customizations are missing and the new layers introduced with a cluster-managed experiment are not the cause of the problem, contact Determined

because I would always rather encourage users to try to figure it out themselves before coming to us, and also because a user might find it very difficult to determine if "the new layers introduced with a cluster-managed experiment are not the cause of the problem"

gh-determined-ai: Agree. Fixed.

docs/training-debug/index.txt
gh-determined-ai: fixed

docs/training-debug/index.txt
gh-determined-ai: fixed.

docs/training-debug/index.txt
gh-determined-ai: fixed.

docs/training-debug/index.txt
gh-determined-ai: Fixed.

rb-determined-ai (approved):
This is great!  It looks like professional documentation now : )

PR #3688, opened 2022-03-03
test: log filter test [DET-6787]

hkang1:
## Description

Write interaction test for `LogViewerFilters.tsx`.

## Test Plan

- [ ] check test coverage increased for `LogViewerFilters.tsx`

## Commentary (optional)

Something to note is that Ant Design v4 (v3 is not impacted) select dropdown has [an issue](https://github.com/ant-design/ant-design/issues/23009#issuecomment-929766415) where when opening a dropdown menu, the options list container has the style of "opacity: 0; event-pointer: none;" applied in the testing environment. The workaround was it either to use the lower level `fireEvent` from `react-testing-library` or pass a `skipPointerEventsCheck` value of true to `userEvent.click()`.

## Checklist

### DIFF
--- a/webui/react/src/utils/string.ts
+++ b/webui/react/src/utils/string.ts
@@ -43,6 +43,16 @@ export const generateAlphaNumeric = (
   return result;
 };
 
+export const generateUUID = (): string => {
+  return [
+    generateAlphaNumeric(8),
+    generateAlphaNumeric(4),
+    generateAlphaNumeric(4),
+    generateAlphaNumeric(4),
+    generateAlphaNumeric(12),
+  ].join('-');
+};
+
 export const generateLetters = (length = DEFAULT_ALPHA_NUMERIC_LENGTH): string => {
   return generateAlphaNumeric(length, LETTERS);
 };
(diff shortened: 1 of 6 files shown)


webui/react/src/utils/string.ts
```diff
@@ -47,6 +47,16 @@ export const generateLetters = (length = DEFAULT_ALPHA_NUMERIC_LENGTH): string =
   return generateAlphaNumeric(length, LETTERS);
 };
 
+export const generateUUID = (): string => {
```
mapmeld, line 50: Since tests are running on Node / major browsers, let's `import crypto` to natively call `crypto.randomUUID()` and we won't need the test either

hkang1: ah nice, good call!

mapmeld (approved):

webui/react/src/utils/string.ts
hkang1: doh! So it looks like crypto.randomUUID is not supported in node 14.16 which is what our CI is based off of. I don't really want to rock the boat by upgrading to 14.18 . I could do a polyfill but that one I function I had before would be a lot less to maintain then another library.

The other issue was that typescript doesn't have native definition for crypto.randomUUID until 4.6.2 (their most recent release) and when I upgrade to tag, there are a slew of other fixes that needed address (specifically the unknown error type handling, no longer forgiving)

PR #3693, opened 2022-03-04
chore: add AWS EC2 p4d.24xlarge

varlogtim:
You should be able to `det d aws up --compute-agent-instance-type p4d.24xlarge ...` and have it just work without modifying the master config now.


ioga (approved):
should we also add `g5.*` instance types?

vishnu2kmohan:
> should we also add `g5.*` instance types?

Agreed with @ioga, let's please use this opportunity to piggyback and add the `g5.*` instances as well, @varlogtim :pray:

PR #3694, opened 2022-03-04
refactor: add all gather for v2 rendezvous

stoksc:
## Description
This change adds functionality to perform an all gather through the master. This is useful to do rendezvous without any shared_fs or other external coordination tools in a world where the RM cannot provide IP addrs.

## Test Plan
- [x] will add units

## Checklist

### DIFF
--- a/master/internal/task/all_gather.go
+++ b/master/internal/task/all_gather.go
@@ -1,0 +1,146 @@
+package task
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/determined-ai/determined/master/pkg/actor/actors"
+	"github.com/determined-ai/determined/master/pkg/ptrs"
+
+	"github.com/pkg/errors"
+
+	"github.com/determined-ai/determined/master/pkg/actor"
+)
+
+var (
+	// DefaultAllGatherTimeout is the default timeout for all gather.
+	DefaultAllGatherTimeout = 10 * time.Minute
+	// AllGatherTimeoutMessage is the error returned when an all gather times out.
+	AllGatherTimeoutMessage = "some ranks are taking a long time to connect to master" +
+		"during all gather; when running on kubernetes this may happen " +
+		"because only some of the pods have been scheduled; it is possible " +
+		"that some pods will never be scheduled without adding compute " +
+		"resources or pausing / killing other experiments in the cluster"
+)
+
+type (
+	allGather struct {
+		id       uuid.UUID
+		watchers map[uuid.UUID]chan AllGatherInfoOrError
+		data     []*structpb.Struct
+		numPeers *int
+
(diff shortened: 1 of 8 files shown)


master/internal/task/all_gather.go
```diff
@@ -0,0 +1,169 @@
...
+		data     []*structpb.Struct
+		numPeers *int
+
+		readyPassed  bool
+		lastPeerJoin time.Time
+	}
+
+	// AllGatherWatcher contains a channel which can be polled for all gather completion.
+	AllGatherWatcher struct {
+		C <-chan AllGatherInfoOrError
```
ioga, line 34: since this actor is returning channels anyway... maybe just make this entire chunk of code without actors, in "normal go"?

master/internal/task/allocation.go
```diff
@@ -224,6 +226,25 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
...
+			case WatchAllGather:
+				a.allGather = NewAllGather()
+				a.allGather.PreStart(ctx)
+			case UnwatchAllGather, allGatherTimeout:
+				// Ignore without active all gather.
+				return nil
+			}
+		}
+
+		switch finished, err := a.allGather.ReceiveMsg(ctx); {
```
ioga, line 241: ah, so there's no explicit `ctx.Respond` here because you call `ReceiveMsg` directly, which does `ctx.Respond` internally. 
so the `AllGather` is not really an actor of its own anyway, even though it looks like one. another reason for it not to be "actor-y".

master/internal/task/all_gather.go
```diff
@@ -0,0 +1,169 @@
...
+	return true
+}
+
+// checkTimeout checks if the task should timeout waiting for rendezvous.
+func (g *AllGather) checkTimeout(id uuid.UUID) error {
+	if g == nil {
+		return nil
+	}
+
+	if g.id == id && time.Now().UTC().After(g.lastPeerJoin.Add(DefaultAllGatherTimeout)) {
```
ioga, line 146: this is checking for timeout after the last peer, but `allGatherTimeout` is only triggered once in `PreStart`. should it be fired off periodically instead, or after every `.watch` call?

stoksc: my thought was just if an all gather takes longer than 10m from start to finish it prints the timeout, rather than adding any logic to restart the timer.

ioga: then maybe this should just check if any peers have joined at all? because if anyone has joined then lastPeerJoin will always be less than DefaultAllGatherTimeout.

stoksc: oh i see what you mean, duh. i think i'll just rm lastPeerJoin and make it allGatherDeadline (time.Now() at start + timeout). sorry for confusion.

ioga (approved):

PR #3695, opened 2022-03-04
ci: deploy-release-party is no longer used in releases

mackrorysd:
## Description

This is broken / bit-rotted and it isn't even used in our release process anymore. YANK!

## Test Plan

As long as CircleCI doesn't complain about syntax errors on this config...


mackrorysd:
Those unit tests are pretty clearly unrelated. But will follow-up to make sure someone looks at them...

justin-determined-ai (approved):

PR #3702, opened 2022-03-07
ci: temporarily stop test flakes

stoksc:
## Checklist


mackrorysd (approved):
Well obviously it's not ideal, but it's better than having test flakes if this is really the issue. And it's better if one looks at it from the standpoint of wanting to make a statement about lag on the logs at least being <10 seconds reliably. Thank you for filing the JIRA to investigate further.

Conventional commits specifically says that other prefixes are allowed. I wonder if this would go better with the "ci:" prefix? And of course please look into the other CircleCI failures.

stoksc:
> And of course please look into the other CircleCI failures.

which others?

PR #3704, opened 2022-03-07
fix: trial rendezvous timeout should just warn

stoksc:
## Description
In 0.17.0 (push arch) we had a regression such that failing to rendezvous after 10m would cause a failure rather than just a warning. This PR comprises 2 small changes. First, the behavior is reverted back to the old. Second, the deadline does not start until the first connection to avoid spurious warnings for things such as long start up scripts. Both are done rather than just the second since for k8s the kill resulting from not suppressing the error could still be incorrect.

## Test Plan
- [ ] add a release note, at least

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -328,7 +328,6 @@ func (a *Allocation) ResourcesAllocated(ctx *actor.Context, msg sproto.Resources
 
 	if a.req.DoRendezvous {
 		a.rendezvous = NewRendezvous(a.model.AllocationID, a.reservations)
-		a.rendezvous.PreStart(ctx)
 	}
 
 	if cfg := a.req.IdleTimeout; cfg != nil {
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -215,7 +215,12 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 			}
 			return nil
 		}
-		if err := a.rendezvous.ReceiveMsg(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


master/internal/task/allocation.go
```diff
@@ -215,7 +215,12 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 			}
 			return nil
 		}
-		if err := a.rendezvous.ReceiveMsg(ctx); err != nil {
+		switch err := a.rendezvous.ReceiveMsg(ctx).(type) {
+		case nil:
+			return nil
+		case ErrTimeoutExceeded:
```
stoksc, line 221: I did check similar code, but preemption should _not_ ignore ErrTimeoutExceeded

ioga (commented):
is it possible *not* erroring out will cause indefinite hangs in some situations? thinking about xxxx complaining about their experiment freezes.

stoksc:
@ioga i think that's pretty unlikely, but this is just restoring the original behavior because this could totally happen with k8s without the scheduler plugin, so i think it has to be this way until with want to swap out this behavior in that case.

stoksc:
i mean, it's possible, but i don't think it's the job of the rendezvous timeout to catch indefinite hangs, except those with rendezvous, and those would be very unlikely i tihnk.

stoksc:
like, maybe you could see this if you did `sleep 10*DET_RANK_ID` in your startup script, haha.

stoksc:
landed through failing python tests that are fixed upstream

PR #3705, opened 2022-03-08
fix: ensure allocation end time is always valid

stoksc:
## Description
An edge case arose with our cluster heartbeat scheme where we save the heartbeat, an allocation is started and the cluster dies. In this case, the end time will be set to the most recent cluster heartbeat but that is before end time so the logs show `failed to aggregate resource allocation: failed to add aggregate: ERROR: range lower bound must be less than or equal to range upper bound (SQLSTATE 22000)  id="allocation-aggregator" system="master" type="allocationAggregator"`.

## Checklist

### DIFF
--- a/master/internal/db/postgres_cluster_intg_test.go
+++ b/master/internal/db/postgres_cluster_intg_test.go
@@ -7,9 +7,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


nrajanee (approved):

nrajanee (commented):
postgres_cluster_intg_test is failing in integration tests

PR #3709, opened 2022-03-09
docs: change master and agent config parameter listing to indexable tabular format

gh-determined-ai:
## Description

Reformatted master and agent configuration parameter documentation. In the CLUSTER SETUP section -> Basics, see Master Configuration Reference and Agent Configuration Reference. At least for the master, config sections are available in the right-side nav sidebar and any of these sections can be indexed from within the doc set.

The Config Cluster topic, itself, still needs some work.

## Checklist


varlogtim:
The problem here is that the config options aren't a list. They are hierarchical. Imagine like clicking on directories in Windows OS. My dream would be something like this:

1. Start
+<Agent Config> (completely collapsed)

2. Click the <Agent Config> and it expands down with all of the first level options, all of which are collapsed.
<Agent Config>
  ...
  + <security>
3. Click the <security> and that expands those options out:
<Agent Config>
  ...
  - <security>
    + <tls>
4. etc...
The point here is that it is easy to determine where in the hierarchy a config option exists in. When this stuff is used in the config file, it needs to have the proper parentage. The agent config isn't as bad as the master, which has many different possible options.

I feel like this would make it easier to navigate to the thing I wanted and even just read the config.

gh-determined-ai:
@varlogtim  I agree that we need to show the hierarchy. One problem I had is that I couldn't find the master or agent JSON schemas anywhere. Do you know where those might be defined? Also, it would be good to have real-world examples, fairly fleshed out, to go with the descriptions.
I think the dynamic action you describe is probably more than is needed but I'll investigate the possibility for such an add-on to our static site generator. I think it's better to show the full structure at a glance (the JSON isn't that large) as long as the hierarchy is clearly displayed. Now, if I can only get that hierarchy.

varlogtim:
@gh-determined-ai Here is the structure: https://docs.determined.ai/latest/sysadmin-basics/cluster-config.html?highlight=master%20config#master-configuration, which is the problem I am talking about. It is so long that it is difficult to determine the hierarchy of the options when you are in the middle of it. There aren't links to specific sections in it from other areas of the docs. You can see though that a table would make the master config worse.

I really feel like this should be auto-generated somehow. There might be back end work that needs to happen before this.

As far as a schema for the master and agent, I am not an authority on this, but I feel like it is only inside the Golang code (`master/internal/config.go` `Config`). The experiment config schema is in `schemas/expconf/v0/experiment.json`, but I don't think there is an equivalent for the master.

gh-determined-ai:
@varlogtim Thanks for sticking with this.
I've sussed out that the schema is in the code but haven't deciphered the full structure yet. I'm aware of the document you linked to and that's what I'm looking to improve. A bullet list is definitely not the solution (a common misconception even among tech writers) and there are all kinds of tables (some readable, some not). A lot of people don't realize that we have a primary navigation mechanism, which is the sidebar on the right. (And I want to get people used to using the navigation correctly, which had been subverted by some of the unfortunate constructs in the existing docs, but that another project.) Properly instantiated, the nav panel can provide the needed indexing regardless of the size of the table(s). It's no different than expandable lists within the content, itself. If you'll look at this branch, that's exactly what I did, although, I'm sure it can be improved and that's what I'm looking to you for.
If I didn't find the schema in the source, it's doubtful it can be autogenerated without adding the source for it.
Would rather not add JavaScript when the nav panel gives us, essentially, the same thing.

varlogtim:
> A lot of people don't realize that we have a primary navigation mechanism, which is the sidebar on the right.

So, I guess you are saying that if we put the entire config inside the navigation, then we can scroll through that and all the details are in the middle?

I don't particularly like the tables ... the data structure doesn't fit the data.

I guess I might want some auto-magical config docs generator thing where the docs are in the code and the docs come from the code :mindblown: ... but I guess it isn't really needed.

Another idea I had to solve my problem of "where do I put this in the config" is when mousing over, show the `the.path[0].to.the.config.option`

PR #3712, opened 2022-03-09
fix: run all migrations in transactions

stoksc:
## Description
This change updates all past migrations to run in transactions.

## Test Plan
- [x] start a cluster, all migrations run fine in a transaction

## Checklist


ioga (approved):
Test plan:
1. `cd master/internal/migrations/`
2. `./migration-create.sh wakeup`
3. in the migration file, `CREATE TABLE ...` and then `SELECT pg_sleep(20);`
4. Run postgres with `cmdline: ["postgres", "-c", "log_statement=all"]`
5. Run master; kill it when it starts running pg_sleep.
6. Check if the table you're supposed to create exists (that'll happen after pg_sleep finishes, i.e. there's a delay up to 20 seconds).
7. It shouldn't appear after this PR.

gh-determined-ai (approved):

PR #3717, opened 2022-03-09
fix: allow ranks to report independent early exits [DET-6792, DET-6835]

stoksc:
## Description
This change modifies the searcher logic to allow multiple invalid HP exits per trial. Exits after the first are ignored. This is useful for multi-rank trials to invalid HP without coordination. The change in behavior in this change is that now, rather than the experiment failing violently on the receipt of the 2nd invalid HP, nothing will happen. Nothing includes no kill, cancel, etc. The master will just wait for the trial to exit on its own, so it is a bit susceptible to hangs.

## Test Plan
- [ ] run invalid hp dtrain, see less failures

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -54,10 +54,18 @@ type (
 	trialGetSearcherState struct {
 		requestID model.RequestID
 	}
-	// trialClosed is used to replay closes missed when the master dies between when a trial closing in
-	// its actor.PostStop and when the experiment snapshots the trial closed.
+
+	// trialClosed is used to replay closes missed when the master dies between when a trial closing
+	// in its actor.PostStop and when the experiment snapshots the trial closed.
 	trialClosed struct {
 		requestID model.RequestID
+	}
+
+	// userInitiatedEarlyExit is a user-injected message, provided through the early exit API. It
+	// _should_ indicate the user is exiting, but in the event they don't, we will clean them up.
+	userInitiatedEarlyExit struct {
+		requestID model.RequestID
+		reason    model.ExitedReason
 	}
 )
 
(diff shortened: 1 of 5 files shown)


master/internal/trial.go
```diff
@@ -235,6 +242,25 @@ func (t *trial) maybeAllocateTask(ctx *actor.Context) error {
 	return nil
 }
 
+const (
+	// InvalidHPKillDelay the delay before we forcibly kill a trial that said it had an invalid HP.
+	InvalidHPKillDelay = 10 * time.Second
```
liamcli, line 247: can we lower this to 3 seconds.  that matches the grace period we give the launched processes in the harness

stoksc, line 247:
```suggestion
	InvalidHPKillDelay = 3 * time.Second
```

master/internal/experiment.go
```diff
@@ -54,11 +54,19 @@ type (
...
+
+	// trialClosed is used to replay closes missed when the master dies between when a trial closing
+	// in its actor.PostStop and when the experiment snapshots the trial closed.
 	trialClosed struct {
 		requestID model.RequestID
 	}
+
+	// userInitiatedEarlyExit is a user-injected message, provided through the early exit API. It
+	// _should_ indicate the user is exiting, but in the event they don't, we will clean them up.
+	userInitiatedEarlyExit struct {
```
liamcli, line 66: Do we still need type `trialReportEarlyExit` above and related code in experiment?

stoksc: yeah. the flow here now is a `userInitiatedEarlyExit` is sent to the trial and the trial tells the experiment `trialReportEarlyExit` once it is done and closed.

liamcli (approved):
i tested locally and lgtm

PR #3718, opened 2022-03-09
fix: download cifar10 only once per node in cifar10_pytorch example

coreystatendet:
## Description

Download cifar10 only once per node in cifar10_pytorch example. [DET-6801]

## Test Plan

Run the example at `computer_vision/cifar10_pytorch/distributed.yaml`, i.e. from root directory run:
```
cd computer_vision/cifar10_pytorch/distributed.yaml && det experiment create distributed.yaml .
```

The trial logs should contain the string `Downloading https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz to data/cifar-10-python.tar.gz` exactly once per node.  It's a 16 GPU experiment, so if nodes have 8 GPUs it should appear twice.

## Commentary (optional)

Previously, a different temp directory was made for each rank to hold independent copies of the dataset.  Instead, this changes the model to use a single fixed directory `data`, and FileLocks to ensure only one rank attempts to download at once.  All other ranks after the first on a node will skip downloading, as the files will already be fully downloaded.

## Checklist
(None of the below should apply, minor fix.)

- [] User-facing API changes need the "User-facing API Change" label.
- [] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-6801]: https://determinedai.atlassian.net/browse/DET-6801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/computer_vision/cifar10_pytorch/model_def.py
+++ b/examples/computer_vision/cifar10_pytorch/model_def.py
@@ -104,23 +104,22 @@ class CIFARTrial(PyTorchTrial):
         accuracy = accuracy_rate(output, labels)
         return {"validation_accuracy": accuracy, "validation_error": 1.0 - accuracy}
 
-    def build_training_data_loader(self) -> Any:
+    def _download_dataset(self, train: bool) -> Any:
         transform = transforms.Compose(
             [transforms.ToTensor(), transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))]
         )
+        # Use a file lock so that workers on the same node attempt the download one at a time.
+        # The first worker will actually perform the download, while the subsequent workers will
+        # see that the dataset is downloaded and skip.
         with filelock.FileLock(os.path.join(self.download_directory, "lock")):
-            trainset = torchvision.datasets.CIFAR10(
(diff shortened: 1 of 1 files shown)


examples/computer_vision/cifar10_pytorch/model_def.py
```diff
@@ -104,17 +108,19 @@ def build_training_data_loader(self) -> Any:
         transform = transforms.Compose(
             [transforms.ToTensor(), transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))]
         )
-        trainset = torchvision.datasets.CIFAR10(
-            root=self.download_directory, train=True, download=True, transform=transform
-        )
+        with filelock.FileLock(os.path.join(self.download_directory, "lock")):
```
rb-determined-ai, line 111: Can you put a comment above this line that says:
```python
# Use a file lock so only one worker on each node does the download.
```

I don't think it's obvious, especially to non-dtrain-aware users, why we would use a filelock here.

rb-determined-ai (approved):

PR #3723, opened 2022-03-10
feat: Accept model.id for lookups, show model.id in UI [DPS-49]

mapmeld:
## Description

This change would accept  `/det/models/:model_id` and `/api/v1/models/:model_id` as valid routes. A model name can't be all digits, so there should be no collision between model names and IDs.  The `GetModel` Go code is also used in PatchModel, GetModelVersion, PostModelVersion, and GetModelVersions, but we don't need to support IDs there explicitly.

The second part of the request is **displaying** the ID and **making the permalink visible**. I've added the number to breadcrumbs in the PR as a start. We could change to use the model ID in WebUI's links, so you are always on the permalink URL. Or we can make a permalink button.

Current proposal for updated breadcrumbs:

Screen Shot 2022-03-10 at 11 40 51 AM

## Test Plan

Go to `/api/v1/models` to see your list of models

- Given any model, `/api/v1/models/:model_name` and `/api/v1/models/:model_id` should get the same response
- Given a numeric ID for a model which doesn't exist, the API should return not found

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_model_registry.py
+++ b/e2e_tests/tests/cluster/test_model_registry.py
@@ -32,7 +32,7 @@ def test_model_registry() -> None:
         # Confirm we can look up a model by its ID
         db_model = d.get_model_by_id(mnist.model_id)
         assert db_model.name == "mnist"
-        db_model = d.get_model(str(mnist.model_id))
+        db_model = d.get_model(mnist.model_id)
         assert db_model.name == "mnist"
 
         # Confirm DB assigned username
--- a/e2e_tests/tests/cluster/test_model_registry.py
+++ b/e2e_tests/tests/cluster/test_model_registry.py
@@ -29,6 +29,12 @@ def test_model_registry() -> None:
         assert mnist.metadata == db_model.metadata
         assert mnist.metadata == {"testing": "metadata"}
 
+        # Confirm we can look up a model by its ID
+        db_model = d.get_model_by_id(mnist.model_id)
+        assert db_model.name == "mnist"
+        db_model = d.get_model(str(mnist.model_id))
+        assert db_model.name == "mnist"
+
         # Confirm DB assigned username
         assert db_model.username == "determined"
 
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -155,14 +155,15 @@ class Determined:
 
         return model.Model._from_json(r.json().get("model"), self._session)
 
-    def get_model(self, name: str) -> model.Model:
+    def get_model(self, identifier: Union[str, int]) -> model.Model:
         """
         Get the :class:`~determined.experimental.Model` from the model registry
-        with the provided name. If no model with that id is found in the registry,
+        with the provided name or ID. If no corresponding model is found in the registry,
         an exception is raised.
         """
-        r = self._session.get("/api/v1/models/{}".format(name))
-        return model.Model._from_json(r.json().get("model"), self._session)
+        r = self._session.get("/api/v1/models/{}".format(identifier)).json()
+        assert r.get("model", False)
+        return model.Model._from_json(r.get("model"), self._session)
 
     def get_model_by_id(self, model_id: int) -> model.Model:
         """
@@ -170,10 +171,7 @@ class Determined:
         with the provided id. If no model with that id is found in the registry,
         an exception is raised.
         """
-        r = self._session.get("/api/v1/models?id={}".format(model_id))
-        models = r.json().get("models")
-        assert len(models) == 1
-        return model.Model._from_json(models[0], self._session)
+        return self.get_model(model_id)
 
     def get_models(
         self,
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -183,9 +183,9 @@ class Determined:
            an integer-type model ID.
         """
         warnings.warn(
-            "Determined.get_model_by_id() has been deprecated and will be removed",
-            "in a future version.\n",
-            "Please call Determined.get_model() with either a string-type name or",
+            "Determined.get_model_by_id() has been deprecated and will be removed"
+            "in a future version.\n"
+            "Please call Determined.get_model() with either a string-type name or"
             "an integer-type model ID.",
             FutureWarning,
         )
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -1,4 +1,5 @@
 import pathlib
+import warnings
 from typing import Any, Dict, List, Optional, Union, cast
 
 from determined.common import check, context, util, yaml
@@ -130,7 +131,7 @@ class Determined:
         Get the :class:`~determined.experimental.Checkpoint` representing the
         checkpoint with the provided UUID.
         """
-        r = self._session.get("/api/v1/checkpoints/{}".format(uuid)).json()
+        r = self._session.get(f"/api/v1/checkpoints/{uuid}").json()
         return checkpoint.Checkpoint._from_json(r["checkpoint"], self._session)
 
     def create_model(
@@ -158,10 +159,14 @@ class Determined:
     def get_model(self, identifier: Union[str, int]) -> model.Model:
         """
         Get the :class:`~determined.experimental.Model` from the model registry
-        with the provided name or ID. If no corresponding model is found in the registry,
+        with the provided identifer, which is either a string-type name or an
+        integer-type model ID. If no corresponding model is found in the registry,
         an exception is raised.
+
+        Arguments:
+            identifier (string, int): The unique name or ID of the model.
         """
-        r = self._session.get("/api/v1/models/{}".format(identifier)).json()
+        r = self._session.get(f"/api/v1/models/{identifier}").json()
         assert r.get("model", False)
         return model.Model._from_json(r.get("model"), self._session)
 
@@ -170,7 +175,20 @@ class Determined:
         Get the :class:`~determined.experimental.Model` from the model registry
         with the provided id. If no model with that id is found in the registry,
         an exception is raised.
+
+        .. warning::
+           Determined.get_model_by_id() has been deprecated and will be removed
+           in a future version.
+           Please call Determined.get_model() with either a string-type name or
+           an integer-type model ID.
         """
+        warnings.warn(
+            "Determined.get_model_by_id() has been deprecated and will be removed",
+            "in a future version.\n",
+            "Please call Determined.get_model() with either a string-type name or",
+            "an integer-type model ID.",
+            FutureWarning,
+        )
         return self.get_model(model_id)
 
     def get_models(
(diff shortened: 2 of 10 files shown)


mapmeld:
update: changed the WebUI URLs, made Model Name/ID interchangeable in the remaining few API routes (archive/delete/unarchive model).
I didn't need to update delete model version because we look up the model version by its unique ID.

gh-determined-ai (changes requested):
Suggest release note wording change to:

> The model registry API now  accepts either the ID or model name in ``/api/v1/models/:id`` or ``/api/v1/models/:name``. This applies to all API routes for models and model versions.


Also, include something from the Jira issue that describes the need for this and what this accomplishes ... something from:

> Objective: Track genesis of a model with a permalink that will never change so they don't have to worry if someone updates the name.
> Concerns with name: They have ran into issues when users change names because of org changes.
> They are fine with using id but their users would have to use the CLI just to get the id (for the workshop they have) and then go back to the WebUI hence the UI request.
>

gh-determined-ai (approved):

master/internal/api_model.go
```diff
@@ -243,8 +252,13 @@ func (a *apiServer) PatchModel(
 
 func (a *apiServer) ArchiveModel(
 	ctx context.Context, req *apiv1.ArchiveModelRequest) (*apiv1.ArchiveModelResponse, error) {
+	getResp, err := a.GetModel(ctx, &apiv1.GetModelRequest{ModelName: req.ModelName})
```
rb-determined-ai, line 255: This looks strange.  It looks like you are invoking an API endpoint handler as if it were a helper function?

I think you should refactor the part that is a helper function into a standalone helper function and call it from `a.GetModel` and also the places where you call `a.GetModel`.

e2e_tests/tests/cluster/test_model_registry.py
```diff
@@ -29,6 +29,12 @@ def test_model_registry() -> None:
         assert mnist.metadata == db_model.metadata
         assert mnist.metadata == {"testing": "metadata"}
 
+        # Confirm we can look up a model by its ID
+        db_model = d.get_model_by_id(mnist.model_id)
+        assert db_model.name == "mnist"
+        db_model = d.get_model(str(mnist.model_id))
+        assert db_model.name == "mnist"
```
rb-determined-ai, line 36: What does this mean for the python API?

If the interface is really:
```python
def get_model(identifier: Union[str, int]):
  ...
```
Then why not make that explicit, and deprecate `get_model_by_id()`?

I dunno, that seems like a sloppy interface.  But I guess if both types of key are guaranteed unique, then it's probably ok.

master/internal/api_model.go
stoksc: we're missing the 'service' layer from the controller/service/repo pattern everywhere in determined (except users) so this is fairly common. that pattern exists for exactly this, otherwise it gets annoying if you e.g. ever need to change get model, or you do a refactor of a bunch of model related stuff or `GetModel` panics because it thinks its always a top level api handler, etc etc. it is some overhead (for devs), and so it's probably most useful to add this layer only when you start to do these sorts of calls, because this is the point we're you begin walking into a hole for not having it.

rb-determined-ai: @stoksc are you agreeing with my refactor suggestion or disagreeing and saying that it would be hurtful to do the refactor suggestion now?

e2e_tests/tests/cluster/test_model_registry.py
mapmeld: I agree that it's not great to have one API method for both, but I can modify this to accept both types. I don't want to remove `get_model_by_id` in case someone updated their code to use this, but I will have it just call `get_model`

master/internal/api_model.go
mapmeld: In this latest commit I separate out `ModelFromIdentifier` and `ModelVersionFromId` to be reusable helper functions instead of using the API calls:  https://github.com/determined-ai/determined/pull/3723/commits/219e8011df3a864a178b9997d4f14616d690ee45

From my perspective it's a small-footprint change, just let me know if there's a preference about whether these (and `clearModelName`) should be in another file or named differently.

harness/determined/common/experimental/determined.py
```diff
@@ -155,25 +155,23 @@ def create_model(
...
+    def get_model(self, identifier: Union[str, int]) -> model.Model:
         """
         Get the :class:`~determined.experimental.Model` from the model registry
-        with the provided name. If no model with that id is found in the registry,
+        with the provided name or ID. If no corresponding model is found in the registry,
         an exception is raised.
         """
-        r = self._session.get("/api/v1/models/{}".format(name))
-        return model.Model._from_json(r.json().get("model"), self._session)
+        r = self._session.get("/api/v1/models/{}".format(identifier)).json()
```
rb-determined-ai, line 164: can you update this to use the newer f-string style (`.get(f"/api/v1/models/{identifier}"`)?

The reason this code was written with `.format()` was because it was written when the python cli still had to support python 3.5, but that's no longer true.

harness/determined/common/experimental/determined.py
```diff
@@ -155,25 +155,23 @@ def create_model(
 
         return model.Model._from_json(r.json().get("model"), self._session)
 
-    def get_model(self, name: str) -> model.Model:
+    def get_model(self, identifier: Union[str, int]) -> model.Model:
         """
         Get the :class:`~determined.experimental.Model` from the model registry
-        with the provided name. If no model with that id is found in the registry,
+        with the provided name or ID. If no corresponding model is found in the registry,
```
rb-determined-ai, line 161: perhaps clarify to say `with the provided identifer, which is either a string-type name or an integer-type model ID`.

harness/determined/common/experimental/determined.py
```diff
@@ -155,25 +155,23 @@ def create_model(
...
         an exception is raised.
         """
-        r = self._session.get("/api/v1/models/{}".format(name))
-        return model.Model._from_json(r.json().get("model"), self._session)
+        r = self._session.get("/api/v1/models/{}".format(identifier)).json()
+        assert r.get("model", False)
+        return model.Model._from_json(r.get("model"), self._session)
 
     def get_model_by_id(self, model_id: int) -> model.Model:
```
rb-determined-ai, line 168: can you modify this docstring to show it is deprecated, and also include a deprecation warning like [this one](https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/checkpoint/_checkpoint.py#L224)?

harness/determined/experimental/client.py
```diff
@@ -233,7 +233,7 @@ def get_model_by_id(model_id: int) -> Model:
         model_id (int): The unique id of the model.
```
rb-determined-ai, line 233: can you also deprecate `client.get_model_by_id()`, separately from `Determined.get_model_by_id()`?

What I mean is:
- mark it as deprecated in this doc string
- mark issue a distinct deprecation warning
- make client.get_model_by_id() call `_determined.get_model()` directly, so that a user who calls `client.get_model_by_id()` isn't hit with two deprecation warnings for a single call.

Basically, my reasoning is that when a user calls a public API function and they get a deprecation warning for some other function called by that public API function, it looks more like a bug in the library and less like a useful message.

mapmeld: ✅  This reminded me to allow `client.get_model()` to accept both types as well

harness/determined/common/experimental/determined.py
mapmeld: ✅

gh-determined-ai:
@mapmeld  I know. Not singling you out personally, you just happened to pop up. Someday all `please`, `note`, and `**` attention-getters will be removed, globally, as just so much unhelpful noise. You can leave it as is.

rb-determined-ai:
@mapmeld what's the plan with this pr?

mapmeld:
@rb-determined-ai I'm ready to merge it, it has the deprecation notices that you requested if you think it's ready to go..

master/internal/api_model.go
```diff
@@ -22,19 +22,50 @@ import (
...
+		err = a.m.db.QueryProto("get_model", m, identifier)
+	}
+	switch err {
 	case db.ErrNotFound:
 		return nil, status.Errorf(
-			codes.NotFound, "model \"%s\" not found", req.ModelName)
+			codes.NotFound, "model \"%s\" not found", identifier)
+	default:
+		return m, errors.Wrapf(err,
+			"error fetching model \"%s\" from database", identifier)
```
rb-determined-ai, line 41: what is the reason that you have different error handling in this final default case between ModelFromIdentifier ModelVersionFromID?

master/internal/api_model.go
```diff
@@ -469,7 +496,7 @@ func (a *apiServer) PatchModelVersion(
 
 	finalModelVersion := &modelv1.ModelVersion{}
 	err = a.m.db.QueryProto("update_model_version", finalModelVersion, req.ModelVersionId,
-		parentModel.Model.Id, currModelVersion.Name, currModelVersion.Comment, currModelVersion.Notes,
+		parentModel.Id, currModelVersion.Name, currModelVersion.Comment, currModelVersion.Notes,
```
rb-determined-ai, line 499: I don't understand what the point of this change is.  It looks like we haven't modified the proto so it seems like this shouldn't be changed, unless it was broken before.

rb-determined-ai (commented):
oh snap, I totally did not realize this was blocked on me.  The python side looks good, a couple questions left in the master side I think.

master/internal/api_model.go
mapmeld: I changed from `parentModel := a.GetModel` (where Model was inside of the returned GetModelResponse object) to `parentModel := currModelVersion.Model`

rb-determined-ai: ah, this makes sense

master/internal/api_model.go
```diff
@@ -22,19 +22,54 @@ import (
...
+	switch err = a.m.db.QueryProto(
+		"get_model_version", mv, parentModel.Id, versionID); {
+	case err == db.ErrNotFound:
+		return nil, status.Errorf(
+			codes.NotFound, "model \"%s\" version %d not found", modelIdentifier, versionID)
 	default:
-		return &apiv1.GetModelResponse{Model: m},
-			errors.Wrapf(err, "error fetching model \"%s\" from database", req.ModelName)
+		return mv, errors.Wrapf(err,
+			"error fetching model \"%s\" version %d from database", modelIdentifier, versionID)
```
rb-determined-ai, line 60: suggestion: use `%q` instead of `\"%s\"` and `%v` instead of `%d`:
- `%q` is easier to read that `\"%s\"` and offers more correct quoting
- `%d` offers nothing that that `%v` doesn't offer, only `%d` is more strict (easier to mess up).

Another suggestion: reword to emphasize what was not found: `"error fetching version %v for model %q from database"`

rb-determined-ai (approved):
looks good to me

master/internal/api_model.go
mapmeld: will do :+1:

PR #3724, opened 2022-03-10
feat: remove global_batch_size requirement

rb-determined-ai:
## Description

Previously, the master enforced that the global_batch_size
hyperparameter existed, but in the future we will have training loops
that do not require this hyperparameter (DeepspeedTrial, Core API).

In the experiment config schemas, this PR simply drops an old
constraint, allowing configurations that were not previously allowed.

In the master code, this removes any ability to attempt to read the
global_batch_size hyperparameter in go code.

In python code, removes .per_slot_batch_size and .global_batch_size from
the internal EnvContext object and moves
EnvContext._calculate_batch_sizes to det.util, where it can be called by
each training loop that still requires the global_batch_size
hyperparameter.  This implies that what used to be a configuration error
(caught before an experiment id was even assigned) is now a runtime
error.  This is unfortunate but unavoidable; it should never have been
the master's responsibility to enforce training loop inputs.

## Test Plan

Added tests to ensure run-time checks apply to each of our current training loops.


liamcli:
looks good to me so far

harness/determined/keras/_tf_keras_context.py
```diff
@@ -417,5 +439,10 @@ class TFKerasExperimentalContext(_data_layer.DataLayerContext):
...
 
-    def __init__(self, env: det.EnvContext, distributed_context: _core.DistributedContext) -> None:
-        super().__init__(env=env, distributed_context=distributed_context)
+    def __init__(
+        self,
+        env: det.EnvContext,
+        distributed_context: _core.DistributedContext,
+        per_slot_batch_size: int,
+    ) -> None:
+        super().__init__(env, distributed_context, per_slot_batch_size)
```
ioga, line 448: this entire method looks redundant

harness/determined/keras/_tf_keras_context.py
```diff
@@ -417,5 +439,10 @@ class TFKerasExperimentalContext(_data_layer.DataLayerContext):
     the ``context.experimental`` namespace.
```
ioga, line 439: it doesn't extend `EstimatorTrialContext`

harness/determined/keras/_tf_keras_context.py
rb-determined-ai: yes, it sure looks like it is.

harness/determined/keras/_tf_keras_context.py
rb-determined-ai: no it sure doesn't

PR #3725, opened 2022-03-10
test: add test for cross-version master/agent clusters

dzhu:
## Description

For rolling upgrades to work properly, we need the master to be able to
communicate with agents from previous versions of Determined. (At least
for now, we're focusing on an upgrade strategy where the master is
upgraded first and the agents second.) And a first step toward ensuring
that is running tests on a cluster with mismatched versions, which this
adds using a custom devcluster config.

## Commentary

Strategy for determining which previous version(s) to use in general
TBD; perhaps 0.x.0 for each minor version, and maybe an occasional later
0.x.y, though that won't work right now because of a recent change to
logging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1434,12 +1434,16 @@ jobs:
       postgres-version:
         type: string
         default: '10'
+      agent-version:
+        type: string
+        default: ''
     machine:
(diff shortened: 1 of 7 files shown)


ioga (approved):
great!

PR #3728, opened 2022-03-11
feat: Add workspaces, projects, and related DB schema [DET-6815]

mapmeld:
## Description

Based on the ERD doc, adds tables to the database to support workspaces and projects.

- `username` fields via `user_id` + JOIN
- `lastExperimentStartedAt`, `numExperiments`, and `numActiveExperiments` fields would come from query of experiments with matching `project_id`
- Notes table joined in on their `project_id` field (unless we want these to be JSON attached to the Project table)
- `checkpoint_id` cannot refer to `checkpoints` directly because it's a view, not a table.

## Checklist

### DIFF
--- a/master/static/migrations/20220311095512_add_workspace_tables.up.sql
+++ b/master/static/migrations/20220311095512_add_workspace_tables.up.sql
@@ -1,0 +1,25 @@
+CREATE TABLE workspaces (
+  id SERIAL PRIMARY KEY,
+  name TEXT UNIQUE NOT NULL,
+  archived BOOLEAN,
+  user_id INT REFERENCES users(id)
+);
+CREATE TABLE projects (
+  id SERIAL PRIMARY KEY,
+  name TEXT UNIQUE NOT NULL,
+  description TEXT,
+  archived BOOLEAN,
+  workspace_id INT REFERENCES workspaces(id),
+  user_id INT REFERENCES users(id)
+);
+CREATE TABLE notes (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  contents TEXT,
+  project_id INT REFERENCES projects(id)
+);
+CREATE TABLE project_models (
+  name TEXT,
+  checkpoint_id INT
+);
+ALTER TABLE experiments ADD COLUMN project_id INT REFERENCES projects(id) NULL;
--- a/master/static/migrations/20220311095512_add_workspace_tables.up.sql
+++ b/master/static/migrations/20220311095512_add_workspace_tables.up.sql
@@ -4,21 +4,26 @@ CREATE TABLE workspaces (
   name TEXT UNIQUE NOT NULL,
   archived BOOLEAN,
   created_at timestamp with time zone NOT NULL DEFAULT NOW(),
-  user_id INT REFERENCES users(id)
+  user_id INT REFERENCES users(id),
+  immutable BOOLEAN DEFAULT FALSE
 );
 CREATE TABLE projects (
   id SERIAL PRIMARY KEY,
-  name TEXT UNIQUE NOT NULL,
+  name TEXT NOT NULL,
   description TEXT,
   archived BOOLEAN,
   created_at timestamp with time zone NOT NULL DEFAULT NOW(),
   notes JSONB,
   workspace_id INT REFERENCES workspaces(id),
-  user_id INT REFERENCES users(id)
+  user_id INT REFERENCES users(id),
+  immutable BOOLEAN DEFAULT FALSE,
+  UNIQUE (name, workspace_id)
 );
 CREATE TABLE project_models (
   name TEXT,
-  checkpoint_id INT
+  checkpoint_id INT,
+  created_at timestamp with time zone NOT NULL DEFAULT NOW(),
+  project_id INT REFERENCES projects(id)
 );
 ALTER TABLE experiments ADD COLUMN project_id INT REFERENCES projects(id) NULL;
 
@@ -26,6 +31,7 @@ ALTER TABLE experiments ADD COLUMN project_id INT REFERENCES projects(id) NULL;
 CREATE INDEX ix_experiments_project_id ON public.experiments USING btree (project_id);
 CREATE INDEX ix_projects_workspace_id ON public.projects USING btree (workspace_id);
 CREATE INDEX ix_projects_user_id ON public.projects USING btree (user_id);
+CREATE INDEX ix_project_models_project_id ON public.project_models USING btree (project_id);
 CREATE INDEX ix_workspaces_user_id ON public.workspaces USING btree (user_id);
 
 -- Default workspace and project objects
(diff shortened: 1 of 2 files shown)


EmilyBonar (changes requested):
This has the basics, but there's some stuff I think we need to address, in addition to the comments. Shouldn't the database indexes be created in add_workspace_tables.up? Also, I think it should include creating the default workspace and project (both named `Uncategorized`), and then setting all existing experiments to being within that project. Finally, I don't know enough SQL to know if this is possible, but if there's some way for those default workspaces and project to be immutable that would be good, so that people couldn't accidentally fuck up their environment.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE workspaces (
+  id SERIAL PRIMARY KEY,
+  name TEXT UNIQUE NOT NULL,
+  archived BOOLEAN,
+  user_id INT REFERENCES users(id)
```
hkang1, line 5: **question:** does this need creation time?

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,25 @@
...
+);
+CREATE TABLE projects (
+  id SERIAL PRIMARY KEY,
+  name TEXT UNIQUE NOT NULL,
+  description TEXT,
+  archived BOOLEAN,
+  workspace_id INT REFERENCES workspaces(id),
+  user_id INT REFERENCES users(id)
+);
```
hkang1, line 14: **question:** does this need creation time?

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE TABLE notes (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  contents TEXT,
+  project_id INT REFERENCES projects(id)
+);
+CREATE TABLE project_models (
+  name TEXT,
+  checkpoint_id INT
+);
```
hkang1, line 24: **question:** this table seems like it's missing a couple of things. I can't remember the full context but some that I can think of are:
```
project_id INT REFERENCES projects(id)
creation_time timestamp with time zone NOT NULL
```

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE workspaces (
+  id SERIAL PRIMARY KEY,
+  name TEXT UNIQUE NOT NULL,
```
hkang1, line 3: **note:** Looking at some of the other migration sql, for fields like this, it seems like the common pattern they use is...
```
name character varying UNIQUE NOT NULL
```
I think they might be the [same under the hood](https://stackoverflow.com/a/4849030/5402432), so implementation-wise it doesn't matter, but might be worth consideration if we want to keep them consistent.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,25 @@
...
+  archived BOOLEAN,
+  workspace_id INT REFERENCES workspaces(id),
+  user_id INT REFERENCES users(id)
+);
+CREATE TABLE notes (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  contents TEXT,
+  project_id INT REFERENCES projects(id)
+);
```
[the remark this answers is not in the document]
hkang1: Definitely not a DB expert but this seems like it's optimizing for pulling projects without notes and then joining when notes are necessary. I looked up some info regarding this and it seems like if project `notes` are `nullable` then querying the table without notes is [just as efficient](https://dba.stackexchange.com/a/18406/228845). cc @stoksc <== DB expert

master/static/migrations/20220311095512_add_workspace_tables.up.sql
stoksc: personally i prefer text unless you specifically need varchar(n)/char(n).

master/static/migrations/20220311095512_add_workspace_tables.up.sql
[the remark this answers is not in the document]
stoksc: The entire post @hkang1 linked is everything I would say and more. My guess would be a separate table makes sense unless every query that ever queries projects also always wants the notes, and we're sure that will be true basically forever.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
mapmeld: yes it does, I noticed when I was working on the 2nd part of this... I will add it to this migration to keep it cleaner

master/static/migrations/20220311095512_add_workspace_tables.up.sql
mapmeld: I will keep it as text, but thanks for checking on this. We don't have any set reason to limit name length with varchar type.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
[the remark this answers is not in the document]
mapmeld: I mentioned in the PR that I could see this being a JSONB column -- we plan to return the full Project from any API query of Projects. I will switch to JSONB

mapmeld:
@EmilyBonar I've added an `immutable` column (default false) to projects and workspace, and when adding the default / Uncategorized project, it'll have it set to true.  I'm assuming that the user_id here is that for the admin user, but maybe we can leave user_id = NULL ?

When we reach the DELETE ticket, I will suggest that we hide rather than actually delete projects or their descendant experiment relationships.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
mapmeld: added these in, plus an index for the project_id

master/static/migrations/20220311095512_add_workspace_tables.up.sql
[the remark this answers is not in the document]
stoksc: I would recommend keeping them separate, personally.

> we're sure that will be true basically forever.

like for this, i dont quite believe this will be true. e.g., if we do historical allocation aggregation queries and we want to join to projects/workspaces to aggregate over them, then separate would be better since that query won't need notes and postgres is a row store. for context that particular query is already the hardest on the system by a lot.

master/static/migrations/20220311095512_add_workspace_tables.up.sql
```diff
@@ -0,0 +1,53 @@
...
+  user_id INT REFERENCES users(id),
+  immutable BOOLEAN DEFAULT FALSE
+);
+CREATE TABLE projects (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  description TEXT,
+  archived BOOLEAN,
+  created_at timestamp with time zone NOT NULL DEFAULT NOW(),
+  notes JSONB,
```
EmilyBonar, line 16: I think we want this as a JSONB array. That would let us store multiple "pages" of notes as in the initial design.

EmilyBonar (approved):
As far as I can tell this looks good, but someone who has a better grasp of SQL should look it over as well.

PR #3732, opened 2022-03-14
fix: set allocation start time accurately (specifically for Kubernetees) [DET-6713]

nrajanee:
## Description

fix: set allocation start time accurately (specifically for Kubernetees) [DET-6713]

## Test Plan

Tested in unit test master/internal/task/allocation_test.go. No manual test is required. 

[DET-6713]: https://determinedai.atlassian.net/browse/DET-6713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -98,12 +98,14 @@ VALUES (:task_id, :allocation_id, :slots, :resource_pool, :agent_label, :start_t
 
 // CompleteAllocation persists the end of an allocation lifetime.
 func (db *PgDB) CompleteAllocation(a *model.Allocation) error {
-	if _, err := db.sql.Exec(`
-	UPDATE allocations
-	SET start_time = $1
-	WHERE start_time is NULL`, a.EndTime); err != nil {
-		return errors.Wrap(err,
-			"setting start time to cluster heartbeat when it's assigned to zero value")
+	if a.StartTime == nil {
+		if _, err := db.sql.Exec(`
+		UPDATE allocations
+		SET start_time = $2
+		WHERE allocation_id = $1`, a.AllocationID, a.EndTime); err != nil {
+			return errors.Wrap(err,
+				"setting start time to cluster heartbeat when it's assigned to zero value")
+		}
 	}
 
 	err := db.namedExecOne(`
@@ -112,7 +114,6 @@ SET end_time = :end_time
 WHERE allocation_id = :allocation_id
 `, a)
 	return err
-
 }
 
 // CompleteAllocationTelemetry returns the analytics of an allocation for the telemetry.
@@ -199,7 +200,7 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 	return err
 }
 
-// UpdateAllocationStartTime stores the latest start time
+// UpdateAllocationStartTime stores the latest start time.
 func (db *PgDB) UpdateAllocationStartTime(a model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -198,8 +198,8 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 func (db *PgDB) UpdateAllocationStartTime(a model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
-		SET start_time=$2
-		WHERE allocation_id=$1
+		SET start_time = $2
+		WHERE allocation_id = $1
 	`, a.AllocationID, a.StartTime)
 	return err
 }
--- a/master/internal/db/postgres_tasks.go
+++ b/master/internal/db/postgres_tasks.go
@@ -99,9 +99,9 @@ VALUES (:task_id, :allocation_id, :slots, :resource_pool, :agent_label, :start_t
 // CompleteAllocation persists the end of an allocation lifetime.
 func (db *PgDB) CompleteAllocation(a *model.Allocation) error {
 	if _, err := db.sql.Exec(`
-UPDATE allocations
-SET start_time = cluster_heartbeat FROM cluster_id
-WHERE start_time = $1`, time.Time{}); err != nil {
+	UPDATE allocations
+	SET start_time = $1
+	WHERE start_time is NULL`, a.EndTime); err != nil {
 		return errors.Wrap(err,
 			"setting start time to cluster heartbeat when it's assigned to zero value")
 	}
@@ -111,7 +111,6 @@ UPDATE allocations
 SET end_time = :end_time
 WHERE allocation_id = :allocation_id
 `, a)
-
 	return err
 
 }
@@ -201,7 +200,6 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 }
 
 // UpdateAllocationStartTime stores the latest start time
-
 func (db *PgDB) UpdateAllocationStartTime(a model.Allocation) error {
 	_, err := db.sql.Exec(`
 		UPDATE allocations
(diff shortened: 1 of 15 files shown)


master/internal/task/allocation.go
```diff
@@ -395,6 +395,7 @@ func (a *Allocation) TaskContainerStateChanged(
 	switch msg.Container.State {
 	case cproto.Pulling:
 		a.state = model.MostProgressedAllocationState(a.state, model.AllocationStatePulling)
+		a.model.StartTime = time.Now().UTC().Truncate(time.Millisecond)
```
stoksc, line 398: we need to make sure this gets saved immediately, i'm not sure when the next time the model is sync is

master/internal/db/postgres_tasks.go
```diff
@@ -98,11 +98,22 @@ VALUES (:task_id, :allocation_id, :slots, :resource_pool, :agent_label, :start_t
...
+	}
+
+	err := db.namedExecOne(`
 UPDATE allocations
 SET end_time = :end_time
 WHERE allocation_id = :allocation_id
 `, a)
+
+	return err
+
```
stoksc, line 116: i'm not sure i understand these changes, and the others in this file. i think we should just be setting start time in allocation.go and saving it.

master/internal/task/allocation.go
nrajanee: Got it! I see an UpdateAllocationState method in master/db/database.go but that only updates the allocation state. Should I create a similar method that updates the start time?

master/internal/db/postgres_tasks.go
nrajanee: These changes are related to these concerns we had when we initially discussed the implementation: - CompleteAllocation if start time is zero value set to cluster_heartbeat. This can happen when we call cleanup after a Poststop message. This message is sent after panic or error. 
- CloseOpenAllocations if start time is zero value set to cluster_heartbeat.  This can happen when cluster fails. 

If we don't do this then we can end up with a zero value in start time and end_time as the current value or cluster_heartbeat which is incorrect.

master/internal/task/allocation.go
```diff
@@ -134,7 +134,7 @@ func NewAllocation(
 			Slots:        req.SlotsNeeded,
 			AgentLabel:   req.Name,
 			ResourcePool: req.ResourcePool,
-			StartTime:    time.Now().UTC(),
+			StartTime:    time.Time{},
```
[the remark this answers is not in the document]
nrajanee: done

master/internal/task/allocation.go
nrajanee: done. Should I add a check for this to the postgres_task_intg_test.go?

master/internal/db/postgres_tasks.go
```diff
@@ -99,20 +99,14 @@ VALUES (:task_id, :allocation_id, :slots, :resource_pool, :agent_label, :start_t
...
-		}
+		a.StartTime = a.EndTime
 	}
 
-	err := db.namedExecOne(`
-UPDATE allocations
-SET end_time = :end_time
-WHERE allocation_id = :allocation_id
-`, a)
+	_, err := db.sql.Exec(`
```
stoksc, line 105: the diff between this and the former version is that we used to make sure w only affected 1 row. maybe leave that check in (nice to know it happens somewhere, even those it _should_ never be broken).

master/internal/db/postgres_tasks.go
```diff
@@ -189,10 +194,28 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
 	return err
 }
 
+// UpdateAllocationStartTime stores the latest start time.
+func (db *PgDB) UpdateAllocationStartTime(a model.Allocation) error {
+	_, err := db.sql.Exec(`
+		UPDATE allocations
+		SET state=$2
```
stoksc, line 201:
```suggestion
		SET start_time=$2
```

master/internal/db/postgres_tasks.go
```diff
@@ -189,10 +194,28 @@ func (db *PgDB) UpdateAllocationState(a model.Allocation) error {
...
+		WHERE allocation_id=$1
+	`, a.AllocationID, a.StartTime)
+	return err
+}
+
 // CloseOpenAllocations finds all allocations that were open when the master crashed
 // and adds an end time.
 func (db *PgDB) CloseOpenAllocations() error {
 	if _, err := db.sql.Exec(`
+	UPDATE allocations
```
stoksc, line 211: nit, here and elsewhere format the queries consistently. also i would run these queries in a transaction.

master/internal/task/allocation.go
```diff
@@ -395,6 +394,11 @@ func (a *Allocation) TaskContainerStateChanged(
 	switch msg.Container.State {
 	case cproto.Pulling:
 		a.state = model.MostProgressedAllocationState(a.state, model.AllocationStatePulling)
+		a.model.StartTime = ptrs.TimePtr(time.Now().UTC().Truncate(time.Millisecond))
+		if err := a.db.UpdateAllocationStartTime(a.model); err != nil {
```
stoksc, line 398: i think we can just log this error and move on. log an error that says something along the lines of 'this allocation maybe not be properly accounted for'.

master/internal/task/allocation_test.go
```diff
@@ -102,8 +103,17 @@ func TestAllocation(t *testing.T) {
...
 				require.NoError(t, system.Ask(self, containerStateChanged).Error())
+
+				beforePulling := time.Now().UTC().Truncate(time.Millisecond)
 				containerStateChanged.Container.State = cproto.Pulling
+				db.On("UpdateAllocationStartTime", mock.Anything).Return(nil)
 				require.NoError(t, system.Ask(self, containerStateChanged).Error())
+				afterPulling := time.Now().UTC().Truncate(time.Millisecond)
+				outOfRange := a.model.StartTime.Before(beforePulling) || a.model.StartTime.After(afterPulling)
+				require.False(t, outOfRange,
+					"Expected start time of open allocation should be in between %q and %q but it is = %q instead",
```
stoksc, line 114: if you use a format string you need to use `require.Falsef` instead. and you can drop the `.String()` calls and just use `%s` i think.

master/internal/db/postgres_tasks.go
nrajanee: But we're using allocation_id here so that means we'll only affect one row?
Leave the check in as in check for a.startTime == nil? Then'll I'll have to have two separate queries like before?

nrajanee: > the diff between this and the former version is that we used to make sure w only affected 1 row. maybe leave that check in (nice to know it happens somewhere, even those it _should_ never be broken).

But we're using allocation_id here so that means we'll only affect one row?
Leave the check in as in check for a.startTime == nil? Then'll I'll have to have two separate queries like before?

master/internal/db/postgres_tasks.go
nrajanee: Oops missed this! thank you.

master/internal/db/postgres_tasks.go
nrajanee: will do!

master/internal/db/postgres_tasks.go
stoksc: no i mean leave the check that we only affected one row.

> But we're using allocation_id here so that means we'll only affect one row?

true yeah, i was being extra cautious since id is generated by us but i think there is a unique constraint on id so it's ok.

master/internal/db/postgres_tasks.go
nrajanee: locally, ran the queries in a transaction and they work fine!

stoksc (approved):
lgtm modulo final comments

stoksc:
@nrajanee make sure to rebase and ensure your migration is last still before landing

master/static/migrations/20220315162743_drop-not-null-allocation-start-time.tx.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE allocations
+   ALTER column start_time SET NOT NULL; 
+
```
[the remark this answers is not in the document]
stoksc: @nrajanee i think this got missed, fyi

PR #3733, opened 2022-03-14
fix: always show progress bar for experiments

johnkim-det:
## Description

In the experiment details header and in the progress column of the experiment list table:
instead of hiding progress bar based on experiment state or 0 progress value, always show progress bar.

## Test Plan
Make sure an experiment's progress bar appears even if state is not 'Active', 'Scheduled', or 'ScheduledBackfilled', and/or if it's 0%:

- [ ] on the experiment details header

Screen Shot 2022-03-14 at 3 27 19 PM
canceled_progress
inprogress
zero_progress

- [ ] in the experiment list table's progress column
Screen Shot 2022-03-14 at 3 22 21 PM

## Checklist


hamidzr:
(here as part of the release party test)

> and/or if it's 0%:

IIRC it was the case that the progress bar would reset to 0% where it should have been 100%. If that's still the case this could add to the confusion. Do you know if that was changed? 

On another point: I read the ticket but I'm not sure if we wanted dashboard cards to be included or not. Although this works on its own as well

PR #3737, opened 2022-03-15
fix: Restore det preview-search in CLI [DET-6850]

mapmeld:
## Description

Returns an API endpoint which is used by the CLI's `det preview-search`. Re-implementing the code in the new API would take an additional time commitment.

## Test Plan

`det preview-search path/to/configfile.yaml`

## Checklist


stoksc (approved):
can you make a ticket for moving the CLI over to the new one?

PR #3738, opened 2022-03-15
fix: create user with password broken [DET-6852]

varlogtim:
## Test Plan

Create a user with a password via the REST API.

```
# Set a curl'able master URL:
export DET_MASTER=http://localhost:8080

# Login as admin
export TOKEN=$(curl -X POST -d '{"username": "admin", "password": ""}' ${DET_MASTER}/api/v1/auth/login | jq '.token' | sed 's/"//g')

# Create a user with password:
curl -H "Authorization: Bearer ${TOKEN}" -H  "accept: application/json" -X POST "${DET_MASTER}/api/v1/users" -d '{"user": {"username": "mytestuser", "admin": false, "active": true}, "password": "asdf"}'

# Try to login as the user using the password "asdf"
det user login mytestuser
```

### DIFF
--- a/master/internal/api_user.go
+++ b/master/internal/api_user.go
@@ -88,6 +88,7 @@ func (a *apiServer) GetUser(
 func (a *apiServer) PostUser(
 	ctx context.Context, req *apiv1.PostUserRequest) (*apiv1.PostUserResponse, error) {
 	curUser, _, err := grpcutil.GetUser(ctx, a.m.db, &a.m.config.InternalConfig.ExternalSessions)
+
 	if err != nil {
 		return nil, err
 	}
@@ -105,7 +106,7 @@ func (a *apiServer) PostUser(
 		Admin:    req.User.Admin,
 		Active:   req.User.Active,
 	}
-	if err = user.UpdatePasswordHash(req.Password); err != nil {
+	if err = user.UpdatePasswordHash(replicateClientSideSaltAndHash(req.Password)); err != nil {
 		return nil, err
 	}
 	var agentUserGroup *model.AgentUserGroup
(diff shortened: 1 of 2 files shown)


stoksc (approved):
approved by @mapmeld should look too

master/pkg/model/test_utils.go
```diff
@@ -37,7 +37,7 @@ func ExperimentModel(opts ...ExperimentModelOption) *Experiment {
 				RawMaxLength: &maxLength,
 			},
 		},
-		RawEntrypoint: &expconf.Entrypoint{RawEntrypoint: "model_def:SomeTrialClass"},
+		RawEntrypoint:      &expconf.Entrypoint{RawEntrypoint: "model_def:SomeTrialClass"},
```
mapmeld, line 40: I don't know if this is picked up by lint, but I think you can remove this change and keep the file unchanged

mapmeld:
I think I understand, you're creating the user from the command line in one POST request.

[The CLI also calls `POST /api/v1/users`](https://github.com/determined-ai/determined/blob/master/harness/determined/cli/user.py#L172) but without a password so we hadn't checked this issue before

master/pkg/model/test_utils.go
varlogtim: I feel like this was caused by lint as I didn't make it.

varlogtim:
> I think I understand, you're creating the user from the command line in one POST request.
> 
> [The CLI also calls `POST /api/v1/users`](https://github.com/determined-ai/determined/blob/master/harness/determined/cli/user.py#L172) but without a password so we hadn't checked this issue before

More details: an OSS user asked why their POST user with password didn't work, I glanced at the endpoint and it looked like it was doing stuff with the password and should work, I verified that it did not work, now we are here.

I am thinking the person who made the CLI used the same kind of principals as most linux tools, i.e., always prompt for passwords.

mapmeld (approved):

master/pkg/model/test_utils.go
stoksc: this is probably lint not getting run with tags properly or something. i'll fix.

PR #3740, opened 2022-03-16
feat: Create and read endpoints for workspaces and projects  [DET-6816]

mapmeld:
## Description

Adds first set of API endpoints for Workspaces:
- GET /workspaces
- POST /workspaces
- GET /workspaces/{id}
- GET /workspaces/{id}/projects
- POST /workspaces/{workspace_id}/projects
- GET /projects/{id}
- GET /projects/{id}/experiments
- POST /projects/{id}/notes

Includes a new file in `e2e_tests` which posts and gets sample workspaces, projects, and notes, including testing sort order.

## Test Plan

`cd e2e_tests && pytest tests/cluster/test_workspace_org.py`

We don't have a delete feature yet. How to clear test objects before re-running the e2e tests:

```sql
DELETE FROM projects WHERE name LIKE '_Test%';
DELETE FROM workspaces WHERE name LIKE '_Test%';
```

## Commentary

- We aren't making Workspaces visible yet, so I don't think we need a release note on it.
- Adds `project_id` to `get_experiments.sql` to reuse it - should not change behavior of existing experiment API
- `api_workspace.go` and `api_project.go` can be combined into one file if it seems cleaner.

## Checklist


master/internal/db/postgres.go
```diff
@@ -250,6 +252,19 @@ func (db *PgDB) RawQuery(queryName string, params ...interface{}) ([]byte, error
 	return db.rawQuery(db.queries.getOrLoad(queryName), params...)
 }
 
+func (db *PgDB) GetWorkspace(ctx context.Context, id int32) (*workspacev1.Workspace, error) {
+	w := &workspacev1.Workspace{}
```
rb-determined-ai, line 256: @stoksc @ioga What do you think about using the proto object directly as the bun model?

Pros I see:
- It's simpler, at least when it works.

Cons I see:
- It violates the separation of user api from data model implementation

stoksc: i think we shouldn't at all. it will not work in general and is more susceptible to weird bugs (weird fields aren't mapped right/are mapped at all). separate models are inevitable the first time the models diverge, so just make them separate, otherwise the first time they diverge may just result in a bug.

stoksc: also, you can't change generated models so you can't embed a `bun.BaseModel` in it. does this even work?

ioga: I agree with @stoksc , that's not going to work as a general solution.

stoksc: https://github.com/determined-ai/ci-shortcuts/blob/master/cibuild/models.go#L28-L53 is the pattern we should be using, imo

PR #3741, opened 2022-03-16
chore: updates to deepspeed launcher

azhou-determined:
## Description

- util function for socket sshd checks
- wrap chief worker's SSHD in a context to terminate on finish
## Checklist

### DIFF
--- a/harness/determined/launch/autodeepspeed.py
+++ b/harness/determined/launch/autodeepspeed.py
@@ -7,7 +7,6 @@ import argparse
 import logging
 import os
 import pathlib
-import socket
 import subprocess
 import sys
 import time
@@ -16,7 +15,7 @@ from typing import List
 from deepspeed.launcher.runner import DEEPSPEED_ENVIRONMENT_NAME
 
 import determined as det
-from determined import constants
+from determined import constants, util
 from determined.common import api
 from determined.common.api import certs
 
@@ -139,46 +138,6 @@ def main(train_entrypoint: str) -> int:
     # slot case.
     os.environ["USE_DEEPSPEED"] = "1"
 
-    if len(info.container_addrs) > 1:
-        # Create the environment file that will be passed by deepspeed to individual ranks.
-        create_deepspeed_env_file()
-        # Set custom PDSH args:
-        # * bypass strict host checking
-        # * -p our custom port
-        # * other args are default ssh args for pdsh
-        os.environ["PDSH_SSH_ARGS"] = (
-            "-o PasswordAuthentication=no -o StrictHostKeyChecking=no "
-            f"-p {constants.DTRAIN_SSH_PORT} -2 -a -x %h"
-        )
-
-        # Chief worker also needs to run sshd when using pdsh.
-        sshd_process = subprocess.Popen(run_sshd_command)
-
-        # Chief machine waits for every worker's sshd to be available.  All machines should be
-        # close to in-step by now because all machines just finished synchronizing rendezvous info.
-        deadline = time.time() + 20
-        for peer in info.container_addrs:
-            while True:
-                with socket.socket() as sock:
-                    sock.settimeout(1)
-                    try:
-                        # Connect to a socket to ensure sshd is listening.
-                        sock.connect((peer, constants.DTRAIN_SSH_PORT))
-                        # The ssh protocol requires the server to serve an initial greeting.
-                        # Receive part of that greeting to know that sshd is accepting/responding.
-                        data = sock.recv(1)
-                        if not data:
-                            raise ValueError("no sshd greeting")
-                        # This peer is ready.
-                        break
-                    except Exception:
-                        if time.time() > deadline:
-                            raise ValueError(
-                                f"Chief machine was unable to connect to sshd on peer machine at "
-                                f"{peer}:{constants.DTRAIN_SSH_PORT}"
-                            )
-                        time.sleep(0.1)
-
     # The chief has several layers of wrapper processes:
     # - a top-level pid_server, which causes the whole container to exit if any local worker dies.
     # - deepspeed, which launches $slots_per_trial copies of the following layers:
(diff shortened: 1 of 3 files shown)


harness/determined/launch/autodeepspeed.py
```diff
@@ -69,6 +71,34 @@ def create_run_command(master_address: str, hostfile_path: str) -> List[str]:
...
+
+    sshd_process = subprocess.Popen(sshd_cmd)
+
+    # Chief machine waits for every worker's sshd to be available.  All machines should be
+    # close to in-step by now because all machines just finished synchronizing rendezvous info.
+    check_sshd(addrs, time.time() + 20, constants.DTRAIN_SSH_PORT)
+    try:
+        yield
+    finally:
+        sshd_process.send_signal(signal.SIGKILL)
```
rb-determined-ai, line 98: let's use `sshd_process.kill()` instead, which is the more common way to [do this](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate).

harness/determined/launch/autodeepspeed.py
```diff
@@ -4,21 +4,23 @@
...
-from typing import List
+from typing import Iterator, List
 
 from deepspeed.launcher.runner import DEEPSPEED_ENVIRONMENT_NAME
 
 import determined as det
 from determined import constants
 from determined.common import api
 from determined.common.api import certs
+from determined.util import check_sshd
```
rb-determined-ai, line 23: please follow [google style guide for python imports](https://google.github.io/styleguide/pyguide.html#22-imports).

```
from determined import util
...
util.check_sshd(...)
```

harness/determined/launch/autodeepspeed.py
```diff
@@ -69,6 +71,34 @@ def create_run_command(master_address: str, hostfile_path: str) -> List[str]:
...
+    # Create the environment file that will be passed by deepspeed to individual ranks.
+    create_deepspeed_env_file()
+    # Set custom PDSH args:
+    # * bypass strict host checking
+    # * -p our custom port
+    # * other args are default ssh args for pdsh
+    os.environ["PDSH_SSH_ARGS"] = (
+        "-o PasswordAuthentication=no -o StrictHostKeyChecking=no "
+        f"-p {constants.DTRAIN_SSH_PORT} -2 -a -x %h"
+    )
```
rb-determined-ai, line 88: why is this stuff in this context manager?

harness/determined/util.py
```diff
@@ -293,3 +294,35 @@ def calculate_batch_sizes(
         )
 
     return per_gpu_batch_size, effective_batch_size
+
+
+def check_sshd(peers: List[str], deadline: float, port: int) -> None:
```
rb-determined-ai, line 299: can you make this function operate on a single peer, and call it in a loop when you want to run it against multiple peers.

That makes it a bit more composable I think, which is good for helper functions.

azhou-determined: I initially did it by single peer, but I figured maybe the deadline would make more sense in the context of a peer group/rendezvous. But you're right, single peer is probably better for a util function.

harness/determined/launch/autodeepspeed.py
```diff
@@ -237,9 +227,12 @@ def main(train_entrypoint: str) -> int:
...
 
-    exit_code = subprocess.Popen(full_cmd).wait()
-
-    return exit_code
+    # Chief worker also needs to run sshd when using pdsh and multi-machine training.
+    multi_machine = len(info.container_addrs) > 1
+    with run_sshd(
+        info.container_addrs, run_sshd_command
+    ) if multi_machine else contextlib.nullcontext():
+        return subprocess.Popen(full_cmd).wait()
```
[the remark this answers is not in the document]
azhou-determined: was trying to avoid calling `subprocess.Popen(full_cmd).wait()` twice, but yeah it's probably overkill

harness/determined/launch/autohorovod.py
```diff
@@ -90,28 +89,8 @@ def main(train_entrypoint: str) -> int:
...
-                    break
-                except Exception:
-                    if time.time() > deadline:
-                        raise ValueError(
-                            f"Chief machine was unable to connect to sshd on peer machine at "
-                            f"{peer}:{DTRAIN_SSH_PORT}"
-                        )
-                    time.sleep(0.1)
+    for peer_addr in info.container_addrs[1:]:
+        util.check_sshd(peer_addr, time.time() + 5, DTRAIN_SSH_PORT)
```
rb-determined-ai, line 93: Let's not modify the deadline code.  20 seconds for all workers to become ready is not equivalent to 5 seconds for the first worker to come up, 5 more seconds for the next worker, etc.

Set a deadline once before the loop and pass it in repeatedly.

harness/determined/launch/autodeepspeed.py
```diff
@@ -237,9 +196,36 @@ def main(train_entrypoint: str) -> int:
...
+        sshd_process = subprocess.Popen(run_sshd_command)
+
+        try:
+            # Chief machine waits for every worker's sshd to be available.  All machines should be
+            # close to in-step by now because all machines just finished synchronizing rendezvous
+            # info.
+            for peer_addr in info.container_addrs:
+                util.check_sshd(peer_addr, time.time() + 5, constants.DTRAIN_SSH_PORT)
+
+            return_code = subprocess.Popen(full_cmd).wait()
```
rb-determined-ai, line 222: just return here, the `finally` clause will still work

harness/determined/launch/autodeepspeed.py
```diff
@@ -237,9 +196,36 @@ def main(train_entrypoint: str) -> int:
...
+        finally:
+            sshd_process.kill()
+            sshd_process.wait()
+        return return_code
 
-    return exit_code
+    return subprocess.Popen(full_cmd).wait()
```
rb-determined-ai, line 228: very small nitpick:

since this line is so much smaller than the above function, can you invert the if logic?

```python
if not multi_machine:
    return subprocess.Popen(full_cmd).wait()
...
# Create the environment file that will be passed by deepspeed to individual ranks.
create_deepspeed_env_file()
```

harness/determined/launch/autodeepspeed.py
```diff
@@ -237,9 +196,36 @@ def main(train_entrypoint: str) -> int:
 
     full_cmd = pid_server_cmd + cmd + pid_client_cmd + log_redirect_cmd + harness_cmd
 
-    exit_code = subprocess.Popen(full_cmd).wait()
+    # Chief worker also needs to run sshd when using pdsh and multi-machine training.
```
rb-determined-ai, line 199: tiny nit: move this comment under the if statement, since it only sometimes applies on this line, but it always applies after the if statement

rb-determined-ai (approved):
looks great

PR #3742, opened 2022-03-16
fix: use displayName in tables

johnkim-det:
## Description

Filter tables by user should use displayName if it exists: ModelRegistry, TaskList
Sort tables by user should use displayName if it exists: ExperimentList, TaskList

## Test Plan

- [ ] verify that sorting by user uses displayNames on ExperimentList and TaskList tables
- [ ] verify that filtering by user shows displayNames on ModelRegistry and TaskList tables

## Checklist


proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -73,6 +73,8 @@ message Experiment {
   bool archived = 7;
   // The number of trials linked to the experiment.
   int32 num_trials = 8;
+  // The display name of the user that created the experiment.
+  string display_name = 18;
```
mapmeld, line 77: If I understand this right, tasks needed to have a displayName on them to be sorted client-side,

If that's the case , I don't think that experiments need it?

johnkim-det: i think i added this because i thought the backend sorting needed it too. if adding it to `get_experiments.sql` is enough for the QueryProtof call to be sorted, I'll go ahead and remove.

johnkim-det: ok looks like it was necessary so keeping that change in

mapmeld (approved):

PR #3743, opened 2022-03-16
fix: replace loading skeleton with empty state [DET-6800]

EmilyBonar:
## Description
The skeletons we were using right after creating an experiment gave the impression that the page was waiting for data to load, when in fact the situation was that there isn't any data to show. The skeletons have been replaced with an empty state.

## Test Plan
To test, fork an experiment in the webui. You should see a page similar to the "New state" screenshot below, not the "Old state".

Old state:
image

New state:
image

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
+++ b/webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
@@ -14,13 +14,14 @@ import { checkpointSize } from 'utils/workload';
 
 interface Props {
   experiment: ExperimentBase;
-  trial: TrialDetails;
+  trial?: TrialDetails;
 }
 
 const TrialInfoBox: React.FC<Props> = ({ trial, experiment }: Props) => {
   const [ showBestCheckpoint, setShowBestCheckpoint ] = useState(false);
 
   const bestCheckpoint: CheckpointDetail | undefined = useMemo(() => {
+    if (!trial) return;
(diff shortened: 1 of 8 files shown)


webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
```diff
@@ -30,34 +31,39 @@ const TrialInfoBox: React.FC<Props> = ({ trial, experiment }: Props) => {
...
+          <OverviewStats title="Start Time">
+            <TimeAgo datetime={trial.startTime} />
+          </OverviewStats>
+        )}
+        {totalCheckpointsSize && (
+          <OverviewStats title="Total Checkpoint Size">
+            {totalCheckpointsSize}
+          </OverviewStats>
+        )}
         {bestCheckpoint && (
```
johnkim-det, line 67: question: would something like `{bestCheckpoint && trial && (` make sense here? instead of `trial?.id` in title string?

johnkim-det (approved):
lgtm, one minor suggestion

webui/react/src/pages/TrialDetails/TrialInfoBox.tsx
EmilyBonar: Yeah, that makes sense. I don't think there's any way for bestCheckpoint to exist without trial existing, but better to be explicit.

PR #3744, opened 2022-03-16
docs: remove some outdated information.

neilconway:
Information on how to upgrade from pre-0.15.0 releases is no longer useful.

Also remove docs on how to deal with a relatively obscure distutils error; the
error is a property of distutils, anyway, and is not specific to Determined.

## Checklist


gh-determined-ai (approved):
Nice cleanup in general. Can more people do this?
Old release notes are always likely to be out of date, eventually. That's why links into docs are no longer permitted. The policy should be that old release notes are note updated but care is taken to make them future-safe.

neilconway:
The failing lint check is unrelated, so I'll land this.

PR #3747, opened 2022-03-17
fix: prevent pytorch geometric nightly test timeout

coreystatendet:
## Description

Prevent pytorch geometric nightly test timeout

## Test Plan
`test_protein_pytorch_geometric` in nightly E2E GPU job should run successfully.

No testing needed for release / outside of CI.

## Commentary (optional)
Needed to update image/wheel for install to be quick enough, due to order of installs and incompatibility with the current `torch-geometric` version.

Also removed `cpu` images since it appears CUDA is required to run the model.


ioga (approved):
lgtm

PR #3748, opened 2022-03-17
chore: better log metadata

stoksc:
## Description
This change tries to ensure it is easier to string compotents' logs together. The general strategy is each child component also logs it's parent, so you can join them like with FKs in postgres.

## Test Plan
- [x] run master, :eye: logs

## Commentary (optional)
pros: @varlogtim can maybe easier grep the history of a task.
cons: really big logs

i'd love if we just do this and move to json structured logs one day and return those to the uis and let them solve the 'how to read this crap' problem. same for trial logs.

## Checklist

### DIFF
--- a/agent/internal/container.go
+++ b/agent/internal/container.go
@@ -235,7 +235,9 @@ func (c *containerActor) containerStopped(ctx *actor.Context, stopped aproto.Con
 		return
 	}
 
-	ctx.Log().Infof("transitioning state from %s to %s", c.State, newState)
+	ctx.Log().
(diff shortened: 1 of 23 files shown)


rb-determined-ai (approved):

ioga (approved):
I am somewhat concerned the log context management (i.e. passing it around, merging it here and there) is very manual. There's no easy pattern to follow around.

stoksc:
yeah, i think eventually we could sneak the entire object into a `context.Context` and pass it around that way.

PR #3749, opened 2022-03-17
fix: check_idle.py is quiet when jupyter takes a while to start up

rb-determined-ai:
Previously, when jupyter took too long to start, check_idle.py would
inject a terrifying stacktrace into the notebook logs.

## Test Plan:

Tested manually by putting a `sleep 15` into `notebook-entrypoint.sh`.

### DIFF
--- a/master/static/srv/check_idle.py
+++ b/master/static/srv/check_idle.py
@@ -1,7 +1,8 @@
 import enum
 import logging
 import os
-from time import sleep
+import socket
+import time
 
 import requests
 from determined.common import api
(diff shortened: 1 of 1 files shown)


master/static/srv/check_idle.py
```diff
@@ -57,21 +79,20 @@ def main():
         )
         idle_type = IdleType.KERNELS_OR_TERMINALS
 
-    while True:
-        sleep(1)
+    wait_for_jupyter(("127.0.0.1", int(port)))
```
mpkouznetsov, line 82: 127.0.0.1 ???

mpkouznetsov (approved):
Please appease the linter and put in the _real_ address.
Then it will be good to go.

master/static/srv/check_idle.py
rb-determined-ai: Yes, this is inside the container, and jupyter is also inside the container.

`127.0.0.1` is used for the notebook_server address as well.

PR #3751, opened 2022-03-18
build: use mockery v2.10.0 which is go1.18-friendly

rb-determined-ai:
## Commentary

I accidentally ran `pacman -Syu` and now I can't build master without fixing this.


stoksc (approved):
confirmed this doesn't cause issues when still on go 1.17

PR #3755, opened 2022-03-18
chore: Update required fields, remove repeat fields from model registry [DET-6861] 

mapmeld:
## Description

- Required fields in Model Registry's request and response objects are now labeled as required for typing purposes.
- In `PATCH /api/v1/models/:model_name`, the request body would be a PatchModel instead of a PatchModelRequest wrapper
- In `PATCH /api/v1/models/:model_name/versions/:version_id`, the request body will be a PatchModelVersion instead of a PatchModelVersionRequest wrapper
- Removes username field from Model. Models continue to be associated with the signed-in user.

## Commentary

If this seems like a good change, I plan to write a release note on the changes:

- the big breaking change is PATCH Model and ModelVersion, and it's optional if we want to leave it alone
- for other API requests, if someone has a functioning script, it already fills any required fields (for example: GetModelRequest now required model_name, which was already in the URL).
- for API responses, typing is now one correct response-wrapper type instead of `ResponseObject or None`, this makes our bindings a little neater
- username is not a breaking change because it wasn't used before and won't throw an error if added now

## Checklist


ioga (approved):
Don't know if there was a decision on the breaking change, but the code lgtm.

EmilyBonar (approved):
LGTM

gh-determined-ai (approved):

PR #3760, opened 2022-03-19
chore: add mock tests for deepspeed launcher

azhou-determined:
## Description

mock tests for deepspeed launcher. a small refactor of command methods will follow. 

## Checklist

### DIFF
--- a/harness/determined/launch/autodeepspeed.py
+++ b/harness/determined/launch/autodeepspeed.py
@@ -38,6 +38,64 @@ def create_hostlist_file(
     return trial_runner_hosts[0]
 
 
+def create_pid_server_cmd(allocation_id: str, num_workers: int) -> List[str]:
+    return [
+        "python3",
+        "-m",
+        "determined.exec.pid_server",
+        "--on-fail",
+        "SIGTERM",
+        "--on-exit",
+        "SIGTERM",
+        "--grace-period",
+        "5",
+        f"/tmp/pid_server-{allocation_id}",
+        str(num_workers),
+        "--",
+    ]
+
+
+def create_pid_client_cmd(allocation_id: str) -> List[str]:
+    return [
+        "python3",
+        "-m",
+        "determined.exec.pid_client",
+        f"/tmp/pid_server-{allocation_id}",
+        "--",
+    ]
+
+
+def create_log_redirect_cmd() -> List[str]:
+    return [
+        "python3",
+        "-m",
+        "determined.exec.worker_process_wrapper",
+        "RANK",
+        "--",
+    ]
+
+
+def create_harness_cmd(train_entrypoint: str) -> List[str]:
+    return [
+        "python3",
+        "-m",
+        "determined.exec.harness",
+        "--train-entrypoint",
+        train_entrypoint,
+    ]
+
+
+def create_sshd_cmd() -> List[str]:
+    return [
+        "/usr/sbin/sshd",
+        "-p",
+        str(constants.DTRAIN_SSH_PORT),
+        "-f",
+        "/run/determined/ssh/sshd_config",
+        "-D",
+    ]
+
+
 def create_deepspeed_env_file() -> None:
     """Create an env var export file to pass Determined vars to the deepspeed launcher.
 
@@ -92,14 +150,7 @@ def main(train_entrypoint: str) -> int:
     os.environ["DET_CHIEF_IP"] = chief_ip
 
     # All ranks will need to run sshd.
-    run_sshd_command = [
-        "/usr/sbin/sshd",
-        "-p",
-        str(constants.DTRAIN_SSH_PORT),
-        "-f",
-        "/run/determined/ssh/sshd_config",
-        "-D",
-    ]
+    run_sshd_command = create_sshd_cmd()
 
     if info.container_rank > 0:
         # Non-chief machines just run sshd.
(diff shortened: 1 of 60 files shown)


harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,229 @@
+import contextlib
+import os
+import time
+from typing import Any, Iterator, List
+from unittest.mock import MagicMock, call, patch
```
rb-determined-ai, line 5: [google style guide](https://google.github.io/styleguide/pyguide.html#22-imports)

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,229 @@
...
+        "-p",
+        str(constants.DTRAIN_SSH_PORT),
+        "-f",
+        "/run/determined/ssh/sshd_config",
+        "-D",
+    ]
+    mock_subprocess.assert_called_once_with(expected_cmd)
+
+
+def get_mock_cluster_info(container_addrs: List[str], container_rank: int) -> ClusterInfo:
```
rb-determined-ai, line 205: rename to `make_mock_cluster_info()` since it is not a getter at all

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,229 @@
...
+        allocation_id="allocationId",
+        session_token="sessionToken",
+        task_type="TRIAL",
+        rendezvous_info=rendezvous_info_mock,
+    )
+    return cluster_info_mock
+
+
+@contextlib.contextmanager
+def set_env_vars() -> Iterator[None]:
```
rb-determined-ai, line 224: rename to `set_container_id_env_var` or something, so its name matches what it does

harness/determined/launch/autodeepspeed.py
```diff
@@ -19,6 +19,8 @@
 from determined.common import api
 from determined.common.api import certs
 
+hostfile_path = "/tmp/hostfile.txt"
```
rb-determined-ai, line 22: slight nit: maybe just pass this into `main()` from from the `if __name__ == "__main__"` clause?  This shouldn't really need to be overwritten as part of the test.

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,229 @@
...
+    sshd_cmd = [
+        "/usr/sbin/sshd",
+        "-p",
+        str(constants.DTRAIN_SSH_PORT),
+        "-f",
+        "/run/determined/ssh/sshd_config",
+        "-D",
+    ]
+
+    launch_cmd = [
```
rb-determined-ai, line 35: nit: break this list into chunks with comments to make it easier to read.

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,229 @@
...
+
+    launch_proc_mock().wait.assert_called_once()
+
+    sshd_proc_mock().kill.assert_called_once()
+    sshd_proc_mock().wait.assert_called_once()
+
+
+@patch("subprocess.Popen")
+@patch("determined.get_cluster_info")
+def test_launch_chief(mock_cluster_info: MagicMock, mock_subprocess: MagicMock) -> None:
```
rb-determined-ai, line 114: rename to `test_launch_one_slot`, maybe?

rb-determined-ai (commented):
This looks pretty good.  I'd like to see some tests for failure cases too.

harness/determined/launch/autodeepspeed.py
azhou-determined: feels like this belongs more as a constant in the class? rather than having to pass in the same path from a test.

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,359 @@
+import contextlib
+import os
+import time
+import unittest.mock as mock
+from typing import Any, Iterator, List
+
+import pytest
+
+from determined import ClusterInfo, RendezvousInfo, constants
```
rb-determined-ai, line 9: google style guide.

```python
import determined as det
from determined import constants

det.ClusterInfo
constants.WHATEVER_CONSTANT
```

harness/determined/launch/autodeepspeed.py
rb-determined-ai: I guess I'd assume you'd pass in a different value than `/tmp/hostfile.txt`.  But I guess `/tmp` is basically always writable, so maybe it doesn't matter at all.

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,359 @@
...
+        if cmd == sshd_cmd:
+            return sshd_proc_mock(*args, **kwargs)
+        if cmd == launch_cmd:
+            return launch_proc_mock(*args, **kwargs)
+        return None
+
+    mock_subprocess.side_effect = mock_process
+
+    with set_container_id_env_var():
+        with pytest.raises(ValueError):
```
rb-determined-ai, line 219: can you include a `matches=` arg here, so we don't get a false positive?  Also it would make the test easier to understand when reading it, so it's a double-win

rb-determined-ai (commented):
This is a failure test for when sshd fails to be contactable, right?  Is there a failure test for when the training subprocess exits nonzero?  No need to test `pid_server` or `pdsh` here, just make sure that if the command that gets launched fails, that `main()` fails too.

harness/tests/launch/test_autodeepspeed.py
```diff
@@ -0,0 +1,359 @@
...
+    sshd_cmd = [
+        "/usr/sbin/sshd",
+        "-p",
+        str(constants.DTRAIN_SSH_PORT),
+        "-f",
+        "/run/determined/ssh/sshd_config",
+        "-D",
+    ]
+
+    pid_server_cmd = [
```
rb-determined-ai, line 37: Also, these launch commands look like an alarming amount of code duplication, maybe break some of these up into helper functions in `autodeepspeed.py`, and call those from here or something?

rb-determined-ai (approved):

azhou-determined:
merged into ds_trial branch

PR #3763, opened 2022-03-21
Generics

rb-determined-ai:
## Description

Remove the type casting that was necessary when the expconf project landed, but is no longer necessary with go generics.

## Commentary (optional)

For whenver go 1.18 is ready for mac.

## Test Plan

Code is already tested, only now it's got much better compile-time safety.

I added tests for the pseudointerface matching in reflect code.


ioga (approved):
cool

rb-determined-ai:
It seems mockery [just doesn't support generics](https://github.com/vektra/mockery/#note-on-golang-118).

cc @stoksc.  I'm guessing we just sit on this until something changes?

stoksc:
I think we're going to `rm -rf ` mockery in favor anyway, shortly. so sit a bit?

rb-determined-ai:
Oh, I like that strategy.

rb-determined-ai:
Update: mockery seems to support generics now.

Issue #2926, opened 2021-09-08
Static type checker for Pytorch custom metric reducer


felix-schneider:
If I use `wrap_reducer` to wrap a simple python function, I get back a `_SimpleReducer` instance, which supports the `update()` method which is not in `MetricReducer`. As a result, the static type checker complains.

![Screenshot 2021-09-08 152930](https://user-images.githubusercontent.com/208336/132518636-7bd3c56a-4b8d-46d6-aada-67373ecc798d.png)

This could be fixed by defining an overload:

```python
@overload
def wrap_reducer(
        self,
        reducer: Callable[List,Dict[str,float]],
        name: Optional[str] = None,
        for_training: bool = True,
        for_validation: bool = True,
    ) -> _SimpleReducer:
    ...
```

vishnu2kmohan:
@rb-determined-ai have you perhaps already addressed this request, and if not would you? :pray:

rb-determined-ai:
Better [late](https://github.com/determined-ai/determined/pull/3764) than never, right?

Sorry for the slow response on this one, @fe

PR #3764, opened 2022-03-21
fix: update types for pytorch custom reducers

rb-determined-ai:
## Description

Use @typing.overload and typing.TypeVar to make sure that the following
code snippets avoid static analysis issues:

    # simple api case
    reducer = wrap_reducer(my_reducer_fn, ...)
    reducer.update()  # avoid "MetricReducer has no .update"

    # full api case
    class MyReducer():
        def my_update(...):
            ...
    reducer = wrap_reducer(MyReducer(), ...)
    reducer.my_update()  # avoid "MetricReducer has no .my_update"

## Commentary

`mypy` is capable enough to handle this, but IMO this means that the `wrap_reducer` API should really be two separate calls:
- `wrap_simple_reducer(fn: Callable) -> SimpleReducer`
- `register_reducer(metric_reducer: MetricReducer) -> None`

## Test Plan

I verified that running `mypy model_def.py` on the following model definition reproduced the errors without this fix, and I made sure that the errors were gone with this fix.

```python
from typing import Any, Dict, Sequence, Tuple, Union, cast, List, Callable

import torch
from torch import nn

from determined import pytorch

class OnesDataset(torch.utils.data.Dataset):
    def __len__(self) -> int:
        return 64

    def __getitem__(self, index: int) -> torch.Tensor:
        return torch.Tensor([1.0])


def custom_reducer(ins: List[Any]) -> Any:
    return 2.5


class CustomReducer(pytorch.MetricReducer):
    def __init__(self) -> None:
        pass

    def reset(self) -> None:
        pass

    def per_slot_reduce(self) -> None:
        pass

    def cross_slot_reduce(self, psm: List) -> Any:
        return 3.5

    def my_update(self, t: torch.Tensor) -> None:
        print('my update', t)



class OneVarPytorchTrial(pytorch.PyTorchTrial):
    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
        self.context = context

        self.model = context.wrap_model(nn.Linear(1, 1, False))
        # initialize weights to 0
        self.model.weight.data.fill_(0)
        self.opt = context.wrap_optimizer(
            torch.optim.SGD(self.model.parameters(), lr=0.001), backward_passes_per_step=2
        )
        self.fnreducer = self.context.wrap_reducer(custom_reducer, "custom_metric")
        self.clsreducer = self.context.wrap_reducer(CustomReducer(), "CustomMetric")

    def train_batch(
        self, batch: pytorch.TorchData, epoch_idx: int, batch_idx: int
    ) -> Dict[str, torch.Tensor]:
        loss = torch.nn.MSELoss()(self.model(batch), batch)
        self.context.backward(loss)
        self.context.step_optimizer(self.opt)
        self.fnreducer.update(1)
        self.clsreducer.my_update(loss)
        return {"loss": loss}

    def evaluate_batch(self, batch: pytorch.TorchData, batch_idx: int) -> Dict[str, Any]:
        data = labels = batch
        loss = torch.nn.MSELoss()(self.model(data), labels)
        return {"loss": loss}

    def build_training_data_loader(self) -> pytorch.DataLoader:
        return pytorch.DataLoader(
            OnesDataset(), batch_size=self.context.get_per_slot_batch_size()
        )

    def build_validation_data_loader(self) -> pytorch.DataLoader:
        return pytorch.DataLoader(
            OnesDataset(), batch_size=self.context.get_per_slot_batch_size()
        )
```

### DIFF
--- a/harness/determined/pytorch/_reducer.py
+++ b/harness/determined/pytorch/_reducer.py
@@ -1,7 +1,7 @@
 import abc
 import enum
 import itertools
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, TypeVar, Union, overload
 
 import numpy as np
 
@@ -155,8 +155,8 @@ class MetricReducer(metaclass=abc.ABCMeta):
 
         The return value should either be:
            -  A dict mapping string metric names to metric values, if the call to
-              context.wrap_metric() omitted the `name` parameter, or
-           -  A non-dict metric value if the call to context.wrap_metric() had name set to a string
+              context.wrap_reducer() omitted the `name` parameter, or
+           -  A non-dict metric value if the call to context.wrap_reducer() had name set to a string
               (an error will be raised if a dict-type metric is returned but name was set).
 
         This will be called after per_slot_reduce.
@@ -265,6 +265,9 @@ class _WrappedReducer:
         return result
 
 
+T = TypeVar("T", bound=MetricReducer)
+
+
 class _PyTorchReducerContext:
     """
     _PyTorchReducerContext is a component of the PyTorchTrialContext which does not interact with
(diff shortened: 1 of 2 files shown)


harness/determined/pytorch/_reducer.py
```diff
@@ -282,6 +285,28 @@ def reset_reducers(self) -> None:
         for wrapped in self._wrapped_reducers:
             wrapped.reset()
 
+    # When wrap_reducer is passed a Callable, it returns a _SimpleReducer.
+    @overload
```
mpkouznetsov, line 289: Neat; I was not aware of this machinery.

mpkouznetsov (approved):
👏

harness/determined/pytorch/_reducer.py
rb-determined-ai: neither was I, it was @felix-schneider who [explained](https://github.com/determined-ai/determined/issues/2926) it to me.

PR #3766, opened 2022-03-21
chore: add ProtoConverter

rb-determined-ai:
## Description

"just write the struct as json and unmarshal it into protobuf" is not a
computationally free operation or an acceptable strategy overall.

Without some syntactic sugar though, coverting e.g. a
map[string]interface{} in go to a to a protobuf struct.Struct is
actually a pain because the conversion function can return an error.

ProtoConverter just captures and holds those errors until you've called
all of your conversion functions.

## Commentary

I pulled this out of the checkpoints feature branch as its own PR in case we wanted to land it independently, but I don't plan on overhauling existing code to use this, so I can land it with the rest of the checkpoints feature branch if reviewers prefer that.

### DIFF
--- a/master/pkg/protoutils/converter.go
+++ b/master/pkg/protoutils/converter.go
@@ -1,0 +1,68 @@
+package protoutils
+
+import (
+	"math"
+	"time"
+
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/determined-ai/determined/proto/pkg/checkpointv1"
+)
+
+// ProtoConverter just holds errors, allowing you to convert a large number of fields on a protobuf
+// struct without checking errors until afterwards.
+type ProtoConverter struct {
+	err error
+}
+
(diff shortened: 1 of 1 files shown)


master/pkg/protoutils/converter.go
```diff
@@ -0,0 +1,68 @@
...
+	return out
+}
+
+// ToTimestamp converts a time.Time into a timestamppb.Timestamp.
+func (c *ProtoConverter) ToTimestamp(x time.Time) *timestamppb.Timestamp {
+	return timestamppb.New(x)
+}
+
+// ToCheckpointv1State converts a model.State string into a checkpointv1.State
+func (c *ProtoConverter) ToCheckpointv1State(state string) checkpointv1.State {
```
ioga, line 44: Would be cool to have a generic solution for string -> PB enum value.

ioga (approved):

master/pkg/protoutils/converter.go
rb-determined-ai: I don't quite understand what that would look like right now, but I'd be happy to fixup the code when there's enough of these that the pattern would be obvious.

PR #3768, opened 2022-03-21
feat: idle gpu watcher for det >= 0.17.7 [DET-6865]

varlogtim:


examples/tutorials/mnist_pytorch/idle_gpu_watcher.py
```diff
@@ -0,0 +1,174 @@
...
+    parser.add_argument("min_percentage", type=int, default=5,
+                        help="Minimum GPU usage which triggers a failure")
+    parser.add_argument("sample_freq", type=int, default=5,
+                        help="How frequently to sample each GPU usage")
+    parser.add_argument("num_samples", type=int, default=12,
+                        help="Number of samples to look at when evaluating GPU usage")
+    parser.add_argument("delay_samples", type=int, default=12,
+                        help="Number of samples to wait before evaluating GPU usage")
+    parser.add_argument("debug", type=bool, default=True,
+                        help="Display debug information")
```
rb-determined-ai, line 162: recommend:
```python
    # make these optional flags, so the cli is more readable
    parser.add_argument("--min_percentage", type=int, default=5,
                        help="Minimum GPU usage which triggers a failure")
    parser.add_argument("--sample_freq", type=int, default=5,
                        help="How frequently to sample each GPU usage")
    parser.add_argument("--num_samples", type=int, default=12,
                        help="Number of samples to look at when evaluating GPU usage")
    parser.add_argument("--delay_samples", type=int, default=12,
                        help="Number of samples to wait before evaluating GPU usage")
    
    # use flag with action="store_true" to do what `--debug` would normally do
    parser.add_argument("--debug", action="store_true",
                        help="Display debug information")
```

Then change the config to:
```yaml
entrypoint:
  - python3
  - -m
  - idle_gpu_watcher
  - --min-percentage
  - "5"
  - --sample-freq
  - "5"
  - --sample-window
  - "10"
  - --startup-delay-samples
  - "6"
  - --debug
  - python3 
  - -m
  - determined.launch.autohorovod
  - model_def:MNistTrial
```

- I took out the unnecessary double-quotes
- I recommend using --sample-window instead of --num-samples, since "window" is a very common concept when writing filters in DSP, which is sort of what we're doing)
- I recommend using --startup-delay-samples instead of --delay-samples because it becomes totally clear what it means

varlogtim: Done! Good suggestions.

examples/tutorials/mnist_pytorch/idle_gpu_watcher.py
```diff
@@ -0,0 +1,176 @@
...
+                print("IDLE_GPU_WATCHER: Training process still alive, killing.")
+                p.kill()  # SIGKILL because it is likely hung
+
+        return p.wait()
+
+
+class AParser(argparse.ArgumentParser):
+    def error(self, message):
+        description = """
+To use with Determined >= 0.17.7, add the following to your experiment config:
```
rb-determined-ai, line 124: Maybe use [this](https://docs.python.org/3/library/argparse.html#description) instead?

examples/tutorials/mnist_pytorch/idle_gpu_watcher.py
```diff
@@ -0,0 +1,176 @@
...
+  - model_def:Trial"""
+        self.print_help()
+        print(description)
+        sys.stderr.write(f"\nerror: {message}\n")
+        sys.exit(2)
+
+
+if __name__ == "__main__":
+
+    parser = AParser(description=
```
rb-determined-ai, line 151: oh, you are using description.  Why do you also subclass it?

examples/tutorials/mnist_pytorch/idle_gpu_watcher.py
varlogtim: I could use this but I would also have to add FormatterClass things to get the spacing right, seems like overkill right here.

examples/tutorials/mnist_pytorch/idle_gpu_watcher.py
varlogtim: The subclass is to override the .error() behavior. Maybe the variable names are making things seem strange?

varlogtim:
This should still be usable. Closing as this was never meant to be merged.

PR #3769, opened 2022-03-21
chore: add bun.DB singleton in master

rb-determined-ai:
The bun.DB singleton will be the basis of future database code.

It was agreed that making the database connection a singleton is fine
because at runtime, because we only support one database anyway.  This
will allow for very concise and easy interfaces, like trials.List(...)
to list trials from the database.

When we want to unit test something like the trial actor (which makes a
few calls to the database), we can pass it a small interface to use, and
mock just that interface in the unit test.

### DIFF
--- a/master/internal/db/postgres.go
+++ b/master/internal/db/postgres.go
@@ -5,15 +5,48 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	_ "github.com/jackc/pgx/v4/stdlib" // Import Postgres driver.
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/dialect/pgdialect"
+	"github.com/uptrace/bun/extra/bundebug"
 
 	"github.com/determined-ai/determined/master/pkg/model"
 )
+
+var bunMutex sync.Mutex
+var theOneBun *bun.DB
+
+func initTheOneBun(db *sql.DB) {
+	bunMutex.Lock()
+	defer bunMutex.Unlock()
+	if theOneBun != nil {
+		panic("Can't initialize bun twice!  Did you initialize the database twice?")
+	}
+	theOneBun = bun.NewDB(db, pgdialect.New())
+
+	// This will print every query that runs.
+	// theOneBun.AddQueryHook(bundebug.NewQueryHook(bundebug.WithVerbose(true)))
+
+	// This will print only the failed queries.
+	theOneBun.AddQueryHook(bundebug.NewQueryHook())
+}
+
+// Bun returns the singleton database connection through the bun library. bun is the database
+// library we we have decided to use for new code in the future due to its superior composability
+// over bare SQL, and its superior flexibility over e.g. gorm.  New code should not use the old bare
+// SQL tooling.
+func Bun() *bun.DB {
+	if theOneBun == nil {
+		panic("Bun is not yet initialized!  Did you use the database before initializing it?")
+	}
+	return theOneBun
+}
 
 // PgDB represents a Postgres database connection.  The type definition is needed to define methods.
 type PgDB struct {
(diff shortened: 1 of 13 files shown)


master/internal/db/postgres.go
```diff
@@ -23,12 +26,25 @@ type PgDB struct {
...
+// and its superior flexibiliy over e.g. gorm.  New code should not use the old bare SQL tooling.
+var Bun *bun.DB
+
 // ConnectPostgres connects to a Postgres database.
 func ConnectPostgres(url string) (*PgDB, error) {
 	numTries := 0
 	for {
 		sql, err := sqlx.Connect("pgx", url)
 		if err == nil {
+			Bun = bun.NewDB(sql.DB, pgdialect.New())
```
stoksc, line 40: @ioga a singleton that needs initialization is a bit stranger than one that does not. wydt looking at just this? should we at least wrap ConnectPostgres in a `sync.Once`?

ioga: I am fine with this. the way the master code is structured, `ConnectPostgres` is called only once. adding an extra layer of stuff doesn't seem necessary to me: it doesn't solve anything, and the "singleton" comment already makes it pretty clear that users should use it as a singleton.

ioga (approved):

master/internal/db/postgres.go
rb-determined-ai: I put it here because, as @ioga says, this is already called once, very early, by design.  So "it works".

I don't know how you would write a singleton that doesn't need initialization.  The master config has to be used to set this up.  I could make the singleton a function, like `db.Bun()`, that panics with an error if you haven't set up the singleton?

The only place I could see that happening is in unit tests, if e.g. you have some `TrialDB` interface you give to the trial actor, and some new code tries to use `db.Bun` directly.  Wdyt?

stoksc: > I don't know how you would write a singleton that doesn't need initialization.

if it does initialization in `init()` or doesn't need initialization.

i was just pointing out that a singleton that needs initialization at runtime is different than one that doesn't because it has the propensity to be misused whereas `net/http`'s `DefaultClient` does not. i understand it always _should_ be initialized once and early because of how the master code is structured but i think the difference between a singleton and explicitly injecting dependencies here is that a refactor of initialization code can now break at runtime whereas the other way it breaks at compile time. of course, if we had an issue where we refactored core.go so that we made db calls before the db was ready, i think we would catch it, but i'm evaluating singletons as a pattern, not just this specific instance.

concretely, i'm fine with singletons, but I think we at least wrap it in a `sync.Once` so it is at least impossible to accidentally reinitialize it, or even panic if `ConnectPostgres` is called twice, idk. with sync.once then it would be easier to reuse between tests without some strange fixture or reinitialization it every time, yes.

rb-determined-ai: I went with a `db.Bun()` function that panics if it is used wrong.  That should only happen in cases of gross developer error.

I agree with @ioga that we'd most likely never have problems if didn't write any protection around the singleton,  but even more than that, I like code that is bulletproof, and it's nearly zero overhead to do that here.

rb-determined-ai:
re-requested @ioga's review since he approved the approach that I elected not to use

ioga (approved):
I think this is overengineered. I would have preferred the previous approach. 

`ConnectPostgres` function is producing an instance of our legacy `PgDB` which is later stored as a property by a variety of actors. What if connectpostgres is called again, like in the issue we're trying to solve?.. none of the actors would use the new connection. I don't see much value in starting to bother about it now.

However the overall overhead at call site is just two extra characters: doing `db.Bun().` instead of `db.Bun.`, so I am not going to protest this. I also like the variable name `theOneBun`.

stoksc:
> I think this is overengineered. I would have preferred the previous approach.

my point is just.. it seems we're about to use singletons everywhere, it which case having a pattern for singletons that it harder to mess up and easier to use in tests is nice. which is why i proposed `sync.Once` (and i like the fn with checks). i feel we ended up with a compromise that is inconsistent in its choices, so i would be fine with the old way too.

rb-determined-ai:
> i feel we ended up with a compromise that is inconsistent in its choices

I agree with this, but I think it's because we have:
1. a clear vision for the future, but
2. no desire to implement the entire vision right now.

I think the behavior of the bun singleton is accurate to what we would want it to be.  When it's mixed with the ConnectPostgres function, it seems weird, but in the vision of the future, there's no PgDB at all, and it won't be weird anymore.

stoksc:
I like this pattern, it is basically what i was proposing https://sourcegraph.com/github.com/Dreamacro/clash@30d4668008b493b65e7d6982b4c43bb34871e13c/-/blob/component/profile/cachefile/cache.go?L160-165

PR #3771, opened 2022-03-21
fix: nil pointer exception when restoring experiments

eecsliu:
## Description
When restoring experiments, we encounter a nil pointer exception due to the lack of `Owner.Username`. This makes sure we populate the field during restoration.

## Checklist

### DIFF
--- a/master/internal/restore.go
+++ b/master/internal/restore.go
@@ -85,7 +85,7 @@ func (m *Master) restoreExperiment(expModel *model.Experiment) error {
 	taskSpec.TaskContainerDefaults = taskContainerDefaults
 	owner, err := m.db.UserByUsername(expModel.Username)
 	if err != nil {
-		return errors.Wrap(err, "invalid user")
+		return errors.Wrapf(err, "retrieving full user on restart")
 	}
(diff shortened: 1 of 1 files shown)


master/internal/restore.go
```diff
@@ -83,6 +83,11 @@ func (m *Master) restoreExperiment(expModel *model.Experiment) error {
 	taskContainerDefaults := m.getTaskContainerDefaults(poolName)
 	taskSpec := *m.taskSpec
 	taskSpec.TaskContainerDefaults = taskContainerDefaults
+	owner, err := m.db.UserByUsername(expModel.Username)
+	if err != nil {
+		return errors.Wrap(err, "invalid user")
```
stoksc, line 88: error may not be "invalid user"
```suggestion
		return errors.Wrapf(err, "retrieving full user on restart")
```

stoksc (approved):
lgtm. only nit is maybe it makes sense to grab the user back out in the bulk read in `NonTerminalExperiments` instead of making maybe 100-1000s single calls, but i dont think this will hurt too bad.

PR #3775, opened 2022-03-22
chore: update pytorch data loader args

rb-determined-ai:
## Description

New args need to be accepted.

## Commentary (optional)

Still need to think through the implications of allowing new args with old versions of torch.

### DIFF
--- a/harness/determined/pytorch/_data.py
+++ b/harness/determined/pytorch/_data.py
@@ -16,6 +16,7 @@ from typing import (
 
 import numpy as np
 import torch
+from packaging import version
 from torch.utils.data import (
     BatchSampler,
     Dataset,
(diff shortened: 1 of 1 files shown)


harness/determined/pytorch/_data.py
```diff
@@ -95,6 +104,11 @@ def __init__(
         drop_last: bool = False,
         timeout: float = 0,
         worker_init_fn: _worker_init_fn_t = None,
+        multiprocessing_context=None,
```
mpkouznetsov, line 107: are you going to document this argument as well?

mpkouznetsov: I now see that torch does not document it, either. Maybe we could leave it undocumented. Or maybe you can steal from https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataloader.py#L335 . Something like
`A valid context object or a string specifying the start method. (default: None)`.

mpkouznetsov (approved):
Looks good (check the linter).

harness/determined/pytorch/_data.py
rb-determined-ai: We are not here to fix or alter the DataLoader from pytorch, except for exactly the changes we need, so I left this undocumented as they did.

rb-determined-ai:
@mpkouznetsov please re-review; I had to make some fairly significant changes.

mpkouznetsov (approved):
LGTM

PR #3776, opened 2022-03-22
chore: remove unused websockets

hkang1:
## Description

Remove unused websocket dependencies.

## Checklist

### DIFF
--- a/webui/react/package-lock.json
+++ b/webui/react/package-lock.json
@@ -6143,15 +6143,6 @@
         }
       }
     },
-    "@types/websocket": {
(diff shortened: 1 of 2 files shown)


johnkim-det (approved):
lgtm. only question i have is whether there might be websocket stuff on the backend that also could be cleaned up?

hkang1:
> lgtm. only question i have is whether there might be websocket stuff on the backend that also could be cleaned up?

Thanks for the review!

I believe the backend still uses websockets to communicate between the master and agents, and other nodes, beyond API.

PR #3777, opened 2022-03-23
feat: add slurm rendezvous

stoksc:
## Description
Add support to do rendezvous through the master, using some information from slurm on slurm installations.

## Test Plan
- [x] run on a slurm cluster, get ips

## Checklist

### DIFF
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -12,7 +12,7 @@ import psutil
 import determined as det
 from determined import constants
 from determined.common import util
-from determined.common.api import authentication, bindings, certs, request
+from determined.common.api import bindings, certs, request
 from determined.common.experimental import Determined, Session
 
 
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -13,7 +13,7 @@ import determined as det
 from determined import constants
 from determined.common import util
 from determined.common.api import bindings, certs, request
-from determined.common.experimental import Determined, Session
+from determined.common.experimental import Session
 
 
 def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -206,4 +206,4 @@ if __name__ == "__main__":
         det.ResourcesInfo._by_inspection()._to_file()
 
     if args.rendezvous:
-        do_rendezvous(Determined()._session, info.allocation_id)
+        do_rendezvous(sess, info.allocation_id)
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -11,7 +11,8 @@ import psutil
 
 import determined as det
 from determined import constants
-from determined.common.api import bindings, certs, request
+from determined.common import util
+from determined.common.api import authentication, bindings, certs, request
 from determined.common.experimental import Determined, Session
 
 
@@ -196,6 +197,7 @@ if __name__ == "__main__":
         info._to_file()
 
     cert = certs.default_load(info.master_url)
+    sess = Session(info.master_url, util.get_container_user_name(), None, cert)
 
     if args.trial:
         trial_prep(info, cert)
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -2,7 +2,12 @@ import argparse
 import base64
 import io
 import os
+import socket
 import tarfile
+import uuid
+from typing import List, Optional, cast
+
+import psutil
 
 import determined as det
 from determined import constants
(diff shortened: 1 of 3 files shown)


harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+    if not rendezvous_iface:
+        raise ValueError("unable to resolve rendezvous iface")
+
+    return rendezvous_iface.split(",")
+
+
+def get_eth_interface_name() -> Optional[str]:
+    net_if_addrs = list(psutil.net_if_addrs())
+    for interface in net_if_addrs:
+        if interface.startswith("eth"):
```
stoksc, line 136: afaiu, this will work when it needs to and we're sure of that. i started poking around to find the first nic that is ipv4, up and encap 'ether' instead but it wasn't entirely clear to me what python library to use that, or that it was even a good idea.

harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+
+# The canonical definitions of these consts live in Go code.
+RESOURCES_TYPE_K8S_POD = "k8s-pod"
+RESOURCES_TYPE_DOCKER_CONTAINER = "docker-container"
+RESOURCES_TYPE_SLURM_JOB = "slurm-job"
+
+
+def do_rendezvous(info: det.ClusterInfo, cert: certs.Cert) -> None:
+    # Even though resources_id and resources type is not part of the ClusterInfo API, we still
+    # depend on them in all current Determined backends.
```
rb-determined-ai, line 166: No need to make excuses here, I think.  ClusterInfo is a public API and this is an internal file... it wouldn't make sense for us to pass internal info through the public API.

harness/determined/exec/prep_container.py
rb-determined-ai: sounds like you need to write a comment that explains why hardcoding "eth" is a good idea.

harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+        raise ValueError("unable to resolve rendezvous iface")
+
+    return rendezvous_iface.split(",")
+
+
+def get_eth_interface_name() -> Optional[str]:
+    net_if_addrs = list(psutil.net_if_addrs())
+    for interface in net_if_addrs:
+        if interface.startswith("eth"):
+            return cast(str, interface)
```
rb-determined-ai, line 137: can you do:
```python
assert isinstance(interface, str)
return interface
```
instead?

cast() offers no type safety at runtime or lint time... it's really pretty useless.

harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+    if interface not in net_if_addrs:
+        available = list(net_if_addrs.keys())
+        raise ValueError(
+            f"{interface} is not a valid network interface. "
+            f"Valid network interfaces are: {available}"
+        )
+
+    for info in net_if_addrs[interface]:
+        if info.family == socket.AF_INET:
+            return cast(str, info.address)
```
rb-determined-ai, line 153: ditto on not using cast

harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+
+    raise ValueError(f"interface {interface} doesn't have an IPv4 address")
```
rb-determined-ai, line 155: question: why does it have to be ipv4?

harness/determined/exec/prep_container.py
```diff
@@ -54,22 +59,125 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
...
+    for rendezvous_iface in rendezvous_ifaces():
+        try:
+            rendezvous_ip = get_ip_from_interface(rendezvous_iface)
+            break
+        except ValueError as e:
+            resolution_error = e
+    if not rendezvous_ip:
+        raise resolution_error or ValueError("unable to resolve rendezvous ip")
+
+    r = request.post(
```
rb-determined-ai, line 96: how hard is it to use the generated bindings instead?  This is a new REST API usage, so we should be using the new bindings.  While you're modifying the script, you should also consider migrating the old direct-request code.

rb-determined-ai (approved):
looks pretty good to me.

harness/determined/exec/prep_container.py
stoksc: yeah, this was just me saying 'mypy sshhh I know it is a str'

PR #3779, opened 2022-03-23
chore: fix up job queue issues after merge

eecsliu:
## Description
A couple of things got left behind during the merge of `jq-write` branch. This PR fixes/cleans up two errors: the agent RM was returning "unexpected message" errors and deleting a job would result in excessive "no job id found" messages 

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/agent_resource_manager.go
+++ b/master/internal/resourcemanagers/agent_resource_manager.go
@@ -167,8 +167,8 @@ func (a *agentResourceManager) Receive(ctx *actor.Context) error {
 				ctx.Log().WithError(aMsg).Error("")
 				ctx.Respond(aMsg)
 				return nil
-			case jobv1.QueueStats:
-				qStats.Stats = &aMsg
+			case *jobv1.QueueStats:
+				qStats.Stats = aMsg
 				resp.Results = append(resp.Results, &qStats)
 			default:
 				return fmt.Errorf("unexpected response type: %T", aMsg)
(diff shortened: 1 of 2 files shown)


master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -167,8 +167,8 @@ func (a *agentResourceManager) Receive(ctx *actor.Context) error {
 				ctx.Log().WithError(aMsg).Error("")
 				ctx.Respond(aMsg)
```
stoksc, line 168: if there are multiple errors we will ctx.Respond multiple times, but a ctx.Respond closes the channel it sends on and sends on close channels panic, so this will crash the rm if a resource pool crashes.

stoksc, line 168: if there are multiple errors we will ctx.Respond multiple times, but a ctx.Respond closes the channel it sends on and sends on close channels panic, so this will crash the rm if a resource pool crashes.

eecsliu: doesn't it return nil right after?

stoksc: 1. i haven't had coffee yet, woke up 5m ago
2. you're up!?

duh, i can't read. but i do think it makes sense to accumulate the error.

stoksc (approved):

master/internal/resourcemanagers/agent_resource_manager.go
hamidzr: yeah why are you guys up?!!

hamidzr:
🚢

PR #3791, opened 2022-03-23
fix: Restore /model_def / getExperimentModelDefinition

mapmeld:
## Description

The "Download Experiment Code" button uses this legacy API route, which we now return. The Web UI would not be able to convert the encoded ZIP / JSON from the new `/api/v1/experiments/:id/model_def` endpoint without some additional work.

## Test Plan

On latest-master, go to an experiment detail page and click "Download Experiment Code". If the button is not visible, click the three dots in the top right to select it from the dropdown.

## Checklist


johnkim-det (approved):
not sure if i'm the best person to review since it's go code for a feature i'm not too familiar with. i just happened to be the one who noticed the issue. but i don't see any problems, and i guess if it's restoring code that was already there previously, it should be fine?

PR #3796, opened 2022-03-24
chore: API requests that return username should also return userId

johnkim-det:
## Description

Part of a longer-term project to only refer to/look up users by userId. 

This PR updates the following to include userId:
- Experiment, Model, ModelVersion, Command, Shell, Notebook, Tensorboard endpoints
- GET /api/v1/job-queues
- GET resources/allocation/raw

## Test Plan

verify that all of the endpoints that have a `username` in the response also include the corresponding `userId`:
- [ ] Experiments, Models, ModelVersions GETs/POSTs/PATCHes
- [ ] Commands, Shells, Notebooks, Tensorboard GETs/POSTs
- [ ] JobQueues GET
- [ ] ResourceAllocationRaw GET

## Commentary (optional)

Username will be removed from these responses later, but both username and userId are being included for now so that this project can be completed progressively over multiple PRs.

## Checklist


master/static/srv/get_model_versions.sql
```diff
@@ -10,13 +10,14 @@ WITH mv AS (
...
     username,
+    user_id,
     last_updated_time
   FROM model_versions
   LEFT JOIN users ON users.id = model_versions.user_id
   WHERE model_id = $1
 ),
 m AS (
-  SELECT m.id, m.name, m.description, m.notes, m.metadata, m.creation_time, m.last_updated_time, array_to_json(m.labels) AS labels, u.username, m.archived, COUNT(mv.version) as num_versions
+  SELECT m.id, m.name, m.description, m.notes, m.metadata, m.creation_time, m.last_updated_time, array_to_json(m.labels) AS labels, u.username, m.user_id, m.archived, COUNT(mv.version) as num_versions
```
mapmeld, line 20: this is a problem in the existing master branch too, but some fields (username, user_id) are SELECT-ed properly here but also need to be added to the SELECT at the end of this query.  If it is missing you get `"username": "", "userId": 0,` back

hamidzr:
> verify that all of the endpoints that have a username in the response also include the corresponding userId:

a good/easy place to check this is the swagger spec that's generated under proto/build (grep for username?)

mapmeld (approved):

PR #3800, opened 2022-03-24
fix: bugs in PIDServer

rb-determined-ai:
* If a worker dies, we guarantee a nonzero exit.

  We were sending SIGTERM to sshd containers, which were "gracefully"
  exiting zero, causing non-chief containers with failed workers to act
  like they didn't fail.

  Now, we ensure that if a worker or child process dies, we always
  always exit nonzero.

* Guarantee the last worker to exit is checked for a graceful exit.

  There was a codepath where the exit status of the final worker to exit
  on a node was being ignored.


liamcli (approved):
need to change the ipc test case return code but lgtm otherwise

PR #3806, opened 2022-03-25
chore: use userId as url param for user endpoints

johnkim-det:
## Description

Replacing `username` with `userId` as the url parameter to be used when fetching or updating a user.

Also includes minor change to form validation on the Change Password modal.

## Test Plan

- [ ] Change Display Name modal is functional
- [ ] Change Password modal is functional
- [ ] User avatar and display name are fetched correctly

## Checklist


johnkim-det:
@mapmeld @hamidzr not sure how to address the lint-go error for `make -C proto check`, any ideas?

mapmeld:
@johnkim-det That file is sort of a binary compiled record of our proto API, I think we are supposed to keep it up-to-date, generally it comes up in situations like this where you've replaced a field... I've heard it's better practice to remove `string username = 1;` and make `int32 user_id = 2;` even if there is no  field 1 anymore.

You can re-generate that file with `make -C proto gen-buf-image`

PR #3807, opened 2022-03-25
feat: support for torch native dtrain

azhou-determined:
## Description

launch layer + trial code changes needed for supporting torch native dtrain.

tests to follow.

## Checklist

### DIFF
--- a/e2e_tests/tests/fixtures/pytorch_identity/model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_identity/model_def.py
@@ -51,7 +51,12 @@ class IdentityPyTorchTrial(pytorch.PyTorchTrial):
         data, label = batch
 
         loss = self.loss_fn(self.model(data), label)
-        weight = self.model.weight.data.item()
+
+        if self.context._distributed_backend.use_torch():
+            weight = self.model.module.weight.data.item()
+        else:
+            weight = self.model.weight.data.item()
+
         return {"val_loss": loss, "weight": weight}
 
     def build_training_data_loader(self) -> pytorch.DataLoader:
--- a/e2e_tests/tests/fixtures/pytorch_identity/model_def.py
+++ b/e2e_tests/tests/fixtures/pytorch_identity/model_def.py
@@ -52,10 +52,7 @@ class IdentityPyTorchTrial(pytorch.PyTorchTrial):
 
         loss = self.loss_fn(self.model(data), label)
 
-        if self.context._distributed_backend.use_torch():
-            weight = self.model.module.weight.data.item()
-        else:
-            weight = self.model.weight.data.item()
+        weight = self.model.weight.data.item()
 
         return {"val_loss": loss, "weight": weight}
 
--- a/harness/determined/_trial_controller.py
+++ b/harness/determined/_trial_controller.py
@@ -16,9 +16,13 @@ class _DistributedBackend:
 
     HOROVOD = "USE_HOROVOD"
     DEEPSPEED = "USE_DEEPSPEED"
+    TORCH = "USE_TORCH_DISTRIBUTED"
 
     def use_horovod(self) -> bool:
         return bool(os.environ.get(self.HOROVOD, None))
+
+    def use_torch(self) -> bool:
+        return bool(os.environ.get(self.TORCH, None))
 
     def use_deepspeed(self) -> bool:
         return bool(os.environ.get(self.DEEPSPEED, None))
@@ -49,6 +53,7 @@ class TrialController(metaclass=abc.ABCMeta):
 
         distributed_backend = _DistributedBackend()
         self.use_horovod = distributed_backend.use_horovod()
+        self.use_torch = distributed_backend.use_torch()
         self._check_if_trial_supports_configurations(env)
 
         self.scheduling_unit = self.env.experiment_config.scheduling_unit()
--- a/harness/determined/exec/harness.py
+++ b/harness/determined/exec/harness.py
@@ -95,6 +95,8 @@ def main(train_entrypoint: Optional[str]) -> int:
             distributed = _core.DistributedContext.from_horovod(horovod.hvd)
         elif distributed_backend.use_deepspeed():
             distributed = _core.DistributedContext.from_deepspeed()
+        elif distributed_backend.use_torch():
+            distributed = _core.DistributedContext.from_torch_distributed()
         elif len(info.container_addrs) > 1 or len(info.slot_ids) > 1:
             raise ValueError(
                 "In multi-slot tasks, the determined.exec.harness module must not be invoked "
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -1,0 +1,116 @@
+import argparse
+import logging
+import os
+import subprocess
+import sys
+from typing import List, Tuple
+
+import determined as det
+
+C10D_PORT = 29400
+
+
+def main(override_args: List[str], script: List[str]) -> int:
+    override_args = override_args or []
+
+    info = det.get_cluster_info()
+    os.environ["USE_TORCH_DISTRIBUTED"] = "True"
+
+    single_slot = len(info.container_addrs) == 1 and len(info.slot_ids) == 1
+
+    # Detect single-slot trials and skip distributed launch
+    if single_slot:
+        return subprocess.Popen(script).wait()
+
+    chief_ip = info.container_addrs[0]
+    os.environ["DET_CHIEF_IP"] = chief_ip
+
+    torch_distributed_cmd = [
+        "python3",
+        "-m",
+        "torch.distributed.run",
+        "--nnodes",
+        str(len(info.container_addrs)),
+        "--nproc_per_node",
+        str(len(info.slot_ids)),
+        "--node_rank",
+        str(info.container_rank),
+        "--max_restarts",
+        "1",
+        "--master_addr",
+        "localhost" if len(info.container_addrs) == 1 else chief_ip,
+        "--master_port",
+        str(C10D_PORT),
+        "--module",
+    ]
+
+    torch_distributed_cmd.extend(override_args)
+
+    log_redirect_cmd = [
+        "determined.exec.worker_process_wrapper",
+        "RANK",
+        "--",
+    ]
+
+    logging.debug(
+        f"Torch distributed launching with: {torch_distributed_cmd + log_redirect_cmd + script}"
+    )
+
+    return subprocess.Popen(torch_distributed_cmd + log_redirect_cmd + script).wait()
+
+
+def parse_args(args: List[str]) -> Tuple[List[str], List[str]]:
+    if "--" in args:
+        split = args.index("--")
+        override_args = args[:split]
+        args = args[split + 1 :]
+    else:
+        override_args = []
+
+    parser = argparse.ArgumentParser(
+        usage="%(prog)s [[TORCH_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)",
+        description=("Launch a script under pytorch distributed on a Determined cluster"),
+        epilog=(
+            "TORCH_OVERRIDES may be a list of arguments to pass directly to torch.distributed.launch to override "
+            "the values set by Determined automatically.  When provided, the list of override "
+            "arguments must be terminated by a `--` argument."
+        ),
+    )
+    # For legacy Trial classes.
+    parser.add_argument(
+        "--trial",
+        help=(
+            "use a Trial class as the entrypoint to training.  When --trial is used, the SCRIPT "
+            "positional argument must be omitted."
+        ),
+    )
+    # For training scripts.
+    parser.add_argument(
+        "script",
+        metavar="SCRIPT...",
+        nargs=argparse.REMAINDER,
+        help="script to launch for training",
+    )
+    parsed = parser.parse_args(args)
+
+    script = parsed.script or []
+
+    if parsed.trial is not None:
+        if script:
+            # When --trial is set, any other args are an error.
+            parser.print_usage()
+            print("error: extra arguments to --trial:", script, file=sys.stderr)
+            sys.exit(1)
+        script = det.util.legacy_trial_entrypoint_to_script(parsed.trial)
+    elif not script:
+        # There needs to be at least one script argument.
+        parser.print_usage()
+        print("error: empty script is not allowed", file=sys.stderr)
+        sys.exit(1)
+
+    return override_args, script
+
+
+if __name__ == "__main__":
+    override_args, script = parse_args(sys.argv[1:])
+    sys.exit(main(override_args, script))
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -29,6 +29,7 @@ def create_launch_cmd(
         master_addr,
         "--master_port",
         str(C10D_PORT),
+        "--module",
     ]
 
     cmd.extend(override_args)
@@ -120,8 +121,6 @@ def parse_args(args: List[str]) -> Tuple[List[str], List[str]]:
             print("error: extra arguments to --trial:", script, file=sys.stderr)
             sys.exit(1)
         script = det.util.legacy_trial_entrypoint_to_script(parsed.trial)
-        # Add --module as an override arg
-        override_args.append("--module")
     elif not script:
         # There needs to be at least one script argument.
         parser.print_usage()
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -48,6 +48,8 @@ def main(override_args: List[str], script: List[str]) -> int:
     override_args = override_args or []
 
     info = det.get_cluster_info()
+    assert info is not None, "must be run on-cluster"
+
     os.environ["USE_TORCH_DISTRIBUTED"] = "True"
 
     single_slot = len(info.container_addrs) == 1 and len(info.slot_ids) == 1
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -29,7 +29,6 @@ def create_launch_cmd(
         master_addr,
         "--master_port",
         str(C10D_PORT),
-        "--module",
     ]
 
     cmd.extend(override_args)
@@ -121,6 +120,8 @@ def parse_args(args: List[str]) -> Tuple[List[str], List[str]]:
             print("error: extra arguments to --trial:", script, file=sys.stderr)
             sys.exit(1)
         script = det.util.legacy_trial_entrypoint_to_script(parsed.trial)
+        # Add --module as an override arg
+        override_args.append("--module")
     elif not script:
         # There needs to be at least one script argument.
         parser.print_usage()
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -50,14 +50,14 @@ def main(override_args: List[str], script: List[str]) -> int:
     info = det.get_cluster_info()
     assert info is not None, "must be run on-cluster"
 
-    os.environ["USE_TORCH_DISTRIBUTED"] = "True"
-
     single_slot = len(info.container_addrs) == 1 and len(info.slot_ids) == 1
 
     # Detect single-slot trials and skip distributed launch
     if single_slot:
         return subprocess.Popen(script).wait()
 
+    os.environ["USE_TORCH_DISTRIBUTED"] = "True"
+    
     chief_ip = info.container_addrs[0]
     os.environ["DET_CHIEF_IP"] = chief_ip
 
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -36,7 +36,7 @@ def main(override_args: List[str], script: List[str]) -> int:
         "--node_rank",
         str(info.container_rank),
         "--max_restarts",
-        "1",
+        "0",
         "--master_addr",
         "localhost" if len(info.container_addrs) == 1 else chief_ip,
         "--master_port",
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -10,7 +10,9 @@ import determined as det
 C10D_PORT = 29400
 
 
-def create_launch_cmd(num_nodes: int, proc_per_node: int, node_rank: int, master_addr: str, override_args: List[str]) -> List[str]:
+def create_launch_cmd(
+    num_nodes: int, proc_per_node: int, node_rank: int, master_addr: str, override_args: List[str]
+) -> List[str]:
     cmd = [
         "python3",
         "-m",
@@ -57,11 +59,13 @@ def main(override_args: List[str], script: List[str]) -> int:
     chief_ip = info.container_addrs[0]
     os.environ["DET_CHIEF_IP"] = chief_ip
 
-    torch_distributed_cmd = create_launch_cmd(len(info.container_addrs),
-                                              len(info.slot_ids),
-                                              info.container_rank,
-                                              "localhost" if len(info.container_addrs) == 1 else chief_ip,
-                                              override_args)
+    torch_distributed_cmd = create_launch_cmd(
+        len(info.container_addrs),
+        len(info.slot_ids),
+        info.container_rank,
+        "localhost" if len(info.container_addrs) == 1 else chief_ip,
+        override_args,
+    )
 
     log_redirect_cmd = create_log_redirect_cmd()
 
@@ -84,9 +88,9 @@ def parse_args(args: List[str]) -> Tuple[List[str], List[str]]:
         usage="%(prog)s [[TORCH_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)",
         description=("Launch a script under pytorch distributed on a Determined cluster"),
         epilog=(
-            "TORCH_OVERRIDES may be a list of arguments to pass directly to torch.distributed.launch to override "
-            "the values set by Determined automatically.  When provided, the list of override "
-            "arguments must be terminated by a `--` argument."
+            "TORCH_OVERRIDES may be a list of arguments to pass directly to "
+            "torch.distributed.launch to override the values set by Determined automatically.  "
+            "When provided, the list of override arguments must be terminated by a `--` argument."
         ),
     )
     # For legacy Trial classes.
--- a/harness/determined/launch/torch_distributed.py
+++ b/harness/determined/launch/torch_distributed.py
@@ -10,6 +10,38 @@ import determined as det
 C10D_PORT = 29400
 
 
+def create_launch_cmd(num_nodes: int, proc_per_node: int, node_rank: int, master_addr: str, override_args: List[str]) -> List[str]:
+    cmd = [
+        "python3",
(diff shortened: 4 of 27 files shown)


model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -61,16 +62,15 @@ def __init__(self, context: det_torch.PyTorchTrialContext) -> None:
...
-        # horovod's allreduce primitive instead of pytorch distributed's all_reduce function.
-        utils.monkey_patch_reduce_mean(self.model)
-        # We need to convert pytorch SyncBatchNorm layers to horovod's.
-        utils.convert_syncbn_model(self.model)
+        model = mmdet.models.build_detector(self.cfg.model)
 
         # Initialize model
-        self.model.init_weights()
+        model.init_weights()
+        self.model = torch.nn.parallel.DistributedDataParallel(model)
```
rb-determined-ai, line 72: Why don't we wrap the model in DistributedDataParallel on behalf of the user?  That would be more inline with our existing api, where we wrap the user's optimizer for them, only in dtrain scenarios

model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -61,16 +62,15 @@ def __init__(self, context: det_torch.PyTorchTrialContext) -> None:
         # We will control how data is moved to GPU.
         self.context.experimental.disable_auto_to_device()
 
+        dist.init_process_group(backend="gloo")
```
liamcli, line 65: can we `init_process_group` as part of the controller's pre execute hook similar to what we do for deepspeed and enable users to override?
https://github.com/determined-ai/determined/blob/master/harness/determined/pytorch/deepspeed/_deepspeed_trial.py#L70

liamcli: we should probably default to `nccl` backend instead of `gloo`

harness/determined/launch/torch_distributed.py
```diff
@@ -0,0 +1,50 @@
...
+        "-m",
+        "torch.distributed.run",
+        "--nnodes",
+        str(len(info.container_addrs)),
+        "--nproc_per_node",
+        str(len(info.slot_ids)),
+        "--node_rank",
+        str(info.container_rank),
+        "--max_restarts",
+        str(max_restarts),
```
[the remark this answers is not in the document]
azhou-determined: 25?! how? i thought default max_restarts was 5 if not configured in experiment config?

model_hub/model_hub/mmdetection/_trial.py
azhou-determined: done

model_hub/model_hub/mmdetection/_trial.py
azhou-determined: done

harness/determined/launch/torch_distributed.py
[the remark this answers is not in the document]
rb-determined-ai: it looks to me like this is introducing a second layer of restarts, where the launch layer restarts 5 times and the master restarts the launch layer 5 times.

azhou-determined:
> a couple overall questions:
> 
> * Is this change sufficient to allow some users to use images that don't contain horovod?  And if so, is that a first-class experience?
> * Is there new information we need to put into the [pytorch/README.md](https://github.com/determined-ai/determined/blob/master/harness/determined/pytorch/README.md)?

- yes, it should be. _pytorch_trial.py now only initializes horovod if the horovod launch layer is used. i'll build an image without horovod to test this though.
- we'll need to add docs, yes. i'll sync with Gene on this.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -185,6 +187,9 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
         if self.experimental._auto_amp:
             model = self.autocast_forward_pass(model)
 
+        if self._distributed_backend.use_torch():
+            model = torch.nn.parallel.DistributedDataParallel(model)
+
```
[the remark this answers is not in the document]
azhou-determined: i removed DataParallel from this. are we sure it isn't in use?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -168,16 +170,8 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
             check.false(self._use_apex, "Must call wrap_model() before configure_apex_amp.")
 
             model = model.to(self.device)
-            if not self.distributed.size > 1 and self.n_gpus > 1:
```
rb-determined-ai, line 171: what happens if somebody sets slots_per_trial = 1?

model_hub/examples/mmdetection/yolov3.yaml
```diff
@@ -45,6 +45,6 @@ environment:
     - OMP_NUM_THREADS=1 # Following pytorch dtrain, this environment variable is set to 1 to avoid overloading the system.
 
 resources:
-  slots_per_trial: 8 # max number of GPUs a trial is allowed to individually use
+  slots_per_trial: 2 # max number of GPUs a trial is allowed to individually use
```
rb-determined-ai, line 48: don't forget to restore this before landing this pr

harness/determined/launch/torch_distributed.py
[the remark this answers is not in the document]
azhou-determined: made this 1 since we control the master restart through max_restarts

model_hub/examples/mmdetection/yolov3.yaml
azhou-determined: added e2e tests for multi-node torch experiments

harness/determined/_trial_controller.py
```diff
@@ -15,10 +15,14 @@ class _DistributedBackend:
     """
 
     HOROVOD = "USE_HOROVOD"
+    TORCH = "USE_TORCH_DIST"
```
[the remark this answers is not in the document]
azhou-determined: seems like a confusing format for when to use `.` vs `_`. i thought torch_dist was shorter but i wouldn't be opposed to `USE_TORCH_DISTRIBUTED`?

harness/determined/pytorch/_pytorch_context.py
azhou-determined: they should be able to run torch DDP on a single machine single-slot. are you saying we should have some check to prevent them from doing this?

harness/determined/_trial_controller.py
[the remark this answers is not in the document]
rb-determined-ai: yeah I think I agree, USE_TORCH_DISTRIBUTED is better.

e2e_tests/tests/fixtures/pytorch_identity/model_def.py
```diff
@@ -51,7 +51,7 @@ def evaluate_batch(self, batch: pytorch.TorchData) -> Dict[str, Any]:
         data, label = batch
 
         loss = self.loss_fn(self.model(data), label)
-        weight = self.model.weight.data.item()
+        weight = self.model.module.weight.data.item()
```
rb-determined-ai, line 54: does this work, even for non-ddp models?

harness/determined/exec/harness.py
```diff
@@ -95,12 +95,16 @@ def main(train_entrypoint: Optional[str]) -> int:
...
+        elif distributed_backend.use_torch():
+            distributed = _core.DistributedContext.from_torch_distributed()
         elif len(info.container_addrs) > 1 or len(info.slot_ids) > 1:
             raise ValueError(
                 "In multi-slot tasks, the determined.exec.harness module must not be invoked "
                 "directly.  Instead, it must be wrapped in one of the following launch layers: "
                 "determined.launch.horovod, determined.launch.deepspeed"
             )
+        else:
+            distributed = _core.DummyDistributed()
```
rb-determined-ai, line 107: I'd prefer we keep the defaulting in one layer, and `core.init()` already does this.

harness/determined/launch/torch_distributed.py
```diff
@@ -0,0 +1,93 @@
...
+    print(f"launching {torch_distributed_cmd + log_redirect_cmd + script}")
+
+    return subprocess.Popen(
+        torch_distributed_cmd + log_redirect_cmd + script
+    ).wait()
+
+
+def parse_args(args: List[str]) -> List[str]:
+    parser = argparse.ArgumentParser(
+        usage="%(prog)s (--trial TRIAL)|(SCRIPT...)",
```
rb-determined-ai, line 51: do we want to allow overrides to the `torch.distributed.run` command, the way we do for horovodrun?

We need to decide now because it is a breaking change to add them later.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: so as-written, the user story is:
- if you want to run vanilla pytorch, use horovod backend with a single-slot trial
- if you want horovod dtrain, use horovod backend with a multi-slot trial
- if you want DDP dtrain, use torch.distributed backend

It seems weird to me that in the single-slot case, we would special-case horovod but not DDP.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -168,16 +174,11 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
...
-                    "frequency greater than 1 for single machine multi-GPU "
-                    "training.",
-                )
-                model = nn.DataParallel(model)
-                logging.debug("Initialized model for native parallel training.")
+
+            if self._distributed_backend.use_torch():
+                ddp_model = torch.nn.parallel.DistributedDataParallel(model)
+                self._wrapped_models[ddp_model] = model
+                model = ddp_model
```
rb-determined-ai, line 181: Code is usually more readable if you can avoid overwriting local variables.  This is a good example, where it would be harder to read the code wrong if you wrote it like this:

```python
if self._distributed_backend.use_torch():
    wrapped_model = torch.nn.parallel.DistributeDataParallel(model)
else:
    wrapped_model = model

# ... use wrapped_model hereafter
```

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -572,7 +580,7 @@ def backward(
                         gradient=gradient, retain_graph=retain_graph, create_graph=create_graph
                     )
 
-                if self.distributed.size > 1 and self._should_communicate_and_update():
+                if self.distributed.size > 1 and self._should_communicate_and_update() and self._distributed_backend.use_horovod():
                     # When we exit out of this context manager, we need to finish
                     # communicating gradient updates before they are unscaled.
```
rb-determined-ai, line 585: this looks to me like a bug:

- Losses are unscaled when we exit the apex.amp.scale_loss() context manager.
- Losses need to be synchronized before they are unscaled or the scalers can get out-of-sync, causing training to diverge
- In DDP we won't be synchronizing gradients inside the scale_loss context manager, and therefore we have a bug.

Please take a look, and correct me if I'm wrong.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -572,7 +580,7 @@ def backward(
...
                     #
```
rb-determined-ai, line 586: github won't let me comment on lines that are distant from changed lines, but I think you have a fairly serious networking inefficiency by not using [DDP.no_sync](https://pytorch.org/docs/master/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel.no_sync) during gradient aggregation.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: hm... is it that DDP losses are synchronized during the call to .backward?  That would be worth a comment if it is the case.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -659,7 +667,7 @@ def clip_grads(params):
         # before we apply gradient clipping and `step()`. In the case of APEX
         # this is called in backward() instead, so that it's inside the context
         # manager and before unscaling.
-        if self.distributed.size > 1 and not self._use_apex:
+        if self.distributed.size > 1 and self._distributed_backend.use_horovod() and not self._use_apex:
```
rb-determined-ai, line 670: IIUC, this is ok because DDP synchronizes gradients during loss.backward().  If that's true, this is worth another comment.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -689,7 +697,7 @@ def step_fn() -> None:
         else:
             step_fn = optimizer.step  # type: ignore
 
-        if self.distributed.size > 1:
+        if self.distributed.size > 1 and self._distributed_backend.use_horovod():
```
rb-determined-ai, line 700: another line deserving a comment to explain

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -230,7 +238,7 @@ def cleanup_iterator() -> None:
                 ) as load_path:
                     self._load(pathlib.Path(load_path))
 
-            if self.context.distributed.size > 1:
+            if self.context.distributed.size > 1 and self.use_horovod:
                 hvd.broadcast_parameters(self.context._main_model.state_dict(), root_rank=0)
                 for optimizer in self.context.optimizers:
                     hvd.broadcast_optimizer_state(optimizer, root_rank=0)
```
rb-determined-ai, line 244: why does ddp not need this step?

I mean, I understand this is a horovod-specific API, but why doesn't DDP need something equivalent?

model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -62,15 +63,11 @@ def __init__(self, context: det_torch.PyTorchTrialContext) -> None:
         self.context.experimental.disable_auto_to_device()
 
         # Build model and make sure it's compatible with horovod.
-        self.model = mmdet.models.build_detector(self.cfg.model)
-        # We need to monkey patch the reduce_mean function from mmdet.core.utils so that it uses
-        # horovod's allreduce primitive instead of pytorch distributed's all_reduce function.
-        utils.monkey_patch_reduce_mean(self.model)
-        # We need to convert pytorch SyncBatchNorm layers to horovod's.
-        utils.convert_syncbn_model(self.model)
```
rb-determined-ai, line 70: wheeeeew! :tada:

rb-determined-ai (commented):
Ok, I think all of this looks correct, but you need to add a lot of detail to the developer's guide to pytorch trial.

Basically, all of the questions I asked here need to be answerable from that document I think.

harness/determined/pytorch/_pytorch_context.py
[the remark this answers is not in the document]
rb-determined-ai: yes, or if anyone was using it, I'm sure things were badly broken for them.

harness/determined/pytorch/_pytorch_trial.py
azhou-determined: torch `DistributedDataParallel` does this in the constructor. it broadcasts the state_dict() from the local module at rank 0 to all the other ranks in the group.

harness/determined/pytorch/_pytorch_context.py
azhou-determined: added single-slot detection for DDP. do you think an internal argument like `--autohorovod` is necessary in this case?

e2e_tests/tests/fixtures/pytorch_identity/model_def.py
azhou-determined: changed this test model to check depending on DDP or not.

harness/determined/launch/torch_distributed.py
azhou-determined: added. it should be supported since quite a few parameters like rendezvous backend/port should be able to be overriden.

harness/determined/pytorch/_pytorch_context.py
azhou-determined: introduced no_sync.

harness/determined/pytorch/_pytorch_context.py
azhou-determined: yep. DDP receives a signal at the backward() pass to synchronize gradients. i will update the README on all of this.

harness/determined/exec/harness.py
```diff
@@ -95,6 +95,8 @@ def main(train_entrypoint: Optional[str]) -> int:
             distributed = _core.DistributedContext.from_horovod(horovod.hvd)
         elif distributed_backend.use_deepspeed():
             distributed = _core.DistributedContext.from_deepspeed()
+        elif distributed_backend.use_torch():
```
mpkouznetsov, line 98: Is it the intended behavior that if both horovod and torch are selected we use horovod?

mpkouznetsov (approved):
This looks good to me overall (I raised a couple of questions).
Unfortunately, my review is much shallow than that provided by Ryan. For example I learned about `no_sync` today :) .

liamcli:
i think we should include documentation on how to use native pytorch dtrain with this pr.

harness/determined/exec/harness.py
azhou-determined: we don't support multiple dtrain launch layers. in theory this would never happen, harness is launched through a launch layer, and only the launch layer sets environment variables that determines this flag. i guess it's possible that somehow a user sets some of these environment variables themselves for some reason, but this is uncharted/unsupported behavior territory.

model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -179,7 +176,7 @@ def train_batch(self, batch: Any, epoch_idx: int, batch_idx: int) -> Dict[str, t
         batch = {key: batch[key].data[0] for key in batch}
 
         losses = self.model(**batch)
-        loss, log_vars = self.model._parse_losses(losses)
+        loss, log_vars = self.model.module._parse_losses(losses)
```
liamcli, line 179: we're suppose to rely on passthrough here right?  since otherwise this wouldn't work for single slot training.

model_hub/model_hub/mmdetection/_callbacks.py
```diff
@@ -140,7 +140,7 @@ def __init__(self, context: det_torch.PyTorchTrialContext, model: torch.nn.Modul
         self.model = model
 
     def on_validation_start(self) -> None:
-        if self.context.distributed.get_size() > 1:
+        if self.context.distributed.get_size() > 1 and self.context._distributed_backend.use_horovod():
```
liamcli, line 143: i think we should have logic to switch between what broadcast operation to use; either torch.distributed.broadcast or hvd's broadcast.

model_hub/model_hub/mmdetection/_trial.py
```diff
@@ -63,14 +64,10 @@ def __init__(self, context: det_torch.PyTorchTrialContext) -> None:
 
         # Build model and make sure it's compatible with horovod.
         self.model = mmdet.models.build_detector(self.cfg.model)
-        # We need to monkey patch the reduce_mean function from mmdet.core.utils so that it uses
```
liamcli, line 66: we should probably convert everything to work with native pytorch dtrain only since we are going to remove the monkey patches here.

e2e_tests/tests/fixtures/pytorch_identity/model_def.py
```diff
@@ -51,7 +51,12 @@ def evaluate_batch(self, batch: pytorch.TorchData) -> Dict[str, Any]:
         data, label = batch
 
         loss = self.loss_fn(self.model(data), label)
-        weight = self.model.weight.data.item()
+
+        if self.context._distributed_backend.use_torch():
+            weight = self.model.module.weight.data.item()
```
liamcli, line 56: should we remove this branching here so we test the passthrough?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -498,12 +525,30 @@ def configure_apex_amp(
...
             else 0,
         )
+
         if not isinstance(models, list):
             self.models = [models]
+
+        if self.distributed.size > 1 and self._distributed_backend.use_torch():
+            # If Torch DDP is in use, re-wrap the models
+            self.models = [
+                torch.nn.parallel.DistributedDataParallel(model) for model in self.models
```
liamcli, line 535: we need to use passthrough class here no?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -168,25 +174,39 @@ def wrap_model(self, model: torch.nn.Module) -> torch.nn.Module:
...
+                    when using PyTorch DDP
+                    """
+
+                    def __getattr__(self, name: str) -> Any:
+                        try:
+                            return super().__getattr__(name)
+                        except AttributeError:
+                            return getattr(self.module, name)
+
+                wrapped_model = PyTorchDistributedDataParallel(model)
```
liamcli, line 194: can you add a comment here about why we don't need to specify device_ids and output_device when initializing DDP module?  
cf. https://github.com/NVIDIA/apex/blob/master/examples/imagenet/README.md
cf. https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/distributed.py#L428

liamcli:
a few outstanding things i'd like to include in this PR:

- have wrap_scalar return a subclassed scalar when using native pytorch amp that overrides `update` to only call scalar.update when it's after an optimizer step to properly account for gradient aggregation steps.  this is so the user does not have to write 
`if batch_idx%aggregation_frequency==0` in their own code since aggregation_frequency should otherwise not show up in trial code.
- modify load function to work in the case of loading a ddp checkpoint into a non-ddp model.  this can happen when a user sets source_uuid or source_trial_id but changes slots to 1 or loads from a distributed checkpoint but trial has 1 slot.  we should be able to use a consume_prefix_in_state_dict helper function that pytorch has
- make sure apex works with distributed training and aggregation_frequency > 1 and remove the check in `configure_apex_amp`
- look into whether we need to modify the load when using apex amp as is done here to prevent loss spike on resume: https://github.com/NVIDIA/apex/issues/480#issuecomment-587154020
- look into why we do not allow multiple backward passes with apex: https://github.com/determined-ai/determined/blob/master/harness/determined/pytorch/_pytorch_context.py#L560.  it looks like it should be pretty easy to support: https://nvidia.github.io/apex/advanced.html#multiple-models-optimizers-losses

azhou-determined:
> a few outstanding things i'd like to include in this PR:
> 
> * have wrap_scalar return a subclassed scalar when using native pytorch amp that overrides `update` to only call scalar.update when it's after an optimizer step to properly account for gradient aggregation steps.  this is so the user does not have to write
>   `if batch_idx%aggregation_frequency==0` in their own code since aggregation_frequency should otherwise not show up in trial code.

i thought scaler updates with gradient aggregation was handled by `_pytorch_trial:_should_update_scaler` which checks for aggregation frequency in the train step? or are you talking about specific cases where the user code calls `context.scaler.update()` directly?


> * look into why we do not allow multiple backward passes with apex: https://github.com/determined-ai/determined/blob/master/harness/determined/pytorch/_pytorch_context.py#L560.  it looks like it should be pretty easy to support: https://nvidia.github.io/apex/advanced.html#multiple-models-optimizers-losses

according to [the developer README](https://github.com/determined-ai/determined/tree/master/harness/determined/pytorch), it's because we can't map losses to their optimizers, which apex requires to scale_loss. so we synchronize all (only support 1) optimizers on each backwards pass.

azhou-determined:
> a few outstanding things i'd like to include in this PR:
> 
> * modify load function to work in the case of loading a ddp checkpoint into a non-ddp model.  this can happen when a user sets source_uuid or source_trial_id but changes slots to 1 or loads from a distributed checkpoint but trial has 1 slot.  we should be able to use a consume_prefix_in_state_dict helper function that pytorch has

done
> * make sure apex works with distributed training and aggregation_frequency > 1 and remove the check in `configure_apex_amp`
> * look into whether we need to modify the load when using apex amp as is done here to prevent loss spike on resume: [loss spike after checkpoint reload NVIDIA/apex#480 (comment)](https://github.com/NVIDIA/apex/issues/480#issuecomment-587154020)

amp issues will be addressed in a separate PR

harness/determined/launch/torch_distributed.py
```diff
@@ -0,0 +1,135 @@
...
+        str(proc_per_node),
+        "--node_rank",
+        str(node_rank),
+        "--max_restarts",
+        "0",
+        "--master_addr",
+        master_addr,
+        "--master_port",
+        str(C10D_PORT),
+        "--module",
```
liamcli, line 32: i think we should exclude the --module argument from the launch command and instead support it as an override arg.  what do you think?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -498,12 +525,44 @@ def configure_apex_amp(
             if self.distributed.get_rank() == 0 or self.env.experiment_config.debug_enabled()
             else 0,
         )
+
         if not isinstance(models, list):
             self.models = [models]
+
+        if self.distributed.size > 1 and self._distributed_backend.use_torch():
+            # If Torch DDP is in use, re-wrap the models
+            class PyTorchDistributedDataParallel(
```
liamcli, line 534: let's not repeat this class

model_hub/model_hub/mmdetection/_callbacks.py
```diff
@@ -140,7 +140,10 @@ def __init__(self, context: det_torch.PyTorchTrialContext, model: torch.nn.Modul
         self.model = model
 
     def on_validation_start(self) -> None:
-        if self.context.distributed.get_size() > 1:
+        if (
+            self.context.distributed.get_size() > 1
+            and self.context._distributed_backend.use_horovod()
```
liamcli, line 145: let's assert that backend is use_torch and then switch the broadcast statements below to those from the torch.distributed library

model_hub/model_hub/mmdetection/_trial.py
liamcli: we should remove the monkey patch in utils.py and also update the entrypoint in all the experiment configs to go through torch_distributed

harness/determined/launch/torch_distributed.py
azhou-determined: done

azhou-determined: done

model_hub/model_hub/mmdetection/_callbacks.py
azhou-determined: discussed, mmdetection disables torch DDP's `broadcast_buffers` for some reason, which handles this.

azhou-determined: discussed, mmdetection disables torch DDP's `broadcast_buffers` for some reason, which handles this.

harness/determined/pytorch/_pytorch_context.py
azhou-determined: done

azhou-determined: done

model_hub/model_hub/mmdetection/_trial.py
azhou-determined: done

azhou-determined: done

liamcli (approved):
Lgtm

PR #3809, opened 2022-03-25
chore: add trialIds to GET experiment response

johnkim-det:
## Description

Add `trialIds` field to response when fetching an experiment. 

## Test Plan

- [ ] experiments with trials should be fetched with `trialIds` as an array of its trials' ids
- [ ] experiments with trials should be fetched with an accurate `numTrials` field
- [ ] experiments without trials should get fetched with 0 `numTrials` and an empty `trialIds` array

## Checklist


mapmeld:
- I'm a little worried about experiments with lots of trials. Do we have any index in the database related to experiments and trials?
- Doesn't affect this query, but worth noting for future work that trial IDs are not guaranteed to be in order
- Python lint will pass on rebase or merge

johnkim-det:
> * I'm a little worried about experiments with lots of trials. Do we have any index in the database related to experiments and trials?
> * Doesn't affect this query, but worth noting for future work that trial IDs are not guaranteed to be in order
> * Python lint will pass on rebase or merge

@stoksc do you have any thoughts on the performance concern here?

stoksc:
> I'm a little worried about experiments with lots of trials. Do we have any index in the database related to experiments and trials?

I don't expect this to be much worse than the count that was already happening. there is an index that should help

```
CREATE INDEX ix_trials_experiment_id ON public.trials USING btree (experiment_id);
```

> Doesn't affect this query, but worth noting for future work that trial IDs are not guaranteed to be in order

good point, if users are expecting it. could be worthwhile to sort them.

mapmeld (approved):

master/static/srv/get_experiment.sql
```diff
@@ -9,14 +9,17 @@ SELECT
...
     e.end_time AS end_time,
     'STATE_' || e.state AS state,
-    (SELECT COUNT(*) FROM trials t WHERE e.id = t.experiment_id) AS num_trials,
     e.archived AS archived,
     e.progress AS progress,
     e.job_id AS job_id,
     e.parent_id AS forked_from,
     e.owner_id AS user_id,
-    u.username AS username
+    u.username AS username,
```
stoksc, line 17: more of a personal preference thing, but instead of using a group by, i probably would've just written
```
EXPLAIN ANALYZE WITH trial_ids AS (
	SELECT id
	FROM trials
	WHERE experiment_id = 1
)
SELECT
	e.id AS id,
	e.config->>'name' AS name,
	e.config->>'description' AS description,
	e.config->'labels' AS labels,
	e.config->'resources'->>'resource_pool' as resource_pool,
	e.config->'searcher'->'name' as searcher_type,
	e.notes AS notes,
	e.start_time AS start_time,
	e.end_time AS end_time,
	'STATE_' || e.state AS state,
	e.archived AS archived,
	e.progress AS progress,
	e.job_id AS job_id,
	e.parent_id AS forked_from,
	e.owner_id AS user_id,
	u.username AS username,
	(SELECT json_agg(id) FROM trial_ids) AS trial_ids,
	(SELECT count(id) FROM trial_ids) AS num_trials
FROM experiments e
JOIN users u ON e.owner_id = u.id
WHERE e.id = 1;
```
for whatever that is worth.

PR #3810, opened 2022-03-25
chore: add missing FK constraints for `trials` and `allocation_sessions`

ioga:
## Description

Some foreign keys were missing:
- Add `task_id` foreign key on `trials`.
- Add `allocation_id` foreign key on `allocation_sessions`.
- Refactor `checkpoint_gc` tasks to be more like normal allocations.

Note: To make sure migration doesn't fail, we delete offending `task_ids`. There shouldn't be any, but if they exist then they're probably not useful anyway.

## Test Plan

N/A

## Checklist


master/static/migrations/20220325095020_add-task-allocation-fks.tx.up.sql
```diff
@@ -0,0 +1,14 @@
+-- NULLify entries that would cause an integrity error.
+UPDATE trials SET task_id = NULL WHERE task_id NOT IN (SELECT task_id FROM tasks);
```
stoksc, line 2: I don't think this is possible, but I suppose better safe than sorry

stoksc (approved):
lgmt

stoksc (approved):
+1000 the checkpoint gc updates. thanks.

PR #3812, opened 2022-03-26
ci: delete useless and slow noop test

rb-determined-ai:
Once upon a time, the master would kill the trial runner if it requested
a TRAIN_FOR_STEP and didn't hear back from the trial in less than 60
seconds, hence this regression test.

With push architecture, where there is no longer a persistent connection
of any kind, this test really just adds a couple extra minutes of test
time while offering absolutely no new codepaths of test coverage.


e2e_tests/tests/fixtures/no_op/single-long-train-step.yaml
```diff
@@ -1,26 +0,0 @@
...
-  metrics_progression: decreasing
-  metrics_base: 0.9
-  metrics_sigma: 0
-  training_batch_seconds: 1
-searcher:
-  metric: validation_error
-  smaller_is_better: true
-  name: single
-  max_length:
-    batches: 100
```
rb-determined-ai, line 18: somewhere along the way, this test stopped being a "single long train step" and became "a long string of short steps" anyway.  i.e. it doesn't even do as advertised anymore.

rb-determined-ai:
did some digging, here are the original commits where this test was added:
```
commit 0026c688581b99b8fcba4e25e3912d87a6887b56
Date:   Tue Dec 11 14:45:39 2018 -0800

    Add an integration test for a >60 second training step.

commit 3c6d4d2e068a0dec6e0c5aab14a7cca736a3a908
Date:   Tue Dec 11 14:43:15 2018 -0800

    Modify uwsgi config to correctly handle long-lived TR websockets.

    e2f8e71bca9d3acc4fe93c7826199ad8ff808c1b removed pings from the trial
    runner codebase and inadvertantly caused the failures of any workloads
    that ran for longer than 60 seconds. This commit modifies the correct
    uWSGI connections to amend this.
```

stoksc (approved):

PR #3816, opened 2022-03-28
fix: keep page from crashing when trial/hyperparameters are not loaded yet [DET-6951]

trentwatt:
## Description

fix: keep page from crashing when trial/hyperparameters are not loaded yet [DET-6951]

## Test Plan

- `cd examples/computer_vision/cifar10_pytorch`

- `det e create distributed.yaml .`

navigate to /experiments/<new_experimentid> and click on hyperparameters before experiment is able to start.

page should not crash

## Commentary (optional)

It looks like it first crops up in [3743](https://github.com/determined-ai/determined/pull/3743), which prevented the loading limbo state.

seems like if there is only one trial for an experiment we shouldn't have to wait for it to start in order to be able to display the hyperparameters, but [this](https://github.com/determined-ai/determined/blob/e35972d83ca67b5fe3b4672248e0db6d940e97b5/webui/react/src/pages/ExperimentDetails/ExperimentSingleTrialTabs.tsx#L64) is unable to fetch. 

more investigation could probably be made into being able to show hyperparams for queued experiments..

## Checklist

[DET-6951]: https://determinedai.atlassian.net/browse/DET-6951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
+++ b/webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
@@ -36,6 +36,7 @@ const TrialDetailsHyperparameters: React.FC<Props> = ({ trial }: Props) => {
   ], []);
 
   const dataSource: HyperParameter[] = useMemo(() => {
+    if (trial?.hyperparameters == null) return [];
     return Object.entries(trial.hyperparameters).map(([ hyperparameter, value ]) => {
       return {
         hyperparameter,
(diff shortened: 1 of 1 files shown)


EmilyBonar (approved):
LGTM!

PR #3817, opened 2022-03-28
feat: Prevent PATCH/DELETE on default workspace/project [DET-6952]

mapmeld:
## Description

Handles the special case of immutable workspaces and projects (the default ones created to hold existing experiments). This should prevent PATCH to change their name or DELETE to remove them.  The column already existed but here we're adding it to the proto / API.

I added some lines to the `e2e_test`

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_workspace_org.py
+++ b/e2e_tests/tests/cluster/test_workspace_org.py
@@ -2,7 +2,7 @@ from typing import List
 
 import pytest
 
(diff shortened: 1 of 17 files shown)


EmilyBonar (approved):
This all looks good. I see that the api can return immutable as `undefined`. Ideally it would always return true or false, but as long as the default workspace/project always have it equal true it doesn't really matter.

PR #3828, opened 2022-03-29
fix: use RoutePagination to navigate between trials

johnkim-det:
## Description

Allow users to navigate between trials for the same experiment with buttons/keys.

image

## Test Plan

- [ ] multi-trial experiment "Trial detail" pages show next/previous buttons.
- [ ] clicking "next" shows the trial that was created after the current trial
- [ ] clicking "previous" shows the trial that was created after the current trial
- [ ] tab persists between traversal (e.g. if viewing "Hyperparameters" tab when clicking "next/previous, i should still see the "Hyperparameters" tab on the newly navigated to trial)
- [ ] Start and end of list shows disabled buttons.
- [ ] User can use keyboard arrow keys to traverse experiments. (Left arrow goes to previous trial. Right arrow goes to next trial)

## Checklist

### DIFF
--- a/webui/react/src/components/RoutePagination.module.scss
+++ b/webui/react/src/components/RoutePagination.module.scss
@@ -1,0 +1,5 @@
+.base :global(.ant-pagination-item),
+.base :global(.ant-pagination-jump-prev),
+.base :global(.ant-pagination-jump-next) {
+  display: none;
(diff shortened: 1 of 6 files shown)


gt2345 (approved):
LGTM

PR #3829, opened 2022-03-29
chore: update StorageManager and extend CheckpointContext

rb-determined-ai:
## Description

- Take checkpointing business logic out of StorageManager; functions now
  operate in terms of src and dst, not storage_id (checkpoint uuid).
  This is not clearly necessary yet, but when we support sharded
  checkpointing we will certainly need to separate the storage layer
  from the checkpointing layer (so as to support multiple storage layer
  calls for a single checkpointing operation).

- Add upload(src,dst) and download(src,dst) to the StorageManager
  interface. Also add unit tests for upload/download.

- Redesign StorageManager to use pathlib.Path everywhere, which makes
  the src/dst distinction a bit clearer.

- Introduce CloudStorageManager, which defines restore_path() and
  post_store_path() for all cloud-based storage managers in terms of
  upload. This saves a good chunk of duplicate code.

- Settle on argument ordering; be consistent with (src,dst) everywhere.

- Move checkpointing business logic to core.CheckpointContext and
  use that everywhere.

- (TODO, not this pr): add sharding extensions to core.Checkpointing.

## Test plan

I added unit tests for upload/download.

### DIFF
--- a/harness/determined/common/storage/base.py
+++ b/harness/determined/common/storage/base.py
@@ -1,22 +1,30 @@
 import abc
 import contextlib
 import os
-import shutil
-import uuid
-from typing import Any, Dict, Iterator, Optional, Tuple
+import pathlib
+from typing import Any, Dict, Iterator, Optional, Union
 
 from determined.common.check import check_gt, check_true, check_type
 
 
-class StorageManager:
+class StorageManager(metaclass=abc.ABCMeta):
     """
     Abstract base class for storage managers. Storage managers need to
-    support three operations: creating, loading, and deleting storages.
+    support five operations:
 
-    Configuration for storage managers is represented as a dictionary of key
-    value pairs. The primary key in the dictionary is the `type` defining
-    which storage backend to use. Additional keys may be required to
-    instantiate some implementations of the storage manager.
+       1.  Upload an existing directory to storage
+       2.  Download from storage to a target directory
+       3.  Provide a path to write to, in order to upload
+       4.  Provide a path to read from, after a download
+       5.  Delete a directory in storage
+
+    Advanced methods 3. and 4.  allow shared_fs to do do optimized zero-copy checkpointing.
+    Cloud-based implementations can subclass the CloudStorageManager, which will define 3. and 4. in
+    terms of 1. and 2.
+
+    Configuration for storage managers is represented as a dictionary of key value pairs. The
+    primary key in the dictionary is the `type` defining which storage backend to use. Additional
+    keys may be required to instantiate some implementations of the storage manager.
     """
 
     def __init__(self, base_path: str) -> None:
@@ -29,7 +37,8 @@ class StorageManager:
         """from_config() just calls __init__() unless it is overridden in a subclass."""
         return cls(**config)
 
-    def post_store_path(self, storage_id: str, storage_dir: str) -> None:
+    @abc.abstractmethod
+    def post_store_path(self, src: str, dst: str) -> None:
         """
         post_store_path is a hook that will be called after store_path(). Subclasess of
         StorageManager should override this in order to customize the behavior of store_path().
@@ -37,18 +46,14 @@ class StorageManager:
         pass
 
     @contextlib.contextmanager
-    def store_path(self, storage_id: str = "") -> Iterator[Tuple[str, str]]:
+    def store_path(self, dst: str) -> Iterator[pathlib.Path]:
         """
-        Prepare a local directory that will become a checkpoint.
+        Prepare a local directory to be written to the storage backend.
 
-        This base implementation creates the temporary directory and chooses a
-        random checkpoint ID, but subclasses whose storage backends are in
-        remote places are responsible for uploading the data after the files are
-        created and deleting the temporary checkpoint directory.
+        This base implementation creates the dst directory, but subclasses whose storage
+        backends are in remote places are responsible for uploading the data after the files are
+        created and deleting the temporary dst directory.
         """
-
-        if storage_id == "":
-            storage_id = str(uuid.uuid4())
 
         # Set umask to 0 in order that the storage dir allows future containers of any owner to
         # create new checkpoints. Administrators wishing to control the permissions more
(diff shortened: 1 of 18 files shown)


harness/determined/common/storage/base.py
```diff
@@ -75,31 +78,20 @@ def restore_path(self, storage_id: str) -> Iterator[str]:
...
-        storage_dir = os.path.join(self._base_path, storage_id)
-
-        check_true(
-            os.path.exists(storage_dir), "Storage directory does not exist: {}".format(storage_dir)
-        )
-        check_true(
-            os.path.isdir(storage_dir), "Storage path is not a directory: {}".format(storage_dir)
-        )
+    @abc.abstractmethod
+    def upload(self, src: str, dst: str) -> None:
```
ioga, line 82: I feel like this API makes it hard to distinguish which parameters are concrete local filesystem paths, and which are "remote" destinations which sometimes may be paths and sometimes may look like paths but really aren't, such as s3 key names or prefixes.

I understand this may be a bit of work but we could solve it through typing, in particular have `Union[str, Path]` for concrete local fs paths.

so upload will become
```suggestion
    def upload(self, src: Union[str, Path], dst: str) -> None:
```

For return values, returning `pathlib.Path` for things that are paths also may be helpful in a similar vein. This applies for `restore_path` which may return `-> Iterator[Path]`.

harness/determined/common/storage/base.py
```diff
@@ -1,22 +1,29 @@
...
-    which storage backend to use. Additional keys may be required to
-    instantiate some implementations of the storage manager.
+       1.  Upload an existing directory to storage
+       2.  Download from storage to a target directory
+       3.  Provide a path to write to, in order to upload
+       4.  Provide a path to read from, after a download
+       5.  Delete a directory in storage
+
+    Notice that the 3. and 4. are a little odd, but the mechanics allow shared_fs to do zero-copy
+    checkpointing.  Cloud-based implementations can subclass the CloudStorageManager, which will
```
ioga, line 21:
```suggestion
    Advanced operations 3. and 4. enable shared_fs to do optimized zero-copy
    checkpointing.  Cloud-based implementations can subclass the CloudStorageManager, which will
```

harness/determined/common/storage/base.py
rb-determined-ai: done, but I used `Union[str, os.PathLike]` instead of `Union[str, pathlib.Path]`.

harness/determined/_core/_checkpointing.py
```diff
@@ -32,8 +35,69 @@ def __init__(
...
+        """
+        # XXX really?  I find my own docstring highly un-convincing.  But I can't shake the feeling
+        #     that an API with a .upload() should have an equally easy .download() method.
+
+        ckpt_dir = os.fspath(ckpt_dir)
+
+        self._storage_manager.download(src=storage_id, dst=ckpt_dir)
+
+        # XXX get real metadata
+        return {}
```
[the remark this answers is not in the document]
ioga: we've got a training `Context` object that has preemption, searcher, checkpointing, and they form the "Core API" aka "use this API in training". Python SDK is initialized separately, so in part it has its own life. Maybe having more than one way to do a thing is okay.

if you want to deduplicate this with Python SDK, then I'd say
- `Context` object should have `.client` on it which is a python sdk client - so users don't have to explicitly instantiate it, i.e. if they don't have to explicitly initialize the 2nd explicit API.
- users will call `context.client.get_checkpoint(...).download`
- add `Checkpoint#upload` method, we're missing it today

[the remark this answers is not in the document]
rb-determined-ai: > Maybe having more than one way to do a thing is okay.

I think I'm going to go with this in the short term.

rb-determined-ai:
@ioga this is ready for final review/approval.

harness/determined/common/storage/shared.py
```diff
@@ -60,9 +68,32 @@ def restore_path(self, storage_id: str) -> Iterator[str]:
...
+        shutil.copytree(src, os.path.join(self._base_path, dst))
+
+    def download(self, src: str, dst: Union[str, os.PathLike]) -> None:
+        dst = os.fspath(dst)
+        try:
+            shutil.copytree(os.path.join(self._base_path, src), dst)
+            return None
+        except FileNotFoundError:
+            pass
+        raise errors.CheckpointNotFound(f"Did not find checkpoint {src} in shared_fs storage")
```
ioga, line 99:
```suggestion
        try:
            shutil.copytree(os.path.join(self._base_path, src), dst)
        except FileNotFoundError:
            raise errors.CheckpointNotFound(f"Did not find checkpoint {src} in shared_fs storage")
```

harness/determined/estimator/_estimator_trial.py
```diff
@@ -748,9 +748,9 @@ def _init_paths(self) -> None:
             logging.debug(f"Load path set to {self.estimator_dir}.")
 
             # Load WorkloadSequencer state.
-            wlsq_path = os.path.join(load_path, "workload_sequencer.pkl")
-            if self.wlsq is not None and os.path.exists(wlsq_path):
-                with open(wlsq_path, "rb") as f:
+            wlsq_path = load_path.join("workload_sequencer.pkl")
```
ioga, line 751: I don't think `pathlib.Path` has `.join`? prob should be `joinpath` or `/`?
```suggestion
            wlsq_path = load_path / "workload_sequencer.pkl"
```

ioga (approved):

harness/determined/common/storage/shared.py
rb-determined-ai: I see the clarity of the error messages produced by a library as part of the measure of the quality of a library, and so I dislike giving longer stacktraces than necessary, which is why I wrote it the way I did.

The requests library is guilty of excessively long stacktraces.  They'll give you a 3-step traceback covering a page and a half of your terminal, all to say "yo that request timed out".

So the way I wrote this, we would discard some stacktrace going into shutil.copytree (which is useless) and raise a CheckpointNotFound error with the src in the error message (which is like 99.9% likely to be enough of a stack trace).

harness/determined/estimator/_estimator_trial.py
rb-determined-ai: good catch.  I'm going to see if I can't figure out any why this wasn't caught by the linter, and look for similar errors.  e2e tests indicate there are a couple more.

harness/determined/common/storage/shared.py
rb-determined-ai: as per your out-of-band suggestion, I'll change this to `raise CheckpointNotFound() from None` to discard the earlier stacktrace.

harness/determined/estimator/_estimator_trial.py
rb-determined-ai: It's not caught because apparently TrialController.context has always been an `Any`.

I experimented with fixing it but it looks, surprisingly, very involved.  So I don't have time to do it today

PR #3833, opened 2022-03-30
refactor: consolidate allocation state.

ioga:
## Description

Remove `Allocation.state` field in favor of `Allocation.model.State`.

## Commentary (optional)

I am contemplating inserting `Allocation` into database on `PreStart` (instead of when resources are allocated).

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -37,10 +37,6 @@ type (
 		// The persisted representation.
 		model model.Allocation
 
-		// State for the primary behavior of an allocation.
-		// The state of the allocation, just informational.
-		state model.AllocationState
-
 		// State of all our resources.
 		resources resourcesList
 		// Separates the existence of resources from us having started them.
@@ -205,7 +201,6 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 	// These messages allow users (and sometimes an orchestrator, such as HP search)
 	// to interact with the allocation. The usually trace back to API calls.
 	case AllocationReady:
-		a.model.State = &a.state // TODO(Brad): Consolidate where state is store else bugs.
 		a.model.IsReady = ptrs.BoolPtr(true)
 		if err := a.db.UpdateAllocationState(a.model); err != nil {
 			a.Error(ctx, err)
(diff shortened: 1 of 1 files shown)


master/internal/task/allocation.go
```diff
@@ -205,7 +201,6 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 	// These messages allow users (and sometimes an orchestrator, such as HP search)
 	// to interact with the allocation. The usually trace back to API calls.
 	case AllocationReady:
-		a.model.State = &a.state // TODO(Brad): Consolidate where state is store else bugs.
```
ioga, line 208: I was slightly perplexed by this line: did it do anything?

stoksc: no, it didn't

master/internal/task/allocation.go
```diff
@@ -356,13 +351,13 @@ func (a *Allocation) Cleanup(ctx *actor.Context) {
...
 func (a *Allocation) ResourcesAllocated(ctx *actor.Context, msg sproto.ResourcesAllocated) error {
-	if a.state != model.AllocationStatePending {
+	if a.getModelState() != model.AllocationStatePending {
 		// If we have moved on from the pending state, these must be stale (and we must have
 		// already released them, just the scheduler hasn't gotten word yet).
 		return ErrStaleResourcesReceived{}
 	}
 
-	a.state = model.AllocationStateAssigned
+	a.setModelState(model.AllocationStateAssigned)
```
stoksc, line 360: ideally, at this point it would get automagically persisted or something (the entire model) at the end of a receive block. could `defer persist model` at the top of receive and use a flag on the model to decide to persist. also, eventually bun.

stoksc (approved):

master/internal/task/allocation.go
stoksc: maybe ideally is too strong. "i had a thought that, " instead.

ioga: yes I agree we could do that one day.
also some of the methods here do two updates within one function without a transaction, which also would be a nice improvement once we have bun here.

PR #3836, opened 2022-03-31
feat: allow the CLI to accept incomplete task UUIDs [DET-6706]

dzhu:
## Description

For user convenience, this allows the CLI to accept an incomplete prefix
of a task UUID as long as there is a unique UUID that it corresponds to.

## Test Plan

- [x] manual testing
- [ ] new tests (partially done, want to cover all task types without being totally awful)

### DIFF
--- a/docs/release-notes/cli-uuid-prefixes.txt
+++ b/docs/release-notes/cli-uuid-prefixes.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**New Features**
+
+-  CLI: Allow the CLI to accept any unique prefix of a task UUID to refer to the task, rather than
+   requiring the entire UUID. In some places, Determined only displays the first few characters of a
+   UUID.
--- a/harness/determined/cli/command.py
+++ b/harness/determined/cli/command.py
@@ -1,9 +1,10 @@
 import base64
 import json
+import re
 from argparse import Namespace
 from collections import OrderedDict, namedtuple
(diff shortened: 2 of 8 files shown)


hamidzr (approved):
LGTM. This is helpful. The only thing I can think of is that we want to persist command longer than 24hrs and that might mean a paginated API at which point we'd practically need to do the expansion server side. Would that still be reasonable? It seems ok. We wouldn't wanna take the feature away later.

docs/release-notes/cli-uuid-prefixes.txt
```diff
@@ -0,0 +1,5 @@
+:orphan:
+
+**New Features**
```
gh-determined-ai, line 3: Sorry for the delay.
This could use more context, such as describing the problem this feature solved, from the DET-6706 description field.
Also, is this really a 'new feature' or more of a 'fixed'?

gh-determined-ai (approved):

docs/release-notes/cli-uuid-prefixes.txt
dzhu: Okay, I added more context. I'd say it's a new feature; it's not incorrect for the CLI to require a full UUID, just not the most convenient possible functionality.

dzhu:
@hamidzr Yeah, that should be fine to adapt to later if required.

PR #3837, opened 2022-03-31
chore: add conditionalRender prop to Spinner

johnkim-det:
## Description

Add a prop to the `<Spinner>` component that provides the option to not render content while loading/spinning [DET-6275]
Screen Shot 2022-03-31 at 9 34 28 AM
Screen Shot 2022-03-31 at 9 34 00 AM

## Test Plan

- [ ] try using the prop in Page.tsx and confirm that the page content is not rendered

## Checklist

[DET-6275]: https://determinedai.atlassian.net/browse/DET-6275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/Spinner.tsx
+++ b/webui/react/src/components/Spinner.tsx
@@ -8,6 +8,7 @@ import css from './Spinner.module.scss';
 
 interface Props extends Omit<SpinProps, 'size'>, SpinState {
   center?: boolean;
+  conditionalRender?: boolean;
   inline?: boolean;
   size?: IconSize;
 }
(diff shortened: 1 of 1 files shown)


trentwatt (approved):
Looks good to me. 

`Take it or leave it` It does look kind of bare though with the conditional render. One possibility would be to render a skeleton instead of null as the child, something like this:

image

This is taken from our storybook, at [this page](http://localhost:9009/?path=/docs/skeleton-skeletonsection--with-custom-children#with-custom-children) assuming you do `npm run storybook`

PR #3839, opened 2022-03-31
fix: remove python 3.10 union-type syntax

rb-determined-ai:
## Description

This removes a [python 3.10-specific syntax](https://peps.python.org/pep-0604/) from the codebase.

I can't figure out how the code in question runs at all.  It must be some sort of runtime optimization that isn't considering types at all, because a simple command like:
```
python -c 'a: int | str = 1'
```
throws an error:
```
TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

I was occasionally (but not consistently) seeing mypy errors in make check.

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -9,7 +9,7 @@ import time
 from argparse import FileType, Namespace
 from pathlib import Path
 from pprint import pformat
-from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Set, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Set, Tuple, Union
 
 import tabulate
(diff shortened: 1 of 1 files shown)


mapmeld:
I don't have an explanation for that either.
I was trying to add a suggestion ( `Union[A, B]` instead of `Union[A | B]` ) but it was actually a commit, hope that makes sense as the better backwards-compatible way to do it

rb-determined-ai:
I mean, normally I would raise an eyebrow at somebody committing to my PR, but this is fine.  Silly error on my part.

mapmeld (approved):
Approved with a tiny change. Sorry I could have done the requested change part better.

rb-determined-ai:
no need to be sorry, not a problem at all here

PR #3843, opened 2022-04-01
ci: fail building docs when sphinx-build fails

rb-determined-ai:
~Our little sphinx-wrapper needs `set -o pipefail`, otherwise
sphinx-build can fail without our build step failing.~

Nope, debian dash doesn't have `set -o pipefail`


varlogtim:
Yup, looks good a second time around again.

PR #3851, opened 2022-04-04
feat: add opentel to determined master [DET-6775]

nrajanee:
## Description

feat: add opentel to determined master [DET-6775] 

## Test Plan

- [ ] Run `make devcluster`
- [ ] Run `docker run --rm -v “$HOME/Desktop/opentel/config.yaml" -p 1888:1888 -p 8888:8888 -p 8889:8889 -p 13133:13133  -p 4317:4317 -p 4318:4318 otel/opentelemetry-collector` to start the opentel collector. 
- [ ] Go to localhost:8080 and check that you are seeing logs when you hit diff endpoints. 

[DET-6775]: https://determinedai.atlassian.net/browse/DET-6775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/internal/core.go
```diff
@@ -740,6 +742,8 @@ func (m *Master) Run(ctx context.Context) error {
 	// Initialize the HTTP server and listen for incoming requests.
 	m.echo = echo.New()
 	m.echo.Use(middleware.Recover())
+	tp := otelConfig()
+	m.echo.Use(otelecho.Middleware("determined-master", otelecho.WithTracerProvider(tp)))
```
philipnrmn, line 746: Did we test whether it's necessary to pass the tracer provider here? I think this might be belt-and-braces, so it's not necessarily an issue, just curious.

philipnrmn:
LGTM! Do we need to make this configurable, is that coming in a separate PR?

master/internal/core.go
nrajanee: I will try that!

nrajanee:
> LGTM! Do we need to make this configurable, is that coming in a separate PR?

nrajanee:
> LGTM! Do we need to make this configurable, is that coming in a separate

master/internal/core.go
```diff
@@ -738,6 +740,8 @@ func (m *Master) Run(ctx context.Context) error {
 	// Initialize the HTTP server and listen for incoming requests.
 	m.echo = echo.New()
 	m.echo.Use(middleware.Recover())
+	otelConfig()
+	m.echo.Use(otelecho.Middleware("determined-master"))
```
philipnrmn, line 744: Brilliant. One last thing, let's move this down to the bottom of the middleware (line 788 at present) so it times the whole call, not just the inner call.

philipnrmn (approved):
Splendid. Looking forward to seeing a dashboard!

PR #3852, opened 2022-04-04
feat: track task stats [DET-6872, DET-6926, DET-6927]

gt2345:
## Description

Track image pulled time and queued time for each allocation.
Image pulled time is accessed from `get_raw_allocation` and queued time is probably accessed through api once decided.

## Test Plan

Run task with new image, 
verify that image pulled start and end time coincident with CSV report from cluster page,


master/internal/resourcemanagers/agent/agent.go
```diff
@@ -69,6 +70,8 @@ type (
 		opts *aproto.MasterSetAgentOptions
 
 		agentState *AgentState
+
+		db db.DB
```
ioga, line 74: per the previous comment - let's not add `db` here.

master/pkg/model/task.go
```diff
@@ -101,6 +101,15 @@ type Allocation struct {
 // AllocationState represents the current state of the task. Value indicates a partial ordering.
 type AllocationState int
 
+// TaskStats is the model for task stats in the database.
+type TaskStats struct {
+	AllocationID AllocationID `db:"allocation_id"`
+	ResourcePool string       `db:"resource_pool"`
+	EventType    string       `db:"event_type"`
+	StartTime    *time.Time   `db:"start_time"`
+	EndTime      *time.Time   `db:"end_time"`
```
ioga, line 110: bun tags are `bun:` instead of `db:` so I suspect these end up doing nothing, but the field/column names luckily match anyway.

master/pkg/tasks/task.go
```diff
@@ -79,6 +79,7 @@ type TaskSpec struct {
 	Devices                []device.Device
 
 	UserSessionToken string
+	TaskType         model.TaskType
```
ioga, line 82: if we won't check for tasktype on agent, this probably won't be necessary as well

ioga, line 82: if we won't check for tasktype on agent, this probably won't be necessary as well

master/static/migrations/20220330212121_add_task_stats_table.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TYPE public.stats_type AS ENUM (
+    'QUEUED',
+    'IMAGEPULL'
+);
+
+CREATE TABLE public.task_stats (
+    allocation_id text NOT NULL REFERENCES public.allocations(allocation_id),
+    resource_pool text NOT NULL,
```
ioga, line 8: I don't think this field is needed, because the referenced allocations table has a "resource_pool" column.

agent/internal/containers.go
```diff
@@ -43,8 +43,9 @@ type containerManager struct {
 	Labels        map[string]string `json:"labels"`
 	Devices       []device.Device   `json:"devices"`
 
-	fluentPort int
-	docker     *client.Client
+	fluentPort   int
+	docker       *client.Client
+	ResourcePool string
```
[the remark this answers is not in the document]
gt2345: done

agent/internal/docker.go
```diff
@@ -123,6 +126,23 @@ func (d *dockerActor) pullImage(ctx *actor.Context, msg pullImage) {
 		sendErr(ctx, errors.Wrapf(err, "error checking if image exists: %s", ref.String()))
 		return
 	}
+	if taskNeedsRecording(msg) {
+		ctx.Tell(ctx.Self().Parent(), aproto.DockerImagePull{
+			EndStats: false,
+			Stats: &model.TaskStats{
```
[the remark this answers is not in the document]
gt2345: done

agent/internal/docker.go
```diff
@@ -169,9 +189,14 @@ func (d *dockerActor) pullImage(ctx *actor.Context, msg pullImage) {
 		sendErr(ctx, errors.Wrap(err, "error closing log stream"))
 		return
 	}
+
 	ctx.Tell(ctx.Sender(), imagePulled{})
 }
 
+func taskNeedsRecording(msg pullImage) bool {
+	return msg.TaskType == string(model.TaskTypeTrial)
+}
```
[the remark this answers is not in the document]
gt2345: done

master/static/migrations/20220330212121_add_task_stats_table.up.sql
gt2345: I'd like to cache the resource pool name to avoid joining table after when aggregating

master/internal/resourcemanagers/agent/agent.go
gt2345: done

agent/internal/docker.go
```diff
@@ -123,6 +126,23 @@ func (d *dockerActor) pullImage(ctx *actor.Context, msg pullImage) {
 		sendErr(ctx, errors.Wrapf(err, "error checking if image exists: %s", ref.String()))
 		return
 	}
+	if taskNeedsRecording(msg) {
+		ctx.Tell(ctx.Self().Parent(), aproto.DockerImagePull{
```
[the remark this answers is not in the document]
gt2345: done

master/internal/db/postgres_tasks.go
```diff
@@ -346,6 +347,30 @@ WHERE task_id = $1
...
+
+// RecordTaskEndStats record end stats for tasks.
+func (db *PgDB) RecordTaskEndStats(stats *model.TaskStats) error {
+	now := time.Now().UTC()
+	stats.EndTime = &now
+	_, err := Bun().NewUpdate().Model(stats).Column("end_time").Where(
+		"allocation_id = ? AND event_type = ? AND end_time IS NULL", stats.AllocationID, stats.EventType,
+	).Exec(context.TODO())
+	return err
+}
```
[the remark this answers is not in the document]
gt2345: Revised as discussed

master/pkg/tasks/task.go
gt2345: I want to keep it so that master have access to TaskType

master/pkg/model/task.go
gt2345: This table is used by both bun and plain sql, I wonder if changing to `bun` may cause any problem

ioga: if these aren't necessary here than maybe omitting them keeps the code more concise. 

but if you had to have them here, you could totally have multiple tags space-separated inside backticks, e.g.:
```
type Allocation struct {
	bun.BaseModel `bun:"table:allocations"`

	AllocationID AllocationID     `db:"allocation_id" bun:"allocation_id,notnull,unique"`
	TaskID       TaskID           `db:"task_id" bun:"task_id,notnull"`
	Slots        int              `db:"slots" bun:"slots,notnull"`
	AgentLabel   string           `db:"agent_label" bun:"agent_label,notnull"`
	ResourcePool string           `db:"resource_pool" bun:"resource_pool,notnull"`
	StartTime    *time.Time       `db:"start_time" bun:"start_time"`
	EndTime      *time.Time       `db:"end_time" bun:"end_time"`
	State        *AllocationState `db:"state" bun:"state"`
	IsReady      *bool            `db:"is_ready" bun:"is_ready"`
}
```

master/static/migrations/20220330212121_add_task_stats_table.up.sql
ioga: hmm, won't this join will only happen when updating the aggregate tables? that's a rare periodic query. is this join that bad?

gt2345: After rethinking it, I think you are right, query only happens (on average) once a day, and we already have foreign key. Thank you for pointing this out, I've fixed the aggregate sql

ioga (approved):

PR #3853, opened 2022-04-04
feat: make launch layers generally useful

rb-determined-ai:
## Description

Help text from `python -m determined.launch.horovod`:
```
usage: horovod.py [[HVD_OVERRIDES...] --] [--autohorovod] (--trial TRIAL)|(SCRIPT...)

Launch a script under horovodrun on a Determined cluster, with automatic
handling of IP addresses, sshd containers, and shutdown mechanics.

positional arguments:
  SCRIPT...      script to launch for training

optional arguments:
  -h, --help     show this help message and exit
  --trial TRIAL  use a Trial class as the entrypoint to training. When --trial
                 is used, the SCRIPT positional argument must be omitted.

HVD_OVERRIDES may be a list of arguments to pass directly to horovodrun to
override the values set by Determined automatically. When provided, the list
of override arguments must be terminated by a `--` argument.
```

Help text from `python -m determined.launch.deepspeed`:
```
usage: deepspeed.py [[DS_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)

Launch a script under deepspeed on a Determined cluster, with automatic
handling of IP addresses, sshd containers, and shutdown mechanics.

positional arguments:
  SCRIPT...      script to launch for training

optional arguments:
  -h, --help     show this help message and exit
  --trial TRIAL  use a Trial class as the entrypoint to training. When --trial
                 is used, the SCRIPT positional argument must be omitted.
```

## Test Plan

Added tests for new behaviors

## Commentary

I would like to land this in wednesday's release.  It still needs documentation updates I think.


harness/tests/launch/test_launch.py
```diff
@@ -1,12 +1,12 @@
-from unittest.mock import MagicMock, patch
+from unittest import mock
 
 import determined as det
 from determined.exec import launch
 
 
-@patch("subprocess.Popen")
-@patch("determined.exec.harness.main")
-def test_launch_native(mock_harness: MagicMock, mock_subprocess: MagicMock) -> None:
```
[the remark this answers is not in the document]
rb-determined-ai: oops, yeah I forgot about this whole file.

harness/determined/_core/_distributed.py
```diff
@@ -355,3 +405,10 @@ def __init__(self) -> None:
             cross_rank=0,
             cross_size=1,
         )
+
+
+def _get_default_port_offset() -> int:
```
azhou-determined, line 410: is this helper function necessary? the port offset is defaulted to 0 in init. also will this ever be called outside of a trial task type?

rb-determined-ai: My reasoning was: there are dtrain scenarios where we need to port offsets to ensure multiple sub-entire-node dtrain jobs can coexist on a node.  I don't think there's any other reason to set that port offset.  When you create the `DistributedContext()` manually (the lowest-level API) you can set that value, but that's the sort of detail I'd rather not expose in the helper functions like `DistributedContext.from_horovod()` or `DistributedContext.from_deepspeed()`.  But I'd like those helper functions to do the right thing by default when they are used in sub-entire-node dtrain jobs, so picking the right port_offset somehow is necessary.

And since you'll be adding a `DistributedContext.from_torch_distributed()` soon, having 3 copies of that logic seems like its worth having a helper function.  It doesn't get exposed outside of this file though, which I like.

rb-determined-ai: Perhaps "default" is not a good name for it.  I'm open to renaming it.  "_get_training_port_offset()" maybe?

azhou-determined: i think `_get_training_port_offset` would make more sense, implying that different implementations can set different offsets

azhou-determined:
should we maybe add an e2e test for these changes? seems to cover a lot of "glue" code.

harness/determined/exec/harness.py
```diff
@@ -149,6 +132,6 @@ def main(train_entrypoint: Optional[str]) -> int:
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
-    parser.add_argument("--train-entrypoint")
+    parser.add_argument("train-entrypoint")
```
azhou-determined, line 135: IIRC, i think this "--" is needed by argparse if you have a "-" in the argument name.

azhou-determined, line 135: IIRC, i think this "--named_arg" is needed by argparse if you have a "-" in the argument name like "named-arg".

rb-determined-ai: Um, I don't think that's accurate

harness/tests/launch/test_launch.py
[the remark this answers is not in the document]
rb-determined-ai: just rewrote it.

harness/determined/exec/harness.py
rb-determined-ai: well shut my mouth... I'll fix this.

rb-determined-ai: fixed, using `"train_entrypoint"` now, thank you, and sorry for my sass lol

rb-determined-ai:
> should we maybe add an e2e test for these changes? seems to cover a lot of "glue" code.

What do you have in mind?  I'm thinking that we already have pretty good coverage of these launch layers in e2e testing.

(Full disclosure, I started work on a more general core-api-based script, but I've run into some issues with the core_context.searcher so I can't include that test here.)

harness/determined/launch/deepspeed.py
```diff
@@ -113,7 +103,7 @@ def create_deepspeed_env_file() -> None:
                 f.write(f"{k}={v}\n")
 
 
-def create_run_command(master_address: str, hostfile_path: str) -> List[str]:
+def create_run_command(master_address: str, hostfile_path: str, ds_args: List[str]) -> List[str]:
```
liamcli, line 106: I'm somewhat hesitant to let users pass args to the deepspeed launcher since our launcher is fairly specific to the pdsh backend and many of the arguments need to be set by us.  The only ones I can see value in exposing are --include and --exclude in case specific nodes/gpus aren't working correctly.

liamcli: Let's remove the argument for now and we can add if users request it.

liamcli:
I'm not sure if we should have users supply an arg for --trial or --script.  My concern is that having an arg for --trial may signal that trials are not first class and push users away from our Trial APIs.

rb-determined-ai:
>  My concern is that having an arg for --trial may signal that trials are not first class and push users away from our Trial APIs.

I am not sure why `--trial` would signal that.

Also, in the very long run I expect that PyTorchTrial is the only trial we continue supporting, and I expect that the standard way to use PyTorchTrial will transition from `--trial` to calling a training script which invokes a Trainer API that takes a PyTorchTrial.

harness/determined/launch/deepspeed.py
rb-determined-ai: Removed it now, but watch out... it will be a breaking change to put it back.  Anybody whose script had a `--` arg in it would be broken unless they added an extra `--` right at the beginning of their script.

I could add a check that enforces that if `--` is present in the script that it must also be added at the beginning.  That would be a really weird thing to enforce but later it would let us add DS_OVERRIDES back in without a breaking change.  My judgement call was "that's too weird" so the current PR doesn't do that.

liamcli:
>

liamcli:
> 

Ok, I'm onboard with --trial

rb-determined-ai:
result: gke tests all failed due to quotas, test-e2e-gpu-distributed-nightly never saw any agents, and one real failure that has been resolved on the pr, landing now.

liamcli (approved):
lgtm

PR #3855, opened 2022-04-04
chore: Revert "chore: use userId as url param for user endpoints (#3806)"

johnkim-det:
This reverts commit e35972d83ca67b5fe3b4672248e0db6d940e97b5.

## Description

Previous PR link: https://github.com/determined-ai/determined/pull/3806

Reverting for this sprint/release. Realized it's a breaking API change that doesn't need to be. Will recommit later when done in a non-breaking way.

## Test Plan

No testing required

## Checklist


mapmeld:
In https://github.com/determined-ai/determined/pull/3723/files I included user_id in a SQL file to match the new format, but which would be outside of your commit/revert. Could you change `get_model_by_id.sql` file to return the same fields as `get_model.sql` ?  Or test if it's an issue.

johnkim-det:
> In https://github.com/determined-ai/determined/pull/3723/files I included user_id in a SQL file to match the new format, but which would be outside of your commit/revert. Could you change `get_model_by_id.sql` file to return the same fields as `get_model.sql` ? Or test if it's an issue.

I'm seeing the same SELECT statement in both files, and actually i think fetching both user_id and username for models is not related to what's being reverted here (the breaking change is that users are fetched with user_id instead of username. an earlier commit that returns user_ids in addition to usernames is not being reverted)

mapmeld (approved):
LGTM then ✅

PR #3859, opened 2022-04-05
feat: generic checkpoints

stoksc:
##### Checkpoint class
- Move training-specific attributes into an optional training substructure
- Add references to task_id and allocation_id
- Add a metadata Dict for holding user-defined checkpoint metadata along with system-defined metadata
      - `format`
      - `framework`
      - `determined_version`
      - `steps_completed`
##### Database
Instead of a potentially heavy migration of existing checkpoints, use
views to obtain hydrated checkpoint objects backed by both old and new tables.
#### steps_completed instead of latest_batch
Across the board,`latest_batch` has been renamed `steps_completed` to emphasize the abstract nature of units and the fact that it equals the number of units of work completed
#### Make Core API public
- Move it from determined._core to determined.core
- Provide a tutorial showing how to augment a simple training script with Core API features

## Test Plan
- [ ] @mpkouznetsov is working on adding coverage
- [ ] i will manually test with curl, CLI, etc.

## Commentary (optional)
We probably need a new API to get at generic checkpoints for them to be remotely useful. What's the plan for this @rb-determined-ai / @mpkouznetsov ?

Also, I plan to follow up to 'bun-ify' some of these things, but thought it was most critical, first, to unblock the projects that are held up by this (this that work may take more time, and frankly the work can probably be done faster in 2 pieces).

## Checklist


master/static/srv/get_checkpoints_for_trial.sql
```diff
@@ -1,23 +1,4 @@
...
-    (v.metrics->'validation_metrics'->>(e.config->'searcher'->>'metric'))::float8 AS searcher_metric
-FROM checkpoints c
-LEFT JOIN validations v ON v.total_batches = c.total_batches AND v.trial_id = c.trial_id
-JOIN trials t ON c.trial_id = t.id
-JOIN experiments e ON t.experiment_id = e.id
-WHERE t.id = $1
-ORDER BY end_time DESC
+SELECT *
+FROM proto_checkpoints_view c
+WHERE c.training->>'trial_id' = $1
```
mpkouznetsov, line 3: I think the same CAST(... AS integer) is needed here as in get_checkpoints_for_experiment.sql

stoksc: good catch, i think only the webui uses this right now so my manual testing missed it.

master/pkg/model/experiment.go
```diff
@@ -422,6 +440,49 @@ type Checkpoint struct {
...
+// CheckpointTrainingMetadata is a substruct of checkpoints encapsulating training specific
+// information.
+type CheckpointTrainingMetadata struct {
+	TrialID           int     `db:"trial_id"`
+	ExperimentID      int     `db:"experiment_id"`
+	ExperimentConfig  JSONObj `db:"experiment_config"`
+	HParams           JSONObj `db:"hparams"`
+	TrainingMetrics   JSONObj `db:"training_metrics"`
+	ValidationMetrics JSONObj `db:"validation_metrics"`
+	SearcherMetrics   JSONObj `db:"searcher_metric"`
```
mpkouznetsov, line 464: There is an inconsistency between this being a JSONObj and model.experiment considering it a float64. Which is it?

gh-determined-ai (approved):

examples/tutorials/core_api/1_metrics.py
```diff
@@ -0,0 +1,33 @@
...
+
+def main(core_context, increment_by):
+    x = 0
+    for batch in range(100):
+        x += increment_by
+        time.sleep(.1)
+        print("x is now", x)
+        # NEW: report training metrics to Determined.
+        if batch % 10 == 9:
+            core_context.train.report_training_metrics(latest_batch=batch, metrics={"x": x})
```
ioga, line 24: it'd seem more aligned if x was reported at the same frequency as `print`?

examples/tutorials/core_api/3_hpsearch.py
```diff
@@ -0,0 +1,87 @@
...
+    batch = starting_batch
+    last_checkpoint_batch = None
+    for op in core_context.searcher.operations():
+        # NEW: we will use a while loop for easier accounting of absolute lengths.
+        while batch < op.length:
+            x += increment_by
+            time.sleep(.1)
+            print("x is now", x)
+            batch += 1
+            if batch % 10 == 0:
```
ioga, line 45: this used to be `== 9` in previous stages

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -40,47 +50,37 @@ class Checkpoint(object):
...
-        self.model_version = model_version
-        self.model_id = model_id
         self.metadata = metadata
+        self.state = state
+        self.training = training
+
+    def get_training(self) -> CheckpointTrainingMetadata:
+        if self.training is None:
+            raise RuntimeError("Non-training checkpoints are not yet supported.")
+        return self.training
```
ioga, line 75: 1. I don't think such getters are necessary when there's no extra I/O (like network fetch) is involved. so maybe use `@property`? 
2. looking at `checkpoint.get_training()` or `checkpoint.training` it's not clear for me what the result is. so maybe `training_metadata`? i.e.:


```suggestion
    @property
    def training_metadata(self) -> CheckpointTrainingMetadata:
        if self._training is None:
            raise RuntimeError("Non-training checkpoints are not yet supported.")
        return self._training
```

3. how can a user tell from API if a checkpoint has this training metadata? do we need a `has_training_metadata` method? why is throwing an exception here necessary? Maybe no need for this property/getter at all, just have it `None` when it's not available?

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -40,47 +50,37 @@ class Checkpoint(object):
...
-        self.determined_version = determined_version
-        self.model_version = model_version
-        self.model_id = model_id
         self.metadata = metadata
+        self.state = state
+        self.training = training
+
+    def get_training(self) -> CheckpointTrainingMetadata:
+        if self.training is None:
+            raise RuntimeError("Non-training checkpoints are not yet supported.")
```
ioga, line 74:
```suggestion
            raise NotImplementedError("Non-training checkpoints are not yet supported.")
```

harness/determined/core/_checkpoint.py
```diff
@@ -222,32 +222,30 @@ def _report_checkpoint(
...
+            "metadata": metadata,
+            "state": bindings.determinedcheckpointv1State.STATE_COMPLETED,  # YYY a new const?
         }
         logger.info(f"Reported checkpoint to master {storage_id}")
-        self._session.post(self._api_path, data=det.util.json_encode(body))
+        api_path = "/api/v1/checkpoints"
+        self._session.post(api_path, data=det.util.json_encode(body))
+        # XXX: confirm that this implementation is ok for now
+        # instead of
+        # bindings.post_ReportCheckpoint(self._session, body=ckpt)
```
ioga, line 248: my vote is for bindings.

master/internal/api_experiment.go
```diff
@@ -664,23 +686,41 @@ func (a *apiServer) GetExperimentCheckpoints(
...
 		}
 
-		if len(req.ValidationStates) != 0 && !found {
-			return false
+		if req.OrderBy == apiv1.OrderBy_ORDER_BY_DESC {
+			aj, ai = ai, aj
 		}
 
-		return true
+		switch req.SortBy {
```
ioga, line 704: do/can we plan to do these in SQL?

e2e_tests/tests/experiment/test_core.py
```diff
@@ -295,22 +292,32 @@ def test_end_to_end_adaptive() -> None:
...
-    assert checkpoint.metadata == {"testing": "metadata"}
+    # Beginning with 0.18 the system contributes a few items to the dict
+    assert checkpoint.metadata.get("testing") == "metadata"
+    assert checkpoint.metadata.keys() == {"format", "framework", "latest_batch", "testing"}
     assert checkpoint.metadata == db_check.metadata
 
     checkpoint.add_metadata({"some_key": "some_value"})
     db_check = d.get_checkpoint(checkpoint.uuid)
-    assert checkpoint.metadata == {"testing": "metadata", "some_key": "some_value"}
+    assert checkpoint.metadata.items() > {"testing": "metadata", "some_key": "some_value"}.items()
```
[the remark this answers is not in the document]
mpkouznetsov: `>` on sets is a super-set check, as I am sure you know. What I did not know until recently is that `dict.items()` is a set-like data structure if values are hashable, which in this test they are.
From https://docs.python.org/3/library/stdtypes.html#dict-views
> If all values are hashable, so that (key, value) pairs are unique and hashable, then the items view is also set-like.

```
>>> d1 = {'abc': 1, 'de': 2}
>>> d2 = {'aaa': 34, 'de': 2, 'abc': 1}
>>> d1.items() < d2.items()
True
>>> {('de', 2)} < d2.items()
True
>>> {('de', 3)} < d2.items()
False
>>> {('abc', 1), ('aaa', 34)} < d1.items()
False
>>> {('abc', 1), ('aaa', 34)} < d2.items()
True
```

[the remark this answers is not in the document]
ioga: oh wow I did not know this either

harness/determined/core/_checkpoint.py
mpkouznetsov: I know it is :) I will give it a try.

harness/determined/common/experimental/checkpoint/_checkpoint.py
mpkouznetsov: This is a good point. In CLI I am going to check for Noneness and render empty string when training is None. There is one usage, in experimental.Checkpoint.download() that relies on it being a training checkpoint to get to the checkpoint_storage. I am going to modify it; let me know what you think.

examples/tutorials/core_api/2_checkpoints.py
```diff
@@ -0,0 +1,88 @@
...
+def load_state(trial_id, path):
+    import pathlib
+    path = pathlib.Path(path)
+    with path.joinpath("state").open("r") as f:
+        x, latest_batch, ckpt_trial_id = [int(field) for field in f.read().split(",")]
+    if ckpt_trial_id == trial_id:
+        return x, latest_batch
+    else:
+        # This is a new trial; load the "model weight" but not the batch count.
+        return x, 0
```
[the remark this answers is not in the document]
mpkouznetsov: yes

examples/tutorials/core_api/1_metrics.py
mpkouznetsov: Do you mean `print` is too frequent? I would agree with that.

ioga: I mean it'd be nice to have 1:1 mapping between manual prints and the metrics "showing up in the webui" - should make it clearer where these are coming from.

examples/tutorials/core_api/2_checkpoints.py
```diff
@@ -0,0 +1,93 @@
...
+import pathlib
+import sys
+import time
+
+import determined as det
+
+
+# NEW: given a checkpoint_directory of type pathlib.Path, save our state to a file.
+# You can save multiple files, and use any file names or directory structures.
+# All files nested under `checkpoint_directory` path fill be included into the checkpoint.
```
ioga, line 26: I had a typo
```suggestion
# All files nested under `checkpoint_directory` path will be included into the checkpoint.
```

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -132,13 +126,17 @@ def download(self, path: Optional[str] = None) -> str:
         if not any(p.exists() for p in potential_metadata_paths):
             # If the target directory doesn't already appear to contain a
             # checkpoint, attempt to fetch one.
-            if self.experiment_config["checkpoint_storage"]["type"] == "shared_fs":
-                src_ckpt_dir = self._find_shared_fs_path()
+            if not self.training:
```
ioga, line 129: lgtm, but as a nitpick:
```suggestion
            if self.training is None:
```

ioga (commented):
changes outside of the remaining `XXX` and `YYY` lgtm. what's the plan for them, the metadata conundrum in particular?

examples/tutorials/core_api/1_metrics.py
mpkouznetsov: it's a bit tricky for me to guess Ryan's intent. But I think the idea was to make explicit the "learning" - processing of individual batches. You will notice that 0_start does not collect metrics at all:
```
[2022-04-25T18:48:44.442200Z] bbedc350 || INFO: [1] root: Launching: ['sh', '-c', './0_start.py']
[2022-04-25T18:48:44.567337Z] bbedc350 || x is now 1
[2022-04-25T18:48:44.672570Z] bbedc350 || x is now 2
[2022-04-25T18:48:44.774656Z] bbedc350 || x is now 3
[2022-04-25T18:48:44.879599Z] bbedc350 || x is now 4
[2022-04-25T18:48:44.981441Z] bbedc350 || x is now 5
[2022-04-25T18:48:45.082459Z] bbedc350 || x is now 6
[2022-04-25T18:48:45.184186Z] bbedc350 || x is now 7
[2022-04-25T18:48:45.285437Z] bbedc350 || x is now 8
[2022-04-25T18:48:45.388754Z] bbedc350 || x is now 9
[2022-04-25T18:48:45.491606Z] bbedc350 || x is now 10
[2022-04-25T18:48:45.595408Z] bbedc350 || x is now 11
```
The next step, 1_metrics, keeps printing the value of x at every "batch" but in addition reports the metrics every 10 batches:
```
[2022-04-25T18:49:19.250446Z] 87c021c1 || INFO: [1] root: Launching: ['sh', '-c', './1_metrics.py']
[2022-04-25T18:49:20.788894Z] 87c021c1 || x is now 1
[2022-04-25T18:49:20.891923Z] 87c021c1 || x is now 2
[2022-04-25T18:49:20.995577Z] 87c021c1 || x is now 3
[2022-04-25T18:49:21.099430Z] 87c021c1 || x is now 4
[2022-04-25T18:49:21.204740Z] 87c021c1 || x is now 5
[2022-04-25T18:49:21.308930Z] 87c021c1 || x is now 6
[2022-04-25T18:49:21.411429Z] 87c021c1 || x is now 7
[2022-04-25T18:49:21.516599Z] 87c021c1 || x is now 8
[2022-04-25T18:49:21.619109Z] 87c021c1 || x is now 9
[2022-04-25T18:49:21.720999Z] 87c021c1 || x is now 10
[2022-04-25T18:49:21.721534Z] 87c021c1 || INFO:determined.core:report_training_metrics(latest_batch=9, metrics={'x': 10})
[2022-04-25T18:49:21.868868Z] 87c021c1 || x is now 11
[2022-04-25T18:49:21.971171Z] 87c021c1 || x is now 12
```

mpkouznetsov: If you are not convinced, I can reduce the frequency to 1 in 10.

ioga: okay, maybe I am overthinking this.

mpkouznetsov:
> changes outside of the remaining XXX and YYY lgtm. what's the plan for them, the metadata conundrum in particular?

Addressing those probably deserves a PR against this branch. And potentially a discussion.

examples/tutorials/core_api/3_hpsearch.py
rb-determined-ai: yeah, this was intentional.    The way things were counted changed slightly so this needed to as well.

examples/tutorials/core_api/2_checkpoints.py
```diff
@@ -0,0 +1,93 @@
...
+            # NEW: write checkpoints at regular intervals to limit lost progress in case of a crash
+            # during training.
+            checkpoint_metadata = {"latest_batch": batch}
+            with core_context.checkpoint.store_path(checkpoint_metadata) as (path, uuid):
+                save_state(x, batch, trial_id, path)
+
+            # NEW: check for a preemption signal.  This could originate from a higher-priority task
+            # bumping us off the cluster, or for a user pausing the experiment via the WebUI or CLI.
+            if core_context.preempt.should_preempt():
+                # In this case, we just saved a checkpoint, so we can just shut down early and
```
rb-determined-ai, line 69: nit: in this ~case~ example

The phrase "in this case" might read as "when should_preempt() returns true" rather than "when running this particular example"

(I know, this is me commenting on my own example)

examples/tutorials/core_api/2_checkpoints.py
```diff
@@ -0,0 +1,93 @@
...
+    if latest_checkpoint is not None:
+        with core_context.checkpoint.restore_path(latest_checkpoint) as path:
+            x, starting_batch = load_state(trial_id, path)
+
+    for batch in range(starting_batch, 100):
+        x += increment_by
+        time.sleep(.1)
+        print("x is now", x)
+        if batch % 10 == 9:
+            core_context.train.report_training_metrics(latest_batch=batch, metrics={"x": x})
```
rb-determined-ai, line 58: just realized, I think this is going to save the wrong latest batch into the checkpoint; it saves the after-increment X but the before-increment batch number.

examples/tutorials/core_api/3_hpsearch.py
```diff
@@ -0,0 +1,90 @@
...
+    batch = starting_batch
+    last_checkpoint_batch = None
+    for op in core_context.searcher.operations():
+        # NEW: we will use a while loop for easier accounting of absolute lengths.
+        while batch < op.length:
+            x += increment_by
+            time.sleep(.1)
+            print("x is now", x)
+            batch += 1
+            if batch % 10 == 9:
```
rb-determined-ai, line 48: this is going to checkpoint at slightly wrong times.  What do you think of switching to `while batch < op.length` in all the examples to avoid counting inconsistencies between examples?  Originally I avoided that because the `for i in range(n)` is so idiomatic in the early examples.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -3,6 +3,7 @@
 import pathlib
 import shutil
 import warnings
+from dataclasses import dataclass
```
rb-determined-ai, line 6: follow the google style guide for imports:

```python
import dataclasses
...
@dataclasses.dataclass
class ...
```

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -171,20 +167,8 @@ def write_metadata_file(self, path: str) -> None:
...
-                    "trial_id": self.trial_id,
-                    "hparams": self.hparams,
-                    "experiment_config": self.experiment_config,
-                    "metadata": self.metadata,
-                },
-                f,
-                indent=2,
-            )
+            # YYY: only dump user metadata
+            json.dump(self.metadata, f, indent=2)
```
rb-determined-ai, line 171: why is this a `# YYY`?  I think this is the right strategy.

harness/determined/core/_checkpoint.py
```diff
@@ -79,8 +79,13 @@ def upload(
         ckpt_dir = os.fspath(ckpt_dir)
 
         storage_id = str(uuid.uuid4())
-        self._storage_manager.upload(src=ckpt_dir, dst=storage_id)
         resources = self._storage_manager._list_directory(ckpt_dir)
+
+        # add metadata pre-upload but without counting it among resources
+        if metadata is not None:
+            self._write_metadata_file(ckpt_dir, metadata)
```
rb-determined-ai, line 86: maybe always write it, but write `metadata or {}` instead?

harness/determined/core/_checkpoint.py
```diff
@@ -158,6 +163,8 @@ def store_path(
         with self._storage_manager.store_path(storage_id) as path:
             yield path, storage_id
             resources = self._storage_manager._list_directory(path)
+            self._write_metadata_file(os.fspath(path), metadata)
```
rb-determined-ai, line 166: `metadata or {}` here too

harness/determined/core/_checkpoint.py
```diff
@@ -213,6 +220,12 @@ def delete(self, storage_id: str) -> None:
         """
         self._storage_manager.delete(storage_id)
 
+    def _write_metadata_file(self, ckpt_dir: str, metadata: Optional[Dict[str, Any]]) -> None:
```
rb-determined-ai, line 223: readability nit: disallow `metadata=None` and leave the defaulting behavior to the calling code, or rename this to `_maybe_write_metadata_file` (which seems less desirable to me).

IMO, a function called write_some_file should always write that file.

harness/determined/core/_checkpoint.py
```diff
@@ -222,32 +235,27 @@ def _report_checkpoint(
...
+            resources={k: str(v) for k, v in resources.items()},
+            taskId=self._task_id,
+            training=bindings.v1CheckpointTrainingMetadata(),
+            uuid=storage_id,
+            reportTime=datetime.now(timezone.utc).isoformat(),
+            state=bindings.determinedcheckpointv1State.STATE_COMPLETED,
+        )
+        bindings.post_ReportCheckpoint(self._session, body=ckpt)
         logger.info(f"Reported checkpoint to master {storage_id}")
-        self._session.post(self._api_path, data=det.util.json_encode(body))
```
rb-determined-ai, line 250: did you figure out a way to use our custom `json_encode()` with our bindings?  Otherwise, this will cause problems when users try to put e.g. metrics into their metadata (which seems pretty reasonable to me).

examples/tutorials/core_api/3_hpsearch.py
mpkouznetsov: in examples 1 and 2 it means `while batch < 100`

examples/tutorials/core_api/2_checkpoints.py
mpkouznetsov: I think that is perfectly fine: at batch 9 we report x = 10*increment_by

harness/determined/core/_checkpoint.py
mpkouznetsov: no I am not using it. What did I miss?
```
            checkpoint_metadata = {"latest_batch": batch, "metrics": {"x": x}}
            with core_context.checkpoint.store_path(checkpoint_metadata) as (path, uuid):
                save_state(x, batch, trial_id, path)
```
^ this works.
What wouldn't?

rb-determined-ai: There's quite a few corner cases we've had to add to `json_encode()` in the past.  Some of the more important ones are np arrays, nans, and infs.

mpkouznetsov: ok, no I did not take care of that. We agreed to carve it out and address it later: https://determinedai.atlassian.net/browse/DET-7166

examples/tutorials/core_api/2_checkpoints.py
mpkouznetsov: We agreed to deal with it here https://github.com/determined-ai/determined/pull/4048

master/test/integration/api/api_checkpoints_test.go
```diff
@@ -346,28 +346,35 @@ func testGetTrialCheckpoints(
 
 }
 
-func createPrereqs(t *testing.T, db *db.PgDB) (
+func createPrereqs(t *testing.T, pgDB *db.PgDB) (
```
mpkouznetsov, line 349: This change is catching up with https://github.com/determined-ai/determined/pull/3959 and https://github.com/determined-ai/determined/pull/3946 .

rb-determined-ai (approved):
ship it!

But please do take the time write a detailed commit message, this PR includes quite a few details and quite a few decisions more than just generic checkpoints.

mpkouznetsov:
Merging despite some GPU-based tests failing:
- GKE could not spin up due to insufficient resources
- model_hub / transformers test failures are not related to this work

PR #3861, opened 2022-04-05
feat: make core api searcher multiworker-safe

rb-determined-ai:
## Description

core_context.searcher.ops() needs to be coordinated across workers
the same way that core_context.preemption.should_preempt() is
coordinated across workers.

This change also includes tests for the new functionality.


rb-determined-ai:
did not mean to make upstream PR

PR #3869, opened 2022-04-06
feat: Search models by name and description substring [DET-6939]

mapmeld:
## Description

- On `/api/v1/models` endpoint, the name and description search use ILIKE and % for case-insensitive substring matches
- Change `clearModelName` model name collision check - no longer uses the GetModels function
- Remove unneeded NameCaseInsensitive parameter (the endpoint is now case-insensitive)

## Test Plan

Use `/api/v1/models?name=_&description=_` to make queries of name and description.
- Confirm description substring match
- Confirm name substring match
- On the create model dialog or `det model create` confirm that new models are not allowed to have name collisions

## Checklist


EmilyBonar (approved):
This seems like a good change, but it's breaking tests. Make sure those pass before merging.

PR #3871, opened 2022-04-06
feat: make core.Searcher multiworker-safe

rb-determined-ai:
## Description

core_context.searcher.ops() needs to be coordinated across workers
the same way that core_context.preemption.should_preempt() is
coordinated across workers.

This change also includes tests for the new functionality.

## Commentary

I labeled this as a "user-facing change" because it's a deviation from previously-advertised Core API specs, even though technically it's still not a released API so it isn't any sort of breaking change.

### DIFF
--- a/harness/determined/_core/_searcher.py
+++ b/harness/determined/_core/_searcher.py
@@ -93,9 +93,8 @@ class Searcher:
 
     .. code:: python
 
-       # We'll pretend we configured we configured the searcher
-       # in terms of batches, so we will interpet the the op.length as a
-       # count of batches.
+       # We'll pretend we configured the searcher in terms of batches,
+       # so we will interpet the the op.length as a count of batches.
        # Note that you'll have to load your starting point from a
        # checkpoint if you want to support pausing/continuing training.
        batches_trained = 0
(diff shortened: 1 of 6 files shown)


harness/determined/_core/_searcher.py
```diff
@@ -119,12 +126,14 @@ class Searcher:
     def __init__(
         self,
         session: Session,
+        dist: _core.DistributedContext,
```
mpkouznetsov, line 129: Could you remove the extra `we configured` on line 96?

harness/tests/test_ipc.py
```diff
@@ -148,92 +148,61 @@ def test_zmq_server_client() -> None:
...
-    expect = [
-        set(range(cross_rank * local_size, (cross_rank + 1) * local_size))
-        for cross_rank, _ in itertools.product(range(cross_size), range(local_size))
-    ]
-    assert results == expect, "not all threads ran allgather_local correctly"
-
-    # Close all contexts.
-    for context in contexts:
-        context.close()
+    with parallel.Execution(size, local_size=local_size, make_distributed_context=False) as pex:
```
mpkouznetsov, line 151: lovely

mpkouznetsov (approved):
Makes sense!

PR #3872, opened 2022-04-06
chore: Clean up tests with fewer Optional types and asserts

mapmeld:
## Description

During earlier changes to e2e_tests (#3539) I avoided issues with `Optional[]` types with some additional code -- `assert training_workloads is not None` or `for w in (workloads or []):`. Even if a trial has 0 workloads, the API will return an empty array, so we can mark this field as required and avoid `Optional[]` typing.  Similar reasoning to require metrics on a `MetricsWorkload`.

Marking these as required allows us to remove or simplify some code in e2e_tests.

I also replace `workloads_for_mode(wl, "training")` with `workloads_with_training(wl)` and other specific functions to avoid type differences between MetricWorkload and CheckpointWorkload.

## Checklist


stoksc (approved):
lgtm, modulo whatever is going on with the react tests

PR #3873, opened 2022-04-06
feat: show agent version in `/agents` and `det agent list` [DET-6847]

dzhu:
## Description

Simply make the agent report its Determined version as a query parameter
when it connects to the master, and then thread that information through
to the API endpoint and CLI.

## Test Plan

- [x] manual: start cluster, check `det a` output, kill agent, change `VERSION` file, rebuild and rerun agent, check `det a` output
- [x] add test that API endpoint reports current version

## Commentary (optional)

The test should ideally be a cross-version test so that we can confirm that the endpoint returns the actual agent version and not always the current master version, but we can't do that right now because older agents don't report their versions at all. (At least using older agents just leads to the master treating the version as an empty string, not erroring out.)

### DIFF
--- a/e2e_tests/tests/cluster/test_agent.py
+++ b/e2e_tests/tests/cluster/test_agent.py
@@ -1,0 +1,19 @@
+import os
+
+import pytest
+
+import determined
+from determined.common import api
+from tests import config as conf
+
+
+# TODO: This should be marked as a cross-version test, but it can't actually be at the time of
+# writing, since older agent versions don't report their versions.
+@pytest.mark.e2e_cpu
+def test_agent_version() -> None:
+    # DET_AGENT_VERSION is available and specifies the agent version in cross-version tests; for
+    # other tests, this evaluates to the current version.
+    target_version = os.environ.get("DET_AGENT_VERSION") or determined.__version__
+    agents = api.get(conf.make_master_url(), "agents").json()
+
+    assert all(agent["version"] == target_version for agent in agents.values())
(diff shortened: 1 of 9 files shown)


e2e_tests/tests/cluster/test_agent.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+# TODO: This should be marked as a cross-version test, but it can't actually be at the time of
+# writing, since older agent versions don't report their versions.
+@pytest.mark.e2e_cpu
+def test_agent_version() -> None:
+    # DET_AGENT_VERSION is available and specifies the agent version in cross-version tests; for
+    # other tests, this evaluates to the current version.
+    target_version = os.environ.get("DET_AGENT_VERSION") or determined.__version__
+    agents = api.get(conf.make_master_url(), "agents").json()
```
ioga, line 17: it'd be great to start using the "new" (`/api/v1/`) APIs instead of the plain ones. also maybe make use of the autogenerated python bindings? or python sdk in the future.

e2e_tests/tests/cluster/test_agent.py
```diff
@@ -0,0 +1,19 @@
+import os
+
+import pytest
+
+import determined
+from determined.common import api
+from tests import config as conf
+
+
+# TODO: This should be marked as a cross-version test, but it can't actually be at the time of
```
ioga, line 10: 👍

ioga (approved):

e2e_tests/tests/cluster/test_agent.py
dzhu: Makes sense, though we want this in for the release, so I'm going to land now and follow up with that switch.

PR #3876, opened 2022-04-07
fix: avoid double-timestamps in logs

rb-determined-ai:
We attach timestamps to all logs after-the-fact, so the python logging
library needs to be configured to not also emit timestamps.

Apparently, this was broken in distributed training since 0.17.0, and
recently it became broken in non-distributed training as well.

The source of the bug was that there used to be two common paths by
which we configured logs, which I did not realize, and so in the push
architecture refactor I unwittingly took the worse one and used it on a
new code path.

This fix deletes that way to configure logs and uses the remaining way
everywhere.

### DIFF
--- a/harness/determined/common/_logging.py
+++ b/harness/determined/common/_logging.py
@@ -11,6 +11,6 @@ def set_logger(debug_enabled: bool) -> None:
 
     handler = logging.StreamHandler(sys.stdout)
     handler.setLevel(logging.DEBUG if debug_enabled else logging.INFO)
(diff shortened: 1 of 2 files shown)


varlogtim (changes requested):
I never actually merged this: https://github.com/determined-ai/determined/pull/3785, because I was intending to add named loggers for harness so that we don't have "root" listed in places.

Can we close that out and add it to this review?

varlogtim (approved):
Yeah, I realized I could edit the file during review. Can we add the following?

PR #3880, opened 2022-04-07
fix: update checkpoint decoder in react

mpkouznetsov:
## Description

Update the decoder to avoid errors. There could be additional UI changes necessary; the purpose of this update is, again, just to avoid ts compilation errors.


stoksc (approved):
could've sworn i ran build-bindings. that's all this is from, right?

mpkouznetsov:
> could've sworn i ran build-bindings. that's all this is from, right?

Part of it is bindings out of sync and part of it is data structure changes just not reflected in the typescript code.

PR #3888, opened 2022-04-12
feat: 

rb-determined-ai:
## Description

Accepting a `chief_only: bool` argument is the classic "bools can't be extended" design flaw.  This was apparent when the first real user to use the preemption API wanted to have each worker ask the chief independently for the preemption signal.

This PR rewrites the two chief_only cases to be enums instead of bools.  In the case of the searcher, it makes absolutely no sense to offer a workers-ask-master mode, so it still only accepts two values, but even those values are clearer as to what is actually happening under the hood.

## Test Plan

I will add tests for the Preemption API before landing.


harness/determined/_core/_preemption.py
```diff
@@ -221,10 +251,19 @@ def acknowledge_preemption_signal(self) -> None:
 class DummyPreemption(Preemption):
     """Present an Preemption API that never returns True."""
```
mpkouznetsov, line 252:
```suggestion
    """Present a Preemption API that never returns True."""
```

harness/determined/_core/_preemption.py
```diff
@@ -236,7 +275,25 @@ def __enter__(self) -> "Preemption":
...
-    def should_preempt(self, chief_only: bool = False, auto_ack: bool = True) -> bool:
+    def should_preempt(self, auto_ack: bool = True) -> bool:
+        if not self._started:
+            # Match the requirements of the real Preemption class.
+            raise RuntimeError(
+                "you cannot call Preemption.should_preempt() before Preemption.start()"
+            )
+        if self._dist.rank == 0:
+            if self._preempt_mode == PreemptMode.WorkersAskChief:
+                # Even though we always return False, preserve the synchronization behavior to avoid
```
mpkouznetsov, line 286: 👍

mpkouznetsov (approved):

PR #3892, opened 2022-04-12
fix: hide sizeChanger on RoutePagination

johnkim-det:
## Description

By default, the antd Pagination component shows a dropdown to set the page size if total > 50. Hide this for the RoutePagination use case.

Screen Shot 2022-04-12 at 12 19 21 PM

## Test Plan

- [ ] with an experiment that has more than 50 trials, open the trial details page for one of the trials. make sure this dropdown does not appear there.

## Checklist


hkang1 (approved):
LGTM!

Unable to test it via netlify :(

PR #3893, opened 2022-04-12
fix: use otel only if enabled 

nrajanee:
## Description

Ensure that otel middleware is used only when otel is enabled. 
Add otel_enabled (default = false) and otel_endpoint to config (default = localhost:4317) . 

## Test Plan
**When otel_enabled is not set to true in the devcluster.yaml file:** 
- [ ]  Run make devcluster
- [ ] Ensure that you don't see any logs like this: 
`max retry time elapsed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:4317: connect: connection refused"`

**When otel_enabled is set to true in devcluster.yaml file** (`telemetry:
          otel_enabled: true`): 
- [ ] Run make devcluster'
- [ ] Set endpoint telemetry:
          otel_endpoint: {insert_endpoint}` in devcluster.yaml (endpoint eg: "localhost:4317". Note port number here is 4317.) 
- [ ]  Run docker run --rm -v “$HOME/Desktop/opentel/config.yaml" -p 1888:1888 -p 8888:8888 -p 8889:8889 -p 13133:13133 -p **{port_number}**:4317 -p 4318:4318 otel/opentelemetry-collector to start the opentel collector. 
- [ ] Go to localhost:8080 and check that you are seeing logs (with http.server_name: STRING(determined-master)) when you hit diff endpoints.


master/internal/core.go
```diff
@@ -785,8 +785,11 @@ func (m *Master) Run(ctx context.Context) error {
 		m.echo.Use(auditLogMiddleware())
 	}
 
-	otelConfig()
-	m.echo.Use(otelecho.Middleware("determined-master"))
+	if m.config.Telemetry.OtelEnabled {
+		endpoint := m.config.Telemetry.OtelExportedOtlpEndpoint
+		otelConfig(endpoint)
```
stoksc, line 790: I also feel `otelConfig` should be `configureOtel` or something instead.
```suggestion
		otelConfig(m.config.Telemetry.OtelExportedOtlpEndpoint)
```

master/internal/config/config.go
```diff
@@ -89,9 +89,11 @@ func DefaultConfig() *Config {
 		HarnessPath: "/opt/determined",
 		Root:        "/usr/share/determined/master",
 		Telemetry: config.TelemetryConfig{
-			Enabled:          true,
-			SegmentMasterKey: DefaultSegmentMasterKey,
-			SegmentWebUIKey:  DefaultSegmentWebUIKey,
+			Enabled:                  true,
+			OtelEnabled:              false,
+			OtelExportedOtlpEndpoint: "localhost:4317",
```
stoksc, line 94: since this default is problem wrong more than it is right, can we just drop it?

master/cmd/determined-master/init.go
```diff
@@ -129,4 +129,6 @@ func registerConfig() {
 		defaults.Telemetry.SegmentMasterKey, "the Segment write key for the master")
 	registerString(flags, name("telemetry", "segment-webui-key"),
 		defaults.Telemetry.SegmentWebUIKey, "the Segment write key for the WebUI")
+	registerBool(flags, name("telemetry", "otel_enabled"), defaults.Telemetry.OtelEnabled, "enable otel")
+	registerString(flags, name("telemetry", "otel_endpoint"), defaults.Telemetry.OtelExportedOtlpEndpoint, "set otel endpoint")
```
ioga, line 133: looking at the code above, it uses dash (e.g. `segment-webui-key`), so probably

```suggestion
	registerBool(flags, name("telemetry", "otel-enabled"), defaults.Telemetry.OtelEnabled, "enable otel")
	registerString(flags, name("telemetry", "otel-endpoint"), defaults.Telemetry.OtelExportedOtlpEndpoint, "set otel endpoint")
```

(and update the docs accordingly)

docs/sysadmin-basics/cluster-config.txt
```diff
@@ -168,6 +168,11 @@ To disable telemetry reporting in both the master and the WebUI, start the maste
 setting an environment variable, as with any other configuration option). Disabling telemetry
 reporting will not affect the functionality of Determined in any way.
 
+Separate from the telemetry reporting mentioned above we also provide functionality to use opentelemetry to collect traces.
```
ioga, line 171: I'd add a section header and a hyperlink (similar to `.. _master-configuration:` below), so this docs can be more easily referred to.

ioga (approved):

philipnrmn:
Looks great! One suggestion is that this should probably live in the `observability` configuration block rather than telemetry.

PR #3896, opened 2022-04-13
chore: sweeping rename of Core API components

rb-determined-ai:
Part 1: shorten names and use a consistent part of speech:

    core_context.training -> core_context.train
    core_context.checkpointing -> core_context.checkpoint
    core_context.preemption -> core_context.preempt
    # core_context.distributed stays because it matches e.g. torch.distributed
    # core_context.searcher stays because we like it better.

Part 2: Change the class name of each component of the Core API to
include "Context".  This won't really affect users but it's a bit clearer.

    core.Searcher -> core.SearcherContext
    core.Training -> core.TrainContext
    core.Preemption -> core.PreemptContext
    core.Checkpointing -> core.CheckpointContext
    # core.DistributedContext remains unchanged.

Part 3: avoid abbreviations

    core_context.search.ops -> core_context.search.operations
    core.SearcherOp -> core.SearcherOperation

Part 4: anything that must only be called by the chief is named report_*

    SearcherOperation.complete() -> SearcherOperation.report_completed()

Part 5: expose DistributedContext._zmq_*, stripping the _zmq_ prefix.
Originally, the prefix was meant to make it clear that
context.distributed.allgather() was definitely not hvd.allgather(), but
We care enough about not marrying the name to a particular communication
library that we are willing to sacrifice the clear hvd/zmq distinction
to avoid that.

    DistributedContext._zmq_allgather() -> DistributedContext.allgather()
    DistributedContext._zmq_gather() -> DistributedContext.gather()
    DistributedContext._zmq_broadcast() -> DistributedContext.broadcast()
    # etc

## Commentary

This PR adds a small bit of warning text to the _zmq_* methods, but otherwise is just one big rename PR.

## Test Plan

Normal tests should still pass.

### DIFF
--- a/harness/tests/test_ipc.py
+++ b/harness/tests/test_ipc.py
@@ -164,23 +164,23 @@ def test_distributed_context(cross_size: int, local_size: int, force_tcp: bool) 
             )
 
         # Perform a broadcast.
-        results = pex.run(lambda: contexts[pex.rank]._zmq_broadcast(pex.rank))  # type: ignore
+        results = pex.run(lambda: contexts[pex.rank].broadcast(pex.rank))  # type: ignore
         assert results == [0] * size, "not all threads ran broadcast correctly"
 
         # Perform a local broadcast.
-        results = pex.run(lambda: contexts[pex.rank]._zmq_broadcast_local(pex.rank))
+        results = pex.run(lambda: contexts[pex.rank].broadcast_local(pex.rank))
         expect = [rank - (rank % local_size) for rank in range(size)]  # type: Any
 
         assert results == expect, "not all threads ran broadcast_local correctly"
 
         # Perform a gather.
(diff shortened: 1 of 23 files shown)


harness/tests/test_ipc.py
```diff
@@ -123,7 +123,7 @@ def health_check() -> None:
                 assert all(g == 2 * msg for g in gathered)
 
 
-def test_zmq_server_client() -> None:
+def testserver_client() -> None:
```
mpkouznetsov, line 126: Did you mean?
```suggestion
def test_server_client() -> None:
```

mpkouznetsov (approved):
Awesome!
Please look at my one question.

harness/tests/test_ipc.py
rb-determined-ai: lol sed got a little overzealous

PR #3898, opened 2022-04-13
chore: add filtering by userIds to API endpoints (DET-7019)

johnkim-det:
## Description

API endpoints for fetching Experiments, Models, and Tasks (Commands, Notebooks, Shells, Tensorboards) should filter by users with an array of userIds.

## Test Plan
- [ ] make sure that filtering Experiments and Models tables by users still works (This should test both API and front-end changes at once).
- [ ] make sure that filtering TaskList by users still works. 
- [ ] filtering by usernames with the "users" field in the request body should still work for all 6 modified endpoints (probably easiest to test this via swagger docs at `<API deployment url>/docs/rest-api/`)
- [ ] filtering by userIds with the "userIds" field in the request body works for Task endpoints (commands, notebooks, shells, tensorboards) (also test via swagger)

## Commentary (optional)

This is part of a larger project to always refer to users by their ids instead of their usernames.

TaskList uses front-end filtering instead of making API calls when filters are updated. My guess is that this is due to performance issues related to fetching four different types of tasks at once. Even if the app is not currently using the users/userIds param for these endpoints, we still want to test that they correctly filter by userId. At some point we may want to look into replacing the front-end filtering with API calls, but that's outside the scope of this PR.

## Checklist


master/internal/api_experiment.go
```diff
@@ -253,6 +253,10 @@ func (a *apiServer) GetExperiments(
 	}
 	stateFilterExpr := strings.Join(allStates, ",")
 	userFilterExpr := strings.Join(req.Users, ",")
+	userIdFilterExpr := strings.Trim(
+		strings.Join(strings.Split(fmt.Sprint(req.UserIds), " "), ","),
```
mapmeld, line 257: I'm a little confused on this, is this to handle `req.UserIds` being a null or empty string?

johnkim-det: the intent is to turn an array of ints `[1 2 3]` into a string that can be used in the sql query `1,2,3`. not sure if there's a cleaner way to write this in go, or i can just add a clarifying comment.

johnkim-det:
@mapmeld trying to follow the error in test-e2e-cpu. assuming that the arguments used for the `get_model` query need to be changed but i don't see where the test would be making that query. any ideas?

master/static/srv/get_models.sql
```diff
@@ -4,8 +4,9 @@ LEFT JOIN users as u ON u.id = m.user_id
 WHERE ($1 = 0 OR m.id = $1)
 AND ($2 = '' OR m.archived = $2::BOOL)
 AND ($3 = '' OR (u.username IN (SELECT unnest(string_to_array($3, ',')))))
-AND ($4 = '' OR (m.labels <@ string_to_array($4, ',')))
-AND ($5 = '' OR m.name ILIKE $5)
-AND ($6 = '' OR m.description ILIKE $6)
+AND ($4 = '' OR m.user_id IN (SELECT unnest(string_to_array($4, ',')::int [])))
+AND ($5 = '' OR (m.labels <@ string_to_array($5, ',')))
+AND ($6 = '' OR m.name ILIKE $6))
```
mapmeld, line 9: I think the issue is just the double parens here: ))

proto/src/determined/api/v1/command.proto
```diff
@@ -32,8 +32,11 @@ message GetCommandsRequest {
   int32 offset = 3;
   // Limit the number of commands. A value of 0 denotes no limit.
   int32 limit = 4;
-  // Limit commands to those that are owned by the specified users.
-  repeated string users = 5;
+  // Limit commands to those that are owned by users with the specified userIds.
+  repeated int32 user_ids = 6;
+  // Limit commands to those that are owned by users with the specified
+  // usernames.
+  repeated string users = 7;
```
mapmeld, line 39: I don't think you need to make a new users field here, you can change the comment without changing the type

johnkim-det: the intent here was to make users use `7` so that its argument would come later in the function defined in `api-ts-sdk/api.ts`, and then `detApi.Commands.getCommands` could be called with the last `users` argument omitted.

master/static/srv/get_models.sql
johnkim-det: looks like that was it, thanks for catching that!

proto/src/determined/api/v1/command.proto
mapmeld: I don't think that this is a good way of handling users argument being optional or less-used (?) -- if the API added a new field in the future then it would negate the benefit of changing users like this

johnkim-det: that's a good point. i wanted to avoid using `undefined` as the value for `users` but i guess if we're keeping both fields we might as well be explicit about the fact that we're ignoring one of them.

master/internal/api_experiment.go
stoksc: we've done `"{"+strings.Join(trialIds, ",")+"}"` elsewhere with an `unnest` on the sql side (see proto_get_trials_plus .. or something). (bun should also make this sort of thing less bad-feeling)

master/internal/api_model.go
```diff
@@ -153,7 +158,7 @@ func (a *apiServer) clearModelName(ctx context.Context, modelName string) error
 
 	getResp := &apiv1.GetModelsResponse{}
 	err := a.m.db.QueryProtof("get_models", []interface{}{"id"},
-		&getResp.Models, 0, "", "", "", modelName, "")
+		&getResp.Models, 0, "", "", "", "", modelName, "")
```
stoksc, line 161: bun should also help this sort of risky business. good catch remembering this.

stoksc (commented):
go side lgtm but I'll hold approving in case @mapmeld has comments I missed

webui/react/src/pages/TaskList.tsx
```diff
@@ -121,6 +121,10 @@ const TaskList: React.FC = () => {
         getJupyterLabs({ signal: canceler.signal }),
         getShells({ signal: canceler.signal }),
         getTensorBoards({ signal: canceler.signal }),
+        // getCommands({ signal: canceler.signal, users: settings.user }),
+        // getJupyterLabs({ signal: canceler.signal, users: settings.user }),
+        // getShells({ signal: canceler.signal, users: settings.user }),
+        // getTensorBoards({ signal: canceler.signal, users: settings.user }),
```
mapmeld, line 127: I would say since these are replaced, you can remove the commented lines from this file

mapmeld (approved):
LGTM, I did make one additional request to remove some commented lines but then it should be ready to go

PR #3901, opened 2022-04-13
chore: update dep requirements for react

hamidzr:
## Checklist

### DIFF
--- a/webui/react/package.json
+++ b/webui/react/package.json
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "engines": {
-    "node": "^14.16.0"
+    "node": ">=14.0.0 <15.0.0",
+    "npm": "<7.0.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


hkang1 (approved):
LGTM!

webui/react/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "engines": {
-    "node": "^14.16.0"
+    "node": ">=14.0.0 <15.0.0",
```
hkang1, line 6: do we need to go so far back to 14.0.0? vs just keeping the lower end 14.16.0?
Also `^14.16.0` implies that it will only support 14.x.x and not anything higher such as 15.x.x

hamidzr: I thought so too but having the caret in front of it it was still complaining when the user had 14.15 (or maybe it was 14.17). Although it was just a warning.

I can update to keep the node engine as is and just add the bit for `npm`

PR #3903, opened 2022-04-13
chore: Draft using bun for GetProject and GetWorkspaceProjects

mapmeld:
## Description

Replaces the responses for GetProject and GetWorkspaceProjects (including filters, sort-by, order-by)

**Edit: added GetProjectExperiments**

Ongoing issues:
- JOIN-ing the experiments table in the view is very slow, we should maybe create a view or function (possible to do WITH?) to get num experiments and num active experiments
- users=a,b,c is not converting to a list query properly, it is only working for an exact match like `users=admin`
- `apiOrderByToSQL` and changes to `internal/api.go` and `internal/db/postgres.go` is in common with checkpoint PR, we will need to coordinate on one PR to make these shared changes
 
## Checklist


master/internal/api/projects/model.go
```diff
@@ -7,12 +7,69 @@ import (
...
+
 	"github.com/determined-ai/determined/master/internal/db"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/master/pkg/protoutils/protoconverter"
+	"github.com/determined-ai/determined/proto/pkg/experimentv1"
 	"github.com/determined-ai/determined/proto/pkg/projectv1"
 )
 
+type ExperimentMetadata struct {
+	bun.BaseModel `bun:"select:experiments_view"`
```
stoksc, line 20: thoughts on view vs just forming the equivalent of the view in bun?

PR #3904, opened 2022-04-13
feat: add det support bundle [DET-5886]

nrajanee:
## Description

feat: add det support bundle [DET=5886]

## Test Plan

Tested in an e2e test. 

Manual test: 

- [ ] Run an experiment locally. 
- [ ] Get a trial id of the trials associated with this experiment. 
- [ ] Run `det trial support-bundle {trial_id} -o {output-dir}`
- [ ] Ensure that {output_dir}.tar.gz is created and the following .json files are in this tar folder: api_experiment_call.json, api_trial_call.json, master_logs.json, trial_logs.json.


e2e_tests/tests/conftest.py
```diff
@@ -110,7 +110,7 @@ def cluster_log_manager(request: SubRequest) -> Iterator[Optional[ClusterLogMana
         # Yield `None` so that pytest handles the no log manager case correctly.
         yield None
 
-    compare_stats()
+    #compare_stats()
```
nrajanee, line 113: @guangqing mentioned that she needs to fix something here. Once she does I will fetch from upstream/master. https://determined-ai.slack.com/archives/CSG3W08JY/p1649877002656269?thread_ts=1649870028.978399&cid=CSG3W08JY

harness/determined/cli/trial.py
```diff
@@ -120,6 +125,71 @@ def trial_logs(args: Namespace) -> None:
...
+@authentication.required
+def output_logs(args: Namespace): 
+    if args.output_dir is None: 
+        raise Exception('Please provide the output directory you would like to direct the logs to.')
+    
+    output_dir_tar = tarfile.open(f'{args.output_dir}.tar.gz', 'w:gz')
+
+    trial_logs_file = write_trial_logs(args, args.trial_id, output_dir_tar)
+    output_dir_tar.add(f'{trial_logs_file}.json')
+    os.remove(f'{trial_logs_file}.json')
```
varlogtim, line 137: It seems strange to me that half of the filename is defined inside a function a step down, and half is defined outside. 
```
stack: output_logs -> write_trial_logs -> create_json_file_in_dir
   "{filename}.json"         ^ filename            ^ "{filename}.json"
```

I feel like the extension should just be added in the `write_*` functions so there isn't an agreement between the other functions that they need to add `.json`.

harness/determined/cli/trial.py
```diff
@@ -120,6 +125,71 @@ def trial_logs(args: Namespace) -> None:
         timestamp_after=args.timestamp_after,
     )
 
+@authentication.required
+def output_logs(args: Namespace): 
+    if args.output_dir is None: 
+        raise Exception('Please provide the output directory you would like to direct the logs to.')
+    
+    output_dir_tar = tarfile.open(f'{args.output_dir}.tar.gz', 'w:gz')
```
varlogtim, line 133: We should, by default, write to PWD and it should be something like:

```
if output_dir is None:
    output_dir = os.getcwd()
filename = "det-bundle-trail<trial_num>-<datetime>.tar.gz"
fullpath = os.path.join(output_dir, filename)
bundle = tarfile.open(fullpath, "w:gz")
...
```

Maybe use Temp Dirs and/or Temp Files so that they get auto-cleaned up, like this:
```
with tempfile.TemporaryDirectory() as temp_dir:
    trial_logs_filepath = write_trial_logs(args, temp_dir)
    bundle.add(trial_logs_filepath, arcname=os.path.basename(trial_logs_filepath))
    ...
```

I didn't test the above but you get the idea.

varlogtim (changes requested):
Overall it looks pretty good. I am glad to see work is being started on this area.

Oh, also, I think we should use `"` instead of `'` ... I think this is specified somewhere but I can't find where.

harness/determined/cli/trial.py
```diff
@@ -120,6 +125,71 @@ def trial_logs(args: Namespace) -> None:
...
+    os.remove(f'{api_trail_file}.json')
+    os.remove(f'{api_experiment_file}.json')
+
+    output_dir_tar.close()
+
+    return f'{args.output_dir}.tar.gz'
+
+    
+
+def write_trial_logs(args,trial_id, output_dir):  #difference between this trial_logs and api call? There doesn't seem to be any.
```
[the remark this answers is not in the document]
varlogtim: Actually, `trial_id` is a property of `args, so, you can just pass in `args` and access the prop in the function,

harness/determined/cli/trial.py
```diff
@@ -120,6 +125,71 @@ def trial_logs(args: Namespace) -> None:
...
+
+    
+
+def write_trial_logs(args,trial_id, output_dir):  #difference between this trial_logs and api call? There doesn't seem to be any. 
+    trial_logs = api.trial_logs(args.master, trial_id)
+    # use trial logs to get the experiment id 
+    file_name = 'trial_logs'
+    trial_logs_list = []
+    for log in trial_logs: 
+        trial_logs_list.append(log)
```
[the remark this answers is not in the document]
varlogtim: I would also like it if there was a way to control the order of the keys when we `json.dumps()`

trial example:
```
trial example: 
{"id": "37", "timestamp": "2022-04-14T02:28:51.946635Z", "message": "[2022-04-14T02:28:51.946635Z] 0bf890e4 || INFO [28]: determined.core: report_training_metrics(latest_batch=10, metrics={'loss': 1.9621983})\n", "level": "LOG_LEVEL_UNSPECIFIED"}

master example:
{'id': 34, 'message': 'container cbdc387b-a42b-42f6-a0ac-3f20261e25d8 terminated  actor-local-addr="agent2" actor-system="master" error="<nil>" go-type="agent"', 'time': '2022-04-13T22:28:59.571143493-04:00', 'level': 'info'}
```

I would like it if it was always, [id, timestamp, level, message], but I am not sure if there is a handy way to do that.

harness/determined/cli/trial.py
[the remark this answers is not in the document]
nrajanee: > Actually, `trial_id` is a property of `args, so, you can just pass in `args` and access the prop in the function,

Yes, sorry I missed that! I literally passed in args.trial_id :D

harness/determined/cli/trial.py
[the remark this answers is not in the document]
nrajanee: Will look into this!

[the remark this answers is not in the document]
varlogtim: Thinking about this more, I wonder if we should just print only the `message` part for trial logs - basically so they look like trial logs.

And for the master logs, we make some kind of printf which prints them in the same fashion that they appear in the master logs as. Basically like:

```
   for log in master_logs:
      f.write(f"{log['time']} [{log['level']}]: {log['message']}")
```

harness/determined/cli/trial.py
nrajanee: Yes this definitely makes things a lot cleaner! Thank you.

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
+
+    bundle = None
+    if args.output_dir is None:
+        output_dir = os.getcwd()
+        tar = f"det-bundle-trail_{args.trial_id}-{datetime.now().strftime('%Y-%m-%d_%H:%M:%S')}.tar.gz"
+        fullpath = os.path.join(output_dir, tar)
+        bundle = tarfile.open(fullpath, "w:gz")
+
+    else:
+        bundle = tarfile.open(f"{args.output_dir}.tar.gz", "w:gz")
```
varlogtim, line 147: We don't want whatever the user specifies as the output directory to be the name of the file. For example, if the user says, `det t support-bundle 1234 -o /tmp`, we don't want it to try and write `/tmp.tar.gz`. We want it to do something like `/tmp/det-bundle-trial-1234-20220415T105733.tar.gz`


```
output_dir = os.getcwd()
if args.output_dir is not None:
    output_dir = args.output_dir

filename_printf = "det-bundle-trial-{trial_id}-{datetime}.tar.gz"
tar_filename = filename_printf.format(trial_id=args.trail_id, datetime=datime.now().strftime('%Y%m%dT%H%M%S'))
fullpath = os.path.join(output_dir, tar_filename)
bundle = tarfile.open(fullpath, "w:gz")
```

... I also changed the format of the time a bit. I don't know if Windows will get angry about `:` in paths, we should probably avoid it.

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
+
+    bundle.add(trial_logs_filepath, arcname=os.path.basename(trial_logs_filepath))
+    bundle.add(master_logs_filepath, arcname=os.path.basename(master_logs_filepath))
+    bundle.add(api_trail_filepath, arcname=os.path.basename(api_trail_filepath))
+    bundle.add(api_experiment_filepath, arcname=os.path.basename(api_experiment_filepath))
+
+    bundle.close()
```
varlogtim, line 154: Maybe print the full path of the file you wrote here.

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
-                args.master, "logs"
-            )
-    file_name = 'master_logs'
-    create_json_file_in_dir(response.json(),file_name, output_dir)
-    return file_name
-
-def write_api_call(args, output_dir): 
-    file_name1 = 'api_experiment_call'
-    file_name2 = 'api_trial_call'
+def output_logs(args: Namespace):
```
varlogtim, line 130: Possibly rename this to like `generate_support_bundle`

varlogtim (changes requested):
Sorry to flip flop on my responses :)

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
+
+
+def write_master_logs(args, temp_dir):
+    response = api.get(args.master, "logs")
+    file_path = os.path.join(temp_dir.name, "master_logs.jsonl")
+    create_json_file_in_dir(response.json(), file_path)
+    return file_path
+
+
+def write_api_call(args, temp_dir):
```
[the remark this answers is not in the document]
nrajanee: I kept it as one function because I'm getting the experiment id from the trial object. If I use a different function I'll have to return it from the trial api call function and send it as a parameter to the exp api call function. I can do that too if you think that's cleaner?

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
+    bundle.add(api_trail_filepath, arcname=os.path.basename(api_trail_filepath))
+    bundle.add(api_experiment_filepath, arcname=os.path.basename(api_experiment_filepath))
+
+    bundle.close()
+
+
+def write_trial_logs(args, temp_dir):
+    trial_logs = api.trial_logs(args.master, args.trial_id)
+    file_path = os.path.join(temp_dir.name, "trial_logs.jsonl")
+    create_jsonl_file_in_dir(trial_logs, file_path)
```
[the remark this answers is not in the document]
nrajanee: Yes, I saw that comment! I pinged you about it on slack. But this looks good to me! Thank you.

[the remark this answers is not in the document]
nrajanee: I'm going to make these .txt files is that okay?

harness/determined/cli/trial.py
```diff
@@ -125,71 +125,72 @@ def trial_logs(args: Namespace) -> None:
...
+    return file_path
+
+
+def write_master_logs(args, temp_dir):
+    response = api.get(args.master, "logs")
+    file_path = os.path.join(temp_dir.name, "master_logs.jsonl")
+    create_json_file_in_dir(response.json(), file_path)
```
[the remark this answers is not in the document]
nrajanee: I looked into this I'm not calling the logs function in harness/determined/cli/master.py. I'm only making the api call: api.get(args.master, "logs") and this will trigger the function getMasterLogs in master/internal/core.go and I don't see any sleeps over there so I don't think we need to handle polling.

harness/determined/cli/trial.py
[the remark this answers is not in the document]
varlogtim: or `.log`

harness/determined/cli/trial.py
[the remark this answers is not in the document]
varlogtim: Ahh, I see. I guess it is fine this way.

harness/determined/cli/trial.py
```diff
@@ -138,33 +138,43 @@ def output_logs(args: Namespace):
...
+    )
+    fullpath = os.path.join(output_dir, tar_filename)
+    bundle = tarfile.open(fullpath, "w:gz")
+
     bundle.add(trial_logs_filepath, arcname=os.path.basename(trial_logs_filepath))
     bundle.add(master_logs_filepath, arcname=os.path.basename(master_logs_filepath))
     bundle.add(api_trail_filepath, arcname=os.path.basename(api_trail_filepath))
     bundle.add(api_experiment_filepath, arcname=os.path.basename(api_experiment_filepath))
 
+    print(f"bundle directory: {bundle.name}")
```
varlogtim, line 157: We are generating a file not a directory. Maybe just `print(f"bundle path: {fullpath}")`

harness/determined/cli/trial.py
```diff
@@ -138,33 +138,43 @@ def output_logs(args: Namespace):
     bundle = None
     if args.output_dir is None:
         output_dir = os.getcwd()
-        formatted_dt = datetime.now().strftime('%Y-%m-%d_%H:%M:%S')
-        tar = f"det-bundle-trail_{args.trial_id}-{formatted_dt}.tar.gz"
-        fullpath = os.path.join(output_dir, tar)
-        bundle = tarfile.open(fullpath, "w:gz")
 
     else:
         bundle = tarfile.open(f"{args.output_dir}.tar.gz", "w:gz")
```
varlogtim, line 143: This should be:

```
output_dir = args.output_dir
if output_dir is None:
    output_dir = os.getcwd()
...
```

varlogtim (changes requested):

e2e_tests/tests/trial/support-bundle.py
```diff
@@ -0,0 +1,24 @@
+import subprocess
```
stoksc, line 1: meta comment, I'd put this in `e2e_tests/tests/cluster/test_support_bundle.py` instead

e2e_tests/tests/trial/trial.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from determined.common.api import authentication, bindings, certs
+from determined.common.api.bindings import determinedexperimentv1State
+from determined.common.experimental import session
+from tests import config as conf
+from tests import experiment as exp
+
+
+def create_exp_get_trial_id():
```
stoksc, line 12: `e2e_tests/tests/experiment/experiment.py` has a lot of similar helpers. maybe put it there for the sake of anyone else being able to find this/use it.

e2e_tests/tests/trial/trial.py
```diff
@@ -0,0 +1,30 @@
...
+    session = test_session()
+    trials = bindings.get_GetExperimentTrials(session, experimentId=exp_id).trials
+
+    assert len(trials) > 0
+    trial_example = trials[0]
+    trial_id = trial_example.id
+    return trial_id
+
+
+def test_session() -> session.Session:
```
stoksc, line 26: this already exists somewhere. maybe just import it?

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,77 @@ def trial_logs(args: Namespace) -> None:
     )
 
 
+@authentication.required
+def generate_support_bundle(args: Namespace):
+    temp_dir = tempfile.TemporaryDirectory()
```
stoksc, line 131: I think we need to clean up this temp directory? maybe use a context manager or try/finally.

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,77 @@ def trial_logs(args: Namespace) -> None:
...
+    master_logs_filepath = write_master_logs(args, temp_dir)
+
+    api_experiment_filepath, api_trail_filepath = write_api_call(args, temp_dir)
+
+    bundle = None
+    if args.output_dir is None:
+        output_dir = os.getcwd()
+
+    else:
+        bundle = tarfile.open(f"{args.output_dir}.tar.gz", "w:gz")
```
stoksc, line 143: same comment about context managers. may be able to use `contextlib.closing` to help.

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,77 @@ def trial_logs(args: Namespace) -> None:
...
+
+    filename_printf = "det-bundle-trial-{trial_id}-{datetime}.tar.gz"
+    tar_filename = filename_printf.format(
```
stoksc, line 146: you can use f-strings. I think we're requiring at least 3.6 now.

harness/determined/cli/trial.py
stoksc, line 143: something isn't quite right here - why are we opening two tar files? did we just mean to set `output_dir = args.output_dir` here instead? `--output-dir` doesn't work right now, if I try to use it I see this because we never set output dir. 
```
(det) ➜  determined git:(b46d5721a) ✗ det t support-bundle --output-dir tmp 4   
Traceback (most recent call last):
  File "/Users/bradleylaney/Code/determined/harness/determined/cli/cli.py", line 247, in main
    parsed_args.func(parsed_args)
  File "/Users/bradleylaney/Code/determined/harness/determined/common/api/authentication.py", line 397, in f
    return func(namespace)
  File "/Users/bradleylaney/Code/determined/harness/determined/cli/trial.py", line 149, in generate_support_bundle
    fullpath = os.path.join(output_dir, tar_filename)
UnboundLocalError: local variable 'output_dir' referenced before assignment
Failed to support bundle
```

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,77 @@ def trial_logs(args: Namespace) -> None:
...
+        for log in trial_logs:
+            f.write(log["message"])
+
+    return file_path
+
+
+def write_master_logs(args, temp_dir):
+    response = api.get(args.master, "logs")
+    file_path = os.path.join(temp_dir.name, "master_logs.txt")
+    with open(file_path, "w") as f:
```
stoksc, line 175: @varlogtim thoughts on full master logs vs. just a slice from `(t.start_time - 5m, t.end_time + 5m)`. I actually kinda like just getting the whole thing. It's max 25k lines.

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,77 @@ def trial_logs(args: Namespace) -> None:
...
+
+def write_api_call(args, temp_dir):
+    file_path1 = os.path.join(temp_dir.name, "api_experiment_call.json")
+    file_path2 = os.path.join(temp_dir.name, "api_trial_call.json")
+
+    trial_obj = bindings.get_GetTrial(setup_session(args), trialId=args.trial_id).trial
+    experiment_id = trial_obj.experimentId
+    exp_obj = bindings.get_GetExperiment(setup_session(args), experimentId=experiment_id)
+
+    trial_obj.to_json(), exp_obj.to_json()
```
stoksc, line 189: I don't think this does anything?

stoksc (commented):
looking pretty good

harness/determined/cli/trial.py
varlogtim: > I think we need to clean up this temp directory? maybe use a context manager or try/finally.

I tested this and it does appear that the temp_dir gets cleaned up, but I agree that a context manager would make this more explicit.

harness/determined/cli/trial.py
varlogtim: @stoksc I like the full master logs.

However, I think that the max-25k-lines is problem. I would like the master->logs interface to support time deltas and a much larger range ... but that is another discussion.

harness/determined/cli/trial.py
varlogtim: Yeah, I put a review somewhere pointing this out but I think it might have gotten lost somewhere...

harness/determined/cli/trial.py
nrajanee: yeah that wasn't supposed to be there, thanks!

harness/determined/cli/trial.py
nrajanee: Oops just realized I missed that requirement! Should I add that to this PR?

harness/determined/cli/trial.py
nrajanee: Yes, I think I didn't commit this change correctly, I fixed it now!

harness/determined/cli/trial.py
nrajanee: I used f-strings somewhere else and lint is complaining.

harness/determined/cli/trial.py
nrajanee: yes, sorry I didn't commit that change correctly.

harness/determined/cli/trial.py
nrajanee: added a with statement!

harness/determined/cli/trial.py
nrajanee: added a with statement

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,69 @@ def trial_logs(args: Namespace) -> None:
...
 
 
+@authentication.required
+def generate_support_bundle(args: Namespace) -> None:
+    output_dir = args.output_dir
+    if output_dir is None:
+        output_dir = os.getcwd()
+
+    elif not os.path.exists(output_dir):
+        os.mkdir(output_dir)
```
stoksc, line 136: I would probably just raise a value error here instead.

harness/determined/cli/trial.py
```diff
@@ -121,6 +126,69 @@ def trial_logs(args: Namespace) -> None:
...
+
+def write_api_call(args: Namespace, temp_dir: str) -> Tuple[str, str]:
+    file_path1 = os.path.join(temp_dir, "api_experiment_call.json")
+    file_path2 = os.path.join(temp_dir, "api_trial_call.json")
+
+    trial_obj = bindings.get_GetTrial(setup_session(args), trialId=args.trial_id).trial
+    experiment_id = trial_obj.experimentId
+    exp_obj = bindings.get_GetExperiment(setup_session(args), experimentId=experiment_id)
+
+    create_json_file_in_dir(exp_obj.to_json(), file_path1)
```
stoksc, line 182: `file_path1` and `file_path2` don't really tell me anything, I have to read what they are and how they are used anyway. which is to say maybe just put them where they're used, or closer to where they're used, or give them better names.
```suggestion
    create_json_file_in_dir(exp_obj.to_json(), os.path.join(temp_dir, "api_experiment_call.json"))
```

varlogtim:
Is it possible to squash all these changes so that I can read them as a single unit?

Maybe I just don't know how to read github correctly to get everything to display together - if there is a way to just read the entire change, please let me know. :)

e2e_tests/tests/conftest.py
```diff
@@ -110,7 +112,7 @@ def cluster_log_manager(request: SubRequest) -> Iterator[Optional[ClusterLogMana
         # Yield `None` so that pytest handles the no log manager case correctly.
         yield None
 
-    compare_stats()
+    # compare_stats()
```
stoksc, line 115: don't forget to remove this comment when merging

stoksc (approved):
lgtm, maybe let @varlogtim take one last look.

varlogtim (changes requested):
Last review, I swear! :)

harness/determined/cli/trial.py
```diff
@@ -121,6 +127,81 @@ def trial_logs(args: Namespace) -> None:
...
+            bundle.add(
+                api_trail_filepath,
+                arcname=os.path.join(tar_filename, os.path.basename(api_trail_filepath)),
+            )
+            bundle.add(
+                api_experiment_filepath,
+                arcname=os.path.join(tar_filename, os.path.basename(api_experiment_filepath)),
+            )
+
+            print(f"bundle path: {fullpath}")
```
varlogtim, line 163: Sorry if I was confusing in my last comment.

Putting the files inside of a "directory" whose name is also the name of the `tar.gz` file it is, causes a problem when extracting. Because it tries to write the files into a path as if the `tar.gz` file was a directory. See:

```
(dev-3.7.11) [ttucker@gir tmp]$ tar zxvf det-bundle-trial-11-20220420T085359.tar.gz
det-bundle-trial-11-20220420T085359.tar.gz/trial_logs.txt
tar: det-bundle-trial-11-20220420T085359.tar.gz/trial_logs.txt: Cannot open: Not a directory
det-bundle-trial-11-20220420T085359.tar.gz/master_logs.txt
tar: det-bundle-trial-11-20220420T085359.tar.gz/master_logs.txt: Cannot open: Not a directory
det-bundle-trial-11-20220420T085359.tar.gz/api_trial_call.json
tar: det-bundle-trial-11-20220420T085359.tar.gz/api_trial_call.json: Cannot open: Not a directory
det-bundle-trial-11-20220420T085359.tar.gz/api_experiment_call.json
tar: det-bundle-trial-11-20220420T085359.tar.gz/api_experiment_call.json: Cannot open: Not a directory
tar: Exiting with failure status due to previous errors
```


This is what I put in my last review:

```
    dt = datetime.now().strftime("%Y%m%dT%H%M%S")
    bundle_name = f"det-bundle-trial-{args.trial_id}-{dt}"
    fullpath = os.path.join(output_dir, f"{bundle_name}.tar.gz")

    with tempfile.TemporaryDirectory() as temp_dir, tarfile.open(fullpath, "w:gz") as bundle:
        trial_logs_filepath = write_trial_logs(args, temp_dir)
        master_logs_filepath = write_master_logs(args, temp_dir)
        api_experiment_filepath, api_trail_filepath = write_api_call(args, temp_dir)

        bundle.add(trial_logs_filepath,
                   arcname=os.path.join(bundle_name, os.path.basename(trial_logs_filepath)))
        bundle.add(master_logs_filepath,
                   arcname=os.path.join(bundle_name, os.path.basename(master_logs_filepath)))
        bundle.add(api_trail_filepath,
                   arcname=os.path.join(bundle_name, os.path.basename(api_trail_filepath)))
        bundle.add(api_experiment_filepath,
                   arcname=os.path.join(bundle_name, os.path.basename(api_experiment_filepath)))

        print(f"bundle path: {fullpath}")
```

We need to make a `bundle_name` which will become the directory that the files get extracted into. The name of the `tar.gz` file should be the `bundle_name` with `tar.gz` at the end.

I usually find it helpful to run the these features from the point of view of different personas, Sales Engineers, Support, Customers, etc. and in various scenarios, single trial, HP search, huge logs, etc. and it helps me to spot these kinds of issues / improvements.

varlogtim (changes requested):
Sorry if my last comment was confusing! :) Just one more change.

harness/determined/cli/trial.py
nrajanee: done!

varlogtim (approved):

PR #3906, opened 2022-04-13
fix: run scheduling on agent connection/enable events, reconnectBacklog replay.

ioga:
## Description

We didn't run scheduler in case when an agent reconnects or gets enabled, and no other action is happening. This'd mean a pending task would stay pending, even though a new capacity has appeared.

- add an auto test for this.
- add a bit of a more structured logging for agent messages in RP actor.
- also remove unused `uuid` field on agent.
- also remove unused `internal/sproto/container.go`, gone as of #3663.

## Test Plan
In a single agent setup,
- Disable the agent.
- Submit a task.
- Enable the agent.
- It should get scheduled.

## Checklist


master/internal/resourcemanagers/resource_pool.go
```diff
@@ -368,29 +369,47 @@ func (rp *ResourcePool) Receive(ctx *actor.Context) error {
 }
 
 func (rp *ResourcePool) receiveAgentMsg(ctx *actor.Context) error {
+	var agentID string
 	switch msg := ctx.Message().(type) {
+	// TODO(ilia): I hope go will have a good way to do this one day.
```
stoksc, line 374: previously I've just added an interface to get `msg.Agent` and used that. but it's really not better.

stoksc (approved):

master/internal/resourcemanagers/resource_pool.go
ioga: yeah, then you need to implement interface + 3 methods? not much of a difference...

stoksc: yeah I agree. I really meant " but it's really not better.".

PR #3908, opened 2022-04-13
fix: RM crashes when setting cmd priority

eecsliu:
## Description
Missed a spot when moving to `Asks` from `Tells`. When actor is not expecting a message, it returns an error and causes a crash. 

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -375,10 +375,7 @@ func (c *command) setPriority(ctx *actor.Context, priority int, forward bool) er
 			c.jobType)
 	}
 
-	curPriority := c.Config.Resources.Priority
-	if curPriority == nil || (curPriority != nil && *curPriority != priority) {
-		c.Config.Resources.Priority = &priority
-	}
+	c.Config.Resources.Priority = &priority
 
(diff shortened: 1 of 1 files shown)


master/internal/command/command.go
```diff
@@ -375,7 +372,11 @@ func (c *command) setPriority(ctx *actor.Context, priority int, forward bool) er
 		return fmt.Errorf("setting priority for job type %s in kubernetes is not supported",
 			c.jobType)
 	}
-	c.Config.Resources.Priority = &priority
+
+	currPriority := c.Config.Resources.Priority
+	if currPriority != nil && *currPriority != priority {
```
hamidzr, line 377: do we ever need to set priority if the currPriority is nil? or do we expect Priority to always have a default value whenever this function called

also nit: `cur` vs `curr` feel free to ignore.

eecsliu: ahh good catch

hamidzr (approved):
the code change on its own looks good to me but I haven’t followed all the user error logs to be able to tell if it solves their specific issue or not.

PR #3909, opened 2022-04-14
fix: parse different time format in compare stats script [DET-7039]

gt2345:
## Description

Parse different time format in stats script

## Test Plan

Run `e2e_test` based on local/cloud instance should all work

### DIFF
--- a/e2e_tests/tests/nightly/compute_stats.py
+++ b/e2e_tests/tests/nightly/compute_stats.py
@@ -15,7 +15,20 @@ REMOVE_KEY = "removing"
 
 
 def iso_date_to_epoch(iso_date: str) -> int:
-    return calendar.timegm(datetime.strptime(iso_date[:-4], DATE_PATTERN).timetuple())
+    iso_format = re.compile(r"^\d{4}-\d{2}-\d{2}\S\d{2}:\d{2}:\d{2}\.\d{3,}-\d{2}:\d{2}$")
+    utc_format = re.compile(r"^\d{4}-\d{2}-\d{2}\S\d{2}:\d{2}:\d{2}\.\d+Z$")
+
+    if iso_format.match(iso_date):
+        # because time may came with different number of digits, only keep the first 3 digits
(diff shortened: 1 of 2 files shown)


e2e_tests/tests/nightly/compute_stats.py
```diff
@@ -15,7 +15,20 @@
 
 
 def iso_date_to_epoch(iso_date: str) -> int:
-    return calendar.timegm(datetime.strptime(iso_date[:-4], DATE_PATTERN).timetuple())
+    iso_format = re.compile(r"^\d{4}-\d{2}-\d{2}\S\d{2}:\d{2}:\d{2}\.\d{3,}[-|\+]\d{2}:\d{2}$")
```
ioga, line 18: Perhaps use `dateutil.parser` instead of doing custom parsing, e.g.

```
from dateutil import parser
...
return int(parser.parse(iso_date).timestamp())
```

e2e_tests/tests/nightly/compute_stats.py
```diff
@@ -34,6 +47,9 @@ def parse_log_for_gpu_stats(log_path: str) -> Tuple[int, str, str]:
             match_date = date_parsing_re.match(line)
             if match_date:
                 ts = iso_date_to_epoch(match_date.group(1))
+                if ts == -2:
```
ioga, line 50: `parser.parse` will throw an exception (`dateutil.parser.ParserError`) - no need for "custom" return code

e2e_tests/tests/nightly/compute_stats.py
```diff
@@ -15,7 +16,7 @@
 
 
 def iso_date_to_epoch(iso_date: str) -> int:
-    return calendar.timegm(datetime.strptime(iso_date[:-4], DATE_PATTERN).timetuple())
```
ioga, line 18: is `DATE_PATTERN` unused now?

ioga (approved):
great

e2e_tests/tests/nightly/compute_stats.py
gt2345: Thank you, it's been fixed

PR #3910, opened 2022-04-14
feat: add core.DownloadMode

rb-determined-ai:
In multi-worker settings, it is important that workers on the same node
share a single copy of a downloaded checkpoint.  On an 8-GPU compute
node, that amounts to an 8x bandwidth savings, and also it prevents
workers from stepping on each other's files.

This PR introduces DownloadMode.LocalWorkersShareDownload and
DownloadMode.NoSharedDownload settings, which apply to both
CheckpointContext.download() and CheckpointContext.restore_path().
This allows users to use .download() with or without this coordination,
in case they have special use cases that don't need coordination.

This PR also introduces unit tests for the CheckpointContext.

### DIFF
--- a/harness/determined/_core/_checkpoint.py
+++ b/harness/determined/_core/_checkpoint.py
@@ -1,4 +1,5 @@
 import contextlib
+import enum
 import logging
 import os
 import pathlib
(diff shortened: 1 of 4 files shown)


harness/determined/_core/_checkpoint.py
```diff
@@ -106,44 +142,70 @@ def store_path(
...
+            raise RuntimeError(
+                "cannot call CheckpointContext.store_path() from non-chief worker "
                 f"(rank={self._dist.rank})"
             )
 
         storage_id = str(uuid.uuid4())
         with self._storage_manager.store_path(storage_id) as path:
             yield path, storage_id
-            resources = storage.StorageManager._list_directory(path)
+            resources = self._storage_manager._list_directory(path)
```
mpkouznetsov, line 160: isn't `_list_directory` a static method? If yes isn't the existing call preferable?

mpkouznetsov: oh, is it for mockability??

rb-determined-ai: it is a static method.  Getting it from the storage_manager lets the storage_manager instance override the list_directory via class inheritance, while `storage.StorageManager._list_directory` is referring to a specific class's implementation.

I would argue that if you don't want to use the instances' _list_directory, I'd argue that it shouldn't be a staticmethod at all, but a helper function in the storage module.

mpkouznetsov: 👍

mpkouznetsov (approved):

PR #3918, opened 2022-04-14
fix: Hide archived experiments unless using --all [DET-7042]

mapmeld:
## Description

As we were working on the Determined Project CLI, I noticed that the experiments CLI claims that the `--all` flag is needed to show archived experiments in `det experiment list`. Yet currently archived experiments are included by default.

Bug originated in #3418 because we started using the new API instead of sending filter=all to the legacy API.

This does not affect the Experiments API -- it is just changing what the CLI calls.

## Test Plan

Run `det experiment list` to see which experiments are visible and completed running.

Pick one and archive it: `det experiment archive [id]`
Try `det experiment list` (archived experiment is now hidden)
Try `det experiment list --all` (archived experiment is included)

If you have access to an instance with other users' experiments, `det -m URL experiment list --all` should show those.

## Checklist

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -501,17 +501,15 @@ def wait(args: Namespace) -> None:
 
 @authentication.required
 def list_experiments(args: Namespace) -> None:
-    users: List[str] = []
+    kwargs = {
+        "limit": args.limit,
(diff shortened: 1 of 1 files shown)


trentwatt (approved):
Good fix.

PR #3921, opened 2022-04-15
chore: bring webui checkpoint types into alignment with CoreAPI [DET-6676]

trentwatt:
## Description

chore: bring webui checkpoint types into alignment with CoreAPI [DET-6676]

Many fields previously required in checkpoints are now optional. This PR brings the WebUI types into alignment and provides a fallback UX on the ModelVersionDetails page for when experiment_id, trial_id, etc. are not available.

## Test Plan

Check out the branch and build devcluster from scratch. May have to `rm -rf ~/.postgres` (proceed with care) in order to get the migrations to run properly.

Create a simple mnist pytorch with a checkpoint, register the checkpoint to a model, and go to model version page. things should be ok.

The real test is that with the new schema, many fields are optional that were previously required. model checkpoints automatically have most of those fields, but one way to simulate their absence is to change the return statement of `decodeCheckpoint` in `decoders.ts` to:

```
return {
    allocationId: data.allocationId,
    metadata: {},
    resources: {},
    state: decodeCheckpointState(Sdk.Determinedcheckpointv1State.UNSPECIFIED),
    taskId: data.taskId,
    totalBatches: 0,
    uuid: data.uuid,
  };
```

as well as to various intermediate states between that and the one in the PR.

## Commentary (optional)

There are two main 'checkpoint' types in the API that are actually completely separate: `CheckpointWorkload`, from `trials.proto` and `Checkpoint` from `checkpoint.proto`. As of this PR, their corresponding types in the WebUI are called `CheckpointWorkloadExtended` and `CoreApiGenericCheckpoint`, respectively. The latter is the one we are concerned with here. 

However, in the existing code there is quite a bit of entanglement between them. Specifically, data coming from `CheckpointWorkload` is augmented and cast as `Checkpoint`. As a result `CheckpointModal.tsx` is erroneously called with props of both types, even though the data only ever comes from `CheckpointWorkload`. 

This PR disentangles them, so that the checkpoint prop type of `CheckpointModal` can be narrowed to just that of `CheckpointWorkload`, and we don't have to special-case based on props. This is especially useful given that the backend work in this branch has the types diverging further.

That said, I have included the adaptations needed for `CheckpointModal` to handle `Checkpoint` as well, in case we want to make it available for future use. 

Now is probably a good time to decide whether to keep those or not. My vote would be no: it's kinda messy and promotes more entanglement.

The UI changes in the `CheckpointModal` to accommodate potential missing data exactly parallel those in `ModelVersionDetails`.

If you would like to test checkpoint modal against actual `Checkpoint` type, you can paste this somewhere into `ModelVersionDetails.tsx` (do with decoder modifications for the full effect)

```
      <CheckpointModal
        checkpoint={modelVersion.checkpoint}
        config={modelVersion.checkpoint.experimentConfig || {} as ExperimentConfig}
        show={true}
        title={'Modal!'}
      />
```

## Checklist

[DET-6676]: https://determinedai.atlassian.net/browse/DET-6676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/CheckpointModal.tsx
+++ b/webui/react/src/components/CheckpointModal.tsx
@@ -9,7 +9,6 @@ import { paths } from 'routes/utils';
 import {
   CheckpointStorageType,
   CheckpointWorkloadExtended,
-  CoreApiGenericCheckpoint,
   ExperimentConfig,
   RunState,
 } from 'types';
@@ -21,7 +20,7 @@ import css from './CheckpointModal.module.scss';
 import Link from './Link';
 
 interface Props {
-  checkpoint: CheckpointWorkloadExtended | CoreApiGenericCheckpoint;
+  checkpoint: CheckpointWorkloadExtended;
   config: ExperimentConfig;
   onHide?: () => void;
   searcherValidation?: number;
@@ -31,7 +30,7 @@ interface Props {
 
 const getStorageLocation = (
   config: ExperimentConfig,
-  checkpoint: CheckpointWorkloadExtended | CoreApiGenericCheckpoint,
+  checkpoint: CheckpointWorkloadExtended,
 ): string => {
   const hostPath = config.checkpointStorage?.hostPath;
   const storagePath = config.checkpointStorage?.storagePath;
@@ -106,16 +105,11 @@ const CheckpointModal: React.FC<Props> = (
     showRegisterCheckpointModal({ checkpointUuid: checkpoint.uuid });
   }, [ checkpoint.uuid, onHide, showRegisterCheckpointModal ]);
 
-  const searcherMetric =
-    props.searcherValidation ??
-    ('searcherMetric' in checkpoint ? checkpoint.searcherMetric : undefined);
+  const searcherMetric = props.searcherValidation;
 
-  const checkpointTime =
-    'endTime' in checkpoint
-      ? checkpoint.endTime
-      : 'reportTime' in checkpoint
-        ? checkpoint.reportTime
-        : undefined;
+  if (!checkpoint.experimentId || !checkpoint.trialId) {
+    return null;
+  }
 
   return (
     <Modal
@@ -127,28 +121,21 @@ const CheckpointModal: React.FC<Props> = (
       width={768}
       onCancel={onHide}>
       <div className={css.base}>
-        {renderRow('Source', checkpoint.experimentId ? (
-          <div className={css.source}>
-            <Link path={paths.experimentDetails(checkpoint.experimentId.toString())}>
-              Experiment {checkpoint.experimentId}
-            </Link>
-            {!!checkpoint.trialId && (
-              <><span className={css.sourceDivider} />
-                <Link path={paths.trialDetails(checkpoint.trialId, checkpoint.experimentId)}>
-                  Trial {checkpoint.trialId}
-                </Link>
-              </>
-            )}
-            {!!checkpoint.totalBatches && (
-              <>
-                <span className={css.sourceDivider} />
-                <span>Batch {checkpoint.totalBatches}</span>
-              </>
-            )}
-          </div>
-        ) : 'taskId' in checkpoint
-          ? (<div className={css.source}><span>Task {checkpoint.taskId}</span></div>)
-          : null)}
+        {renderRow(
+          'Source', (
+            <div className={css.source}>
+              <Link path={paths.experimentDetails(checkpoint.experimentId)}>
+                Experiment {checkpoint.experimentId}
+              </Link>
+              <span className={css.sourceDivider} />
+              <Link path={paths.trialDetails(checkpoint.trialId, checkpoint.experimentId)}>
+                Trial {checkpoint.trialId}
+              </Link>
+              <span className={css.sourceDivider} />
+              <span>Batch {checkpoint.totalBatches}</span>
+            </div>
+          ),
+        )}
         {renderRow('State', <Badge state={state} type={BadgeType.State} />)}
         {checkpoint.uuid && renderRow('UUID', checkpoint.uuid)}
         {renderRow('Location', getStorageLocation(config, checkpoint))}
@@ -158,7 +145,7 @@ const CheckpointModal: React.FC<Props> = (
             <HumanReadableNumber num={searcherMetric} /> {`(${config.searcher.metric})`}
           </>,
         )}
-        {checkpointTime && renderRow('Checkpoint Time', formatDatetime(checkpointTime))}
+        {checkpoint.endTime && renderRow('End Time', formatDatetime(checkpoint.endTime))}
         {renderRow('Total Size', totalSize)}
         {resources.length !== 0 && renderRow(
           'Resources', (
--- a/webui/react/src/components/CheckpointModal.tsx
+++ b/webui/react/src/components/CheckpointModal.tsx
@@ -6,17 +6,21 @@ import HumanReadableNumber from 'components/HumanReadableNumber';
 import useCreateModelModal from 'hooks/useCreateModelModal';
 import useRegisterCheckpointModal from 'hooks/useRegisterCheckpointModal';
 import { paths } from 'routes/utils';
-import { CheckpointDetail, CheckpointStorageType, CheckpointWorkload, CheckpointWorkloadExtended,
-  ExperimentConfig, RunState } from 'types';
+import {
+  CheckpointStorageType,
+  CheckpointWorkloadExtended,
+  ExperimentConfig,
+  RunState,
+} from 'types';
 import { formatDatetime } from 'utils/datetime';
 import { humanReadableBytes } from 'utils/string';
-import { checkpointSize, getBatchNumber } from 'utils/workload';
+import { checkpointSize } from 'utils/workload';
 
 import css from './CheckpointModal.module.scss';
 import Link from './Link';
 
 interface Props {
-  checkpoint: CheckpointWorkloadExtended | CheckpointDetail;
+  checkpoint: CheckpointWorkloadExtended;
   config: ExperimentConfig;
   onHide?: () => void;
   searcherValidation?: number;
@@ -26,7 +30,7 @@ interface Props {
 
 const getStorageLocation = (
   config: ExperimentConfig,
-  checkpoint: CheckpointDetail | CheckpointWorkload,
+  checkpoint: CheckpointWorkloadExtended,
 ): string => {
   const hostPath = config.checkpointStorage?.hostPath;
   const storagePath = config.checkpointStorage?.storagePath;
@@ -101,11 +105,9 @@ const CheckpointModal: React.FC<Props> = (
     showRegisterCheckpointModal({ checkpointUuid: checkpoint.uuid });
   }, [ checkpoint.uuid, onHide, showRegisterCheckpointModal ]);
 
-  const totalBatchesProcessed = getBatchNumber(checkpoint);
+  const totalBatchesProcessed = checkpoint.totalBatches;
 
-  const searcherMetric = props.searcherValidation !== undefined ?
-    props.searcherValidation :
-    ('validationMetric' in checkpoint ? checkpoint.validationMetric : undefined);
+  const searcherMetric = props.searcherValidation;
 
   if (!checkpoint.experimentId || !checkpoint.trialId) {
     return null;
--- a/webui/react/src/components/CheckpointModal.tsx
+++ b/webui/react/src/components/CheckpointModal.tsx
@@ -9,6 +9,7 @@ import { paths } from 'routes/utils';
 import {
   CheckpointStorageType,
   CheckpointWorkloadExtended,
+  CoreApiGenericCheckpoint,
   ExperimentConfig,
   RunState,
 } from 'types';
@@ -20,7 +21,7 @@ import css from './CheckpointModal.module.scss';
 import Link from './Link';
 
 interface Props {
-  checkpoint: CheckpointWorkloadExtended;
+  checkpoint: CheckpointWorkloadExtended | CoreApiGenericCheckpoint;
   config: ExperimentConfig;
   onHide?: () => void;
   searcherValidation?: number;
@@ -30,7 +31,7 @@ interface Props {
 
 const getStorageLocation = (
   config: ExperimentConfig,
-  checkpoint: CheckpointWorkloadExtended,
+  checkpoint: CheckpointWorkloadExtended | CoreApiGenericCheckpoint,
 ): string => {
   const hostPath = config.checkpointStorage?.hostPath;
   const storagePath = config.checkpointStorage?.storagePath;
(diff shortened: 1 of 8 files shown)


webui/react/src/services/decoder.ts
```diff
@@ -444,30 +444,28 @@ const decodeCheckpointWorkload = (data: Sdk.V1CheckpointWorkload): types.Checkpo
...
+    hparams: data.training.hparams,
+    metadata: data.metadata,
+    reportTime: data.reportTime?.toString(),
+    resources: resources,
+    searcherMetric: data.training.searcherMetric,
     state: decodeCheckpointState(data.state || Sdk.Determinedcheckpointv1State.UNSPECIFIED),
-    trialId: data.training.trialId || -1, // TODO maybe it becomes required again
+    taskId: data.taskId,
+    totalBatches:
+      data.metadata['latest_batch'] ?? data.training.trainingMetrics?.batchMetrics?.length ?? 0,
```
trentwatt, line 464: placeholder, pending advice from Maksim

webui/react/src/types.ts
```diff
@@ -406,18 +398,32 @@ export interface Step extends WorkloadGroup, StartEndTimes {
   training: MetricsWorkload;
 }
 
-export interface Metrics {
-  numInputs?: number;
-  validationMetrics?: Record<string, number>;
+export interface Metrics extends Api.V1Metrics {
+  // these two fields are present in the protos
+  // as a struct and list of structs, respectively
+  // here, we are being a bit more precise
```
trentwatt, line 404: is this assumption valid? pending confirmation from @mpkouznetsov

webui/react/src/services/decoder.ts
trentwatt, line 464: placeholder, pending advice from @mpkouznetsov

mpkouznetsov: Yes, this looks good.

webui/react/src/types.ts
mpkouznetsov: Yes.

hamidzr:
I added the userfacing api change to the pr. feel free to update it if it's not accurate

hamidzr (commented):
I looked over the proto and WebUI code changes. I'll build the UI and check that out as well.

question: with checkpoints becoming an artifact that the trial runner can submit could these two types become one or do we still need both?
`checkpointWorkloadExtended | CoreApiGenericCheckpoint;`

proto/src/determined/checkpoint/v1/checkpoint.proto
```diff
@@ -30,45 +22,68 @@ enum State {
...
+  // Aggregate user-generated metrics
+  google.protobuf.Struct avg_metrics = 1;
+  // User-generated metrics for each batch
+  repeated google.protobuf.Struct batch_metrics = 2;
+}
+
+// CheckpointTrainingMetadata is specifically metadata about training.
+message CheckpointTrainingMetadata {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [] }
```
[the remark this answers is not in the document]
mpkouznetsov: do you mean if required array is empty the whole `option` section should be skipped? yes we can remove it if you prefer.

trentwatt:
> I looked over the proto and WebUI code changes. I'll build the UI and check that out as well.
> 
> 
> 
> question: with checkpoints becoming an artifact that the trial runner can submit could these two types become one or do we still need both?
> 
> `checkpointWorkloadExtended | CoreApiGenericCheckpoint;` 

I asked @mpkouznetsov whether there was an intent for the types to become consolidated. He said there is not.

webui/react/src/components/ResponsiveTable.tsx
```diff
@@ -185,11 +185,11 @@ const ResponsiveTable: ResponsiveTable = ({
           }
           scroll={tableScroll}
           tableLayout="auto"
-          onRow={(record) => ({
+          onRow={(record) => areRowsRightClickable ? {
```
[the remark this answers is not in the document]
trentwatt: I didn't mean to include this sorry, this was just to have some peace in the console. It has been fixed in master.

webui/react/src/types.ts
```diff
@@ -406,18 +398,32 @@ export interface Step extends WorkloadGroup, StartEndTimes {
...
+  avgMetrics: Record<string, number>;
+  batchMetrics?: Array<Record<string, number>>;
 }
 
 export type Metadata = Record<RecordKey, string>;
 
-export interface CheckpointDetail extends Checkpoint {
-  batch: number;
+export interface CoreApiGenericCheckpoint {
```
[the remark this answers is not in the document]
trentwatt: The purpose of the new name is to make the distinction between the types clear and to avoid the kind of entanglement from before:
- half the places checkpointDetail was being used before, it was not the right type for the data
- the new name makes clear the origin of the data

webui/react/src/services/decoder.ts
trentwatt: You mentioned over slack that it should be metadata['latest_batch'] else null to align with the CLI, so I will take out the fallback to training_batches.length.

webui/react/src/types.ts
trentwatt: my default would be to wait until it appears in another structure, but i am not opposed.

webui/react/src/types.ts
```diff
@@ -406,18 +398,32 @@ export interface Step extends WorkloadGroup, StartEndTimes {
...
+  allocationId: string;
+  experimentConfig?: ExperimentConfig;
   experimentId?: number;
-  metadata?: Metadata;
-  metrics?: Metrics;
+  hparams?: TrialHyperparameters;
+  metadata: Metadata;
+  reportTime?: string;
```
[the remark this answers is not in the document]
trentwatt: @mpkouznetsov it looks like report_time is optional in `checkpoint.proto`. Does that sound right?

webui/react/src/types.ts
```diff
@@ -406,18 +398,32 @@ export interface Step extends WorkloadGroup, StartEndTimes {
...
+  resources: Record<string, number>;
```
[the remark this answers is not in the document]
trentwatt: definitely worth looking into. can we hold off for now though?

webui/react/src/types.ts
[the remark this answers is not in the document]
mpkouznetsov: Not sure. In the old format `end_time` was already optional for some reason.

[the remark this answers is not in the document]
hamidzr: ah that being a carryover from the previous end_time would make sense. probably to report checkpoints in active state. 

```
  // The checkpoint is in an active state.
  STATE_ACTIVE = 1;

```

proto/src/determined/checkpoint/v1/checkpoint.proto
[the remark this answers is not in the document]
hamidzr: ah that doesn't matter much IMO. I was just checking if can make guarantees on any of those fields being present then we should otherwise the generated code treats all of them as nullable. eg experiment_id or trial_id

webui/react/src/components/CheckpointModal.tsx
```diff
@@ -121,21 +127,28 @@ const CheckpointModal: React.FC<Props> = (
...
+              Experiment {checkpoint.experimentId}
+            </Link>
+            {!!checkpoint.trialId && (
+              <><span className={css.sourceDivider} />
+                <Link path={paths.trialDetails(checkpoint.trialId, checkpoint.experimentId)}>
+                  Trial {checkpoint.trialId}
+                </Link>
+              </>
+            )}
+            {!!checkpoint.totalBatches && (
```
hamidzr, line 142: just checking that we don't expect a totalBatches of `0` to render this row. it probably wouldn't make sense for it anyway

webui/react/src/types.ts
```diff
@@ -406,18 +398,33 @@ export interface Step extends WorkloadGroup, StartEndTimes {
   training: MetricsWorkload;
 }
 
-export interface Metrics {
-  numInputs?: number;
-  validationMetrics?: Record<string, number>;
+type MetricStruct = Record<string, number>;
+export interface Metrics extends Api.V1Metrics {
```
hamidzr, line 402: IMO we don't get much out of the extension right now and it only adds the risk of introducing unwanted, and unpopulated, fields to the object in the future. do we use it for something?

hamidzr (approved):
LGTM. great work and nice to see the detailed pr description.


> Now is probably a good time to decide whether to keep those or not. My vote would be no: it's kinda messy and promotes more entanglement.

I'd say if you see the opportunity for it go ahead and drop the support for the unused type. IIRC that was a remnant from the old APIs

I ran into the migration issue, as mentioned, not a related issue to this pr. I cleared my db to get over it. I'd think this get fixed by an updated migration number for releasing into master.
```
ERRO[2022-04-18T12:35:37-07:00]                                               actor-local-addr=3 actor-system=master error="ERROR: relation \"checkpoints_view\" does not exist (SQLSTATE 42P01)" experiment-id=3 go-type=experiment job-id=ec0b7a29-9058-4ba0-bf83-e99be52270ab
```

webui/react/src/components/CheckpointModal.tsx
trentwatt: No, I am treating `0` as a null value for `totalBatches`. Is there a preferred way to indicate this? Explicit `>`?

webui/react/src/types.ts
trentwatt: In `V1Metrics`, the types for `avgMetrics` and `batchMetrics` are `any` and `Array<any>`, respectively corresponding to their proto schemas of `Struct` and `repeated Struct`.

@mpkouznetsov said that they will always be of the form `Record<string, number>` and `Array<Record<string, number>>`. This extension makes that explicit. The attributes in the extension are present in the original interface, but here we are refining them.

Maybe it is bad form though to encode assumptions not present in the schema itself though? I can instead parse them according to the expected format and leave them null otherwise (even though `validationMetrics` is a required field for `Metrics`, `Metrics` itself is only ever present as an optional field).

webui/react/src/types.ts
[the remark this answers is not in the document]
trentwatt: It seems like a `report_time` would always exist by definition though.

webui/react/src/types.ts
trentwatt: ^^went ahead and did this, i.e. validate, parse and narrow.

trentwatt:
> I'd say if you see the opportunity for it go ahead and drop the support for the unused type. IIRC that was a remnant from the old APIs
Went ahead and dropped it.

webui/react/src/types.ts
hamidzr: > This extension makes that explicit. 

maybe I'm reading this incorrectly I'm not sure I see how the `extends Api.V1Metrics` at the end make it more explicit compared to having the type as you want it with Record<string, number> ... and then using the decoder like you already do.

> Maybe it is bad form though to encode assumptions not present in the schema itself though?

we already do someof this and it's a tradeoff. we eventually have to make some of these assumptions to be able to use the data. what we used to do was use iotypes package to enforce the assumptions clientside. now with the new apis we don't need as much of that but when we do we do the check in the decoder module (IIRC)

webui/react/src/services/decoder.ts
```diff
@@ -444,6 +444,26 @@ const decodeCheckpointWorkload = (data: Sdk.V1CheckpointWorkload): types.Checkpo
   };
 };
 
+/* using any here because this comes from the api as any /*
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+export const decodeMetricStruct = (data: any): Record<string, number> => {
+  const metrics: Record<string, number> = {};
+  Object.entries(data || {}).forEach(([ metric, value ]) => {
+    if (typeof metric === 'string' && (typeof value === 'number' || typeof value === 'string')) {
+      const numberValue = (typeof value === 'number') ? value : parseFloat(value);
```
hamidzr, line 453: re any: 
nit: I think this could be more robust if have the function take `unknown`and check whether it's an object and go from there. that should be safer because it'll force the implementation to handle other types as well.

```
nesh> Object.entries('aj223' || {})
[
  [ '0', 'a' ],
  [ '1', 'j' ],
  [ '2', '2' ],
  [ '3', '2' ],
  [ '4', '3' ]
]
```

it might be safer to just ignore value === 'string' if we don't expect them to come through as string since parseFloat accepts this 
```
nesh> parseFloat('1.2h34asdf')
1.2
```

webui/react/src/types.ts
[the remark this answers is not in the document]
hamidzr: you're right yeah 🤔
would this change every time the checkpoint is "reported"? or would it be empty until the checkpoint is in a terminal state like endtime used to behave.

webui/react/src/types.ts
hamidzr: > what we used to do was use iotypes package to enforce the assumptions clientside. now with the new apis we don't need as much of that but when we do we do the check in the decoder module (IIRC)

I saw your change wrt to this and those make sense to me👍🏼

webui/react/src/services/decoder.ts
trentwatt: re: re: any:
i'm happy to do this, but I guess I'm a little confused about the implications:
- the values supplied to this function correspond to either `Metrics.avg_metrics` of items of `Metrics.batch_metrics` as defined in `checkpoint.proto`
- so, based on the protos, `data` will always be a `Struct`/`Record`/_Real Object_™ here.
- so, my assumption would be if you trust the protos, the check is redundant (and potentially confusing for those reading it later)
- and if you don't trust the protos, what can you trust??

In order to make it clear that the assumptions of this function are only valid for the values being supplied here, I would propose putting `decodeMetricStruct` inside of `decodeMetrics`.

trentwatt:
Merging this for now, let's continue discussion though.

webui/react/src/services/decoder.ts
trentwatt: I guess maybe one concern would be that the protos could change and this code wouldn't know it.

I thought about doing something like this:
```
  const decodeMetricStruct = (data: Sdk.V1Metrics['avgMetrics']): Record<string, number> => {
```
but then if the protos change to a allow a string for that type, it's going to stop your from doing `Object.entries` like you were saying.

webui/react/src/types.ts
trentwatt: > how the `extends Api.V1Metrics` at the end makes it more explicit 
yeah, i don't think I articulated it very well. What I meant was that:
- having a type `OurMetrics` that extends `V1Metrics` while narrowing its attributes, AND
- asserting that data that is officially of type `V1Metrics` is actually of type `OurMetrics`
makes the assumption about `V1Metrics` explicit. anyway I guess it's a moot point now.

PR #3922, opened 2022-04-15
fix: crash on upgrade to InteractiveTable [DET-7036]

trentwatt:
## Description

fix: resolve crash on upgrade to InteractiveTable [DET-7036]

## Test Plan
Basic:
`make live`
`git checkout 0.17.12`

add Trials to columns in column picker modal

`git checkout master`
behold the beauty of a web page crashing

`git checkout table-upgrade-crash-fix`
the page should not crash

explanation: `trials` column name changed to `numTrials` in `columns.settings`, app did a lookup on previous name derived from url/settings, came up blank and didn't even know it. 

this fix includes a check on the property, as well as a number of additional edge case safeguards.

try stress testing by (among others):
http://localhost:3000/det/experiments?columns=nopeNotAColumn&columns=nopeNeitherIsThis
http://localhost:3000/det/experiments?columns=nopeNotAColumn&columns=nopeNeitherIsThis&columns=name
http://localhost:3000/det/experiments?columns=id&columns=name&columns=description&columnWidth=
http://localhost:3000/det/experiments?columns=1&columnWidth=

it should generally try to salvage whatever makes sense of the query, and just reset things otherwise

## Checklist

[DET-7036]: https://determinedai.atlassian.net/browse/DET-7036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -319,7 +319,7 @@ const InteractiveTable: InteractiveTable = ({
         ?.getPropertyValue('--theme-sizes-layout-big').slice(0, -2),
     );
     if (typeof pagePadding !== 'number') pagePadding = 16;
-    return columnsWidths.reduce((a, b) => a + b) + 2 * WIDGET_COLUMN_WIDTH + pagePadding;
+    return columnsWidths.reduce((a, b) => a + b, 0) + 2 * WIDGET_COLUMN_WIDTH + pagePadding;
   }, []);
 
   const getUpscaledWidths = useCallback(
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -480,19 +480,20 @@ const InteractiveTable: InteractiveTable = ({
   const renderColumns: ColumnsType<RecordType> = useMemo(
     () =>
       [
-        ...settings.columns.map((columnName, index) => {
-          const column = columnDefs[columnName];
-          const columnWidth = widthData.widths[index];
-          const sortOrder =
+        ...settings.columns.filter(columnName => columnDefs[columnName])
+          .map((columnName, index) => {
+            const column = columnDefs[columnName];
+            const columnWidth = widthData.widths[index];
+            const sortOrder =
             column.key === settings.sortKey ? (settings.sortDesc ? 'descend' : 'ascend') : null;
 
-          return {
-            onHeaderCell: onHeaderCell(index, column),
-            sortOrder,
-            width: columnWidth,
-            ...column,
-          };
-        }),
+            return {
+              onHeaderCell: onHeaderCell(index, column),
+              sortOrder,
+              width: columnWidth,
+              ...column,
+            };
+          }),
         { ...columnDefs.action, width: WIDGET_COLUMN_WIDTH },
       ] as ColumnsType<RecordType>,
 
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -298,8 +298,8 @@ const InteractiveTable: InteractiveTable = ({
   const { width: pageWidth } = useResize(containerRef);
   const tableRef = useRef<HTMLDivElement>(null);
   const [ widthData, setWidthData ] = useState({
-    dropLeftStyles: settings?.columnWidths.map(() => ({})),
-    dropRightStyles: settings?.columnWidths.map(() => ({})),
+    dropLeftStyles: settings?.columnWidths.map(() => ({})) ?? [],
+    dropRightStyles: settings?.columnWidths.map(() => ({})) ?? [],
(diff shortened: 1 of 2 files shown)


hkang1 (commented):
I took a look at what might be the potential reason as to why it's crashing and I think I have it narrowed down, and it might some extra checks.

Here's a screenshot of what is in my "old" local storage where it's causing the crash:
Screen Shot 2022-04-15 at 8 11 44 AM

Here's one of a working one:
Screen Shot 2022-04-15 at 8 19 14 AM

So the difference is that `columnWidths` exists on the working one. So on on older version, it's going to be missing because it's something new that was added. What that basically means is that the code needs to be able to handle the case when `settings.columnWidths` is `undefined`. Not sure if is should fallback to an empty (if that will work) or something else...

hkang1:
> I took a look at what might be the potential reason as to why it's crashing and I think I have it narrowed down, and it might some extra checks.
> 
> Here's a screenshot of what is in my "old" local storage where it's causing the crash: Screen Shot 2022-04-15 at 8 11 44 AM
> 
> Here's one of a working one: Screen Shot 2022-04-15 at 8 19 14 AM
> 
> So the difference is that `columnWidths` exists on the working one. So on on older version, it's going to be missing because it's something new that was added. What that basically means is that the code needs to be able to handle the case when `settings.columnWidths` is `undefined`. Not sure if is should fallback to an empty (if that will work) or something else...

nm, I further tested by deleting some of these local storage values (`columnWidths` included) and it seems to handle it nicely!

webui/react/src/components/InteractiveTable.tsx
```diff
@@ -298,8 +298,8 @@ const InteractiveTable: InteractiveTable = ({
   const { width: pageWidth } = useResize(containerRef);
   const tableRef = useRef<HTMLDivElement>(null);
   const [ widthData, setWidthData ] = useState({
-    dropLeftStyles: settings?.columnWidths.map(() => ({})),
-    dropRightStyles: settings?.columnWidths.map(() => ({})),
+    dropLeftStyles: settings?.columnWidths.map(() => ({})) ?? [],
+    dropRightStyles: settings?.columnWidths.map(() => ({})) ?? [],
```
hkang1, line 302: 👍  was this where it was breaking? where columnWidths didn't map to anything causing these styles to be undefined instead of an array?

webui/react/src/pages/ExperimentList.tsx
```diff
@@ -489,6 +489,29 @@ const ExperimentList: React.FC = () => {
...
+      updateSettings({ columns: actualColumns });
+    }
+    if (settings.columnWidths.length !== actualColumns.length) {
+      updateSettings({
+        columnWidths:
+        actualColumns.map((columnName) => DEFAULT_COLUMN_WIDTHS[columnName]),
+      });
+    }
+  }, [ settings.columns, settings.columnWidths, columns, resetSettings, updateSettings ]);
+
```
hkang1, line 514: **suggestion:** slight update to reduce the number of consecutive `updateSettings` call:
```
  useEffect(() => {
    // This is the failsafe for when column settings get into a bad shape.
    if (!settings.columns?.length || !settings.columnWidths?.length) {
      updateSettings({
        columns: DEFAULT_COLUMNS,
        columnWidths: DEFAULT_COLUMNS.map((columnName) => DEFAULT_COLUMN_WIDTHS[columnName]),
      });
    } else {
      const columnNames = columns.map(column => column.dataIndex as ExperimentColumnName);
      const actualColumns = settings.columns.filter(name => columnNames.includes(name));
      const newSettings: Partial<ExperimentListSettings> = {};
      if (actualColumns.length < settings.columns.length) {
        newSettings.columns = actualColumns;
      }
      if (settings.columnWidths.length !== actualColumns.length) {
        newSettings.columnWidths = actualColumns.map(name => DEFAULT_COLUMN_WIDTHS[name]);
      }
      if (Object.keys(newSettings).length !== 0) updateSettings(newSettings);
    }
  }, [ settings.columns, settings.columnWidths, columns, resetSettings, updateSettings ]);
```

hkang1 (approved):
This looks good, especially with the failsafe. I had one suggestion but the rest looks good!

webui/react/src/pages/ExperimentList.tsx
trentwatt: incorporated!

webui/react/src/components/InteractiveTable.tsx
trentwatt: initially it was breaking [here](https://github.com/determined-ai/determined/blob/575cc21f7fb55fdfb0b80f65babb6ddde85106d9/webui/react/src/components/InteractiveTable.tsx#L487) but after an initial fix it was broken here as well. added some defensive coding throughout.

PR #3930, opened 2022-04-15
chore: wait for process substition log filters [DET-6712]

stoksc:
## Description
To prevent losing logs from very short commands in determined on kubernetes, all task container entrypoints are updated to wait on stdout and stderr process substitution filters before exiting. Additionally, the recommended fluentbit version is changed to 1.9.3, due to enhancements in SIGTERM handling. As a side effect from relying on so many shell features to fix this, we can no longer `exec` the container main process.

## Test Plan
- [x] test this correctly lets something like `det cmd run ls` report all its logs.
- [x] test that SIG_ still makes it to containers OK
- [x] test that commands/shells/tb/notebooks/etc still all work OK

## Checklist

### DIFF
--- a/master/pkg/tasks/const.go
+++ b/master/pkg/tasks/const.go
@@ -7,6 +7,12 @@ const (
 	taskLoggingSetupScript = "task-logging-setup.sh"
 	taskLoggingSetupMode   = 0744
 
+	taskLoggingTeardownScript = "task-logging-teardown.sh"
+	taskLoggingTeardownMode   = 0744
+
+	taskSignalHandlingScript = "task-signal-handling.sh"
+	taskSignalHandlingMode   = 0744
+
 	// Put as many ssh-related files in /run/determined as possible. In particular, it is very
 	// important that we don't overwrite the user's host $HOME/.ssh/id_rsa, if the user happens to
 	// mount their host $HOME into the container's $HOME. Since we control the invocation of sshd,
--- a/master/static/srv/notebook-entrypoint.sh
+++ b/master/static/srv/notebook-entrypoint.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
+source /run/determined/task-signal-handling.sh
 source /run/determined/task-logging-setup.sh
+trap 'source /run/determined/task-logging-teardown.sh' EXIT
 
 set -e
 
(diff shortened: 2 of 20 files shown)


master/static/srv/task-logging-teardown.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
+    # Replace overriden stdout and stderr with original and close them, since the command is finished.
+    exec >&1- >&2- 1>&3 2>&4
+
+    for (( i=0;i<$LOG_WAIT_COUNT;i++ )); do
+        # read returns 1 on EOF, but it's a fifo so that is OK.
+        read <$LOG_WAIT_FIFO || true
+    done
+fi
```
stoksc, line 11:
```suggestion
fi

```

stoksc: ?? come on GitHub

master/static/srv/notebook-entrypoint.sh
```diff
@@ -41,7 +42,7 @@ test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
 "$DET_PYTHON_EXECUTABLE" /run/determined/jupyter/check_idle.py &
 
 READINESS_REGEX='^.*Jupyter Server .* is running.*$'
-exec jupyter lab --ServerApp.port=${NOTEBOOK_PORT} \
+jupyter lab --ServerApp.port=${NOTEBOOK_PORT} \
```
dzhu, line 45: Move the following lines left too, I guess. I mean, actually we should just start using `shfmt`.

master/pkg/tasks/const.go
```diff
@@ -7,6 +7,9 @@ const (
 	taskLoggingSetupScript = "task-logging-setup.sh"
 	taskLoggingSetupMode   = 0744
 
+	taskLoggingTeardownScript = "task-logging-teardown.sh"
+	taskLoggingTeardownMode   = 0755
```
dzhu, line 11: Doesn't need to be executable, I believe (and this and the setup script can even be 644 instead of 744).

master/static/srv/task-logging-setup.sh
```diff
@@ -27,4 +27,14 @@ if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
...
     mkdir -p -m 755 $STDERR_ROTATE_DIR
     exec > >(multilog n2 "$STDOUT_ROTATE_DIR")  2> >(multilog n2 "$STDERR_ROTATE_DIR")
+
+    # Create a fifo to monitor process substitution exits, and a count to know how many to wait on.
+    LOG_WAIT_FIFO=/run/determined/train/logs/wait.fifo
+    LOG_WAIT_COUNT=0
+    mkfifo $LOG_WAIT_FIFO
+
+    exec 3>&1 1> >(multilog n2 "$STDOUT_ROTATE_DIR"; : >$LOG_WAIT_FIFO) \
+         4>&2 2> >(multilog n2 "$STDERR_ROTATE_DIR"; : >$LOG_WAIT_FIFO)
```
dzhu, line 37: Uh, could you explain the extra redirections with 3 and 4? Also, do we need _both_ this exec and the one above that was here previously?

dzhu (approved):

master/static/srv/task-logging-setup.sh
```diff
@@ -27,4 +27,14 @@ if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
     mkdir -p -m 755 $STDOUT_ROTATE_DIR
     mkdir -p -m 755 $STDERR_ROTATE_DIR
     exec > >(multilog n2 "$STDOUT_ROTATE_DIR")  2> >(multilog n2 "$STDERR_ROTATE_DIR")
```
stoksc, line 29: fix this brad

stoksc:
landing through some known failures on master, soon

PR #3936, opened 2022-04-19
test: user experiment delete

hamidzr:
## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -99,7 +99,7 @@ func (a *apiServer) GetExperiment(
 		JobSummary: &jobv1.JobSummary{},
 	}
 
-	if model.TerminalStates[model.StateFromProto(exp.State)] {
+	if model.ActiveState != model.StateFromProto(exp.State) {
 		return &resp, nil
 	}
 
@@ -114,7 +114,8 @@ func (a *apiServer) GetExperiment(
 		// easy deducible how long that would block. So the best we can really do is return without
 		// an error if we're in this case and log. This is a debug log because of how often the
 		// happens when polling for an experiment to end.
-		if !strings.Contains(err.Error(), actorDidNotRespond) {
+		if !strings.Contains(err.Error(), actorDidNotRespond) || !strings.Contains(err.Error(), "could not be found") {
+			// REMOVEME: the actor could not be found issue seems to be allowing exp delete test to pass
 			return nil, err
 		}
 		logrus.WithError(err).Debugf("asking for job summary")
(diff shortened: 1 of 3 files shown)


master/internal/api_experiment.go
```diff
@@ -99,7 +99,7 @@ func (a *apiServer) GetExperiment(
 		JobSummary: &jobv1.JobSummary{},
 	}
 
-	if model.TerminalStates[model.StateFromProto(exp.State)] {
+	if model.ActiveState != model.StateFromProto(exp.State) {
```
hamidzr, line 102: maybe I should've taken out the part to fetch jobsummary entirely to demo. my expectation is that that would result in the same behavior.

master/internal/api_experiment.go
```diff
@@ -114,7 +114,9 @@ func (a *apiServer) GetExperiment(
 		// easy deducible how long that would block. So the best we can really do is return without
 		// an error if we're in this case and log. This is a debug log because of how often the
 		// happens when polling for an experiment to end.
-		if !strings.Contains(err.Error(), actorDidNotRespond) {
+		if !strings.Contains(err.Error(), actorDidNotRespond) ||
```
stoksc, line 117: either consts for both, or remove the const I added, for consistency? either is OK with me.

master/internal/api_experiment.go
```diff
@@ -114,7 +114,9 @@ func (a *apiServer) GetExperiment(
...
+			!strings.Contains(err.Error(), "could not be found") {
+			// REMOVEME: the actor could not be found issue seems to be allowing exp delete test to pass
```
stoksc, line 119: It's not impossible that this could be the test legitimately seeing a race condition. E.G.
* we query DB state, it is active
* we get de-scheduled by the go run time
* experiment finishes it inbox, runs poststop https://github.com/determined-ai/determined/blob/master/master/pkg/actor/ref.go#L330-L337
* and sends it parent that it stopped https://github.com/determined-ai/determined/blob/master/master/pkg/actor/ref.go#L344
* then the parent ('experiments' actor) wakes up and deletes the experiment from the system https://github.com/determined-ai/determined/blob/master/master/pkg/actor/ref.go#L245
* then we wake back up and get "could not be found" rather than "did not respond" (which happens if the experiment gets the message but it was already closing).

so basically.. keep this.

master/internal/api_experiment.go
stoksc: this condition makes more sense to me.

stoksc (approved):
lgtm

hamidzr:
https://app.circleci.com/pipelines/github/determined-ai/determined/21105/workflows/416f85cd-6d08-4295-8341-3fe4f209c836/jobs/663437 shouldn't happen 🤔

hamidzr:
the test in question passed with the inclusion of #3943

hamidzr:
closing at it has surved its purpose. fixes are in through two other prs

PR #3942, opened 2022-04-19
chore: Remove legacy API trials endpoint [DET-6684]

mapmeld:
## Description

Following other legacy-API-deprecating PRs, this removes the old `core_trial.go` API, which is replaced by /api/v1/trials and other endpoints on experiments

@stoksc  - let me know if this is duplicating other work. 

## Test Plan

Automated tests should all pass because previous testing already migrated to the REST API

## Checklist


master/internal/core.go
```diff
@@ -908,10 +908,6 @@ func (m *Master) Run(ctx context.Context) error {
 	searcherGroup := m.echo.Group("/searcher", authFuncs...)
 	searcherGroup.POST("/preview", api.Route(m.getSearcherPreview))
 
-	trialsGroup := m.echo.Group("/trials", authFuncs...)
-	trialsGroup.GET("/:trial_id", api.Route(m.getTrial))
-	trialsGroup.GET("/:trial_id/metrics", api.Route(m.getTrialMetrics))
```
stoksc, line 913: I did end up finding out `get t describe` uses this still.

mapmeld: thanks. I'll see if the tests flag this, then see if the Trials CLI can be changed (will look in +1 week or so?)

stoksc: yeah. updating the trial CLI would be nice, anyway. though it is work.

gh-determined-ai (approved):

master/internal/core.go
mapmeld: I've replaced the e2e_tests use of legacy trials API, so I think this is about to pass tests 🤞

e2e_tests/tests/experiment/experiment.py
```diff
@@ -258,12 +258,37 @@ def is_terminal_state(state: determinedexperimentv1State) -> bool:
     )
 
 
-def trial_metrics(trial_id: int) -> Dict[str, Any]:
-    certs.cli_cert = certs.default_load(conf.make_master_url())
-    authentication.cli_auth = authentication.Authentication(conf.make_master_url(), try_reauth=True)
-    r = api.get(conf.make_master_url(), "trials/{}/metrics".format(trial_id))
-    json = r.json()  # type: Dict[str, Any]
-    return json
+class Step:
```
stoksc, line 261: maybe make this match the new proto 'metrics', https://github.com/determined-ai/determined/blob/4fb1886b8004d7bdf4d4faad7e550bbee36d06df/proto/src/determined/checkpoint/v1/checkpoint.proto#L25-L34, including optional `batch_metrics`. maybe
`def trial_metrics(trial_id: int) -> Iterable[Tuple[int, Metrics]]:`

e2e_tests/tests/experiment/test_core.py
```diff
@@ -52,17 +52,14 @@ def test_metric_gathering() -> None:
     # Check training metrics.
     full_trial_metrics = exp.trial_metrics(trials[0].trial.id)
     batches_trained = 0
-    for step in full_trial_metrics["steps"]:
-        metrics = step["metrics"]
-
-        actual = metrics["batch_metrics"]
-        assert len(actual) == scheduling_unit
+    for step in full_trial_metrics:
+        assert step.batches == batches_trained + scheduling_unit
```
stoksc, line 56: this change doesn't make sense to me. the new check is very different, and doesn't test the same as the old.

e2e_tests/tests/nightly/test_convergence.py
```diff
@@ -12,19 +12,17 @@ def test_mnist_pytorch_accuracy() -> None:
...
-    trial_metrics = exp.trial_metrics(trials[0].trial.id)
+    trial_metrics = exp.trial_validation_metrics(trials[0].trial.id)
 
-    validation_accuracies = [
-        step["validation"]["metrics"]["validation_metrics"]["accuracy"]
-        for step in trial_metrics["steps"]
-        if step.get("validation")
-    ]
+    validation_accuracies = list(
+        map(lambda step: float(step.avg_metrics["accuracy"]), trial_metrics)
```
stoksc, line 18: what's our python code standard/pref between comprehensions and map/filter

e2e_tests/tests/experiment/experiment.py
stoksc: definitely can't be step, since steps (in this meaning of the word) have been intentionally removed from the system

PR #3944, opened 2022-04-19
feat: Add numProjects field to workspaces API, CLI [DET-7064]

mapmeld:
## Description

Adds numProjects / num_projects count associated with each workspace. Affects API and CLI.

## Test Plan

- The /api/v1/workspaces response should include numProjects in each record (with workspace 1 having numProjects: 1)
- The /api/v1/workspaces/1 (default workspace) should include numProjects: 1
- In the CLI, `det w list` should list workspaces table with "# Projects" column, default workspace has "1"
- `det w describe Uncategorized` should print table with numProjects = 1
- `det w create test` should create a workspace and print out table numProjects = 0
- `det w describe test`, `det w edit test test`, etc. should return table with numProjects = 0

## Checklist

### DIFF
--- a/harness/determined/cli/workspace.py
+++ b/harness/determined/cli/workspace.py
@@ -11,15 +11,19 @@ from determined.common.experimental import session
 from . import render
 
 PROJECT_HEADERS = ["ID", "Name", "Description", "# Experiments", "# Active Experiments"]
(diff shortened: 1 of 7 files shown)


EmilyBonar (approved):
LGTM!

PR #3945, opened 2022-04-19
fix: replace carriage returns with newlines in task output [DET-5302]

dzhu:
## Description

A task may output carriage return characters (\r) do something mildly
fancy with the terminal like update a progress bar in place on one
line. Python's tqdm library is a common way to do this. That works
poorly with our logging, since Fluent Bit interprets everything as one
line, causing it to mash everything together and buffer the output for
way too long. Since we're not going to do anything like interpreting the
carriage returns in our log displays, here we simply replace them all
with newlines to get a reasonable effect in those cases.

An alternative might be trying to manipulate tqdm itself, as the
original ticket proposes. But that comes with the issue that dealing
with `det cmd run ...` case would require us to ship a patched copy of
tqdm, since we don't have direct control over the Python interpreter in
that case. And of course that doesn't address similar output from other
sources.

## Test Plan

- [x] run tqdm in commands and trials, check that output comes out without excessive buffering and shows up in the UI

### DIFF
--- a/master/static/srv/task-logging-setup.sh
+++ b/master/static/srv/task-logging-setup.sh
@@ -12,6 +12,18 @@ mkdir -p "$(dirname "$STDOUT_FILE")" "$(dirname "$STDERR_FILE")"
 ln -sf /proc/$$/fd/1 "$STDOUT_FILE"
 ln -sf /proc/$$/fd/2 "$STDERR_FILE"
 
+# Create a FIFO to monitor process substitution exits, and a count to know how
+# many to wait on.
+DET_LOG_WAIT_FIFO=/run/determined/train/logs/wait.fifo
+DET_LOG_WAIT_COUNT=0
+mkfifo $DET_LOG_WAIT_FIFO
+
+# Save the original stdout and stderr. Process substitutions we'll be doing
+# below block until their stdin is closed and, when we clean up, by saving these
+# we can close them safely and replace stdout and stderr for the shell with the
+# original.
+exec {ORIGINAL_STDOUT}>&1 {ORIGINAL_STDERR}>&2
+
 if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
 	# To do logging with a sidecar in Kubernetes, we need to log to files that
 	# can then be read from the sidecar. To avoid a disk explosion, we need to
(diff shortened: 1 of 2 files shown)


stoksc:
@dzhu wydt about incorporating https://github.com/determined-ai/determined/pull/3930 since this may also be affected.

stoksc:
or does `stdbuf` somehow magically fix it?

dzhu:
Ah yeah, we should let that go in first; I'll go look at that. I remembered we talked about logs, but I totally forgot the specifics. I guess `stdbuf` does reduce the amount of loss you'd expect, in that it drops the buffering from 4k or whatever to 0, but it doesn't avoid the fundamental issue. Now that we know what the issue is, and since the urgency of the fix here is pretty low, let's do the right thing first.

stoksc (approved):
lgtm. I can pull this the other way, into mine.

stoksc:
got a while before next release and this is a pure improvement so I think it's fine that way

master/static/srv/task-logging-setup.sh
```diff
@@ -46,3 +46,15 @@ if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
...
+# A task may output carriage return characters (\r) to do something mildly fancy
+# with the terminal like update a progress bar in place on one line. Python's
+# tqdm library is a common way to do this. That works poorly with our logging,
+# since Fluent Bit interprets everything as one line, causing it to mash
+# everything together and buffer the output for way too long. Since we're not
+# going to do anything like interpreting the carriage returns in our log
+# displays, here we simply replace them all with newlines to get a reasonable
+# effect in those cases. This must be after the multilog exec, since exec
+# redirections are applied in reverse order.
+exec > >(stdbuf -o0 tr '\r' '\n') 2> >(stdbuf -o0 tr '\r' '\n')
```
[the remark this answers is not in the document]
dzhu: Oh right, I was counting on the part above to have run and saved the originals, but of course that doesn't happen unconditionally. I think we need to have an `exec {ORIGINAL_STDOUT}>&1 {ORIGINAL_STDERR}>&2` that always runs, and then everything else can just do the stuff on top of that without worrying about having to save them. Does that make sense?

master/static/srv/task-logging-setup.sh
```diff
@@ -27,22 +39,33 @@ if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
...
+# everything together and buffer the output for way too long. Since we're not
+# going to do anything like interpreting the carriage returns in our log
+# displays, here we simply replace them all with newlines to get a reasonable
+# effect in those cases. This must be after the multilog exec, since exec
+# redirections are applied in reverse order.
+exec > >(
+	stdbuf -o0 tr '\r' '\n'
+	: >$DET_LOG_WAIT_FIFO
+) 2> >(
+	stdbuf -o0 tr '\r' '\n'
```
stoksc, line 67:
```suggestion
	stdbuf -o0 tr '\r' '\n' >&2
```
do we need anything like this? my head hurts lol

dzhu: We do, good catch!

stoksc (approved):
lgtm.

PR #3946, opened 2022-04-19
chore: store allocation resources and agent RM containers in DB.

ioga:
## Description

 - task creation for trials and allocation creation now happen earlier, so the resource tables could FK on these objects.
 - the data is unused for now, but will be needed for rolling upgrades and some EE work.
 - since some of these are using Bun, switch allocation tests to use the real database.

## Test Plan

N / A

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/resource_pool.go
+++ b/master/internal/resourcemanagers/resource_pool.go
@@ -782,19 +782,26 @@ func (c containerResources) Kill(ctx *actor.Context, logCtx logger.Context) {
 	})
 }
 
+// Single asserts there's a single element in the map and take it.
+func Single[K comparable, V any](m map[K]V) (kr K, vr V, ok bool) {
+	// TODO(ilia): move it into a shared utilities package when
+	// it'll be used elsewhere.
+	if len(m) != 1 {
+		return kr, vr, false
+	}
+	for k, v := range m {
+		kr = k
+		vr = v
+	}
+	return kr, vr, true
+}
+
 func (c containerResources) Persist() error {
 	summary := c.Summary()
 
-	var agentID aproto.ID
-
-	switch agentCount := len(summary.AgentDevices); agentCount {
-	case 1:
-		for key := range summary.AgentDevices {
-			agentID = key
-			break
-		}
-	default:
-		return errors.New(fmt.Sprintf("%d agents in containerResources summary", agentCount))
+	agentID, _, ok := Single(summary.AgentDevices)
+	if !ok {
+		return fmt.Errorf("%d agents in containerResources summary", len(summary.AgentDevices))
 	}
 
 	snapshot := agent.ContainerSnapshot{
--- a/master/internal/resourcemanagers/resource_pool.go
+++ b/master/internal/resourcemanagers/resource_pool.go
@@ -1,6 +1,7 @@
 package resourcemanagers
 
 import (
+	"context"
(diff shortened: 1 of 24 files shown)


master/internal/task/resources_list.go
```diff
@@ -1,44 +1,81 @@
...
+		Rank:         rank,
+		ResourceID:   summary.ResourcesID,
+		AllocationID: summary.AllocationID,
+	}
+}
+
+// WipeResourcesState deletes all database contents.
+func WipeResourcesState() error {
+	// Bun requires at least one WHERE for updates and deletes.
+	_, err := db.Bun().NewDelete().Model((*ResourcesWithState)(nil)).Where("1=1").Exec(context.TODO())
```
stoksc, line 52: could truncate instead?

stoksc (commented):
mostly lgtm. the only funky stuff is the db state if we rollback after trying to allocate things and the number of saves throughout allocation.go that may be nice to batch at the end of message handling or something.

master/internal/task/resources_list.go
stoksc: I mean, I don't think we do this anywhere, either. Truncate to me more clearly says 'we're wiping this entire table'. Plus I think (?) it is faster because it works a bit differently internally.

ioga: truncate is not MVCC-safe, but none of these cleanups run in a tx at the moment.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -163,31 +177,45 @@ func (rp *ResourcePool) allocateResources(ctx *actor.Context, req *sproto.Alloca
...
 			return false
 		default:
 			panic(fmt.Sprintf("bad AllocateFreeDevices response: %s", resp))
 		}
 	}
 
+	// Persist allocation_resources and container_resources.
+	for _, cr := range resources {
+		rs := task.NewResourcesState(cr, -1)
+		if err := rs.Persist(); err != nil {
```
stoksc, line 190: seems better to me to do all the db persists then the agent ones. so you can't have a container on an agent that the master doesn't know about if it crashes.

master/internal/resourcemanagers/resource_pool.go
```diff
@@ -753,3 +781,27 @@ func (c containerResources) Kill(ctx *actor.Context, logCtx logger.Context) {
...
+
+	switch agentCount := len(summary.AgentDevices); agentCount {
+	case 1:
+		for key := range summary.AgentDevices {
+			agentID = key
+			break
+		}
+	default:
+		return errors.New(fmt.Sprintf("%d agents in containerResources summary", agentCount))
+	}
```
stoksc, line 798:
```suggestion
func Single[K comparable, V any](m map[K]V) (kr K, vr V, ok bool) {
	if len(m) != 1 {
		return kr, vr, false
	}

	for k, v := range m {
		kr = k
		vr = v
	}
	return kr, vr, true
}
```

🤷 or even add

master/internal/resourcemanagers/resource_pool.go
stoksc: I think I misread here.

master/internal/resourcemanagers/resource_pool.go
ioga: 💪

PR #3948, opened 2022-04-19
chore: slurm proxy support

stoksc:
## Description
This change allows containers which would like proxies in slurm to advertise their proxy address to the master. When this address is advertised, the proxy is registered in the master and endpoints concerning proxies will list addresses for the container.

## Test Plan
- [x] test ancillary rendezvous stuff on a slurm
- [x] test proxy stuff on casablanca

## Commentary (optional)
There are a million ways to write this code. I just went with the simplest I could find for now.

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -65,6 +65,8 @@ type (
 		idleTimeoutWatcher *IdleTimeoutWatcher
 		// proxy state
 		proxies []string
+		// proxyAddress is provided by determined.exec.prep_container if the RM doesn't provide it.
+		proxyAddress *string
 		// active all gather state
 		allGather *allGather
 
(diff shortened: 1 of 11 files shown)


master/internal/task/allocation.go
```diff
@@ -66,6 +66,8 @@ type (
 		idleTimeoutWatcher *IdleTimeoutWatcher
 		// proxy state
 		proxies []string
+		// proxyAddress is provided by determined.exec.prep_container if the RM doesn't provide it.
+		proxyAddress *string
```
ioga, line 70: we need to persist this field to the db [so we are able to load it back on master restart].

stoksc: ok, I suppose I need to let you land your pr then.

stoksc: well, I'll hold off until we do 'slurm fault tolerance' or something. is that OK?

ioga: no problem

ioga (approved):

PR #3955, opened 2022-04-20
feat: Add overall allocation bar to new cluster page [DET-7074]

gt2345:
## Description

Add overall allocation bar and historical usage page to new cluster page, allocation bar and historical page should behave exactly like the old one.

## Test Plan

Navigate to `/clusters` page to verify the allocation bar and `Historical Usage` tab exists.

Screen Shot 2022-04-21 at 5 08 10 PM

### DIFF
--- a/webui/react/src/pages/Clusters.tsx
+++ b/webui/react/src/pages/Clusters.tsx
@@ -8,7 +8,9 @@ import { paths } from 'routes/utils';
 import { ResourceType } from 'types';
 import { percent } from 'utils/number';
 
+import ClusterHistoricalUsage from './Cluster/ClusterHistoricalUsage';
 import ClusterLogs from './ClusterLogs';
+import css from './Clusters.module.scss';
 import ClustersOverview from './Clusters/ClustersOverview';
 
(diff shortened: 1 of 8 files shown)


webui/react/src/pages/Clusters.tsx
```diff
@@ -57,7 +59,10 @@ const Clusters: React.FC = () => {
         <TabPane key="overview" tab="Overview">
           <ClustersOverview />
         </TabPane>
-        <TabPane key="logs" tab="Cluster Logs">
+        <TabPane className={css.tab} key="historical-usage" tab="Historical Usage">
```
EmilyBonar, line 62: question: Why is historical usage the only tab pane with this CSS applied?

EmilyBonar (approved):
LGTM!

webui/react/src/pages/Clusters.tsx
gt2345: Because `Historical Usage` itself does not have padding, it was wrapped into `Page` at another component.

PR #3956, opened 2022-04-20
feat: add drag and drop functionality to experiment list column table [DET-7044]

julian-determined-ai:
## Description

feat: add drog and drop functionality to experiment list column table [DET-7044]

Adds the ability to drag and drop rows in the "Customize Columns" label on the "Experiment List" page. 

## Test Plan
 1. Navigate to the "Experiment List" page and click the "Columns" button.
 2. Under the "Visible" column it should be possible to drag and drop the column items.
     - Drag columns around and then hit the "save" button, the column order on the experiments list table should match the new column order
     - Drag columns around in the table on the experiment list page, and then repeat the above step and ensure the column order still matches the "visible" list.
     - Enter  a search term and ensure the drag and drop functionality still works as expected. 

## Commentary 

The `withDragAndDropProvider` was defined as a way to enable easier usage of the `DndProvider` to provide drag and drop functionality to components. It likely makes more sense to define this function elsewhere and I am open to suggestions.

Additionally, even though the `DndProvider` is defined higher in the component tree, the `useDrag` functionality did not work without the above hook. The issue raised [here](https://github.com/react-dnd/react-dnd/issues/1494#issuecomment-604307618) describes the error that occurs when trying to use `react-dnd` functionality without wrapping the immediate parent with the Provider.

Currently the `dropTarget` which is the item that will be replaced by the item being dragged will show a border around the entire item as defined in `DraggableListItem.module.scss`

## Checklist

[DET-7044]: https://determinedai.atlassian.net/browse/DET-7044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/DraggableListItem.tsx
+++ b/webui/react/src/components/DraggableListItem.tsx
@@ -12,10 +12,11 @@ interface Props {
   onDrop: (indexOne: number, indexTwo: number) => void;
   style: CSSProperties;
 }
-
 interface DroppableItemProps {
   index: number
 }
+
+const DraggableTypes = { COLUMN: 'COLUMN' };
 
 /* eslint-disable-next-line @typescript-eslint/ban-types */
 const withDragAndDropProvider = <T extends {}>
@@ -25,8 +26,6 @@ const withDragAndDropProvider = <T extends {}>
       <Component {...props} />
     </DndProvider>
   );
-
-const DraggableTypes = { COLUMN: 'COLUMN' };
 
 const DraggableListItem: React.FC<Props> = ({ index, style, onClick, children, onDrop }: Props) => {
 
--- a/webui/react/src/components/DraggableListItem.tsx
+++ b/webui/react/src/components/DraggableListItem.tsx
@@ -1,0 +1,63 @@
+import { number } from 'fp-ts';
+import React, { CSSProperties } from 'react';
+import {useDrag, useDrop} from 'react-dnd';
+import { DndProvider } from 'react-dnd'
+import { HTML5Backend } from 'react-dnd-html5-backend'
+
+import css from './DraggableListItem.module.scss';
+
+interface Props {
+  style: CSSProperties;
+  onClick: (event: React.MouseEvent) => {};
+  children: React.ReactChildren;
+  onDrop: (indexOne: number, indexTwo: number) => {};
+  index: number;
+}
+
+interface DroppableItemProps {
+  index: number
+}
+
+const withDragAndDropProvider = (Component: React.FunctionComponent<any>) => (props: any) =>{
+  return (
+    <DndProvider backend={HTML5Backend}>
+        <Component {...props}/>
+    </DndProvider>
+  )
+}
+
+const DraggableTypes = {
+  COLUMN: 'COLUMN'
+}
+
+
+  const DraggableListItem: React.FC<Props>= ({index, style ,onClick, children, onDrop }: Props) => {
+
+    const [{ isOver }, drop] = useDrop(
+      () => ({
+        accept: DraggableTypes.COLUMN,
+        drop: (item: DroppableItemProps) => {
+          onDrop(index, item.index)
+        },
+        collect: (monitor) => ({
+          isOver: !!monitor.isOver(),
+          canDrop: !!monitor.canDrop()
+        })
+      }),
+      []
+    )
+
+    const [, drag] = useDrag(() => ({
+      type: DraggableTypes.COLUMN,
+      item: {index},
+    }))
+    
+      return (
+        <span ref={drop}>
+          <li ref={drag} className={isOver ? css.dropTarget: undefined} style={style} onClick={onClick}>
+              {children}
+          </li>
+        </span>
+      )
+    }
+  export default withDragAndDropProvider(DraggableListItem);
--- a/webui/react/src/components/DraggableListItem.tsx
+++ b/webui/react/src/components/DraggableListItem.tsx
@@ -1,63 +1,64 @@
-import { number } from 'fp-ts';
 import React, { CSSProperties } from 'react';
-import {useDrag, useDrop} from 'react-dnd';
-import { DndProvider } from 'react-dnd'
-import { HTML5Backend } from 'react-dnd-html5-backend'
+import { useDrag, useDrop } from 'react-dnd';
+import { DndProvider } from 'react-dnd';
+import { HTML5Backend } from 'react-dnd-html5-backend';
 
 import css from './DraggableListItem.module.scss';
 
 interface Props {
-  style: CSSProperties;
-  onClick: (event: React.MouseEvent) => {};
-  children: React.ReactChildren;
-  onDrop: (indexOne: number, indexTwo: number) => {};
+  children: React.ReactNode;
   index: number;
+  onClick: (event: React.MouseEvent) => void;
+  onDrop: (indexOne: number, indexTwo: number) => void;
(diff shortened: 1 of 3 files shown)


webui/react/src/components/DraggableListItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const [ , drag ] = useDrag(() => ({
+    item: { index },
+    type: DraggableTypes.COLUMN,
+  }));
+
+  return (
+    <span ref={drop}>
+      <li
+        className={isOver ? css.dropTarget : undefined}
```
trentwatt, line 54: `RE: drop styles`

`className={isOver ? (dropDirection === 'above' ? ...`

trentwatt (commented):
Looking really good! There are two main things:

- drag and drop does not play nicely with column search (see video)
- dropTarget styles should reflect which columns you are dropping between

for the first, i think you basically just need to get the index for the source and destination columns in `visibleColumns`. right now you are swapping columns in `visibleColumns` based on their indices in `filteredVisibleColumns`. one possible fix is to add the column name to your dnd item, then in your onDrop callback, look up its index in `visibleColumns`. I have sketched out a few pieces of that approach under the heading `RE: column -> visibleColumnsIndex`.

for the second, the current style for the `dropTarget` highlights the entire row of the target column. from this, it is not clear where the column will end up: before the highlighted column or after. with the way the code is written, the answer is: before if it's coming from below, after if it's coming from above (b/c the splicing shifts the indices of the columns after). to indicate this, would just need to determined where the drop is coming from and do a `border-bottom` or `border-top` accordingly. I sketched out some pieces of this under the heading `RE: drop styles`.

really great work though.

https://user-images.githubusercontent.com/28383080/164343769-2df483b4-5467-46d3-9e1e-ccc726642dd1.mov


A heads up: if you play around with column search, you may encounter [DET-6837](https://determinedai.atlassian.net/browse/DET-6837): just wanted to let you know in advance that it's not related to your PR.

webui/react/src/components/DraggableListItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+    [],
+  );
+
+  const [ , drag ] = useDrag(() => ({
+    item: { index },
+    type: DraggableTypes.COLUMN,
+  }));
+
+  return (
+    <span ref={drop}>
```
trentwatt, line 52: `question`: what does the span achieve here? vs just putting the ref on the `li`?

webui/react/src/hooks/useModal/useModalCustomizeColumns.tsx
```diff
@@ -100,15 +101,40 @@ const useModalCustomizeColumns = ({
...
+    );
+  }, [ renderColumnName ]);
+
   const renderHiddenRow = useCallback(({ index, style }) => {
     const row = filteredHiddenColumns[index];
     return renderRow(row, style, () => makeVisible(row));
   }, [ filteredHiddenColumns, makeVisible, renderRow ]);
 
   const renderVisibleRow = useCallback(({ index, style }) => {
     const row = filteredVisibleColumns[index];
```
trentwatt, line 129: `nit`/ `question` do you think it would make the code easier to read if all instances of `row`/`Row` were changed to `column`/`Column`. I get that the actual ui elements are rows of a list, but just wanted to throw it out there.

trentwatt:
Also, probably worth noting somewhere in a comment that the modal drag and drop reordering won't work for `ResponsiveTable.tsx`, since that component takes the column order from the `props.columns` instead of `settings.columns`

Currently none of the pages with `ResponsiveTable` use the customize columns modal, but would be worth calling it out.

julian-determined-ai:
The index comparison has been updated in the PR to ensure that `columns` are compared based on their index in the `visibleColumns` array. This ensures that the drag and drop functionality works as expected, including in cases where this is a filter being applied on the `column` name.

A video of this feature working can be seen below: 

CD8F6F22-49E1-4D5E-8933-7CA1E593055F_1_102_o

webui/react/src/components/DraggableListItem.tsx
trentwatt: nvm, i see it.

webui/react/src/hooks/useModal/useModalCustomizeColumns.tsx
julian-determined-ai: I initially found this confusing because the terms `row` and `column` usually refer to the same matrix or table. But here we are referencing a list `row` from the list of `columnNames`. I updated several variables to use `columnName` instead of just `column` which makes things clearer to me.

julian-determined-ai:
> Also, probably worth noting somewhere in a comment that the modal drag and drop reordering won't work for `ResponsiveTable.tsx`, since that component takes the column order from the `props.columns` instead of `settings.columns`
> 
> Currently none of the pages with `ResponsiveTable` use the customize columns modal, but would be worth calling it out.

Could you let me know where would be the best place to call this out?

trentwatt:
> > Also, probably worth noting somewhere in a comment that the modal drag and drop reordering won't work for `ResponsiveTable.tsx`, since that component takes the column order from the `props.columns` instead of `settings.columns`
> > Currently none of the pages with `ResponsiveTable` use the customize columns modal, but would be worth calling it out.
> 
> @trentwatt Could you let me know where would be the best place to call this out?

i guess it's not needed. i was going to say [here](https://github.com/determined-ai/determined/blob/5d65f6443ee0106e5d29b5caff240f1d1fb95011/webui/react/src/pages/ExperimentList.tsx#L610) since that's where people would be copying the pattern from, but that's going to be replaced as a part of the work going on in the workspaces feature branch.

webui/react/src/hooks/useModal/useModalCustomizeColumns.tsx
```diff
@@ -128,11 +136,11 @@ const useModalCustomizeColumns = ({
...
     return renderDraggableRow(
-      index,
       row,
+      index,
       style,
       () => makeHidden(row),
-      (itemIndex: number, newIndex:number) => switchRowOrder(itemIndex, newIndex),
+      (column: string, newColumnNeighbor: string) => switchRowOrder(column, newColumnNeighbor),
```
trentwatt, line 143: `nit` i think this can just be `switchRowOrder`

webui/react/src/components/DraggableListItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { HTML5Backend } from 'react-dnd-html5-backend';
+
+import css from './DraggableListItem.module.scss';
+
+interface Props {
+  children: React.ReactNode;
+  columnName: string;
+  index: number;
+  onClick: (event: React.MouseEvent) => void;
+  onDrop: (column: string, newNeighborColumnName: string) => void;
```
trentwatt, line 13: `nit` elsewhere the second arg is called `newNeighborColumn`

webui/react/src/hooks/useModal/useModalCustomizeColumns.tsx
trentwatt: (also elsewhere `newColumnNeighbor` is referred to as `newNeighborColumn`, but this redundant if you make the first change)

trentwatt (approved):
Nice, LGTM!

couple nits I leave to your discretion.

PR #3963, opened 2022-04-20
chore: apply filtering to task logs

hkang1:
## Description

Trial Logs currently have log message filtering applied. The standard timestamp is filtered out, and the kubernetes container id is filtered out as well. Apply the same logic to task log messages as well.

BEFORE (double timestamp):
2022-04-20 at 3 10 PM

AFTER:
2022-04-20 at 3 09 PM

UPDATED FORMAT:
Screen Shot 2022-04-21 at 3 40 58 PM

## Test Plan

- [ ] Check that the task logs don't display the timestamp twice (2nd time because it's being displayed in the message)
- [ ] Check that in kubernetes cluster task logs, the container id `[8 hex-digits]` doesn't show up

## Checklist

### DIFF
--- a/webui/react/src/services/decoder.ts
+++ b/webui/react/src/services/decoder.ts
@@ -525,41 +525,44 @@ const decodeV1LogLevelToLogLevel = (level: Sdk.V1LogLevel): types.LogLevel | und
   return logLevelMap[level];
 };
 
-const defaultRegex = /^\[([^\]]+)\]\s([\s\S]*)(\r|\n)$/im;
-const kubernetesRegex = /^\s*([0-9a-f]+)\s+(\[[^\]]+\])\s\|\|\s(\S+)\s([\s\S]*)(\r|\n)$/im;
+const newlineRegex = /(\r\n|\n)$/g;
+const messageRegex = new RegExp(
+  [
+    '^',
+    '\\[([^\\]]+)\\]\\s',                             // timestamp
+    '([0-9a-f]{8})?\\s?',                             // container id
+    '\\|\\|',                                         // divider ||
+    '(\\s(CRITICAL|DEBUG|ERROR|INFO|WARNING):\\s)?',  // log level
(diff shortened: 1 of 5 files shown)


stoksc (approved):
can't approve the code, but I tested locally and it works great

webui/react/src/services/decoder.ts
```diff
@@ -528,32 +528,36 @@ const decodeV1LogLevelToLogLevel = (level: Sdk.V1LogLevel): types.LogLevel | und
 const defaultRegex = /^\[([^\]]+)\]\s([\s\S]*)(\r|\n)$/im;
 const kubernetesRegex = /^\s*([0-9a-f]+)\s+(\[[^\]]+\])\s\|\|\s(\S+)\s([\s\S]*)(\r|\n)$/im;
 
+const filterLogMessage = (message: string): string => {
```
hamidzr, line 531: nit: rename? translation or formatting describes the operation better

webui/react/src/services/decoder.ts
```diff
@@ -528,32 +528,36 @@ const decodeV1LogLevelToLogLevel = (level: Sdk.V1LogLevel): types.LogLevel | und
...
+  let filteredMessage = message.trim();
+
+  if (defaultRegex.test(message)) {
```
hamidzr, line 534: I'm not sure how .test is implemented but if it's just checking for matches underthe hood just doing `.match` once and storing the results should be faster considering we might be processing a lot of log lines during scroll events.

hamidzr (approved):
lgtm

webui/react/src/services/decoder.ts
EmilyBonar: According to [this website](https://www.measurethat.net/Benchmarks/Show/11043/0/stringmatch-vs-regextest), `.test` is about 58% faster than `.match`.

hkang1: good call

PR #3965, opened 2022-04-20
fix: add test warmup command e2e [DET-5803] 

nrajanee:
## Description

fix: add test warmup command e2e [DET=5803] 

## Test Plan

Made sure the e2e_tests run fine (even with the timeouts specified for tests in test_logging.py and and test_profiling.py)

### DIFF
--- a/e2e_tests/tests/conftest.py
+++ b/e2e_tests/tests/conftest.py
@@ -91,7 +91,7 @@ def instantiate_gpu() -> None:
     ]
 
     subprocess.run(
-        command, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env
+        command, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
     )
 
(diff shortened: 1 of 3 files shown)


e2e_tests/tests/conftest.py
```diff
@@ -81,6 +81,13 @@ def pytest_addoption(parser: Parser) -> None:
     parser.addoption("--follow-local-logs", action="store_true", help="Follow local docker logs")
 
 
+@pytest.fixture(scope="session", autouse=True)
+def instantiate_gpu() -> None:
+    command = ["det", "cmd", "--config", "resources.slots=1", "'sleep 30'"]
```
ioga, line 86: should save a tiny bit of time
```suggestion
    command = ["det", "cmd", "--config", "resources.slots=1", "'sleep 1'"]
```

ioga (approved):
could you please fix the title to have `DET-5803` instead of `=`? someway somehow jira integration doesn't care and works anyway, but we use dash everywhere.

PR #3976, opened 2022-04-21
Uplot refactor refactor

hkang1:
## Checklist


webui/react/src/components/UPlot/UPlotChart.tsx
```diff
@@ -63,145 +67,98 @@ const shouldUpdate = (
...
+        scalesZoomData.current[scaleKey] = {
+          isZoomed: currentMax < max || currentMin > min,
+          max,
+          min,
+        };
+      } ],
     },
-    options || {},
-  ) as uPlot.Options;
+  }, options ?? {}) as uPlot.Options, [ options ]);
```
trentwatt, line 107: `options` is constantly getting reassigned, so I don't think the `useMemo` is doing anything for us here

trentwatt: and even it was doing something we could be closing over stale values from `scalesZoomData`

webui/react/src/components/UPlot/UPlotChart.tsx
```diff
@@ -63,145 +67,98 @@ const shouldUpdate = (
...
    */
   useEffect(() => {
     if (!chartDivRef.current) return;
-    scalesZoomData.current = {};
-    const data = [ [], [ [] ] ] as unknown as uPlot.AlignedData;
-    if (!chartRef?.current) {
-      chartRef.current = new uPlot(optionsRef.current, data, chartDivRef.current);
-      setChartIsMounted(true);
-    }
+    if (!shouldRecreate(chartRef.current, fullOptions, prevFullOptions)) return;
```
trentwatt, line 116: this is conditional and the destroy is not so if this returns false the charts will be destroyed. 

`if (!chartRef?.current) {` is conditional too but only on the chart existing in the first place

webui/react/src/components/UPlot/UPlotChart.tsx
trentwatt: ^^wait is that true though? My thinking was that since max/min are numbers they will get passed in by value and not change if the ref changes, but I am not 100%

Also RE: options getting reassigned: it looks like most places besides the profiler that is not happening. I am working on a refactor of the profiler metrics themselves that allows us to avoid reassigning options, so this would be useful if we can land that.

hkang1:
@trentwatt, so I looked at your PR and mine and did like a hybrid code meld. 100% agree with you regarding using `optionsRef` btw. I purposely left the `console.logs` in there so if you were running this code you can see the lifecycle of the component. This seems work pretty well with profiler, single and multi-trial experiment charts. Let me know what you think and if there are areas of concern.

Some of the themes are:

- keeping all chart params as `useRefs` to better control the chart, particularly separating when to update and when to recreate.
- an update occurs when data, series focus, chart container and scroll position changes
- a recreate occurs when chart options changes (with more fine grain control via `shouldRecreate` logic)
- 3 key types useEffects.
  - one to create and recreate the chart when `shouldRecreate` is true
  - one to update the chart, particularly the one the handles the data changes
  - final one to simply cleanup and teardown the chart when the chart component unmounts.

What I didn't do and things not addressed:
- tune the `shouldRecreate` (basically the same as your `shouldUpdate`) to work properly. For example, in HpScatterPlots & HpHeatMaps, when you change the filters, it doesn't properly detect that it should recreate the chart, based on the option changes.
- tune `zoomInfo` to ensure zooming is properly preserved, but from what I've observed this seems to be working

trentwatt:
changes incorporated

PR #3977, opened 2022-04-21
feat: Projects CLI, add sort-by fields

mapmeld:
## Description

Work in progress on a handful of workspaces / projects issues:
- adding archive/unarchive to projects cli, and moving experiments between projects
- removing a.sort to fix a bug in GetProjectExperiments
- add two sortable fields (forked-from and resource-pool); Due to a weird bug where GetExperiments and GetProjectExperiments were too similar for the Python bindings, I added a third field
- make experiment labels filter-able by project_id

## Checklist


harness/determined/cli/experiment.py
```diff
@@ -747,6 +748,18 @@ def unarchive(args: Namespace) -> None:
     print("Unarchived experiment {}".format(args.experiment_id))
 
 
+@authentication.required
+def move_project(args: Namespace) -> None:
```
EmilyBonar, line 752: question: Is this meant to be called `move_project`? The code seems to be implementing move_experiment.

master/internal/api_workspace.go
```diff
@@ -85,8 +85,6 @@ func (a *apiServer) GetWorkspaceProjects(ctx context.Context,
 		return nil, err
 	}
 
-	a.sort(resp.Projects, req.OrderBy, req.SortBy,
```
EmilyBonar, line 88: question: Does this still work even with this `a.sort` removed?

EmilyBonar (approved):
LGTM

harness/determined/cli/experiment.py
mapmeld: ✅  fixed

master/internal/api_workspace.go
mapmeld: My new understanding is that `a.sort` is for objects which aren't sorted by the database for some reason (for example, it's used on checkpoints and in notebooks, but never on GetExperiments).
e2e_tests include sorting by id and name to verify orders that workspaces and projects are returned

webui/react/src/services/apiConfig.ts
```diff
@@ -400,11 +400,11 @@ export const getExpTrials: Service.DetApi<
 };
 
 export const getExperimentLabels: Service.DetApi<
-  Service.EmptyParams, Api.V1GetExperimentLabelsResponse, string[]
+  Service.ExperimentLabelsParams, Api.V1GetExperimentLabelsResponse, string[]
 > = {
   name: 'getExperimentLabels',
   postProcess: (response) => response.labels || [],
-  request: (options) => detApi.Experiments.getExperimentLabels(options),
+  request: (options) => detApi.Experiments.getExperimentLabels(options.project_id || 0),
```
EmilyBonar, line 407: I believe this should be `request: (params, options) => detApi.Experiments.getExperimentLabels(params.project_id, options),`

PR #3984, opened 2022-04-22
fix: avoid permanent filtered state in model registry [DET-6946]

trentwatt:
## Description

fix: avoid permanent filtered state in model registry [DET-6946]

## Test Plan

- preferably have your own cluster to test the actual no model state (rather than deleting all the models on latest-master, I guess you can though). 
- go to `/det/models`
- if there are existing models delete them
- you should see the 'No Models Registered' Page
- add a new model, see it appear in the table
- apply a bunch of different kinds of filters. even when there are no models matching the filters, you should still have the option to clear the filters
## Commentary (optional)

Note:
if someone deletes all their models while in a filtered state, they will not get back to the  'No Models Registered' Page, it will just say "No data," like it does when there's none matching filters.

the filtering happens API side so there's no way to know that there are no models from the response to a filtered request. 
you would have to make an extra API request to get the unfiltered models and do a check on that. seems like too much to address what is already an acceptable result in an edge case. 

## Checklist

[DET-6946]: https://determinedai.atlassian.net/browse/DET-6946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ModelRegistry.tsx
+++ b/webui/react/src/pages/ModelRegistry.tsx
@@ -435,12 +435,19 @@ const ModelRegistry: React.FC = () => {
     showModal({});
   }, [ showModal ]);
 
+  // archive === false is a filter
+  const filtered = (settings.archived != null
+  || settings.description
+  || settings.tags
+  || settings.name
+  || settings.users);
+
   return (
     <Page docTitle="Model Registry" id="models" loading={isLoading}>
       <Section
         options={<Button onClick={showCreateModelModal}>New Model</Button>}
         title="Model Registry">
-        {(models.length === 0 && !isLoading) ?
+        {(models.length === 0 && !isLoading && !filtered) ?
           (
             <div className={css.emptyBase}>
               <div className={css.icon}>


gt2345:
Great work on filters, but I noticed couple other minor issues while testing

Not able to create new model with name shorter than existing model.
Screen Shot 2022-04-22 at 11 01 32 AM

When filtering by tag, models without tag are still showing.
Screen Shot 2022-04-22 at 11 13 09 AM

gt2345 (approved):
Great work on descriptions and test plan

trentwatt:
> Great work on filters, but I noticed couple other minor issues while testing
> 
> Not able to create new model with name shorter than existing model. Screen Shot 2022-04-22 at 11 01 32 AM
> 
> When filtering by tag, models without tag are still showing. Screen Shot 2022-04-22 at 11 13 09 AM

yes, I noticed a (probably related) issue with the filter but that is a good catch. i will get on those next.  

thanks for thorough test.

PR #3985, opened 2022-04-22
chore: bring in can-ndjson-stream and apply patch [DET-7121]

hkang1:
## Description

This only impacts our dev environment but there is a lot of noise when we call the AbortController cancel for a ndjson streaming endpoint. The issue is that there are uncaught errors within the `can-ndjson-stream` library.

Sample error we encounter on the browser client.
image (1)

Different browsers behave differently. FireFox notably shows a full page error and hitting escape allows you to continue on. On Chrome, it shows up in the console as an error but continues to function.

Decided to bring this MIT library into our project because...
1. it doesn't seem to be an active project anymore
2. there is a patch we want to apply to fix the describe problem above
3. it is only ~60 lines and fairly easy to maintain

## Test Plan

- [ ] run a local web development instance (via `make live`) and ensure that there are no `The operation was aborted by the user` or any message in that similar vein.

## Checklist

### DIFF
--- a/webui/react/src/services/can-ndjson-stream.js
+++ b/webui/react/src/services/can-ndjson-stream.js
@@ -1,0 +1,77 @@
+"use strict";
+/*exported ndjsonStream*/
+
+var namespace = require('can-namespace');
+
+/**
+ * Brought this MIT library in to our source because...
+ * 1. it doesn't seem to be an active project anymore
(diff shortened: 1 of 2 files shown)


trentwatt (commented):
Fix needed per discussion.

PR #3986, opened 2022-04-22
docs: add release notes for 0.17.15

mackrorysd:
Add release notes for 0.17.15.

### DIFF
--- a/docs/release-notes.txt
+++ b/docs/release-notes.txt
@@ -8,6 +8,31 @@
  Version 0.17
 **************
 
+Version 0.17.15
+===============
+
+**Release Date:** April 22, 2022
+
+**Breaking Changes**
+
+-  API: Endpoints for getting or updating a user now accept a ``userId`` instead of ``username`` as
+   the path parameter.
+
+**Bug Fixes**
+
+-  Fix an issue where deleted experiments would get stuck in a ``DELETING`` state indefinitely due
+   to their checkpoint GC tasks not to completing.
+
+-  API: Fix an issue where a reported job state could be stale due to a faulty caching mechanism.
+   This could have resulted in an experiment showing in `queued` or `scheduled` state, either in CLI
+   or WebUI, when it was in the other state.
+
+**New Features**
+
+-  Add a translation of DeepSpeed's DCGAN example using the new DeepSpeedTrial API.
+
+----
+
 Version 0.17.14
 ===============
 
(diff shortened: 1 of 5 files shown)


docs/release-notes.txt
```diff
@@ -8,6 +8,31 @@
...
+
+**Breaking Changes**
+
+-  API: Endpoints for getting or updating a user now accept a ``userId`` instead of ``username`` as
+   the path parameter.
+
+**Bug Fixes**
+
+-  Fix an issue where deleted experiments would get stuck in a ``DELETING`` state indefinitely due
+   to their checkpoint GC tasks not to completing.
```
gh-determined-ai, line 24: Change to "... not completing."

gh-determined-ai (changes requested):

docs/release-notes.txt
```diff
@@ -8,6 +8,31 @@
...
+
+-  API: Fix an issue where a reported job state could be stale due to a faulty caching mechanism.
+   This could have resulted in an experiment showing in `queued` or `scheduled` state, either in CLI
```
gh-determined-ai, line 27: Change to "... state in the CLI or WebUI when it was in the other state."
Or, should it really be "... in another state."?

gh-determined-ai (changes requested):

mackrorysd:
The Semantic PR complaint is because the action is being smart and detecting the same commit ID between this one and #3987, which I created manually and proposed merging to the wrong back. #3986 (this one) is correct in both aspects: right branch, correctly formatted title. I'll update the release tooling and docs to stop telling people to create a PR when the tooling has already done it correctly.

docs/release-notes.txt
mackrorysd: I think the original wording is correct but I'll have @hamidzr confirm.

docs/release-notes.txt
mackrorysd: Fixed!

docs/release-notes.txt
hamidzr: I'm using "other" to communicate that it could have been only between two options eg if it wasn't `scheduled` it should have been `queued` and not just any other state.
"another" seems correct as well, but I think it would be less accurate.

gh-determined-ai: I agree with leaving it as is in that case, although, it's not clear that those are the only possible states; if you think most users should know that, it's fine.

gh-determined-ai (approved):

mackrorysd:
Merging because test-e2e is going to wait for all the options to time out

PR #3992, opened 2022-04-22
fix: handle infinite validation metric values in more cases

dzhu:
## Description

Since JSON can't represent infinity, we store infinite metric values as
the string `"Infinity"` or `"-Infinity"` in JSON in the database. Some
ways of taking those strings from PostgreSQL and converting them to
numbers work and preserve the infinite values, and some do not. The most
correct, and also most natural, way is to use `(map->>key)::float8`; the
similar `(map->key)::text::float8` adds quotes to string values and so
fails the last cast. (Also, we were using the `numeric` type before when
we did that, which does not support infinite values anyway.) Another
thing we were doing was running queries that returned jsonb values from
PostgreSQL back to Go and then attempting to do a standard
deserialization into a float64, which would fail on these string values;
we can take advantage of the same in-PostgreSQL cast as above to do the
right thing there.

We also need to add handling for when we have such an infinite value and
want to serialize it out to JSON. That's accomplished with a wrapper
type; for now, it's defined and used in the one place I've seen it to be
necessary so far, but it may make sense to move it somewhere and make it
public if there are other places it could be helpful.

## Test Plan

- [x] run an experiment that produces infinite validation metrics, check that this change makes it go from showing the error we were seeing at the customer to producing a working page

## Commentary

Arguably no one should ever actually want to produce an infinite metric value, and so these values should be ignored instead. But for now I'm going to take the stance that these values do have a specific meaning and we should go with that meaning, and if the experiment isn't meant to be producing these values then it should be fixed.

### DIFF
--- a/master/internal/db/postgres_experiments.go
+++ b/master/internal/db/postgres_experiments.go
@@ -214,7 +214,7 @@ func (db *PgDB) TrainingTrialsSnapshot(experimentID int, minBatches int, maxBatc
 SELECT
   t.id AS trial_id,
   t.hparams AS hparams,
-  s.metrics->'avg_metrics'->$1 AS metric,
+  (s.metrics->'avg_metrics'->>$1)::float8 AS metric,
   s.end_time AS end_time,
   s.total_batches as batches
 FROM trials t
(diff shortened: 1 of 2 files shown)


stoksc (approved):

dzhu:
Oh, this fixes the page we initially saw the error on, but other ones do similar things, which I should've checked before. I'll get on that.

PR #3993, opened 2022-04-22
fix: pass task time not from logCtx

gt2345:
## Description

In order to avoid breaking in the future, do not read task type from logCtx


master/pkg/tasks/task_command.go
```diff
@@ -81,6 +81,6 @@ func (s GenericCommandSpec) ToTaskSpec(
 	}
 
 	res.UseFluentLogging = true
-
+	res.TaskType = model.TaskTypeCommand
```
stoksc, line 84: generic command spec is actually for commands, shells, tensorboards and notebooks, since they used to always be the same

gt2345: Fixed

master/pkg/tasks/task_command.go
```diff
@@ -31,6 +31,7 @@ type GenericCommandSpec struct {
 // ToTaskSpec generates a TaskSpec.
 func (s GenericCommandSpec) ToTaskSpec(
 	keys *ssh.PrivateAndPublicKeys,
+	taskType model.TaskType,
```
stoksc, line 34: `keys` is only passed here because it isn't know until task spec creation time. since the other is known right off the bat, we can just specific it when the GenericCommandSpec is created, I'd think.

stoksc (approved):

PR #3994, opened 2022-04-22
feat: add modal to explain why users cannot delete items [DET-6998]

julian-determined-ai:
## Description

feat: add modal to explain why users cannot delete items [DET-6998]

The "Delete" button is no longer disabled when deleting a model or deregistering a model version. Instead a modal is shown explaining that only an admin or the user who created the item is able to delete it. 

20FB050E-AD2E-44BE-BC4E-181BA4B8FDA4

Screen Shot 2022-04-25 at 2 23 50 PM

## Test Plan

1. Using an admin account or the account of a user that has created the model or version, attempt to delete a model or deregister version. Everything should work as expected.
2. Using a non admin account or account of a user that did not create a model or version. Attempt to deregister a model or version. The "Delete" or "Deregister" button should be enabled. After clicking the button, a modal should show explaining that "Only the item creator or an admin can delete items from the model registry".

## Commentary (optional)

The `showCannotDelete` function has been defined several times in this PR but would seeming be more appropriate defined in a single place such as in a `helpers` file. However, it does not seem as though any such directory is currently defined. I wondered if that was due in an effort to explicitly separate function/utility concerns. However, if we do think this function should be defined in one place only I will refactor. 

update: The refactor mentioned above has been implemented.

## Checklist

[DET-6998]: https://determinedai.atlassian.net/browse/DET-6998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ModelVersionDetails/ModelVersionHeader.tsx
+++ b/webui/react/src/pages/ModelVersionDetails/ModelVersionHeader.tsx
@@ -56,6 +56,17 @@ const ModelVersionHeader: React.FC<Props> = (
     });
   }, [ onDeregisterVersion, modelVersion.version ]);
 
+  const showCannotDelete = () => {
+    Modal.confirm({
+      closable: true,
+      content: 'Only the item creator or an admin can delete items from the model registry.',
+      icon: null,
+      maskClosable: true,
+      okText: 'Ok',
+      title: 'Unable to Delete',
+    });
+  };
+
   const infoRows: InfoRow[] = useMemo(() => {
     return [ {
       content: (
@@ -117,9 +128,9 @@ const ModelVersionHeader: React.FC<Props> = (
       },
       {
         danger: true,
-        disabled: !isDeletable,
+        disabled: false,
         key: 'deregister-version',
-        onClick: showConfirmDelete,
+        onClick: () => isDeletable ? showConfirmDelete() : showCannotDelete(),
         text: 'Deregister Version',
       },
     ];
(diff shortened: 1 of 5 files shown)


EmilyBonar (changes requested):
The code itself is fine, but you're correct that redefining the modal that many times is undesirable. Take a look in the `hooks/useModal` directory, and try using that. I think `useModalExperimentDelete.tsx` is a pretty simple one you could base this off of. You might also consider changing the text to be more general so that this modal could be applied outside the Model Registry; something like `Only the item creator or an admin can delete this item.`

Also, this is actually ticket 6998, not 6698.

webui/react/src/pages/ModelVersionDetails/ModelVersionHeader.tsx
```diff
@@ -117,9 +118,10 @@ const ModelVersionHeader: React.FC<Props> = (
       },
       {
         danger: true,
-        disabled: !isDeletable,
+        disabled: false,
```
EmilyBonar, line 121: nit: this line can be removed

EmilyBonar (approved):
LGTM

webui/react/src/pages/ModelVersionDetails/ModelVersionHeader.tsx
EmilyBonar: to be clear, I'm referring to the `disabled` line

julian-determined-ai: agreed! This has been updated

PR #3995, opened 2022-04-22
chore: handle rank id in log entries

hkang1:
## Description

log parsing wasn't handling the rank ids in the log messages. Mapping specific to the [log generation code](https://github.com/determined-ai/determined/blob/master/master/pkg/model/task.go#L284-L321).

## Test Plan

- [ ] run an experiment that generates rank ids (dcgan_deepspeed_mnist) and check that the log formatting looks good (no duplicate timestamps and other weird artifacts).

## Checklist

### DIFF
--- a/webui/react/src/services/decoder.ts
+++ b/webui/react/src/services/decoder.ts
@@ -531,8 +531,9 @@ const messageRegex = new RegExp(
     '^',
     '\\[([^\\]]+)\\]\\s',                             // timestamp
     '([0-9a-f]{8})?\\s?',                             // container id
-    '\\|\\|',                                         // divider ||
-    '(\\s(CRITICAL|DEBUG|ERROR|INFO|WARNING):\\s)?',  // log level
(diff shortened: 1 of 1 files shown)


webui/react/src/services/decoder.ts
```diff
@@ -531,8 +531,9 @@ const messageRegex = new RegExp(
     '^',
     '\\[([^\\]]+)\\]\\s',                             // timestamp
```
hamidzr, line 532: question: do we need double escaping because we're going from string to Regex?

hkang1: exactly, not the sexiest look i know :)

hkang1:
> this looks good to me although it's always hard to ready regexes like this for me. I have to put them into https://regex101.com/ or through a test to verify.
> 
> left = before Screen Shot 2022-04-22 at 14 02 38
> 
> if we think this'll stay around it might pay off to turn it into a test case

I'm really hoping this is temporary until we get the fields extracted by the backend

PR #3996, opened 2022-04-22
feat: Users-specific pinned workspaces

mapmeld:
## Description

- A user can pin a workspace to signify that it is important / current to their work.
- API supports pinning, unpinning, GetWorkspaces optional filtering by pinned/unpinned
- Returned Workspace object (GetWorkspace, GetWorkspaces, PatchWorkspace) include `"pinned": bool` field
- Pins contain a `created_at` column which could be used for sorting, but TODO

## Testing

- Includes e2e_tests for pinning, checking len(pinned) = 1, unpinning, then checking len(pinned) = 0.
- Should test GetWorkspace and GetWorkspaces

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_workspace_org.py
+++ b/e2e_tests/tests/cluster/test_workspace_org.py
@@ -99,6 +99,20 @@ def test_workspace_org() -> None:
             map(lambda w: w.name, list_test_3)
         )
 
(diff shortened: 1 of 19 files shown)


EmilyBonar (approved):
LGTM

PR #4003, opened 2022-04-25
fix: handle zeros correctly in HpTrialTable metricSorter

johnkim-det:
## Description

Fixing how zeros are sorted on Experiment Visualization tables.

## Test Plan

- [ ] For Learning Curve and HP Parallel Coordinates tables, make sure that best validation metric (2nd column from left) correctly sorts zeros as the lowest value.

Screen Shot 2022-04-25 at 10 40 40 AM

## Checklist


trentwatt (approved):
LGTM. 

I checked the other tables as well, looks like we're good across the board.

PR #4004, opened 2022-04-25
refactor: theme architecture [DET-6211]

hkang1:
## Description

The theme needed some update moving forward to:
1. provide rules and structure to make it easier to decide how and what to style the future components
2. simplify structure to minimize developer decisions for styling
3. enable multiple themes designs
4. allow dynamic theme changes

The following covers the most important high level concepts that were incorporated during the design of the theme.

Layers theme structure helps with mimicking some physical layers of the views.
Layers

Auto-generated Strong/Weak allows to keep the actual themes that need to be defined to a much manageable pool of variables.
StrongWeak

Continuing to use status colors to express DL item lifecycles.
Status

Dark Mode
Screen Shot 2022-05-02 at 11 43 36 PM

Light Mode
Screen Shot 2022-05-02 at 11 44 00 PM

## Test Plan

- [ ] validate dark/light mode switching works via OS level mode preferences
- [ ] check that all pages render properly
- [ ] check interactive components, that the sub rendered view (items hidden behind interactions) render and behave properly

## Checklist


webui/react/src/themes.ts
```diff
@@ -269,34 +43,297 @@ const stateColorMapping = {
...
+  statusSuccessStrong: undefined,
+  statusSuccessWeak: undefined,
+  statusSuccessOn: 'rgba(255, 255, 255, 1.0)',
+  statusSuccessOnStrong: undefined,
+  statusSuccessOnWeak: undefined,
+  statusWarning: 'rgba(204, 153, 0, 1.0)',
+  statusWarningStrong: undefined,
+  statusWarningWeak: undefined,
+  statusWarningOn: 'rgba(255, 255, 255, 1.0)',
+  statusWarningOnStrong: undefined,
```
hamidzr, line 211: do we need to individually define these or can we assume that each key here `k` comes with `kWeak` and `kStrong`? can we generate that programmatically as well?

hamidzr: well it seems we can't. it wouldn't apply to all keys. what if we apply it to certain value types and patterns

- if the variations are undefined
- if the original key is `\d+px` or `^rgba`

hamidzr: that should make it easier to manually go through this list

hamidzr (approved):
the on hover effect for the navigation sidebar seems to have been lost or not be visible, on both light and dark themes, but otherwise lgtm. great work

webui/react/src/themes.ts
hkang1: I was looking to see if we can dynamically define them with TypeScript support but I didn't have any luck. There were enough special cases that didn't follow the pattern, that I ended up going this route.

webui/react/package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">=8.0.0"
   },
   "scripts": {
```
hamidzr, line 9: the package-lock diff size doesn't match up with the changes here. are you on npm 8? 

we want the lock file on
```
  "lockfileVersion": 2,
```

hkang1: sorry my node version keeps "slipping" to v14, I'll fix this

webui/react/src/themes.ts
hkang1: discussed async, decided to pursue this in a separate body of work (assuming we can make it work w/ TypeScript, since the issue might be we are unable to handle dynamic record keys when we are defining types from records that have runtime keys being defined)

hkang1:
> the on hover effect for the navigation sidebar seems to have been lost or not be visible, on both light and dark themes, but otherwise lgtm. great work

Thanks! the nav side bar has a very faint hover effect but yeah could probably use a little more. The side bar may be going away in the near workspace-based design too.

PR #4006, opened 2022-04-25
build: set up shared-web submodule [DET-6961]

hamidzr:
## Description

- move message component as an example

## Test Plan

- expect the build to fail with a bad submodule SHA
- check that we still
  - build the stories for shared modules
  - run and include the shared module tests

## Commentary (optional)

shared-web repo: https://github.com/determined-ai/shared-web

## Checklist


webui/react/src/shared/components/message/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './Message';
+import Message from './Message';
+export default Message;
```
hkang1: nice, you got both named and default re-exported

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "webui/react/src/shared"]
+	path = webui/react/src/shared
+	url = https://github.com/determined-ai/shared-web.git
```
hkang1, line 3: looks like this is the magic for this PR

hkang1 (approved):
This PR looks good.

I have some questions about how one would do the following:
- Check out the tip of the submodule in the respective core and saas repo.
  - I think it's `git submodule init`, `git submodule sync`, `git submodule update`?
- Check the changes you want in the `shared` folder into `shared-web` repo? How do we create a PR of just the changes in `shared` folder against the tip of `shared-web` repo?

Would it help to have targets in `Makefile` or `package.json` scripts to do the steps above or should this just be left to our personal dev env setup?

hkang1:
Ok I think I worked through the flow and here are some additional steps I had to take to get the dev flows working.

### Checking out or pulling updates for submodules

1. initialize git submodules via `git submodule init` (only need to do it once in the respective repos, ie. once for core, once for repo) - (you can combine with `update` command with `--init` param as an option)
2. sync submodules via `git submodule sync`
3. update submodule via `git submodule update`

### Pushing changes to the submodule repo

1. install `gh` github CLI (`homebrew install gh` on OSX)
2. [create a github personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) and remember to copy the generated auth token from this step for use in the step below...
3. authenticate with token via `gh auth init` and choose the `Paste an authentication token` option
4. you should be able to go into `shared` directory and create a branch + push for `shared-web` changes to create a PR

webui/react/src/shared/components/message/index.ts
hamidzr: I wonder if there's a shorter syntax for rexporting the default export.

hamidzr:
> * I think it's `git submodule init`, `git submodule sync`, `git submodule update`?

`git update --init` I think you can just stick with this for most cases.

for authentication this would also be a decent option

https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf
in ~/.gitconfig
```
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
```

hamidzr:
- update the release process

webui/react/src/shared/components/message/index.ts
hkang1: I looked into this too when I saw this section, I think this is the most succinct way.

PR #4007, opened 2022-04-25
perf: process trial profiler data more efficiently DET-7050

trentwatt:
## Description

perf: process trial profiler data more efficiently [DET-7050] also [DET-7134]

## Test Plan

[This experiment](https://github.com/determined-ai/determined/files/8619756/mnist_pytorch.zip) has profiling and starts up fairly quick. do a `det e create prof.yaml .` from there and navigate to profiler page. watch data come in, play around, etc. you can fork repeatedly as well.

## Before
There are a number of problematic aspects of the existing code to process profiler data:

1. unnecessary entanglement of the different charts
2. unnecessary entanglement of the between the data and options for the charts
3. reprocessing of entire data every time new data comes in
4. unnecessary use of context (the context is only used by siblings)
5. splitting of data processing between 4 different files `Profiles/utils.tsx`, `useFetchMetrics`, `ProfileFiltersProvider`, `ProfilesEnabled` when the data is all going to one place.

 

Essentially this is what the existing profiler code is doing:
- `ProfileFiltersProvider` does a `useFetchMetrics` for each chart (Throughput, Timing, and System)
- `useFetchMetrics` calls a streaming API where the responses look like this:
```
{
  "result": {
    "batch": {
      "values": [0.0000936872, 0.0000951247, 0.0000956039],
      "batches": [1522, 1523, 1524],
      "timestamps": [
        "2022-04-23T06:58:03.960501Z",
        "2022-04-23T06:58:04.035649Z",
        "2022-04-23T06:58:04.113157Z"
      ],
      "labels": {
        "trialId": 27,
        "name": "from_device_52",
        "agentId": "orangebrain.local",
        "gpuUuid": "",
        "metricType": "PROFILER_METRIC_TYPE_TIMING"
      }
    }
  }
}
```

the ultimate goal is to get them into the `data` and `options` props for the `UPlotChart` component. 
`data` is a 2d array where all the rows have the same length, the first row are the timestamps, and each row corresponds to a value for a metric at that timestamp. if a metric does not exist for a given timestamp, then it is `null`.

as far as `options`, the main thing is that is has an attribute called `series` this an array of metadata for each metric, that is aligned with the indices in `data`

- as responses come in, a dictionary `metrics` is built of the form 
```
{
  "dataByTime": {
   "2022-04-23T06:58:03.960501Z" : { "metricA": 1, "metricB": 2 },
   "2022-04-23T06:58:04.035649Z" : { "metricA": 1, "metricC": 2 }
  },
  "names": ["metricA", "metricB", "metricC" ]
}
```

- this dictionary is returned to `ProfileFiltersProvider`, which bundles all the different metrics into a dictionary (also called) `metrics`, then puts `metrics` as the context value. recently a change was made to throttle the recreation of the dictionary, but previously it was updated every time new data came in.
- profiles enabled then does `useProfilesFilterContext` to get the metrics.
- whenever the `metrics` changes, `ProfilesEnabled` updates `data` and `options`, which are the actual values sent to the `UPlotChart` component. here's how:
  - for each metric type `profilesEnabled` calls `convertMetricsToUplotData` which sorts the data from scratch **(every data update)** and then builds the `data` prop as described above
  - after it does that it creates dummy null arrays for the data from the other charts **(again, every data update)**. this is in order to implement syncing of bounds, but it is inefficient and doesnt work.
  - it also rebuilds `options.series` for those dummy arrays based on the metrics for the other charts. this is significant because each time options is updated the `UPlotChart` component does a compare on the options to find whether it needs to recreate the chart. in general the options actually changing is rare, but it gets recreated every time data comes in so there is a lot of redundancy here.

## After

This PR takes a much more simple and efficient approach.

- it completely gets rid of the entanglement between the different charts
- `useFetchMetrics` now returns `data` in the form needed for the `UPlotChart` component.
- each chart does `useFetchMetrics` separately as its own component
- that's pretty much it

In order to achieve this we are truncating timestamps at the `1/10th` of a second (they are captured at the millisecond). As it turns out, though, the digits beyond the `1/10th` of a second are not meaningful for comparing across metrics (which is the only reason such precision would be useful in the first place).

This is because the metrics that are being captured have wildly different margins of error and interpretations for the timestamps associated with them.
 - some metrics `A` have timestamps corresponding to the moment of capture
 - some metrics `B` are an average of the last `1/10th` of a second. 
 - some metrics `C` are average of the last `1/6th` of a second

in this context we can have `A` reported at `1.105` and `B` reported at `1.110`, so it appears that `B` is after `A`, but the center of the sampling window for `B` is at `1.06`, so `B` would be more properly said to be before `A`.

in light of this and other timing concerns, reporting timestamps to the millisecond gives the end user a misleading picture of the precision of the values.

The following is the performance motivation for truncating: In general timestamps for different metrics are not aligned. so we can have things like this:

- metric 0 is timestamped at 0, 1, 2,
- metric 1 is timestamped at 0.001, 1.001, 2.001
- metric 2 is timestamped at 0.002, 1.002, 2.002

and so on till metric 1000

but then each metrics comes in separately.  so our timestamps on `data` will look like this after each step

- `[0.000, 1.000, 2.000]`
- `[0.000, 0.001, 1.000, 1.001, 2.000, 2.001]`
- `[0.000, 0.001,  0.002, 1.000, 1.001, 1.002, 2.000, 2.001, 2.002]`

In other words, we would be doing a lot of costly insertions, as well as filling our `data` array with `100x` as many `nulls` as we would if the metrics were aligned on the tenth of a second. 

in the new approach we initialize the data at the beginning according to the initial timestamp, and some buffer in case subsequent metrics come in with earlier timestamps. then every time a new batch comes in with a higher max timestamp, we add buckets for each tenth of a second between the current max and the new max to each series. 

then for each metric value it is just a matter of updating the appropriate index of the data.

## Checklist

[DET-7050]: https://determinedai.atlassian.net/browse/DET-7050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DET-7134]: https://determinedai.atlassian.net/browse/DET-7134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mapmeld (approved):
Ran the suggested experiment locally, and profiler is working. LGTM

hkang1 (approved):
LGTM!

Preview worked great. Nothing blocking, but couple of things.
- Some refactor suggestions and questions.
- A follow up ask of more test coverage, in particular for components, hooks and utility functions. If that is too large of a task, I would say focus on covering the most impactful code that is reusable code, so most likely utility functions the hooks.

PR #4009, opened 2022-04-25
feat: Cluster UI resource pool detail page [DET-7072, DET-7073]

gt2345:
## Description

Cluster page UI improvement. 

## Test Plan

Navigate to cluster page, `Historical Usage` tab should be same as before, `Master Logs` tab should be same as cluster logs page. Slot allocation bar should be same as before with removal of CPU bar. 
Screen Shot 2022-04-26 at 11 34 28 AM

Click into resource pool cards, run couple experiments, verify that `Active` `Queued` tabs work as expected.
Screen Shot 2022-04-26 at 11 35 14 AM

`Stats` tab shows average queued time by days, and the data for today is timely (updated when task is removed from queue)
Screen Shot 2022-04-26 at 11 35 27 AM


mapmeld:
I have two issues in compute-pool, one is a minor delay in the CUDA Slots allocation (showing 0/8) when I load the page, I sent you a screenshot.

Second issue is that on Active (scheduledCount, maybe queuedCount?) the number in the tab is not updating when I add an experiment or going back to 0 when all experiments complete.

Screen Shot 2022-04-26 at 12 15 53 PM

webui/react/src/pages/Clusters/utils.ts
```diff
@@ -0,0 +1,4 @@
+export interface JobQueuedTimeChartSeries {
```
mapmeld, line 1: I don't think that you're using this file or interface anymore (only other mention is commented out)

webui/react/src/components/ResourcePoolCardLight.tsx
```diff
@@ -9,20 +9,19 @@ import SlotAllocationBar from 'components/SlotAllocationBar';
...
 import { clone } from 'utils/data';
 
 import Json from './Json';
-import Link from './Link';
 import css from './ResourcePoolCardLight.module.scss';
-import ResourcePoolDetails from './ResourcePoolDetails';
 
 interface Props {
-  computeContainerStates: ResourceState[];
+  poolStats?: Api.V1RPQueueStat | undefined;
```
mapmeld, line 22: nitpick / optional - in this file you have `import * as Api from 'services/api-ts-sdk';` but only use it for `V1RPQueueStat` here. In this and the other files with `import * as Api` I'd recommend importing only what you need. It's just for clarity on what depends on what.

mapmeld:
I tried running experiments on the cluster again and the task number updates and LGTM. I made a few code comments.

My last suggestion is to add a note in `docs/release-notes` to explain that three Navigation SideBar options (Cluster, Job Queue, and Cluster Logs) are combined into the new Cluster UI.  This will make it easier for the release manager to know about this change and to describe the change to users.

webui/react/src/pages/Clusters/utils.ts
gt2345: Thank you for pointing that out, it's been fixed.

webui/react/src/components/ResourcePoolCardLight.tsx
gt2345: That surely would be better, it's been fixed.

webui/react/src/pages/ResourcepoolDetail.tsx
```diff
@@ -0,0 +1,173 @@
...
+import Json from 'components/Json';
+import Page from 'components/Page';
+import { poolLogo } from 'components/ResourcePoolCard';
+import { RenderAllocationBarResourcePool } from 'components/ResourcePoolCardLight';
+import Section from 'components/Section';
+import { useStore } from 'contexts/Store';
+import usePolling from 'hooks/usePolling';
+import { paths } from 'routes/utils';
+import { getJobQStats } from 'services/api';
+import * as Api from 'services/api-ts-sdk';
```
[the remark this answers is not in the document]
gt2345: Thank you for catching this, also a release note has been added~

mapmeld (approved):

master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -500,3 +508,39 @@ func (a *agentResourceManager) createResourcePoolSummary(
 
 	return resp, nil
 }
+
+func (a *agentResourceManager) fetchAvgQueuedTime(pool string) (
+	[]*jobv1.AggregateQueueStats, error,
+) {
+	aggregates := make([]model.ResourceAggregates, 0)
```
ioga, line 515: somewhat of a style preference, but unless you want to preallocate some known capacity (e.g. `make([]model.ResourceAggregates, 0, len(someResourceMap))`, I'd say curly bracket initialization looks nicer
```suggestion
	aggregates := []model.ResourceAggregates{}
```

master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -500,3 +508,39 @@ func (a *agentResourceManager) createResourcePoolSummary(
...
+	err := db.Bun().NewSelect().Model(&aggregates).Where(
+		"aggregation_type = ? AND aggregation_key = ? AND date >= CURRENT_TIMESTAMP - interval '30 days'",
+		"queued", pool,
```
ioga, line 518: also somewhat of a style nitpick. one of the benefits of bun is being able to do chaining, e.g. for `.Where` calls. 
in my opinion one clause per parametrized where plus one or more clause for "fixed" filters is more readable than one clause with multiple `?` and multiple parameters, because the latter leaves you figuring out which parameter goes where. i.e.
```suggestion
		Where("aggregation_type = ?", "queued").
		Where(aggregation_key = ?", pool).
		Where(date >= CURRENT_TIMESTAMP - interval '30 days'").
		Order...
```

webui/react/src/components/NavigationSideBar.tsx
```diff
@@ -113,8 +113,8 @@ const NavigationSideBar: React.FC = () => {
       { icon: 'model', label: 'Model Registry', path: paths.modelList() },
       { icon: 'tasks', label: 'Tasks', path: paths.taskList() },
       { icon: 'cluster', label: 'Cluster', path: paths.cluster() },
-      { icon: 'queue', label: 'Job Queue', path: paths.jobs() },
-      { icon: 'logs', label: 'Cluster Logs', path: paths.clusterLogs() },
+      // { icon: 'queue', label: 'Job Queue', path: paths.jobs() },
+      // { icon: 'logs', label: 'Cluster Logs', path: paths.clusterLogs() },
```
ioga, line 117: remove these?

webui/react/src/pages/Cluster/ClusterOverallStats.tsx
```diff
@@ -0,0 +1,73 @@
...
+      tally.total += rp.auxContainerCapacity;
+      tally.running += rp.auxContainersRunning;
+    });
+    return tally;
+  }, [ resourcePools ]);
+
+  const [ cudaTotalSlots, rocmTotalSlots ] = useMemo(() => {
+    return resourcePools.reduce((acc, pool) => {
+      let index;
+      switch (pool.slotType) {
```
ioga, line 28: per https://determined-ai.slack.com/archives/CLCE8D998/p1651701445031029
I believe we should treat CPU slots same as CUDA/ROCM here 
cc @hamidzr

webui/react/src/constants/states.ts
```diff
@@ -121,6 +121,7 @@ export const resourceStateToLabel: { [key in ResourceState]: string } = {
   [ResourceState.Terminated]: 'Terminated',
   [ResourceState.Pulling]: 'Pulling',
   [ResourceState.Starting]: 'Starting',
+  [ResourceState.Warm]: 'Warm',
```
ioga, line 124: I assume "Warm" is a new term we're going to use to describe these available empty slots to distinguish them from "theoretically" total available ones

webui/react/src/components/ResourcePoolCardLight.tsx
hamidzr: let's chat about which one we prefer to make sure we follow the same style. the team had a chat about this and I think we picked the other style.

webui/react/src/pages/Cluster/ClusterOverallStats.tsx
hamidzr: the change I'm looking at here might be outdated at this point but maybe bringing [this change](https://github.com/determined-ai/determined/pull/4080/files#diff-36cdee0a95644a8d46011b3c54a395f2e849b4b147817bfee1d0cb816c70ffd8R98) in would do this?

master/internal/resourcemanagers/agent_resource_manager.go
gt2345: Fixed

master/internal/resourcemanagers/agent_resource_manager.go
gt2345: Thanks! Fixed.

webui/react/src/pages/Cluster/ClusterOverallStats.tsx
gt2345: I already updated how we calculate total slots according to Hamid's change, feel free to verify and bring up any questions or concerns

gt2345: I already updated how we calculate total slots according to Hamid's change, feel free to verify and bring up any questions or concerns

webui/react/src/constants/states.ts
gt2345: Yes, warm slots are the available slots among connected agents.

gt2345: Yes, warm slots are the available slots among connected agents.

proto/src/determined/job/v1/job.proto
```diff
@@ -129,3 +129,14 @@ message QueueStats {
   // Number of scheduled jobs in the queue.
   int32 scheduled_count = 2;
 }
+
+// Aggregate statistics for a queue.
+message AggregateQueueStats {
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [ "period_start", "seconds" ] }
+  };
+  // Number of queued jobs in the queue.
```
mapmeld, line 138: Just checking on the PR. In this section, update the comments on these fields and rebuild the TypeScript bindings.

gt2345: Thank you for catching that, fixed

webui/react/src/components/SlotAllocationBar.tsx
```diff
@@ -140,30 +147,84 @@ const SlotAllocationBar: React.FC<Props> = ({
     return [ parts.running, parts.pending, parts.free, parts.potential ];
   }, [ totalSlots, stateTallies, pendingSlots, freeSlots, slotsPotential, footer, isAux ]);
 
-  const stateDetails = useMemo(() => {
-    const states = [
+  const totalSlotsNum = useMemo(() => {
+    return slotsPotential && slotsPotential > 0 ? slotsPotential : totalSlots;
```
hamidzr, line 151: nit slotsPotential || totalSlots

hamidzr (commented):
the new page looks great.

- separating queued and active jobs: (I should have caught this earlier) what's the motivation behind separating active and queued jobs? this mostly limits the information we show to the user. There are cases where some queued jobs are ahead of some scheduled jobs and with the two separated we can't communicate this to the user.

- 'scheduled' instead of 'active'?

hamidzr:
> the new page looks great.
> 
> * separating queued and active jobs: (I should have caught this earlier) what's the motivation behind separating active and queued jobs? this mostly limits the information we show to the user. There are cases where some queued jobs are ahead of some scheduled jobs and with the two separated we can't communicate this to the user.
> * 'scheduled' instead of 'active'?

cc @shiyuann

webui/react/src/components/SlotAllocationBar.tsx
gt2345: Thanks

PR #4011, opened 2022-04-26
fix: Ensure table offset does not exceed pagination total [DET-6829]

julian-determined-ai:
## Description

fix: Ensure table offset does not exceed pagination total [DET-6829]

Currently, filtering can cause the ExperimentList table to show "no data" because the table pagination offset may exceed the total number of data points. For example if there are only 20 data points, but the pagination offset is 100, the table will show "no data". 

The following fix addresses the issue by subscribing to pagination changes, most importantly the value of "total". If the offset is determined to be great than the total, then the table is updated to show the correct last page of the current data set. An example can be seen below. 

https://user-images.githubusercontent.com/103522725/165338395-172bef94-d72b-44f4-a624-296c154833eb.mov

## Test Plan

1. Reset any applied filters on the Experiments List table or apply a filter with a larger data set than the filter in step (3). 
2. Navigate to the last page of the data set in the table. 
3. Apply a filter that will result in a smaller data set. 
4. Ensure that the table shows the last page of the new filtered data set.  
5. Apply different filters and page limits in the experiments table and ensure that navigating to the last page in a large data set, then filtering to a smaller data set always results in being shown the last page in the smaller data set. 

## Commentary (optional)

Sometimes the "no data" state is shown for a short amount of time, less than a second or so. Due to the the `isLoading` state being set to `false` during the `fetchExperiments` call. The result of that call is the `total` being updated, and the `total` update will trigger this hook to run and re-render the table. This could potentially be avoided by updating how and when the `isLoading` state is set to `false` in the "fetchExperiments" call but I did not make any changes in the PR. If we find that this leads to a poor experience for users then this can be revisited.    

## Checklist

[DET-6829]: https://determinedai.atlassian.net/browse/DET-6829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -678,10 +678,8 @@ const ExperimentList: React.FC = () => {
 
   useEffect(() => {
     if(settings.tableOffset > total){
-      setIsLoading(true);
       const offset = settings.tableLimit * Math.floor(total / settings.tableLimit);
       updateSettings({ tableOffset: offset });
-      setIsLoading(false);
     }
   }, [ total, settings.tableOffset, settings.tableLimit, updateSettings ]);
 
--- a/webui/react/src/pages/ExperimentList.tsx
+++ b/webui/react/src/pages/ExperimentList.tsx
@@ -676,6 +676,15 @@ const ExperimentList: React.FC = () => {
     return () => canceler.abort();
   }, [ canceler ]);
 
+  useEffect(() => {
+    if(settings.tableOffset > total){
+      setIsLoading(true);
+      const offset = settings.tableLimit * Math.floor(total / settings.tableLimit);
+      updateSettings({ tableOffset: offset });
+      setIsLoading(false);
+    }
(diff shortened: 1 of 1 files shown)


trentwatt (approved):
Looks like you can make it so where the loading spinner is going when there is no data by moving this logic [here](https://github.com/determined-ai/determined/blob/ea4ab88540bd826e8eaf7138e75d18d677c5c19c/webui/react/src/pages/ExperimentList.tsx#L157), and reduce the number of `useEffect`s in the process.

That said, I think the place this kind of fix most properly belongs is at the component level, so [here](https://github.com/determined-ai/determined/blob/ea4ab88540bd826e8eaf7138e75d18d677c5c19c/webui/react/src/components/InteractiveTable.tsx#L358). you can access `total` as `tablePagination.total` there. I tried messing around with it for a bit but didn't get anywhere. 

You can try if you want.

PR #4015, opened 2022-04-26
fix: Prevent archived models from appearing in the Register Checkpoint modal [DET-7132]

mapmeld:
## Description

After a model is archived, it cannot add new Model Versions, so it should not be in the list of selectable models in the Register Checkpoint / new Model Version modal window.

In the current version the model appears, and trying to register a new model version throws an error. 

## Test Plan

- Archive a model
- Go to an experiment with checkpoints
- Check list of models available when you register a checkpoint

## Checklist


EmilyBonar (approved):
LGTM

PR #4016, opened 2022-04-26
fix: remove dead shell start code [DET-7131]

varlogtim:
Tested manually.

To test, see description in DET-7131


ioga (approved):
lgtm, but as discussed needs confirmation on windows and k8s.

varlogtim:
@mpkouznetsov Could you verify the change to here: e2e_tests/tests/cluster/test_users.py

PR #4017, opened 2022-04-26
fix: Human-readable option for empty filters in logs [DET-6781, DET-6999]

mapmeld:
## Description

On the task logs page, the filter UI includes options for allocation, agent, container, rank, and levels.  Levels is constant, but the others are filled in by the database from all options available in the task logs.

When there is 0-1 option in a dropdown, `moreThanOne` already hides this dropdown (only Levels and Container dropdowns visible below).

Screen Shot 2022-04-26 at 10 53 26 AM

When there is an empty string value, the containers dropdown already lists this option as "No Container". This PR adds "No Agent", "No Rank", and "No Allocation" to the other dynamic dropdowns in place of any empty string value.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer/LogViewerFilters.tsx
+++ b/webui/react/src/components/LogViewer/LogViewerFilters.tsx
@@ -90,7 +90,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
             value={values.allocationIds}
             onChange={handleChange('allocationIds', String)}>
             {selectOptions?.allocationIds?.map(id => (
-              <Option key={id} value={id}>{id || 'No Allocation'}</Option>
+              <Option key={id} value={id}>{id || 'No Allocation ID'}</Option>
             ))}
           </MultiSelect>
         )}
@@ -100,7 +100,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
             value={values.agentIds}
             onChange={handleChange('agentIds', String)}>
             {selectOptions?.agentIds?.map(id => (
-              <Option key={id} value={id}>{id || 'No Agent'}</Option>
+              <Option key={id} value={id}>{id || 'No Agent ID'}</Option>
             ))}
           </MultiSelect>
         )}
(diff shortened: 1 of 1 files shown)


webui/react/src/components/LogViewer/LogViewerFilters.tsx
```diff
@@ -89,15 +89,19 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values
             itemName={LABELS.allocationIds}
             value={values.allocationIds}
             onChange={handleChange('allocationIds', String)}>
-            {selectOptions?.allocationIds?.map(id => <Option key={id} value={id}>{id}</Option>)}
+            {selectOptions?.allocationIds?.map(id => (
+              <Option key={id} value={id}>{id || 'No Allocation'}</Option>
```
stoksc, line 93: It's not so much "no allocation" as it is "no allocation _filters_"

stoksc: @mapmeld I think this reasoning applies to every filter

mapmeld: I should have asked about this before changing -
when you have a container, the options are "All Containers", "No Container", or [container-id]. Assuming that's true of the other filters, the "No _" does not mean to remove filters (as with All), but to select unassociated logs such as

```
[2022-04-26T20:17:11.219319Z] || INFO: Scheduling TensorBoard (really-casual-badger) (id: a770396c-cf53-4aae-bded-4be9eb02378f) <info> [2022-04-26 20:17:11]
[2022-04-26T20:17:11.745773Z] || INFO: TensorBoard (really-casual-badger) was assigned to an agent <info> [2022-04-26 20:17:32]
[2022-04-26T20:17:32.258493Z] || INFO: Service of TensorBoard (really-casual-badger) is available 
```

hkang1 (approved):
the code looks good! Just need to come to a consensus of the wording. I want to keep the wording as short as possible but ensure it is relatively easy for the human :) to be able to figure out what the option means.

webui/react/src/components/LogViewer/LogViewerFilters.tsx
stoksc: Sure, that's fine. "No Allocation Id" meaning "logs on which there is no allocation id".

mapmeld: ok I will update all 4, only one which sounds odd to me is No Rank ID (?)

hkang1: ah good point, for that one just `No Rank`?

PR #4020, opened 2022-04-26
feat: Telemetry of use of workspaces and projects

mapmeld:
## Description

Include our planned telemetry on workspace and project features:
- total count of workspaces and projects
- average projects-per-workspace and average experiments-per-project (excepting those in the default workspace + project)

Uses `::float` for averages, and `greatest(count(*), 1)` to avoid dividing by 0. You cannot make new projects in the default workspace, so there are no user-created projects in this situation.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_workspace_org.py
+++ b/e2e_tests/tests/cluster/test_workspace_org.py
@@ -287,7 +287,7 @@ def test_workspace_org() -> None:
         returned_notes = r5.notes
         assert len(returned_notes) == 2
 
-        # Create an experiment in the default project and move to a test project.
+        # Create an experiment in the default project.
         test_exp_id = run_basic_test(
             conf.fixtures_path("no_op/single.yaml"), conf.fixtures_path("no_op"), 1
         )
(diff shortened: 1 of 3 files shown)


EmilyBonar (approved):
LGTM

e2e_tests/tests/cluster/test_workspace_org.py
```diff
@@ -296,13 +296,17 @@ def test_workspace_org() -> None:
...
+        exp_count = len(bindings.get_GetProjectExperiments(sess, id=made_project.id).experiments)
+        assert exp_count == 0
         mbody = bindings.v1MoveExperimentRequest(
             destinationProjectId=made_project.id, experimentId=test_exp_id
         )
         bindings.post_MoveExperiment(sess, experimentId=test_exp_id, body=mbody)
         modified_exp = bindings.get_GetExperiment(sess, experimentId=test_exp_id).experiment
         assert modified_exp.projectId == made_project.id
+        exp_count = len(bindings.get_GetProjectExperiments(sess, id=made_project.id).experiments)
+        assert exp_count == 1
```
EmilyBonar, line 309: It might also be good to assert that the moved project is no longer present in the uncategorized list - that's the issue I was running into.

mapmeld: ok, updated the test

PR #4022, opened 2022-04-26
chore: add byUserId to aggregated allocation API response

johnkim-det:
## Description

for the `api/v1/resources/allocation/aggregated` API endpoint, return `byUserId` in addition to `byUsername`

## Test Plan

- [ ] make sure API response includes both `byUserId` and `byUsername`

## Commentary (optional)

Will most likely also remove `byUsername` later, as part of DET-6923

## Checklist


hamidzr:
how does jira find this without the ticket number being mentioned? is it the branch name? https://determinedai.atlassian.net/browse/DET-7137

master/internal/core.go
```diff
@@ -382,6 +382,9 @@ func (m *Master) getAggregatedResourceAllocation(c echo.Context) error {
 		if err = writeAggType("username", entry.ByUsername); err != nil {
 			return err
 		}
+		if err = writeAggType("user_id", entry.ByUserId); err != nil {
```
hamidzr, line 385: assuming this part works the code changes lgtm

hamidzr (approved):

master/internal/core.go
johnkim-det: actually it doesn't work -- i was using `map<int32, float>` for `by_user_id` in `master.proto`, and `map[string]float32` is expected in `core.go`. decided to switch that to `map<string, float>`, please lmk if that doesn't seem like a good solution.

PR #4024, opened 2022-04-27
feat: save metadata pre-upload

mpkouznetsov:
## Description

A less ambitious change: save metadata to `metadata.json` as part of upload.

Initially (https://github.com/determined-ai/determined/pull/4010) I tried to push `metadata.json` into the checkpoint storage as part of `Checkpoint.add_metadata` and `Checkpoint.remove_metadata`. That did not work as the client typically does not have write access to the checkpoint storage directory in the case of shared_fs.

Instead we still save metadata into the local directory just before uploading it but only in `CheckpointContext`.
The good news: metadata is saved to checkpoint storage as part of `CheckpointContext.upload` and `CheckpointContext.store_path` (in addition to being saved to DB).
The sad news: if the user mutates metadata locally with python SDK (outside of Contexts/experiment execution) those updates only get to the database but not to the checkpoint storage.

## Docstring change
Note the "Warning" line.
image

### DIFF
--- a/harness/determined/core/_checkpoint.py
+++ b/harness/determined/core/_checkpoint.py
@@ -222,8 +222,8 @@ class CheckpointContext:
 
     def _write_metadata_file(self, ckpt_dir: str, metadata: Optional[Dict[str, Any]]) -> None:
         if metadata is not None:
-            metadata_path = str(pathlib.Path(ckpt_dir).joinpath("metadata.json"))
-            with open(metadata_path, "w") as f:
+            metadata_path = pathlib.Path(ckpt_dir).joinpath("metadata.json")
+            with metadata_path.open("w") as f:
                 json.dump(metadata, f, indent=2)
 
     def _report_checkpoint(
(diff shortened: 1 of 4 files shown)


harness/determined/core/_checkpoint.py
```diff
@@ -214,6 +220,12 @@ def delete(self, storage_id: str) -> None:
         """
         self._storage_manager.delete(storage_id)
 
+    def _write_metadata_file(self, ckpt_dir: str, metadata: Optional[Dict[str, Any]]) -> None:
+        if metadata is not None:
+            metadata_path = str(pathlib.Path(ckpt_dir).joinpath("metadata.json"))
+            with open(metadata_path, "w") as f:
```
ioga, line 226:
```suggestion
            metadata_path = pathlib.Path(ckpt_dir).joinpath("metadata.json")
            with metadata_path.open("w") as f:
```

ioga (approved):
lgtm

harness/determined/core/_checkpoint.py
mpkouznetsov: 🙇‍♂️

PR #4030, opened 2022-04-28
fix: avoid null allocation id in checkpoint view

mpkouznetsov:
## Description

Testing upgrade (creating a bunch of experiments with 0.17.6, upgrading to the current version, then accessing existing experiments, trials, checkpoints).

Error:
```
time="2022-04-28T03:27:32Z" level=error msg="finished unary call with code Internal" error="rpc error: code = Internal desc = failed to create experiment: failed to get checkpoint for source trial 162: error querying for latest trial checkpoint (162): sql: Scan error on column index 3, name \"allocation_id\": converting NULL to string is unsupported" grpc.code=Internal grpc.method=CreateExperiment grpc.service=determined.api.v1.Determined grpc.start_time="2022-04-28T03:27:32Z" grpc.time_ms=59.749 span.kind=server system=grpc
```

when trying to continue trial created with 0.17.6 after upgrading to the current version.

## Checklist

### DIFF
--- a/master/static/migrations/20220423162505_generic-checkpoints.tx.up.sql
+++ b/master/static/migrations/20220423162505_generic-checkpoints.tx.up.sql
@@ -19,7 +19,12 @@ CREATE OR REPLACE VIEW public.checkpoints_view AS
         c.id AS id,
         c.uuid AS uuid,
         t.task_id,
-        t.task_id||'.'||t.run_id as allocation_id,
+        CASE
+        WHEN t.task_id is NULL THEN
+            NULL
+        ELSE
+            t.task_id || '.' || c.trial_run_id
+        END allocation_id,
         c.end_time as report_time,
         c.state,
         c.resources,
--- a/master/static/migrations/20220423162505_generic-checkpoints.tx.up.sql
+++ b/master/static/migrations/20220423162505_generic-checkpoints.tx.up.sql
@@ -19,7 +19,7 @@ CREATE OR REPLACE VIEW public.checkpoints_view AS
         c.id AS id,
         c.uuid AS uuid,
         t.task_id,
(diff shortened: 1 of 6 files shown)


master/static/migrations/20220423162505_generic-checkpoints.tx.up.sql
```diff
@@ -19,7 +19,7 @@ CREATE OR REPLACE VIEW public.checkpoints_view AS
         c.id AS id,
         c.uuid AS uuid,
         t.task_id,
-        NULL as allocation_id, -- TODO, is there any trickery to deduce allocation_id?
+        t.task_id||'.'||t.run_id as allocation_id,
```
stoksc, line 22: I don't think this is right. this would associated every checkpoint with the last run

stoksc: I think it is truly null for old enough experiments

mpkouznetsov: is `raw_checkpoints.trial_run_id` the right thing to use then?

mpkouznetsov: Reflecting our Slack conversation:
`raw_checkpoints.trial_run_id` _is_ there right thing to use here. However, older checkpoints may still have trouble loading because tasks and allocations did not exist until a certain release. So task_id and allocation_id would conceivably still be null. We decided I would look into the repercussions of making those nullable fields.

mpkouznetsov:
Thing still work when I make task_id and allocation_id optional, surprisingly.
@stoksc and I decided that it is better to keep them optional at this level than filling them in with sentinel values.

stoksc (approved):

PR #4034, opened 2022-04-28
chore: Refactor CheckpointModal to be hook based [DET-7136]

julian-determined-ai:
## Description

chore: Refactor CheckpointModal to be hook based [DET-7136]

The `CheckpointModal` is now `useModalCheckpoint` and is implemented inline with our other hook based modals. 

## Test Plan

1. Clicking on the "Best Checkpoint" area on the experiment overview page should show the checkpoint modal as it did before.
2. Clicking on the "Checkpoint" button (the flag icon) in the Workloads area of the same page as above, should show the checkpoint modal as it did before.
3. Clicking "Register Checkpoint" on the checkpoint modal should work as it does today.

## Commentary (optional)

While attempting to create a storybook entry, I was unable to render showing the modal do to the `useStore` call in the `useCreateModelModal` child component needing a provider. I attempted to add a Provider as a decorator in Storybook but that did not solve the issue. Any insights would be appreciated! For now the storybook renders a disabled button. 

## Checklist


webui/react/src/components/CheckpointDetails.tsx
```diff
@@ -0,0 +1,33 @@
...
+    ...props
+  }: PropsWithChildren<Props>,
+) => {
+  const { modalOpen: openModalDelete } =
+  useModalCheckpoint({ checkpoint: checkpoint, config: experiment.config, title: title });
+  const handleModalCheckpointClick = useCallback(() => openModalDelete(), [ openModalDelete ]);
+
+  return (
+    <span onClick={() => handleModalCheckpointClick()}>
+      {props.children}
```
hkang1, line 28: **suggestion:** what do you think about having the default content defined as...
```
<Tooltip title="View Checkpoint">
  <Button
    aria-label="View Checkpoint"
    icon={<Icon name="checkpoint" />}
  />
</Tooltip>
```
if `props.children` is undefined?

webui/react/src/components/CheckpointDetails.tsx
```diff
@@ -0,0 +1,33 @@
...
+  experiment: ExperimentBase;
+  title: string;
+}
+
+const CheckpointDetails: React.FC<Props> = (
+  {
+    checkpoint,
+    experiment,
+    title,
+    ...props
```
hkang1, line 19: **nit:** could probably just specify `children` since it's the only `props` being used

webui/react/src/components/CheckpointDetails.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const handleModalCheckpointClick = useCallback(() => openModalDelete(), [ openModalDelete ]);
+
+  return (
+    <span onClick={() => handleModalCheckpointClick()}>
+      {props.children}
+    </span>
+  );
+};
+
+export default CheckpointDetails;
```
hkang1, line 33: **blocking question:** so I'm not sure if this component is needed. What if you used the `useModalCheckpoint` directly where you need it?
```
const { modalOpen: openModalDelete } =
  useModalCheckpoint({ checkpoint: checkpoint, config: experiment.config, title: title });

// somewhere in the JSX
<SomeComponent onClick={openModalDelete} />
```

webui/react/src/components/OverviewStats.tsx
```diff
@@ -3,14 +3,15 @@ import React, { PropsWithChildren } from 'react';
 import css from './OverviewStats.module.scss';
 
 interface Props {
+  clickable?: boolean
```
hkang1, line 6: just noting that this change would not be needed if we consumed the hook directly

webui/react/src/hooks/useModal/useModalCheckpoint.tsx
```diff
@@ -69,23 +69,10 @@ const renderResource = (resource: string, size: string): React.ReactNode => {
...
-const CheckpointModal: React.FC<Props> = (
-  { config, checkpoint, onHide, show, title, ...props }: Props,
-) => {
-  const { showModal: showCreateModelModal } = useCreateModelModal();
-  const state = checkpoint.state as unknown as RunState;
-
-  const totalSize = useMemo(() => {
-    return humanReadableBytes(checkpointSize(checkpoint));
-  }, [ checkpoint ]);
+const useModalCheckpoint = ({ checkpoint, config, title, ...props }: Props): ModalHooks => {
```
hkang1, line 72: nice looking hook! 👍

hkang1 (commented):
This PR looks great, especially the hook itself! I tried it out on netlify and it works great. I just have the one blocking question of if we can get away of just consuming the hook directly wherever needed, to avoid having to wrap a component where it's being used.

webui/react/src/components/CheckpointDetails.tsx
julian-determined-ai: I have made this update.

webui/react/src/components/CheckpointDetails.tsx
julian-determined-ai: Initially I had implemented the hook this way but the implementation did not work as intended due to how the `checkpoint` state is being set. For example, in the current `ExperimentTrials` the `activeCheckpoint` is set [as a result of a button click](https://github.com/determined-ai/determined/blob/ad06cd43145725c5f1ea52ed36254bd6035f5fb1/webui/react/src/pages/ExperimentDetails/ExperimentTrials.tsx#L52). 

Therefore on initial page load, using the modal such as `useModalCheckpoint({ checkpoint: checkpoint, config: experiment.config, title: title });` would initialize the hook with a `checkpoint` that was undefined. And clicking on the the button that would reveal the modal actual would not render as you would expect since the the `activeCheckpoint` would not be set until after the modal has shown, due to the state not having been updated by the time the hook was called. 

The other pages using the `CheckpointModal` follow a similar pattern, where the `checkpoint` being used would not necessarily be set by the time the `useModalCheckpoint` was called. The choice to use an extra component was made to resolve this issue. 

With the current implementation, the hook will always be called with the correct checkpoint since the `CheckpointDetails` will always be rendered with the appropriate `checkpoint`. 

Let me know if that explanation makes sense and sounds reasonable.

hkang1: Ah I see, so I guess `CheckpointDetails` acts as the modal trigger. So couple of things:

**suggestion:** This is more of a pattern discussion and FYI for future modals too. If there are things you want to pass to the hook but aren't read during hook definition, e.g. `checkpoint`, then you can move it from a prop over to `openOrUpdate` as a function param instead. So when you are ready to open the modal, the variable dependency will be available then.

**nit suggestion:** thoughts on renaming `CheckpointDetails` to be `CheckpointModalTrigger` to be a bit more specific for what it is doing?

julian-determined-ai: The naming update to `CheckpointModalTrigger` has been made.

webui/react/src/storybook/shared/generateTestExperiments.ts
```diff
@@ -0,0 +1,221 @@
+import { CheckpointDetail,
```
hkang1, line 1: 👍

hkang1 (approved):
This looks great! Before you land this, can you write tests for `CheckpointModalTrigger.tsx` and `useModalCheckpoint.tsx` 🙏 ?

PR #4036, opened 2022-04-29
chore: Recommend git clone --recursive for submodules

mapmeld:
## Description

Update our docs where we `git clone` to include the new submodules with `git clone --recursive`

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ Base AMI (Ubuntu)".
 Start by cloning the Determined repo:
 
 ```sh
-git clone git@github.com:determined-ai/determined.git
+git clone --recursive git@github.com:determined-ai/determined.git
 ```
 
 #### Prerequisites
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ Base AMI (Ubuntu)".
 Start by cloning the Determined repo:
 
 ```sh
-git clone --recursive git@github.com:determined-ai/determined.git
+git clone --recurse-submodules git@github.com:determined-ai/determined.git
 ```
 
 #### Prerequisites
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ det deploy local cluster-up --no-gpu
 # Access web UI at localhost:8080. By default, "determined" user accepts a blank password.
 
 # Navigate to a Determined example.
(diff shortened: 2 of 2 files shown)


hamidzr:
thanks for updating the docs. don't we want this one instead? `--recursive` doesn't show up in `git clone --help`

```
--recurse-submodules[=<pathspec>]
             After the clone is created, initialize and clone
             submodules within based on the provided pathspec. If no
             pathspec is provided, all submodules are initialized
             and cloned. This option can be given multiple times for
             pathspecs consisting of multiple entries. The resulting
             clone has submodule.active set to the provided
             pathspec, or "." (meaning all submodules) if no
             pathspec is provided.

             Submodules are initialized and cloned using their
             default settings. This is equivalent to running git
             submodule update --init --recursive <pathspec>
             immediately after the clone is finished. This option is
             ignored if the cloned repository does not have a
             worktree/checkout (i.e. if any of --no-checkout/-n,
             --bare, or --mirror is given)

```

hamidzr (commented):
thanks for updating the docs. what does --recursive do? we might want this one  instead `--recursive` doesn't show up in `git clone --help`

```
--recurse-submodules[=<pathspec>]
             After the clone is created, initialize and clone
             submodules within based on the provided pathspec. If no
             pathspec is provided, all submodules are initialized
             and cloned. This option can be given multiple times for
             pathspecs consisting of multiple entries. The resulting
             clone has submodule.active set to the provided
             pathspec, or "." (meaning all submodules) if no
             pathspec is provided.

             Submodules are initialized and cloned using their
             default settings. This is equivalent to running git
             submodule update --init --recursive <pathspec>
             immediately after the clone is finished. This option is
             ignored if the cloned repository does not have a
             worktree/checkout (i.e. if any of --no-checkout/-n,
             --bare, or --mirror is given)

```

hamidzr:
I'm considering whether we want to add a `git submodule update` to our `get-deps` step.

mapmeld:
oh I had used `--recursive` before, it looks like it's only an alias https://explainshell.com/explain?cmd=git+clone+--recursive

I'll update this PR and then if you think it should combine in the `git submodule update` I can look it up

hamidzr:
thanks.
re 
> I'll update this PR and then if you think it should combine in the git submodule update I can look it up

i'll try this in saas and see how that works out first.

hamidzr (approved):

PR #4038, opened 2022-04-29
fix: Alter Boolean arg default handling

### LINKED ISSUE #4037: boolean arg help text is confusing
The way defaults for boolean CLI parameters are created produces potentially confusing help text.  For example, running `det deploy local cluster-up --help` returns this:
```
  --gpu                 enable GPU support for agent (default: True)
  --no-gpu              disable GPU support for agent (default: True)
  --no-autorestart      disable container auto-restart (recommended for local
                        development) (default: False)
```

I suspect that the intent is for a default of "True" to convey that `--gpu` defaults to "specified" from the perspective of the command line, while `--no-gpu` actually defaults to "not specified."  However, things are slightly confused by the `--no-autorestart` flag, which double-negative defaults to False.  Does that mean autorestart is actually on by default?  Or would it follow the above model, where no-autorestart would actually be the default behavior if `False` was the default.  Ok, that one is a bad example, but user confusion still potentially ensues with the default value sometimes referring to what's effectively an internal use rather than "was the flag specified on CLI or not." :)

It'd be nice if the complementary boolean args could just print all on one line, as `--[no-]gpu` or something like that, but I don't think there's an easy way to do that in legacy argparse.  One could add a third literal argument combining the two, I suppose, then suppress help output for the "real" args, but that feels pretty gross.  That also breaks the existing "provide true and false help text" interface, requiring more changes across the code.  I'll open a PR with a suggestion which makes me feel slightly less gross in a second.

dannysauer:
## Description

Adjust the default value when adding Boolean args for improved help output clarity.
Closes #4037

## Test Plan

The `make test` is misbehaving on this old machine, but here's sketchy PoC output in the interim:
```python
Python 3.6.12 (default, Dec 02 2020, 09:44:23) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import argparse
>>> p = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
>>> k = False
>>> out = None
>>> p.add_argument("--k", dest="out", action="store_true", help="enable k", default=k)
_StoreTrueAction(option_strings=['--k'], dest='out', nargs=0, const=True, default=False, type=None, choices=None, help='enable k', metavar=None)
>>> p.add_argument("--no-k", dest="out", action="store_false", help="disable k", default=argparse.SUPPRESS)
_StoreFalseAction(option_strings=['--no-k'], dest='out', nargs=0, const=False, default='==SUPPRESS==', type=None, choices=None, help='disable k', metavar=None)
>>> p.parse_args()
Namespace(out=False)
>>> p.parse_args(["--k"])
Namespace(out=True)
>>> p.parse_args(["--no-k"])
Namespace(out=False)
>>> p.print_help()
usage: [-h] [--k] [--no-k]

optional arguments:
  -h, --help  show this help message and exit
  --k         enable k (default: False)
  --no-k      disable k
```

## Commentary (optional)

First: I'm not yet quite comfortable enough with the codebase to be sure this is all done the happiest way possible, so feedback is very much encouraged. I just set this all up for the first time this morning. 😅  So, "Draft PR" FTW.

With python versions 3.9 and newer, argparse can automatically create the `--no-x` option and resolve this whole mess (using `add_argument('--someflag', action=argparse.BooleanOptionalAction)`), but presumably the intent is to maintain compatibility with earlier python versions.

In those earlier versions, setting the default of the false parameter to `argparse.SUPPRESS` suppresses output of a default in the help text without breaking the dest's default value as set in a previous call.  I learned while playing with this that `None` does not suppress output (it prints `default: None`). 🤷 By only showing dest's default on the "positive" argument, the end user doesn't have to do any mental gymnastics with double negatives or knowing the code's internals and, IMO, ends up less confused.

It appears (from [the argparse code](https://github.com/python/cpython/blob/11652ceccf1dbf9dd332ad52ac9bd41b4adff274/Lib/argparse.py#L1844-L1849)) that the first default added for a given destination is the one used for that destination, so defaults on subsequent arguments are redundant as far as assigning a value to the dest (same code block linked above), but still matter to the output formatter.  So, by adding the true argument first with the desired default value, this all continues working as expected.  See sketchy proof of concept output above.

## Checklist

### DIFF
--- a/harness/determined/common/declarative_argparse.py
+++ b/harness/determined/common/declarative_argparse.py
@@ -201,12 +201,19 @@ def add_args(parser: ArgumentParser, description: List[Any], depth: int = 0) -> 
 
         elif isinstance(thing, BoolOptArg):
             parser.add_argument(
-                thing.true_name, dest=thing.dest, action="store_true", help=thing.true_help
+                thing.true_name,
+                dest=thing.dest,
+                action="store_true",
+                help=thing.true_help,
+                default=thing.default,
             )
             parser.add_argument(
-                thing.false_name, dest=thing.dest, action="store_false", help=thing.false_help
+                thing.false_name,
+                dest=thing.dest,
+                action="store_false",
+                help=thing.false_help,
+                default=SUPPRESS,
             )
-            parser.set_defaults(**{thing.dest: thing.default})
 
     # If there are any subcommands but none claimed the default action, make
     # the default print help.


ioga:
Danny, welcome to the team!

1. I did not know about `default=argparse.SUPPRESS`, this is cool, and seems to solve the problem.
2. >presumably the intent is to maintain compatibility with earlier python versions.

That's correct, we claim to support python 3.7+ at the moment.

3. The circleci checks report a python lint failure that looks legit. Running `make fmt` in `harness` should reformat the code the way linter wants and address it. Can you please do that?
4. `test-intg` failure seems to be a flake 🤔 , ignore it for now.
5. Let me ask around about the CLA requirement.
6. There was no need to open a github issue for that, the PR is already sufficiently descriptive.

ioga:
@cla-bot[bot] check

ioga:
okay, we should be good on the CLA.

one more thing: under semantic PR titles, this is probably a `fix: ` cause it's a material fix to the code, not just a `style: ` change.

dannysauer:
I always waffle on bug v/s style for little stuff like this - but "fix" makes me feel better about the branch name. ;)

So I guess that means there should be release notes as well?

ioga (approved):
lgtm. this is probably minor enough to not require a release note.

PR #4043, opened 2022-05-02
refactor: simplify apiConfig to reduce redundancy

hkang1:
## Description

The `apiConfig.ts` is a common area for making a mistake because it required two api objects to be updated when there is a new api module to consume. The 2nd one is really easy to miss, so this update reduces it to one area to update.

## Test Plan

- [ ] general check that pages load and API calls work.
- [ ] Master logs would be good to just do a quick check.

## Checklist


EmilyBonar (approved):
LGTM! Thanks for simplifying this!

PR #4049, opened 2022-05-02
chore: remove unnecessary use of username in webui (DET-6922)

johnkim-det:
## Description

Changing to use `userId` whenever appropriate.

usage of `username` should be limited to:
- authentication/login
- creating sample users for testing
- displaying all user info (`UserSettings` modal)
- `getDisplayName` util function
- resource allocation charts (to be handled separately in DET-7137)

all other usage of username should be replaced with `userId` and `getDisplayName` with `users` in the store.

## Test Plan

Could test everything that involves user data, or just make sure changes pass regression testing.

## Checklist


webui/react/src/hooks/useCreateModelModal.tsx
```diff
@@ -80,7 +80,6 @@ const useCreateModelModal = (): ModalHooks => {
         labels: tags,
         metadata: metadata,
         name: modelName,
-        username: user?.username,
```
hkang1, line 83: **question:** just checking to see if is ok to create a model without a user id. Does it figure it out on the backend based on the auth token of who is creating a model?

webui/react/src/hooks/useTelemetry.ts
```diff
@@ -82,7 +82,7 @@ class Telemetry {
           clusterName: info.clusterName,
           edition: 'OSS',
           masterId: info.masterId,
-          username: auth.user.username,
+          userId: auth.user.id,
```
hkang1, line 85: **chore:** for segment traits, the user ID trait key is `id` according to [this](https://segment.com/docs/connections/spec/identify/#traits)

webui/react/src/utils/user.ts
```diff
@@ -1,5 +1,10 @@
 import { DetailedUser, User } from 'types';
 
-export function getDisplayName (user: DetailedUser | User | undefined): string {
+interface UserNameFields {
+  displayName?: string;
+  username: string;
```
hkang1, line 5: **question:** should `username` be optional?

hkang1 (approved):
LGTM! I have a few questions, posted, but nothing seems to be blocking, one chore somewhere in there.

webui/react/src/hooks/useCreateModelModal.tsx
johnkim-det: yep, it checks `CurrentUser`

webui/react/src/hooks/useTelemetry.ts
johnkim-det: reverting changes to this file

webui/react/src/utils/user.ts
johnkim-det: yep missed this, thanks

PR #4050, opened 2022-05-02
fix: When a user creates a workspace, it is pinned by that user

mapmeld:
## Description

Users will now automatically add a pin to their created workspaces. This also updates e2e_tests

## Checklist


EmilyBonar (approved):
LGTM as long as the tests pass

PR #4051, opened 2022-05-02
fix: show notification when delete experiment fail [DET-6811]

gt2345:
## Description

When deleting experiment from experiment list and the action fails from server side error, fix notification to pop. 

## Test Plan

- Verify delete experiment function works as expected at experiment list page and experiment detail page. 
- Create server side error for `func (a *apiServer) DeleteExperiment` and verify delete experiment pops notification as expected. 
- As suggested by Trent, another way to create server side error during testing is to disconnect WIFI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "webui/react/src/shared"]
 	path = webui/react/src/shared
 	url = https://github.com/determined-ai/shared-web.git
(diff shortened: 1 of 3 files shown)


trentwatt (approved):
Nice, LGTM.

One easy way to create an error for experiment delete is to disconnect WifI, might be worth mentioning for test plan.

PR #4052, opened 2022-05-02
chore: remove filtering by username (DET-7095)

johnkim-det:
## Description

Remove the ability to filter Experiments, Models, and Tasks by username. Should only be able to filter by userId.

This PR is dependent on https://github.com/determined-ai/determined/pull/4049.

## Test Plan

Regression testing to make sure these endpoints still work, with filtering by userId
- [ ] Get all experiments
- [ ] Get all models
- [ ] Get all commands
- [ ] Get all notebooks
- [ ] Get all shells
- [ ] Get all tensorboards

## Commentary (optional)

This was initially done as a non-breaking change that allowed filtering by either username or userId. With further discussion among the web team, it was decided that properly addressing the tech debt of relying on usernames as identifiers requires removing the username filtering, even if it is a breaking change.

## Checklist


harness/determined/common/api/authentication.py
```diff
@@ -81,7 +82,9 @@ def _init_session(
             token = util.get_container_user_token()
 
         if token is not None:
-            return Session(session_user, token)
+            return Session(
+                session_user, token, _get_current_user_id(self.master_address, token, cert)
```
stoksc, line 86: seems like a sort of unfortunate new round trip on every start up. though at least when we do a login we save the trip (nice). only way I see to get around it would also be to store the user id in the token store (and inject it for task containers somehow), which is kind of a pain.

stoksc (approved):
lgtm (had some non-blocking comments, just to chat)

harness/determined/common/api/authentication.py
ioga: I agree - why don't we store the user id in the token store?
then I think we'd only need to fetch it for users who are upgrading from the previous version without requiring them to log in again, if that.

docs/release-notes/4052-filter-by-user.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Breaking Changes**
+
+-  API: Endpoints for getting a list of experiments, models, commands, notebooks, shells, and
+   tensorboards now accept a ``userIds`` field of user IDs instead of a ``users`` field of usernames
+   for filtering by user.
```
stoksc, line 7:
```suggestion
-  API: Endpoints for getting lists of experiments, models, commands, notebooks, shells, or
   tensorboards are now filterable by ``userIds`` instead of ``users``.
```
maybe just slightly clearer, but I'm bad a grammar.

harness/determined/common/api/authentication.py
johnkim-det: updating to use token store instead

harness/determined/common/api/authentication.py
```diff
@@ -138,7 +145,7 @@ def do_login(
     username: str,
     password: str,
     cert: Optional[certs.Cert] = None,
-) -> str:
+) -> Dict[str, str]:
```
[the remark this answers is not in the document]
johnkim-det: makes sense, will update

docs/release-notes/4052-filter-by-user.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Breaking Changes**
+
+-  API: Endpoints for getting lists of experiments, models, commands, notebooks, shells, and
+   tensorboards now filter by ``userIds`` instead of ``users`` (usernames)
```
stoksc, line 6:
```suggestion
   tensorboards now filter by ``userIds`` instead of ``users`` (usernames),
```

harness/determined/common/api/authentication.py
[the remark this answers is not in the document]
stoksc: even better may be to just use the new API and bindings here instead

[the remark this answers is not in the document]
johnkim-det: hmm not sure i know what you're referring to, do you mean using `LoginRequest` instead of `postLogin`?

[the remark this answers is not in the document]
stoksc: using https://github.com/determined-ai/determined/blob/master/harness/determined/common/api/bindings.py#L6765 from the new api bindings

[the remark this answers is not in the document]
johnkim-det: got it, will switch to using that.

harness/determined/common/api/authentication.py
```diff
@@ -138,38 +138,18 @@ def get_session_token(self, must: bool = True) -> str:
...
-        master_address,
-        "login",
-        json={"username": username, "password": password},
-        authenticated=False,
-        cert=cert,
+) -> bindings.v1LoginResponse:
+    certs.cli_cert = certs.default_load(master_address)
+    cli_auth = Authentication(master_address, try_reauth=True)
+    sess = det.common.experimental.session.Session(
+        master_address, "determined", cli_auth, certs.cli_cert
```
stoksc, line 150: we definitely can't hardcode the determined user. can we not make the request without a session or something initially? (login would be the only request this makes sense for, mostly)

johnkim-det: it looks like all the requests require a `session` for `session._do_request`. is that something i can just change in `bindings.py`, to use `api.get` instead? not sure if that's something that's okay for me to modify.

stoksc: It appears all the arguments to a session are optional, so I think you could just construct a session like
```
sess = det.common.experimental.session.Session(master_address, None, None, certs.cli_cert)
```

ioga: maybe even `Session(master_address, None, None, None)`?
something similar happened recently in https://github.com/determined-ai/determined/pull/4096/files

but I am starting to get worried about proliferation of this entire pattern because it *must* be easier to make API calls without worrying about this.

harness/determined/common/api/authentication.py
```diff
@@ -81,7 +82,7 @@ def _init_session(
             token = util.get_container_user_token()
 
         if token is not None:
-            return Session(session_user, token)
+            return Session(session_user, token, self.token_store.get_active_user_id())
```
ioga, line 85: Question: what happens to the currently logged in CLI users after upgrading to this?

johnkim-det:
@stoksc @ioga running into some test failures after switching to the new api bindings that i'm not sure how to address. the errors seem to indicate a problem with mocking the POST requests, even though i think the path is correct. any thoughts?

ioga:
> @stoksc @ioga running into some test failures after switching to the new api bindings that i'm not sure how to address. the errors seem to indicate a problem with mocking the POST requests, even though i think the path is correct. any thoughts?

the new path is probably not mocked?
there's no automatic mock for all URLs. here's how it's done: https://github.com/determined-ai/determined/blob/master/harness/tests/cli/test_cli.py#L51

johnkim-det:
> > @stoksc @ioga running into some test failures after switching to the new api bindings that i'm not sure how to address. the errors seem to indicate a problem with mocking the POST requests, even though i think the path is correct. any thoughts?
> 
> the new path is probably not mocked? there's no automatic mock for all URLs. here's how it's done: https://github.com/determined-ai/determined/blob/master/harness/tests/cli/test_cli.py#L51

thanks, was able to address that but now seeing another error. i think a test in `test_multimaster.py` is trying to reach `/api/v1/auth/login` at `https://localhost:12345/` and getting a 404. not sure what's going on there with the 12345, is there somewhere else that i need to update `/login` to `/api/v1/auth/login` or something?

ioga:
1. have you looked at what `test_multimaster.py` does? it runs its own mock server to test the certificates better.
2. I haven't seen this test being a flake, so it's probably the changes.

johnkim-det:
> 1. have you looked at what `test_multimaster.py` does? it runs its own mock server to test the certificates better.
> 2. I haven't seen this test being a flake, so it's probably the changes.

i guess what i meant is that the `test_users.py` failures do seem to be occurring as a result of using the new API bindings, but i'm having trouble following how exactly those changes are causing the `exitstatus` to be different on specific log in attempts. looking into it but please let me know if you happen to have any thoughts or context that i might be missing.

johnkim-det:
@bradley @ioga the more I think about this, I think that changing the CLI login to use the new API bindings is really outside the scope of this ticket, which is supposed to be about using userIds instead of usernames for filtering experiments. If you think that's fair, I'd like to go ahead and revert back to the commit before that specific change and get this landed.

harness/determined/common/api/authentication.py
ioga: this question still stands... do we need to shim the old token store to fetch the user id for existing logins?..

johnkim-det: what do you think about using the `users/me` call -- i think this should only occur one time, in the case where there's a token but no `active_user_id` in the token store?

ioga: sounds like a plan to me

harness/determined/common/api/authentication.py
```diff
@@ -81,7 +82,12 @@ def _init_session(
             token = util.get_container_user_token()
 
         if token is not None:
-            return Session(session_user, token)
+            if self.token_store.get_active_user_id() is None:
+                return Session(
+                    session_user, token, _get_current_user_id(self.master_address, token, cert)
```
ioga, line 87: once we fetched it here, should it then be saved back to the token store?

johnkim-det: thanks, will add

harness/determined/common/api/authentication.py
```diff
@@ -133,12 +143,24 @@ def get_session_token(self, must: bool = True) -> str:
         return self.session.token
 
 
+class AuthContext:
```
rb-determined-ai, line 147: We already have too many authentication-related objects, can you just return a tuple of (token, user_id) from do_login instead?

stoksc: I asked for this, instead of a dict[str,str], lol. Maybe a namedtuple or regular tuple is best?

stoksc: disadvantage of dict is that it breaks accidentally very easy, whereas I guess a type or tuple both don't.

johnkim-det:
I let this get stale for a while but it was actually one small change away from landing so trying to get it done now.

I’m seeing some test failures, but I’m pretty sure they’re not related to the work in this PR. Not sure of the best way to address, or if they should block this getting merged.

@hkang1 thoughts on the netlify errors?

@stoksc @ioga @rb-determined-ai the e2e test failures look like they're specifically related to tests that were recently updated in https://github.com/determined-ai/determined/pull/4429. again, it doesn't look to me like the failures are related to my user/auth changes in this PR, but maybe i'm missing something.

loksonarius (approved):

PR #4055, opened 2022-05-03
chore: remove remainder of native api

rb-determined-ai:
The Native API, deprecated in 0.13.5 (September 2020) is now completely
removed.


azhou-determined:
do we need to remove the det.experimental native usage from user-facing docs?

harness/determined/_experiment_config.py
```diff
@@ -8,9 +8,6 @@ def debug_enabled(self) -> bool:
     def scheduling_unit(self) -> int:
         return int(self.get("scheduling_unit", 100))
 
-    def native_enabled(self) -> bool:
-        return "internal" in self and self["internal"] is not None and "native" in self["internal"]
-
     def native_parallel_enabled(self) -> bool:
```
azhou-determined, line 11: completely unrelated, but what even is this?

azhou-determined (approved):
🥳

harness/determined/_experiment_config.py
rb-determined-ai: more dead code, I'm pretty sure.  Actually, now that I look around it's totally unused, so I'll just remove it.

PR #4057, opened 2022-05-03
fix: Add Workspace and Project name params to experiment yaml

mapmeld:
## Description

The YAML for an experiment loaded from the command line, forked from another experiment, etc. can include optional Workspace and Project parameters.  These should be the names of a current Workspace and nested Project.  If provided, the experiment should become associated with the Project.   If the named workspace/project is not found, or only one of the two parameters is provided, it should raise an error and prevent the experiment from being created or run.

## Checklist

### DIFF
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -270,6 +270,8 @@ type CreateExperimentParams struct {
 	GitCommitter  *string         `json:"git_committer"`
 	GitCommitDate *time.Time      `json:"git_commit_date"`
 	ValidateOnly  bool            `json:"validate_only"`
+	Project       *string         `json:"project"`
+	Workspace     *string         `json:"workspace"`
 }
 
 func (m *Master) parseCreateExperiment(params *CreateExperimentParams, user *model.User) (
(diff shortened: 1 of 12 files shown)


master/internal/core_experiment.go
```diff
@@ -350,9 +352,23 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams, user *mod
 	taskSpec.UserSessionToken = token
 	taskSpec.Owner = user
 
+	if (config.Workspace() == "" && config.Project() != "") || (config.Workspace() != "" && config.Project() == "") {
+		return nil, false, nil, errors.New("Workspace and Project must both be included in config if one is provided")
```
EmilyBonar: lint: has to all be lowercase

master/internal/db/postgres_experiments.go
```diff
@@ -26,6 +28,24 @@ const (
...
+	if err != nil {
+		return 1, err
+	}
+	p := projectv1.Project{}
+	err = db.Query("get_project_from_name", &p, w.Id, projectName)
+	if err != nil {
+		return 1, err
+	}
+	if p.Id < 1 {
+		return 1, fmt.Errorf("Workspace and Project names did not match any known project.")
```
EmilyBonar: lint: has to all be lowercase with no period at the end

schemas/expconf/v0/experiment.json
```diff
@@ -171,6 +171,13 @@
             "default": {},
             "optionalRef": "http://determined.ai/schemas/expconf/v0/profiling.json"
         },
+        "project": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "default": ""
```
EmilyBonar, line 179: nit: I think default should be `null`?

schemas/expconf/v0/experiment.json
```diff
@@ -225,6 +232,13 @@
             ],
             "default": null,
             "optionalRef": "http://determined.ai/schemas/expconf/v0/tensorboard-storage.json"
+        },
+        "workspace": {
+            "type": [
+                "string",
+                "null"
+            ],
+            "default": ""
```
EmilyBonar, line 241: nit: same as above, I think default should be `null`?

EmilyBonar (approved):
Looks good, I think the go test errors are coming from `TestExperimentConfig` in `master/pkg/schemas/expconf/schema_test.go`

PR #4058, opened 2022-05-03
chore: remove PBT searcher

stoksc:
## Description
This change removes PBT.

## Test Plan
- [x] tests should pass

## Checklist


docs/release-notes/remove-pbt.txt
```diff
@@ -0,0 +1,5 @@
+:orphan:
+
+**Breaking Changes**
+
+-  HP Search: Remove the PBT searcher.
```
rb-determined-ai, line 5: Remove the PBT searcher, which was deprecated in 0.17.6 (Jan 2022).

stoksc, line 5: @rb-determined-ai / @gh-determined-ai  any more context I should add here? can't strictly say "because asha is better". maybe "due to lack of use and engineering convenience" .. idk. I think this is probably just sufficient.

rb-determined-ai: I do think it's useful to include when it was deprecated, to show users who are strongly anti-breaking changes that they had plenty of time to adjust.

I don't think there's much else to say here.  "because it wasn't as good as any of our others" or "because nobody used it" seem not worth writing.

rb-determined-ai (approved):

docs/release-notes/remove-pbt.txt
stoksc, line 5:
```suggestion
-  HP Search: Remove the PBT searcher, which was deprecated in 0.17.6 (Jan 2022).
```

PR #4059, opened 2022-05-03
fix: use displayNames in ClusterHistoricalUsage

johnkim-det:
## Description

"Compute Hours by User" chart is not displaying `displayName` for users, still uses `username` only.

Screen Shot 2022-05-03 at 12 39 37 PM

## Test Plan

- [ ] make sure "Computer Hours by User" chart shows the display name if it exists

## Checklist

### DIFF
--- a/webui/react/src/pages/Cluster/utils.ts
+++ b/webui/react/src/pages/Cluster/utils.ts
@@ -1,10 +1,10 @@
 import { V1ResourceAllocationAggregatedEntry } from 'services/api-ts-sdk';
+import { DetailedUser } from 'types';
 import { sumArrays } from 'utils/array';
 import { secondToHour } from 'utils/datetime';
+import { getDisplayName } from 'utils/user';
 
 import { GroupBy } from './ClusterHistoricalUsage.settings';
-import { DetailedUser } from 'types';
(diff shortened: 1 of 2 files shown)


webui/react/src/pages/Cluster/utils.ts
```diff
@@ -17,18 +19,34 @@ export interface ResourceAllocationChartSeries {
...
+  period: Record<string, number>,
+  users: DetailedUser[],
+): Record<string, number> => {
+  const result:Record<string, number> = {};
+  Object.keys(period).forEach(key => {
+    const user = users.find(u => u.username === key);
+    const displayName = getDisplayName(user);
+    result[displayName] = period[key];
+  });
+  return result;
```
EmilyBonar, line 47: nit: you could do this as a reduce, something like
```
return (Object.keys(period).reduce((result, key) => {
    result[getDisplayName(users.find(u => u.username === key)] = period[key];
    return result;
}, {}));
```
But the way you have it works too, and might be easier to read.

EmilyBonar (approved):
LGTM!

hkang1 (approved):
👍

PR #4063, opened 2022-05-03
fix: correct the return type for readStream

hkang1:
## Description

There is an issue of the `readStream` of not returning the buffer data upon a terminal state. When the promise has resolved, the data is not returned back. The logic was previously broken by me 🤦  Fortunately found this issue before a release!

## Test Plan

- [ ] Load task logs page that already has logs, and it should load the initial set properly right away. When broken, it shows "no logs" message until a streaming event occurs.

## Checklist

### DIFF
--- a/webui/react/package-lock.json
+++ b/webui/react/package-lock.json
@@ -10,7 +10,6 @@
       "dependencies": {
         "ansi-to-html": "^0.7.1",
         "antd": "^4.16.13",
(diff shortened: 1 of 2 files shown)


trentwatt (approved):
LGTM.

hkang1:
removed `can-ndjson-stream` references in `package-lock.json`, great catch and thank you!

PR #4065, opened 2022-05-04
fix: fix old checkpoints

stoksc:
## Checklist

### DIFF
--- a/master/static/migrations/20220502180510_generic-checkpoints.tx.up.sql
+++ b/master/static/migrations/20220502180510_generic-checkpoints.tx.up.sql
@@ -99,7 +99,12 @@ CREATE OR REPLACE VIEW public.proto_checkpoints_view AS
             'experiment_id', c.experiment_id,
             'experiment_config', c.experiment_config,
             'hparams', c.hparams,
-            'training_metrics', c.training_metrics,
+            -- construct training metrics from the untyped jsonb deterministically, since older
+            -- versions may have old keys (e.g., num_inputs) and our unmarshaling is strict.
+            'training_metrics', jsonb_build_object(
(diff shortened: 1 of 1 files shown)


master/static/migrations/20220502180510_generic-checkpoints.tx.up.sql
```diff
@@ -99,7 +99,12 @@ CREATE OR REPLACE VIEW public.proto_checkpoints_view AS
...
             'experiment_config', c.experiment_config,
             'hparams', c.hparams,
-            'training_metrics', c.training_metrics,
+            -- construct training metrics from the untyped jsonb deterministically, since older
+            -- versions may have old keys (e.g., num_inputs) and our unmarshaling is strict.
+            'training_metrics', jsonb_build_object(
+                'avg_metrics', c.training_metrics->'avg_metrics',
+                'batch_metrics', c.training_metrics->'batch_metrics'
+            ),
             'validation_metrics', json_build_object('avg_metrics', c.validation_metrics),
```
rb-determined-ai, line 108: question, why not only apply this to the old checkpoints, via the pre-UNION part of the checkpoints_view definition?

My thinking is that the legacy logic should be as close to the parts that need legacy logic as possible, rather than applying legacy logic to new parts of the code where we know it to be a no-op.

rb-determined-ai (approved):
tested by downloading a checkpoint from 0.13.13, and this fixes it.

master/static/migrations/20220502180510_generic-checkpoints.tx.up.sql
stoksc: makes sense. I just thought, to everything else in the system, this is just jsonb so who cares what is it in.

PR #4067, opened 2022-05-04
chore: update deprecated reducer usage

rb-determined-ai:
Fix breakages introduced by #4060.


mpkouznetsov (approved):
I guess that's how we know the stuff is gone. Is it possible to create a lightweight experiment that would have caught this? Or maybe looking backwards is pointless?

rb-determined-ai:
I think the models in model_hub are sort of enormous by nature

PR #4068, opened 2022-05-04
feat: allow build_model attribute for general use

rb-determined-ai:
Also, avoid the backwards-compatible hasttr("build_model") check on
PyTorchTrial checkpoints saved by determined>=0.17.9, so users can
safely use the build_model attribute for their own purposes again.

## Test Plan

Added a new test with old checkpoints to guarantee the old behavior keeps working.


harness/determined/pytorch/_load.py
```diff
@@ -79,23 +85,29 @@ def load_trial_from_checkpoint_path(path: str, **kwargs: Any) -> pytorch.PyTorch
...
+            model_func = util.get_member_func(trial, "build_model")
+            if model_func is not None:
+                model = cast(torch.nn.Module, model_func())
+                model.load_state_dict(checkpoint["model_state_dict"])
+                # Note, with the very old checkpoints, we actually return the bare model.
+                return model  # type: ignore
+            raise errors.InvalidCheckpointException()
+
         # Backward compatible with older checkpoint.
         model_func = util.get_member_func(trial, "build_model")
```
coreystatendet, line 105: Could break `model_func = ...` out to above both branches.

PR #4071, opened 2022-05-04
feat: auth proxies

stoksc:
## Description
This change adds authentication to proxies.

## Test Plan
- [x] launch a notebook, without a token you are prompted to login and then redirected, with a token works as usual

## Checklist

### DIFF
--- a/master/internal/proxy/proxy.go
+++ b/master/internal/proxy/proxy.go
@@ -19,10 +19,10 @@ type (
 	// Register registers the service name with the associated target URL. All requests with the
 	// format ".../:service-name/*" are forwarded to the service via the target URL.
 	Register struct {
-		ServiceID   string
-		URL         *url.URL
-		ProxyTCP    bool
-		UseHTTPAuth bool
+		ServiceID       string
+		URL             *url.URL
+		ProxyTCP        bool
+		Unauthenticated bool
 	}
 	// Unregister removes the service from the proxy. All future requests until the service name is
 	// registered again will be responded with a 404 response. If the service is not registered with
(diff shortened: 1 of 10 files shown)


master/pkg/tasks/task_command.go
```diff
@@ -21,8 +21,9 @@ type GenericCommandSpec struct {
 
 	Keys *ssh.PrivateAndPublicKeys
 
-	Port     *int
-	ProxyTCP bool
+	Port        *int
+	ProxyTCP    bool
+	UseHTTPAuth bool
```
dzhu, line 26: Perhaps it should be inverted into `AllowUnauthenticated` or something so that the default is secure.

master/internal/user/service.go
```diff
@@ -107,6 +107,30 @@ func (s *Service) ProcessAuthentication(next echo.HandlerFunc) echo.HandlerFunc
...
+func (s *Service) ProcessProxyAuthentication(c echo.Context) (done bool, err error) {
+	token, err := s.extractToken(c)
+	if err != nil {
+		return true, c.Redirect(
+			http.StatusSeeOther,
+			fmt.Sprintf("/det/login?redirect=%s", c.Request().URL.Path),
+		)
+	}
+
+	switch user, _, err := s.db.UserByToken(token, s.extConfig); err {
```
dzhu, line 121: At least in isolation, I think the failure cases here (except maybe the default?) should also redirect instead of throwing a hard error. Though maybe that doesn't match up with the behavior elsewhere.

master/internal/user/service.go
```diff
@@ -107,6 +107,30 @@ func (s *Service) ProcessAuthentication(next echo.HandlerFunc) echo.HandlerFunc
...
 }
 
+// ProcessProxyAuthentication is a middleware processing function that attempts
+// to authenticate incoming HTTP requests coming through proxies.
+func (s *Service) ProcessProxyAuthentication(c echo.Context) (done bool, err error) {
+	token, err := s.extractToken(c)
+	if err != nil {
+		return true, c.Redirect(
+			http.StatusSeeOther,
+			fmt.Sprintf("/det/login?redirect=%s", c.Request().URL.Path),
```
dzhu, line 117: This throws away query parameters. Probably not that big a deal for the current use cases, but probably something to be avoided. Just dropping `.Path` seems to preserve the query, though there might still be encoding issues or something, I don't know.

master/internal/proxy/proxy.go
```diff
@@ -41,12 +42,19 @@ type Service struct {
 	URL           *url.URL
 	LastRequested time.Time
 	ProxyTCP      bool
+	UseHTTPAuth   bool
 }
 
+// ProxyHTTPAuth process a proxy request, returning true if the request should terminate
```
dzhu, line 48:
```suggestion
// ProxyHTTPAuth processes a proxy request, returning true if the request should terminate
```

dzhu (approved):

PR #4076, opened 2022-05-04
fix: filter out NaN, +/- Infinity metric values for charts for now.

hkang1:
## Description

Metric values can now come through as `NaN`, `Infinity` and `-Infinity`. Our charts can't render these values, so typically the chart just renders a blank chart. For now, filter out these values for the purposes of charting, and will address handling them better in the near future.

Before:
Screen Shot 2022-05-04 at 3 25 20 PM

After:
Screen Shot 2022-05-04 at 3 25 09 PM

## Test Plan

- [ ] HP Viz (Learning Curve, HP Scatter Plot, HP Heat Map, Parallel Coordinates) should be able to render despite containing NaN, +/-Inf metric values.
- [ ] Trial detail or single trial experiment metric chart should be able to render despite containing NaN, +/-Inf metric values.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
@@ -44,7 +44,7 @@ interface HpData {
   hpLabelValues: Record<string, number[]>;
   hpLabels: Record<string, string[]>;
   hpLogScales: Record<string, boolean>;
-  hpMetrics: Record<string, number[]>;
+  hpMetrics: Record<string, (number | null)[]>;
   hpValues: HpValue;
   metricRange: Range<number>;
   trialIds: number[];
@@ -213,7 +213,7 @@ const HpHeatMaps: React.FC<Props> = ({
 
     const canceler = new AbortController();
     const trialIds: number[] = [];
-    const hpMetricMap: Record<number, Record<string, number>> = {};
+    const hpMetricMap: Record<number, Record<string, number | null>> = {};
     const hpValueMap: Record<number, Record<string, Primitive>> = {};
     const hpLabelMap: Record<string, string[]> = {};
     const hpLabelValueMap: Record<string, number[]> = {};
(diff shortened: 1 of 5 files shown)


webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
```diff
@@ -247,19 +247,25 @@ const HpHeatMaps: React.FC<Props> = ({
...
               const key = generateHpKey(hParam1, hParam2);
-              hpMetricMap[trialId][key] = trial.metric;
+              hpMetricMap[trialId][key] = trialMetric;
             });
           });
 
-          if (trial.metric < metricRange[0]) metricRange[0] = trial.metric;
-          if (trial.metric > metricRange[1]) metricRange[1] = trial.metric;
+          if (trialMetric !== null && trialMetric < metricRange[0]) metricRange[0] = trialMetric;
+          if (trialMetric !== null && trialMetric > metricRange[1]) metricRange[1] = trialMetric;
```
trentwatt, line 268: `nit question`

would this be easier to read as 

`metricRange[0] = Math.min(metricRange[0], trialMetric ?? Number.MAX_SAFE_INTEGER)`

trentwatt (approved):
LGTM.

Nice, btw:
image

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpHeatMaps.tsx
hkang1: I think there are cases where the two lines would produce different results. The one I can think of is if `trialMetric` happens to be `null`. With how it's written currently, `metricRange[0]` will not actually be set to anything, where as the rewrite would produce `NaN` if `metricRange[0]` is `undefined`, or it would be set to whatever value `metricRange[0]` is set to, because it would be less than `Number.MAX_SAFE_INTEGER`.

hkang1:
thanks for the review @trentwatt! 🙏

PR #4077, opened 2022-05-04
fix: wrap torch.distributed launch in pid server/client 

azhou-determined:
## Description

due to a bug in Pytorch which doesn't catch NCCL failures, wrap the launch command in PID server/client

## Checklist

### DIFF
--- a/harness/determined/exec/pid_client.py
+++ b/harness/determined/exec/pid_client.py
@@ -7,7 +7,7 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("addr")
     parser.add_argument("cmd")
-    parser.add_argument("cmd_args", nargs="*")
+    parser.add_argument("cmd_args", nargs=argparse.REMAINDER)
     args = parser.parse_args()
 
     addr = ipc.read_pid_server_addr(args.addr)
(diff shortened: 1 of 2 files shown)


harness/determined/exec/pid_client.py
```diff
@@ -7,7 +7,7 @@
     parser = argparse.ArgumentParser()
     parser.add_argument("addr")
     parser.add_argument("cmd")
-    parser.add_argument("cmd_args", nargs="*")
+    parser.add_argument("cmd_args", nargs=argparse.REMAINDER)
```
rb-determined-ai, line 10: Are you sure this is going to work right?  In my former experiments, the normal_arg + REMAINDER behaved strangely, but using just REMAINDER to mean command+cmd_args worked as intended.

azhou-determined: just ran a script with `localhost -- python3 -m determined.exec.worker_process_wrapper RANK --`, and got the expected output of (addr='localhost', cmd='python3', cmd_args=['-m', 'determined.exec.worker_process_wrapper', 'RANK', '--'])

rb-determined-ai (approved):
this looks good, but have you confirmed that the change to pid_client cli isn't going to alter the way that any other `--`'s are treated by other launchers?

azhou-determined:
> this looks good, but have you confirmed that the change to pid_client cli isn't going to alter the way that any other `--`'s are treated by other launchers?

the unit tests ran successfully, but i will manually test horovod and deepspeed before merging.

rb-determined-ai:
I think unit tests of the launcher behavior is the most important thing here.  We just need to make sure we have the right coverage I think.

PR #4078, opened 2022-05-04
fix: Use CLI user in place of determined [DET-7179]

mapmeld:
## Description

In experiment interactions in the CLI, instead of the default "determined" user, we create a session for the CLI user.  Another option discussed was to use an existing session if available.

## Test Plan

CLI should successfully create an experiment and track logs. It should work even if the "determined" user is missing or inactive.

## Checklist


mapmeld:
Closing now that we have #4096

PR #4083, opened 2022-05-05
fix: Support rendering rank of 0 [DET-7186]

mapmeld:
## Description

Supports '0' as a Rank and not as a falsey value rendered as 'No Rank'.

## Test Plan

Includes a test that 0 and undefined appear as options (as '0' and 'No Rank').
'0' appears somewhere else in the DOM, so we check that '0' appears twice.

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer/LogViewerFilters.test.tsx
+++ b/webui/react/src/components/LogViewer/LogViewerFilters.test.tsx
@@ -63,6 +63,24 @@ describe('LogViewerFilter', () => {
     });
   });
 
+  it('should render filters with rank 0 and no rank', async () => {
+    const values: Filters = {
+      agentIds: [],
+      allocationIds: [],
+      containerIds: [],
+      levels: [],
+      rankIds: [ 0, undefined ],
+    };
+    setup(values, { ...values, rankIds: [] });
+
+    const agentOption1 = screen.getByText('All Ranks');
+    userEvent.click(agentOption1, undefined, { skipPointerEventsCheck: true });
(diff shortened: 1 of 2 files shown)


webui/react/src/components/LogViewer/LogViewerFilters.test.tsx
```diff
@@ -63,6 +63,24 @@ describe('LogViewerFilter', () => {
...
+      containerIds: [],
+      levels: [],
+      rankIds: [ 0, undefined ],
+    };
+    setup(values, { ...values, rankIds: [] });
+
+    const agentOption1 = screen.getByText('All Ranks');
+    userEvent.click(agentOption1, undefined, { skipPointerEventsCheck: true });
+    await waitFor(async () => {
+      expect(await screen.findAllByText('0')).toHaveLength(2);
```
hkang1, line 79: **question:** are there 2 because one in the dropdown menu and one in the dropdown picker itself?

hkang1, line 79: **question:** are there 2 because one in the dropdown menu and one in the dropdown picker itself?

hkang1 (approved):
LGTM! Thanks for adding the test!

webui/react/src/components/LogViewer/LogViewerFilters.test.tsx
hkang1: I see that there is a "ghost" div under "list-box" in the menu that has a duplicate list of the menu options

PR #4087, opened 2022-05-05
chore: HAL-2879 Cleanly shutdown all sshd servers on exit (#176)

jerryharrow:
## Description

Change the pidserver settings for sshd launches to SIGTERM during
pidserver exit.  Previously the setting was to wait for the sshd
processes, but they never exit so it can cause a hang.  With
this change when when the horovodrun command completes, and the
pidserver is about to exit, it instead sends SIGTERM to the
sshd processes and all containers can exit cleanly.

## Test Plan

Verified with successful run and test completion and shutdown using. SLURM  on casablanca & horizon with cifar10_pytorch distributed (8 slots).

## Checklist

### DIFF
--- a/harness/determined/launch/horovod.py
+++ b/harness/determined/launch/horovod.py
@@ -31,7 +31,7 @@ def create_sshd_worker_cmd(allocation_id: str, num_slot_ids: int) -> Tuple[List[
         "--on-fail",
         "SIGTERM",
         "--on-exit",
-        "WAIT",
+        "SIGTERM",
         f"/tmp/pid_server-{allocation_id}",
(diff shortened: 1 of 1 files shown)


rb-determined-ai (approved):
Great, thank you!

While this is not directly related to the SLURM integration work, it's worth noting that this will only fix our `determined.launch.horovod`-based training... there may be other work to do with the `determined.launch.deepspeed`-based training, and we need to at least document that daemon-mode containers won't work with slurm.

jerryharrow:
Thanks @rb-determined-ai -- note that I don't have write access, so you additionally have to submit the PR.

PR #4088, opened 2022-05-05
fix: Assorted fixes for Workspaces/Projects

mapmeld:
## Description

- Workspaces and projects have a name between 1 and 80 chars.
- Forking an experiment in the WebUI keeps it in the same project.
- Allow user to delete workspaces with a project, projects with an experiment, or combined (deleting workspaces with a project with an experiment).  The experiments are not deleted and go to the Uncategorized project (project_id = 1).

## Test Plan

- [x] Workspaces and projects have a name between 1 and 80 chars.
- [x] Forking an experiment in the WebUI keeps it in the same project.
- [x] Allow user to delete workspaces with a project, projects with an experiment, or combined (deleting workspaces with a project with an experiment). The experiments are not deleted and go to the Uncategorized project (project_id = 1).

## Checklist

### DIFF
--- a/harness/determined/common/api/bindings.py
+++ b/harness/determined/common/api/bindings.py
@@ -906,6 +906,7 @@ class v1CreateExperimentRequest:
         config: "typing.Optional[str]" = None,
         modelDefinition: "typing.Optional[typing.Sequence[v1File]]" = None,
         parentId: "typing.Optional[int]" = None,
+        projectId: "typing.Optional[int]" = None,
         validateOnly: "typing.Optional[bool]" = None,
     ):
         self.modelDefinition = modelDefinition
@@ -913,6 +914,7 @@ class v1CreateExperimentRequest:
         self.validateOnly = validateOnly
         self.parentId = parentId
         self.activate = activate
+        self.projectId = projectId
 
     @classmethod
     def from_json(cls, obj: Json) -> "v1CreateExperimentRequest":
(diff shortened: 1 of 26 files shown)


trentwatt (approved):
RE: Forking an experiment in the WebUI keeps it in the same project.

I wanted to test that this was the case after moving experiment, but it looks like there are some issues around moving an experiment. 

after move:
 - experiment shows up in experiment list for destination project
 - experiment details page shows experiment as being in original project (even after 30+ mins)
 - CLI does not show experiment in destination project

hopefully these are all related in an easily resolvable way. but maybe not.

i was able to test forking into a different project by editing config.

if i leave out the workspace and project when forking experiment though, it goes to uncategorized. that is not what i was expecting. i was thinking it would remain in the same project. i guess if we're always updating the experiment config whenever the experiment moves, that's okay, but i thought we wanted to keep that kind of metadata from being tied to config.

EmilyBonar:
Moving all of the experiments from a deleted project or workspace is not necessarily _wrong_, but if we do that we need to update the copy on `useModalProjectDelete` (seen below) and `useModalWorkspaceDelete`. This would also add an annoying second step to the process if some one really did want to entirely delete the project/workspace, experiments and all.
Screenshot 2022-05-06 at 11-20-16 Determined

mapmeld:
@EmilyBonar ok I forget where we landed on where the experiments go. I've made a commit to delete the experiments, too (which involves deleting the experiment's trial and task logs first).

PR #4091, opened 2022-05-06
docs: rework quickstart for ml developers

gh-determined-ai:
## Checklist

### DIFF
--- a/docs/getting-started.txt
+++ b/docs/getting-started.txt
@@ -2,13 +2,14 @@
  Quickstart for Model Developers
 #################################
 
-This quickstart uses the MNIST dataset to walk you through the basic steps to install Determined,
-run training experiments, and visualize the results in your browser. It uses three examples to
-demonstrate the scalability and enhanced functionality gained through configuration settings:
+This quickstart uses the MNIST dataset to demonstrate basic Determined functionality, and walks you
+through the steps needed to install Determined, run training jobs, and visualize experiment results
+in your browser. Three examples show the scalability and enhanced functionality gained from
+configuration setting changes:
 
--  Training on a local, single CPU or GPU.
--  Running a distributed training job on multiple GPUs.
--  Using hyperparameter tuning.
+-  Train on a local, single CPU or GPU.
+-  Run a distributed training job on multiple GPUs.
+-  Use hyperparameter tuning.
 
 An experiment is a training job that consists of one or more variations, or trials, of the same
 model. Using the Determined API, you automatically get metric frequency output, plots, and
--- a/docs/getting-started.txt
+++ b/docs/getting-started.txt
@@ -57,9 +57,9 @@ Hardware
 Docker
 ======
 
-Install Docker to run containerized workloads. If
-you do not already have Docker installed, follow the :doc:`Installing Docker
-</sysadmin-deploy-on-prem/requirements>` instructions to install and run Docker on Linux or macOS.
+Install Docker to run containerized workloads. If you do not already have Docker installed, follow
+the :doc:`Installing Docker </sysadmin-deploy-on-prem/requirements>` instructions to install and run
+Docker on Linux or macOS.
 
 Environment Variables
 =====================
@@ -125,7 +125,9 @@ you might want to review them to see how to call the Determined API from your co
 | ``model_def.py`` | Model definition and training/validation loops.                        |
 +------------------+------------------------------------------------------------------------+
 
-After gaining basic familiarity with Determined tools and operations, you can replacing these files with your model data and code, and setting configuration parameters for the kind of experiments you want to run.
+After gaining basic familiarity with Determined tools and operations, you can replacing these files
+with your model data and code, and setting configuration parameters for the kind of experiments you
+want to run.
 
 *****************************************
  Run a Local Single CPU/GPU Training Job
@@ -156,7 +158,9 @@ schedules to run.
 
       det experiment create const.yaml .
 
-   The last dot (.) argument uploads all of the files in the current directory as the *context directory* for your model. Determined copies the model context directory contents to the trial container working directory.
+   The last dot (.) argument uploads all of the files in the current directory as the *context
+   directory* for your model. Determined copies the model context directory contents to the trial
+   container working directory.
 
    You should receive confirmation that the experiment is created:
 
@@ -209,8 +213,8 @@ schedules to run.
 
    TBD: What is this image showing me that the user needs to know?
 
-With this fundamental understanding of Determined, you are ready to scale to distributed
-training in the next example.
+With this fundamental understanding of Determined, you are ready to scale to distributed training in
+the next example.
 
 ***************************************
  Run a Remote Distributed Training Job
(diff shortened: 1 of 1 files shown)


docs/getting-started.txt
```diff
@@ -27,75 +28,117 @@ code.
...
+you do not already have Docker installed, follow the :doc:`Installing Docker
 </sysadmin-deploy-on-prem/requirements>` instructions to install and run Docker on Linux or macOS.
 
-Training Example
-================
+Environment Variables
+=====================
+
+To run Determined on a remote master instance, set the remote IP address and port number in the
+``DET_MASTER`` environment variable:
```
rb-determined-ai, line 68: ~To run Determined on a remote master instance~ To connect to a Determined master running on a remote instance

Reason: setting environment variables locally will not cause a Determined master process to start running on any remote instance.

rb-determined-ai: From a doc-layout perspective, I think it's weird to have a layout like this:

- Prerequisites
  - Environment variables (for remotely-running master)
- Install and Deploy Determined Locally
- ...

It seems weird because the envvar section is listed as a prereq, which implicitly assumes the master is already running remotely as a prereq.  But then the first step is "install and deply determined locally".

Perhaps a more correct layout would be:

- Prerequisites:
  - ...
- Install and Deploy Determined...
  - Option 1: ...on the Local Machine
  - Option 2: ...on a Remote Machine
  
     In the body of this subsection, we include the the content from the Environment Variables section

docs/getting-started.txt
```diff
@@ -27,75 +28,117 @@ code.
...
+      P100, V100, A100, GTX 1080, GTX 1080 Ti, TITAN, or TITAN XP.
+   -  4GB RAM
+   -  50GB of free disk space.
 
 Docker
 ======
 
-For the examples that use an agent, install Docker to run containerized workloads. If you do not
-already have Docker installed, follow the :doc:`Installing Docker
+For the distributed example that uses an agent, install Docker to run containerized workloads. If
```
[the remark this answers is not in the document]
gh-determined-ai: Logic error, I guess. So, we don't even need to say "that use agent" because they all do. From the wording of the examples, it seemed like only example 2 used an agent.

docs/getting-started.txt
```diff
@@ -107,64 +150,29 @@ might want to review them to gain an understanding of how the Determined API is
...
-
-   export DET_MASTER=<ipAddress>:8080
-
-You can also include the master IP address in the command line:
-
-.. code:: bash
-
-   det -m http://<ipAddress>:8080 experiment create distributed.yaml .
+After gaining basic familiarity with Determined tools and operations, you can experiment in your
+domain by replacing these files with your model data and code, and setting configuration parameters
```
rb-determined-ai, line 154: After gaining basic familiarity with Determined tools and operations, you can ~experiment in your
domain by replacing these files with your model data and code, and setting configuration parameters~ replaces these files with your own model data, code, and configuration.

Or maybe... what is "in your domain" supposed to tell users?  To me it reads like, "feel free to replace this computer-vision model with a model related to the field of study you are in" which is not something I think we need to bother saying.

docs/getting-started.txt
gh-determined-ai: Yeah, that's how it was originally but a prerequisite is a prerequisite. Presumably, the reader is going to do all 3 examples. I prefer your first option, which clarifies when the envar applies.

docs/getting-started.txt
```diff
@@ -107,64 +150,29 @@ might want to review them to gain an understanding of how the Determined API is
...
 
    .. code:: bash
 
       det experiment create const.yaml .
 
-   The last argument dot notation (.) uploads all files in the current directory to the Determined
-   cluster and makes them available in the working directory in which each trial is run. This
-   directory is called the context directory.
+   The last dot notation (.) argument uploads all of the example files to the Determined cluster and
+   makes them available in the working, or *context*, directory in which each trial is run.
```
rb-determined-ai, line 173: I don't think `working directory` and `context directory` are interchangable terms, so `working, or context, directory` seems imprecise.

The context directory is what you upload to the server.  The working directory for training (which is an implementation detail and therefore not documented exactly) is always `/run/determined/workdir`.  We will copy the contents of the context directory you uploaded into your trial's working directory.

So maybe a rephrasing like this would be more appropriate:

The last dot (``.``) argument uploads all of the files in the current directory as the *context directory* for your model.  Determined will copy the contents of the model's context directory into the working directory of the Trial container.

docs/getting-started.txt
```diff
@@ -180,130 +188,178 @@ hyperparameters on a single slot.
...
-to use and how many model configurations to explore.
-
-#. Submit the experiment specifying the configuration file:
+This example demonstrates hyperparameter search. The example uses the ``adaptive.yaml``
+configuration file, which is similar to the ``const.yaml`` file in the first example but includes
+additional hyperparameter settings:
+
+.. code:: yaml
+
+   hyperparameters:
```
rb-determined-ai, line 279: I recommend using the `includeliteral` thing that I used in the core API docs to keep one source of truth for this code block.

This recommendation applies throughout the doc, there are other codeblocks which are copy/pasted but which don't have to be.

docs/getting-started.txt
```diff
@@ -180,130 +188,178 @@ hyperparameters on a single slot.
...
+
+#. Create and run the experiment:
 
    .. code:: bash
 
       det experiment create adaptive.yaml .
 
-#. Enter your cluster address in the browser.
+#. To view the WebUI dashboard, enter your cluster address in the browser address bar, accept the
+   default determined username, and click **Sign In**. No password is required.
```
rb-determined-ai, line 327: didn't we already say this in this same doc?

docs/getting-started.txt
rb-determined-ai: I didn't give two options, I showed what you have, then I should what I proposed.  The original was:

- Prerequeisites
  - ...
- Install and Deploy Determined Locally
- Use a Remote Determined Instance

The indentation of the hierarchies is not the same across between what was originally there and what I proposed.  In the original version installing locally and connecting to a remote installation are consecutive steps, which is confusing.

I'm ok with installing and deploying being part of prerequisites, or as the first step after the prereq.  But the hierarchy should show that only one needs to be executed in any case.

docs/getting-started.txt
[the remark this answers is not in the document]
gh-determined-ai: Done.

docs/getting-started.txt
gh-determined-ai: I think I effectively changed this as you suggested.

docs/getting-started.txt
gh-determined-ai: Done.

docs/getting-started.txt
gh-determined-ai: I think that might work for reference docs but I'm less inclined to do that for non-reference docs. The manual and mental effort needs to be put into maintaining this kind of content without having a (somewhat) hidden dependency. Using ``literalinclude`` doesn't hide the maintenance issue, particularly if the whole file isn't included.

docs/getting-started.txt
gh-determined-ai: I don't mind repeating if it saves the user from making a temporal leap for this kind of content. "Spacial adjacency trumps temporal stacking" - Tufte. Which is also why I'm moving the env var mention to where it applies.

PR #4095, opened 2022-05-06
chore: use reported slots available for on prem deployments

hamidzr:
## Description

on-premise deployments shouldn't rely on max agents for calculating or reporting total slots or agents.

- remove duplicate type definitions for ResourcePool type

## Checklist

### DIFF
--- a/webui/react/src/components/ResourcePoolCardLight.tsx
+++ b/webui/react/src/components/ResourcePoolCardLight.tsx
@@ -61,7 +61,12 @@ const poolAttributes = [
   {
     key: 'numAgents',
     label: 'Connected Agents',
-    render: (x: ResourcePool) => `${x.numAgents}/${x.maxAgents}`,
+    render: (x: ResourcePool) => {
+      if (x.maxAgents > 0) {
+        return `${x.numAgents}/${x.maxAgents}`;
+      }
+      return x.numAgents;
+    },
   },
   { key: 'slotsPerAgent', label: 'Slots Per Agent' },
   { key: 'auxContainerCapacityPerAgent', label: 'Aux Containers Per Agent' },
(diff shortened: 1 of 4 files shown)


gt2345 (approved):
LGTM

webui/react/src/pages/Cluster/ClusterOverview.tsx
```diff
@@ -41,7 +41,10 @@ const defaultSorter = { descend: false, key: 'name' };
  * @param pool resource pool
  */
 export const maxPoolSlotCapacity = (pool: ResourcePool): number => {
-  return pool.maxAgents * (pool.slotsPerAgent ?? 0);
+  if (pool.maxAgents) return pool.maxAgents * (pool.slotsPerAgent ?? 0);
+  // on-premise deployments don't have dynamic agents and we don't know how many
+  // agents might connect.
+  return pool.slotsAvailable;
```
hamidzr, line 47: @ioga does this look right to you?

webui/react/src/components/ResourcePoolCardLight.tsx
```diff
@@ -61,7 +61,12 @@ const poolAttributes = [
   {
     key: 'numAgents',
     label: 'Connected Agents',
-    render: (x: ResourcePool) => `${x.numAgents}/${x.maxAgents}`,
+    render: (x: ResourcePool) => {
+      if (x.maxAgents > 0) {
```
hamidzr, line 65: ditto @ioga

webui/react/src/pages/Cluster/ClusterOverview.tsx
ioga: seems like it'll work.

on a related note, the `pool.slotsPerAgent ?? 0` has always seemed "wrong" to me because the backend still returns `-1` for static pools. however `maxAgents` for static pools is always 0, which saves us from any negative numbers appearing here. So, probably not a big deal. But it'd be great if some day backend uses "null" value for unknown `slotsPerAgent`.

ioga (approved):

webui/react/src/pages/Cluster/ClusterOverview.tsx
hamidzr: it doesn't hurt to take slotsPerAgent < 1 out of consideration here. since now we're reusing this bit it should be a one-line change. I'll update it

PR #4098, opened 2022-05-07
feat: add per nic net throughput to the profiler

stoksc:
## Description
This change allow users to collect per NIC net throughput metrics through the system metrics profiler.

## Test Plan
- [x] run a job with `profiling.per_nic_net_throughput` both enabled and disable, ensure both work as expected.

## Checklist

### DIFF
--- a/docs/release-notes/per-nic-profiling.txt
+++ b/docs/release-notes/per-nic-profiling.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Improvements**
+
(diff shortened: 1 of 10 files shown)


trentwatt:
what is the unit? for webui

stoksc:
> what is the unit? for webui

same as the other net throughputs. good thought.

stoksc:
we really need the backend to just include a unit and let the web read it from the labels for the series

trentwatt:
we can do that

stoksc:
ok, I'll let mlsys take it from here if that is ok.

PR #4101, opened 2022-05-09
chore: explicit naming of preemption and coscheduler resources [DET-7140]

eecsliu:
## Description
Previously for the preemption scheduler, we would piggyback off of the coscheduler resources and just rename the scheduler. This was causing some issues when interpreting if the job was actually preemptible or not because it would still be listed for the "coscheduler" instead of the "preemption" scheduler. This PR changes the resources to be explicitly "coscheduler" or "preemption".

## Test Plan
Test feature:
1. Deploy k8s cluster
2. Submit a job
3. Go to the jobs task, click "Manage Job" for your job, and check if `Is Preemtible` == Yes

Test Upgrading:
1. Deploy k8s cluster using an old determined version
4. Use `helm upgrade` to upgrade it to the current one
5. Ensure it works smooth

## Checklist

### DIFF
--- a/helm/charts/determined/templates/master-config.yaml
+++ b/helm/charts/determined/templates/master-config.yaml
@@ -67,7 +67,7 @@ data:
       {{- if .Values.defaultScheduler}}
       {{- $schedulerType := .Values.defaultScheduler | trim}}
       {{- if or (eq $schedulerType "coscheduler") (eq $schedulerType "preemption")}}
-      default_scheduler: "coscheduler"
+      default_scheduler: {{ $schedulerType }}
       {{- end }}
       {{- end }}
       {{- if (ne (default "gpu" .Values.slotType) "gpu") }}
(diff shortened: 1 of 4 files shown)


eecsliu:
Thoughts on naming of the scheduler - `{{ $schedulerType }}` is either "coscheduler" or "preemption" and in all cases "coscheduler" makes sense, but "preemption" on its own might not. Users won't see this unless they probe into the scheduler pod or its service account.

We could change the scheduler type from "preemption" to "preemption-scheduler" to fix this, but that would impact the user. Alternatively, we could use `{{ $schedulerType }}-scheduler`, which would make "preemption" --> "preemption-scheduler" internally, but it would also make "coscheduler --> coscheduler-scheduler". Thoughts?

ioga (approved):

hamidzr:
> houghts on naming of the scheduler - `{{ $schedulerType }}` is either "coscheduler" or "preemption" and in all cases "coscheduler" makes sense, but "preemption" on its own might not. Users won't see this unless they probe into the scheduler pod or its service account.

if the name would be accompanied by a descriptive key when it's viewed I'd slightly lean toward not adding the `-scheduler` suffix. Either one works

PR #4109, opened 2022-05-10
chore: fix rank for horovod with mpi

stoksc:
## Description
this just affects logging, it seems.

## Checklist

### DIFF
--- a/harness/determined/launch/horovod.py
+++ b/harness/determined/launch/horovod.py
@@ -77,7 +77,7 @@ def create_worker_wrapper_cmd(allocation_id: str) -> List[str]:
         "python3",
         "-m",
         "determined.launch.wrap_rank",
-        "HOROVOD_RANK",
+        "HOROVOD_RANK,OMPI_COMM_WORLD_RANK,PMI_RANK",
         "--",
     ]
 
--- a/harness/determined/launch/wrap_rank.py
+++ b/harness/determined/launch/wrap_rank.py
@@ -57,7 +57,6 @@ def main() -> int:
     )
     args = parser.parse_args()
 
-    rank = None
     if set("0123456789") >= set(args.rank):
         # Rank is provided as a number.
         rank = int(args.rank)
--- a/harness/determined/launch/wrap_rank.py
+++ b/harness/determined/launch/wrap_rank.py
@@ -47,8 +47,9 @@ def main() -> int:
         "rank",
         metavar="RANK",
         help=(
-            "Can be an integer rank or the name of an "
-            "environment variable containing an integer rank."
+            "Can be an integer rank or a comma-separated list of "
+            "names of environment variables which are tried, in order, "
+            "to determine an integer rank."
         ),
     )
     parser.add_argument(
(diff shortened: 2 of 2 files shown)


jerryharrow (approved):

harness/determined/launch/wrap_rank.py
```diff
@@ -61,13 +62,16 @@ def main() -> int:
         rank = int(args.rank)
     else:
         # Rank is provided as the name of an environment variable.
-        if args.rank not in os.environ:
+        for rank in args.ranks.split(","):
```
jerryharrow, line 65: You need rank=None at 59:

stoksc: I think this time there is no way to get through without rank being defined like before. but I'll add anyway because I don't like python

stoksc: fixed my terrible code though, too

jerryharrow (approved):

harness/determined/launch/horovod.py
```diff
@@ -77,7 +77,7 @@ def create_worker_wrapper_cmd(allocation_id: str) -> List[str]:
         "python3",
         "-m",
         "determined.launch.wrap_rank",
-        "HOROVOD_RANK",
+        "HOROVOD_RANK,OMPI_COMM_WORLD_RANK,PMI_RANK",
```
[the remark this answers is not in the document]
rb-determined-ai: Never knew about this behavior, this is an excellent improvement.

harness/determined/launch/wrap_rank.py
```diff
@@ -47,27 +47,32 @@ def main() -> int:
...
+            "names of environment variables which are tried, in order, "
+            "to determine an integer rank."
         ),
     )
     parser.add_argument(
         "script", nargs=argparse.REMAINDER, metavar="SCRIPT...", help="The worker command."
     )
     args = parser.parse_args()
 
+    rank = None
```
rb-determined-ai, line 60: This looks like a bug.  It is never allowed to be None.

I mean... it should never _be_ None, but this line only confuses a reader.

rb-determined-ai (approved):

harness/determined/launch/wrap_rank.py
jerryharrow: It needs to be defined before the if/else otherwise they are only visible inside the if/else.

https://www.studytonight.com/python-howtos/scope-of-the-variable-initialised-in-the-if-statement

stoksc: I don't think that is true
```
>>> if True:
...     rank = 0
... 
>>> rank
0
```
python is really weird.

stoksc: or basically, scope in python is very different, maybe because it is interpreted.

stoksc, line 60:
```suggestion
```

jerryharrow: Hmm.  So must have been misunderstanding what lint was saying.   Or maybe it was more related to the prior version of your code where some path resulting it rank not being set which caused the issue.

PR #4112, opened 2022-05-10
chore: move out reusable webui utils and types

hamidzr:
## Description

goal: share core WebUI utils with saas.

- move utils, including routes, and services, constants, types to a shareable structure
- when needed move parts of a module to keep the changes limited to a simple move
- avoid making manual code changes in this pr. these should be made in a separate, smaller, and more easily reviewable pr. 

## Checklist

- [ ] note to self: once the pr is approved merge the shared-web branch first then update the submodule hash here to match before landing


johnkim-det (approved):
LGTM. one thought is that it could be a little confusing or annoying to manage when there are files with the same name in core and shared (like `types.ts`). maybe we could import the shared version into the core version so that other files don't have to worry about importing from both?

PR #4113, opened 2022-05-10
refactor: cleanup slurm shell scripts

jagadeesh545:
## Description

refactor: cleanup slurm shell scripts

## Checklist


stoksc:
@cla-bot check

master/static/srv/tensorboard-entrypoint.sh
```diff
@@ -2,10 +2,8 @@
 
 source /run/determined/task-signal-handling.sh
 source /run/determined/task-logging-setup.sh
-trap 'source /run/determined/task-logging-teardown.sh' EXIT
 
 set -e
-set -x
```
stoksc, line 8: let's leave the `set -x` for now. or maybe @rb-determined-ai you were mentioning the logs, want to just put it around the startup script? I think that's acceptable. though it may be nice to do something like `if debug == true, set -x`.

stoksc: maybe just drop it. if users want to add it to their startup script they can. and then in a separate PR we can enable it when debug is on.

master/static/srv/task-logging-setup.sh
```diff
@@ -69,3 +69,5 @@ exec > >(
 )
 
 ((DET_LOG_WAIT_COUNT += 2))
+
+trap 'source /run/determined/task-logging-teardown.sh' EXIT
```
stoksc, line 73: don't forget a newline

stoksc (approved):
lgtm mod final comments. I would probably call this `refactor: cleanup slurm shell scripts`

master/static/srv/task-logging-setup.sh
jagadeesh545: Updated.

master/static/srv/tensorboard-entrypoint.sh
rb-determined-ai: I think it's a good idea to just drop it.  We're past the point of "so simple it's fine to show users every detail" now.

PR #4117, opened 2022-05-11
chore: add a local docs server

rb-determined-ai:
## Description

The live.py local docs server supports autorebuilding whenever input
files change, and client-side autoreloading whenever a doc being viewed
is updated on disk.


gh-determined-ai (approved):
Works! A beauty.

PR #4118, opened 2022-05-11
ci: docs-only ci

rb-determined-ai:
## Description

PRs which are docs-only should not trigger the full CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,22 @@ commands:
       - run: git submodule sync
       - run: git submodule update --init
 
+  skip-if-docs-only:
+    steps:
+      - run: git remote add upstream https://github.com/determined-ai/determined
+      - run: git fetch upstream
+      - run:
+          name: check for docs-only changes
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -89,6 +89,22 @@ commands:
...
+  skip-if-docs-only:
+    steps:
+      - run: git remote add upstream https://github.com/determined-ai/determined
+      - run: git fetch upstream
+      - run:
+          name: check for docs-only changes
+          command: |
+            MERGE_BASE="$(git merge-base upstream/master HEAD)"
+            DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE" HEAD)"
+            if [ "$DIFF_DIRS" = ".circleci" ] ; then
```
rb-determined-ai, line 101: Before landing, this should be `"docs"` not `".circleci"`, but setting it to .circleci makes it easy for us to test the docs-only behavior in this PR

ioga (approved):

PR #4119, opened 2022-05-11
chore: add job ID and experiment labels to prometheus endpoint mappings [DET-6964]

azhou-determined:
## Description
Updates to prometheus endpoint:
- add job ID to det-state mappings
- add experiment labels on experiment creation

## Checklist

### DIFF
--- a/master/internal/prom/prom.go
+++ b/master/internal/prom/prom.go
@@ -112,9 +112,9 @@ func AssociateJobExperiment(jID model.JobID, eID string, labels expconf.Labels) 
 	AssociateExperimentIDLabels(eID, expLabels)
 }
 
-// DisassociateTaskExperiment disassociates a task ID with experiment info.
-func DisassociateTaskExperiment(tID model.TaskID, eID string, labels expconf.Labels) {
-	jobIDToExperimentID.WithLabelValues(tID.String(), eID).Dec()
+// DisassociateJobExperiment disassociates a job ID with experiment info.
+func DisassociateJobExperiment(jID model.JobID, eID string, labels expconf.Labels) {
+	jobIDToExperimentID.WithLabelValues(jID.String(), eID).Dec()
 	expLabels := make([]string, len(labels))
 
 	for l, _ := range labels {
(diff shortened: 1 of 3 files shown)


master/internal/prom/prom.go
```diff
@@ -25,14 +26,20 @@ Exposes mapping of allocation ID to container ID`,
...
 	containerIDToRuntimeID = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Subsystem: "det",
 		Name:      "container_id_runtime_container_id",
 		Help:      "a mapping of the container ID to the container ID given be the runtime",
 	}, []string{"container_runtime_id", "container_id"})
 
+	taskIDToExperimentID = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "det",
+		Name:      "task_id_experiment_id",
+		Help:      "a mapping of the experiment ID to the labels",
```
stoksc, line 40:
```suggestion
		Help:      "a mapping of the task ID to the experiment ID",
```

stoksc (commented):
what's our strategy on many map[x]y series vs (almost just) one map[x]map[y]map[z]... series? I'm not sure what is better. (though queries are way easier with the latter)

stoksc:
also, why `task_id_experiment_id` as opposed to `job_id_experiment_id`

azhou-determined:
> why task_id_experiment_id as opposed to job_id_experiment_id

no good reason. made the changes before adding job_id to the mappings. is job_id a better mapping? and why? 

> what's our strategy on many map[x]y series vs (almost just) one map[x]map[y]map[z]... series? I'm not sure what is better. (though queries are way easier with the latter)

yeah, it's a tradeoff between easier querying and easier code readability/cleanliness. doing a single mapping would require pulling in random data from various parts of the backend. also, a lot of queries are specific to filters and don't need to join all of the mappings (like experiment labels, jobs, etc) so it seems more straightforward as a strategy to try and add what we need piece-by-piece.

stoksc:
> no good reason. made the changes before adding job_id to the mappings. is job_id a better mapping? and why?

mostly just how the hierarchy of things works, it makes more sense. but it's not the end of the world.

stoksc (approved):
lgtm, though I think job_id_experiment_id makes a lot more sense

azhou-determined:
> lgtm, though I think job_id_experiment_id makes a lot more sense

made the change. 👍

PR #4120, opened 2022-05-11
chore: multiple distributed jobs collide on horovod run sshd port [HAL-2894]

rcorujo:
… port

## Description

Allow the user to change the ports used by Determined do deal with the situation where those ports may already be in use by another process on the compute node.

## Test Plan

Tested using the "cifar10_pytorch" experiment with the "distributed.yaml" file and these environment variables.  The experiment completed successfully.

```
environment:
  environment_variables:
     - DTRAIN_SSH_PORT=13000
     - HOROVOD_GLOO_RENDEZVOUS_PORT=13010
     - INTER_TRAIN_PROCESS_COMM_PORT_1=13020
```

Bradley said that users will typically be instructed to set these environment variables in "startup-hook.sh", rather than in the experiment's YAML file.  I set them in the experiment's YAML file for testing.

```
you dont even need to make configurations for all of this, for now
it can be 'secret'
we source startup-hook.sh where the hook is user provided
they can just set there if they need
it's a lot of work and much more public to open configs
```

## Checklist

### DIFF
--- a/harness/determined/constants.py
+++ b/harness/determined/constants.py
@@ -1,3 +1,5 @@
+import os
+
 # The maximum number of slots we expect any agent to have. Since we offset some ports
 # by the min(device_id) belonging to the trial, if we have two ports offset in this way,
 # we separate them by the max(min(device_id)) to avoid collisions. The two rendezvous ports
@@ -24,16 +26,21 @@ DEFAULT_EXP_CFG = {
     "optimizations": DEFAULT_OPTIMIZATIONS,
 }
 
+# Until we implement a more automatic solution, expose a temporary workaround of
+# allowing ports to be changed using envionment variables for the rare case that
+# the default ports are already in use by other processes.
+
 # TODO (DET-1189): Use port registry to allocate ssh port.
 # SSH port used for agents during dtrain (currently used with horovod and deepspeed backend).
-DTRAIN_SSH_PORT = 12350
+DTRAIN_SSH_PORT = int(str(os.getenv("DTRAIN_SSH_PORT", "12350")))
 
 # GLOO port used by Horovod for the Gloo controller.
-HOROVOD_GLOO_RENDEZVOUS_PORT = 12355
+HOROVOD_GLOO_RENDEZVOUS_PORT = int(str(os.getenv("HOROVOD_GLOO_RENDEZVOUS_PORT", "12355")))
 
 # Port for communicating between training processes. Used for reducing
 # validation metrics.
-INTER_TRAIN_PROCESS_COMM_PORT_1 = 12360
+INTER_TRAIN_PROCESS_COMM_PORT_1 = int(str(os.getenv("INTER_TRAIN_PROCESS_COMM_PORT_1", "12360")))
+
 INTER_TRAIN_PROCESS_COMM_PORT_2 = INTER_TRAIN_PROCESS_COMM_PORT_1 + MAX_SLOTS_PER_AGENT
 
 # How many seconds horovod waits for startup to complete before failing.


stoksc (commented):
lgtm, though I'll let @rb-determined-ai review since this is his area of expertise

harness/determined/constants.py
```diff
@@ -24,16 +26,39 @@
...
+#      - INTER_TRAIN_PROCESS_COMM_PORT_1=13020
+DTRAIN_SSH_PORT_ENV_VAR = "DTRAIN_SSH_PORT"
+HOROVOD_GLOO_RENDEZVOUS_PORT_ENV_VAR = "HOROVOD_GLOO_RENDEZVOUS_PORT"
+INTER_TRAIN_PROCESS_COMM_PORT_1_ENV_VAR = "INTER_TRAIN_PROCESS_COMM_PORT_1"
+
 # TODO (DET-1189): Use port registry to allocate ssh port.
 # SSH port used for agents during dtrain (currently used with horovod and deepspeed backend).
-DTRAIN_SSH_PORT = 12350
+# Note: Use of str() is to prevent lint complaint:
+#   error: Argument 1 to "int" has incompatible type "Optional[str]"
```
rb-determined-ai, line 46: I'd prefer you leave these `str()` comments out.  This doesn't quite raise enough of an eyebrow to justify a comment, and certainly not enough to justify three comments in close proximity.

harness/determined/constants.py
```diff
@@ -24,16 +26,39 @@
     "optimizations": DEFAULT_OPTIMIZATIONS,
 }
 
+# Allow users to override the ports using the environment variables, in case
+# there are other processes running on the compute node which are already using
+# our default port. Users will likely set these in their "startup-hook.sh".
+# Below is an example how to set these in the experiment's YAML file for testing.
```
rb-determined-ai, line 32: I spoke about this strategy for a while with @stoksc, and I agree that the overall approach is the right approach for now.  But I don't really want to give this feature "first-class" status, because some day a port registry or some sort of choose-port-then-allgather-chosen-port-numbers logic is probably a better strategy.

My real concern is that there would be people who would see this port-configuration logic you have here and would coopt it to accomplish something unrelated to the collision we are currently trying to avoid, and when we have a port registry or something, then we will have trouble removing these environment variables.

I think the right approach is to just set user expectations in this comment.  This isn't a first-class feature and we aren't promising it's always around.  So instead of talking about it like a normal feature, like "allow users to override... users will likely...", let's use phrasing more like "expose a temporary workaround in the rare case that ... until we implement a more automatic solution".

harness/determined/constants.py
```diff
@@ -24,16 +26,39 @@
...
+#
+# environment:
+#   environment_variables:
+#      - DTRAIN_SSH_PORT=13000
+#      - HOROVOD_GLOO_RENDEZVOUS_PORT=13010
+#      - INTER_TRAIN_PROCESS_COMM_PORT_1=13020
+DTRAIN_SSH_PORT_ENV_VAR = "DTRAIN_SSH_PORT"
```
rb-determined-ai, line 39: can we avoid this `LONG_NAME_VAR = "LONG_NAME"` and just use the `"LONG_NAME"` string directly?

There's a tiny type-checking benefit to the `LONG_NAME_VAR` symbol if we ever change the name, but that doesn't really matter because:
1. the only sane way for us to change an environment variable involves grepping for all instances of the old name anyway
2. changing the name would break all users who depend on it, so it really can't change anyway.

rb-determined-ai (commented):
I agree with this approach as a workaround.  Thank you for adding it.

rb-determined-ai:
Also, I don't understand the `... port` in the PR description.... did something get truncated?

stoksc:
> Also, I don't understand the ... port in the PR description.... did something get truncated?

This is just what happens if the title is too long, GitHub runs it into the description (but I fixed up the title).

harness/determined/constants.py
rcorujo: Will change the comment.

harness/determined/constants.py
rcorujo: Will change.

rb-determined-ai (approved):
thank you!

stoksc:
Allow the user to change the ports used by Determined do deal with the 
situation where those ports may already be in use by another process on 
the compute node.

PR #4123, opened 2022-05-12
fix: use custom image for tensor board [DET-7242]

gt2345:
## Description

When start a tensor board, disable override environment image when custom image is presented.

### DIFF
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -12,6 +12,7 @@ from tests.filetree import FileTree
 AWAITING_METRICS = "TensorBoard is awaiting metrics"
 SERVICE_READY = "TensorBoard is running at: http"
 num_trials = 1
+custom_image_for_testing = "custom_alpine"
 
 
 def shared_fs_config(num_trials: int) -> str:
@@ -159,3 +160,50 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
                 break
         else:
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
+
+
+@pytest.mark.slow
+@pytest.mark.e2e_gpu
+@pytest.mark.tensorflow2
+@pytest.mark.parametrize("prefix", [None, "my/test/prefix"])
+def test_start_tensorboard_with_custom_image(
+    tmp_path: Path, secrets: Dict[str, str], prefix: Optional[str]
+) -> None:
+    """
+    Start a random experiment configured with the shared_fs backend, start a
+    TensorBoard instance pointed to the experiment, and kill the TensorBoard
+    instance.
+    """
+    with FileTree(tmp_path, {"config.yaml": s3_config(1, secrets, prefix)}) as tree:
+        config_path = tree.joinpath("config.yaml")
+        experiment_id = exp.run_basic_test(
+            str(config_path), conf.fixtures_path("no_op"), num_trials
+        )
+
+    command = [
+        "tensorboard",
+        "start",
+        str(experiment_id),
+        "--no-browser",
+        "--detach",
+        "--config",
+        f"environment.image={custom_image_for_testing}",
+    ]
+    with cmd.interactive_command(*command) as tensorboard:
+        t_id = tensorboard.task_id
+        commandt = ["tensorboard", "config", t_id]
+        with cmd.interactive_command(*commandt, task_id=t_id) as tensorboard_config:
+            for line in tensorboard_config.stdout:
+                if "cpu" in line or "cuda:" in line or "rocm" in line:
+                    if custom_image_for_testing in line:
+                        break
+                    else:
+                        raise AssertionError(f"Setting custom image not working properly: {line}")
+            else:
+                raise AssertionError("Did not find custom image in output")
+        #     if SERVICE_READY in line:
+        #         break
+        #     if AWAITING_METRICS in line:
+        #         raise AssertionError("Tensorboard did not find metrics")
+        # else:
+        #     raise AssertionError(f"Did not find {SERVICE_READY} in output")
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -163,13 +163,11 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
 
 @pytest.mark.slow
 @pytest.mark.e2e_cpu
-def test_start_tensorboard_with_custom_image(
-    tmp_path: Path
-) -> None:
+def test_start_tensorboard_with_custom_image(tmp_path: Path) -> None:
     """
     Start a random experiment configured with the shared_fs backend, start a
-    TensorBoard instance pointed to the experiment with custom image, verify the image has been set, and kill the TensorBoard
-    instance.
+    TensorBoard instance pointed to the experiment with custom image, verify 
+    the image has been set, and kill the TensorBoard instance.
     """
     with FileTree(tmp_path, {"config.yaml": shared_fs_config(1)}) as tree:
         config_path = tree.joinpath("config.yaml")
--- a/e2e_tests/tests/command/test_tensorboard.py
+++ b/e2e_tests/tests/command/test_tensorboard.py
@@ -12,7 +12,6 @@ from tests.filetree import FileTree
 AWAITING_METRICS = "TensorBoard is awaiting metrics"
 SERVICE_READY = "TensorBoard is running at: http"
 num_trials = 1
-custom_image_for_testing = "custom_alpine"
 
 
 def shared_fs_config(num_trials: int) -> str:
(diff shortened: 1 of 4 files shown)


master/internal/api_tensorboard.go
```diff
@@ -317,6 +321,31 @@ func (a *apiServer) LaunchTensorboard(
...
+	if err != nil {
+		return false
+	}
+
+	type DummyImg struct {
+		Image string `json:"image"`
+	}
+	type DummyEnv struct {
+		Environment DummyImg `json:"environment"`
+	}
```
rb-determined-ai, line 339: This has a bug, since `environment.image` can be things other than a plain string.

I think you should have a dummy struct containing a `*model.RuntimeItem`, that when `nil` you know it was not set:

```go
type DummyEnv struct {
    Image *model.RuntimeItem `json:"image"`
}
type DummyConfig struct {
    Environment DummyEnv `json:"environment"`
}
```

By using `model.RuntimeItem` you can just borrow the unmarshaling logic, but by using a pointer you can check for nil, which is not normally possible with a `model.Environment` struct.

Also notice that calling the top-level dummy config `DummyEnv` is a bit confusing since it's a dummy replacement for the command config, and what you have as `DummyImg` is really a replacement for the model.Environment struct.

rb-determined-ai (commented):
Overall looks good.  I think this is the right strategy.

I think it's reasonable to discard errors in your `usingCustomImage` field but I think you should include a comment explaining why that's ok.  Otherwise somebody reading the code with less context might be confused.

Lastly, you might consider placing `usingCustomImage` in the `pkg/model/environment_config.go` file instead of where it is.  Your call on that one though, since it's only used by tensorboards.  My reasoning is that it might be easier to avoid subtle breakages if the schema changes if this function is closer to the schema definition.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +160,44 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
                 break
         else:
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
+
+
+@pytest.mark.slow
+@pytest.mark.e2e_gpu
+@pytest.mark.tensorflow2
+@pytest.mark.parametrize("prefix", [None, "my/test/prefix"])
```
rb-determined-ai, line 168: no need for any of these decorators I believe

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +160,44 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+def test_start_tensorboard_with_custom_image(
+    tmp_path: Path, secrets: Dict[str, str], prefix: Optional[str]
+) -> None:
+    """
+    Start a random experiment configured with the shared_fs backend, start a
+    TensorBoard instance pointed to the experiment, and kill the TensorBoard
+    instance.
```
rb-determined-ai, line 175: docstring is out-of-sync with the code

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +160,44 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    """
+    with FileTree(tmp_path, {"config.yaml": s3_config(1, secrets, prefix)}) as tree:
```
rb-determined-ai, line 177: can you allow the default configuration to take over, rather than force an s3 backend?  This is the sort of e2e test which should not depend on cloud credentials.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -12,6 +12,7 @@
 AWAITING_METRICS = "TensorBoard is awaiting metrics"
 SERVICE_READY = "TensorBoard is running at: http"
 num_trials = 1
+custom_image_for_testing = "custom_alpine"
```
rb-determined-ai, line 15: what is the `custom_alpine` image?

Also, I recommend using `alpine` because it's tiny and so the test will run quickly... but tensorboard will also fail immediately in that image.  So I would recommend hardcoding `"alpine"` in the test rather than putting it here.  It's a rather useless image for any other purpose.  Also, `custom_image_for_testing` doesn't really explain what the variable represents in the first place.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +160,44 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    command = [
+        "tensorboard",
+        "start",
+        str(experiment_id),
+        "--no-browser",
+        "--detach",
+        "--config",
+        f"environment.image={custom_image_for_testing}",
+    ]
+    with cmd.interactive_command(*command) as tensorboard:
```
rb-determined-ai, line 192: No need to make this interactive; the `--detach` arg will cause it to exit almost immediately.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +160,44 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+        t_id = tensorboard.task_id
+        commandt = ["tensorboard", "config", t_id]
+        with cmd.interactive_command(*commandt, task_id=t_id) as tensorboard_config:
```
rb-determined-ai, line 195: no need for this to be interactive either; you could pass the entire output of the command into a yaml parser and check the output that way, which would be clearer to read as well.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +159,39 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    Start a random experiment configured with the shared_fs backend, start a
+    TensorBoard instance pointed to the experiment with custom image, verify
+    the image has been set, and kill the TensorBoard instance.
+    """
+    with FileTree(tmp_path, {"config.yaml": shared_fs_config(1)}) as tree:
+        config_path = tree.joinpath("config.yaml")
+        experiment_id = exp.run_basic_test(
+            str(config_path), conf.fixtures_path("no_op"), num_trials
+        )
+    not_a_real_image = "not_a_real_image"
```
rb-determined-ai, line 176: this variable is pointless, just hardcode the string in place

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    """
+    Start a random experiment configured with the shared_fs backend, start a
+    TensorBoard instance pointed to the experiment with custom image, verify
+    the image has been set, and kill the TensorBoard instance.
+    """
+    with FileTree(tmp_path, {"config.yaml": shared_fs_config(1)}) as tree:
+        config_path = tree.joinpath("config.yaml")
+        experiment_id = exp.run_basic_test(
+            str(config_path), conf.fixtures_path("no_op"), num_trials
+        )
```
rb-determined-ai, line 178: Couldn't all of this just be:

```python
experiment_id = exp.run_basic_test(
    conf.fixtures_path("no_op/single-one-short-step.yaml"),
    conf.fixtures_path("no_op"),
    1,
)
```

I see that the `shared_fs_config` function in this file is from the old pedl repo (that is, it's very old) but I think it's just tech debt, that it doesn't add anything, and you should avoid it here.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+        str(experiment_id),
+        "--no-browser",
+        "--detach",
+        "--config",
+        "environment.image=alpine",
+    ]
+    env = os.environ.copy()
+    env["DET_DEBUG"] = "true"
+    res = subprocess.run(
+        command, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env
```
rb-determined-ai, line 194: drop `stderr=subprocess.PIPE` because it will show up in pytest without any extra effort.

drop `env` and DET_DEBUG because it's not necessary.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    )
+    assert res.returncode == 0, "\nstdout:\n{} \nstderr:\n{}".format(res.stdout, res.stderr)
```
rb-determined-ai, line 196: Instead of asserting the returncode, just put `check=True` in the `subprocess.run()` call.

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    t_id = res.stdout.replace("\n", "")
```
rb-determined-ai, line 197: replace with `res.stdout.strip()`

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
...
+    res = subprocess.run(
+        command, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env
+    )
+    assert res.returncode == 0, "\nstdout:\n{} \nstderr:\n{}".format(res.stdout, res.stderr)
+    config = yaml.safe_load(res.stdout)
+    assert (
+        config["environment"]["image"]["cpu"] == "alpine"
+        and config["environment"]["image"]["cuda"] == "alpine"
+        and config["environment"]["image"]["rocm"] == "alpine"
+    ), "\nSetting custom image not working properly:\n{}".format(config["environment"]["image"])
```
rb-determined-ai, line 208: Avoid `.format()` in any new code.  There was a time when we supported python 3.5 but now we don't even support 3.6.  Use f-strings instead.

But in this case, just omit the string entirely.  If it fails, people will know this info from the fact that `test_start_tensorboard_with_custom_image` failed.

Also, just include the whole raw output.  That way the failing test logs are as useful as possible ("useful" meaning "devs can understand what failed with just the test logs and the test code that ran"):

```
assert (...), res.stdout
```

e2e_tests/tests/command/test_tensorboard.py
```diff
@@ -159,3 +161,48 @@ def test_start_tensorboard_for_multi_experiment(tmp_path: Path, secrets: Dict[st
                 break
         else:
             raise AssertionError(f"Did not find {SERVICE_READY} in output")
+
+
+@pytest.mark.slow
```
rb-determined-ai, line 166: remove this.  This isn't a slow test at all.

Separately, I don't know if this mark even does anything anymore.

rb-determined-ai (approved):
Fantastic work!

Fixes a nasty bug in a small amount of code with no frills, complete with a test that is clearly written and intuitive.

PR #4126, opened 2022-05-12
fix: return user data when new user is created

johnkim-det:
## Description

Bug in POST users endpoint that returns an error because the newly created user cannot be found.

Bug was caused by looking for a user ID in the request to query by, instead of using the newly created user's ID.
Changing the `AddUser` method to return a userID, which can be used to query for the new user.

## Test Plan
- [ ] Create a new user directly with the API endpoint and make sure the new user's data is returned, instead of an error

## Checklist


master/internal/db/postgres_users.go
```diff
@@ -234,10 +234,10 @@ func deleteAgentUserGroup(tx *sqlx.Tx, userID model.UserID) error {
 }
 
 // AddUser creates a new user.
-func (db *PgDB) AddUser(user *model.User, ug *model.AgentUserGroup) error {
+func (db *PgDB) AddUser(user *model.User, ug *model.AgentUserGroup) (model.UserID, error) {
```
ioga, line 237: perhaps return a nil pointer here instead of 0 value?

```suggestion
func (db *PgDB) AddUser(user *model.User, ug *model.AgentUserGroup) (*model.UserID, error) {
```

docs/release-notes/4126-fix-post-user.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  API: Fix an issue where the ``POST /users`` endpoint always returned an error instead of the
```
ioga, line 5: any chance we could add a test for this?

master/internal/db/postgres_users.go
johnkim-det: do you think this is a blocker? tried using the pointer and nil values and starting running into a bunch of different errors that led to other errors... happy to address if it's important but also want to try to get this fix in ASAP.

docs/release-notes/4126-fix-post-user.txt
johnkim-det: do we have any similar tests that cover an API endpoint like this? trying to find something existing to reference.

ioga: couple options:
1. Integration tests in master: https://github.com/determined-ai/determined/blob/master/master/test/integration/api/api_checkpoints_test.go
2. can be just another python test in `e2e_tests/tests/cluster`, using bindings to hit the API.

master/internal/db/postgres_users.go
ioga: What kind of errors do you see?

docs/release-notes/4126-fix-post-user.txt
johnkim-det: adding test to `e2e_tests/test/cluster`

ioga (approved):
lgtm

e2e_tests/tests/cluster/test_users.py
```diff
@@ -181,6 +182,22 @@ def extract_id_and_owner_from_exp_list(output: str) -> List[Tuple[int, str]]:
...
 
+@pytest.mark.e2e_cpu
+def test_post_user_api(clean_auth: None) -> None:
+    master_url = conf.make_master_url()
+    authentication.cli_auth = authentication.Authentication(
+        conf.make_master_url(), requested_user="admin", password="", try_reauth=True
+    )
+    new_username = get_random_string()
+
+    user = bindings.v1User(active=True, admin=False, username=new_username, id=0)
```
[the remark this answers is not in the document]
ioga: I like your solution.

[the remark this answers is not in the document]
mapmeld: Do you know if there's a good way to put the commit on this PR? I've edited individual files in the web UI before but I don't know for a whole commit. Otherwise I'll make a new PR with just that change.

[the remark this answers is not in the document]
ioga: I don't think there's a way to commandeer a PR unless it's an upstream branch, so yes probably create a new one.

PR #4127, opened 2022-05-12
fix: No redirecting url when model name is changed

mapmeld:
## Description

For complicated reasons, we have linked to models by name (`/det/models/:name`) in the past. When the model name changed, we used #3559 to redirect to the new model URL.
We now prefer a permanent link with the numeric id  (`/det/models/:id`) so this redirect can be removed.
I don't think there's anywhere else in Web UI where the model name can be changed.

## Test Plan

- Create a model
- Visit that model's detail page
- Change the name of the model (we should not be redirected)
- Refresh the page to confirm that the name was successfully changed

## Checklist

### DIFF
--- a/webui/react/src/pages/ModelDetails.tsx
+++ b/webui/react/src/pages/ModelDetails.tsx
@@ -298,7 +298,6 @@ const ModelDetails: React.FC = () => {
         body: { name: editedName },
         modelName,
       });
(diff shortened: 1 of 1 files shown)


thiagodallacqua-hpe (approved):
Seems ok to me, but let us 🐻 in mind any possible fallout on user-facing interactions. 😉

PR #4128, opened 2022-05-12
feat: miscellaneous workspace ui functionalities + fixes [DET-7189]

trentwatt:
## Description

feat: miscellaneous workspace ui functionalities + fixes [DET-7189]

main things:

- got rid stale data in a lot of places
- QOL stuff around moving
- notes fixes
- disentangle task and experiment types
- consolidate logic to determine available experiment actions in `experiment.ts` (rather than having it duplicated across `TaskActionDropdown`, `ProjectDetails`, and `ExperimentDetailsHeader`)
- forked experiments stay in same project

QOL stuff around moving:
- batch move of experiments from project details table
- move experiments defaults to current workspace
- move experiments remembers last destination while on same page
- after move experiment show toast with link to destination project
- after move project show toast with link to destination workspace
- current project shows up destination list but disabled

lots of notes fixes:
- edit note title by clicking on it
- single click instead of double click to select note
- block editing notes for archived project
- fix max update depth exceeded for project empty notes
- fix cropping of blue highlight for inline editor for note title
- make border for selected note more defined
- fix stale note content after accepting unsaved changes message
- fix stale note content after switching notes
- go to new note when note is created
- stay on same note when note is deleted
- save title without saving note
- focus title when going to new note
- double click in content area to edit

etc:
- block unarchiving project from archived workspace
- block forking experiment in archived project from experiment details
- make name change in workspace details inline editor work
- more distinct highlight for selected project
- project name quicker to update
- fix various rerender loops
- fix issue where project header not update with navigation between projects
- fix stale enabled state for move button after going to new workspace
- adjust dropdown width to not crop "Others' workspaces"
- fixed locked archive toggle

## Checklist

[DET-7189]: https://determinedai.atlassian.net/browse/DET-7189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mapmeld (approved):
Looks like this works except for:
- block forking experiment in archived project from experiment details
- note editing issues which we discussed

I'll add forking/archived interaction to my current open PR so these other changes can go in

webui/react/src/utils/task.ts
```diff
@@ -129,7 +129,7 @@ export const generateExperiments = (count = 30): Type.ExperimentItem[] => {
         resourcePool: `ResourcePool-${Math.floor(Math.random() * 3)}`,
         searcherType: 'single',
         username: user.username,
-      } as Type.ExperimentItem;
+      } as unknown as Type.ExperimentItem;
```
mapmeld, line 132: Not a fan of this change. `generateExperiments` is for testing, so you ought to be able to add filler values if they are expecting `parentArchived: false` or other fields

trentwatt: yeah, all the experiment-specific code in that file needs to be moved to `experiment.ts` and disentangled from the tasks.

this is just a placeholder for now.

webui/react/src/components/InteractiveTable.tsx
```diff
@@ -319,7 +319,8 @@ const InteractiveTable: InteractiveTable = ({
         ?.getPropertyValue('--theme-sizes-layout-big').slice(0, -2),
     );
     if (typeof pagePadding !== 'number') pagePadding = 16;
-    return columnsWidths.reduce((a, b) => a + b, 0) + 2 * WIDGET_COLUMN_WIDTH + pagePadding;
+    const padding = 24;
+    return columnsWidths.reduce((a, b) => a + b, 0) + 2 * WIDGET_COLUMN_WIDTH + 2 * padding ;
```
mapmeld, line 323: is there any reason to put `padding = 24` here?
do we no longer use pagePadding?

webui/react/src/pages/ExperimentDetails.tsx
```diff
@@ -29,6 +29,7 @@ const ExperimentDetails: React.FC = () => {
   const [ canceler ] = useState(new AbortController());
   const [ experiment, setExperiment ] = useState<ExperimentBase>();
   const [ trial, setTrial ] = useState<TrialDetails>();
+  /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
   const [ valHistory, setValHistory ] = useState<ValidationHistory[]>([]);
```
mapmeld, line 33: this is a new one for me, but you can write `const [ , setValHistory ] = useState` to not receive a variable
removes need for the eslint exception

webui/react/src/hooks/useModal/useModalExperimentMove.tsx
```diff
@@ -1,29 +1,83 @@
...
+  storagePath: 'experiment-destination',
+};
+
+const moveExperimentWithHandler = async (
+  experimentId: number,
+  destinationProjectId: number,
+): Promise<number> => {
+  try {
+    await moveExperiment({ destinationProjectId, experimentId });
+    return 0;
```
mapmeld, line 68: nit / question, is there a benefit to making this return 1/0 instead of true/false?

mapmeld (approved):

webui/react/src/pages/ExperimentDetails.tsx
trentwatt: yeah so the only place it was being used [before](https://github.com/trentwatt/determined/blob/de46c36d7e46300cc4c9ce902f2d335f32d90bfe/webui/react/src/pages/ExperimentDetails.tsx#L46) was to see if it was equal to the previous value. 

i didn't want to just get rid of it, at least this way we call attention to its not being used when comes time to merge.

webui/react/src/components/InteractiveTable.tsx
trentwatt: this is basically just replicated [this](https://github.com/determined-ai/determined/pull/4138) it's to make the horizontal scrollbars not have to appear.

we can take it out as part of merge.

webui/react/src/hooks/useModal/useModalExperimentMove.tsx
trentwatt: no, i was initially going to add them but then i forgot `js` doesnt have array sum.

PR #4132, opened 2022-05-12
fix: user filter on dashboard [DET-7251]

hkang1:
## Description

Fix the issue of handling older local storage user filter settings that are still using usernames instead of user ids. The usernames in the dashboard filters and other areas with stick user pickers cause `userIds=` to be sent to the API quick causes `400 Bad Request` response.

## Checklist


johnkim-det (approved):
LGTM, thanks for figuring this out!

PR #4136, opened 2022-05-13
chore: fix boolean urlparams for grpc

rb-determined-ai:
The generated python bindings would emit boolean url parameters as
`val=True` but apparently grpc wants `val=true` instead.

## Test Plan

`det e list` now uses real booleans instead of string booleans, so test with and without `-a`.

## Commentary (optional)

It bugs me that we use generated bindings for "type safety" but that we also have this `limit_offset_paginator` which destroys type safety.  I spent quite a while learning about python's 3.10 [typing.ParamSpec feature](https://sobolevn.me/2021/12/paramspec-guide) but even with that I couldn't get things to work.  And I like what the function does so for now I have no response to it.

It might be worth making the generated bindings do the repeated paging at some point, which would be a lot more ergonomic for these paginated endpoints.

### DIFF
--- a/bindings/generate_bindings.py
+++ b/bindings/generate_bindings.py
@@ -28,6 +28,14 @@ class TypeAnno:
         # Only Refs to empty structs ever return True; we skip generating them.
         return False
 
+    def need_urlparam_dump(self) -> bool:
+        """
+        Dump a value for url parameters.
+
+        Defaults to need_parse(), since dump_as_urlparam() defaults to dump()
+        """
+        return self.need_parse()
+
     def dump_as_urlparam(self, val: Code) -> Code:
         """
         Dump a value for url parameters.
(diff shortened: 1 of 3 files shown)


mapmeld (commented):
- This works! :+1: I plan to remove a str(val).lower() patch in the workspaces/projects branch
- The `or not param.required` leads to many more `if {param} is not None else None` lines in the bindings, for example on most pagination variables. e2e_tests fail without it, but we didn't need it here before?

rb-determined-ai:
> The `or not param.required` leads to many more `if {param} is not None else None` lines in the bindings, for example on most pagination variables. e2e_tests fail without it, but we didn't need it here before?

Yeah, I realized there was a bug in my generator logic without it.  But it is a little over-complex right now, in the ways you are noticing.  Let me see if I can simplify it.

rb-determined-ai:
Ok, now if you remove the `is not None` from the `if {param.name} is not None else None` line in the generator, you will see that the generated code is identical, except for the boolean parameters, which have an obvious bug (hence the need for the `is not None` part)l.

mapmeld (approved):

PR #4137, opened 2022-05-13
docs: add launcher description to run training code

gh-determined-ai:
## Description

Add 'launcher' overview to 'Run Training Code'. Still needs some cleanup but review can start.

## Checklist


gh-determined-ai:
@rb-determined-ai  I need to put a pointer to this 'launcher' description in the Core API docs. Where's the best place to put it? Launcher and endpoint are only mentioned in your Distributed Training section of Core API.

gh-determined-ai:
Let me know who else should review this.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+
+   entrypoint: python3 -m (LAUNCH_LAYER) (TRIAL_DEFINITION)
+
+or a custom script:
+
+.. code:: yaml
+
+   entrypoint: python3 script.py arg1 arg2
+
+Entrypoints for pre-configured launch layers may differ slightly in arguments, but have the same
```
rb-determined-ai, line 32: @gh-determined-ai I thought we were avoiding the phrase "launch layer", in favor of "launcher"?   This appears in several places in the doc

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
 
 ********************
  Pause and Activate
 ********************
 
-An important feature of Determined is the ability to have trials stop running and then start again
-later without losing any training progress. The scheduler might choose to stop running a trial to
-allow a trial from another experiment to run, but a user can also manually pause an experiment at
-any time, which causes all of its trials to stop.
+Trials can be stopped and restarted without losing training progress. The scheduler can stop running
```
rb-determined-ai, line 240: let's say "trials can be paused and reactivated" since we use pause/stop buttons in the webui, where pause is the button that lets you continue, and the word "restart" could imply "from the beginning".  Also "activate" is the word we use in the cli and the python sdk for making a paused train start training again.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+
+Entrypoint supports nesting multiple launch layers in a single script. This can be useful for tasks
+that need to be run before the training code starts, such as profiling tools (dlprof), custom memory
+management tools (numactl), or data preprocessing.
+
+Example:
+
+.. code:: bash
+
+   dlprof --mode=simple python3 -m determined.launch.autohorovod --trial model_def:MnistTrial
```
rb-determined-ai, line 173: question for perhaps @liamcli or @coreystatendet or @garrett361, since I have no idea and couldn't find the answer online... Will this command line work?  There will be several processes between the dlprof main process and the process that actually does training.  I'm concerned that dlprof isn't actually going to work with those extra processes (determined.launch.horovod and horovodrun).

Separately, the name is no longer `autohorovod` but just `horovod`

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+
+Horovod Launcher
+================
+
+Format:
+
+``determined.launch.horovod [[HVD_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)``
+
+The module accepts arguments to be passed directly to the Horovod launcher called under-the-hood,
+horovodrun, which overrides automatically set values. See the official Horovod documentation for
```
rb-determined-ai, line 90: link to official horovodrun docs

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+with arguments passed as expected.
+
+Horovod Launcher
+================
+
+Format:
+
+``determined.launch.horovod [[HVD_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)``
+
+The module accepts arguments to be passed directly to the Horovod launcher called under-the-hood,
```
rb-determined-ai, line 89: > The module accepts arguments to be passed directly to the Horovod launcher called under-the-hood, horovodrun, 

Awkward phrasing.  How about:

```
The horovod launcher is a wrapper around 
`horovodrun <https://horovod.readthedocs.io/en/stable/summary_include.html#running-horovod>`_ 
which automatically configures the workers for the trial. You can pass arguments directly to 
``horovodrun`` (overriding any values set by Determined) as ``HVD_OVERRIDES``, which must 
end with a ``--`` to separate the overrides from the normal args.
```

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+
+where filepath is the location of your training class file, and ClassName is the name of the Python
+training class
+
+Custom Script
+-------------
+
+.. code:: bash
+
+   script.py [args...]
```
rb-determined-ai, line 77: the `.py` seems to imply that custom scripts must be written in python, which is not strictly necessary.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
  Run Training Code
 ###################
 
-In this document, we will cover how to run your training code. You can submit your code to a cluster
-and run them as experiments. We will go over the whole cycle of it.
+This section covers how to run training code by submitting your code to a cluster and running it as
+an experiment.
```
rb-determined-ai, line 8: I think we're missing a preamble that explains the doc at the very highest level.  I think that preamble should explain:
- your `entrypoint` is either the legacy bare-Trial-class form, or it's a command with args that we run in the container (that's so simple!).
- For distributed training, it's often wise to separate a "launcher" that kicks off some number of distributed workers from your training script (which is typically run on each worker).
- You can write your own custom launcher, or use one of our predefined launchers.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+training code. A set of default container images applicable to many deep learning tasks is provided,
+but you can also specify a :ref:`custom image <custom-docker-images>`. If the specified container
+images do not exist locally, the trial container fetches the images from the registry.
 
-After starting the containers, each trial runs the ``startup-hook.sh`` that exists in the context
+After starting the containers, each trial runs the ``startup-hook.sh`` script in the context
 directory.
 
-However, this whole process might take very long before each trial starts training.
+The pre-training activity can take a long time before each trial begins training.
```
rb-determined-ai, line 222: To make sure we don't look too bad, we should point out that in the ideal case this process only takes a couple seconds

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
 
 ****************
  Start Training
 ****************
```
rb-determined-ai, line 226: scrap this section; it reflects the pre-0.17.0 architecture where the master dictated how training went.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
  Pause and Activate
 ********************
 
-An important feature of Determined is the ability to have trials stop running and then start again
-later without losing any training progress. The scheduler might choose to stop running a trial to
-allow a trial from another experiment to run, but a user can also manually pause an experiment at
-any time, which causes all of its trials to stop.
+Trials can be stopped and restarted without losing training progress. The scheduler can stop running
+a trial to allow a trial from another experiment to run, but you can also manually pause, or
+*checkpoint*, an experiment at any time, which causes all of its trials to stop. After a trial is
```
rb-determined-ai, line 242: > you can also manually pause, or *checkpoint*,

pause != checkpoint, so this appositive is inaccurate.  It's true that pausing a Trial-based training results in a checkpoint before training actually stops, but it's also true that the majority of checkpoints are automatic and have nothing to do with pausing.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+set to stop, it takes a checkpoint at the next available opportunity, such as after the current
+workload completes, and stops running. This frees the slots used by the trial. When the trial
+resumes running, because more slots become available in the cluster or when you activate the
+experiment, it loads the saved checkpoint, continuing training from the saved state.
+
+***************
+ Best Practice
+***************
```
rb-determined-ai, line 250: local test mode only works with the legacy bare-trial-class entrypoint mechanism.  The new equivalent would be to just run your training script directly, but the ergonomics of running a core-api script off-cluster are not really worked out yet.

Not sure the best way to proceed from a docs perspective, but the strategy described here is on its way to being obsolete.

docs/training-run/index.txt
gh-determined-ai: Yep. Will fix.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+format:
+
+.. code:: yaml
+
+   filepath:ClassName
+
+where filepath is the location of your training class file, and ClassName is the name of the Python
+training class
+
+Custom Script
```
rb-determined-ai, line 72: Since a user can submit a bare script without a launcher, I'm not sure quite where this comment belongs.

But if a dtrain user writes their own launcher, it needs to have the following properties:

- Launch all desired workers, passing any required peer info (like rank or chief ip address) to each worker
- Monitor workers.  If any worker exits nonzero, the launcher should kill the remaining workers and exit nonzero.
- Exit nonzero after all workers exit nonzero

Those requirements ensure that distributed training jobs don't hang after a single worker failure.

rb-determined-ai: edit to add that a custom launcher SHOULD wrap each rank worker in the 
```
python3 -m determined.launch.wrap_rank $RANK CMD [ARGS...]
```
script, so that in the webui the final logs can be separated according to rank.  This is also described in the Core API distributed training example.

docs/training-run/index.txt
azhou-determined: @rb-determined-ai i've tested dlprof for pytorch before. it depends on the level of profiling you want to do. `dlprof --mode=simple` works as a launcher for broad GPU usage metrics. for something more granular (ie --mode=pytorch), you need to implement some hooks in the pytorch training code to tell dlprof where the markers are. but this is just an example of a launch layer, so i think that level of detail belongs elsewhere

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
-definition. The ``.detignore`` file uses the same syntax as `.gitignore
-<https://git-scm.com/docs/gitignore>`__. Note that byte-compiled Python files (e.g., ``.pyc`` files
-or ``__pycache__`` directories) are always ignored.
+DeepSpeed Launcher
+==================
 
-Local Test Mode
+Format:
+
+``determined.launch.deepspeed [[DEEPSPEED_ARGS...] --] (--trial TRIAL)|(SCRIPT...)``
```
azhou-determined, line 124: i believe deepspeed doesn't support additional override args (right @liamcli ?) so maybe we should remove this

docs/training-run/index.txt
azhou-determined: yeah, i wasn't quite sure where this belonged either. there's two levels of "custom" in the context of entrypoint: custom launcher, and custom script. perhaps this belongs in the entrypoint doc

azhou-determined:
i think the only minor change left is the deepspeed override args, we don't support those. otherwise, looks good.

docs/training-run/index.txt
gh-determined-ai: Not resolved. Please, review where I put it in the latest doc update.

docs/training-run/index.txt
gh-determined-ai: Not resolved.

docs/training-run/index.txt
gh-determined-ai: Left it as-is.

docs/training-run/index.txt
```diff
@@ -4,106 +4,255 @@
...
+Horovod Launcher
+================
+
+Format:
+
+``determined.launch.horovod [[HVD_OVERRIDES...] --] (--trial TRIAL)|(SCRIPT...)``
+
+The module accepts arguments to be passed directly to the Horovod launcher called under-the-hood,
+horovodrun, which overrides automatically set values. See the official Horovod documentation for
+more information about the horovodrun launcher. The optional override arguments must end with a
```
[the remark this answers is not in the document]
gh-determined-ai: Done

docs/training-run/index.txt
gh-determined-ai: Generalized it, leaving this as an example.

docs/training-run/index.txt
gh-determined-ai: Done

docs/training-run/index.txt
gh-determined-ai: Done

docs/training-run/index.txt
gh-determined-ai: Fixed

docs/training-run/index.txt
gh-determined-ai: Removed section.

gh-determined-ai:
@rb-determined-ai @azhou-determined  Please review again and let's find a solution for the unresolved items. If "custom launcher" and "custom script" are still unclear/ambiguous in this doc, that's an issue that needs to be cleared up in this doc.

docs/training-run/index.txt
```diff
@@ -4,106 +4,259 @@
...
+********************
+ Pause and Activate
+********************
 
-Checkpointing is essential to this ability. After a trial is set to be stopped, it takes a
-checkpoint at the next available opportunity (i.e., once its current workload finishes running) and
-then stops running, freeing up the slots it was using. When it resumes running, either because more
-slots become available in the cluster or because a user activates the experiment, it loads the saved
-checkpoint, allowing it to continue training from the same state it had before.
+Trials can be paused and reactivated without losing training progress. The scheduler can stop
```
rb-determined-ai, line 256: The scheduler can ~stop running a trial to allow a trial from another experiment to run~ pause a running trial to free its resources for another task ...

rb-determined-ai, line 256: Trials can be paused and reactivated without losing training progress.  Pausing trials preserve their progress by saving a checkpoint before exiting the cluster.

-----
This is a more natural point to explain that pausing entails a checkpoint operation than later in the text

docs/training-run/index.txt
```diff
@@ -4,106 +4,259 @@
...
+running a trial to allow a trial from another experiment to run, but you can also manually pause an
+experiment at any time, which causes all of its trials to stop. After a trial is set to stop, it
+takes a checkpoint at the next available opportunity, such as after the current workload completes,
+and stops running. This frees the slots used by the trial. When the trial resumes running, because
```
rb-determined-ai, line 260: running a trial to allow a trial from another experiment to run, but you can also manually pause an
experiment at any time, ~which causes all of its trials to stop. After a trial is set to stop, it
takes a checkpoint at the next available opportunity, such as after the current workload completes,
and stops running.~ which pauses all trials in the experiment.

docs/training-run/index.txt
```diff
@@ -4,106 +4,259 @@
...
+more slots become available in the cluster or when you activate the experiment, it loads the saved
```
rb-determined-ai, line 261: ~activate~ reactivate

docs/training-run/index.txt
rb-determined-ai: I think it's good where it's at.

docs/training-run/index.txt
gh-determined-ai: Done

docs/training-run/index.txt
gh-determined-ai: Done

docs/training-run/index.txt
gh-determined-ai: Done

gh-determined-ai:
@azhou-determined  Around line 159, "See the official DeepSpeed documentation ..." do you mean our DeepSpeed API doc or the DeepSpeed, itself, doc. If DS, itself, which of the 2 docs entry points:
[https://www.deepspeed.ai/getting-started/#writing-deepspeed-models/](https://www.deepspeed.ai/getting-started/#writing-deepspeed-models/)
or
[https://deepspeed.readthedocs.io/en/latest/](https://deepspeed.readthedocs.io/en/latest/)
?

gh-determined-ai:
@rb-determined-ai  I still think we need to mention the launchers or provide a link to this in the Core API doc.

docs/training-run/index.txt
gh-determined-ai: Done

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+images do not exist locally, the trial container fetches the images from the registry.
+
+After starting the containers, each trial runs the ``startup-hook.sh`` script in the context
+directory.
 
-#. train the model for a few batches on the training dataset;
-#. checkpoint the states of the model and other objects;
-#. validate the model on the validation dataset.
+The pre-training activity can incur a delay before each trial begins training but typically only
+takes a few seconds.
```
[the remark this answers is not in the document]
rb-determined-ai: there was some discussion on this point before.  An earlier version said something like, "this might take several minutes" and I was like, "don't make us look bad, when often it only takes a couple seconds".

I personally think any step which might take multiple minutes of waiting is worth documenting why it takes that long, but also worth mentioning that it can be nearly instantaneous under the right conditions (aka cloud spinup delays are not our fault).

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+
+For distributed training, it is good practice to separate the launcher that starts a number of
+distributed workers from your training script, which typically runs each worker. The distributed
+training launcher must implement the following logic:
+
+-  Launch all of the workers you want, passing any required peer info, such as rank or chief ip
+   address, to each worker.
+-  Monitor workers. If a worker exits with non-zero, the launcher should terminate the remaining
+   workers and exit with non-zero.
+-  Exit non-zero after all workers exit non-zero.
```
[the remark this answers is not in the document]
rb-determined-ai: Yes, I agree.  This was originally my typo.

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+an experiment.
+
+You run an experiment by providing a *launcher* and specify the launcher in the experiment
+configuration ``endpoint`` field. Launcher options are:
+
+-  a legacy bare-Trial-class
+
+-  a command with arguments that is run in the container
+
+   You can write a custom launcher or use one of the Determined predefined launchers.
```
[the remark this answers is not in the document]
rb-determined-ai: > but from reading this, I might think that it's normal to do so.

yeah, I agree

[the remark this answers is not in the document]
gh-determined-ai: done

docs/training-run/index.txt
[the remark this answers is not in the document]
gh-determined-ai: done

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+
+   entrypoint: python3 -m (LAUNCHER) (TRIAL_DEFINITION)
+
+or a custom script:
+
+.. code:: yaml
+
+   entrypoint: python3 script.py arg1 arg2
+
+Entry points for pre-configured launcher may differ slightly in arguments, but have the same format:
```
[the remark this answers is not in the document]
gh-determined-ai: Many more errors in addition to typo. Fixed.

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+
+To specify a trial class to be trained, the launcher accepts a TRIAL argument in the following
+format:
+
+.. code:: yaml
+
+   filepath:ClassName
+
+where filepath is the location of your training class file, and ClassName is the name of the Python
+training class
```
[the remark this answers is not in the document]
gh-determined-ai: I think at the start of this section we'll just say refer to the `endpoint` reference in the Experiment Config doc, which itself might need some improvement.

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+   dlprof --mode=simple python3 -m determined.launch.autohorovod --trial model_def:MnistTrial
 
-Our library automatically communicates with the master to get what needs to run and report back the
-results. It might:
+**********************
+ Create an Experiment
+**********************
+
+Currently, the CLI is the only supported mechanism for creating an experiment, although, you can
+create an experiment from an existing experiment or trial using the WebUI. To create an experiment:
```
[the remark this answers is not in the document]
gh-determined-ai: @azhou-determined Can you confirm that the API should be called out here?

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+Byte-compiled Python files, including ``.pyc`` files and ``__pycache__`` directories, are always
+ignored.
+
+********************
+ Pre-training Setup
+********************
+
+Trials are created to train the model. The :ref:`Hyperparameter Tuning <hyperparameter-tuning>`
+searcher specified in the experiment configuration file defines a set of hyperparameter
+configurations. Each set corresponds to a single trial.
```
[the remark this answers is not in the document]
gh-determined-ai: done

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+
+After the context and experiment configuration reach the master, the experiment waits for the
+scheduler to assign slots. If there are insufficient idle slots available, the master automatically
+scales up the resource pools.
```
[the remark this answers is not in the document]
gh-determined-ai: @azhou-determined  Please, confirm what you want here.

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
+
+When a trial is ready to run, the master communicates with the agent, or :ref:`distributed training
+<multi-gpu-training>` agents, which create(s) containers that have the configured environment and
+training code. A set of default container images applicable to many deep learning tasks is provided,
+but you can also specify a :ref:`custom image <custom-docker-images>`. If the specified container
+images do not exist locally, the trial container fetches the images from the registry.
```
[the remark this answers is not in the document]
gh-determined-ai: done

docs/training-run/index.txt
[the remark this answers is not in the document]
gh-determined-ai: Sounds good as is, and the user will not be surprised if it does take long.

docs/training-run/index.txt
```diff
@@ -4,106 +4,269 @@
...
 
-To start training, we load and run the entrypoint, which is the user-defined Python class and
-specified in the experiment configuration.
+.. code:: bash
 
-The user-provided class must be a subclass of a trial class included in Determined. Each trial class
-is designed to support one deep learning application framework. While training or validating models,
-the trial may need to load data from an external source therefore the training code needs to define
-data loaders.
+   dlprof --mode=simple python3 -m determined.launch.autohorovod --trial model_def:MnistTrial
```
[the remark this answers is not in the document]
azhou-determined: it's a bit of both a feature of determined and dependent on the script. this refers to what the determined.launch layers support, and i think the example mainly just points out this as a possibility. maybe "launch layers support" would be a better wording?

docs/training-run/index.txt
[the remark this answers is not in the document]
azhou-determined: i'm not sure what the future of `determined.experimental.client` is. do we intend to keep first-class support for that client? @rb-determined-ai 
maybe a better wording here is `the CLI is the recommended way to create an experiment`

docs/training-run/index.txt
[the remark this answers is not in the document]
azhou-determined: Maybe something like: "The master handles allocating necessary resources as defined in the cluster configuration."

azhou-determined:
don't see an approve option on this PR for some reason, but approved 👍

PR #4138, opened 2022-05-16
fix: adjust upscaling of column widths [DET-7220]

trentwatt:
## Description

fix: adjust upscaling of column widths [DET-7220]

cursor tracking of column resize was disturbed by theme architecture.

there was an existing issue as well where the default widths led to there being slight amount of horizontal scroll that would slightly hide filters on the rightmost column if it had any.

this addressses both. 

the basic thing is: 
- if we upscale too much, we cause horizontal scroll
- if we upscale too little, we lose resize tracking (proportionally to how far off we are)

My sense is that causing unnecessary horizontal scroll is worse than a slight loss of resize tracking, so I am erring on the side of upscaling too little here.

tried to do it based on computed values before but it wasn't working.

## Checklist

[DET-7220]: https://determinedai.atlassian.net/browse/DET-7220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -105,7 +105,9 @@ interface CellProps {
   isCellRightClickable?: boolean;
 }
 
-export const PAGE_PADDING = 16;
+const getAdjustedColumnWidthSum = (columnsWidths: number[]) => {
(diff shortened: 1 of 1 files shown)


hkang1 (approved):
Ooops, didn't mean to break this, thanks for fixing this!

PR #4140, opened 2022-05-16
feat: Enable sending empty string for displayname with fallback to username [DET-7031]

thiagodallacqua-hpe:
## Description

[Original Ticket]: DET-7031

We would like to enable users to send an empty string as a display name, which will fall back to its `username`.

## Test Plan

I ran the application locally and tried submitting an empty string as a display name, which resulted in the expected fallback.

image

image

image

## Commentary (optional)

I've noticed that, in the frontend, we already accept an empty string as a parameter for the request, and even return a toast message of success, although nothing really happens.

So, I've changed the users api to handle that request and, if the `displayname` parameter comes as `""`, it will fallback to the current user `username`.

This is my first PR, and I'm not really knowledgeable in Go... therefore I've taken a naive approach to the problem, which seems to work, but please do advise if there's a better/propper way to achieve this.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_users.py
+++ b/e2e_tests/tests/cluster/test_users.py
@@ -880,35 +880,31 @@ def test_experiment_delete() -> None:
 @pytest.mark.e2e_cpu
 @pytest.mark.e2e_cpu_postgres
 def test_change_displayname(clean_auth: None) -> None:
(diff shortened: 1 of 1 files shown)


mapmeld (approved):
Now LGTM. The remaining issues in lint are not in sections related to this code

PR #4142, opened 2022-05-16
feat: Tensorboard profiles from all hosts

mpkouznetsov:
## Description

This pull request addresses the following 2 issues:
https://determinedai.atlassian.net/browse/DET-6783
https://determinedai.atlassian.net/browse/DET-7238

TensorBoard profiler plugin was only displaying one host. This change introduces 2 strategies for handling metrics and profiling data for use by TensorBoard (`TensorboardMode`): `MANUAL` and `AUTO`.
In `AUTO` mode CoreContext writes metrics and uploads TB data for chief only.
In `MANUAL` mode it does nothing, and it is up to the user to write the metrics and anything else they want to make available for TensorBoard, and to call a new method, `TrainContext.upload_tensorboard_files()` to upload files saved by workers in the containers to checkpoint storage.
Trial API uses the `MANUAL` mode but all the TrialController classes are modified to write the metrics on the chief and upload tensorboard files from the chief _and all the workers_ while renaming them to avoid collisions.

### Background
Tensorboard profiling consists of 2 parts: the [callback](https://github.com/determined-ai/determined/blob/b1829b8c8e1f3bdd189f1296fd6eefd9e21a278f/harness/determined/keras/callbacks.py#L660) that writes logs and profiling data and the Tensorboard executable that picks up these logs and profiling data from a specified location.
In determined the logger/profiler writes to a location in a container from which we then copy TB files to a location under mounted checkpoint storage.

Since the only metrics that make sense are the ones on chief, we previously did not copy TB files for non-chief-workers out of the container. In this PR I introduce a method, `TrainContext.upload_tensorboard_files()` copying TB profiler plugin data out of the container to checkpoint storage. This method can be used to copy profiler logs to rank-aware locations where TB expects to find them. It takes an optional mangler function argument that the caller can provide to inject the rank into the file names. Note: *we decided that it is up to the user to call `upload_tensorboard_files()` and to provide the mangler that works for them.*

I also ported https://github.com/determined-ai/determined-ee/pull/194 to make sure different trials do not overwrite each other's metrics and profiling data.

## Testing
Add a profiling callback to a Keras trial:
```
class FashionMNISTTrial(TFKerasTrial):
    def __init__(self, context: TFKerasTrialContext) -> None:
        self.context = context

    def keras_callbacks(self):
        tboard_callback = callbacks.TensorBoard(
                                                 histogram_freq = 1,
                                                 profile_batch = (10,100))
        return [tboard_callback]
    def build_model(self):
```
Run the trial, then run TB.

This is what the Hosts dropdown looks like after this change:
image

## Commentary (optional)

We provide a basic level of service to the customer who asked for this feature. Instead of us automatically uploading and renaming files in Keras/PyTorch case (maybe based on some setting), we leave it up to the user to add a Keras callback that would call `upload_tensorboard_files` with a correct implementation of mangler provided to it.

## Checklist

### DIFF
--- a/harness/determined/util.py
+++ b/harness/determined/util.py
@@ -341,7 +341,11 @@ def force_create_symlink(src: str, dst: str) -> None:
             os.unlink(dst)
         else:
             shutil.rmtree(dst)
-        os.symlink(src, dst, target_is_directory=True)
+        try:
+            os.symlink(src, dst, target_is_directory=True)
+        except FileExistsError:
+            # in case of a race between two workers
+            pass
 
 
 @contextlib.contextmanager


harness/determined/tensorboard/build.py
```diff
@@ -27,18 +27,18 @@ def get_rank_if_horovod_process_else_return_zero() -> Optional[int]:
     return int(os.getenv("HOROVOD_RANK", 0))
 
 
-def get_base_path(checkpoint_config: Dict[str, Any], manager: bool = False) -> pathlib.Path:
+def get_base_path(checkpoint_config: Dict[str, Any]) -> pathlib.Path:
+    allocation_id = os.environ.get("DET_ALLOCATION_ID", "")
     rank = get_rank_if_horovod_process_else_return_zero()
 
     if checkpoint_config.get("base_path"):
-        return pathlib.Path(checkpoint_config["base_path"]).joinpath("tensorboard")
```
mpkouznetsov, line 34: I wonder whether the real cause of the collision troubles we are trying to fix is this line, i. e. the fact that if the user specified a `base_path` we will not differentiate by rank.

mpkouznetsov: If we introduce the differentiation of path by rank then the only way there will be collisions is if the user runs multiple trials with _the same checkpoint storage_. In other words, if we don't introduce the allocation_id suffix but do introduce the rank ending, and they want to avoid collisions they still can if only they run each trial with a different checkpoint storage in the experiment configuration. Now that I typed this, I feel that it is not a reasonable requirement so I guess we still should have an allocation_id suffix.

rb-determined-ai: No, I don't know how long ago base_path was part of the checkpoint storage config, but it's not allowed anymore (see `schemas/expconf/v0/checkpoint-storage.json`)

It's probably time to just remove this check.

harness/determined/tensorboard/util.py
```diff
@@ -26,3 +36,17 @@ def find_tb_files(base_dir: pathlib.Path) -> List[pathlib.Path]:
...
+        if p.match(f"*{ext}"):
+            print(f"matching *{ext}")
+            num_parts = ext.count(".")
+            while num_parts > 0:
+                p = p.with_suffix("")
+                num_parts -= 1
+            p = p.with_name(f"{p.name}#{rank}")
+            p = p.with_suffix(ext)
+            return p
+    return p
```
rb-determined-ai, line 52: can you explain to me what this is doing?  I don't think I understand what it's trying to accomplish.

harness/determined/core/_train.py
```diff
@@ -231,6 +240,9 @@ def report_validation_metrics(self, steps_completed: int, metrics: Dict[str, Any
             f"report_validation_metrics(steps_completed={steps_completed} metrics={metrics})"
         )
 
+    def upload_tb_profile(self, rank: int = 0) -> None:
+        logger.info(f"upload_tb_profile({rank})")
+
```
rb-determined-ai, line 245: For going into a public api, this is too specific to the current bug we are looking to fix.  The behavior as-written is:
- we automatically upload chief metrics to tensorboard
- you can upload profiling metrics manually 

I think a better setup would be:
- you upload the tensorboard paths you want manually on every rank
- optionally: we could have an "autotensorboard" mode which is exactly what we have on tip-of-master.

harness/determined/tensorboard/util.py
mpkouznetsov: I added a docstring explaining the purpose. It converts `/tmp/tensorboard-30.ed56557c-000c-4794-aa3e-eaffac608adc.1/plugins/profile/2022_05_13_15_25_41/ip-172-31-8-212.input_pipeline.pb` to `/tmp/tensorboard-30.ed56557c-000c-4794-aa3e-eaffac608adc.1/plugins/profile/2022_05_13_15_25_41/ip-172-31-8-212#7.input_pipeline.pb` for `rank=7`.

docs/release-notes/4142-tensorboard-profiler-data-from-workers.txt
```diff
@@ -0,0 +1,16 @@
+:orphan:
+
+**New Features**
+
+-  TensorBoard:
+      -  For Keras trials, the TensorBoard profiling data from all workers is uploaded to the
```
rb-determined-ai, line 6: Be precise here.  A lower-case trial is a run of an experiment.  A "Keras trial" could be a TFKerasTrial or a Core API training that uses Keras.  So rephrase as:  `In TFKerasTrial, the TensorBoard profiling...`

Same for "PyTorch trials"

docs/release-notes/4142-tensorboard-profiler-data-from-workers.txt
```diff
@@ -0,0 +1,16 @@
...
+         checkpoint storage with file names modified in such a way that the TensorBoard profiler
+         plugin displays data from different workers as coming from different hosts.
```
rb-determined-ai, line 8: Probably users don't care about the filename mangling.

harness/determined/core/_train.py
```diff
@@ -16,6 +19,27 @@ class EarlyExitReason(enum.Enum):
     USER_REQUESTED_STOP = "EXITED_REASON_USER_REQUESTED_STOP"
 
 
+class TensorboardSyncMode(enum.Enum):
```
rb-determined-ai, line 22: Need to add this class to the Core API reference docs

harness/determined/tensorboard/util.py
```diff
@@ -10,6 +10,16 @@
     "*.pb",
 ]
 
+profiler_file_extensions = [
```
rb-determined-ai, line 13: question: why not upload everything in `profile/`?

harness/determined/tensorboard/shared.py
```diff
@@ -25,11 +28,18 @@ def __init__(self, storage_path: str, *args: Any, **kwargs: Any) -> None:
...
+        self,
+        selector: Callable[[pathlib.Path], bool] = lambda _: True,
+        mangler: Callable[[pathlib.Path, int], pathlib.Path] = lambda p, __: p,
+        rank: int = 0,
+    ) -> None:
+        for path in self.to_sync(selector):
             shared_fs_path = self.shared_fs_base.joinpath(path.relative_to(self.base_path))
-            pathlib.Path.mkdir(shared_fs_path.parent, parents=True, exist_ok=True)
-            shutil.copy(path, shared_fs_path)
+            mangled_path = mangler(shared_fs_path, rank)
```
rb-determined-ai, line 39: Can we pass relative paths to the user functions?

E.g. if we're writing tensorboard files to `/tmp/SOME_RANDOM_NAME/profile/FILE` let's not leak the implementation details of `/tmp/SOME_RANDOM_NAME` to the user.  Instead select based on `selector("profile/FILE")` and mangle with input of `profile/FILE`?

harness/determined/tensorboard/azure.py
```diff
@@ -27,10 +28,20 @@ def __init__(
...
+        selector: Callable[[pathlib.Path], bool] = lambda _: True,
+        mangler: Callable[[pathlib.Path, int], pathlib.Path] = lambda p, __: p,
+        rank: int = 0,
+    ) -> None:
+        for path in self.to_sync(selector):
             whole_path = self.sync_path.joinpath(path.relative_to(self.base_path))
-            self.client.put("{}/{}".format(self.container, str(whole_path.parent)), path.name, path)
+            mangled_path = mangler(whole_path, rank)
+            self.client.put(
+                "{}/{}".format(self.container, str(mangled_path.parent)),
```
rb-determined-ai, line 41: let's update this to use f-strings

rb-determined-ai: python3.5 is so 2020

harness/determined/core/_train.py
```diff
@@ -16,6 +19,27 @@ class EarlyExitReason(enum.Enum):
...
+    storage, and the workers upload their profiling data. The profiling data file names are
+    altered to include the workers' rank in order to avoid collision and allow displaying each
+    worker's profile under a separate host in Tensorboard.
+
+    In ``Manual`` mode neither metrics nor profiling data are reported. It is up to the user
+    to call ``TrainContext.upload_tb_files()``
+    """
+
+    ChiefOnly = "CHIEF_ONLY"
+    ChiefAndWorkers = "CHIEF_AND_WORKERS"
```
rb-determined-ai, line 39: Why would a user want ChiefOnly, rather than just ChiefAndWorkers?

I mean, why not just have Auto and Manual, where Auto is really ChiefAndWorkers?

harness/determined/core/_train.py
```diff
@@ -221,6 +282,16 @@ def report_validation_metrics(self, steps_completed: int, metrics: Dict[str, Any
...
+    def _auto_upload_tb_files(self) -> None:
+        logger.info("_auto_upload_tb_files")
+
+    def upload_tb_files(
+        self,
+        selector: Callable[[pathlib.Path], bool] = lambda _: True,
+        mangler: Callable[[pathlib.Path, int], pathlib.Path] = lambda p, __: p,
+    ) -> None:
+        logger.info("upload_tb_files()")
+
```
rb-determined-ai, line 294: extraneous edit, it looks like

rb-determined-ai (commented):
I have a few thoughts.

Stepping back, it does seem a bit odd that we are even bothering to write metrics but not upload them.  If we only write training and validation metrics from rank=0, why wouldn't we upload all files all the time?  And offer just a mangler?  I mean, I guess we'd still want a selector on every rank for special cases.  But... why allow users to customize _when_ the upload happens?  Oh yeah, that's because they need to be synchronized.  Or... do they need to be synchronized?

Ok, this is a different idea: since `upload_tensorboard_files` is an idempotent synchronization API, which is pretty fast at ignoring pre-existing updates, what we added the `upload_tensorboard_files()` function to the API but left tensorboard_mode out entirely?  We could still call `upload_tensorboard_files()` internally in the same places and users could add additional synchronization points if they want.

Originally, I had it in my mind that Manual sync mode would make training faster (not introducing tb-related overhead), but I now realize that if you want to go fast, you should just _not_ write tensorboard events in the first place.  We should add that config, not add a config to write them but not upload them.  Does that make sense?

rb-determined-ai:
oh, also you are missing an API for giving the user a tensorboard sync path I think.

harness/determined/tensorboard/util.py
mpkouznetsov: We do upload everything. But we only mangle things we recognize.

Unfortunately, the profilers generate files with a variable number of "extensions". So currently I just remove known extensions, add `#<rank>` to what is left, then add the extensions back. An alternative would be to insert `#<rank>` before the first dot. However, if we ever get a host name with embedded dots we would be screwed.

harness/determined/core/_train.py
mpkouznetsov: I was thinking of giving an option to have the exact same behavior as before 0.18.1 but I guess there is little value in that?

harness/determined/core/_train.py
mpkouznetsov: Not sure what you mean. I have to explicitly override `upload_tb_files` because the parent version refers to `self._tb_mgr` which is not defined in `DummyTrainContext`.

rb-determined-ai: oh, I'm a dummy (lol)

harness/determined/core/_train.py
mpkouznetsov: done.

docs/release-notes/4142-tensorboard-profiler-data-from-workers.txt
```diff
@@ -0,0 +1,9 @@
+:orphan:
+
+**New Features**
+
+-  TensorBoard:
+      A new method, `TrainContext.upload_tensorboard_files()`, copies TensorBoard profiling data
+      files to checkpoint storage while optionally renaming them. It enables uploading profiling
+      data from non-chief workers (TensorBoard metrics and profiling data from chief is uploaded
+      automatically).
```
rb-determined-ai, line 9: This looks good, but is going to need additional explanation that:
- the call is only exposed to Core API users
- Trial API uses this new call under the hood to ship profiling data from all workers

harness/determined/_trial_context.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         self.env = env
 
         self.distributed = self._core.distributed
+        self.train = self._core.train
```
rb-determined-ai, line 23: I don't want to expose `core.train` to users, can add a getter on the TrialContext for the tensorboard path instead of handing users the whole TrainContext?

harness/determined/core/_train.py
```diff
@@ -16,6 +19,19 @@ class EarlyExitReason(enum.Enum):
...
+    """
+    ```TensorboardSyncMode`` defines how Tensorboard metrics and profiling data are retained.
+    In ``Auto`` mode only the chief uploads its metrics and profiling data to checkpoint
+    storage. This is the same behavior that existed prior to 0.18.2.
+    In ``Manual`` mode neither metrics nor profiling data are reported. It is up to the user
+    to call ``TrainContext.upload_tensorboard_files()``
+    """
+
+    Auto = "AUTO"
+    Manual = "MANUAL"
```
rb-determined-ai, line 32: python enums should be in all-caps.

(I see that the ones I recently added are not, but I also see that python enums support [aliasing](https://docs.python.org/3/library/enum.html#duplicating-enum-members-and-values) so I'll make a PR to correct that in a non-breaking way)

harness/determined/core/_train.py
```diff
@@ -89,10 +112,55 @@ def report_training_metrics(
...
+        self,
+        steps_completed: int,
+        metrics: Dict[str, Any],
+        batch_metrics: Optional[List[Dict[str, Any]]] = None,
+    ) -> None:
+        if self._tbd_writer:
             self._tbd_writer.on_train_step_end(steps_completed, metrics, batch_metrics)
-            self._tbd_mgr.sync()
+
+    def get_tensorboard_base_path(self) -> Optional[pathlib.Path]:
```
rb-determined-ai, line 129: take the word `base` out of the name, since it doesn't really add any meaning.

Also, is there any way we can make sure this returns a Path instead of an Optional[Path]?  It seems like it would be very inconvenient to callers for this to sometimes return None.

harness/determined/core/_train.py
```diff
@@ -48,6 +68,9 @@ def set_status(self, status: str) -> None:
         logger.debug(f"set_status({status})")
         self._session.post(f"/api/v1/trials/{self._trial_id}/runner/metadata", json=body)
 
+    def set_tensorboard_metric_writer(self, writer: tensorboard.BatchMetricWriter) -> None:
+        self._tbd_writer = writer
+
```
rb-determined-ai, line 73: This is a form of lazy initialization, and it makes it harder to think about the object because of the ways it can mutate over time.

Let's get rid of it and keep the API behavior simpler.  I think either you set up Auto mode and we picked a metric writer for you, or you set up Manual mode and we shouldn't allow you to set the metric writer afterwards.

harness/determined/core/_train.py
```diff
@@ -89,10 +112,55 @@ def report_training_metrics(
...
+                self._tbd_mgr.sync()
+
+    def write_training_tensorboard_metrics(
+        self,
+        steps_completed: int,
+        metrics: Dict[str, Any],
+        batch_metrics: Optional[List[Dict[str, Any]]] = None,
+    ) -> None:
+        if self._tbd_writer:
             self._tbd_writer.on_train_step_end(steps_completed, metrics, batch_metrics)
```
rb-determined-ai, line 127: hm... I know I thought this `write_training_tensorboard_metrics()` was a good idea, but now that I see it, I think it's a rather bad idea.  The calling pattern here would be:

- user creates core api with manual tb mode
- user creates one of our tensorboard.MetricWriter objects, which is our own internal interface, wrapped in a BatchMetricWriter, another one of our internal interfaces.
- user passes the BatchMetricWriter to set_tensorboard_metric_writer()
- user then is able to call write_training_tensorboard_metrics()
- write_training_tensorboard_metrics() just passes its args to the BatchMetricWriter that they already created
- user calls upload_tensorboard_metrics() to sync the tensorboard to persistent storage


This is pretty complex.  I'd rather the manual tb calling pattern look like:

- user creates core api with manual tb mode
- user writes metrics literally however they usually write metrics (not using our internal MetricWriter interfaces)
- user calls upload_tensorboard_metrics() to sync the tensorboard to persistent storage

harness/determined/core/_train.py
```diff
@@ -89,10 +112,55 @@ def report_training_metrics(
...
+        if self._distributed.rank == 0:
+            self.upload_tensorboard_files(
+                selector=lambda _: True,  # upload metrics and profiling data
+                mangler=get_rank_aware_path,
+            )
+        else:
+            self.upload_tensorboard_files(
+                selector=lambda p: not p.match("*tfevents*"),
+                mangler=get_rank_aware_path,
+            )
```
rb-determined-ai, line 148: The Core API is consumed by the TrialControllers, and so this should maybe be a helper function on the base TrialController, or maybe a standaline utility function somewhere.  But the Core API should not have helper functions for our training loops which consume the Core API.

harness/determined/_trial_controller.py
```diff
@@ -106,6 +106,11 @@ def supports_mixed_precision(cls: Type["TrialController"]) -> bool:
     def supports_averaging_training_metrics(cls: Type["TrialController"]) -> bool:
         return False
 
+    @classmethod
+    def get_metric_writer(cls: Type["TrialController"]) -> tensorboard.BatchMetricWriter:
+        # default to the original kludge to avoid regression in custom trial controllers
+        return tensorboard.get_metric_writer()
+
```
rb-determined-ai, line 113: I don't think there is a single custom TrialController in existence (outside of our own test fixtures).

If there is, they are super-hackers since they never once asked us for help, and as super-hackers they can figure this out on their own.

That said, if you want to leave it in "just in case" that's ok, but if you get rid of `TrainContext.set_tensorboard_metric_writer()` (which you absolutely should) you don't need this anymore.

harness/determined/core/_train.py
```diff
@@ -221,6 +296,19 @@ def report_validation_metrics(self, steps_completed: int, metrics: Dict[str, Any
...
+    def _auto_upload_tb_files(self) -> None:
+        logger.info("_auto_upload_tb_files")
+
+    def upload_tensorboard_files(
+        self,
+        selector: Callable[[pathlib.Path], bool] = lambda _: True,
+        mangler: Callable[[pathlib.Path, int], pathlib.Path] = lambda p, __: p,
+    ) -> None:
+        logger.info("upload_tensorboard_files()")
+
```
rb-determined-ai, line 311: these look extraneous?

harness/determined/core/_train.py
```diff
@@ -16,6 +19,19 @@ class EarlyExitReason(enum.Enum):
     USER_REQUESTED_STOP = "EXITED_REASON_USER_REQUESTED_STOP"
 
 
+class TensorboardSyncMode(enum.Enum):
```
rb-determined-ai, line 22: Does the word "Sync" belong in the name?  It actually governs both sync an metric writing, right?  Maybe just `TensorboardMode`?

harness/tests/tensorboard/test_base.py
```diff
@@ -30,7 +30,7 @@ def test_setting_optional_variable(tmp_path: pathlib.Path) -> None:
         env.det_cluster_id, env.det_experiment_id, env.det_trial_id, checkpoint_config
     )
     assert isinstance(manager, tensorboard.SharedFSTensorboardManager)
-    assert manager.base_path == pathlib.Path("test_value/tensorboard")
+    assert manager.base_path == pathlib.Path("test_value/tensorboard-")
```
rb-determined-ai, line 33: what is this for?

harness/tests/tensorboard/test_util.py
```diff
@@ -1,13 +1,44 @@
...
+            print(f"matching *{ext}")
+            num_parts = ext.count(".")
+            while num_parts > 0:
+                path = path.with_suffix("")
+                num_parts -= 1
+            path = path.with_name(f"{path.name}#{rank}")
+            path = path.with_suffix(ext)
+            return path
+    return path
+
```
rb-determined-ai, line 41: why define this inside the test file?

harness/determined/core/_train.py
mpkouznetsov: It sure _is_ lazy initialization. The reason I have to use it is that we have not instantiated a trial/controller pair at the moment of `core.init()` and therefore do not know which metric writer we intend to use.
The order of initialization is:
`core.init() -> core_context -> trial_context -> trial -> controller -> metric_writer -> core_context.set_tensorboard_metric_writer`. The alternative that avoids lazy initialization is what I call the "original kludge": to step on a potential land mine, and if we explode then use a second life to instantiate another writer (with no guarantee that we will actually be using the correct writer). Is there another alternative?

harness/tests/tensorboard/test_base.py
mpkouznetsov: The tensorboard path is now of form f"{tensorboard}-{allocation_id}" or f"{tensorboard}-{allocation_id}-{rank}". As you can see, at least one dash is now there.

harness/determined/core/_train.py
mpkouznetsov: I changed it to `TensorboardMode` and now I am like, wait a minute, writing _is_ part of sync in some sense, isn't it? idk, let's go with `TensorboardMode`.

harness/determined/_trial_context.py
mpkouznetsov: Since all trial classes write metrics and upload profiling data for all workers already, I just removed this line.

harness/tests/tensorboard/test_util.py
mpkouznetsov: removed.

harness/determined/_trial_controller.py
```diff
@@ -106,6 +108,11 @@ def supports_mixed_precision(cls: Type["TrialController"]) -> bool:
     def supports_averaging_training_metrics(cls: Type["TrialController"]) -> bool:
         return False
 
+    @classmethod
+    @abc.abstractmethod
+    def _create_metric_writer(cls: Type["TrialController"]) -> tensorboard.BatchMetricWriter:
```
rb-determined-ai, line 113: whole object is a internal API... no real need for `_`

examples/computer_vision/mnist_pl/mnist.py
```diff
@@ -15,14 +15,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import data
```
rb-determined-ai, line 18: I think this sorting of import is worse, since data should be treated as a local module rather than 3rd party (that is, it should below, as before).

examples/computer_vision/mnist_pl/model_def.py
```diff
@@ -4,12 +4,12 @@
 LightningModule with Determined's PyTorch control loop.
 """
 
-from determined.pytorch import PyTorchTrialContext, DataLoader
-from determined.pytorch.lightning import LightningAdapter
-
 import data
 import mnist
```
rb-determined-ai, line 8: ditto

harness/determined/core/_context.py
```diff
@@ -123,6 +124,9 @@ def init(
             ``core_context.preempt.should_preempt()`` method.  See
             :class:`~determined.core.PreemptMode` for more detail.  Defaults to ``WorkersAskChief``.
         storage_manager: Internal use only.
+        tensorboard_mode (``core.TensorboardMode``, optional): Define how Tensorboard
+            metrics and profiling data are retained. See
+            :class:`~determined.core.TensorboardMode`` for more detail.
```
rb-determined-ai, line 129: Nit:  include a `Defaults to` sentence at the end, according to [this guide](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

harness/determined/core/_tensorboard_mode.py
```diff
@@ -0,0 +1,14 @@
+import enum
+
+
+class TensorboardMode(enum.Enum):
+    """
+    ```TensorboardMode`` defines how Tensorboard metrics and profiling data are retained.
+    In ``Auto`` mode only the chief uploads its metrics and profiling data to checkpoint
+    storage. This is the same behavior that existed prior to 0.18.2.
+    In ``Manual`` mode neither metrics nor profiling data are reported. It is up to the user
+    to call ``TrainContext.upload_tensorboard_files()``
```
rb-determined-ai, line 10: You say "defines how Tensorboard **metrics and profiling** data are **retained**", but this applies to all tensorboard files of every kind (not just metrics and profiling), and it isn't really a retention policy.

Recommended change:
```
``TensorboardMode`` defines how Tensorboard artifacts are handled.

In ``Auto`` mode the chief automatically writes any reported training or validation 
metrics to the Tensorboard path (see meth:`core_context.trian.get_tensorboard_path<
determined.core.TrainContext.get_tensorboard_path>`), and automatically 
uploads all of its own tensorboard artifacts to checkpoint storage.  Tensorboard 
artifacts written by non-chief workers will not be uploaded at all. This is the 
same behavior that existed prior to 0.18.2.

In ``Manual`` mode no Tensorboard artifacts are written or uploaded at all. 
It is entirely up to the user to write their desired metrics and upload 
them with calls to :ref:`core_context.train.upload_tensorboard_files<
determined.core.TrainContext.upload_tensorboard_files>`.
```

harness/determined/core/_train.py
```diff
@@ -89,10 +95,30 @@ def report_training_metrics(
...
+        mangler: Callable[[pathlib.Path, int], pathlib.Path] = lambda p, __: p,
+    ) -> None:
+        """
+        Upload files generated for consumption by Tensorboard to checkpoint storage.
+        :param selector: optional function returning True for a file that should be included.
+        If not provided, all files are uploaded.
+        :param mangler: optional function modifying the destination file names based on rank.
+        """
+        logger.info("upload tb profile")
+        self._tensorboard_manager.sync(selector, mangler, self._distributed.rank)
```
rb-determined-ai, line 121: question: should upload_tensorboard_metrics be allowed in non-AUTO mode?

harness/determined/tensorboard/build.py
```diff
@@ -27,18 +27,18 @@ def get_rank_if_horovod_process_else_return_zero() -> Optional[int]:
...
-
-    if manager or rank == 0:
-        # In a distributed training job the manager should monitor the chief
-        # trials logs and ignore all other trials.
-        return pathlib.Path("/", "tmp", "tensorboard")
+        base_path = pathlib.Path(checkpoint_config["base_path"])
+    else:
+        base_path = pathlib.Path("/", "tmp")
```
rb-determined-ai, line 37: while we're here, we can just delete the `checkpoint_config.get("base_path")` case, it's _really_ old.

harness/determined/core/_train.py
mpkouznetsov: `upload_tensorboard_files` is there specifically for the non-AUTO (MANUAL) mode. I am confused by the question.

harness/determined/tensorboard/metric_writers/callback.py
```diff
@@ -34,6 +35,7 @@ def on_train_step_end(
         metrics: Dict[str, Any],
         batch_metrics: Optional[List[Dict[str, Any]]] = None,
     ) -> None:
+        logging.debug("Write training metrics")
```
rb-determined-ai, line 38: nit: can you make this log message elaborate that this is a tensorboard thing?

harness/determined/tensorboard/metric_writers/callback.py
```diff
@@ -53,6 +55,7 @@ def on_train_step_end(
         self.writer.reset()
 
     def on_validation_step_end(self, steps_completed: int, metrics: Dict[str, Any]) -> None:
+        logging.debug("Write validation metrics")
```
rb-determined-ai, line 58: ditto

harness/tests/tensorboard/test_util.py
```diff
@@ -1,13 +1,25 @@
...
+import pytest
 
 import determined as det
 from determined.tensorboard import SharedFSTensorboardManager, get_base_path, get_sync_path
 from determined.tensorboard.metric_writers import util as metric_writers_util
+from determined.tensorboard.util import get_rank_aware_path
 
 BASE_PATH = pathlib.Path(__file__).resolve().parent.joinpath("fixtures")
 
+profiler_file_extensions = [
```
rb-determined-ai, line 13: why do we have this defined once in tests and once in the main code?

harness/determined/core/_tensorboard_mode.py
```diff
@@ -0,0 +1,14 @@
+import enum
+
+
+class TensorboardMode(enum.Enum):
```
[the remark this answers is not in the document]
mpkouznetsov: I prefer to leave it here. I tried putting it in _train.py but that lead to circular imports because `TensorboardMode` is used by `TrainContext`, `CheckpointContext`, and `_context.core.init()`, and importing it in `checkpoint.py` from `core._train` messes things up.

rb-determined-ai (commented):
We still need to give the Trial API a `context.get_tensorboard_path()` call, so users can stop writing to /tmp/tensorboard, and we also need to configure `/tmp/tensorboard` to be a symlink in determined containers (for backwards compatibility).

Overall though, this looks fantastic.

harness/determined/core/_tensorboard_mode.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    ``TensorboardMode`` defines how Tensorboard artifacts are handled.
```
rb-determined-ai, line 6: tiniest nit: add an empty line after this line, so that there is a paragraph separation in the rendered docs.

harness/determined/core/_train.py
rb-determined-ai: sorry, I meant to say "in AUTO mode".

harness/determined/tensorboard/shared.py
```diff
@@ -1,10 +1,13 @@
+import logging
 import os
 import pathlib
 import shutil
-from typing import Any
+from typing import Any, Callable
 
 from determined.tensorboard import base
 
+logger = logging.getLogger("determined.core")
```
rb-determined-ai, line 9: This should probably be `logging.getLogger("determined.tensorboard")`

harness/determined/tensorboard/build.py
mpkouznetsov: A few tests apparently depend on the ability to override the tensorboard path.

harness/determined/exec/prep_container.py
```diff
@@ -61,6 +61,11 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
         model_def.extractall(path=constants.MANAGED_TRAINING_MODEL_COPY)
         model_def.extractall(path=".")
 
+    # pre-0.18.2 code wrote tensorboard stuff under /tmp/tensorboard
+    tensorboard_path_on_chief = f"/tmp/tensorboard-{info.allocation_id}-0"
+    os.makedirs(tensorboard_path_on_chief, exist_ok=True)
+    os.symlink(src=tensorboard_path_on_chief, dst="/tmp/tensorboard")
```
rb-determined-ai, line 67: This will raise a os.FileExists error any time that the user has `/tmp` mounted (either manually or with slurm).

I would recommend in the os.FileExists case, we override what's there, because I think that will result in the least offensive behavior; it would be annoying if somebody bind-mounted /tmp and the same symlink persisted through many, many trials.

harness/determined/exec/prep_container.py
```diff
@@ -61,6 +61,11 @@ def trial_prep(info: det.ClusterInfo, cert: certs.Cert) -> None:
         model_def.extractall(path=constants.MANAGED_TRAINING_MODEL_COPY)
         model_def.extractall(path=".")
 
+    # pre-0.18.2 code wrote tensorboard stuff under /tmp/tensorboard
```
rb-determined-ai, line 64: I think this will be pre 0.18.3 code, right?  This PR won't land in the 0.18.2 release.

There may be other instances of "0.18.2" in the content of this pr, come to think of it.

rb-determined-ai (commented):
so close!

harness/determined/util.py
```diff
@@ -330,6 +330,20 @@ def legacy_trial_entrypoint_to_script(trial_entrypoint: str) -> List[str]:
...
+    try:
+        os.symlink(src, dst, target_is_directory=True)
+    except FileExistsError:
+        if os.path.islink(dst):
+            os.unlink(dst)
+        elif os.path.isfile(dst):
+            os.unlink(dst)
+        else:
+            shutil.rmtree(dst)
+        os.symlink(src, dst, target_is_directory=True)
```
rb-determined-ai, line 344: maybe consider a try/except:pass on the second one?  Because if two workers start at the same time, it's ok if one of the second time.

PR #4149, opened 2022-05-16
fix: return user data when new user is created [DET-7255]

mapmeld:
## Description

Bug in POST users endpoint returns an error because the newly created user cannot be found.

This PR creates a user from the Python bindings without returning an error code or error message. Includes John's e2e_test for creating a new user.

Replaces #4126 (adds only one commit no longer requiring users.id)

## Checklist


ioga (approved):
lgtm, although the CI doesn't seem to be doing well

PR #4150, opened 2022-05-17
feat: allow NaN validation metrics [DET-7177]

dzhu:
## Description

In several places, the harness had checks to filter out NaN validation
metric values, which previously could not be handled. This change
removes the checks, and testing has seemed to indicate that NaN values
are serialized and otherwise handled just fine without causing any
errors.

As a historical note, I found while tracing the history of this code
that NaN values were originally considered an error, then changed to not
being an error, then changed back again. The comment I changed was
actually incorrect until just now, as it applied to an older version of
the code.

## Test Plan

- [x] add test for NaN validation metrics

## Commentary

I noticed while

### DIFF
--- a/e2e_tests/tests/experiment/test_noop.py
+++ b/e2e_tests/tests/experiment/test_noop.py
@@ -59,6 +59,21 @@ def test_noop_pause() -> None:
 
 
 @pytest.mark.e2e_cpu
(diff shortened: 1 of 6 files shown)


ioga (approved):
cc @rb-determined-ai for vis

PR #4154, opened 2022-05-17
feat: k8s detect non-det tasks

eecsliu:
## Description
Allow `det slots list` and the WebUI to accurately show how many slots are available after accounting for non-determined tasks.

## Test Plan
Deploy a k8s cluster and, outside of determined, submit several pods that request GPUs. Check the WebUI and `det slot list` to make sure the cluster page, "Available GPUs", and the CLI all show accurate slot counts.

## Checklist

### DIFF
--- a/harness/determined/cli/agent.py
+++ b/harness/determined/cli/agent.py
@@ -78,15 +78,16 @@ def list_slots(args: argparse.Namespace) -> None:
     }
 
     def get_task_name(containers: Dict[str, Any], slot: Dict[str, Any]) -> str:
-        if slot["container"] and slot["container"]["id"] in containers:
-            return str(containers[slot["container"]["id"]]["name"])
+        container_id = slot["container"]["id"]
+        if slot["container"] and container_id in containers:
+            return str(containers[container_id]["name"])
         if slot["container"] and (
-            "determined-master-deployment" in slot["container"]["id"]
-            or "determined-db-deployment" in slot["container"]["id"]
+            "determined-master-deployment" in container_id
+            or "determined-db-deployment" in container_id
         ):
-            return "Determined System Task"
+            return f"Determined System Task: {container_id}"
         if slot["container"]:
-            return "Non-Determined Task"
+            return f"Non-Determined Task: {container_id}"
         return "FREE"
 
     slots = [
@@ -100,8 +101,8 @@ def list_slots(args: argparse.Namespace) -> None:
                 (
                     "allocation_id",
                     c_names[slot["container"]["id"]]["allocation_id"]
-                    if slot["container"] and slot["container"]["id"] in c_names
-                    else (slot["container"]["id"] if slot["container"] else "FREE"),
+                    if slot["container"]
+                    else "FREE",
                 ),
                 ("task_name", get_task_name(c_names, slot)),
                 ("type", slot["device"]["type"]),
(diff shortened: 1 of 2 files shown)


master/internal/resourcemanagers/kubernetes/pods.go
```diff
@@ -614,11 +641,59 @@ func (p *pods) summarize(ctx *actor.Context) map[string]model.AgentSummary {
...
+	}
+
+	for _, p := range nonDetPods {
+		if _, ok := nodeToTasks[p.Spec.NodeName]; !ok {
+			continue
+		}
+		reqs := 0
+		for _, c := range p.Spec.Containers {
+			if deviceType == device.CPU {
+				reqs += int(c.Resources.Requests.Cpu().Value())
```
ioga, line 688: this is a number of vCPUs but we have `p.slotResourceRequests.CPU` to define how many real vCPUs are in one determined slot:
https://github.com/determined-ai/determined/blob/master/master/internal/resourcemanagers/kubernetes/pods.go#L566

so perhaps this should similarly do:

```suggestion
				reqs += int64(float32(c.Resources.Requests.Cpu().MilliValue()) / (1000. * p.slotResourceRequests.CPU))
```

maybe even move that math to helper

harness/determined/cli/agent.py
```diff
@@ -76,6 +76,18 @@ def list_slots(args: argparse.Namespace) -> None:
...
+    def get_task_name(containers: Dict[str, Any], slot: Dict[str, Any]) -> str:
+        if slot["container"] and slot["container"]["id"] in containers:
+            return str(containers[slot["container"]["id"]]["name"])
+        if slot["container"] and (
+            "determined-master-deployment" in slot["container"]["id"]
+            or "determined-db-deployment" in slot["container"]["id"]
+        ):
+            return "Determined System Task"
+        if slot["container"]:
+            return "Non-Determined Task"
```
ioga, line 88: since we already know this information, would it be nicer for the user if we actually print the task name, prefixed with these, e.g.:

`Determined system task: determined-master-deployment`
and `Non-determined task: user-container-name`

then they'll immediately know *what* is taking the slot.

eecsliu: Currently have the name of the pod as `allocation-id` but I'll change it to read this and for `allocation-id` to read "OCCUPIED" for non-det tasks.

master/internal/resourcemanagers/kubernetes/pods.go
```diff
@@ -597,6 +602,28 @@ func (p *pods) summarize(ctx *actor.Context) map[string]model.AgentSummary {
...
 
+		for _, taskName := range nodeToTasks[node.Name] {
+			for i := int64(0); i < taskSlots[taskName]; i++ {
+				if curSlot >= int(numSlots) {
+					ctx.Log().Warnf("too many pods mapping to node %s", node.Name)
+					continue
+				}
+
+				slotsSummary[strconv.Itoa(curSlot)] = model.SlotSummary{
+					ID:      strconv.FormatInt(i, 10),
```
ioga, line 613: I think `Itoa` is a more popular way to this in out codebase?
```suggestion
					ID:      strconv.Itoa(i),
```

ioga (approved):

master/internal/resourcemanagers/kubernetes/pods.go
eecsliu: yeah `Itoa` doesn't work with int64? So I used `FormatInt` instead

PR #4156, opened 2022-05-17
chore: update deepspeed launcher to work with newer versions

liamcli:
## Description
There have been changes to the deepspeed launcher in newer versions that break compatibility with our launcher.  This PR addresses those compatibility issues introduced in 

- https://github.com/microsoft/DeepSpeed/pull/1832
- https://github.com/microsoft/DeepSpeed/pull/1907

Using deepspeed versions > 0.6.4 will depend on this upstream PR https://github.com/microsoft/DeepSpeed/pull/1957

## Test Plan
Tested single and multinode for 0.5.10 and 0.6.5.dev with the PR above

## Checklist

### DIFF
--- a/harness/determined/launch/deepspeed.py
+++ b/harness/determined/launch/deepspeed.py
@@ -6,16 +6,19 @@ It launches the entrypoint script using DeepSpeed's launch process.
 import argparse
 import logging
 import os
-import pathlib
+import shlex
 import subprocess
 import sys
 import tempfile
 import time
-from typing import List
+from typing import List, Optional
 
+import deepspeed
 from deepspeed.launcher.runner import DEEPSPEED_ENVIRONMENT_NAME
+from packaging import version
 
 import determined as det
+import determined.common
 from determined import constants, util
 from determined.common import api
 from determined.common.api import certs
@@ -43,7 +46,9 @@ def is_nccl_socket_ifname_env_var_set() -> bool:
         return True
 
 
-def get_hostfile_path() -> str:
+def get_hostfile_path(multi_machine: bool) -> Optional[str]:
+    if not multi_machine:
+        return None
 
     global hostfile_path
 
@@ -69,8 +74,11 @@ def get_hostfile_path() -> str:
     return hostfile_path
 
 
+deepspeed_version = version.parse(deepspeed.__version__)
+
+
 def create_hostlist_file(
-    hostfile_path: pathlib.Path, num_proc_per_machine: int, ip_addresses: List[str]
+    hostfile_path: Optional[str], num_proc_per_machine: int, ip_addresses: List[str]
 ) -> str:
     trial_runner_hosts = ip_addresses.copy()
     # In the single node case, deepspeed doesn't use pdsh so we don't need to launch sshd.
@@ -78,10 +86,11 @@ def create_hostlist_file(
     if len(ip_addresses) == 1:
         trial_runner_hosts[0] = "localhost"
 
-    os.makedirs(hostfile_path.parent, exist_ok=True)
-    with open(hostfile_path, "w") as hostfile:
-        lines = [f"{host} slots={num_proc_per_machine}\n" for host in trial_runner_hosts]
-        hostfile.writelines(lines)
+    if hostfile_path is not None:
+        os.makedirs(os.path.dirname(hostfile_path), exist_ok=True)
+        with open(hostfile_path, "w") as hostfile:
+            lines = [f"{host} slots={num_proc_per_machine}\n" for host in trial_runner_hosts]
+            hostfile.writelines(lines)
     return trial_runner_hosts[0]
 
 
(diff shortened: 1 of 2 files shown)


harness/determined/launch/deepspeed.py
```diff
@@ -7,34 +7,38 @@
...
     # Instead, the deepspeed launcher will use localhost as the chief worker ip.
     if len(ip_addresses) == 1:
         trial_runner_hosts[0] = "localhost"
 
-    os.makedirs(hostfile_path.parent, exist_ok=True)
-    with open(hostfile_path, "w") as hostfile:
-        lines = [f"{host} slots={num_proc_per_machine}\n" for host in trial_runner_hosts]
-        hostfile.writelines(lines)
+    if hostfile_path is not None:
+        os.makedirs(pathlib.Path(hostfile_path).parent, exist_ok=True)
```
rb-determined-ai, line 38: nit: stick to either str or pathlib, don't go back and forth.

Replace `pathlib.Path(hostfile_path).parent` with `os.path.dirname(hostfile_path)`

harness/determined/launch/deepspeed.py
```diff
@@ -100,29 +104,42 @@ def create_deepspeed_env_file() -> None:
         environ = os.environ.copy()
         for k, v in environ.items():
             if k in INCLUDE:
-                f.write(f"{k}={v}\n")
+                if deepspeed_version >= version.parse("0.6.2"):
```
rb-determined-ai, line 107: when you introduce a version check like this, can you include a comment explaining the difference and motivation?

harness/determined/launch/deepspeed.py
```diff
@@ -100,29 +104,42 @@ def create_deepspeed_env_file() -> None:
...
-        "--no_python",
-        "--no_local_rank",
-        "--",
-    ]
+    deepspeed_process_cmd = ["deepspeed"]
+    if hostfile_path is not None:
+        deepspeed_process_cmd += ["-H", hostfile_path]
+    deepspeed_process_cmd += ["--master_addr", master_address, "--no_python", "--no_local_rank"]
+    if deepspeed_version > version.parse("0.6.4"):
+        deepspeed_process_cmd.append("--no_ssh_check")
```
rb-determined-ai, line 120: question: how did we avoid the need for this flag before?

harness/determined/launch/deepspeed.py
```diff
@@ -100,29 +104,42 @@ def create_deepspeed_env_file() -> None:
...
 
 
+def check_deepspeed_version(multi_machine: bool) -> None:
+    if multi_machine:
+        compatible = True
+        if deepspeed_version > version.parse("0.6.0") and deepspeed_version <= version.parse(
+            "0.6.4"
+        ):
+            compatible = False
+        assert compatible, f"launcher incompatible with deepspeed version {deepspeed.__version__}"
```
rb-determined-ai, line 132: nit: simply the logic/indentation here:

```python
def check_deepspeed_version(multi_machine: bool) -> None:
    if not multi_machine:
        return
    if deepspeed_version <= version.parse("0.6.0"):
        return
    if deepspeed_version > version.parse("0.6.4"):
        return
    raise ValueError(...)
```

Also, expand the error message to explain which versions _are_ allowed (so the user knows what to do) and explain why the disallowed versions are disallowed (so the user understands why we can't support them).

rb-determined-ai (commented):
question: is there any benefit to landing this before an upstream release with `--no_ssh_check` is cut?

harness/determined/launch/deepspeed.py
liamcli: one of the PRs linked above recently added ssh check to the deepspeed launcher but it doesn't have all the flags we require for ssh (i.e. StrictHostKeyChecking=no)

liamcli:
> question: is there any benefit to landing this before an upstream release with `--no_ssh_check` is cut?

i think better to wait.  i'll also update our environment image once that is released.

rb-determined-ai:
@liamcli is [now](https://pypi.org/project/deepspeed/#history) good?

harness/determined/launch/deepspeed.py
```diff
@@ -153,28 +162,53 @@ def create_deepspeed_env_file() -> None:
...
+    deepspeed_process_cmd.append("--")
     return deepspeed_process_cmd
 
 
+def check_deepspeed_version(multi_machine: bool) -> None:
+    if not multi_machine:
+        return
+    # Upstream deepspeed added an ssh check from 0.6.1 onwards that does not have the
+    # StrictHostKeyChecking=no arg and fails with our agents.  A PR adding a no_ssh_check arg
+    # to bypass this should land for versions 0.6.5 and onwards.
```
rb-determined-ai, line 191: it already did land, right?

rb-determined-ai (approved):
nice

harness/determined/launch/deepspeed.py
liamcli: yes, latest version will work with our code

PR #4159, opened 2022-05-18
fix: avoid potential race between AllocationReady and Running state

jerryharrow:
## Description

AllocationReady comes from the container for shells, notebooks,
and tensorboards, but the clients also depend upon the state
being running immediately after AllocationReady, so to avoid
a race condition where the dispatcher RM discovers the running
state by polling for dispatcher RM, force the state to running upon reception
of the AllocationReady event.

## Test Plan

Manually verified that `det shell start` now works with slurm dispatcher RM.

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -213,6 +213,10 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 	// These messages allow users (and sometimes an orchestrator, such as HP search)
 	// to interact with the allocation. The usually trace back to API calls.
 	case AllocationReady:
+		// AllocationReady only comes from the running container, so to
+		// avoid a race condition with the slower transition to running state
+		// which comes via polling for dispatcher RM, move the state to running now.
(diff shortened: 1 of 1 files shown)


master/internal/task/allocation.go
```diff
@@ -217,6 +217,10 @@ func (a *Allocation) Receive(ctx *actor.Context) error {
 		if err := a.db.UpdateAllocationState(a.model); err != nil {
 			a.Error(ctx, err)
 		}
+		// AllocationReady only comes from the running container, so to
+		// avoid a race condition with the slower transition to running state
+		// which comes via polling for dispatcher RM, move the state to running now.
+		a.setMostProgressedModelState(model.AllocationStateRunning)
```
stoksc, line 223: should be moved up a line so it is persisted.

stoksc (approved):

stoksc:
landing through lint fixes, since they are fixed on master as of ~30m ago

PR #4163, opened 2022-05-18
chore: check git state in lock-api-state.sh

stoksc:
## Checklist

### DIFF
--- a/tools/scripts/lock-api-state.sh
+++ b/tools/scripts/lock-api-state.sh
@@ -1,5 +1,11 @@
 #!/bin/bash -ex
 
+# check for dirty changes
+if [[ -n "$(git status --porcelain)" ]]; then
+    echo "untracked or dirty files are not allowed, cleanup before running lock-api-state.sh"
+    exit 1
+fi
+
 ## lock in current protobuf state
 
 # make gen-buf-image ensures that it starts with a clean git state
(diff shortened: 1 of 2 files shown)


ioga (approved):

tools/scripts/lock-api-state.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash -ex
 
+# check for dirty changes
```
dzhu, line 3: There's a check already in the Makefile, I just... commented it out at some point, I guess. (Also, I think this condition 
is reversed?)

stoksc: uhhh.. what makefile? I couldn't find anything calling this script?

dzhu: In `proto/Makefile` under `gen-buf-image` (which this calls, rather than the other way around).

stoksc: oh I see, it doesn't call the script, but the gen step has the check

stoksc: I will remove from the make file and leave it here. I remember it being annoying in the make file - when you want to make a breaking change, you know you have uncommitted changes, but when you are cutting a release, you want the check.

stoksc: look good to you @dzhu ?

dzhu: Makes sense!

PR #4169, opened 2022-05-18
docs: document database SSL options

dzhu:
## Test Plan

- [x] render and examine docs


neilconway (approved):
lgtm. Would it make sense to describe what the default SSL behavior is so the user doesn't need to consult the PG docs to figure that out, or give an example of how to configure a reasonably-secure Postgres DB connection?

dzhu (commented):
I put in a bit about defaults and which mode values ensure that SSL is used, but I feel like it's not the place to put in _too_ much about how SSL works in general.

Also changed the links to `current`. Good catch!

PR #4172, opened 2022-05-18
fix: sync slot usage for k8s [DET-7350]

gt2345:
## Description

Agents for K8s clusters do not have resource pool, and we assume that k8s clusters only have 1 resource pool named 'kubernetes'. 

## Test Plan

On k8s cluster, run tasks and verify slot allocation bar on resource pool card sync with overall slot allocation.

### DIFF
--- a/docs/release-notes/7073-cluster-ui-improvement
+++ b/docs/release-notes/7073-cluster-ui-improvement
@@ -2,5 +2,6 @@
 
 **Improvements**
 
-- Combine sidebar options `Cluster`, `Job Queues` and `Cluster Logs` into `Cluster` page.
+- Centralize sidebar options `Cluster`, `Job Queues` and `Cluster Logs` into `Cluster` page for a simplified layout.
+- In order to provide a more precise view of resource pools, new fields like `accelerator` and `warm slots` have been added.
 - Clicking on resource pool cards will lead to a detail page, which also includes a `Stats` tab showing average queued time by day. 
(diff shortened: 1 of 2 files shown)


trentwatt:
Whups.. that was actually me (was messing around with setting my `.gitconfig` to Emily's email as a way of giving her attribution for `workspaces` commit, but then we decided it was better just to add her as a co-author, and I forgot to change it back.

And also... double whups, I thought when I pushed it was going to set up a branch for me on my fork with the same name, but apparently I was able to push to your branch... as Emily.. and so then I reverted... and then reverted changes.... then force pushed back.... oh god :anguished: 

Anyway, I had some suggested changes I was going to open a PR to your branch to indicate, I will mention those in the review comment.

trentwatt (approved):
LGTM, this resolves the main issue.

There was another sync issues as well though unless im misinterpreting.

The cluster allocation bar says 3/8 CUDA slots allocated, 5 free.

The resource pool bar looks the exact same (and there is only one resource pool). 

So I would expect resource pool text to say: 3 slots allocated, 5 free. But it says 0 allocated, 8 free.

Screen Shot 2022-05-19 at 1 27 58 AM

It looks like the resource pool text considers it allocated if it has ResourceState.Running, and free if it doesn't.

BUT the cluster allocation bar text considers it allocated if it is `Assigned`, `Pulling`, `Starting`, or `Pending`, and free otherwise.

I opened a PR with some change I think might fix, but I'm not very familiar with this part of code.

PR #4177, opened 2022-05-19
chore: delete "conditional" json-schema extension

rb-determined-ai:
Use json-schema Draft 7's if-then-else instead.

I suspect my original intention was to target Draft 4, which didn't have
if-then-else, but I don't fully remember.  In any case, the
"conditional" extension is pure tech-debt at this point.


rb-determined-ai:
@cla-bot[bot] check

PR #4178, opened 2022-05-19
fix: agent can now be control-C while connecting to master [DET-6287]

NicholasBlaskey:
## Description

Previously the agent was unable to be canceled while connecting to master. This makes it so it can be.

## Test Plan

Run
```
./determined-agent --master-host 34.215.4.95 --master-port 8080
```

and verify control-C closes the agent

## Commentary (optional)

This fix won't cause the ```ctx.Self().Stop()``` to be called or the agent message to be printed. Not sure how important that is because the agent connection to master happens before the agent is initialized completely.  

## Checklist


ioga (approved):
nice

PR #4191, opened 2022-05-20
feat: Deleting workspace/project moves its experiments [DET-7368]

mapmeld:
## Description

After our discussion about when experiments can/can't be safely deleted and how that process is async / not easy to do en masse, when a project is deleted we're going to move its experiments back to the Uncategorized workspace.

This updates the language in the CLI and WebUI to inform users. Notes on a project get deleted, so I added that in.

## Checklist


trentwatt (approved):
cccbcrkgnncsfttjgdlufgdjnhrbihjgtdiblhrvjbk

PR #4192, opened 2022-05-20
feat: Return summarized metrics to visualize a trial

mapmeld:
## Description

As part of the larger development of the Metrics API, this adds an API endpoint where we only the data points needed to visually represent a logged metric.  Unlike previous endpoints which involved downsampling, it's based on an individual trial, you can request multiple metrics, and you can include training and validation metrics in a single API call.

Requests have the format:

```
/api/v1/trials/1200/summarize
?maxDatapoints=100
&endBatches=100000
&metricNames=loss,error
&metricType=METRIC_TYPE_UNSPECIFIED
```

## Test Plan

- Should have tests for different start / end batches
- Should have tests for a metric name appearing in training AND validation
- Should have tests for a metric name appearing in training XOR validation

## Commentary

Developing frontend / chart view

## Checklist


webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -29,36 +29,63 @@ const getChartMetricLabel = (metric: MetricName): string => {
...
+  //     const summ = await getTrialSummary({
+  //       endBatches: Math.ceil(max),
+  //       maxDatapoints: 30,
+  //       metricNames: metricNames,
+  //       startBatches: Math.floor(min),
+  //       trialId: trialId,
+  //     });
+  //     setTrialSummary(summ.metrics);
+  //   }
+  // };
```
hkang1: **chore:** drop, if it remains a comment

webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -81,7 +108,7 @@ const TrialChart: React.FC<Props> = ({
       key: trialId,
       legend: { show: false },
       plugins: [ tooltipsPlugin(), trackAxis() ],
-      scales: { x: { time: false }, y: { distr: scale === Scale.Log ? 3 : 1 } },
```
hkang1, line 84: I see, so the numeric scale value was actually mapped to uPlot's scale?

webui/react/src/pages/TrialDetails/TrialChart.tsx
mapmeld: 👍  This resetZoom code is back

webui/react/src/pages/TrialDetails/TrialChart.tsx
mapmeld: originally I was borrowing that, in the end I kept this as-is and use MetricType on the API

mapmeld:
Closing in favor of a new PR to support Single Trial chart + workloads

PR #4195, opened 2022-05-20
ci: make managed devcluster tests run in parallel

dzhu:
## Description

These tests have been the longest part of PR E2E tests, so let's speed
them up.

## Test Plan

- [ ] see how long the CI takes


dzhu:
Not that helpful yet because there are only two relevant test files and CircleCI does the splitting by file. I'm not sure if a more fine-grained splitting is possible or if I'd need to break up the files; I'll investigate at some point.

dzhu:
See #4193.

PR #4197, opened 2022-05-20
fix: allow `internal: null` for pre-0.15.6 experiments

rb-determined-ai:
The internal field was removed with 0.18.0, along with the Native API.
But before 0.15.6, the `internal` field did not have an `omitempty` tag,
so it was still being stored in the database as `null`.  This causes
forking and certain REST API endpoints to fail on older experiments.

## Test Plan

Find a pre-0.15.6 experiment on a test cluster and see if the webui is broken for it or not.


trentwatt (approved):
doge

PR #4200, opened 2022-05-23
fix: pool detail tab count

gt2345:
## Checklist

### DIFF
--- a/webui/react/src/pages/ResourcepoolDetail.tsx
+++ b/webui/react/src/pages/ResourcepoolDetail.tsx
@@ -156,10 +156,10 @@ const ResourcepoolDetail: React.FC = () => {
           defaultActiveKey={tabKey}
           destroyInactiveTabPane={true}
           onChange={handleTabChange}>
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/ResourcepoolDetail.tsx
```diff
@@ -156,10 +156,10 @@ const ResourcepoolDetail: React.FC = () => {
           defaultActiveKey={tabKey}
           destroyInactiveTabPane={true}
           onChange={handleTabChange}>
-          <TabPane key="active" tab={`${poolStats?.stats.scheduledCount} Active`}>
+          <TabPane key="active" tab={`${poolStats?.stats.scheduledCount || 0} Active`}>
```
trentwatt, line 159: What do you think about doing
```suggestion
          <TabPane key="active" tab={`${poolStats?.stats.scheduledCount ?? ''} Active`}>
```
here? 

That way when it's loading, it's blank instead of saying 0.


Also nit: prefer `??` over `||` for specifying fallback values for undefined.

trentwatt (approved):

PR #4204, opened 2022-05-23
feat: Add theme toggle to user settings [DET-7321]

julian-determined-ai:
## Description

feat: Add theme toggle to user settings [DET-7321]

An additional list item has been added to the user settings to allow users to switch the `theme` for the webui.

### Updates:
- Added a theme toggle with the options `Light Mode`, `Dark Mode` and `System Mode`
- The `System Mode` is the default option. 
- Currently the setting is being stored on the client side, with the idea being that the user's choice will be stored on the backend once available. 

## Checklist

[DET-7321]: https://determinedai.atlassian.net/browse/DET-7321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/hooks/useTheme.ts
```diff
@@ -51,13 +52,14 @@ const updateAntDesignTheme = (path: string) => {
  * and storybook Theme decorators and not individual components.
  */
 export const useTheme = (): ThemeHook => {
+  const currentTheme = (() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
```
hkang1, line 55: **super nit:** `currentMode`? theme I think of as the actual theme data we return.

webui/react/src/components/ThemeToggle.tsx
```diff
@@ -0,0 +1,91 @@
...
+    next: ThemeClass.LIGHT,
+  },
+};
+
+const ThemeToggle: React.FC = () => {
+
+  const {
+    settings,
+    updateSettings,
+  } = useSettings<Settings>(config);
```
hkang1, line 39: **suggestion:** I think it'll be better to have `useSettings` inside `useTheme`. Currently, if for whatever reason, `ThemeToggle` isn't included, the theme will not be properly pulled in from local storage (or user preferences in the near future).

webui/react/src/hooks/useTheme.ts
```diff
@@ -51,13 +52,14 @@ const updateAntDesignTheme = (path: string) => {
...
   const [ branding, setBranding ] = useState(BrandingType.Determined);
-  const [ mode, setMode ] = useState(() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
-
+  const [ mode, setMode ] = useState(currentTheme);
+  const [ scheme, setScheme ] = useState<DarkLight>(currentTheme);
```
hkang1, line 58: **suggestion:** Let's call this `SystemMode` and `setSystemMode`, where it's basically to keep track of what the OS is set to.

webui/react/src/hooks/useTheme.ts
```diff
@@ -51,13 +52,14 @@ const updateAntDesignTheme = (path: string) => {
  * and storybook Theme decorators and not individual components.
  */
 export const useTheme = (): ThemeHook => {
+  const currentTheme = (() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
   const [ branding, setBranding ] = useState(BrandingType.Determined);
-  const [ mode, setMode ] = useState(() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
-
+  const [ mode, setMode ] = useState(currentTheme);
```
hkang1, line 57: **suggestion:** we might need a separate type that's different from `DarkLight` because we want to support `Auto` (which mac provides). Maybe something like:
```
export enum Mode {
  Auto = 'auto',
  Dark = DarkLight.Dark,
  Light = DarkLight.Light,
}
```

webui/react/src/hooks/useTheme.ts
```diff
@@ -7,6 +7,7 @@ import { BrandingType } from 'types';
 
 type ThemeHook = {
   mode: DarkLight,
+  scheme: DarkLight,
```
hkang1, line 10: **chore:** if we update `mode` to handle `Auto` (see comment below), we should be ok without `scheme` as a prop.

webui/react/src/hooks/useTheme.ts
```diff
@@ -7,6 +7,7 @@ import { BrandingType } from 'types';
...
   setBranding: Dispatch<SetStateAction<BrandingType>>,
   setMode: Dispatch<SetStateAction<DarkLight>>,
   theme: Theme,
```
hkang1, line 13: **chore:** So I think the code around detecting the system needs to get updated to support detection of `Auto` mode. Currently we can only distinguish between `Dark` vs `Light / Auto`. We can't tell `Light` from `Auto`. To do that, we'll need to do something like this:
```
const MATCH_MEDIA_SCHEME_DARK = '(prefers-color-scheme: dark)';
const MATCH_MEDIA_SCHEME_LIGHT = '(prefers-color-scheme: light)';
```
and instead of `getIsDarkMode`, use something like this:
```
export enum Mode {
  Auto = 'auto',
  Dark = DarkLight.Dark,
  Light = DarkLight.Light,
}

const getSystemMode = (): Mode => {
  const isDark = matchMedia?.(MATCH_MEDIA_SCHEME_DARK).matches;
  if (isDark) return Mode.Dark;

  const isLight = matchMedia?.(MATCH_MEDIA_SCHEME_LIGHT).matches;
  if (sLight) return Mode.Light;

  return Mode.Auto;
};
```

webui/react/src/hooks/useTheme.ts
```diff
@@ -51,13 +52,14 @@ const updateAntDesignTheme = (path: string) => {
  * and storybook Theme decorators and not individual components.
  */
 export const useTheme = (): ThemeHook => {
+  const currentTheme = (() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
   const [ branding, setBranding ] = useState(BrandingType.Determined);
-  const [ mode, setMode ] = useState(() => getIsDarkMode() ? DarkLight.Dark : DarkLight.Light);
-
+  const [ mode, setMode ] = useState(currentTheme);
+  const [ scheme, setScheme ] = useState<DarkLight>(currentTheme);
   const theme = useMemo(() => themes[branding][mode], [ branding, mode ]);
```
hkang1, line 59: **chore:** this `useMemo` is going to need factor in `scheme`/`systemMode`. The idea is that if the `mode` is set to `Auto` then it needs to use `scheme`/`systemMode`. If `systemMode` is also `Auto` then to keep it simple we can default to `Light`. But interestingly enough, in Mac, if it is `Auto`, it will [change through the day](https://support.apple.com/guide/mac-help/use-a-light-or-dark-appearance-mchl52e1c2d2/mac#:~:text=You%20can%20use%20a%20light,System%20Preferences%2C%20then%20click%20General%20.&text=Select%20Light%2C%20Dark%2C%20or%20Auto.).

webui/react/src/components/ThemeToggle.settings.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface Settings {
+  theme: string;
+}
+
+export enum ThemeClass {
+  SYSTEM = 'system',
+  LIGHT = 'light',
+  DARK = 'dark'
+}
```
hkang1, line 11: **nit:** this is the same idea as the `Mode` enum I mentioned throughout the review, but it might be better to have this in `useTheme` to better encapsulate theme changing code into a more central place. Idea of having this code in a shareable hook instead of a UI component.

hkang1 (commented):
It looks and works better, but I think there's a little bit of code re-org / tuning that can help clean this area up (it was missed during my first theme arch round, not something you introduced).

hkang1:
one visual nit, the label shifts when the icon changes in size, can you update it so the space the icon takes up is consistent so the label on the right doesn't move over and out?

webui/react/src/components/ThemeToggle.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const SYSTEM_MODE = 'System Mode';
+const LIGHT_MODE = 'Light Mode';
+const DARK_MODE = 'Dark Mode';
+
+const setup = () => render(<ThemeToggleContainer />);
+
+describe('ThemeToggle', () => {
+  it('System Mode is the default setting', async () => {
```
hkang1, line 29: **nit:** for our `it` tests, we want to write like like a sentence. So for this example, something like this would read better:
```
it('should have System Mode as the default setting', async () => {
```

hkang1 (approved):
This is SO cool! the animation transition is way better this time around!

Nothing blocking, just a bunch of minor nits. Nice work!

PR #4212, opened 2022-05-24
feat: enable agent side mTLS connection to master

eecsliu:
## Description

Waiting on master side changes, or can be pulled in with master side changes. 

## Checklist

### DIFF
--- a/agent/internal/agent.go
+++ b/agent/internal/agent.go
@@ -285,11 +285,20 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 		}
 	}
 
+	var certs []tls.Certificate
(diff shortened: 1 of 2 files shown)


agent/internal/agent.go
```diff
@@ -279,11 +279,21 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 		}
 	}
 
+	var certs []tls.Certificate
+	if a.Options.Security.TLS.ClientTLSEnabled() {
```
dzhu, line 283: No need to do this check, since `ReadClientCertificate` is already doing it. Should just call that and check if the return value is nil.

dzhu (approved):

PR #4217, opened 2022-05-24
style: add gofumpt

stoksc:
## Description
This change adds gofumpt, which is useful, not only for its pretty-good-formatting-opinions, but also for its tendencies to break lines nicely for developers.

## Test Plan
- [x] run tests

## Commentary (optional)
The instructions [here](https://github.com/segmentio/golines#developer-tooling-integration) for running custom tools on save should work just as well for gofumpt, too. I went with `gofumpt` because I like all the extra rules it adds and golines was too slow and had a tendency to format struct tags very weird.

## Checklist


agent/Makefile
```diff
@@ -9,6 +9,7 @@ clean:
 
 .PHONY: get-deps
 get-deps:
+	go install mvdan.cc/gofumpt@latest
```
NicholasBlaskey, line 12: can we pin this to a version for security reasons?

stoksc: sure.

stoksc: I didn't because I figured it was low risk and pinned things always just get stale until they break, but I agree security is a greater concern than that.

PR #4218, opened 2022-05-24
chore: share webui base page

hamidzr:
## Test Plan

run `src/shared/bin/test.py`

## Commentary (optional)

we could go directly to this step and use this for reviewing the part of the diff that goes out of this repo 🤔 ? https://github.com/determined-ai/shared-web/pull/4

The last time I updated shared-web I forgot to update some dependencies and I just caught these when I went to update `saas` to use the shared changes. I added a test to the `shared-web` module to manually run which checks that both products build fine. we'll want to add ci jobs to both products to run this test and catch similar issues. 

## Checklist


hamidzr:
> So, not sure why it's different, but with this PR, switching branches from `master` to `share-page` (and back) doesn't force me to restart `make live` to get it to run, which is nice.

I'm not sure. whenever I get into this situation with webpack what I do is make a redundant change. `touch src/App.tsx` does the job and doesn't make a a change that git would care about.

PR #4220, opened 2022-05-25
feat: add master-side verification for agent mTLS

dzhu:
## Description

<will add something here later, but don't feel like it at the moment>

## Test Plan

- [x] check that the master rejects unauthenticated agent WebSocket requests

### DIFF
--- a/agent/internal/options.go
+++ b/agent/internal/options.go
@@ -109,7 +109,7 @@ func (t TLSOptions) Validate() []error {
 // ReadClientCertificate returns the client certificate described by this configuration (nil if it
 // does not allow TLS to be enabled).
 func (t TLSOptions) ReadClientCertificate() (*tls.Certificate, error) {
-	if t.ClientCert != "" && t.ClientKey != "" {
+	if t.ClientCert == "" || t.ClientKey == "" {
 		return nil, nil
 	}
 	cert, err := tls.LoadX509KeyPair(t.ClientCert, t.ClientKey)
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -4,10 +4,12 @@ import (
 	"bufio"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/csv"
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -37,6 +39,7 @@ import (
 	"github.com/determined-ai/determined/master/internal/cluster"
 	"github.com/determined-ai/determined/master/internal/command"
 	"github.com/determined-ai/determined/master/internal/config"
+	"github.com/determined-ai/determined/master/internal/connsave"
 	detContext "github.com/determined-ai/determined/master/internal/context"
 	"github.com/determined-ai/determined/master/internal/db"
 	"github.com/determined-ai/determined/master/internal/elastic"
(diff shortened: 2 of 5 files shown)


master/internal/resourcemanagers/agent/agents.go
```diff
@@ -70,6 +77,18 @@ func (a *agents) Receive(ctx *actor.Context) error {
 			}
 		}
 	case api.WebSocketConnected:
+		cmuxConn := connsave.GetConn(msg.Ctx.Request().Context()).(*cmux.MuxConn)
+		// Here, we just have to check that there are any certificates at all, since the top-level TLS
+		// config verifies that any certificates that are provided are valid.
+		if tlsConn, ok := cmuxConn.Conn.(*tls.Conn); ok {
+			if a.config.RequireAuthentication && len(tlsConn.ConnectionState().PeerCertificates) == 0 {
```
ioga, line 84: btw we have a [singleton](https://github.com/determined-ai/determined/blob/master/master/internal/config/config.go#L140) for master config
so instead of passing config around into all the objects that need it, you can just use `config.GetMasterConfig().AgentRM.RequireAuthentication` in place

master/internal/resourcemanagers/setup.go
```diff
@@ -97,7 +97,7 @@ func setupAgentResourceManager(
 	)
 	system.Ask(ref, actor.Ping{}).Get()
 
-	agent.Initialize(system, echo, opts)
+	agent.Initialize(system, echo, config.ResourceManager.AgentRM, opts)
```
eecsliu, line 100: should this be `agent.Initialize(system, echo, opts, config.ResourceManager.AgentRM)` ? Looks like the function def has the config and opts reversed.

master/internal/core.go
```diff
@@ -476,10 +479,31 @@ func (m *Master) startServers(ctx context.Context, cert *tls.Certificate) error
 
 	// If configured, set up TLS wrapping.
 	if cert != nil {
+		var clientCAs *x509.CertPool
+		clientAuthMode := tls.NoClientCert
+
+		if agentRM := m.config.ResourceManager.AgentRM; agentRM != nil && agentRM.RequireAuthentication {
+			clientAuthMode = tls.VerifyClientCertIfGiven
```
eecsliu, line 486: Should we set this to `RequireAndVerifyClientCert` instead? If we request it but don't enforce that it's given, then couldn't a malicious party just bypass security by not providing a certificate?

master/internal/resourcemanagers/setup.go
dzhu: Welp, that's what I get for deciding to move things around right at the end of the day.

agent/internal/agent.go
```diff
@@ -293,6 +293,8 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 		certs = append(certs, *cert)
 	}
 
+	fmt.Println("certs:", certs)
```
eecsliu, line 296: Are we purposely printing certs?

agent/internal/options.go
```diff
@@ -109,10 +110,11 @@ func (t TLSOptions) Validate() []error {
 // ReadClientCertificate returns the client certificate described by this configuration (nil if it
 // does not allow TLS to be enabled).
 func (t TLSOptions) ReadClientCertificate() (*tls.Certificate, error) {
-	if t.ClientCert != "" && t.ClientKey != "" {
+	if t.ClientCert == "" || t.ClientKey == "" {
 		return nil, nil
 	}
 	cert, err := tls.LoadX509KeyPair(t.ClientCert, t.ClientKey)
+	fmt.Println(cert, err)
```
eecsliu, line 117: purposeful?

eecsliu (approved):

PR #4227, opened 2022-05-25
fix: det shell start/open, in windows

rb-determined-ai:
tempfile.NamedTemporaryFile does not do what you think it does, at least
not in Windows.  Use mkstemp directly instead.

Also, fix the color-related ANSI escape codes so there's not a bunch of
raw escape codes when using det in cmd.exe.

## Test Plan

Run `det shell start` in windows, see if it works and if the colors look right.

### DIFF
--- a/harness/determined/cli/cli.py
+++ b/harness/determined/cli/cli.py
@@ -1,4 +1,5 @@
 import hashlib
+import os
 import socket
 import ssl
 import sys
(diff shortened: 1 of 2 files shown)


ioga (approved):

varlogtim:
Confirmed that the colors look tasty in Windows. Also, this seems to resolve the other issue I was experiencing on Windows before in which `det shell open` wouldn't work until I manually ran the SSH command.

PR #4230, opened 2022-05-26
chore: reduce use of any in InteractiveTable

trentwatt:
## Description

left one `any` for now, maybe try to get rid of it before merge.

## Test Plan

not needed, only changed types

## Checklist

### DIFF
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -481,8 +481,7 @@ const InteractiveTable: InteractiveTable = ({
     ],
   );
 
-<<<<<<< HEAD
-  const renderColumns= useMemo(
+  const renderColumns = useMemo(
     () => {
       const columns = settings.columns.filter(columnName => columnDefs[columnName])
         .map((columnName, index) => {
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -19,6 +19,7 @@ import { throttle } from 'throttle-debounce';
 
 import useResize from 'hooks/useResize';
 import { UpdateSettings } from 'hooks/useSettings';
+import { Primitive, UnknownRecord } from 'shared/types';
 
 import Spinner from '../shared/components/Spinner/Spinner';
 
@@ -31,11 +32,6 @@ import css from './InteractiveTable.module.scss';
 export const onRightClickableCell = (): React.HTMLAttributes<HTMLElement> =>
   ({ isCellRightClickable: true } as React.HTMLAttributes<HTMLElement>);
 
-/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-type Comparable = any;
-
-/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-type RecordType = any;
 export interface InteractiveTableSettings {
   /**
      * ColumnWidths: Array of column widths, corresponding to columns array below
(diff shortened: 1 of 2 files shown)


webui/react/src/components/InteractiveTable.tsx
```diff
@@ -70,8 +66,7 @@ interface ContextMenuProps {
 export interface ColumnDef<RecordType> extends ColumnType<RecordType> {
   dataIndex: string;
   defaultWidth: number;
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  isFiltered?: (s: any) => boolean; // any extends Settings
+  isFiltered?: (s: unknown) => boolean;
```
EmilyBonar, line 69: suggestion: I think this can be something like `isFiltered?: <T extends InteractiveTableSettings>(s: T) => boolean`

webui/react/src/components/InteractiveTable.tsx
```diff
@@ -367,17 +362,15 @@ const InteractiveTable: InteractiveTable = ({
     pageWidth ]);
 
   const handleChange = useCallback(
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-    (tablePagination: any, tableFilters: any, tableSorter: any): void => {
+    (tablePagination, tableFilters, tableSorter): void => {
```
EmilyBonar, line 365: question: You should be able to pull the types from the antd Table onChange, right? In which case tablePagination is `TablePaginationConfig`, tableFilters is `Record<string | FilterValue | null>`, and tableSorter is `SorterResult<RecordType> | SorterResult<RecordType>[]`

EmilyBonar (approved):
Overall looks good

webui/react/src/components/InteractiveTable.tsx
trentwatt: tried a few variations of this, no luck.

it's definitely the right path though. I found out it's possible to use generic types in such a way that you can use the type parameters in the function body, e.g.

```
const InteractiveTable = <RecordType, Settings extends InteractiveTableSettings>({
  loading,
  scroll,
  dataSource,
  columns,
  containerRef,
  settings,
  updateSettings,
  ContextMenu,
  areRowsSelected,
  ...props
}: InteractiveTableProps<RecordType, Settings>): JSX.Element => {
```

then the specifics for `Settings` should in theory get inferred by virtue of:
```
interface InteractiveTableProps<RecordType, Settings> extends TableProps<RecordType> {
  ContextMenu?: React.FC<ContextMenuProps>;
  areRowsSelected?: boolean;
  columns: ColumnDef<RecordType>[],
  containerRef: MutableRefObject<HTMLElement | null>,
  settings: Settings;
  updateSettings: UpdateSettings<Settings>
}
```

and you can do things like 

`const column: ColumnDef<RecordType> = ...`

even though in this case `RecordType` is just a type parameter of the function.

I think ultimately this is the way to go, I tried it a bit [here](https://github.com/determined-ai/determined/pull/4291) and was able to git rid of all the `any` and type errors within `InteractiveTable` but it breaks downstream consumption, and I'm tired of messing with it for now. 

So just gonna merge and come back later.

webui/react/src/components/InteractiveTable.tsx
trentwatt: This is a situation where we would need to have `RecordType` parameter available in the function, like with the above. Will revisit.

PR #4232, opened 2022-05-26
chore: add logging for e2e gpu test

eecsliu:
## Description
e2e gpu tests were failing because the created agent was not connecting to master and tests were timing out. The agent failed to start because it was mistakenly looking for a certificate when there isn't one. Confusingly, the certificate name is an empty string. This PR adds some additional logging and a precautionary string trimming function.

## Test Plan
e2e tests should be fixed now

## Checklist

### DIFF
--- a/agent/internal/agent.go
+++ b/agent/internal/agent.go
@@ -278,7 +278,7 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 		certData, err := ioutil.ReadFile(certFile) //nolint:gosec
 		if err != nil {
 			msg := fmt.Sprintf("failed to read certificate file %q", certFile)
-			return nil, errors.Wrap(err, msg)
+			return nil, errors.Wrapf(err, msg)
 		}
 		pool = x509.NewCertPool()
 		if !pool.AppendCertsFromPEM(certData) {
@@ -291,7 +291,7 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 	case err != nil:
 		msg := fmt.Sprintf("failed to read agent certificate file %q or certificate key %q",
 			a.Options.Security.TLS.ClientCert, a.Options.Security.TLS.ClientKey)
-		return nil, errors.Wrap(err, msg)
+		return nil, errors.Wrapf(err, msg)
 	case cert != nil:
 		certs = append(certs, *cert)
 	}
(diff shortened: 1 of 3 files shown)


agent/internal/agent.go
```diff
@@ -286,11 +286,13 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 	}
 
 	var certs []tls.Certificate
-	switch cert, err := a.Options.Security.TLS.ReadClientCertificate(); {
-	case err != nil:
-		return nil, errors.Wrap(err, "failed to read agent certificate file")
-	case cert != nil:
-		certs = append(certs, *cert)
+	if agentCert := a.Options.Security.TLS.ClientCert; agentCert != "" {
```
ioga, line 289: I am somewhat groggy at the moment and a bit confused how it works. 
`ReadClientCertificate` itself already checks if `ClientCert == ""`: https://github.com/determined-ai/determined/blob/master/agent/internal/options.go#L112
and then it's supposed to `return nil, nil`, meaning this switch wouldn't hit any cases.
how does the same (?) check on top changes the code flow here?

eecsliu: yeah I'm still looking into this, but the problem was the error "failed to read agent certificate file" and after the change it's not there...

eecsliu: So I'm just more confused by the error and logically it makes no sense, so I think the best path right now is add logging and a precautionary string trimming step.

agent/internal/agent.go
```diff
@@ -277,7 +277,8 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 	if certFile := a.Options.Security.TLS.MasterCert; certFile != "" {
 		certData, err := ioutil.ReadFile(certFile) //nolint:gosec
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to read certificate file")
+			msg := fmt.Sprintf("failed to read certificate file %q", certFile)
+			return nil, errors.Wrap(err, msg)
```
dzhu, line 281: `errors.Wrapf`

agent/internal/agent.go
```diff
@@ -288,7 +289,9 @@ func (a *agent) tlsConfig() (*tls.Config, error) {
 	var certs []tls.Certificate
 	switch cert, err := a.Options.Security.TLS.ReadClientCertificate(); {
 	case err != nil:
-		return nil, errors.Wrap(err, "failed to read agent certificate file")
+		msg := fmt.Sprintf("failed to read agent certificate file %q or certificate key %q",
+			a.Options.Security.TLS.ClientCert, a.Options.Security.TLS.ClientKey)
+		return nil, errors.Wrap(err, msg)
```
dzhu, line 294: `errors.Wrapf`

dzhu (approved):

PR #4233, opened 2022-05-26
feat: gradient clipping support for `LightningAdapter`

lqf96:
## Description
Add gradient clipping support for `LightningAdapter`. Parameter names are borrowed from the latest PyTorch Lightning [`Trainer`](https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#trainer-class-api) class.

## Test Plan
(TBD; not sure where to add tests since this is my first contribution to the project)

## Commentary (optional)
Right now `_gradient_clip_val` and `_gradient_clip_algorithm` are stored within the `LightningAdapter`. Is this okay or should these be moved to other places? (e.g. `_LightningAdapterState`)

## Checklist

### DIFF
--- a/harness/determined/pytorch/lightning/_adapter.py
+++ b/harness/determined/pytorch/lightning/_adapter.py
@@ -1,6 +1,6 @@
 import inspect
 from abc import abstractmethod
-from typing import Any, Dict, List, Sequence, Tuple, Union, cast
+from typing import Any, Dict, Iterable, List, Sequence, Tuple, Union, cast
 
 import pytorch_lightning as pl
 import torch
@@ -9,6 +9,7 @@ from pytorch_lightning.trainer.optimizers import (
     _validate_scheduler_optimizer,
 )
 from pytorch_lightning.utilities.model_helpers import is_overridden
+from torch.nn.utils import clip_grad_norm_, clip_grad_value_
 from torch.optim.lr_scheduler import _LRScheduler
 from torch.optim.optimizer import Optimizer
 from typing_extensions import Literal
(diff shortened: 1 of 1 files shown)


rb-determined-ai:
Hi @lqf96, we were on a company retreat yesterday, and I am pretty busy today but I will try and get this reviewed on monday.

In the mean time, thank you for your PR and your CLA looks good.  @cla-bot[bot] check

harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -131,6 +132,8 @@ def __init__(
         precision: Union[Literal[32], Literal[16]] = 32,
         amp_backend: Union[Literal["native"], Literal["apex"]] = "native",
         amp_level: Literal["O0", "O1", "O2", "O3"] = "O2",
+        gradient_clip_val: Union[int, float, None] = None,
+        gradient_clip_algorithm: Union[Literal["norm"], Literal["value"]] = "norm",
```
rb-determined-ai, line 136: Asking users to pass these values here is a functional solution, meaning that users can use it and it will work.

But it is not a very "PytorchLightning-native" solution.  It would be really nice if we could somehow hook directly into the `configure_gradient_clipping` call directly on the LightningModule that the user sets up.

I don't think this would be terribly hard either, though from looking at LightningModule.clip_gradients, we'd have to makes sure to patch clip_gradients() to not try and read `self.trainer`.  I think it would be a very nice user experience though.

What do you think?

rb-determined-ai:
closing due to inactivity.  Happy to reopen later if @lqf96 is still interested in this work.

PR #4234, opened 2022-05-26
ci: check ulimit before 4x4 distributed test on macs

rb-determined-ai:
Default macbook ulimits cause the 4x4 test to fail.


maxrussell (approved):
LGTM

PR #4240, opened 2022-05-27
ci: add check to prevent ssh git url

rb-determined-ai:
Something keeps setting up hamidzr/omnibar as an ssh url instead of an
https url.  Fix it, add a lint checker, and a fmt fixer.


hamidzr (approved):
👍🏼

PR #4242, opened 2022-05-28
chore: experimenting with an in-container, compiled log shipper

rb-determined-ai:
Try:
```bash
make && det cmd run --context=. sh pick-orchestrator.sh echo hello world
```


rb-determined-ai:
closing unless we decide some day to write our own log shipper.

PR #4246, opened 2022-05-31
feat: add cli and api for delete checkpoints [DET-7119], [DET-7120]

nrajanee:
## Description

feat: add functionality and add an api endpoint for deleting a list of checkpoints given their uuids [DET-7119], [DET-7120]

## Test Plan

Tested well in integration and e2e tests.

[DET-7119]: https://determinedai.atlassian.net/browse/DET-7119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-7120]: https://determinedai.atlassian.net/browse/DET-7120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -45,6 +48,75 @@ def test_gc_checkpoints_lfs() -> None:
...
+    exp_2_checkpoints = bindings.get_GetExperimentCheckpoints(
+        session=test_session, id=exp_id_1
+    ).checkpoints
+    assert len(exp_1_checkpoints) > 0, f"no checkpoints found in experiment with ID:{exp_id_1}"
+    assert len(exp_2_checkpoints) > 0, f"no checkpoints found in experiment with ID:{exp_id_2}"
+
+    indices_exp_1 = [i for i in range(0, len(exp_1_checkpoints))]
+    indices_exp_2 = [i for i in range(0, len(exp_2_checkpoints))]
+    d_exp_1_checkpoint_uuids = [
+        exp_1_checkpoints[d_index].uuid for d_index in random.sample(indices_exp_1, 2)
```
stoksc, line 83: I think instead you can just do `random.sample(range(len(exp_1_checkpoints)), 2)`

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -45,6 +48,75 @@ def test_gc_checkpoints_lfs() -> None:
...
+    ]
+    d_exp_2_checkpoint_uuids = [
+        exp_2_checkpoints[d_index].uuid for d_index in random.sample(indices_exp_2, 2)
+    ]
+
+    d_checkpoint_uuids = d_exp_1_checkpoint_uuids + d_exp_2_checkpoint_uuids
+    delete_body = bindings.v1DeleteCheckpointsRequest(checkpointUuids=d_checkpoint_uuids)
+    bindings.delete_DeleteCheckpoints(session=test_session, body=delete_body)
+    wait_for_gc_to_finish(exp_id_1)
```
stoksc, line 92: we should probably call this after `run_basic_test_with_temp_config` so we can make sure we're actually checkpoint for the deletion of checkpoints that aren't getting deleted by the normal gc job.

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -45,6 +48,75 @@ def test_gc_checkpoints_lfs() -> None:
...
+def ensure_checkpoint_deleted(test_session, config, d_checkpoint_uuid) -> bool:
+    d_checkpoint = bindings.get_GetCheckpoint(
+        session=test_session, checkpointUuid=d_checkpoint_uuid
+    ).checkpoint
+    if d_checkpoint is not None:
+        assert (
+            d_checkpoint.state == determinedcheckpointv1State.STATE_DELETED
+        ), f"checkpoint with uuid {d_checkpoint_uuid} does not have a deleted state"
+    else:
+        raise Exception(
```
stoksc, line 108: here and elsewhere, maybe a good thing to use `pytest.fail` instead of `raise` to distinguish test failures from test exceptions. though we don't really hold this up all over, i was just thinking about it in tests i was writing recently.

harness/determined/cli/checkpoint.py
```diff
@@ -168,6 +182,20 @@ def describe(args: Namespace) -> None:
             "describe checkpoint",
             [Arg("uuid", type=str, help="checkpoint uuid to describe")],
         ),
+        Cmd(
+            "delete",
+            delete_checkpoints,
+            "delete checkpoints",
+            [
+                Arg("checkpoints_uuids", help="comma-separated list of checkpoints to delete"),
```
stoksc, line 190: just thinking, what is better `det c delete 1,2,3` or `det c delete 1 2 3`? thoughts?

harness/determined/exec/gc_checkpoints.py
```diff
@@ -99,8 +101,8 @@ def main(argv: List[str]) -> None:
 
     manager = storage.build(storage_config, container_path=constants.SHARED_FS_CONTAINER_PATH)
 
-    storage_ids = [c["uuid"] for c in args.delete["checkpoints"]]
-
+    logging.info(f"This is args.delete: {args.delete}")
```
stoksc, line 104: don't forget to remove

harness/determined/exec/gc_checkpoints.py
```diff
@@ -22,6 +22,8 @@ def delete_checkpoints(
     logging.info("Deleting {} checkpoints".format(len(to_delete)))
 
     for storage_id in to_delete:
+        if storage_id == "":
```
stoksc, line 25: why was this necessary? fine as a precaution but in general i would expect this validation to happen earlier

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
+package db
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+// FilterForRegisteredCheckpoints gets the deleted checkpoints provided in the model registry.
+func (db *PgDB) FilterForRegisteredCheckpoints(deleteCheckpoints []uuid.UUID) ([]uuid.UUID, error) {
```
stoksc, line 10: `deleteCheckpoints` -> `checkpoints` or `cs`.. this method has no idea they're being deleted.

stoksc: also the comment implies the checkpoints passed are deleted.

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+	var checkpointIDRows []struct {
+		ID uuid.UUID
+	}
+
+	if err := db.queryRows(`
+	SELECT DISTINCT(c.uuid) AS ID FROM checkpoints_view AS c
```
stoksc, line 16: we don't nede to join at all here, we can just `SELECT DISTINCT(mv.checkpoint_uuid) FROM model_versions AS mv WHERE mv.checkpoint_uuid IN (SELECT UNNEST($1::uuid[]))`.

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	return checkpointIDs, nil
+}
+
+// MarkCheckpointsDeleted updates the provided delete checkpoints to DELETED state.
+func (db *PgDB) MarkCheckpointsDeleted(deleteCheckpoints []uuid.UUID) error {
+	_, err := db.sql.Exec(`UPDATE raw_checkpoints c
+    SET state = 'DELETED'
+    WHERE c.uuid IN (SELECT UNNEST($1::uuid[]))`, deleteCheckpoints)
```
stoksc, line 37: we can talk more offline, but let's not use this `unnest` thing going forward. we ran into a really weird perf issue with it in a more complex query.

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+// GroupCheckpointUUIDsByExperimentID creates the mapping of checkpoint uuids to experiment id.
+// The checkpount uuids grouped together are comma separated.
+func (db *PgDB) GroupCheckpointUUIDsByExperimentID(checkpoints []uuid.UUID) (
+	[]*ExperimentCheckpointGrouping, error) {
+	var groupeIDcUUIDS []*ExperimentCheckpointGrouping
+
+	rows, err := db.sql.Queryx(
+		`SELECT e.id AS ExperimentID, string_agg(c.uuid::text, ',') AS CheckpointUUIDSStr
+	FROM experiments e
+	JOIN checkpoints_view c ON c.experiment_id = e.id
```
stoksc, line 67: also don't need a join here, `checkpoints_view` has the experiment id.

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+	WHERE c.uuid IN (SELECT UNNEST($1::uuid[]))
+	GROUP BY e.id`, checkpoints)
+	if err != nil {
+		return nil, fmt.Errorf("grouping checkpoint UUIDs by experiment ids: %w", err)
+	}
+
+	for rows.Next() {
```
stoksc, line 74: just using `queryRows` like above may save all this code

master/internal/db/postgres_experiments.go
```diff
@@ -1220,36 +1221,27 @@ WITH const AS (
...
-                           #- '{experiment_rank}' #- '{trial_rank}' #- '{trial_order_rank}'
-                       ORDER BY id ASC), '[]'::jsonb)
-            FROM selected_checkpoints
-           ) AS checkpoints
-    FROM const
-) x
-`
-
-	return db.rawQuery(ctes+query, id, experimentBest, trialBest, trialLatest)
+	if err := db.queryRows(ctes+query, &checkpointIDRows,
```
stoksc, line 1230: no reason for this string addition now, just add them statically

master/internal/db/postgres_experiments.go
```diff
@@ -1140,14 +1142,13 @@ FROM experiments
...
-// checkpoints are also marked as deleted in the database.
+// ExperimentCheckpointsToGCRaw returns a comma-separated string describing checkpoints
+// that should be GCed according to the given GC policy parameters. If the delete parameter is true,
+// the returned checkpoints are also marked as deleted in the database.
 func (db *PgDB) ExperimentCheckpointsToGCRaw(
 	id int,
 	experimentBest, trialBest, trialLatest int,
-	delete bool,
-) ([]byte, error) {
+) (string, error) {
```
stoksc, line 1151: bit of a nit but, may as well have this return `[]string` and let another layer do the serialization.

master/internal/api_checkpoint.go
```diff
@@ -28,6 +35,82 @@ func (a *apiServer) GetCheckpoint(
...
 
+func (a *apiServer) DeleteCheckpoints(
+	ctx context.Context,
+	req *apiv1.DeleteCheckpointsRequest) (*apiv1.DeleteCheckpointsResponse, error) {
+	curUser, _, err := grpcutil.GetUser(ctx, a.m.db, &a.m.config.InternalConfig.ExternalSessions)
+	if err != nil {
+		return nil, err
+	}
+
+	checkpointsToDeleteStr := req.CheckpointUuids
```
stoksc, line 46: this temp var is longer than `req.CheckpointUuids` and only used once, so let's just get rid of it

master/internal/db/postgres_checkpoints.go
```diff
@@ -0,0 +1,86 @@
...
+		ID uuid.UUID
+	}
+
+	if err := db.queryRows(`
+	SELECT DISTINCT(c.uuid) AS ID FROM checkpoints_view AS c 
+	JOIN model_versions AS mv ON mv.checkpoint_uuid = c.uuid
+	WHERE c.uuid IN (SELECT UNNEST($1::uuid[])); 
+`, &checkpointIDRows, deleteCheckpoints); err != nil {
+		return nil, fmt.Errorf(
+			"querying for all requested delete checkpoints registered in model registry: %w", err)
```
stoksc, line 21: the error is describing the callers context - let them describe that.
`fmt.Errorf("filtering checkpoint uuids (%v) by those registered in the model registry: %w", uuids, err)`

master/internal/api_checkpoint.go
```diff
@@ -28,6 +35,82 @@ func (a *apiServer) GetCheckpoint(
...
+
+	registeredCheckpointUUIDs, err := a.m.db.FilterForRegisteredCheckpoints(checkpointsToDelete)
+	if err != nil {
+		return nil, err
+	}
+
+	// return 400 if model registry checkpoints and include all the model registry checkpoints
+	if len(registeredCheckpointUUIDs) > 0 {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"This subset of checkpoints provided are in the model registry and cannot be deleted: %v.",
```
stoksc, line 62: maybe lower case, since we're sort of consistent about that elsewhere (even right above)

master/internal/api_checkpoint.go
```diff
@@ -28,6 +35,82 @@ func (a *apiServer) GetCheckpoint(
...
+			registeredCheckpointUUIDs)
+	}
+
+	addr := actor.Addr(fmt.Sprintf("checkpoints-gc-%s", uuid.New().String()))
+
+	taskSpec := *a.m.taskSpec
+	agentUserGroup, err := a.m.db.AgentUserGroup(curUser.ID)
```
stoksc, line 69: todo for us: make all the defaulting logic (like agentUserGroup here) way clearer and harder to mess  up

master/internal/api_checkpoint.go
```diff
@@ -28,6 +35,82 @@ func (a *apiServer) GetCheckpoint(
...
+	for _, expIDcUUIDs := range groupCUUIDsByEIDs {
+		exp, eErr := a.m.db.ExperimentByID(expIDcUUIDs.ExperimentID)
+		if eErr != nil {
+			return nil, err
+		}
+
+		jobSubmissionTime := time.Now().UTC().Truncate(time.Millisecond)
+		ckptGCTask := newCheckpointGCTask(a.m, jobID, jobSubmissionTime, taskSpec,
+			exp.ID, exp.Config.AsLegacy(), expIDcUUIDs.CheckpointUUIDSStr, agentUserGroup, curUser)
+		if gcErr := a.m.system.MustActorOf(addr,
```
stoksc, line 105: `gcErr` isn't an error, it's an actor ref that should never be nil (else we would panic, hence the "must" in the name). so i'm suprisied this doesn't always error?

master/internal/checkpoint_gc.go
```diff
@@ -33,6 +36,28 @@ type checkpointGCTask struct {
 	logCtx logger.Context
 }
 
+func newCheckpointGCTask(master *Master, jobID model.JobID, jobSubmissionTime time.Time,
```
stoksc, line 39: let's not pass this a ref to the master. instead just pass the db and rm

master/internal/checkpoint_gc.go
```diff
@@ -77,17 +102,31 @@ func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
...
 		if msg.Child.Address().Local() == t.allocationID.String() {
 			t.completeTask(ctx)
+			return errors.Wrapf(msg.Error, "checkpoint GC task failed (actor.ChildFailed)")
 		}
 	case actor.PostStop:
 	default:
 		return actor.ErrUnexpectedMessage(ctx)
 	}
+
+	conv := &protoconverter.ProtoConverter{}
```
stoksc, line 112: this will run on the first message that gets through the switch case (so `actor.PreStart`) which isn't right. we want to run this only once the allocation actor we kick off to run the GC job has exited successfully, which we will know when we receive the `task.AllocationExited` and it has a nil error

master/internal/experiment.go
```diff
@@ -345,7 +345,6 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 			e.Config.CheckpointStorage().SaveExperimentBest(),
 			e.Config.CheckpointStorage().SaveTrialBest(),
 			e.Config.CheckpointStorage().SaveTrialLatest(),
-			true,
 		)
 		if err != nil {
 			ctx.Log().WithError(err).Error("")
```
stoksc, line 350: we should use `newCheckpointGCTask` below, too (we may need to change the singature a tad).

master/static/migrations/20220512160917_update-job-type-jobs.tx.down.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TYPE public.job_type RENAME TO _job_type; 
+CREATE TYPE public.job_type AS ENUM (
+    'EXPERIMENT',
+    'NOTEBOOK',
+    'SHELL',
+    'COMMAND',
+    'TENSORBOARD'
+);
+ALTER TABLE public.jobs ALTER COLUMN job_type TYPE public.job_type USING (job_type::text::job_type);
+DROP TYPE _job_type;
```
stoksc, line 10: this migration isn't necessarily safe to run if we ran on this version for a while. probably just run a `DELETE FROM public.jobs WHERE job_type = 'CHECKPOINT_GC'` first or something.

stoksc (commented):
looking very good, nearly there. thanks for all the hard work on this.

harness/determined/cli/checkpoint.py
nrajanee: I think 1,2,3 looks more like a list than 1 2 3. But I don't think I have strong preference. What do most cli's do? I saw that in experiment.py (describe experiments) we use commas.

e2e_tests/tests/cluster/test_checkpoints.py
nrajanee: (messaged on slack about this).

harness/determined/exec/gc_checkpoints.py
nrajanee: true! will add a precaution earlier in cli/checkpoints.py to check if it's a valid uuid.

master/internal/db/postgres_checkpoints.go
nrajanee: oops yes! not sure why i used join, thank you!

master/internal/db/postgres_checkpoints.go
nrajanee: for queryRows in the struct I can only use identifiers such as ID. I think they need to be defined in sql? That's why I used StructScan cause then I can use an identifier that is appropriate for the values.

master/internal/db/postgres_experiments.go
nrajanee: then every method that calls this method would have to do the serialization? Isn't it better to do it here?

master/internal/api_checkpoint.go
nrajanee: I'm not sure about this. I will look into it. Even in api_experiment.go (line 215) it doesn't seem to error?

stoksc: oh wait, duh, it ends with .AwaitTermination

stoksc: if you notice, the one I wrote is in a goroutine and we await the actor's successful termination. in this one I think we just block the api call?

nrajanee: Yes that makes sense.

master/static/migrations/20220512160917_update-job-type-jobs.tx.up.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TYPE public.job_type RENAME TO _job_type;
```
nrajanee, line 1: @stoksc there are new migrations in determined dated after this. Do I need to update the dates for these?

e2e_tests/tests/cluster/test_checkpoints.py
```diff
@@ -215,4 +303,4 @@ def test_fail_on_preclose_chechpoint_save() -> None:
         config_obj,
         conf.fixtures_path("no_op"),
         error_log,
-    )
+    )
```
stoksc, line 307: don't forget to re-add this newline

harness/determined/cli/checkpoint.py
stoksc: let's just do commas then if experiment.py does that

harness/determined/cli/checkpoint.py
```diff
@@ -137,6 +137,20 @@ def describe(args: Namespace) -> None:
     render_checkpoint(checkpoint)
 
 
+@authentication.required
+def delete_checkpoints(args: Namespace) -> None:
+    if args.yes or render.yes_or_no(
+        "Deleting checkpoints will result in deletion of all data associated\n"
+        "with each checkpoint in the checkpoint storage. Do you still want to proceed?"
+    ):
+        cUUIDS = args.checkpoints_uuids.split(",")
```
stoksc, line 146:
```suggestion
        c_uuids = args.checkpoints_uuids.split(",")
```

master/internal/api_checkpoint.go
stoksc: sent message in dmg but I think we still need to drop the AwaitTermination() here

stoksc (approved):
lgtm. great work on this.

varlogtim:
This broke `det e set gc-policy` :(

stoksc:
https://determinedai.atlassian.net/browse/DET-8032

PR #4253, opened 2022-06-02
fix: gcs storage upstream test failure

NicholasBlaskey:
## Description

Fix an upstream test failure in which I didn't have the credentials set for a gcs test.

## Checklist

### DIFF
--- a/harness/tests/storage/test_gcs.py
+++ b/harness/tests/storage/test_gcs.py
@@ -43,11 +43,9 @@ def prep_gcs_test_creds(tmp_path: Path) -> Iterator[None]:
         del os.environ["GOOGLE_APPLICATION_CREDENTIALS"]
 
 
-# @pytest.fixture
 def get_live_gcs_manager(
     tmp_path: Path,
     prefix: Optional[str],
-    prep_gcs_test_creds: None,
     require_secrets: bool,
 ) -> storage.GCSStorageManager:
     """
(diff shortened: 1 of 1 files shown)


harness/tests/storage/test_gcs.py
```diff
@@ -43,11 +43,9 @@ def prep_gcs_test_creds(tmp_path: Path) -> Iterator[None]:
         del os.environ["GOOGLE_APPLICATION_CREDENTIALS"]
 
 
-# @pytest.fixture
 def get_live_gcs_manager(
```
ioga, line 46: any reason why this can't be a `pytest.fixture` like before #4238?

NicholasBlaskey: That was a comment that I accidentally left in (that got merged) before in #4238

The reason why in #4238 I changed that to a function rather than a fixture was because initializing the manager requires the ```prefix``` to be passed in, which will vary based on the test using the manager.

Honestly Python testing is pretty new to me so not sure if there is a better way to handle that case.

ioga: I'd say if you want all tests who make use of `live_gcs_manager` to run both prefix `parameters=[None, "test/prefix"]`, then it's nicer to make it into a [parametrized fixture](https://github.com/determined-ai/determined/blob/master/e2e_tests/tests/cluster/managed_cluster.py#L149)
because then all tests using it will be parametrized, and you won't have to specify `pytest.mark.parametrize` for each of them.

and if it's just a single test using these parameters, then it probably doesn't matter much, and a function call like you have here is okay.

NicholasBlaskey: Cool. In this case since it is just the single test using these parameters I think I am going to leave it as a function.

PR #4257, opened 2022-06-03
fix: don't hardcode /bin/which in entrypoints

rb-determined-ai:
This was my bug, introduced in #1967.

I have no idea why I wrote it that way.

## Test Plan

Run a notebook in an image that doesn't have /bin/which, like:

```python
det notebook start --config=environment.image=python:3.9.11-slim
```

Check the logs, ensuring the following log does NOT appear:
```
error: unable to find python3 as "python3"
```

### DIFF
--- a/docs/release-notes/4257-which.txt
+++ b/docs/release-notes/4257-which.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Supports running in containers which do not have a /bin/which path, such as python-slim. The
+   error was caused by accidentally hardcoding ``/bin/which`` instead of letting the shell find
+   ``which`` on the path.
--- a/master/static/srv/entrypoint.sh
+++ b/master/static/srv/entrypoint.sh
@@ -10,7 +10,7 @@ export PATH="/run/determined/pythonuserbase/bin:$PATH"
 if [ -z "$DET_PYTHON_EXECUTABLE" ] ; then
     export DET_PYTHON_EXECUTABLE="python3"
 fi
-if ! /bin/which "$DET_PYTHON_EXECUTABLE" >/dev/null 2>&1 ; then
+if ! which "$DET_PYTHON_EXECUTABLE" >/dev/null 2>&1 ; then
     echo "error: unable to find python3 as \"$DET_PYTHON_EXECUTABLE\"" >&2
     echo "please install python3 or set the environment variable DET_PYTHON_EXECUTABLE=/path/to/python3" >&2
     exit 1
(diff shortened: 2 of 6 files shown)


stoksc (approved):

stoksc:
@rb-determined-ai this probably deserves a release note?

stoksc:
idk, I'm trying to be more diligent with those

rb-determined-ai:
sure

docs/release-notes/4257-which.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Support running in containers which did not have ``/bin/which``, like python-slim. The error was
```
stoksc, line 5: maybe just "Don't require task image's `which` to be located in `/bin/which`"

rb-determined-ai: @gh-determined-ai do you want this detail or not?

(@stoksc I frequently get asked to add detail to the release notes I write)

docs/release-notes/4257-which.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Support running in containers which do not have ``/bin/which``, like python-slim. The error was
```
gh-determined-ai, line 5: suggest:
"Supports running in containers which do not have a ``/bin/which`` path, such as python-slim. The error was caused by accidentally ..."

gh-determined-ai (approved):

dannysauer:
Should containers depend upon `which` existing at all?  The `which` command isn't required by POSIX and isn't a shell built-in; the equivalent is `command -v`, which is implemented as a bash builtin.  It's conceivable that there could be systems with no `which`.

Most Debian-derived systems actually implement `which` as a shell script, so pulling the relevant section out and implementing it as a function would remove any system dependency. Since the output is being discarded anyway, it's just the eight lines below (fewer if you get semicolon-happy :D). Using `command -v` is probably more portable than `which`, though.  It has the same behavior, and if bash exists, built-in `command` definitely will.

```shell
function localwhich {
  local IFS
  IFS=:
  for D in $PATH
  do
    [[ -f "$D/$1" && -x "$D/$1" ]] && return 0
  done
  return 1
}
```

docs/release-notes/4257-which.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Supports running in containers which do not have a /bin/which path, such as python-slim. The
+   error was caused by accidentally accidentally hardcoding ``/bin/which`` instead of letting the
```
gh-determined-ai, line 6: accidentally 2 instances of "accidentally"

rb-determined-ai: doh

rb-determined-ai:
> Should containers depend upon which existing at all?

Eh, I think we can discuss implementing which in shell when the time comes that somebody requests running a container without which.  We have complex enough shell scripts to support all the workflows we know about, and that's one we've never encountered before.

(but it is interesting, I didn't know which was not part of posix)

PR #4258, opened 2022-06-03
fix: det task logs unable to use trial task IDs and checkpoint GC task IDs [DET-7424]

NicholasBlaskey:
## Description

```det task logs``` now works for trial task IDs and checkpoint GC task IDs.

## Test Plan

CLI unit test and running ```det task logs <CHECKPOINT_GC_TASK_ID>```

## Checklist

### DIFF
--- a/harness/determined/cli/command.py
+++ b/harness/determined/cli/command.py
@@ -39,7 +39,7 @@ _CONFIG_PATHS_COERCE_TO_LIST = {
 }
 
 UUID_REGEX = re.compile(
-    "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", re.IGNORECASE
+    "^(?:[0-9]+\.)?[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", re.IGNORECASE
 )
 
 CommandTableHeader = OrderedDict(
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: harness/determined/cli/command.py [hunk 1]
```diff
@@ -143,10 +143,18 @@ def expand_uuid_prefixes(
...
     # Avoid making a network request if everything is already a full UUID.
     if not all(UUID_REGEX.match(p) for p in prefixes):
-        api_path = RemoteTaskNewAPIs[args._command]
-        api_full_path = "api/v1/{}".format(api_path)
-        res = api.get(args.master, api_full_path).json()[api_path]
-        all_ids: List[str] = [x["id"] for x in res]
+        # If we are unable to determine what specific task type something is
+        # such as when we are calling det task logs we need to check all
+        # of the possible API options.
+        api_paths = list(RemoteTaskNewAPIs.values())
```
stoksc, line 149: i don't think this is going to quite work - the CLI subcommands map one to one with the APIs listed in `RemoteTaskNewAPIs`, so if there isn't any entry, calling them all won't get any response that helps us expand the UUID. as a fallback (or really, in every case), we could just call `/tasks` since that should contain trial and gc entries (even better, if we had a parallel for it in the new api, we could call that). but for now, i'm fine with saying 'uuid expansion not supported for task type: ...' and making a ticket/discussing the rest more. probably could use a slight api redesign around these task apis.

i think the main problem here, good catch, was the UUID (task id) regex being wrong for checkpoint gc task ids, so it always thought it was trying to parse a short for uuid for those, so the full uuid stopped working too.

NicholasBlaskey: I'll make this return an error message for trying to expand with ```det task logs```

Created DET-7452 as a follow up.

PR #4260, opened 2022-06-03
fix: backslashes in show_ssh_command for windows

rb-determined-ai:
Avoid showing backslashes in show_ssh_command, because when passed to
VSCode, they break the resulting .ssh/config.

## Test Plan

On windows and *nix:

- [x] make sure det shell show_ssh_command can be copy/pasted
- [x] make sure vscode works with the show_ssh_command
- [x] make sure det shell open still works

## Commentary:

After discussion with @ioga, the best "solution" to the vscode/virtualenv problem is to say that "show_ssh_command must be run from the same virtualenv as vscode, which is usually no virtualenv at all".

### DIFF
--- a/harness/determined/cli/shell.py
+++ b/harness/determined/cli/shell.py
@@ -49,14 +49,14 @@ def start_shell(args: Namespace) -> None:
         return
 
     ready = False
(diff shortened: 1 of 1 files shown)


mpkouznetsov (approved):
A nitpicky comment about a comment, otherwise looks good.

PR #4263, opened 2022-06-06
docs: delete mnist_tf_layers example

rb-determined-ai:
Around TensorFlow 2.0, the recommended way to use TensorFlow core was
to use wrap it in the tf.layers API and use the layers in an Estimator.

Or at least, that's what I remember at the time.  Estimators was also
deprecated in 2.0 so I'm not sure why that would have been the case.

In any case, since Estimators has now been deprecated in TF for years,
and since we want to drop support for EstimatorTrial at some point, and
since the current TF documentation points TF Core users at keras, we
should stop pointing TF Core users at EstimatorTrial.

Also, the mnist_tf_layers was breaking our unit tests for some reason.


liamcli (approved):
lgtm.  i haven't seen estimator code for a long time.

PR #4266, opened 2022-06-07
chore: Restrict display names [DET-7356]

mapmeld:
## Description

- Prevent non-admin users from appearing as "admin" with display name
- Prevent users from appearing as "determined"
- Prevent users from choosing a display name to copy another user's username / display name
- Prevent users from using non-Latin display names (i.e. ASCII, numbers, Latin alphabet accent marks) (I'm open to making this more flexible, but this avoids someone using invisible spaces, look-alike characters, or the RLM char to disrupt other content and rules)
- In proto/backend, changed PatchUser.DisplayName (null or string) from null.String type to google.protobuf.StringValue (more similar to Model and Workspaces code)

## Test Plan

- Change display name to empty string
- Change display name to dêtérmïned , attempt to change to determined
- Attempt to change display name to Admin / admin
- Confirm previous e2e_tests for display name work

Edit: Trent recommended https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt - most of these escaping / scripting strings are already handled by the db driver and react.  I was most worried about the category with strings like "Ṱ̺̺̕o͞ ̷in̝̗͕vokè" , but these get pruned back to a few Latin alphabet markers.

## Checklist


master/static/srv/set_user_display_name.sql
```diff
@@ -0,0 +1,12 @@
+WITH confusable_users AS (
+  SELECT COUNT(*) AS count
+  FROM users
+  WHERE $2 != ''
```
johnkim-det, line 4: question: could this `$2 != ''` conditional cancel out the `LOWER(username)` conditional?

master/internal/api_user.go
```diff
@@ -166,32 +167,37 @@ func (a *apiServer) PatchUser(
...
+		} else {
+			// Remove non-ASCII chars to avoid hidden whitespace, confusable letters, etc.
+			re := regexp.MustCompile("[^\\p{Latin}\\p{N}\\s]")
+			displayName := re.ReplaceAllLiteralString(req.User.DisplayName.Value, "")
+			// Restrict 'admin' and 'determined' in display names.
+			if !(curUser.Admin && curUser.ID == uid) && strings.Contains(strings.ToLower(displayName),
+				"admin") {
+				return nil, status.Error(codes.InvalidArgument, "Non-admin user cannot be renamed 'admin'")
+			}
+			if curUser.Username != "determined" && strings.Contains(strings.ToLower(displayName),
```
johnkim-det, line 188: question: does this need to use `strings.Contains`? or would having "admin" or "determined" as a substring be acceptable?

master/internal/api_user.go
```diff
@@ -166,32 +167,37 @@ func (a *apiServer) PatchUser(
...
-			return nil, errUserNotFound
-		case err != nil:
-			return nil, err
+	if req.User.DisplayName != nil {
+		u := &userv1.User{}
+		if req.User.DisplayName.Value == "" {
+			err = a.m.db.QueryProto("set_user_display_name", u, req.UserId, "")
+		} else {
+			// Remove non-ASCII chars to avoid hidden whitespace, confusable letters, etc.
+			re := regexp.MustCompile("[^\\p{Latin}\\p{N}\\s]")
```
johnkim-det, line 181: question: is it preferred to remove problem characters and make the update over throwing an error?

johnkim-det (approved):
LGTM, a few questions about intended behavior.

master/static/srv/set_user_display_name.sql
mapmeld: This is to find username conflicts -- if the set username is blank it should be 0, otherwise we look at the other WHERE clauses

master/internal/api_user.go
mapmeld: I think it would be difficult to know how to change my display name if I've included one less-common char. Also we don't display the error on the UI so they wouldn't know what had happened.

PR #4267, opened 2022-06-07
chore: secure echo with default authentication [DET-7405] [DET-7378]

eecsliu:
## Description
Adds authentication by default to the echo server, plus a whitelist for endpoints that should be allowed.

## Test Plan
Has been tested locally. Make sure that you are able to login, see the UI, and access common CLI commands. Run const, adaptive, and distributed experiments, as well as shells, notebooks, commands, and tensorboards. None should error out or encounter permissions issues.

## Checklist


master/internal/user/service.go
```diff
@@ -23,6 +24,42 @@ import (
 var once sync.Once
 var userService *Service
 
+// unauthenticatedPathsMap contains the paths that are exempted from authentication.
+var unauthenticatedPathsMap = map[string]bool{
```
dzhu, line 28: Could this possibly be unified with the slice below?

master/internal/user/service.go
```diff
@@ -23,6 +24,42 @@ import (
...
+	"/agents":             true,
+	"/det/*":              true,
+	"/login":              true,
+	"/api/v1/master":      true,
+	"/api/v1/auth/login":  true,
+	"/api/v1/auth/logout": true,
+	"/proxy/:service/*":   true,
+}
+
+// unauthenticatedPointsList contains URIs that are exempted from authentication.
```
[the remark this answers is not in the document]
eecsliu: ah yeah, so the map should be paths and URIs, and the list should just be URIs. It looks like the path and the URI are different for each request.

master/internal/user/service.go
```diff
@@ -23,6 +24,42 @@ import (
...
+var unauthenticatedPointsList = []string{
+	"/api/v1/notebooks/*",
+	"/api/v1/tasks/*",
+	"/api/v1/allocations/*",
+	"/api/v1/tensorboards/",
```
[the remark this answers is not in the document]
eecsliu: interestingly not - the tensorboard URI is indeed this

master/internal/core.go
```diff
@@ -866,6 +864,8 @@ func (m *Master) Run(ctx context.Context) error {
 		m.echo.Use(otelecho.Middleware("determined-master"))
 	}
 
+	m.echo.Use(userService.ProcessAuthentication)
```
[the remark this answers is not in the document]
eecsliu: I'll give this a look, but I imagine the easiest approach is just to add a blacklist alongside the admin and unauth lists...

master/internal/user/service.go
```diff
@@ -23,6 +24,42 @@ import (
...
+	"/api/v1/shells/*",
+	"/api/v1/experiments/*",
+	"/api/v1/trials/*",
+	"/api/v1/checkpoints",
```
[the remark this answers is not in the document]
eecsliu: this does not appear to be the case on checkpoints

master/internal/user/service.go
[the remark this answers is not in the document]
eecsliu: Actually there were some tensorboard paths that needed the star, so I changed to `/api/v1/tensorboards/.*` for regex

master/internal/user/service.go
```diff
@@ -105,17 +142,59 @@ func (s *Service) UserAndSessionFromRequest(
 // ProcessAuthentication is a middleware processing function that attempts
 // to authenticate incoming HTTP requests.
 func (s *Service) ProcessAuthentication(next echo.HandlerFunc) echo.HandlerFunc {
-	return s.processAuthentication(next, false)
+	return s.processAuthentication(next)
 }
 
 // ProcessAdminAuthentication is a middleware processing function that authenticates requests much
 // like ProcessAuthentication but requires the user to be an admin.
 func (s *Service) ProcessAdminAuthentication(next echo.HandlerFunc) echo.HandlerFunc {
```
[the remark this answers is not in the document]
eecsliu: Changed, will submit another pr to EE to resolve any issues stemming from that.

master/internal/user/service.go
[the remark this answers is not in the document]
eecsliu: yeah so there definitely are `../checkpoints/..` endpoints, but `../checkpoints` is specifically for experiments to record new checkpoints i think - in my experimentation this does not work unless I leave it unauthenticated

master/internal/user/service.go
```diff
@@ -29,6 +30,53 @@ var forbiddenError = echo.NewHTTPError(
...
+	"/api/v1/auth/logout",
+	"/proxy/:service/.*",
+	"/api/v1/notebooks/.*",
+	"/api/v1/tasks/.*",
+	"/api/v1/allocations/.*",
+	"/api/v1/tensorboards/.*",
+	"/api/v1/shells/.*",
+	"/api/v1/experiments/.*",
+	"/api/v1/trials/.*",
+	"/api/v1/checkpoints",
```
stoksc, line 63: drive by comment, why are any of these unauthed?

eecsliu: When I was doing testing on this, things wouldn't run unless I had these unauthenticated - i.e. experiments wouldn't run unless `/api/v1/experiments/.*` was unauthenticated, or tensorboards wouldn't run without the tensorboard api path was unauthenticated. I think when the agent is running tasks, it hits some endpoints to send or receive info.

ioga: shouldn't all new style APIs `/api/*` be authed through their own logic?

ioga: however if this auth just checking if user is signed in then it's not clear why it'd be breaking.

eecsliu: yes to the latter

ioga: it could mean the harness is making unauthenticated calls, for example? For the purpose of security I'd say we should get to the bottom of this.

eecsliu: ok after more testing, I can restrict the `/api/v1/experiments` path to `/api/v1/experiments/.*/model_def` and toss any paths unrelated to experiments (shells, notebooks, tasks). But I can't do much about the `trials`, `allocations`, or `checkpoint`.

stoksc: this sounds like all the calls the harness makes to the master?

stoksc: why can't you remove the model def call? that call (every call basically) should be authed.

stoksc: can you debug why that is happening? that endpoint should be authorized.

eecsliu: Yeah looking into it - the harness needs the model_def, but the authorization doesn't see any token so responds with a 401 error.

stoksc: yeah, there should be a session token in the env it uses

master/internal/user/service.go
```diff
@@ -29,6 +30,44 @@ var forbiddenError = echo.NewHTTPError(
...
+	// authStandard indicates a request needs authentication.
+	authStandard = 1
+	// authAdmin indicates a request needs admin authentication.
+	authAdmin = 2
+)
+
+// unauthenticatedPointsList contains URIs that are exempted from authentication.
+var unauthenticatedPointsList = []string{
+	"/",
+	"/det/",
```
dzhu, line 45: This is redundant with `/det/.*`.

master/internal/user/service.go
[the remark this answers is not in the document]
dzhu: Okay, I think now they really should both say the same thing?

dzhu (approved):
I haven't checked the precise list of which endpoints go where, but apart from that the code looks reasonable.

master/internal/user/service.go
```diff
@@ -108,20 +156,34 @@ func (s *Service) UserAndSessionFromRequest(
 	return s.db.UserByToken(token, s.extConfig)
 }
 
+// getAuthLevel returns what level of authentication a request needs.
+func (s *Service) getAuthLevel(c echo.Context) int {
+	switch {
+	case adminAuthPointsPattern.MatchString(c.Request().RequestURI):
+		return authAdmin
+	case unauthenticatedPointsPattern.MatchString(c.Path()):
+		return authNone
```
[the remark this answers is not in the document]
eecsliu: I think the `/agents\\?id=.*` is a path instead of a URI - I'll double check and either update the comment or remove path checking

master/internal/user/service.go
```diff
@@ -29,6 +30,53 @@ var forbiddenError = echo.NewHTTPError(
...
+	"/info",
+	"/task-logs",
+	"/ws/data-layer/\\.*",
+	"/agents",
+	"/det/\\.*",
+	"/login",
+	"/api/v1/master",
+	"/api/v1/auth/login",
+	"/api/v1/auth/logout",
+	"/proxy/:service/\\.*",
```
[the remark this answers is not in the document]
eecsliu: You might have reviewed a slightly older version of the code? Some of these list changes have already been made and/or have been removed.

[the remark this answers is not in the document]
dzhu: Yeah, I had some old comments sitting around. I resolved everything that was old.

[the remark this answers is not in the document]
eecsliu: got it, thanks!

PR #4269, opened 2022-06-07
feat: stress test for agent enable/disable and reconnect spam [DET-6733]

NicholasBlaskey:
## Description

Tests repeatably enabling and disabling agents and adding and deleting existing agents.

## Test Plan

Run locally (ensure you have docker images built locally) 

And verify works in this PR. 

## Checklist

### DIFF
--- a/e2e_tests/tests/deploy/test_local.py
+++ b/e2e_tests/tests/deploy/test_local.py
@@ -4,6 +4,7 @@ import random
 import subprocess
 import time
 from pathlib import Path
+from tempfile import NamedTemporaryFile
 from typing import List, Optional
 
 import docker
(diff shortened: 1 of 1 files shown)


e2e_tests/tests/deploy/test_local.py
```diff
@@ -192,3 +197,79 @@ def test_agent_up_down() -> None:
...
+    for i in range(num_agents):
+        agent_up(["--agent-name", f"agent-{i}"])
+    time.sleep(3)
+
+    for _ in range(steps):
+        for agent_id, agent_is_up in enumerate(agents_are_up):
+            if random.choice([True, False]):  # Flip agents status randomly.
+                continue
+
+            if agent_is_up:
```
ioga, line 224: I'd switch them up / down randomly regardless of their current state, instead of "toggling" from up to down and back. that'd be more chaotic.

e2e_tests/tests/deploy/test_local.py
```diff
@@ -192,3 +197,79 @@ def test_agent_up_down() -> None:
...
+                    "--json",
+                ]
+            ).decode()
+        )
+        assert sum(agents_are_up) <= len(agent_list)
+        for agent in agent_list:
+            agent_id = int(agent["id"].replace("agent-", ""))
+            assert agents_are_up[agent_id] == agent["enabled"]
+
+    for agent_id in range(num_agents):
```
ioga, line 273: it will be useful to also run a command and/or noop experiment to make sure the master can still schedule things? there were failures when `det agent list` still works okay but nothing would schedule because the resource pool is in a bad state.

NicholasBlaskey: Good call out.

The no-op experiment was originally failing with unable to connect to the logging driver at 24224.

Seems my multiple deploy agents locally caused them all to try and create the same fluent container name causing it constantly restart.

e2e_tests/tests/deploy/test_local.py
```diff
@@ -230,3 +258,77 @@ def test_agent_up_down() -> None:
...
+    agents_are_up = [True] * num_agents
+    for i in range(num_agents):
+        agent_up(["--agent-name", f"agent-{i}"], fluent_offset=i)
+    time.sleep(3)
+
+    for _ in range(steps):
+        for agent_id, agent_is_up in enumerate(agents_are_up):
+            if True and False:
+                continue
+
```
ioga, line 285: 🤔

ioga (approved):

PR #4276, opened 2022-06-08
feat: enable open telemetry in agent [DET 7085]

nrajanee:
## Description

enable open telemetry in agent. 

## Test Plan

No testing required during a release. 

I will complete the following steps after enabling API on the agent side. I've ensured that master traces are working fine with these changes by following the same steps and looking for traces with http.server_name: STRING(determined-master). 

**When otel_enabled is set to true in devcluster.yaml file (telemetry: otel_enabled: true):**

- [x]  Run make devcluster
- [x]  Set endpoint telemetry (default is "localhost:4317") :
- [x] otel_endpoint: {insert_endpoint}` in devcluster.yaml (endpoint eg: "localhost:4317". Note port number here is 4317.)
- [x]  Run docker run --rm -v “$HOME/Desktop/opentel/config.yaml" -p 1888:1888 -p 8888:8888 -p 8889:8889 -p 13133:13133 -p {port_number}:4317 -p 4318:4318 otel/opentelemetry-collector to start the opentel collector.
- [x]  Hit the agent api endpoint  and check that you are seeing traces (with http.server_name: STRING(determined-agent)) when you hit diff endpoints.


tools/devcluster.yaml
```diff
@@ -47,6 +47,9 @@ stages:
         log:
           level: debug
         enable_cors: true
+        telemetry:
```
nrajanee, line 50: remove this before you merge

philipnrmn: ?

agent/internal/agent.go
```diff
@@ -513,14 +515,19 @@ func (a *agent) postTaskLog(log model.TaskLog) error {
 	return nil
 }
 
-func runAPIServer(options Options, system *actor.System) error {
```
philipnrmn, line 516: This seems like a fairly major change, I wonder if we can avoid it? Maybe if we added `OtelEnabled` to `Options`, and switched it on within `Run`?

philipnrmn (changes requested):

tools/devcluster.yaml
nrajanee: that's just a note for myself @philipnrmn. Because this is just to change the config to enable otel. This would be on the users side.

agent/internal/agent.go
nrajanee: Yes that's true. I'll look into passing it through options.

philipnrmn (approved):
These changes lgtm; let's fix up lint though :)

PR #4277, opened 2022-06-08
docs: documentation site restructuring

gh-determined-ai:
## Checklist


gh-determined-ai:
A preview can be found here: [https://cool-licorice-354205.netlify.app/](https://cool-licorice-354205.netlify.app/)

hkang1 (commented):
Thanks for doing this work!

- The quick start looks good with updated screens.
- The dedicated WebUI Interface section is great, looks like a great area that we can expand on. Small nit of changing it to `Web Interface` or `Web User Interface` as the acronym `UI` already includes `Interface`.
- Until the Web section is expanded, would it help to have something on there to indicate that TBD on more content, to encourage users to check back later for more content?

Not for this PR:
- Do we have analytics as part of Docs (I can't remember)? If not, would it be beneficial to add it so we can track what user's are looking at?

gh-determined-ai:
> 
@hkang1 ...
- The web interface section definitely could use some content. I'll add it to the list of projects.
- Will rename WebUI Interface.
- Will add a TBD to the section for now. (It seems to have already served its immediate purpose.)
- I'm not aware of doc site analytics. Maybe @neilconway knows. I haven't found analytics to be very helpful in the past, although, I did feel mighty sorry for someone once when I looked at the page they were currently viewing; I think it takes some sophistication to get meaningful information, a project in itself, which a developer advocate or marketing might be better positioned to address. Kind of along the same lines, a feedback option on each page might be useful, although, I think we already know what needs work.

gh-determined-ai:
**Clarification for reviewers**: The documentation restructuring phase is intended to provide a solid foundation for making more substantive changes in the next, doc rewrite phase.
While details of correctness and completeness are important, they are only of interest at this time if you determine that they have a bearing on documentation structure. It's good to keep a list of those details you uncover, and we'll get to them next, in addition to a multitude of other sins.
The intent and focus of this PR is _usability_.

garrett361 (commented):
Nice work, Gene! Some comments, mostly nitpicks:

1. I would prefer having fewer cards on the landing page, say 4 like the NumPy/SciPy docs. In particular, I would consider dropping Tutorials, Example Solutions, and Python API (maybe even Cluster Setup Guide).
2. It is a little confusing when the cards take you to pages whose titles don't match the card title and/or content implied by the card title. Examples:
     * I would expect the Tutorials card to take me to some general Tutorials landing page, rather than the more specific PyTorch Tutorial.
     * The Reference card currently takes me to the Training API page, whereas I would have expected to find a page which also has CLI, Python API, REST API, etc docs links available.
3. I would have expected to find the Tutorials sidebar section up near the Introduction to Determined section, or even find Tutorials as a subsection thereof.
4. The name Example Solutions is confusing to me and I wouldn't know what to find there based on that title. I would find a simple "Examples" clearer.
5. Open Source Licenses is awkwardly placed in the Example Solutions section. Maybe next to the copyright at the bottom of the page would be better?

It's great that the docs are improving; so useful to users and ourselves, both.

neilconway:
@hkang1 @gh-determined-ai  There is Google Analytics on the current doc site; we should preserve it as we make changes.

* The high-level organization of the docs could stand to be further improved, in my opinion. Is changing that intended to be within scope for this current project? e.g., some of the top-level sections are user activities ("Cluster Setup", "Training"), while others describe kinds of documentation ("Tutorials", "Examples", etc.), while others describe parts of the software architecture ("Determined Interfaces", "Model Hub").
* It seems a bit asymmetric to me that top-level sections (e.g., "Training") don't have any content; some level-2 pages (e.g., Security) just content a table of contents for further nested content, whereas some level-2 pages are leaf nodes and contain content directly. Some of the level-2 pages that have sub-pages link to those subpages via a table (e.g., Security), whereas others do not (e.g., Model Management).
* I think we need _some_ kind of succinct explanation of what Determined is, what problems it solves, why you might use it, etc. on the docs index page
* Seems like there is both a "Python API" and "Python API Reference" page, with similar content?
* The first page in the "Training" section is "Introduction to Distributed Training", which does not seem right.
* "Interactive Job Configuration" probably doesn't belong in the "Determined Interfaces" section as a top-level page.

shiyuann:
I think the contents are too long. The training section is at the end of the second page. As a model developer, I would need to skip many sections that are irrelevant to me, which is a bad experience. Can you fold each section into one first level navigation so that users can unfold by clicking the down arrow?

shiyuann:
@gh-determined-ai There are two types of users: individual developer that functions as a MLE and a platform engineer, teams that have different people for MLE and platform engineer roles. For the first type of MLEs, they do need to access **Cluster Setup Guide**. But their usage pattern is to use **Cluster Setup Guide** in a lower frequency when installing and upgrading the cluster while using **Training** in a higher frequency.

Ray is very similar to our product in that it also have two very different domains of readers, MLE and platform engineer. I like https://docs.ray.io/en/latest/ in that its TOC is short and it sorts the pages in a way that is friendlier to ML developers.

docs/index.rst
```diff
@@ -0,0 +1,152 @@
...
+   Concepts <intro/concepts/overview>
+   System Architecture <intro/architecture/overview>
+
+.. toctree::
+   :caption: Determined Interfaces
+   :hidden:
+
+   WebUI Interface <interfaces/webui-if>
+   Commands and Shells <interfaces/commands-and-shells>
+   Python API <interfaces/python-api>
```
rb-determined-ai, line 20: Can we refer to this entity as the "Python SDK"?  Which is distinguished from all other python-related APIs because it's the only entity in python for interacting with the cluster from outside of jobs.  "Python SDK" is what we have normally called it in the past.

rb-determined-ai (commented):
I think this is overall a move in the right direction, but I agree with some of the feedback that there are improvements to be made here.  I think the training and tutorials sections are awfully late in the ordering, for instance.  I would expect tutorials to be very near the top of the list.

gh-determined-ai:
@shiyuann  I'm tending to agree with you that a compressed TOC would be a better fit for this product. At least the option should be available.
@rb-determined-ai  I'll restore the redirects before publishing.

shiyuann:
@gh-determined-ai I think there are a couple of more comments that I have.

1. I think the administration guide and the interfaces should be compressed as the training section. 
2. Model hub should be moved to a higher position, rather than below the interfaces. I'd suggest fold the model hub pages into one level and move it to "getting started".
3. The Training page is indeed the guide for distributed training. It should actually be a landing page for all the training topics.

PR #4281, opened 2022-06-08
feat: install a SIGUSR1 signal handler to print stacktraces

rb-determined-ai:
## Description

Useful to find out why a hung trial is hung, when you forgot to run it
with debug=true.

## Test Plan

- Run a long-running training
- docker exec into the container
- ps aux to find a determined.exec.harness process
- `kill -SIGUSR1 THEPID`
- check for a stacktrace in trial logs

### DIFF
--- a/harness/determined/core/_context.py
+++ b/harness/determined/core/_context.py
@@ -1,4 +1,7 @@
 import logging
+import signal
+import sys
+import traceback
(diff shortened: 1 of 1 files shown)


asnelling (approved):
tested and verified ✅

PR #4282, opened 2022-06-08
feat: normalize URLs for CLI auth.json [DET-7357]

NicholasBlaskey:
## Description

Normalize URLs in auth.json to prevent weird behavior between entries like ```http://example.com``` and ```example.com:80```.

## Test Plan

CLI unit tests.

## Commentary (optional)

Went back and fourth on how much should the CLI try to fix a config with entries of ```localhost``` and ```localhost:8080```. Was not able to come up with a way that could avoid potentially cause surprising behavior, so decided in that case to raise the corrupted config exception which deletes the config. 

## Checklist

### DIFF
--- a/docs/release-notes/normalize-auth-json.txt
+++ b/docs/release-notes/normalize-auth-json.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Improvements**
+
+-  CLI: URLs stored in ``auth.json`` are now normalized to prevent issues with the same URL being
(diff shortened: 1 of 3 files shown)


dzhu (commented):
Since we've got the version system in place, how about we bump to v2, which is defined to always have master URLs normalized? Then we don't have to run the logic every time and I think the code might be simpler.

gh-determined-ai (approved):

PR #4283, opened 2022-06-08
feat: show only selected trials in learning curve

trentwatt:
## Description

allow users to narrow down trials by clicking their checkboxes in trials table below Learning Curve graph

## Test Plan

find you a nice multi-trial experiment,e.g. 562
- click some checkboxes, only selected trials should show in the graph
- clear selection, all trials should show
- click some more checkboxes
- change metric, selection should be preserved
- change top trials to show only one
- that top trial should show regarless of whether it was selected
- go back to top trials = 100
- the previously selected trials should show
- go back to top trials =1 (without top trial being selected)
- click the checkbox for that single trial
- go back to top trials = 100
- just that single trial should show up

preferably also test with multi-trial experiment that is in progress??

## Checklist

### DIFF
--- a/webui/react/src/components/LearningCurveChart.tsx
+++ b/webui/react/src/components/LearningCurveChart.tsx
@@ -13,6 +13,7 @@ interface Props {
   onTrialClick?: (event: MouseEvent, trialId: number) => void;
   onTrialFocus?: (trialId: number | null) => void;
   selectedMetric: MetricName;
+  selectedTrialIds: number[];
   trialIds: number[];
   xValues: number[];
 }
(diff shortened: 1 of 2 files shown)


gt2345 (approved):
Great work showing only selected trial, but one thing could possibly improve is, when no trial is selected, the color of each learning curve is the same as the color tag in the trials table, but when trials are selected, the filtered learning curve color is no longer corresponds with the color tag in the trials table, it would be great if the color of learning curve persist when filtered. 
Screen Shot 2022-06-08 at 9 26 21 PM
Screen Shot 2022-06-08 at 9 26 43 PM

trentwatt:
> Great work showing only selected trial, but one thing could possibly improve is, when no trial is selected, the color of each learning curve is the same as the color tag in the trials table, but when trials are selected, the filtered learning curve color is no longer corresponds with the color tag in the trials table, it would be great if the color of learning curve persist when filtered.

very good catch, thank you!

trentwatt:
fixed!

PR #4284, opened 2022-06-08
chore: stamp out unexpected messages [DET-7492]

eecsliu:
## Description
There has been a mix of asks and tells in the actor, so the safe play is to check if the sender is expecting a response. This currently is done to the priority and weight messages, which are the job messages that have a chance of originating outside of the jobs actor.

## Checklist

### DIFF
--- a/master/internal/command/command.go
+++ b/master/internal/command/command.go
@@ -345,10 +345,16 @@ func (c *command) Receive(ctx *actor.Context) error {
 		ctx.Self().Stop()
 
 	case job.SetGroupWeight:
-		ctx.Respond(c.setWeight(ctx, msg.Weight))
+		err := c.setWeight(ctx, msg.Weight)
+		if ctx.ExpectingResponse() {
+			ctx.Respond(err)
+		}
 
 	case job.SetGroupPriority:
-		ctx.Respond(c.setPriority(ctx, msg.Priority, true))
+		err := c.setPriority(ctx, msg.Priority, true)
+		if ctx.ExpectingResponse() {
+			ctx.Respond(err)
+		}
 
 	case job.RegisterJobPosition:
 		err := c.db.UpdateJobPosition(msg.JobID, msg.JobPosition)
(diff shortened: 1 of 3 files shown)


master/internal/command/command.go
```diff
@@ -345,10 +345,16 @@ func (c *command) Receive(ctx *actor.Context) error {
 		ctx.Self().Stop()
 
 	case job.SetGroupWeight:
-		ctx.Respond(c.setWeight(ctx, msg.Weight))
+		err := c.setWeight(ctx, msg.Weight)
+		if ctx.ExpectingResponse() {
+			ctx.Respond(err)
```
stoksc, line 350: does the error get logged anywhere?

eecsliu:
There appears to only be one more occurrence of these messages outside of the jobs actor, which is in patch experiment. However it seems to me to be better taste to check response instead of ignoring an error after receiving.

master/internal/command/command.go
eecsliu: I'll add it. The intended behavior was to bubble it back up to the user as part of job modification.

stoksc: `ctx.Log().WithError()` logs nothing. need to actually call info/warn/etc.. also, we can just do
```
if err != nil {
 	ctx.Log().WithError(err).Info("setting command job weight")
         if ctx.ExpectingResponse() {
 	        ctx.Respond(err)
         }
 }
```
if you want because the `ctx.Respond(nil)` is implied if the caller is expecting a response.

stoksc:
@eecsliu are you talking about occurrences where this is called, and that it seems we really should be asking anyway, to check the err response. if so, 100% agree. we should do that.

master/internal/command/command.go
```diff
@@ -356,7 +356,7 @@ func (c *command) Receive(ctx *actor.Context) error {
 	case job.SetGroupPriority:
 		err := c.setPriority(ctx, msg.Priority, true)
 		if err != nil {
-			ctx.Log().WithError(err)
+			ctx.Log().WithError(err).Info("setting command job priority")
```
stoksc, line 359: I still would just put the `ctx.Respond()` in the err block but definitely a nit pick lol. I have something against returning an `err` we know is nil from a readability perspective.

stoksc (approved):

PR #4286, opened 2022-06-08
fix: pass signals through wrapper processes

rb-determined-ai:
I thought we were cleaning things up nicely, but when logging started
keeping containers open until all copies of the stdout file descriptor
closed, it became obvious that the pid_client was being killed by
horovodrun without shutting down its child wrap_rank script, causing the
logging to stay active and the container to hang.

This first fix is to pass signals through wrapper processes, so that
normal graceful shutdowns complete as-expected.

## Test Plan

With a 0.18.1 master:
- start a trial with a context.distributed.allgather in one rank, with an exit(1) in another rank
- observe horovod try to shut down training
- observe training does not stop
- observe pid_server gets impatient and SIGKILLs horovodrun
- observe the container is hung
- docker exec into the container, and see that the pid_client is gone, but the wrap_rank process is not.

After this fix, the same procedure should result in everything shutting down properly.

## Commentary

I sheepishly admit that I always wondered how all the processes were getting cleaned up, and I'm very embarrassed that even so I've always been very confident we cleaned things up properly.

### DIFF
--- a/docs/release-notes/4286-passthru.txt
+++ b/docs/release-notes/4286-passthru.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Prevent certain hangs when using one of Determined's built-in launchers, which begin in release
+   0.18.0. These hangs were caused by wrapper processes which were seeing SIGTERM but not passing it
+   to their child process.
--- a/harness/determined/util.py
+++ b/harness/determined/util.py
@@ -1,4 +1,5 @@
 import collections
+import contextlib
 import datetime
 import enum
 import inspect
@@ -11,10 +12,12 @@ import pathlib
 import random
 import re
 import shutil
+import signal
 import socket
+import subprocess
 import time
 import uuid
-from typing import Any, Callable, Dict, List, Optional, Set, SupportsFloat, Tuple, cast
+from typing import Any, Callable, Dict, Iterator, List, Optional, Set, SupportsFloat, Tuple, cast
 
 import determined as det
 from determined import constants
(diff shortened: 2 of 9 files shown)


harness/determined/util.py
```diff
@@ -325,3 +328,28 @@ def match_legacy_trial_class(arg: str) -> bool:
 
 def legacy_trial_entrypoint_to_script(trial_entrypoint: str) -> List[str]:
     return ["python3", "-m", "determined.exec.harness", trial_entrypoint]
+
+
+@contextlib.contextmanager
+def forward_signals(p: subprocess.Popen, *signums: signal.Signals) -> Iterator[None]:
+    """Forward a list of signals to a subprocess, restoring the original handlers afterwards."""
+    if not signums:
```
mpkouznetsov, line 336: I guess the conditionality here could be useful in the future? For now I notice that we always call it with one argument (i.e. without specific signals).

mpkouznetsov (approved):

harness/determined/util.py
rb-determined-ai: :shrug: I could go either way.

docs/release-notes/4286-passthru.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Prevent certain hangs when using one of Determined's built-in launchers, which begin in release
+   0.18.0.  These hangs were caused by wrapper processes which were seeing SIGTERM but not passing
+   it to their child process, which is now fixed.
```
gh-determined-ai, line 7: nit: Probably don't need to say ", which is now fixed."

gh-determined-ai (approved):

PR #4287, opened 2022-06-08
fix: support "short" validations in TFKerasTrial [MLG-29]

rb-determined-ai:
"Short" validations are not empty, but short enough that one or more
workers will have an empyt shard.  hvd.allreduce() must be avoided,
because it doesn't support non-participating workers.

Note that with graph execution, model.evaluate() crashes on empty
validation data, so short validations are effectively an
eager-execution-only feature.

## Test Plan

- [x] short validations work in eager mode (not really possible in graph mode)
- [ ] averaging works with scalars and arrays (graph mode)
- [ ] averaging works with scalars and arrays (eager mode)
- [ ] averaging preserves the types emitted in logs (graph mode)
- [ ] averaging preserves the types emitted in logs (eager mode)
- [ ] check if there are non-scalar, non-array types that can be emitted in logs

I guess we need to check all of these against multiple versions of keras to be sure.

### DIFF
--- a/harness/determined/keras/_tf_keras_trial.py
+++ b/harness/determined/keras/_tf_keras_trial.py
@@ -847,33 +847,17 @@ class TFKerasTrialController(det.TrialController):
     def _allreduce_logs(self, logs: Dict) -> Dict:
         if not (self.context.distributed.size > 1):
             return logs
-        # Reduce logs in key-sorted to be deterministic across workers.
-        keys = sorted(logs)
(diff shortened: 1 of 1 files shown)


varlogtim:
I have somewhat started testing on this. I hope to get this done by end of next week.

rb-determined-ai:
Closing due to:
* not enough time to actually test
* the bug is real but very unlikely
* the bug is in code that is destined for removal.

We can reopen if a human ever hits it.

PR #4288, opened 2022-06-08
chore: restore determined_version in Trial checkpoint metadata

rb-determined-ai:
This was accidentally dropped from the automatically-stored metadata as
part of the generic checkpoint feature branch (4fb1886b).

## Commentary

This was probably my mistake.

### DIFF
--- a/docs/release-notes/4288-version.txt
+++ b/docs/release-notes/4288-version.txt
@@ -1,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Automatically add a ``determined_version`` key to the metadata of checkpoints created by any of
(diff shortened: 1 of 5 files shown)


mpkouznetsov (approved):
👍

docs/release-notes/4288-version.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Fixes**
+
+-  Automatically add a "determined_version" key to the metadata of checkpoints created by any of the
```
gh-determined-ai, line 5: If determined_version is an actual variable, it should be within double-backticks instead of quotes.
Also, "... in release 0.18.0." instead of "... in the 0.18.0 release."

gh-determined-ai (approved):

PR #4289, opened 2022-06-09
fix: limit waiting for logger to 30 seconds

rb-determined-ai:
If the main process exits and the logger is still going after 30
seconds, assumes something is wrong and give up.

The 30 second window is configurabe by setting DET_LOG_WAIT_TIME in the
environment, in case of emergency.

An automated test is included to make sure that we log things after the
main process exits, but not after the DET_LOG_WAIT_TIME is exceeded.

### DIFF
--- a/master/static/srv/task-logging-teardown.sh
+++ b/master/static/srv/task-logging-teardown.sh
@@ -4,7 +4,22 @@
 # command is finished.
 exec >&1- >&2- 1>&$ORIGINAL_STDOUT 2>&$ORIGINAL_STDERR
 
+# We use the bash builtin printf for getting the epoch time in seconds.
+# This requires bash 4.2 (from 2011) and it depends on strftime(3) supporting
+# the %s directive, which is not in posix.
+epoch_seconds () {
+    printf '%(%s)T\n' -1
+}
+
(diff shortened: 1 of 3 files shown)


master/static/srv/task-logging-teardown.sh
```diff
@@ -4,7 +4,14 @@
...
+# Wait for 30 seconds total for the logging to finish, otherwise just exit.
+deadline=$(($(date '+%s') + 30))
 for ((i = 0; i < $DET_LOG_WAIT_COUNT; i++)); do
-	# read returns 1 on EOF, but it's a fifo so that is OK.
-	read <$DET_LOG_WAIT_FIFO || true
+    timeout="$((deadline - $(date '+%s')))"
+    test "$timeout" -lt 0 && break
+    # read returns 1 on EOF or >128 with timeout, but it's a fifo so that is OK.
+    # for read's -t timeout feature to work, we need to open the fifo for
+    # reading and writing for some reason, which is what the `<>` is for.
```
stoksc, line 14: probably because read writes to it or something.. interesting.

stoksc (approved):

PR #4294, opened 2022-06-09
fix: check for errors from actor asks in createResourcePoolSummary [DET-7494]

nrajanee:
## Description

fix: check for errors from actor asks in createResourcePoolSummary


master/internal/resourcemanagers/agent_resource_manager.go
```diff
@@ -495,7 +495,11 @@ func (a *agentResourceManager) createResourcePoolSummary(
 		}
 	}
 
-	resourceSummary := ctx.Ask(a.pools[poolName], GetResourceSummary{}).Get().(ResourceSummary)
+	response := ctx.Ask(a.pools[poolName], GetResourceSummary{})
+	if response.Error() != nil {
+		return &resourcepoolv1.ResourcePool{}, err
```
stoksc, line 500: I would return nil rather than an empty pool
```suggestion
		return nil, err
```

stoksc (approved):
lgtm modulo one requested change

PR #4300, opened 2022-06-09
feat: Chart using Summarized Metrics API; Single Trial Workloads table using Paginated API

mapmeld:
## Description

Based on #4192 - allows the Experiment / Single Trial View to download fewer rows of metrics while displaying a useful chart and paginated table.

- Single Trial Experiment Chart loads metrics from new Summarized Metrics API ; each metric runs through the lttb downsampler (supports linear and log scales).
- Single Trial Experiment Tables load metrics from Paginated Workloads API ( #3635 ), with new fields to handle the filters on that page
- Removes workloads field from TrialDetails (contained all metrics / all workloads)
- Supporting other components and sections which previously assumed that all of the trials were available to them (these need to do an async API call, or a for loop with multiple async API calls, or get data passed by props). These typically want a sum of the checkpoint file sizes (limited to 1,000 checkpoints here?) or the latest metrics for a table

Some onZoom code is here but does not make additional calls.

## Test Plan

- Single Trial Experiment
- Multiple Trial Experiment
- Trial comparison modal from Multiple Trial Experiment, especially cases like Decision Tree / Titanic example where 'loss' is a training metric and a validation metric
- NaN and Infinite values
- Examples with 1,000 metrics or 10,000+ rows of metric-recording / batches

## Checklist


mapmeld:
oh , we might be able to use the compare endpoint to request the checkpointWorkloads or latestWorkloads in one request, instead of the for loop?

master/internal/api_trials.go
```diff
@@ -496,29 +497,110 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+	metricType apiv1.MetricType, maxDatapoints int, startBatches int,
+	endBatches int, log_scale bool) ([]*apiv1.SummarizedMetric, error) {
+	var metricSeries []lttb.Point
+	var startTime time.Time
+	// var endTime time.Time
+	var err error
+
+	var metrics []*apiv1.SummarizedMetric
+
+	for _, name := range metricNames {
```
trentwatt, line 528: for looping queries seems not ideal

master/internal/api_trials.go
```diff
@@ -496,29 +497,110 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+	}
+	return metrics
+}
+
+func (a *apiServer) MultiTrialSample(trialID int32, metricNames []string,
+	metricType apiv1.MetricType, maxDatapoints int, startBatches int,
+	endBatches int, log_scale bool) ([]*apiv1.SummarizedMetric, error) {
+	var metricSeries []lttb.Point
+	var startTime time.Time
+	// var endTime time.Time
```
trentwatt, line 523: remove comment?

master/internal/api_trials.go
```diff
@@ -496,29 +497,110 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+	var err error
+
+	var metrics []*apiv1.SummarizedMetric
+
+	for _, name := range metricNames {
+		if metricType != apiv1.MetricType_METRIC_TYPE_VALIDATION {
```
trentwatt, line 529: why not check if == training?

trentwatt (changes requested):
noticing a thing where the metric-names hangs for single trial experiments.


https://user-images.githubusercontent.com/28383080/173556230-37133c30-d2fc-4757-beee-6abe97589984.mov



my main priority is reducing entanglement between this work and the trial comparison project. I think the backend changes alone are enough to go off of for now, and the frontend is heavily subject to change as a part of project.

also i dont think the migration to move workloads to separate API call needs to happen as a part of this initial PR.

Can we: 
  - make this a PR for the backend changes EXCEPT the removal of workloads from Trial Details
  - do a separate PR later that migrates workloads to separate API
  - do another separate PR later that migrates the TrialDetails page to the new downsampled API?

master/internal/api_trials.go
mapmeld: if the metricType is Training or Unspecified (all types), then we should return training metrics.  I agree this is confusing so I will change it to an OR

mapmeld:
Divided up into new PRs

PR #4303, opened 2022-06-10
fix: Sync epoch_idx across all workers for PyTorchTrial [DET-7488]

garrett361:
## Description

Synchronize `epoch_idx` across all workers for PyTorchTrials, so that actions which key off of epochs occur at the same step for every worker.  Implemented by forcing all sharded samplers to have the same `len`. 

## Test Plan

Tested manually and with a new`e2e` test: `test_pytorch.py::test_epoch_sync`.

## Checklist


garrett361:
Please check the `api-pytorch-advanced.txt` docs changes in particular. The order for repeating and sharding does not seem to matter to me, at least not for the reasons stated in the present version of the docs.

harness/determined/pytorch/samplers.py
```diff
@@ -105,9 +108,9 @@ def __init__(
         self.rank = rank
 
     def __len__(self) -> int:
-        full_global_batches = len(self.batch_sampler) // self.num_workers
-        worker_gets_partial_batch = int(len(self.batch_sampler) % self.num_workers > self.rank)
-        return full_global_batches + worker_gets_partial_batch
+        full_global_batches, extra_batches = divmod(len(self.batch_sampler), self.num_workers)
```
rb-determined-ai, line 111: wow, every time I think I know all the builtins in python, somebody shows me a new one.

docs/training-apis/api-pytorch-advanced.txt
```diff
@@ -62,16 +62,11 @@ training") is easy if you follow a few rules.
...
--  **Always shard, and not before a repeat**: Use Determined's DistributedSampler or
-   DistributedBatchSampler to provide a unique shard of data to each worker based on your sampler or
-   batch_sampler. It is best to always shard your data, and even when you are not doing distributed
-   training, because in non-distributed-training settings, the sharding is nearly zero-cost, and it
-   makes distributed training seamless if you ever want to use it in the future.
-
-   It is generally important to shard after you repeat, unless you can guarantee that each shard of
-   the dataset will have the same length. Otherwise, differences between the epoch boundaries for
-   each worker can grow over time, especially on small datasets. If you shard after you repeat, you
-   can change the number of workers arbitrarily without issue.
```
rb-determined-ai, line 74: This is actually still true... just the on_epoch_ends will do a better job at pretending it's not true.

harness/determined/pytorch/samplers.py
```diff
@@ -83,14 +89,11 @@ class DistributedBatchSampler(torch.utils.data.BatchSampler):
...
 
-    DistributedBatchSampler has the potential gotcha that when wrapping a
-    non-repeating BatchSampler, if the length of the BatchSampler is not
-    divisible by the number of replicas the length of the resulting
-    DistributedBatchSampler will differ based on the rank. In that case, the
-    divergent paths of multiple workers could cause problems during training.
-    PyTorchTrial always uses RepeatBatchSampler during training, PyTorchTrial
-    does not require that the workers stay in step during validation, so this
-    potential gotcha is not a problem in Determined.
+    DistributedBatchSampler reports equivalent len values for every worker, regardless of rank.
```
rb-determined-ai, line 92: Can we have the chief broadcast how long its shard is to the workers, and the workers use that, instead of having the dataset lie about its length?

I think, when we do support true-epoch-based training, this would be an undesirable behavior, and if you made it configurable, it would be difficult to document.  On the other hand, "in non-epoch based training, everybody gets the on_epoch_end call when the chief's epoch ends" feels more intuitive to explain, although it's a bit more complex to implement.

docs/training-apis/api-pytorch-advanced.txt
garrett361: Yeah, my thinking here was that if all of the data is shuffled anyway, then the sharding/repeating order hardly matters, but this makes too many assumption. I'll revert it.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -153,7 +154,14 @@ def _set_data_loaders(self) -> None:
...
 
-        self.context._epoch_len = len(self.training_loader)
+        # All workers use the chief's definition of epoch lengths, which is based on the training
+        # loader's len. If this len does not exist, epoch durations cannot be deduced, and they
+        # default to max int.
+        try:
+            epoch_len = len(self.training_loader)
+        except TypeError:
+            epoch_len = sys.maxsize
+        self.context._epoch_len = self.context.distributed.broadcast(epoch_len)
```
rb-determined-ai, line 164: this is a good start, but what do we do about variable-length datasets?

Eventually we'll need to break the assumption that the dataset length doesn't change, and we'll have to keep counters of epoch idx, rather than calculate it with modulo math.

... why not now?

rb-determined-ai (approved):
Looks good to me.  You can decide how much scope creep you are willing to accept, vs what you just put in a ticket for later.

PR #4306, opened 2022-06-11
ci: skip e2e_tests for webui

trentwatt:
## Checklist


rb-determined-ai (approved):
This seems good to me, albeit dangerous.

The reason I think it is good is that the e2e tests don't actually check the webui at all, so there's no path for a webui-only change to break them.

The reason it's dangerous is that having an easy command to skip some tests is a strong temptation.  But as long as we don't throw it around willy-nilly we should be ok.  Also, IMO, the risk of misuse is not a good reason to make devs wait on pointless tests.

PR #4307, opened 2022-06-13
chore: add eslint rules

hkang1:
## Description

Add 2 eslint rules:

[array-element-newline](https://eslint.org/docs/rules/array-element-newline) - if there is a new line in the array, force each item to be on its own line
[arrow-paren](https://eslint.org/docs/rules/arrow-parens#:~:text=Requires%20parens%20in%20arrow%20function,of%20parentheses%20in%20arrow%20functions.) - always requiring a paren around the arrow function params

NOTE: will land this later with all the corresponding fixes after workspaces.

## Checklist

### DIFF
--- a/webui/react/.eslintrc.js
+++ b/webui/react/.eslintrc.js
@@ -48,7 +48,11 @@ module.exports = {
     ],
     '@typescript-eslint/prefer-optional-chain': [ 'error' ],
     'array-bracket-spacing': [ 'error', 'always' ],
-    'array-element-newline': [ 'error', 'consistent' ],
+    'array-element-newline': [ 'error', {
+      ArrayExpression: 'consistent',
+      ArrayPattern: { multiline: true },
+    } ],
+    'arrow-parens': [ 'error', 'always' ],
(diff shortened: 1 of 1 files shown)


webui/react/.eslintrc.js
```diff
@@ -48,7 +48,11 @@ module.exports = {
     ],
     '@typescript-eslint/prefer-optional-chain': [ 'error' ],
     'array-bracket-spacing': [ 'error', 'always' ],
-    'array-element-newline': [ 'error', 'consistent' ],
+    'array-element-newline': [ 'error', {
+      ArrayExpression: 'consistent',
+      ArrayPattern: { multiline: true },
+    } ],
+    'arrow-parens': [ 'error', 'always' ],
```
hamidzr, line 55: "as-needed" enforces no parens where they can be omitted.
?

hkang1: I was originally on the allowing them to be omitted, but as we write more code, I think forcing the parens is going to be beneficial with a larger team (allowing less variance for diffs in general)

PR #4308, opened 2022-06-13
fix: skip allocation check on checkpoint save to fix pulling preemption

maxrussell:
…  issue for [DET-7511]

## Description

When runs were refactored out of the code, this allocation check was brought forward. @stoksc and I talked and agreed that it seems extraneous.

Fixes [DET-7511].

### DIFF
--- a/master/internal/db/postgres_trial.go
+++ b/master/internal/db/postgres_trial.go
@@ -291,32 +291,17 @@ DO NOTHING
 func (db *PgDB) AddCheckpointMetadata(
 	ctx context.Context, m *model.CheckpointV2,
 ) error {
-	return db.withTransaction("add checkpoint metadata", func(tx *sqlx.Tx) error {
-		var latestAllocationID model.AllocationID
-		switch err := tx.QueryRowx(`
-SELECT a.allocation_id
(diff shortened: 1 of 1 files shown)


master/internal/db/postgres_trial.go
```diff
@@ -291,32 +291,17 @@ DO NOTHING
...
 VALUES
-	(:uuid, :task_id, :allocation_id, :report_time, :state, :resources, :metadata)
-`, m); err != nil {
-			return errors.Wrap(err, "inserting checkpoint")
-		}
-		return nil
-	})
+	(:uuid, :task_id, :allocation_id, :report_time, :state, :resources, :metadata)`
+
+	if _, err := db.sql.NamedExec(query, m); err != nil {
```
stoksc, line 300: let's use `NamedExecContext` and pass the context. one day want to do much better in the system with contexts.

stoksc (approved):
lgtm modulo single comment.

master/internal/db/postgres_trial.go
maxrussell: Good catch! Thanks!

PR #4311, opened 2022-06-13
chore: drop pbt, adaptive and adaptive_simple from docs and web

hkang1:
## Description

The support for the searchers `pbt`, `adaptive` and `adaptive_simple` was dropped previously, updating the loose ends around them.

## Checklist


harness/determined/common/schemas/expconf/_v0.py
```diff
@@ -630,7 +630,7 @@ def __init__(
     ) -> None:
         pass
 
-
+# Leave this alone for backwards compatibility?
 @SearcherConfigV0.member("pbt")
 class PBTConfigV0(schemas.SchemaBase):
```
rb-determined-ai, line 635: 100% safe to delete

webui/react/src/types.ts
```diff
@@ -253,11 +253,8 @@ export type HyperparametersFlattened = {
 };
 
 export enum ExperimentSearcherName {
-  AdaptiveAdvanced = 'adaptive',
```
rb-determined-ai, line 256: Do you use this to parse experiment configs that you get from the master?  If so, this would be high-risk, because old experiments might fail to load if they were created when PBT was allowed.

rb-determined-ai (approved):

PR #4315, opened 2022-06-13
fix: provide unique port offsets to trials

rb-determined-ai:
The DET_UNIQUE_PORT_OFFSET was always coming out to zero, which
prevented host-mode networking in multi-tenant dtrain scenarios.

The source of the bug was that the unique port offset was being
calculated before the TaskSpec.Devices was populated.

## Test Plan

- Configure `task_container_defaults.network_mode: host` in master
- Start an `artificial_slots:4` agent
- Start two simultaneous `slots_per_trial: 2` trials
- Make sure both trials succeed.

## Commentary (optional)

This bug was difficult to track down because the TaskSpec means multiple different things at different points in time (something I did not previously realize).  Initially, a trial actor is started with a "base" TaskSpec.  I think this has some defaults from the master.  Then the TrialSpec (which has a base TaskSpec as its member) creates a new TaskSpec in TrialSpec.ToTaskSpec.  Why does TrialSpec contain a TaskSpec and still have a ToTaskSpec() method?  Seems confusing to me.  Then later, the containerResources.Start fills the remaining fields of the TaskSpec, to create the actually usable TaskSpec.

Probably these should be three different structs, each extending the previous with struct composition, which is pretty nice and easy in Go.

### DIFF
--- a/master/pkg/tasks/task_trial.go
+++ b/master/pkg/tasks/task_trial.go
@@ -100,15 +100,18 @@ func (s TrialSpec) ToTaskSpec(keys *ssh.PrivateAndPublicKeys) TaskSpec {
 	res.Entrypoint = []string{"/run/determined/train/entrypoint.sh"}
 
 	envVars := map[string]string{
-		"DET_EXPERIMENT_ID":      strconv.Itoa(s.ExperimentID),
-		"DET_TRIAL_ID":           strconv.Itoa(s.TrialID),
-		"DET_TRIAL_RUN_ID":       strconv.Itoa(s.TrialRunID),
-		"DET_TRIAL_SEED":         strconv.FormatUint(uint64(s.TrialSeed), 10),
-		"DET_EXPERIMENT_CONFIG":  jsonify(s.ExperimentConfig),
-		"DET_HPARAMS":            jsonify(s.HParams),
-		"DET_STEPS_COMPLETED":    strconv.Itoa(s.StepsCompleted),
-		"DET_UNIQUE_PORT_OFFSET": strconv.Itoa(trialUniquePortOffset(s.Base.Devices)),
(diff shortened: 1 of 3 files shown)


master/pkg/tasks/task_trial.go
```diff
@@ -100,15 +100,18 @@ func (s TrialSpec) ToTaskSpec(keys *ssh.PrivateAndPublicKeys) TaskSpec {
...
+		"DET_EXPERIMENT_ID":     strconv.Itoa(s.ExperimentID),
+		"DET_TRIAL_ID":          strconv.Itoa(s.TrialID),
+		"DET_TRIAL_RUN_ID":      strconv.Itoa(s.TrialRunID),
+		"DET_TRIAL_SEED":        strconv.FormatUint(uint64(s.TrialSeed), 10),
+		"DET_EXPERIMENT_CONFIG": jsonify(s.ExperimentConfig),
+		"DET_HPARAMS":           jsonify(s.HParams),
+		"DET_STEPS_COMPLETED":   strconv.Itoa(s.StepsCompleted),
+		"DET_TASK_TYPE":         string(model.TaskTypeTrial),
+		// DET_UNIQUE_PORT_OFFSET will be overwritten by the agent resource manager when the
+		// container is started, but only on agent resource pools.  This default value will apply
```
stoksc, line 112: It is actually overrides by slurm too, just in a startup hook we run for slurm stuff.

stoksc (approved):
lgtm

master/pkg/tasks/task_trial.go
rb-determined-ai: Not sure I fully understand.  What should this comment say then?

stoksc: Was just a note that this is not the value slurm uses.

PR #4323, opened 2022-06-14
feat: task container checks for occupied gpus [DET-5091]

drh-determined-ai:
## Test Plan

In a Python shell on the agent machine, allocate some memory on the CUDA GPU. For example,
```
import torch
x = torch.FloatTensor([0., 1., 2.]).to(torch.device("cuda:0"))
assert x.is_cuda
```
Keeping this shell open, run `nvidia-smi` and note the pid of the Python shell; create a Determined experiment you expect to utilize GPU(s) as normal. The task should log a warning referencing the Python pid.

## Commentary

Logging configuration was missing from `prep_container.py`. This had to be implemented because it is an entry-level script; the configuration affects `gpu.py`, which contains the new warning.

The call of `nvidia-smi` in `gpu.py` gives `"[Not Found]"` for the process name. This is likely due to the command being run inside a Docker container. The pid is correct, though, and this seems sufficient for now.

## Checklist

### DIFF
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -241,8 +241,14 @@ if __name__ == "__main__":
         info = det.ClusterInfo._from_env()
         info._to_file()
 
+    try:
+        # See the ClusterInfo.trial property for explanation
+        debug = info.trial._debug
+    except (AssertionError, RuntimeError):
+        debug = False
+
     logging.basicConfig(
-        level=logging.DEBUG if info.task_type == "TRIAL" and info.trial._debug else logging.INFO,
+        level=logging.DEBUG if debug else logging.INFO,
         format=det.LOG_FORMAT,
     )
     logging.debug("running prep_container")


harness/determined/exec/prep_container.py
```diff
@@ -255,6 +267,9 @@ def do_proxy(sess: Session, allocation_id: str) -> None:
 
     if args.resources:
         det.ResourcesInfo._by_inspection()._to_file()
+        from determined.gpu import check_for_gpu_processes
```
rb-determined-ai, line 270: always prefer importing at the very top, without very good reason to do otherwise.

In this case, `determined.gpu` is a pretty trivial file to import, so there's not a good reason.

Also, for python imports, please stick to the [google style guide](https://google.github.io/styleguide/pyguide.html#22-imports).  So this should be:

```python
from determined import gpu
...
gpu.check_for_gpu_processes
```

harness/determined/exec/prep_container.py
```diff
@@ -255,6 +267,9 @@ def do_proxy(sess: Session, allocation_id: str) -> None:
...
+
+        check_for_gpu_processes()
```
rb-determined-ai, line 272: Can we separate the business logic (warning the user) from the helper logic (listing processes from nvidia-smi)?

Basically, let the gpu module have a `get_gpu_processes()` call, but emit the warning from here.

harness/determined/exec/prep_container.py
```diff
@@ -233,8 +233,20 @@ def do_proxy(sess: Session, allocation_id: str) -> None:
     parser.add_argument("--resources", action="store_true")
     parser.add_argument("--rendezvous", action="store_true")
     parser.add_argument("--proxy", action="store_true")
+    parser.add_argument(
+        "--log-level",
+        default=os.getenv("DET_LOG_LEVEL", "INFO"),
+        choices=["DEBUG", "INFO", "WARNING", "ERROR"],
+        help="Set the logging level",
+    )
```
rb-determined-ai, line 241: instead of making this configurable as a cli argument (the user cannot control which cli arguments get passed here), can you make it either INFO or DEBUG, based on the value of `info._debug`, which comes from the experiment config?

rb-determined-ai (commented):
looks pretty good.

harness/determined/gpu.py
```diff
@@ -119,3 +119,56 @@ def get_gpus() -> List[GPU]:
...
+                logging.warning(f"Ignoring GPU process with unexpected nvidia-smi output: {fields}")
+    if proc.returncode:
+        logging.warning(f"nvidia-smi exited with failure status code {proc.returncode}")
+
+    logging.info(f"detected {len(processes)} gpu processes")
+    return processes
+
+
+def get_gpu_processes() -> List[GPUProcess]:
+    return _get_nvidia_processes()
```
rb-determined-ai, line 174: why not just name the original one `get_gpu_processes()`?

harness/determined/gpu.py
```diff
@@ -119,3 +119,56 @@ def get_gpus() -> List[GPU]:
...
+                        process_name=fields["process_name"],
+                        gpu_uuid=fields["gpu_uuid"].strip(),
+                    )
+                )
+            except ValueError:
+                logging.warning(f"Ignoring GPU process with unexpected nvidia-smi output: {fields}")
+    if proc.returncode:
+        logging.warning(f"nvidia-smi exited with failure status code {proc.returncode}")
+
+    logging.info(f"detected {len(processes)} gpu processes")
```
rb-determined-ai, line 169: this isn't really helpful to the user.  The only place this function is called, we already print a message when the user needs to care.  So let's delete this line.

harness/determined/exec/prep_container.py
```diff
@@ -255,6 +267,12 @@ def do_proxy(sess: Session, allocation_id: str) -> None:
 
     if args.resources:
         det.ResourcesInfo._by_inspection()._to_file()
+        for process in gpu.get_gpu_processes():
+            logging.warning(
+                f"process {process.process_name} "
+                f"with pid {process.pid} "
+                f"is using gpu with uuid {process.gpu_uuid}"
```
rb-determined-ai, line 274: perhaps expand a bit more on why the user should care about this

harness/determined/gpu.py
drh-determined-ai: I thought it reasonable to include an explicit "0 processes" message at the INFO level. On second thought, though, the number of running processes will only be interesting to the user if and only if they want to be warned about a non-zero number of processes, which they'll get.
So yeah, I'll reduce the clutter.

harness/determined/exec/prep_container.py
drh-determined-ai: I have no idea why a user might care about this. Did the person that requested this feature explain why? If not - or if that's not the only possible reason - all I can do is present the information.
There's more information that I can pull out of `nvidia-smi`, if that's worthwhile.

harness/determined/gpu.py
rb-determined-ai: well... yeah a TODO would also be sufficient explanation I think.

harness/determined/exec/prep_container.py
```diff
@@ -241,6 +241,18 @@ def do_proxy(sess: Session, allocation_id: str) -> None:
         info = det.ClusterInfo._from_env()
         info._to_file()
 
+    try:
+        # See the ClusterInfo.trial property for explanation
+        debug = info.trial._debug
+    except (AssertionError, RuntimeError):
+        debug = False
```
rb-determined-ai, line 248: how about:
```python
debug = info.task_type == "TRIAL" and info.trial._debug
```
?

It's shorter, and follows the usage pattern described in the ClusterInfo docstring.

rb-determined-ai (approved):

harness/determined/exec/prep_container.py
drh-determined-ai: Because that still raises an exception! You can see here: http://34.168.20.150:8080/det/experiments/15/logs

PR #4367, opened 2022-06-15
fix: remove spinner for the InlineEditor

thiagodallacqua-hpe:
## Description

Original ticket: DET-7344.

An inline spinner was showing after changing the experiment name or details overflowing and breaking the UI.

After some debugging, I've managed to pinpoint and add an optional prop for the `InlineEditor` to opt-out of showing the spinner on the `onSave` event.

## Test Plan

Tested out using `make live` and it all seems to be fixed.

## Commentary (optional)

I've removed some additional fetching due to the nature of the code, for more insight, please check the comments.

## Checklist


webui/react/src/pages/ExperimentDetails/ExperimentDetailsHeader.tsx
```diff
@@ -152,7 +152,6 @@ const ExperimentDetailsHeader: React.FC<Props> = ({
   const handleDescriptionUpdate = useCallback(async (newValue: string) => {
     try {
       await patchExperiment({ body: { description: newValue }, experimentId: experiment.id });
-      await fetchExperimentDetails();
```
thiagodallacqua-hpe, line 155: Removed this additional fetching for two reasons.

1. the main reason it is there is to update the state of the experiment, which is ok, but, since it's an atomic change already reflected in the UI, the UI itself already is reflecting the optimistic rendering of the outcome, making this request avoidable.
2. we're wrapping this change in a `try... catch`, which means that, if it goes bad, the catch would alert the user that something didn't work.
3. there are more of this code in the component, but, it felt kinda unsafe to remove for now.

webui/react/src/pages/ExperimentDetails/ExperimentDetailsHeader.tsx
```diff
@@ -163,12 +162,11 @@ const ExperimentDetailsHeader: React.FC<Props> = ({
       });
       return e as Error;
     }
-  }, [ experiment.id, fetchExperimentDetails ]);
+  }, [ experiment.id ]);
 
   const handleNameUpdate = useCallback(async (newValue: string) => {
     try {
       await patchExperiment({ body: { name: newValue }, experimentId: experiment.id });
-      await fetchExperimentDetails();
```
thiagodallacqua-hpe, line 171: same here

trentwatt:
Haven't dug into the code, but it would be good if we can fix the problem at the `InlineEditor.tsx` level-- by making it so where the `Spinner` is superimposed on the text if there is not room for it off to the side.

That way it doesn't appear in other ways later on.

webui/react/src/components/InlineEditor.tsx
```diff
@@ -142,9 +144,13 @@ const InlineEditor: React.FC<Props> = ({
           onKeyPress={handleTextareaKeyPress}
           onKeyUp={handleTextareaKeyUp}
         />
-        <div className={css.spinner}>
-          <Spinner spinning={isSaving} />
-        </div>
+        {
+          showSpinner && (
```
trentwatt, line 148: For the sake of consistency across our UI, I don't think this should be a prop.

webui/react/src/pages/ExperimentDetails/ExperimentDetailsHeader.tsx
```diff
@@ -371,6 +370,7 @@ const ExperimentDetailsHeader: React.FC<Props> = ({
                 isOnDark
                 maxLength={128}
                 placeholder="experiment name"
+                showSpinner={false}
```
trentwatt, line 373: we still want spinner to appear when it's saving, we just don't want it to change the screen layout.

webui/react/src/components/InlineEditor.module.scss
```diff
@@ -51,10 +51,18 @@
...
+    top: 45%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+  }
+  .growWrap > .backdrop {
+    display: none;
+    position: absolute;
+    background-color: rgba(0, 0, 0, .4);
+    width: 100%;
+    height: 100%;
```
thiagodallacqua-hpe, line 65: added a backdrop to improve the visibility and "hierarchy" of the spinner in the component.

https://user-images.githubusercontent.com/104855841/174311746-1752801c-4998-4bee-8db1-bf79d9546f0e.mp4

webui/react/src/components/InlineEditor.module.scss
```diff
@@ -73,10 +81,8 @@
 }
 .base.loading {
   .growWrap > textarea,
-  .growWrap::after {
-    margin-right: 36px;
-  }
```
thiagodallacqua-hpe, line 78: this was making the component sort of bouncy overall...

trentwatt:
Ok, sounds like we got the okay from @alexjdetermined about the spinner position.

But the point about not needing to refetch after update is no longer true, since in latest-master we have experiment name appearing in two places

without the `fetchExperimentDetails()` it is slow to update- depending on where you are in the polling cycle. below is what I get after rebasing your branch off of master


https://user-images.githubusercontent.com/28383080/176355623-9bd818ba-566d-47be-b32c-01b93af01317.mov

@thiagodallacqua-hpe can you rebase yours off of master and add back the `fetchExperimentDetails()` . with the removal of submodules `git` gets mad if you try to rebase in the normal way. but if you do the below, you should be good:


```
git checkout -f master
git fetch upstream
git pull upstream master
git rebase master spinner-overflow
```

also please update the PR title and description to reflect the current state of the branch 🙏

webui/react/src/components/InlineEditor.module.scss
```diff
@@ -51,10 +51,18 @@
...
-    right: 18px;
-    top: 18px;
+    top: 45%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+  }
+  .growWrap > .backdrop {
+    display: none;
+    position: absolute;
+    background-color: rgba(0, 0, 0, .4);
```
trentwatt, line 63: use theme vars instead of hardcoding colors

webui/react/src/components/InlineEditor.module.scss
```diff
@@ -51,10 +51,18 @@
     border-color: var(--theme-colors-monochrome-14);
   }
   .growWrap > .spinner {
-    display: hidden;
+    display: none;
```
trentwatt, line 54: question: what is this for?

trentwatt (approved):
Thanks for this, it looks really good.

Small request about using [theme vars](https://determinedai.atlassian.net/wiki/spaces/ENGINEERIN/pages/1155694597/Theme+Architecture#Interactive) instead of hardcoding colors

PR #4370, opened 2022-06-16
chore: change notebook `report_idle` ping interval from 1 to 30 seconds.

ioga:
## Description

the notebooks [ping master every second](https://github.com/determined-ai/determined/blob/master/master/static/srv/check_idle.py#L95) for the idle check. default idle timeouts for notebooks and tensorboards are 20 minutes and 5 minutes. one second seems excessive on that scale, lets bump it to 30 seconds.

## Commentary (optional)

Discussed in #ml-ag couple weeks back, but then I lost track of it with the offsite and being sick.

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

PR #4371, opened 2022-06-16
feat: rolling upgrades support for generic command types (notebooks etc.) [DET-7218]

ioga:
## Description

- restore/continue notebooks, tensorboards, commands, shells after master restart.
- also enable `det (notebook|shell|tensorboard) list --json`.

## Test Plan

There're some tests.

## Commentary (optional)

This PR has #4370 embedded in it. This'll be solved with a rebase when it is merged.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/managed_cluster.py
+++ b/e2e_tests/tests/cluster/managed_cluster.py
@@ -89,17 +89,7 @@ class ManagedCluster:
         self.dc.restart_stage("agent1", wait=True, timeout=10)
 
         WAIT_FOR_STARTUP = 10
-        for _i in range(WAIT_FOR_STARTUP):
-            agent_data = _get_agent_data(self.master_url)
-            if (
-                len(agent_data) == 1
-                and agent_data[0]["enabled"] is True
-                and agent_data[0]["draining"] is False
-            ):
-                break
-            time.sleep(1)
-        else:
-            pytest.fail(f"Agent didn't restart after {WAIT_FOR_STARTUP} seconds")
+        self.wait_for_agent_ok(WAIT_FOR_STARTUP)
 
     def kill_proxy(self) -> None:
         subprocess.run(["killall", "socat"])
@@ -125,6 +115,19 @@ class ManagedCluster:
         assert agent_data[0]["enabled"] is True
         assert agent_data[0]["draining"] is False
 
+    def wait_for_agent_ok(self, ticks: int) -> None:
+        for _i in range(ticks):
+            agent_data = _get_agent_data(self.master_url)
+            if (
+                len(agent_data) == 1
+                and agent_data[0]["enabled"] is True
+                and agent_data[0]["draining"] is False
+            ):
+                break
+            time.sleep(1)
+        else:
+            pytest.fail(f"Agent didn't restart after {ticks} seconds")
+
     def fetch_config(self) -> Dict:
         with logged_in_user(ADMIN_CREDENTIALS):
             master_config = json.loads(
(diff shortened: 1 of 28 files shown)


e2e_tests/tests/cluster/test_master_restart.py
```diff
@@ -112,3 +137,145 @@ def test_master_restart_kill_works(managed_cluster: ManagedCluster) -> None:
...
+        r = requests.get(url, headers={"Authorization": f"Bearer {token}"}, allow_redirects=True)
+        # Sometimes the TB/JL take a bit of time to stand up, returning 502.
+        # Sometimes it takes a bit of time for master to register the proxy, returning 404.
+        if r.status_code == 502 or r.status_code == 404:
+            time.sleep(1)
+            bad_status_codes.append(r.status_code)
+            continue
+        r.raise_for_status()
+        html = r.content.decode("utf-8")
+        assert name in html  # Brutal.
```
stoksc, line 213: lol. for what it is worth, the way the idle check works may be better, at least for notebooks.

e2e_tests/tests/cluster/utils.py
```diff
@@ -46,7 +47,7 @@ def run_command(sleep: int = 30, slots: int = 1) -> str:
         "run",
         "-d",
         "--config",
-        "resources.slots=0",
+        f"resources.slots={slots}",
```
stoksc, line 50: aha.. yes! good catch.

master/internal/command/command.go
```diff
@@ -195,6 +297,10 @@ func (c *command) Receive(ctx *actor.Context) error {
 			JobActor: ctx.Self(),
 		})
 
+		ctx.Ask(c.allocation, actor.Ping{}).Get()
```
stoksc, line 300: this is a bit of a risk - if it gets resources immediately, it will also ask us back to build the task spec..

master/internal/command/command.go
```diff
@@ -475,17 +586,56 @@ func (c *command) toShell(ctx *actor.Context) *shellv1.Shell {
...
+		Id:             c.stringID(),
 		State:          state.State.Proto(),
 		Description:    c.Config.Description,
 		StartTime:      protoutils.ToTimestamp(ctx.Self().RegisteredTime()),
 		Container:      state.FirstContainer().ToProto(),
-		ServiceAddress: *c.serviceAddress,
-		ExperimentIds:  c.Metadata["experiment_ids"].([]int32),
-		TrialIds:       c.Metadata["trial_ids"].([]int32),
+		ServiceAddress: c.serviceAddress(),
+		ExperimentIds:  recastMetadataToInt32(c.Metadata["experiment_ids"]),
```
stoksc, line 637: what a leaky abstraction. maybe metadata could be a union instead? not blocking at all, just food for thought.

master/internal/command/command.go
ioga: it's the same as trial restoration. if I were asking for `getSummary` like `createGenericActor` does, then yeah it hangs because of this, but works okay for Ping.

e2e_tests/tests/cluster/test_master_restart.py
ioga: looks too complicated for a test

e2e_tests/tests/cluster/managed_cluster.py
```diff
@@ -146,7 +149,7 @@ def log_marker(self, marker: str) -> None:
                 fout.write(marker)
 
 
-@pytest.fixture(scope="session", params=[False, True], ids=["reattach-off", "reattach-on"])
+@pytest.fixture(scope="session", params=[True, False], ids=["reattach-on", "reattach-off"])
```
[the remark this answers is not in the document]
ioga: yep I've also run into it. probably broke in one of my refactorings

master/internal/command/command.go
ioga: @stoksc pushed the static typing. looks much better.

PR #2529, opened 2021-06-08
chore: doc echo apis

stoksc:
## Description
This change adds swag (https://github.com/swaggo/swag) to doc echo APIs. This is useful because some APIs cannot be expressed within the constraints of grpc<->grpc-gateway land, e.g. endpoints that return CSV, or generally endpoints that accept or return MIME types other than application/json.

## Test Plan
- [x] build docs, make sure they look good.

## Commentary (optional)
eh.. the makefile change is kinda bad, build proto/swagger needs to be pulled up a level, I guess, since now it depends on stuff in ./master and ./proto but I wanted to get an opinion on this before I do any of that.

## Checklist

- [x] User-facing API changes need the "User-facing API Change" label.
- [x] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [x] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234

PR #4374, opened 2022-06-16
feat: api to expose experiment model code [DET-7465]

gt2345:
## Checklist


e2e_tests/tests/experiment/test_model_def.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from tests import config as conf
+from tests import experiment as exp
+
+
+@pytest.mark.e2e_cpu
+def test_fetch_experiment_model_def() -> None:
```
rb-determined-ai, line 8: question: is there a way to run this as an integration test in go or something?  It seems a bit expensive to _actually_ run an experiment in order to test a REST API for fetching experiment configs?

master/internal/api_experiment.go
```diff
@@ -1390,3 +1394,47 @@ func (a *apiServer) MoveExperiment(
...
+func (a *apiServer) GetModelDefTree(
+	_ context.Context, req *apiv1.GetModelDefTreeRequest,
+) (*apiv1.GetModelDefTreeResponse, error) {
+	tgz, err := a.m.db.ExperimentModelDefinitionRaw(int(req.ExperimentId))
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"error fetching model definition from database: %d", req.ExperimentId)
+	}
+	var resp apiv1.GetModelDefTreeResponse
+	arc, err := archive.FromTarGz(tgz)
```
rb-determined-ai, line 1407: To me, this suggests that we need to consider _not_ storing model definitions in the database anymore.

This is going to be pretty expensive: we're going to query up to 100MB from the database and gunzip all of it in order to get just a list of files in the checkpoint?

The GetModelDefFile has similar shortcomings, it's going to query/gunzip up to 100MB of data to fetch a single file.

And I'd bet that somewhere, somebody is going to try to call GetModelDefFile once for every file in the GetModelDefTree, which is going to query/gunzip up to 100MB of data once for every single file in the model definition.

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -139,3 +139,17 @@ message ValidationHistoryEntry {
...
+  // Path of file.
+  string path = 1;
+  // Modification time of file.
+  google.protobuf.Timestamp modified_time = 2;
+  // Number of bytes in file content.
+  int32 content_length = 3;
+  // Is this a directory.
+  bool is_dir = 4;
+  // MIME type of file.
+  string content_type = 5;
```
rb-determined-ai, line 154: we don't know the mime-type of files, so we probably shouldn't be pretending that we do.

Also, we support symbolic links in archives, but I don't see that those are actually supported in this message type.  How would we support those?  301 redirects or something?

master/internal/core_experiment.go
```diff
@@ -88,6 +88,37 @@ func (m *Master) getExperimentCheckpointsToGC(c echo.Context) (interface{}, erro
 		args.ExperimentID, args.ExperimentBest, args.TrialBest, args.TrialLatest)
 }
 
+func (m *Master) getExperimentModelFile(c echo.Context) error {
```
rb-determined-ai, line 91: I see we are using the old echo server to download the file.  Is that because the grpc server doesn't let us return the plain content in a browser-friendly fashion?

master/cmd/determined-master/root.go
```diff
@@ -55,6 +56,12 @@ func runRoot() error {
 	}
 	log.Infof("master configuration: %s", printableConfig)
 
+	err = os.MkdirAll(config.Cache.CacheDir, fs.ModePerm)
```
rb-determined-ai, line 59: This is a security issue.  fs.ModePerm is `0777` permissions, which would allow any user on the master node to access the content of model definitions.

Let's use the minimum possible permissions, which would be `0700` (read/write/execute for user only, nothing for group and nothing for world).

master/cmd/determined-master/root.go
```diff
@@ -55,6 +56,12 @@ func runRoot() error {
...
+	if err != nil {
+		log.WithError(err).Errorf("Failed allocating cache direatory at %s", config.Cache.CacheDir)
```
rb-determined-ai, line 61: Nit: it is not called "allocating" a directory.  How about: `failed to make cache directory (%s)`?

proto/src/determined/api/v1/experiment.proto
```diff
@@ -575,3 +575,29 @@ message MoveExperimentRequest {
...
+  // The id of the experiment.
+  int32 experiment_id = 1;
+  // The path of file.
+  string path = 2;
+}
+
+// Response to GetModelDefFileRequest.
+message GetModelDefFileResponse {
+  // Content of file.
+  bytes file = 1;
```
rb-determined-ai, line 602: question:  What does the grpc gateway do with bytes objects?  Does it try to treat them like strings?

Also... why bother with a proto version of this REST API when you need the echo version to return the bare file anyway?

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+func (f *FileCache) getOrCreateFolder(expID int) (*modelDefFolder, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	value, ok := f.caches[expID]
+	if !ok {
+		exp := struct {
+			ModelDefinition []byte
+		}{}
+		err := Bun().NewSelect().TableExpr(
+			"experiments").Column("model_definition").Where("id = ?", expID).Scan(context.TODO(), &exp)
```
rb-determined-ai, line 55: what is the backend-eng team's stance on context.TODO?

It seems to me like it is easy to pass the real request's context into this code.  I would imagine that we should do that here.

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+		err = os.MkdirAll(f.genPath(expID, ""), fs.ModePerm)
+		if err != nil {
+			return nil, err
+		}
+		for _, ar := range arc {
+			if ar.IsDir() {
+				err = os.MkdirAll(f.genPath(expID, ar.Path), fs.ModePerm)
+			} else {
+				err = os.WriteFile(f.genPath(expID, ar.Path), ar.Content, fs.ModePerm)
+			}
```
rb-determined-ai, line 73: What is the desired behavior for symbolic links?

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+		exp := struct {
+			ModelDefinition []byte
+		}{}
+		err := Bun().NewSelect().TableExpr(
+			"experiments").Column("model_definition").Where("id = ?", expID).Scan(context.TODO(), &exp)
+		if err != nil {
+			return nil, err
+		}
+		var fileTree []*experimentv1.FileNode
+		arc, err := archive.FromTarGz(exp.ModelDefinition)
```
rb-determined-ai, line 60: Is there a standard library for untaring tgz files to disk?  It seems like perhaps we should not be writing it by hand.

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+			fileTree:   fileTree,
+			cachedTime: time.Now(),
+		}
+		f.caches[expID] = value
+		f.prune()
+	}
+	return value, nil
+}
+
+func (f *FileCache) prune() {
```
rb-determined-ai, line 97: this prune code does not seem to take into account any files which were present on disk before the master started up.  Effectively, when the master restarts, any files on disk have been "leaked" and will never be reused or cleaned up.

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -139,3 +139,21 @@ message ValidationHistoryEntry {
...
+  // Name of file.
+  string name = 2;
+  // Modification time of file.
+  google.protobuf.Timestamp modified_time = 3;
+  // Number of bytes in file content.
+  int32 content_length = 4;
+  // Is this a directory.
+  bool is_dir = 5;
+  // MIME type of file.
+  string content_type = 6;
```
rb-determined-ai, line 156: how important is it that we deliver the `content_type` in the filetree?  I imagine the answer is "it's pretty important to the webui" but I think it complicates the filecache implementation quite a bit so I want to make sure.

rb-determined-ai (commented):
The file cache implementation has some issues right now:
- it treats the cache dir as being empty on startup, which may not be true
- it keeps an in-memory cache, rather than just using the file cache.  Effectively, this creates two sources of truth.  Each source is immutable, so it's not a terrible situation, but it's also a waste of memory
- pruning is only done based on the contents of the in-memory cache, so the filesystem can easily leak files that never get cleaned up.

If you did rely more on what's present in the filesystem (which I think you should), then you would have another couple issues right away:
- You write to the cache-dir non-atomically.  It would be better to untar things into a temporary space, and rename the top-level directory into place.  That way, if the master node looses power halfway through writing (the worst case scenario, because even `defer` statements wouldn't run) you'd have some parts of the cache which appear valid but which are only half-written.
- You don't currently have any mechanism for storing the content_type on the filesystem.

An alternate strategy would be to stick to the in-memory cache and only use the filesystem cache as a key-value store of file content.  In that world, you would probably want to just assign every file a UUID in the cache (rather than using real filenames or nested directories) and you'd want to empty the cache on startup.  This would be a bit wasteful of cached resources, but maybe it would be easier to write.

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/determined-ai/determined/master/pkg/archive"
+	"github.com/determined-ai/determined/proto/pkg/experimentv1"
+)
+
+type modelDefFolder struct {
```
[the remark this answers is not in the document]
gt2345: The source of model definitions are stored at `experiments` table, if we calculate and store file tree in memory when loading from `experiments` table, we can at least make sure file tree is always accurate (given that something can happen to files on disk), and whenever we notice file system differs from file tree in memory, (like file missing/renamed, not including file content being changed), we can reload from source (`experiments` table)

master/internal/db/cache.go
gt2345: When we call `GetModelDefCache` for the first time, I did `os.RemoveAll(rootDir)` so that file system and memory can sync up, but it's true that after initialization, if something tempers with the folder, `prune` does not take into account that.

master/internal/db/cache.go
gt2345: I tried to test this with symbolic links like `ln -s const.yaml link.yaml` then `det e create const.yaml .`, it doesn't seem to make any difference, `link.yaml` seems to be just a regular plain text file

master/internal/db/cache.go
gt2345: Thank you for catching this, will be fixed

proto/src/determined/api/v1/experiment.proto
gt2345: grpc gateway treats them as json string.

I was under the impression that echo version is only for downloading the file, I'm not sure if it will be the same thing for front end, I will look into it.

master/cmd/determined-master/root.go
gt2345: Makes sense, fixed

master/internal/core_experiment.go
gt2345: I’m not sure how to pass non-json object through proto… but we might be able to trigger download from front end with json object, but for now, can we keep it in echo server

e2e_tests/tests/experiment/test_model_def.py
gt2345: Yes, an integration test has been added at `master/internal/db/cache_intg_test.go`

proto/src/determined/api/v1/experiment.proto
rb-determined-ai: If it treat it as a json string, this will fail for binary files, because json cannot encode arbitrary binary data.

master/internal/db/cache.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, file := range fileTree {
+		if file.Path == path {
+			folder.lock.RLock()
+			defer folder.lock.RUnlock()
+			file, err := os.ReadFile(f.genPath(expID, path))
+			if err != nil {
+				_, ok := err.(*fs.PathError)
+				if ok {
+					// This means memory and file system are out of sync.
+					err = os.RemoveAll(f.rootDir)
```
rb-determined-ai, line 163: this is unexpected enough, we should at least be logging it.

master/internal/db/cache.go
gt2345: Yes absolutely, I should have done that

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -270,14 +270,14 @@ def test_agent_reconnect_keep_cmd(managed_cluster: ManagedCluster) -> None:
     managed_cluster.ensure_agent_ok()
 
     try:
-        command_id = run_command(30, slots=0)
```
rb-determined-ai, line 273: why did this get changed?

gt2345: Thank you for catching this, must have happened when I tried to revert `e2e_tests`

proto/src/determined/api/v1/experiment.proto
gt2345: I try to test it with pdf file and the api does not seem to break. Is this because binary data has already been encoded? I imagine that frontend can decode it when needed, but I can definitely make changes if new requests are risen during integration. After all, I don't think we support non-ascii files for viewing. 
Screen Shot 2022-07-05 at 3 26 25 PM

master/internal/db/cache.go
```diff
@@ -0,0 +1,187 @@
...
+		maxAge:  maxAge,
+		caches:  make(map[int]*modelDefFolder),
+	}
+}
+
+func (f *FileCache) getOrCreateFolder(expID int) (*modelDefFolder, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+	value, ok := f.caches[expID]
+	if !ok {
```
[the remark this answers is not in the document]
gt2345: This has been fixed

master/internal/db/cache.go
gt2345: I don't think `gzip` or `tar` deal with file system directly (just `io.Reader/Writer`), so I'm trying to find is there a way to write to disk from `io.Reader` directly, so we can skip parsing files into `Item`. For now I don't see a clear way how to do it, but am I on the right track?

master/internal/db/cache.go
```diff
@@ -0,0 +1,196 @@
...
+	maxAge  time.Duration
+	caches  map[int]*modelDefFolder
+	lock    sync.Mutex
+}
+
+// NewFileCache initialize FileCache obj.
+func NewFileCache(rootDir string, maxAge time.Duration) *FileCache {
+	err := os.RemoveAll(rootDir)
+	if err != nil {
+		log.WithError(err).Errorf("failed to initialize model def cache at %s", rootDir)
```
rb-determined-ai, line 41: fix this error string; you didn't fail to initialize the model def cache, you just failed to clear the old contents of the model def cache.

master/internal/db/cache.go
```diff
@@ -0,0 +1,196 @@
+package db
```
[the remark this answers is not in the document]
gt2345: The reason it is in db package is due to for integration test in Go, I want to use `requireMockExperiment`, if I move it to another package, I no longer have access to db helper functions...

master/internal/core_experiment.go
```diff
@@ -88,6 +88,29 @@ func (m *Master) getExperimentCheckpointsToGC(c echo.Context) (interface{}, erro
...
```
ioga, line 91: since this is a new echo endpoint that is not going to be in grpc-gateway, and it doesn't produce automated docs, could we please add documentation using the method from PR #2529 here?

master/internal/db/cache.go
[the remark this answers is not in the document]
ioga: you can make it public. e.g. `RequireMockTask` used to be private `requireMockTask`, but now it's reused in some other packages.

master/internal/db/cache.go
```diff
@@ -0,0 +1,196 @@
...
+	return filepath.Join(f.rootDir, strconv.Itoa(expID), path)
+}
+
+func (f *FileCache) getFileTree(expID int) ([]*experimentv1.FileNode, error) {
+	folder, err := f.getOrCreateFolder(expID)
+	if err != nil {
+		return nil, err
+	}
+	folder.lock.RLock()
+	defer folder.lock.RUnlock()
```
ioga, line 126: what's the purpose of this locking?

master/internal/db/cache.go
```diff
@@ -0,0 +1,196 @@
...
+		return []byte{}, err
+	}
+	fileTree, err := f.getFileTree(expID)
+	if err != nil {
+		return []byte{}, err
+	}
+	for _, file := range fileTree {
+		if file.Path == path {
+			folder.lock.RLock()
+			defer folder.lock.RUnlock()
```
ioga, line 157: this is acquiring a read lock, but in line 165 it'd try to `RemoveAll` which is a write operation.

master/internal/db/cache.go
```diff
@@ -0,0 +1,196 @@
...
+			file, err := os.ReadFile(f.genPath(expID, path))
+			if err != nil {
+				_, ok := err.(*fs.PathError)
+				if ok {
+					log.Debugf(`File system cache (%s) is likely out of sync. 
+					File system cache is about to re-initialize and this message should not appear again.`,
+						f.rootDir)
+					err = os.RemoveAll(f.rootDir)
```
ioga, line 165: when does this happen?

master/internal/db/cache.go
gt2345: Log has been added

master/internal/db/cache.go
gt2345: This has been fixed

master/internal/db/cache.go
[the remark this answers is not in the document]
gt2345: Makes sense, thank you for pointing it out

master/internal/cache.go
```diff
@@ -0,0 +1,29 @@
...
+
+var modelDefCache *db.FileCache
+
+const cacheDir = "exp_model_def"
+
+// GetModelDefCache returns FileCache object.
+func GetModelDefCache() *db.FileCache {
+	config := config.GetMasterConfig()
+	if modelDefCache == nil {
+		rootDir := filepath.Join(config.Cache.CacheDir, cacheDir)
```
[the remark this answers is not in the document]
gt2345: This has been fixed

master/internal/db/cache.go
gt2345: Concurrence may happen when `GetFileContent` is triggered with cache being re-initialized

master/internal/db/cache.go
gt2345: Thank you for catching this, this has been fixed

master/internal/db/cache.go
gt2345: One circumstance this may happen is when files in cache folder get deleted or renamed

master/internal/core_experiment.go
gt2345: Document has been added

master/internal/db/cache.go
ioga: Taking a closer look, the write operation here clears out *the entire cache*, not just this single folder.
So maybe
- in this method, we need to acquire `f.lock` but only when we get the error and need to do the delete.
- then `folder.lock` becomes useless and can be removed.

gt2345: You are right, thank you for catching this. Using `f.lock` will definitely does the job, but another way to do it is to just erase the individual folder, so that when `ReadFile` is triggered for one experiment, read regarding other experiments would not be affected.

proto/src/determined/api/v1/experiment.proto
rb-determined-ai: Cool, it looks like `bytes`-type objects are getting base64-encoded, which does make it json-safe.

For what it's worth, I verified with a junk file I wrote like this: 
```bash
python -c 'import os; os.write(1, bytes(range(256)))' > junk
```
which is a more reliable way to have invalid utf8.

master/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	for _, file := range fileTree {
+		if file.Path == path {
+			folder.lock.Lock()
+			defer folder.lock.Unlock()
+			file, err := os.ReadFile(f.genPath(expID, path))
+			if err != nil {
+				_, ok := err.(*fs.PathError)
+				if ok {
+					log.Debugf(`File system cache (%s) is likely out of sync.
```
rb-determined-ai, line 154: Please log at error level.  If this shows up in a user's master log, I want it to be obvious that something is going wrong.

master/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	folder.lock.RLock()
+	defer folder.lock.RUnlock()
+	return folder.fileTree, folder, nil
+}
+
+// GetFileTreeNested returns folder tree structure with given experiment id.
+func (f *FileCache) GetFileTreeNested(expID int) ([]*experimentv1.FileNode, error) {
```
rb-determined-ai, line 132: This is just a nit, but I'm pretty sure that in a go interface, you're [not supposed](https://go.dev/doc/effective_go#Getters) to include the word "get" in a getter's name.  So it might be worth considering removing this.

master/internal/db/cache.go
rb-determined-ai: I see.  You are right, we aren't actually preserving symbolic links at upload time as I thought we were.

No need to worry about this for now I guess.

master/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	err = os.MkdirAll(f.genPath(expID, ""), 0700)
+	if err != nil {
+		return nil, err
+	}
+	for _, ar := range arc {
+		if ar.IsDir() {
+			err = os.MkdirAll(f.genPath(expID, ar.Path), 0700)
+		} else {
+			err = os.WriteFile(f.genPath(expID, ar.Path), ar.Content, 0600)
```
rb-determined-ai, line 80: You have to be _really_ careful about unpacking tarballs, because anybody can craft a tarball with arbitrary paths.  So with the implementation you have here, somebody could have a tarball with a path like `../../../etc/shadow`, which they could use to override the root password, then they could become root on the machine.  If the determined service is running as root (right now it does) then that would be trivially possible.

So you either need to check the resulting path first, to ensure that it falls inside `$CACHE/exp_model_def/$EXP_ID`, or you need to avoid writing to user-defined paths at all.

If you wanted to take the second route, your cache directory could be a flat list of uuid-named files, where each uuid corresponds to a single file you've unpacked, and in memory you keep a mapping of paths you've unpacked and the uuid file that it corresponds to.  This would eliminate some of the complexity and danger of your code, but it would make the resulting system a little less nice to deal with.  For instance, if the cache is written as untarred files, like you have now, then any system admin can look though that directory, and our system seems a little less mysterious.

So because of that, I don't care which way you do it.

master/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/determined-ai/determined/master/internal/db"
+	"github.com/determined-ai/determined/master/pkg/archive"
+	"github.com/determined-ai/determined/proto/pkg/experimentv1"
+)
+
+type modelDefFolder struct {
+	fileTree   []*experimentv1.FileNode
```
rb-determined-ai, line 24: This struct seems like it should be a map of strings to FileNodes.  I think `insertToTree` would be a lot more performant that way.

master/internal/core.go
```diff
@@ -985,6 +985,7 @@ func (m *Master) Run(ctx context.Context) error {
 
 	experimentsGroup := m.echo.Group("/experiments", authFuncs...)
 	experimentsGroup.GET("/:experiment_id/model_def", m.getExperimentModelDefinition)
+	experimentsGroup.GET("/:experiment_id/file/download", m.getExperimentModelFile)
```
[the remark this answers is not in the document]
gt2345: `file/download` returns file as an attachment by setting `Content-Disposition`, while `file/` just return encoded file content as JSON

master/internal/cache/cache.go
gt2345: Makes sense, this is fixed

master/internal/cache/cache.go
gt2345: Sure, thank you for sharing, this has been fixed

master/internal/config/config.go
```diff
@@ -45,6 +45,12 @@ func DefaultDBConfig() *DBConfig {
 	}
 }
 
+// CacheConfig is the configuration for file cache.
+type CacheConfig struct {
+	CacheDir string `json:"cache_dir"`
+	MaxAge   string `json:"max_age"`
```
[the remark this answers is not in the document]
gt2345: I agree, `max_age` has been removed from master config

master/internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+				if ok {
+					log.Debugf(`File system cache (%s) is likely out of sync. 
+File system cache is about to re-initialize and this message should not appear again.`,
+						f.rootDir)
+					delete(f.caches, expID)
+					err = os.RemoveAll(f.genPath(expID, ""))
+					if err != nil {
+						return []byte{}, err
+					}
+					return f.GetFileContent(expID, path)
```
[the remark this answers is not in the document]
gt2345: Thank you for bringing this up, I believe what happened was before recursing, the folder gets deleted from the map along with it's lock, so next time `folder.lock` is called, it is for a new instance of `folder`.
You are right about potentially recursing forever, this has been fixed.

master/internal/cache/cache.go
gt2345: Thank you for catching this, validation for user-defined path has been added

master/internal/cache/cache.go
gt2345: Using map would definitely help with random accessing, but here I think it makes sense to persist the order

rb-determined-ai: Why is that?  Files in directories do not have a natural order.

master/internal/core.go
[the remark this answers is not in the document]
rb-determined-ai: ok

master/internal/cache/cache.go
rb-determined-ai: ping

gt2345: From my point of view, tar file decodes to a list, and the API response is also a list, so it's easier to use list in between. Also I was under the impression that when tar file decodes, it persist some sort of order.

rb-determined-ai: yes, but I see no real value in the API matching the list order of the tarball.  This code is pretty high algorithmic complexity for no real benefit.

Also I would argue that the protobuf should use a map anyway.

gt2345: I think protobuf should stay using list because map is not compatible with the frontend file tree display frame work we use, otherwise frontend will have to somehow transfer the response to a list

gt2345: Not sure if this is related, but persist some sort of order is important because when user refresh page, we don't want folders/files changing their order

rb-determined-ai: > not compatible with the frontend file tree display frame work we use
Is it our code that does the display?  If so, that would be a fine change IMO.

> By "high algorithmic complexity", did you mean what was happening in insertToTree iteration?

Yes.

I think an intermediate solution would be to use a map when building it, but convert it to the list at the end.

Or maybe it's fine as-is.  Probably what I am proposing is too complex.

rb-determined-ai (approved):
great work, especially with the huge ask of "please include our first-ever filesystem cache in-scope".

I'd like to let @ioga approve this as well, since I'm not technically an owner on this part of the system.

ioga (commented):
I think there's 2 concurrency issues

ioga (approved):
lgtm 👍

PR #4187, opened 2022-05-20
chore: remove legacy code/docs for NCCL/Gloo port range config

neilconway:
The port ranges used by NCCL and Gloo were previously configurable, but support
for this feature was removed quite a while ago. This commit removes the vestiges
of the previous feature, including docs and config parsing code. (We continue to
allow `nccl_port_range` and `gloo_port_range` to be specified in `master.yaml`
but those values are now ignored.)

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4376, opened 2022-06-16
chore: Merge Slurm-infrastructure support changes

jerryharrow:
## Description

One of many small changes coming to make the diff more manageable.

## Checklist


harness/determined/launch/deepspeed.py
```diff
@@ -7,20 +7,74 @@
 import logging
```
stoksc, line 7: I think everything in this file is good, already (so we can revert this)

master/pkg/archive/archive.go
```diff
@@ -80,6 +94,17 @@ func RootItem(path string, content []byte, mode int, fileType byte) Item {
 
 // UserItem returns a new Item which will be owned by the user under which the container runs.
 func UserItem(path string, content []byte, mode int, fileType byte, userID int, groupID int) Item {
+	if fileType == tar.TypeSymlink {
```
stoksc, line 97: just for consistency, can you make `RootItem` above behave the same? or, maybe better, remove "Linkname" from "Item" and do the mapping with the rest of the logic in "func tarArchive"

harness/determined/launch/horovod.py
```diff
@@ -19,7 +19,9 @@
 from determined.constants import DTRAIN_SSH_PORT
 
 
-def create_sshd_worker_cmd(allocation_id: str, num_slot_ids: int) -> Tuple[List[str], List[str]]:
+def create_sshd_worker_cmd(
```
stoksc, line 22: @rb-determined-ai can you just look at this file instead of my pr. @jerryharrow wrote it so I'll let him pr it.

master/pkg/tasks/task.go
```diff
@@ -119,7 +120,17 @@ func (t TaskSpec) EnvVars() map[string]string {
...
+	}
+	if t.Owner != nil {
+		e["DET_USER"] = t.Owner.Username
+	}
+
+	if t.TaskContainerDefaults.NCCLPortRange != "" {
+		e["NCCL_PORT_RANGE"] = t.TaskContainerDefaults.NCCLPortRange
+	}
+	if t.TaskContainerDefaults.NCCLPortRange != "" {
+		e["GLOO_PORT_RANGE"] = t.TaskContainerDefaults.NCCLPortRange
```
stoksc, line 133:
```suggestion
		e["GLOO_PORT_RANGE"] = t.TaskContainerDefaults.GlooPortRange
```

master/pkg/tasks/task.go
```diff
@@ -139,6 +150,15 @@ func (t TaskSpec) EnvVars() map[string]string {
 		e["DET_SEGMENT_API_KEY"] = t.SegmentAPIKey
 	}
 
+	if t.LoggingFields != nil {
+		j, err := json.Marshal(t.LoggingFields)
+		if err != nil {
+			// TODO(Dispatcher): propagate errors.
```
stoksc, line 156:
```suggestion
			// TODO(DET-7565): Propagate errors.
```

stoksc: made a ticket

master/pkg/tasks/task.go
```diff
@@ -261,7 +287,10 @@ func injectUserArchive(aug *model.AgentUserGroup, workDir string) cproto.RunArch
 	passwdBytes := []byte(
 		fmt.Sprintf("%v:x:%v:%v::%v:/bin/bash\n", aug.User, aug.UID, aug.GID, workDir),
 	)
-	shadowBytes := []byte(fmt.Sprintf("%v:!!:::::::\n", aug.User))
+	// Add a bogus password in the shadow file.  Cannot use ! as that locks the account
+	// when using SLURM/Singularity.   The password is not actually ever used so OK that
+	// it can never be specified due to our sshd_config is configured with "PasswordAuthentication no"
+	shadowBytes := []byte(fmt.Sprintf("%v:BogusPasswordHash:::::::\n", aug.User))
```
stoksc, line 293: maybe @rb-determined-ai should review these lines, since he wrote our libnss stuff.

master/static/srv/entrypoint.sh
```diff
@@ -1,5 +1,29 @@
 #!/bin/bash
 
+# When the task container is invoked via SLURM, we have
+# to set the slot IDs.
+if [ "$DET_RESOURCES_TYPE" == "slurm-job" ]; then
+    export DET_SLOT_IDS="[${CUDA_VISIBLE_DEVICES}]"
```
stoksc, line 6: maybe drop a comment explaining that slurm sets `CUDA_VISIBLE_DEVICES`

stoksc: also, we probably need this in every entrypoint, so maybe pull it into a script and use that?

master/static/srv/shell-entrypoint.sh
```diff
@@ -44,8 +44,19 @@ test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
 # After openssh 8+ is the only version of openssh supported (that is, after we
 # only support ubuntu >= 20.04), we can use the more obvious SetEnv option and
 # skip this awkwardness.
-blacklist="^(_|HOME|TERM|LANG|LC_.*)"
-vars="$(env | sed -E -e "s/=.*//; /$blacklist/d")"
+#
+# For HPC systems, bash module support uses variables that store functions 
+# of the form, and a wierd %% defintion.
```
stoksc, line 49:
```suggestion
# of the form, and a weird %% definition.
```
and some other spelling/grammar stuff in the comment

master/static/srv/task-logging-setup.sh
```diff
@@ -51,6 +51,33 @@ if [ -n "$DET_K8S_LOG_TO_FILE" ]; then
...
+    if [ -z "$DET_SKIP_PIP_INSTALL" ]; then
+        "$DET_PYTHON_EXECUTABLE" -m pip install -q --user /opt/determined/wheels/determined*.whl
+    else
+        if ! "$DET_PYTHON_EXECUTABLE" -c "import determined" >/dev/null 2>&1 ; then
+            echo "{\"log\": \"error: unable run without determined package\n\", \"timestamp\": \"$(date --rfc-3339=seconds)\"}" >&2
+            exit 1
+        fi
+    fi
+
+    exec 1> >(tee -p >("$DET_PYTHON_EXECUTABLE" /run/determined/enrich_task_logs.py --stdtype stdout) >&1; : >$DET_LOG_WAIT_FIFO) \
```
stoksc, line 75:
```suggestion
    exec 1> >(tee -p >("$DET_PYTHON_EXECUTABLE" /run/determined/enrich_task_logs.py --stdtype stdout) >&1; printf x  >$DET_LOG_WAIT_FIFO) \
```
here and on the other, since how we did this changed out from under us.

stoksc (commented):
seems like `enrich_task_logs.py` didn't actually make it here

harness/determined/launch/deepspeed.py
jerryharrow: I did figure that out earlier and attempted to revert, but apparently not :-)

master/pkg/tasks/task.go
jerryharrow: Of course there is no such thing, which makes that a non-trivial exercise.   
I

jerryharrow:
I dropped the code that has to be re-worked from this review as fixes need to be made/tested on the dispatcher branch before I can move them here.

jerryharrow:
> seems like `enrich_task_logs.py` didn't actually make it here

The Pods unit tests were failing now that we added enrich_task_logs.py into the helper archive (and it doesn't exist in this tree).   I was just creating a stub for that, but I realize now what is happening.    The same test was failing in the same way in the dispatcher tree, but today it started succeeding again.   I will bring the full version of enrich_task_logs.py  and drop in in the appropriate place here -- unless we should instead drop it from the run archive (and put it in via the dispatcher-task.go configuration -- like other slurm-specific files)?

master/static/srv/entrypoint.sh
jerryharrow: Addressing on dispatcher branch.

master/pkg/tasks/task.go
jerryharrow: Updated the comment.

master/pkg/archive/archive.go
jerryharrow: Will address on dispatcher branch where I can test it, before merging.

master/static/srv/task-logging-setup.sh
stoksc: I think you can just fix here and I can fix on rebase

master/pkg/tasks/task.go
stoksc: should be? https://github.com/determined-ai/determined/blob/7b3fee44c0846f3546f7e4bf57b231d227dcd27d/master/pkg/model/task_container_defaults.go#L23

jerryharrow: Ohh.  It is GLOOPortRange, not GlooPortRange -- grep didn't find what you suggested.

stoksc:
> unless we should instead drop it from the run archive (and put it in via the dispatcher-task.go configuration -- like other slurm-specific files)?

this is probably better, but I think either is fine.

harness/determined/launch/horovod.py
```diff
@@ -105,6 +109,9 @@ def main(hvd_args: List[str], script: List[str], autohorovod: bool) -> int:
     experiment_config = info.trial._config
 
     debug = experiment_config.get("debug", False)
+    if debug:
+        print("DEBUG: HOROVOD initilizing", file=sys.stderr)
```
rb-determined-ai, line 113: stray print

harness/determined/launch/horovod.py
```diff
@@ -179,6 +188,18 @@ def main(hvd_args: List[str], script: List[str], autohorovod: bool) -> int:
...
+    # running in a singularity container and SLURM may or may not have
+    # compatible configuration enabled.  We therefore clear the SLURM_JOBID variable
+    # before invoking mpi so that mpirun will honor the args passed via horvod
+    # run to it describing the hosts and process topology, otherwise mpi ends
+    # up wanting to launch all -np# processes on the local causing an oversubscription
+    # error ("There are not enough slots available in the system").
+    slurm_vars_to_clear = ["SLURM_JOBID"]
+    for k, _ in os.environ.items():
+        if k in slurm_vars_to_clear:
+            os.environ.pop(k)
```
rb-determined-ai, line 202: Maybe just a plain:
```python
os.environ.pop("SLURM_JOBID", None)
```
and we can make a loop out of it when we need it to be a loop.

master/pkg/tasks/task.go
rb-determined-ai: Cat we use `*` instead of `BogusPasswordHash`?

I see from `man 5 shadow` that `*` or `!` would both cause password logins to fail, though `!` has the special meaning that the account is locked (which I suppose something in the slurm/singularity stack is looking at).

jerryharrow: Right.  !  was completely blocked.   I can try * instead if you prefer, but gonna delay this til next week, since I can't test on this branch.   I can only test this support on the determined-ee dispatcher branch at the moment.

jerryharrow: It wasn't actually that hard to test -- so pushed up the change to use * instead.

stoksc (approved):
go side lgtm

stoksc:
I'll wait on Ryan for his bits

master/pkg/tasks/task.go
```diff
@@ -261,7 +287,10 @@ func injectUserArchive(aug *model.AgentUserGroup, workDir string) cproto.RunArch
 	passwdBytes := []byte(
 		fmt.Sprintf("%v:x:%v:%v::%v:/bin/bash\n", aug.User, aug.UID, aug.GID, workDir),
 	)
-	shadowBytes := []byte(fmt.Sprintf("%v:!!:::::::\n", aug.User))
+	// Add a bogus password in the shadow file.  Cannot use ! as that locks the account
+	// when using SLURM/Singularity.   Use * which cannot match any password.
+	// Our sshd_config is configured with "PasswordAuthentication no" so * works.
```
rb-determined-ai, line 292: is this line still necessary?  Or is it from the old version?

I don't understand why PasswordAuthentication makes a difference as to whether or not `*` works.

master/static/srv/enrich_task_logs.py
```diff
@@ -0,0 +1,95 @@
...
+def main(
+    master_url: str,
+    task_logging_metadata: Dict[str, Any],
+) -> None:
+    buffer = []
+    # Maximum number of log messages to be send to determined master each time.
+    max_buffer_length = 1024
+    start_time = time.time()
+    # Maximum time in seconds the loop executes, before sending log messages to determined master.
+    frequency = 5
```
rb-determined-ai, line 32: A 5-second buffer seems pretty high.  Certainly watching trial logs would be a jarring experience.  Is .5 sec too short?

master/static/srv/enrich_task_logs.py
```diff
@@ -0,0 +1,95 @@
...
+
+    for line in sys.stdin:
```
rb-determined-ai, line 34: This is always going to block, even long after the 5-second buffer timeout has passed.

You should either use selectors, or use two threads with a queue in between, where the queue has a timeout while waiting for new lines.

master/static/srv/enrich_task_logs.py
jerryharrow: I'm just merging this code, it is only used by Slurm.    You should take it up with @stoksc because he directed the implementation :-)

master/pkg/tasks/task.go
jerryharrow: Updated the comment.

master/static/srv/enrich_task_logs.py
jerryharrow: For @stoksc  as well

master/static/srv/enrich_task_logs.py
stoksc: haha. I mean, for what it is worth I nearly insisted we work off https://github.com/determined-ai/determined-ee/pull/185/files but we took this approach instead.

stoksc: @rb-determined-ai can we fly with this for now, I'll try to land my batching over the next day?

rb-determined-ai: sure I guess.  Still working on one more view comment though.

master/static/srv/shell-entrypoint.sh
```diff
@@ -44,8 +44,19 @@ test -f "${STARTUP_HOOK}" && source "${STARTUP_HOOK}"
...
+#   BASH_FUNC_ml()=() {  eval $($LMOD_DIR/ml_cmd "$@")
+#   BASH_FUNC_module%%=() {  eval `/opt/lib/modulecmd bash $*`
+# so we add variables with parens or % in the name to the blacklist and 
+# filter them out because the additional parens/% in the variable names 
+# causes problems in the escaping code below.   Additionally, we properly
+# enumerate just the names by adding the grep '^\S+=\S' below to remove 
+# the multi-line values from such functions so the blacklist can be
+# properly applied.
+blacklist="^(_|HOME|TERM|LANG|\S+\(\).*|\S+\%\S*|LC_.*)"
+vars="$(env | grep -E '^\S+=\S' | sed -E -e "s/=.*//; /$blacklist/d")"
```
rb-determined-ai, line 59: I'm thinking that we've hit the limit of this blacklist/sed strategy for parsing environment variables.  For example, we are acknowledging that we are dealing with multiline variables, but the grep command is assuming the multiline variables never match `^\S+=\S`.

How about we use `env -0` to get NUL-byte delimited environment variables, then use `mapfile -d ''` to read them into a bash array, then we iterate through the bash array:

```bash
# extglob enables +() notation in patterns of ${parameter/pattern/string} notation
shopt -s extglob

# convert NUL-delimited environment key-value pairs in an array
# -d '' means "use NUL byte as delimiter"
# -t means "strip the delimiter"
mapfile -d '' -t kvps < <(env -0)

# iterate through each key-value pair in the array
options="$(
    for kvp in "${kvps[@]}" ; do
        # Variable name is what comes before the first '='
        var="${kvp/=*/}"
        # Variable content starts after the first '='
        val="${kvp/#+([^=])=/}"

        # Filter names we shouldn't forward
        if [[ "$var" =~ ^(_|HOME|TERM|LANG|LC_.*)$ ]] ; then
            continue
        fi

        # For slurm: filter variables with %% or () in the name
        if [[ "$var" =~ (%%|\(\)) ]] ; then
            continue
        fi

        # Backslash-escape quotes so that sshd works.
        val="${val//\"/\\\"}"
        # Backslash-escape backslashes so that sed doesn't interpret them.
        val="${val//\\/\\\\}"
        # Backslash-escape forward slashes so that sed works.
        val="${val////\\/}"
        echo -n "environment=\"$var=$val\","
    done
)"
```

I tested with:
```
env -i \
 LC_ALL=1 \
 ASDF=A=B=C \
 "$(echo -e 'newline=before\nafter')" \
 "PERCENTS%%=ZXCV" \
 "PARENS()=ZXCV" \
 bash test.sh
```

to confirm that equals and newlines in variables are preserved, but LC_* is still ignored, as are %% and () if they appear.

master/static/srv/enrich_task_logs.py
stoksc: @rb-determined-ai https://github.com/determined-ai/determined-ee/pull/305/files I'll bring it back over once this lands

master/static/srv/shell-entrypoint.sh
jerryharrow: Done

master/pkg/tasks/task.go
```diff
@@ -119,7 +120,17 @@ func (t TaskSpec) EnvVars() map[string]string {
...
-		"DET_USER":          t.Owner.Username,
+		"DET_WORKDIR":       t.WorkDir,
+	}
+	if t.Owner != nil {
+		e["DET_USER"] = t.Owner.Username
+	}
+
+	if t.TaskContainerDefaults.NCCLPortRange != "" {
+		e["NCCL_PORT_RANGE"] = t.TaskContainerDefaults.NCCLPortRange
+	}
```
rb-determined-ai, line 131: can we just drop this?

We once had a fork of NCCL that supported setting port ranges, but we no longer maintain that fork or ship it to users

It's basically an ignored config field now.

(the GLOO_PORT_RANGE is built into gloo, so it's still valid).

jerryharrow: This allows users to override the GLOO port range if they are hitting problems.
With Slurm+Singularity the ports are global on the compute nodes for all experiments.   If more than one experiment is running on the same node there can be conflicts.

jerryharrow: So I'm not sure what's up here.      

chore: remove legacy code/docs for NCCL/Gloo port range config (#4187)

Which removed it, but @stoksc said earlier in this review to change it from:

`e["GLOO_PORT_RANGE"] = t.TaskContainerDefaults.NCCLPortRange`

to 

`e["GLOO_PORT_RANGE"] = t.TaskContainerDefaults.GLOOPortRange`

I guess I can just drop it, unless  @stoksc was really trying to put it back.

stoksc: Let's just leave it for now and I can PR to remove the config entirely from the master shortly?

rb-determined-ai: I think removing it from the config is technically a breaking feature, which is why it was still allowed in the config even after #4187.  

But `NCCL_PORT_RANGE` is an environment variable that [I personally added to our nccl fork three years ago](https://github.com/determined-ai/nccl/commit/8768cb3f881ce198825294b62a435e3f3a804baf).  That fork is no longer used anywhere, and upstream nccl doesn't use `NCCL_PORT_RANGE`, so setting `NCCL_PORT_RANGE` will have no effect (except maybe confusing a user somewhere).

I'm not talking about removing the GLOO_PORT_RANGE environment variable, only the NCCL_PORT_RANGE.

rb-determined-ai (approved):
great work!

master/static/srv/enrich_task_logs.py
```diff
@@ -0,0 +1,183 @@
...
+                        ).isoformat(),
+                        "log": line + "\n",
+                        **self.task_logging_metadata,
+                        **parsed_metadata,
+                    }
+                )
+            except Exception as e:
+                print(f"fatal error collecting log {e}", file=sys.stderr)
+
+        self.ship_queue.put(ShutdownMessage())
```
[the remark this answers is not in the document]
stoksc: pls and ty

master/static/srv/enrich_task_logs.py
```diff
@@ -0,0 +1,185 @@
...
+        task_logging_metadata: Dict[str, Any],
+    ):
+        self.ship_queue = ship_queue
+        self.task_logging_metadata = task_logging_metadata
+        super().__init__()
+
+    def run(self) -> None:
+        try:
+            for line in sys.stdin:
+                print(line, flush=True)
```
rb-determined-ai, line 62: it occurs to me, the slurm logs will have the extra newlines.  Can you put `end=""` in here?

And, can you maybe leave a comment explaining that we are intentionally preserving stdin on stderr so that slurm collects the logs, in addition to shipping them to determined?

jerryharrow: Correct.   Will fix.

rb-determined-ai (approved):
looks good

PR #4381, opened 2022-06-17
fix: reject reconnecting agents with different device configuration  [DET-7568]

ioga:
## Description

When master/agent task restoration is enabled (i.e. `agent_reattach_enabled: true`), and devices change on reconnect, discard the corresponding agent state and tell the agent to shutdown (and restart) as a way to not overly complicate the restore process and keep sanity. 
We use device count and their UUIDs to match the devices. UUIDs for CPUs, CUDA, and ROCM should be stable. In this change, we also introduce stability for artificial slot UUIDs, using agent id as a seed.

## Test Plan

- Automated test added. 
- Normal task restoration should continue working the same for existing setups.

For manually testing the device change handling:
1. ensure `agent_reattach_enabled: true` in master config
2.a for CUDA agent, change `NVIDIA_VISIBLE_DEVICES` on the agent container on restart so it sees different number of cards
2.b for artificial slots agent, change the number of artificial slots exposed.
3. Have agent reconnect back to master
4. Agent state will be wiped and it will be told it needs to restart
5. After another reconnection agent and master should work ok

## Commentary (optional)

In the future it'd be great to handle this case internally, without having agent restart and reconnect again.

## Checklist

### DIFF
--- a/agent/internal/detect.go
+++ b/agent/internal/detect.go
@@ -28,17 +28,29 @@ const (
 	osDarwin = "darwin"
 )
 
+// randFromString returns a random-number generated seeded from an input string.
+func randFromString(seed string) (*rand.Rand, error) {
+	h := sha256.New()
+	h.Write([]byte(seed))
+	rndSource, bytesRead := binary.Varint(h.Sum(nil)[:8])
+	if bytesRead <= 0 {
+		return nil, fmt.Errorf(
+			"failed to init random source for artificial slots ids. bytes read: %d", bytesRead)
+	}
+	rnd := rand.New(rand.NewSource(rndSource)) // nolint:gosec
+	return rnd, nil
+}
+
 func (a *agent) detect() error {
 	switch {
 	case a.ArtificialSlots > 0:
-		h := sha256.New()
-		h.Write([]byte(a.AgentID))
-		rndSource, bytesRead := binary.Varint(h.Sum(nil)[:8])
-		if bytesRead <= 0 {
-			return fmt.Errorf(
-				"failed to init random source for artificial slots ids. bytes read: %d", bytesRead)
+		// Generate random UUIDs consistent across agent restarts as long as
+		// agentID is the same.
+		rnd, err := randFromString(a.AgentID)
+		if err != nil {
+			return err
 		}
-		rnd := rand.New(rand.NewSource(rndSource)) // nolint:gosec
+
 		for i := 0; i < a.ArtificialSlots; i++ {
 			u, err := uuid.NewRandomFromReader(rnd)
 			if err != nil {
--- a/agent/internal/detect.go
+++ b/agent/internal/detect.go
@@ -2,9 +2,12 @@ package internal
 
 import (
 	"bufio"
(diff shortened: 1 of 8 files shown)


agent/internal/detect.go
```diff
@@ -28,8 +31,20 @@ const (
 func (a *agent) detect() error {
 	switch {
 	case a.ArtificialSlots > 0:
+		h := sha256.New()
```
stoksc, line 34: neat trick.

stoksc: maybe adding a comment explaining this is to keep slot ids consistent across restarts?

stoksc: idk, or put it in a func with that name, I guess

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -418,22 +418,39 @@ func (a *agent) handleAPIRequest(ctx *actor.Context, apiCtx echo.Context) {
...
+					Message: aproto.AgentMessage{
+						AgentShutdown: &aproto.AgentShutdown{
+							ErrMsg: aproto.ErrAgentMustReconnect.Error(),
+						},
+					},
+				}
+				if err = ctx.Ask(a.socket, wsm).Error(); err != nil {
+					log.WithError(err).Error("failed to tell agent to reconnect")
+				}
+				check.Panic(err)
```
stoksc, line 442: why not just `panic(err)` in the if

master/internal/resourcemanagers/agent/agent.go
```diff
@@ -418,22 +418,39 @@ func (a *agent) handleAPIRequest(ctx *actor.Context, apiCtx echo.Context) {
...
+				ctx.Self().Stop()
```
stoksc, line 443: so weird to reason about stop not actually happening until we get around to it in the inbox

master/internal/resourcemanagers/agent/agent_state.go
```diff
@@ -255,6 +255,35 @@ func (a *AgentState) agentStarted(ctx *actor.Context, agentStarted *aproto.Agent
...
+	for did, slot := range a.slotStates {
+		ourDevices[did] = slot.device
+	}
+
+	theirDevices := map[device.ID]device.Device{}
+	for _, d := range agentStarted.Devices {
+		theirDevices[d.ID] = d
+	}
+
+	if len(ourDevices) != len(theirDevices) {
```
stoksc, line 271: this logic (maps.Equal with better messaging) is in enough places it may make sense to pull out. the testing assert package has it but wouldn't want use that here I don't think..

stoksc (approved):
lgtm

master/internal/resourcemanagers/agent/agent_state.go
ioga: that's the only place we use `maps.Equal` based on search? I imagine we may be comparing maps by hand in some code which would already benefit just from a move to `maps.Equal`. same for `maps.Keys`.

PR #4387, opened 2022-06-21
fix: always flush tensorboard ready log

stoksc:
## Description
Tensorboards weren't opening reliably because the 'isReady' bit wasn't being set. upon some investigation, the proxy is up because the container is running, and the fetcher has indeed moved the tfevents files over from shared_fs to /tmp/whatever, which should imply it has printed this line. I noticed running on `debug: True` made this never happen and assume it was some weird stream buffering thing. This patch entire removes the issue.

## Test Plan
- [x] reproduce many, many times without this. ensure adding the line makes it unreproducible.

## Commentary (optional)
If a python world person has a better way, pls save me.

## Checklist

### DIFF
--- a/harness/determined/exec/tensorboard.py
+++ b/harness/determined/exec/tensorboard.py
@@ -133,7 +133,7 @@ def start_tensorboard(
                         responsive = check_tensorboard_responsive()
 
                     if responsive and not triggered and num_fetched_files > 0:
-                        print(TENSORBOARD_TRIGGER_READY_MSG)
+                        print(TENSORBOARD_TRIGGER_READY_MSG, flush=True)
                         triggered = True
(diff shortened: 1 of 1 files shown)


rb-determined-ai (commented):
Can we just launch this task with `PYTHONUNBUFFERED=1`?  Or set it in every container we launch?

rb-determined-ai:
er... to be clear, leaving `flush=True` in that spot is still fine with me.

stoksc:
> Or set it in every container we launch?

Seems pretty broad, I wasn't sure how many implications that had. Maybe, sure, in tb/jupyter?

rb-determined-ai (approved):

PR #4391, opened 2022-06-21
fix: cluster overview [DET-7517 DET-7567 DET-7517]

hamidzr:
## Commentary (optional)

I started to fix the mentioned bugs but it was getting confusing with the old cluster modules and the duplicated code so worked on deprecating that as well. I removed the feature to remember the selected resource pool assuming that's the desired effect with the new design but if we want that I should add it back.

@alexjdetermined @gt2345 what is the expected behavior when you visit the cluster page? do we want the user to be dropped into the resource pool they last clicked on (this is the behavior I saw when I tested the new design in the bugbash) or to be dropped into the RP they were last on?

## Checklist


hamidzr:
> Elegant work, I like how you cleaned up job queue component, but probably out of scope for this project, consider removing components like `RPStatsOverview` because they are no longer used.

thanks for noticing that. I wonder if there's a way to automatically find those unused components. are there any other ones you know of?

note to self:

- [ ] release notes.

hamidzr:
> doh, silly macs. Sorry about that.

that confused me the first time as well 🤷🏼

gh-determined-ai (approved):

hamidzr:
https://determined-ai.slack.com/archives/CKQ3EV0G6/p1655835649800519

PR #4394, opened 2022-06-21
feat: persist user web setting [DET-7501]

gt2345:
## Description

Persist user setting from local storage to DB.

## Test Plan

Navigate to a page with user settings (experiment-list), inspect -> application -> local storage, remove one/some of the settings related to this page manually, refresh and the setting should come back. 
Verify that `dev resetUserPreferences` works as expected. 

## Commentary (optional)

Please note that setting not stored in local storage through `useSettings` (start with `u:`) are not persist. 
Also currently data validation is kept at front end.

## Checklist


master/internal/db/postgres_users.go
```diff
@@ -425,3 +427,32 @@ func (db *PgDB) AuthTokenKeypair() (*model.AuthTokenKeypair, error) {
...
+func UpdateUserSetting(setting *model.UserWebSetting) error {
+	if len(setting.Value) == 0 {
+		_, err := Bun().NewDelete().Model(setting).Where(
+			"user_id = ?", setting.UserID).Where(
+			"storage_path = ?", setting.StoragePath).Where(
+			"key = ?", setting.Key).Exec(context.TODO())
+		return err
+	}
+
+	_, err := Bun().NewInsert().Model(setting).On("CONFLICT (user_id, key, storage_path) DO UPDATE").
```
trentwatt, line 441: question:  i would have thought if we were doing `ON CONFLICT UPDATE` here, then we wouldn't need to do the deletion above. do we need both?

master/internal/db/postgres_users.go
```diff
@@ -425,3 +427,32 @@ func (db *PgDB) AuthTokenKeypair() (*model.AuthTokenKeypair, error) {
...
+func GetUserSetting(userID model.UserID) ([]*userv1.UserWebSetting, error) {
+	setting := []*userv1.UserWebSetting{}
+	err := Bun().NewSelect().Model(&setting).Where("user_id = ?", userID).Scan(context.TODO())
+	return setting, err
+}
+
+// ResetUserSetting resets user setting.
+func ResetUserSetting(userID model.UserID) error {
+	var setting model.UserWebSetting
+	_, err := Bun().NewDelete().Model(&setting).Where("user_id = ?", userID).Exec(context.TODO())
```
trentwatt, line 456: question: would it make sense for reset to set the settings to empty as opposed to deleting the row (don't really have a definite idea here, just was wondering)

proto/src/determined/api/v1/user.proto
```diff
@@ -66,3 +66,34 @@ message PatchUserResponse {
   // The updated user.
   determined.user.v1.User user = 1;
 }
+// Get user settings.
+message GetUserSettingRequest {
+  // The id of the user.
+  int32 user_id = 1;
```
trentwatt, line 72: if- later on- we wanted to adapt the GET API to return a particular setting as opposed to all of them for a user, would we be able to accommodate that pretty well?

webui/react/src/hooks/useSettings.ts
```diff
@@ -280,6 +297,34 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
...
+        } else {
+          storage.set(config.storageKey, jsonValue, storagePath);
+        }
+      }
+    });
+
+  }, [ configMap, storage, user ]);
+
+  useEffect(() => {
+    fetchUserSetting();
```
trentwatt, line 325: would this cause `settings` to be updated, if `settings` was defined by `const { settings } = useSettings(...)`?

webui/react/src/hooks/useSettings.ts
```diff
@@ -280,6 +297,34 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
     updateSettings(newSettings);
   }, [ config.settings, updateSettings ]);
 
+  const fetchUserSetting = useCallback(async () => {
+
+    if(!user) return;
+    const userSettingResponse = await getUserSetting({ userId: user.id });
+    userSettingResponse.settings.forEach(setting => {
```
trentwatt, line 304: it might be nice to be able to query for specific settings at some point, to avoid processing the whole thing.

webui/react/src/hooks/useSettings.ts
```diff
@@ -280,6 +297,34 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
     updateSettings(newSettings);
   }, [ config.settings, updateSettings ]);
 
+  const fetchUserSetting = useCallback(async () => {
+
+    if(!user) return;
+    const userSettingResponse = await getUserSetting({ userId: user.id });
```
trentwatt, line 303: can we get the `user` from the auth on the API side as opposed to passing it as a parameter?

trentwatt (commented):
just did a read through of the code, looks good. will do some testing tomorrow

would we want to provide an option for devs to control whether certain storagePath/certain individual settings are synced?

master/internal/db/postgres_users.go
gt2345: Undefined `Value` in body corresponds to `Storage.remove`

master/internal/db/postgres_users.go
gt2345: This was trying to align with front end storage, when we `clear` the local storage, we remove everything related, so at the backend, I wanted to do the same.

webui/react/src/hooks/useSettings.ts
gt2345: I agree, I should do that.

webui/react/src/hooks/useSettings.ts
gt2345: I agree, but I wrote like this because `basePath` is not exposed to this level, so even we can query with a path, we do not know the path outside of `Storage` component.

PR #4402, opened 2022-06-22
chore: migrate shared-web to subtree

trentwatt:
## Description

convert existing submodule at `webui/react/src/shared` to be a subtree in order for us to try it out a bit.

incorporating `shared-web` git history into `core` resulted in some hassles when switching branches, so left that out. history in actual `shared-web` repo  remains intact.

## Test Plan

`git fetch upstream`
`git checkout subtree`

switch back and forth between any branches you have locally (especially older ones) and `subtree`, check that status is clean and you don't get blocked. there may be some issues that are config and/or local-residue dependent. check if issues persist with fresh clone of repo.

### test rebasing

`git remote add hamid https://github.com/hamidzr/determined`
`git checkout share-eslint`
`git rebase subtree`
`git rebase --skip` (for the commits that only touch `shared`)
`git subtree pull --prefix webui/react/src/shared https://github.com/determined-ai/shared-web share-eslint-styles --squash`

[here is a PR reflecting the result of that process](https://github.com/hamidzr/determined/pull/11/files). there are some artifacts outside of web, but looks like that's just because existing branch is not up to date.

### test PRs

[example PR that encompasses core and shared](https://github.com/determined-ai/determined/pull/4403). Can merge it, and then rebase `share-eslint` again.

## Checklist


hamidzr (commented):
lgtm thanks for making the change. I think we should try to fix shared/bin/test.py before we land so that
1. we can partially test the changes against saas when we're making changes in core (ci won't do this here easily because  secret sharing complexity)
2. saas ci jobs won't start failing when they test that core builds fine.

hamidzr (approved):
looks good to me a couple of notes.
- should we add some details to the read me to tell people about the subtree, the commands they might want or the process? as in they should pull in changes and push those back up after to the shared web.
- might make sense to squash all the old shared-web commits into one so it's easier to follow the pr history if we're squash merging into core anyway

webui/react/src/shared/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Limitations:
+- shared dependencies between all the projects would need to have similar APIs
+- no pre-canned component-level versioning solution
+
+Read more about the methodology and tradeoffs [here](https://hpe-my.sharepoint.com/:w:/p/hamid_zare/EZh809x5395CrmypPcu8yiUBO4_Pk1cK_MdRnP3bywOLrQ?e=VdccG4)
+
+## How
+
+Code in this directory should not depend on internal dependencies that leave outside this directory
```
dannysauer, line 21:
```suggestion
Code in this directory should not depend on internal dependencies that live outside this directory
```
Typo?

.github/labeler.yml
```diff
@@ -0,0 +1,2 @@
+shared-web:
+- webui/react/src/shared/**/*
```
dannysauer, line 2:
```suggestion
- webui/react/src/shared/**
```
You only need the trailing `/*` if you want something like `*.js` in any directory. If it's "all the files in any subdirs", then /** matches.

I'm pretty sure it'll still work the way it is, but if the Readme spelling is fixed, may as well clean this up at the same time. :)

PR #4404, opened 2022-06-22
chore: Add nolintlint to deadcode for unused oss method

jerryharrow:
## Description

jobStatsByPool is only used in determined-ee and thus has a nolint:deadcode directive, however, since it is used in
determined-ee it additionally needs ,nolintlint to avoid a linter error that it is not needed.

## Checklist


stoksc (approved):
lol..

PR #4405, opened 2022-06-22
fix: Do not fail if no checkpoints to gc [FOUNDENG-83, DET-7548]

jerryharrow:
## Description

With slurm there may be no checkpoints to gc and if so we end up with a error back from the checkpoint_gc.py script since no argument is provided to the --delete.

## Test Plan

Manually tested with metric_maker on casablanca-login which hits this case, and verified we do invoke the cleanup if there are any checkpoints to be removed.

## Checklist


stoksc:
tagged another ticket since we had one on our side for this, too

PR #4406, opened 2022-06-22
fix: Import Scale from correct location

mapmeld:
## Description

Fixes an immediate issue with building webui on master branch.

## Checklist

### DIFF
--- a/webui/react/src/components/UPlot/UPlotScatter.tsx
+++ b/webui/react/src/components/UPlot/UPlotScatter.tsx
@@ -1,11 +1,11 @@
 import React, { useMemo, useRef } from 'react';
 import uPlot from 'uplot';
 
-import { Scale } from 'components/ScaleSelectFilter';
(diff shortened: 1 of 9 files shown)


webui/react/src/components/UPlot/UPlotScatter.tsx
```diff
@@ -1,11 +1,11 @@
 import React, { useMemo, useRef } from 'react';
 import uPlot from 'uplot';
 
-import { Scale } from 'components/ScaleSelectFilter';
 import UPlotChart, { Options } from 'components/UPlot/UPlotChart';
 import QuadTree, { pointWithin } from 'components/UPlot/UPlotScatter/quadtree';
 
 import { Range } from '../../shared/types';
+import { Scale } from '../../types';
```
hkang1, line 8: **suggestion:** just a FYI, the two imports above can reference from an "absolute" path given how our webpack / typescript has aliases setup. So they can the following instead:
```
import { Range } from 'shared/types';
import { Scale } from 'types';
```

PR #4407, opened 2022-06-22
feat: Connect trial UI to workloads API; pass sort/filter to API

mapmeld:
## Description

Connects the "Workloads" table on the single-trial page to the Workloads API. This code is part of a plan to remove the response of all workloads and metrics from the Trial API (#4300). We're splitting that PR up for easier review.

We already have a paginated workloads API (#3635). For the sortable table to connect to the API, this PR adds a sortKey field (default is batch #, this adds sorting by a metric) and filter field (a dropdown near the table selects All / Checkpoints / Validation / Checkpoints or Validation).

~This commit includes the patch from #4406.~ Done.

Remaining issue with same-name metrics:
- `coalesce(t.metrics, v.metrics)` is odd, can we confirm that no one workload includes both training and validation metrics? 
- When we sort by a metric ("loss"), we sort after the `coalesce(t.metrics, v.metrics)` line distinguishing training and validation metrics.

Exploring the issue:

Screen Shot 2022-06-22 at 8 12 34 AM

In this test, training-loss and validation-loss both show a sorted-column icon. The training and validation losses are displayed on the correct columns. The training workload 100 and validation workload 100 were returned as separate workloads by the API, but got combined into one table row by `workloadsToSteps`, and sorted as the highest value for that metric within that page.

latest-master also combines the 100 workloads, and sorts the row as the highest value of all workloads (i.e. DESC shows it first, ASC shows it last)
 
Screen Shot 2022-06-22 at 8 49 07 AM

## Checklist

### DIFF
--- a/master/internal/api_trials.go
+++ b/master/internal/api_trials.go
@@ -512,7 +512,7 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
 		req.TrialId,
 		nil,
 		nil,
-		"All",
+		"FILTER_OPTION_UNSPECIFIED",
 	); {
 	case err == db.ErrNotFound:
 		return nil, status.Errorf(codes.NotFound, "trial %d workloads not found:", req.TrialId)
--- a/master/internal/api_trials.go
+++ b/master/internal/api_trials.go
@@ -510,8 +510,8 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
 		},
 		&workloads,
 		req.TrialId,
-		0,
-		1000000,
+		nil,
+		nil,
 		"All",
 	); {
 	case err == db.ErrNotFound:
--- a/master/internal/api_trials.go
+++ b/master/internal/api_trials.go
@@ -503,13 +503,16 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
 	switch err := a.m.db.QueryProtof(
 		"proto_get_trial_workloads",
 		[]interface{}{
+			"total_batches",
+			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 		},
 		&workloads,
 		req.TrialId,
-		nil,
-		nil,
+		0,
+		1000000,
+		"All",
 	); {
 	case err == db.ErrNotFound:
 		return nil, status.Errorf(codes.NotFound, "trial %d workloads not found:", req.TrialId)
@@ -635,13 +638,24 @@ func (a *apiServer) GetTrialWorkloads(_ context.Context, req *apiv1.GetTrialWork
 		limit = nil
 	}
 
+	sortCode := "total_batches"
+	if req.SortKey != "" && req.SortKey != "batches" {
+		sortCode = fmt.Sprintf("metrics->>'%s'", strings.ReplaceAll(req.SortKey, "'", ""))
+	}
+
 	switch err := a.m.db.QueryProtof(
 		"proto_get_trial_workloads",
-		[]interface{}{db.OrderByToSQL(req.OrderBy), db.OrderByToSQL(req.OrderBy)},
+		[]interface{}{
+			sortCode,
+			db.OrderByToSQL(req.OrderBy),
+			db.OrderByToSQL(req.OrderBy),
+			db.OrderByToSQL(req.OrderBy),
+		},
 		resp,
 		req.TrialId,
 		req.Offset,
 		limit,
+		req.Filter,
 	); {
 	case err == db.ErrNotFound:
 		return nil, status.Errorf(codes.NotFound, "trial %d workloads not found:", req.TrialId)
--- a/master/internal/api_trials.go
+++ b/master/internal/api_trials.go
@@ -655,7 +655,7 @@ func (a *apiServer) GetTrialWorkloads(_ context.Context, req *apiv1.GetTrialWork
 		req.TrialId,
 		req.Offset,
 		limit,
-		req.Filter,
+		req.Filter.String(),
 	); {
 	case err == db.ErrNotFound:
 		return nil, status.Errorf(codes.NotFound, "trial %d workloads not found:", req.TrialId)
--- a/master/internal/db/postgres_experiments.go
+++ b/master/internal/db/postgres_experiments.go
@@ -484,7 +484,7 @@ FROM trials t
 WHERE t.id=$2
   AND s.state = 'COMPLETED'
   AND total_batches >= $3
-  AND total_batches <= $4
+  AND ($4 <= 0 OR total_batches <= $4)
   AND s.end_time > $5
   AND s.metrics->'avg_metrics'->$1 IS NOT NULL
 ORDER BY batches;`, metricName, trialID, startBatches, endBatches, startTime)
@@ -512,7 +512,7 @@ JOIN validations v ON t.id = v.trial_id
 WHERE t.id=$2
   AND v.state = 'COMPLETED'
   AND v.total_batches >= $3
-  AND v.total_batches <= $4
+  AND ($4 <= 0 OR v.total_batches <= $4)
   AND v.end_time > $5
   AND v.metrics->'validation_metrics'->$1 IS NOT NULL
 ORDER BY batches;`, metricName, trialID, startBatches, endBatches, startTime)
--- a/master/internal/db/postgres_experiments.go
+++ b/master/internal/db/postgres_experiments.go
@@ -484,7 +484,7 @@ FROM trials t
 WHERE t.id=$2
   AND s.state = 'COMPLETED'
   AND total_batches >= $3
-  AND ($4 <= 0 OR total_batches <= $4)
+  AND total_batches <= $4
   AND s.end_time > $5
   AND s.metrics->'avg_metrics'->$1 IS NOT NULL
 ORDER BY batches;`, metricName, trialID, startBatches, endBatches, startTime)
@@ -512,7 +512,7 @@ JOIN validations v ON t.id = v.trial_id
 WHERE t.id=$2
   AND v.state = 'COMPLETED'
   AND v.total_batches >= $3
-  AND ($4 <= 0 OR v.total_batches <= $4)
+  AND v.total_batches <= $4
   AND v.end_time > $5
   AND v.metrics->'validation_metrics'->$1 IS NOT NULL
 ORDER BY batches;`, metricName, trialID, startBatches, endBatches, startTime)
--- a/master/static/srv/proto_get_trial_workloads.sql
+++ b/master/static/srv/proto_get_trial_workloads.sql
@@ -1,5 +1,5 @@
 WITH validations_vt AS (
-  SELECT row_to_json(r1) AS validation, total_batches, end_time
+  SELECT row_to_json(r1) AS validation, total_batches, end_time, metrics
   FROM (
       SELECT 'STATE_' || v.state as state,
         v.end_time,
@@ -8,10 +8,11 @@ WITH validations_vt AS (
         v.metrics->'validation_metrics' as metrics
       FROM validations v
       WHERE v.trial_id = $1
+      AND $4 != 'Has Checkpoint'
     ) AS r1
 ),
 trainings_vt AS (
-  SELECT row_to_json(r1) AS training, total_batches, end_time
+  SELECT row_to_json(r1) AS training, total_batches, end_time, metrics
   FROM (
       SELECT s.end_time,
         'STATE_' || s.state as state,
@@ -20,6 +21,7 @@ trainings_vt AS (
         s.total_batches
       FROM steps s
       WHERE s.trial_id = $1
+      AND $4 = 'All'
     ) AS r1
 ),
 checkpoints_vt AS (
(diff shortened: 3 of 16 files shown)


mapmeld:
update: I think I have a good resolution on table rows and sorting when there are both training-loss and validation-loss.  There would be an issue if typically we have workloads with the same batch number and we *do* want them merged, but I don't think that's the case?

master/internal/api_trials.go
```diff
@@ -501,13 +501,16 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 			db.OrderByToSQL(apiv1.OrderBy_ORDER_BY_ASC),
 		},
 		&workloads,
 		req.TrialId,
-		nil,
-		nil,
+		0,
+		1000000,
```
trentwatt, line 512: so we are assuming trial has <= million batches? What, if anything, goes wrong if this assumption is not correct? 

maybe it's not so important because if you care about workloads you should be using the workloads endpoint?

it looks like we are still getting workloads as a part of trial API. is the plan to get rid of it eventually?

master/internal/api_trials.go
```diff
@@ -501,13 +501,16 @@ func (a *apiServer) GetTrial(_ context.Context, req *apiv1.GetTrialRequest) (
...
+		"All",
```
trentwatt, line 513: so we are assuming trial has <= million batches? What, if anything, goes wrong if this assumption is not correct?

maybe it's not so important because if you care about workloads you should be using the workloads endpoint? 

it looks like we are still getting workloads as a part of trial API. is the plan to get rid of it eventually?

proto/src/determined/api/v1/trial.proto
```diff
@@ -251,6 +251,10 @@ message GetTrialWorkloadsRequest {
   int32 limit = 3;
   // Limit workloads to those that are owned by the specified trial.
   int32 trial_id = 4;
+  // Sort workloads by this column.
+  string sort_key = 5;
+  // Filter workloads with validation and/or checkpoint information.
+  string filter = 6;
```
trentwatt, line 257: can we use an enum for this?

proto/src/determined/api/v1/trial.proto
```diff
@@ -251,6 +251,10 @@ message GetTrialWorkloadsRequest {
   int32 limit = 3;
   // Limit workloads to those that are owned by the specified trial.
   int32 trial_id = 4;
+  // Sort workloads by this column.
+  string sort_key = 5;
```
trentwatt, line 255: can we use an enum for this? or could it make things trickier later?

webui/react/src/pages/TrialDetails/TrialDetailsHeader.tsx
```diff
@@ -12,14 +12,14 @@ import { openOrCreateTensorBoard } from 'services/api';
...
-import { getWorkload, isMetricsWorkload } from 'utils/workload';
+import { isMetricsWorkload } from 'utils/workload';
 import { openCommand } from 'wait';
 
 export const trialWillNeverHaveData = (trial: TrialDetails): boolean => {
   const isTerminal = terminalRunStates.has(trial.state);
   const workloadsWithSomeMetric = trial.workloads
-    .map(getWorkload)
-    .filter(workload => isMetricsWorkload(workload) && !!workload.metrics);
+    .map(w => w.checkpoint || w.training || w.validation)
```
trentwatt, line 21: can we use `??`

master/internal/db/postgres_experiments.go
```diff
@@ -396,7 +396,7 @@ FROM trials t
 WHERE t.id=$2
   AND s.state = 'COMPLETED'
   AND total_batches >= $3
-  AND total_batches <= $4
+  AND ($4 <= 0 OR total_batches <= $4)
```
trentwatt, line 399: question: don't see a need for it, but it seems like just for reasons of consistency, `end_batches = 0` should be a valid filter. is there a reason we need this? 

is it since because we are passing `0` when no value is supplied?

master/static/srv/proto_get_trial_workloads.sql
```diff
@@ -48,7 +50,11 @@ workloads AS (
       t.end_time,
       v.end_time,
       c.end_time
-    ) AS end_time
+    ) AS end_time,
+    coalesce(
+      t.metrics,
+      v.metrics
```
trentwatt, line 56: looks like `checkpoints_view`(also in this query) has both training metrics and validation metrics together. it feels a little weird that we are separating them out and then putting them back together client side.

webui/react/src/utils/workload.ts
```diff
@@ -36,22 +30,23 @@ export const isMetricsWorkload = (
...
-    stepsDict[batchNum].batchNum = batchNum;
-
-    if (workload.checkpoint) {
-      stepsDict[batchNum].checkpoint = workload.checkpoint;
-    } else if (workload.validation) {
-      stepsDict[batchNum].validation = workload.validation;
-    } else if (workload.training) {
-      stepsDict[batchNum].training = workload.training;
+  return workloads.map(workload => {
+    let wltype = 't';
```
trentwatt, line 34: can this be based of off API enum?

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
```diff
@@ -106,8 +110,52 @@ const TrialDetailsWorkloads: React.FC<Props> = ({
...
+    trial?.id,
+    settings.sortDesc,
+    settings.sortKey,
+    settings.tableLimit,
+    settings.tableOffset,
+    settings.filter,
+  ]);
+
+  // const { stopPolling } =
+  usePolling(fetchWorkloads);
```
trentwatt, line 151: if you want to rerun the polling function when the params change, `rerunOnNewFn` will do that (for freshness of data)

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
```diff
@@ -106,8 +110,52 @@ const TrialDetailsWorkloads: React.FC<Props> = ({
...
+
+  useEffect(() => {
+    fetchWorkloads();
+  }, [ fetchWorkloads, settings.sortDesc, settings.sortKey, trial?.id ]);
```
trentwatt, line 155: rerun should allow you to get rid of this

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
```diff
@@ -106,8 +110,52 @@ const TrialDetailsWorkloads: React.FC<Props> = ({
...
+
   const workloadSteps = useMemo(() => {
-    const data = trial?.workloads || [];
+    const data = workloads || [];
```
trentwatt, line 158: can use ??

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
```diff
@@ -167,9 +215,9 @@ const TrialDetailsWorkloads: React.FC<Props> = ({
           pagination={getFullPaginationConfig({
             limit: settings.tableLimit,
             offset: settings.tableOffset,
-          }, workloadSteps.length)}
+          }, workloadCount)}
           rowClassName={defaultRowClassName({ clickable: false })}
-          rowKey="batchNum"
+          rowKey="key"
```
trentwatt, line 220: 😃

master/static/srv/proto_get_trial_workloads.sql
```diff
@@ -58,7 +64,7 @@ page_info AS (
 )
 SELECT (
   SELECT jsonb_agg(w) FROM (SELECT validation, training, checkpoint FROM workloads
-    ORDER BY total_batches %s, end_time %s
+    ORDER BY %s %s NULLS LAST, total_batches %s, end_time %s
```
trentwatt, line 67: i think maybe there could be an extra `NULLS LAST` in here somewhere? sorting by loss asc results in nulls first

image

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
```diff
@@ -106,8 +110,52 @@ const TrialDetailsWorkloads: React.FC<Props> = ({
...
+  }, [
+    trial?.id,
+    settings.sortDesc,
+    settings.sortKey,
+    settings.tableLimit,
+    settings.tableOffset,
+    settings.filter,
+  ]);
+
+  // const { stopPolling } =
```
trentwatt, line 150: stray comment

trentwatt (approved):
Looks good, main caveat is just that sorting by loss ASC results in displaying null first. may have been the case previously to but might as well address now.

hkang1 (commented):
Took a quick peak, looks like @trentwatt got a lot of coverage already and it's looking good! Added some minor things here and there.

master/internal/db/postgres_experiments.go
mapmeld: yes `end_batches = 0` should mean no filter on `end_batches` and make this line TRUE

master/internal/api_trials.go
mapmeld: I'll replace this with a '0' (i.e. no end batch)
We will later remove workloads response from this endpoint, but the frontend still expects it to show total combined checkpoint size, metric names, whether metrics have been logged, TrialsComparisonModal, and (previously, I think?) the chart information

master/static/srv/proto_get_trial_workloads.sql
mapmeld: I don't have access to the same experiment, but as validation_loss would be desc here, I'm thinking that you're sorting by another column? And that column variable exists on both training and validation? I was thinking that it was going OK to sort these together, but I forgot how likely it is that only one of the columns would be displayed in the UI. So I probably want to revisit this and make it possible to treat these differently.

master/internal/api_trials.go
mapmeld: ended up being nil for consistency ✅

webui/react/src/pages/TrialDetails/TrialDetailsWorkloads.tsx
mapmeld: ✅

trentwatt:
> update: I think I have a good resolution on table rows and sorting when there are both training-loss and validation-loss. There would be an issue if typically we have workloads with the same batch number and we _do_ want them merged, but I don't think that's the case?

what do you think we want to do here @hkang1 ?  merge or not? if no merge, we end up with a single batch having separate rows for training metrics, validation metrics, and checkpoints, at least under certain circumstances.

would require significantly changing API structure/query logic though?

image

proto/src/determined/api/v1/trial.proto
mapmeld: Added enum for filter ✅

webui/react/src/services/decoder.ts
```diff
@@ -543,6 +543,26 @@ export const decodeTrialSummary = (
...
+  data: Sdk.V1GetTrialWorkloadsResponse,
+): types.TrialWorkloads => {
+  if (data.workloads) {
+    const workloads = data.workloads.map(ww => ({
+      checkpoint: ww.checkpoint && decodeCheckpointWorkload(ww.checkpoint),
+      training: ww.training && decodeMetricsWorkload(ww.training),
+      validation: ww.validation && decodeMetricsWorkload(ww.validation),
+    }));
+    return {
+      count: data?.pagination?.total || 0,
```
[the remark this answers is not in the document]
mapmeld: ok I got pagination in there ✅

master/internal/db/postgres_experiments.go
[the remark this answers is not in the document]
mapmeld: For infinite batches we could use -1, does that seem like a better option?

master/static/srv/proto_get_trial_workloads.sql
mapmeld: ok I remember what happened here -- I don't know the specifics of `checkpoints_view` so I might be off-base on that.  

But training, validation, and checkpoints are separate events being pulled in from their respective tables (trainings_vt, checkpoints_vt, validations_vt) into one workloads view here with no overlaps, i.e. `FULL JOIN ... ON false`. COALESCE (as used here already for total_batches and end_time) is essentially UNION-ing. Maybe the coalesce is needed for the `v.validation::jsonb AS validation, t.training::jsonb AS training, c.checkpoint::jsonb AS checkpoint` line? I've followed suit with metrics here for our table and for ORDER BY.

hkang1:
> > update: I think I have a good resolution on table rows and sorting when there are both training-loss and validation-loss. There would be an issue if typically we have workloads with the same batch number and we _do_ want them merged, but I don't think that's the case?
> 
> what do you think we want to do here @hkang1 ? merge or not? if no merge, we end up with a single batch having separate rows for training metrics, validation metrics, and checkpoints, at least under certain circumstances.
> 
> would require significantly changing API structure/query logic though?
> 
> image

What's the level of effort to preserve the merged behavior?

We are not quite there yet but the goal is eventually to have a dedicated checkpoints section for each experiment, along with a dedicated data download capability. This workloads table is really for the user to understand what data is getting used to draw the chart and to find if there are problems with training (e.g. looking for exploding values of NaN, Inf or -Inf).

Basically saying if it's easy to preserve the merged behavior, let's do it, otherwise the separate rows are fine as it is a temporary state. The benefits of being able to sort and paginate metrics and checkpoints more efficiently is much needed for performance issues we've had around this that compared to losing merged behavior is acceptable IMO.

PR #4408, opened 2022-06-22
fix: allocation state mismatch with proto [DET-7593]

NicholasBlaskey:
## Description

Fix a proto and Go enum mismatch causing ```/api/v1/tensorboards``` and ```/api/v1/tasks/:id``` to be out of sync.

## Test Plan

Go integration tests. 

In addition verify this migration ran correctly by ensuring that the table ```allocations``` field state is still correct after migration.

This can be done by running a bunch of allocations on a commit before this, stopping devcluster, then rerunning devcluster on this commit and running ```SELECT state FROM allocations``` to make sure the states map correctly.

## Checklist

### DIFF
--- a/master/pkg/model/task.go
+++ b/master/pkg/model/task.go
@@ -98,7 +98,7 @@ type Allocation struct {
 }
 
 // AllocationState represents the current state of the task. Value indicates a partial ordering.
-type AllocationState int
+type AllocationState string
 
 // TaskStats is the model for task stats in the database.
 type TaskStats struct {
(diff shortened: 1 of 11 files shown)


master/static/migrations/20220621152800_update-allocation-state-values.tx.up.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TYPE public.allocation_state as ENUM (
```
stoksc, line 1: can you add this enum to the test `TestExhaustiveEnums`

master/pkg/model/task.go
```diff
@@ -148,57 +148,31 @@ func MostProgressedAllocationState(states ...AllocationState) AllocationState {
...
-	case AllocationStateRunning:
-		return taskv1.State_STATE_RUNNING
-	case AllocationStateTerminating:
-		return taskv1.State_STATE_TERMINATING
-	case AllocationStateTerminated:
-		return taskv1.State_STATE_TERMINATED
-	default:
-		return taskv1.State_STATE_UNSPECIFIED
-	}
+	return taskv1.State(taskv1.State_value[fmt.Sprintf("STATE_%s", s)])
```
stoksc, line 175: idk, I still prefer the exhaustive switch here.. or if there was a test to ensure proto and go (like the go and pg test) were always in sync this is fine.

stoksc (approved):
lgtm. this migration will probably be plenty fast. I guess we'll see.

master/pkg/model/task.go
NicholasBlaskey: Changed it back. 

Just wondering what is the reasoning behind not just using the Proto enum inside the Go model?

PR #4409, opened 2022-06-22
test: tune web tests

hkang1:
## Description

Address the following:
- silence async-validator warnings show around `antd` forms (not test failures but adds a lot of noise)
- fix missing `key` error in `LogViewerFilter` tests

Sample `async-validator` warning:
```
 PASS  src/hooks/useModal/UserSettings/useModalChangeName.test.tsx (42.857 s)
  ● Console

    console.warn
      async-validator: [ "'displayName' cannot be longer than 80 characters" ]

      at Function.warning (node_modules/src/util.js:18:17)
      at callback (node_modules/src/index.js:162:20)
      at Object.validator (node_modules/src/validator/string.js:32:3)
      at func (node_modules/src/index.js:237:22)
      at forEach (node_modules/src/util.js:113:5)
          at Array.forEach (<anonymous>)
      at asyncParallelArray (node_modules/src/util.js:112:7)
      at node_modules/src/util.js:199:9
```

Sample `LogViewerFilter` missing key warning:
```
 PASS  src/components/LogViewer/LogViewerFilters.test.tsx (40.121 s)
  ● Console

    console.error
      Warning: Each child in a list should have a unique "key" prop.
      
      Check the render method of `LogViewerFilters`. See https://reactjs.org/link/warning-keys for more information.
          at Option
          at LogViewerFilters (/home/circleci/project/webui/react/src/components/LogViewer/LogViewerFilters.tsx:38:46)

      117 |         )}
      118 |         {moreThanOne.rankIds && (
    > 119 |           <MultiSelect
          |           ^
      120 |             itemName={LABELS.rankIds}
      121 |             value={values.rankIds}
      122 |             onChange={handleChange('rankIds', Number)}>

```

## Test Plan

- [ ] check test logs to ensure there are no `async-validator` warnings
- [ ] check that LogViewerFilter tests are not warning about missing `key` attribute

## Checklist

### DIFF
--- a/webui/react/src/components/LogViewer/LogViewerFilters.tsx
+++ b/webui/react/src/components/LogViewer/LogViewerFilters.tsx
@@ -90,7 +90,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
             value={values.allocationIds}
             onChange={handleChange('allocationIds', String)}>
             {selectOptions?.allocationIds?.map(id => (
-              <Option key={id} value={id}>{id || 'No Allocation ID'}</Option>
+              <Option key={id || 'no-id'} value={id}>{id || 'No Allocation ID'}</Option>
             ))}
           </MultiSelect>
         )}
@@ -100,7 +100,7 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values 
             value={values.agentIds}
             onChange={handleChange('agentIds', String)}>
             {selectOptions?.agentIds?.map(id => (
-              <Option key={id} value={id}>{id || 'No Agent ID'}</Option>
+              <Option key={id || 'no-id'} value={id}>{id || 'No Agent ID'}</Option>
             ))}
           </MultiSelect>
         )}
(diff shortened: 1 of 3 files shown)


webui/react/src/components/LogViewer/LogViewerFilters.tsx
```diff
@@ -89,8 +89,8 @@ const LogViewerFilters: React.FC<Props> = ({ onChange, onReset, options, values
             itemName={LABELS.allocationIds}
             value={values.allocationIds}
             onChange={handleChange('allocationIds', String)}>
-            {selectOptions?.allocationIds?.map(id => (
-              <Option key={id} value={id}>{id || 'No Allocation ID'}</Option>
+            {selectOptions?.allocationIds?.map((id, index) => (
+              <Option key={id || `no-id-${index}`} value={id}>{id || 'No Allocation ID'}</Option>
```
hkang1, line 93: extra precaution of if there are multiple options that have id of empty strings. This shouldn't happen from the API side, but you never know.

hamidzr: we could probably always go with index. it'd be less stable rendering wise maybe but should do the job.since it's for testing. works as is too

hamidzr (approved):
lgtm. 
test: run this before and after
make test 2>&1 > /dev/null| grep -E "(async-validator|unique)"

webui/react/src/components/LogViewer/LogViewerFilters.tsx
hkang1: I was bit hesitant to just use the `index` in the case where new ids came in (since the log filter options may change over time while a trial is running for instance) and with the index as the key, the options might not re-render properly. Not 100% sure about that.

PR #4411, opened 2022-06-22
fix: End agent stats log to debug level

gt2345:
## Description

Suppress unnecessary log to debug level. 

## Test Plan

Simulate a situation where agent is up but stats is not recorded, then kill the agent, there should be a debug level log `End agent stats call affects 0 row(s), this number is suppose to be 1.` instead of any error.

## Checklist

### DIFF
--- a/master/internal/db/postgres_agent.go
+++ b/master/internal/db/postgres_agent.go
@@ -1,6 +1,10 @@
 package db
 
 import (
+	"context"
+
+	log "github.com/sirupsen/logrus"
+
 	"github.com/determined-ai/determined/master/pkg/model"
 )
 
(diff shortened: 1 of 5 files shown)


master/internal/db/postgres_agent.go
```diff
@@ -17,12 +21,18 @@ WHERE NOT EXISTS (
...
-WHERE agent_id = :agent_id AND end_time IS NULL
-`, a)
+func EndAgentStats(a *model.AgentStats) error {
+	res, err := Bun().NewUpdate().Table("agent_stats").Set(
+		"end_time = (SELECT CURRENT_TIMESTAMP)").Where(
+		"agent_id = ?", a.AgentID).Where("end_time IS NULL").Exec(context.TODO())
+	if err != nil {
+		return err
+	}
+	count, err := res.RowsAffected()
```
ioga, line 31: when zero rows is affected, is `err` here nil or not?

ioga (approved):

master/internal/db/postgres_agent.go
gt2345: `err` here will be nil no matter how many rows are affected

PR #4420, opened 2022-06-23
feat: Redesigning expected experiment states [DET-7278] [DET-7719]

mapmeld:
## Description

This WIP branch is following our Figma flowchart for more intuitive UI for Experiment State.

This change applies three rules:
- While the experiment.state is Active / Job is Scheduled, its UI label is "Running"
- If the experiment requires too many slots to **ever** run on an existing agent (currently using `(e.config->'resources'->>'slots_per_trial')::int <= (SELECT MAX(slots) FROM slots_by_agent)`), its UI label is "Queued" 
- For these two states, the state badge UI will match on the Experiment List and Detail views

This is part of a larger project to change any state badge in the UI (Experiments, Trials, Tasks, Commands, and Jobs).

## Test Plan

SQL to change an experiment slots_per_trial:
`update experiments set config = config||'{"resources":{"slots_per_trial": 3}}' WHERE id = 1887;`

There ought to be a helper function used in the decoder for Experiment and ExperimentList, which then has its own test.

From my testing: pause, stop, and complete appear to be consistently working.

## Checklist


webui/react/src/utils/sort.ts
```diff
@@ -7,18 +7,19 @@ import { NullOrUndefined, Primitive } from '../shared/types';
...
-  [RunState.Completed]: 5,
-  [RunState.StoppingCanceled]: 6,
-  [RunState.Canceled]: 7,
-  [RunState.Deleted]: 7,
-  [RunState.Deleting]: 7,
-  [RunState.DeleteFailed]: 7,
-  [RunState.Insufficient]: 8,
-  [RunState.Unspecified]: 9,
-  [RunState.Pending]: 10,
+  [RunState.Pending]: 2,
```
hamidzr, line 10: assuming typescript can still infer the types:
could we generate this off of an ordered array? that'd make the diffs more readable.

hamidzr (commented):
I realized I have some old comments pending.. here they go hopefully not too outdated

master/static/srv/get_experiment.sql
```diff
@@ -13,7 +24,18 @@ SELECT
...
     e.start_time AS start_time,
     e.end_time AS end_time,
-    'STATE_' || e.state AS state,
+    CASE
+      WHEN e.state = 'PAUSED'
+        THEN 'STATE_PAUSED'
+      WHEN e.state = 'CANCELED'
+        THEN 'STATE_CANCELED'
+      WHEN (e.config->'resources'->>'slots_per_trial')::int > max_slots.count
+        THEN 'STATE_INSUFFICIENT'
```
ioga, line 33: Can we please discuss it in backend office hours or a dedicated meeting? I am not sure this is good long-term plan to implement this feature.

hamidzr: I'd like to listen in if/when this happens ✋🏼

mapmeld: ok, I will plan to move Insufficient to be more like Queued which has to come from Resource Manager.
I created a new PR today #4524 to prevent these experiments from being created in the first place

docs/release-notes/4420-get-experiments-consolidation.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Breaking Changes**
+
+-  API: the GetProjectExperiments endpoint (/projects/:id/experiments) is removed and replaced with
```
gh-determined-ai, line 5: Needs syntax highlighting:
-  API: the ``GetProjectExperiments`` endpoint (``/projects/:id/experiments``) is removed and replaced with
   a *project_id* parameter on the ``GetExperiments`` endpoint (``/experiments``).

gh-determined-ai (approved):

webui/react/src/utils/sort.ts
mapmeld: I've just added a commit to have an array of RunStates, and store them and an index in a JavaScript Map object. Same for CommandStates.

docs/release-notes/4420-get-experiments-consolidation.txt
mapmeld: ✅

master/internal/api_experiment.go
```diff
@@ -126,6 +133,11 @@ func (a *apiServer) GetExperiment(
 		logrus.WithError(err).Debugf("asking for job summary")
 	} else {
 		resp.JobSummary = jobSummary
+		resp.Experiment.JobState = mapJobStateToExperimentState[jobSummary.State]
+		if jobSummary.State == jobv1.State_STATE_QUEUED &&
+			exp.State == experimentv1.State_STATE_ACTIVE {
+			exp.State = experimentv1.State_STATE_QUEUED
+		}
```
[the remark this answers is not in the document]
mapmeld: the goal here is to set experiment.state to QUEUED based on jobState, but only if experiment is Active and not already in an overriding state (for example, paused or cancelled)

master/internal/api_command.go
```diff
@@ -174,14 +176,27 @@ func (a *apiServer) GetCommands(
 	if err = a.ask(commandsAddr, req, &resp); err != nil {
 		return nil, err
 	}
+	for _, task := range resp.Commands {
```
stoksc, line 179: since the command actor is already responsible for converting this to proto, I think we should push these command-related conversions down to `toNotebook`/`toShell`... to try to cut down on the duplication.

master/internal/api_experiment.go
```diff
@@ -115,6 +116,41 @@ func (a *apiServer) GetExperiment(
 		resp.JobSummary = jobSummary
 	}
 
+	expTasks := []*taskv1.ExperimentAllocation{}
+	err = a.m.db.QueryProto(
+		"get_active_allocations",
+		&expTasks,
+		0,
```
stoksc, line 123: don't think this is right

master/internal/api_trials.go
```diff
@@ -479,6 +480,45 @@ func (a *apiServer) GetExperimentTrials(
...
+	for _, trial := range resp.Trials {
+		if trial.State == experimentv1.State_STATE_ACTIVE {
+			if setState, ok := applyIDs[trial.Id]; ok {
+				trial.State = setState
+			}
+			//  else {
+			// 	trial.State = experimentv1.State_STATE_QUEUED
+			// }
+		}
+	}
```
stoksc, line 520: I think we could rewrite all this code (and everything with a similar structure, basically) in a way that is generic to jobs (so not experiment or trial specific). Maybe be helpful to think of it like aggregating the job/task state from its job/task id, then converting job/task states to experiment, trial, etc states. Would save us a lot of duplication and help us get better about maintaining feature parity/identifying gaps in feature parity between job types to have everything be the same under the hood with some thin shims to fit specific types' requirements.

proto/src/determined/task/v1/task.proto
```diff
@@ -22,6 +22,8 @@ enum State {
   STATE_TERMINATED = 6;
   // The task has begun to exit.
   STATE_TERMINATING = 7;
+  // Additional state to cover queueing operations.
+  STATE_QUEUED = 8;
```
stoksc, line 26: with this change, we'll have obsoleted some other states (the api will never return them). I think they should be removed for clarity.

proto/src/determined/task/v1/task.proto
```diff
@@ -47,3 +49,15 @@ message Task {
   // List of Allocations.
   repeated Allocation allocations = 4;
 }
+
+// Task allocations with an associated Experiment.
+message ExperimentAllocation {
```
stoksc, line 54: this should just be a go struct, not be in protos

stoksc (commented):
in the right direction, but I think we need to cut down on the amount of duplication. these are things that are all _sort of_ supposed to be the same and they're all special cased. if you want some more direction there, lmk.

proto/src/determined/task/v1/task.proto
mapmeld: ✅

master/internal/api_trials.go
mapmeld: My current solution for this is to return IsRunning and IsStarting from the SQL query into the struct... for whatever reason my initial query would return the enum's int values instead of strings.  The new version simplifies things but I ought to test it out on more complex Tasks and multi-trial Experiments.

master/static/srv/get_active_allocations.sql
```diff
@@ -0,0 +1,13 @@
...
+  WHERE state IN ('PULLING', 'RUNNING', 'STARTING')
+)
+SELECT trials.task_id AS TaskID,
+  trials.id AS TrialID,
+  trials.experiment_id AS ExperimentID,
+  (SELECT COUNT(*) > 0 FROM activeAllocations WHERE activeAllocations.task_id = trials.task_id AND state = 'RUNNING') AS IsRunning,
+  (SELECT COUNT(*) > 0 FROM activeAllocations WHERE activeAllocations.task_id = trials.task_id) AS IsStarting
+FROM trials
+WHERE ($1 = 0 OR trials.id = $1)
+  AND ($2 = 0 OR trials.experiment_id = $2);
```
stoksc, line 13: this is going to be an expensive query with that CTE on some Postgres versions. can you rewrite as a join?

stoksc: also, maybe write it in terms of jobs/tasks instead of exp/trials

mapmeld: ✅  made an update using tasks table and JOIN

master/internal/api_trials.go
mapmeld: Updated with counts in the SQL for running allocations and all pending/running/starting allocations

proto/src/determined/task/v1/task.proto
mapmeld: would it be reasonable to deprecate states and leave removal for a set future date, in case the user's server version / CLI version / browser version are not all being kept up-to-date

stoksc (commented):
had this pending half-review from days ago i forgot to submit

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -88,7 +88,8 @@ def test_agent_restart_exp_container_failure(managed_cluster_restarts: ManagedCl
         # As soon as the agent is back, the original allocation should be considered dead,
         # but the new one should be allocated.
         state = exp.experiment_state(exp_id)
-        assert state == EXP_STATE.STATE_ACTIVE
+        assert state == EXP_STATE.STATE_QUEUED
+        exp.wait_for_experiment_state(exp_id, EXP_STATE.STATE_ACTIVE)
```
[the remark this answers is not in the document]
mapmeld: in the new system the API returns STATE_ACTIVE to mean "Running" in the UI, but I don't have a server-side STATE_RUNNING constant / object

master/internal/api_tasks.go
```diff
@@ -35,6 +35,16 @@ var (
 	taskNotFound = status.Error(codes.NotFound, "task not found")
 )
 
+var pendStates = []taskv1.State{
+	taskv1.State_STATE_PULLING,
+	taskv1.State_STATE_STARTING,
+}
+
+var queueStates = []taskv1.State{
+	taskv1.State_STATE_PENDING,
```
ioga, line 44: the fact that STATE_PENDING Is not in "pending state" list is certainly confusing. I wish we had better names for these.

e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -88,7 +88,8 @@ def test_agent_restart_exp_container_failure(managed_cluster: ManagedCluster) ->
...
```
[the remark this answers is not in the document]
mapmeld: I ended up making this line into all three possible active states - `assert state in [EXP_STATE.STATE_QUEUED, EXP_STATE.STATE_PENDING, EXP_STATE.STATE_ACTIVE]` and then waiting for the active/running state

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -36,6 +36,22 @@ enum State {
...
+// Job state.
+enum JobState {
+  // Unspecified state.
+  JOB_STATE_UNSPECIFIED = 0;
+  // Job is queued and waiting to be schedlued.
+  JOB_STATE_QUEUED = 1;
+  // Job is scheduled.
+  JOB_STATE_SCHEDULED = 2;
+  // Job is scheduled as a backfill.
+  JOB_STATE_SCHEDULED_BACKFILLED = 3;
```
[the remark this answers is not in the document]
mapmeld: no longer using `JobState`

proto/src/determined/experiment/v1/experiment.proto
```diff
@@ -36,6 +36,10 @@ enum State {
   STATE_DELETE_FAILED = 11;
   // The experiment is killed and is shutting down.
   STATE_STOPPING_KILLED = 12;
+  // The experiment has an allocation which is being spun up for use.
+  STATE_PENDING = 13;
```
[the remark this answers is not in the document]
mapmeld: ✅ - we might talk about this later, but does it work for Pending and Queued to be substates of Active, and then when the API returns Active it means the experiment is running?

stoksc:
@mapmeld can we query plan 'get_active_allocations' in the most insane case and post back here? just to make sure we're not going to regret this.

master/internal/api_experiment.go
```diff
@@ -45,6 +45,61 @@ import (
...
+}
+
+func (a *apiServer) enrichExperimentState(experiments []*experimentv1.Experiment) (
+	[]*experimentv1.Experiment, error,
+) {
+	// filter by JobID when getting a single experiment
+	jobID := ""
+	if len(experiments) == 1 {
+		jobID = experiments[0].JobId
+	}
```
stoksc, line 63: so we would filter by nothing with a list?

master/static/srv/get_active_allocations_by_job.sql
```diff
@@ -0,0 +1,9 @@
+SELECT tasks.task_id AS TaskID,
+  tasks.job_id AS JobID,
+  SUM(case when allocations.state = 'RUNNING' then 1 else 0 end) AS NumRunning,
+  COUNT(allocations) AS NumStarting
+FROM tasks
+JOIN allocations ON allocations.task_id = tasks.task_id
+WHERE tasks.job_id in (SELECT unnest(string_to_array($1, ','))) -- Experiment match
+  AND allocations.state IN ('PULLING', 'RUNNING', 'STARTING')
+GROUP BY tasks.task_id, tasks.job_id;
```
stoksc, line 9: maybe something more like (table aliases, bool_or since we dont really need counts or sums, no group by task since this is for jobs, then the `enrichExperimentState` func can be simplified, too). also we have to join through jobs to get the index.
```
SELECT
  j.job_id,
  BOOL_OR(CASE WHEN a.state IN ('PULLING', 'STARTING') THEN true ELSE false END) AS is_starting,
  BOOL_OR(CASE WHEN a.state = 'RUNNING' THEN true ELSE false END) AS is_running
FROM jobs j
JOIN tasks t ON t.job_id = j.job_id
JOIN allocations a ON a.task_id = t.task_id
WHERE j.job_id in (SELECT unnest(string_to_array($1, ',')))
GROUP BY j.job_id;
```

master/static/srv/get_active_allocations_by_task.sql
```diff
@@ -0,0 +1,9 @@
+SELECT tasks.task_id AS TaskID,
+  tasks.job_id AS JobID,
+  SUM(case when allocations.state = 'RUNNING' then 1 else 0 end) AS NumRunning,
+  COUNT(allocations) AS NumStarting
+FROM tasks
+JOIN allocations ON allocations.task_id = tasks.task_id
+WHERE tasks.task_id IN (SELECT unnest(string_to_array($1, ','))) -- Trial match
+  AND allocations.state IN ('PULLING', 'RUNNING', 'STARTING')
+GROUP BY tasks.task_id, tasks.job_id;
```
stoksc, line 9: similar to above
```
SELECT
  t.task_id,
  BOOL_OR(CASE WHEN a.state IN ('PULLING', 'STARTING') THEN true ELSE false END) AS is_starting,
  BOOL_OR(CASE WHEN a.state = 'RUNNING' THEN true ELSE false END) AS is_running
FROM tasks t
JOIN allocations a ON a.task_id = t.task_id
WHERE t.task_id IN (SELECT unnest(string_to_array($1, ',')))
GROUP BY t.task_id;
```

master/internal/api_experiment.go
```diff
@@ -46,6 +46,60 @@ import (
...
+	applyIDs := make(map[string]experimentv1.State)
+	for _, task := range tasks {
+		if task.NumRunning > 0 {
+			applyIDs[*task.JobID] = experimentv1.State_STATE_ACTIVE
+		} else if task.NumStarting > 0 {
+			_, expStoredState := applyIDs[*task.JobID]
+			if !expStoredState {
+				applyIDs[*task.JobID] = experimentv1.State_STATE_PENDING
+			}
+		}
```
stoksc, line 87: let's make this exhaustive since it is easier to reason about (in terms of my query comments). same for trial state.
```
switch {
case task.IsRunning:
	byJobID[task.JobID] = model.MostProgressedAllocationState(byJobID[task.JobID], experimentv1.State_STATE_ACTIVE)
case task.IsStarting:
	byJobID[task.JobID] = model.MostProgressedAllocationState(byJobID[task.JobID], experimentv1.State_STATE_PENDING)
default:
	byJobID[task.JobID] = model.MostProgressedAllocationState(byJobID[task.JobID], experimentv1.State_STATE_QUEUED)
}
```

proto/src/determined/experiment/v1/experiment.proto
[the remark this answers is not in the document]
stoksc: maybe we talked about it, i think i'd prefer active -> running everywhere, otherwise the states are a bit confusing (active implies pending/queued/running is less confusing than active imples pending/queued/active because its less clear what active actually is)

master/internal/api_experiment.go
```diff
@@ -46,6 +46,60 @@ import (
...
+	[]*experimentv1.Experiment, error,
+) {
+	// filter allocations by JobIDs on this page of experiments
+	jobFilter := []string{}
+	for _, exp := range experiments {
+		jobFilter = append(jobFilter, exp.JobId)
+	}
+
+	// get active or pending tasks
+	tasks := []*experimentAllocation{}
```
stoksc, line 67: use a specialized struct for each query, maybe use bun if you are brave.

stoksc (commented):
looks a lot better with those changes, just a few more comments

master/internal/api_experiment.go
```diff
@@ -71,7 +125,13 @@ func (a *apiServer) getExperiment(experimentID int) (*experimentv1.Experiment, e
...
+
+	expListForm, err := a.enrichExperimentState([]*experimentv1.Experiment{exp})
+	if err != nil {
+		return nil, err
+	}
+
+	return expListForm[0], nil
 }
 
 func (a *apiServer) GetExperiment(
```
ioga, line 137: should GetExperiment also do something new? it'd be weird if the state between `GetExperiment` and `GetExperiments` is different because one is using JobSummary and another is doing the enrichment.

mapmeld: `GetExperiment` (the route) goes through `getExperiment` (the helper function) which goes through enrichment (`a.enrichExperimentState([]*experimentv1.Experiment{exp})`)

On review, I've changed the valid states for when to check `GetJobSummary` vs. ending GetExperiment early

master/internal/api_experiment.go
mapmeld: :+1:

master/static/srv/get_active_allocations_by_job.sql
mapmeld: applied the SQL changes :+1:

mapmeld:
@stoksc Added STATE_RUNNING back into the mix, and rebased to get the new experiments bun code, so hopefully that all runs through

master/internal/api_experiment.go
mapmeld: I made separate taskAllocation and experimentAllocation, is that what you meant?

master/internal/api_experiment.go
```diff
@@ -66,17 +123,23 @@ func (a *apiServer) getExperiment(
 	if err != nil {
 		return nil, err
 	}
-	if ok, err := expauth.AuthZProvider.Get().
+	if ok, authErr := expauth.AuthZProvider.Get().
 		CanGetExperiment(curUser, modelExp); err != nil {
```
NicholasBlaskey, line 127:
```suggestion
		CanGetExperiment(curUser, modelExp); authErr != nil {
```

mapmeld: thanks! 👍

master/internal/command/command.go
```diff
@@ -40,6 +42,25 @@ const terminatedDuration = 24 * time.Hour
...
+		taskv1.State_STATE_PULLING,
+		taskv1.State_STATE_STARTING,
+	}
+
+	queueStates := []taskv1.State{
+		taskv1.State_STATE_PENDING,
+		taskv1.State_STATE_ASSIGNED,
+	}
+
+	if slices.Contains(queueStates, state) {
```
stoksc, line 56: style: use a switch.

master/internal/command/command.go
```diff
@@ -40,6 +42,25 @@ const terminatedDuration = 24 * time.Hour
 // should stop and garbage collect its state.
 type terminateForGC struct{}
 
+func enrichState(state taskv1.State) taskv1.State {
```
stoksc, line 45: the disconnect between what pending means internally and what is means externally (pending isnt a pending state) is probably going to get really confusing really fast. any thoughts?

master/pkg/model/experiment.go
```diff
@@ -465,6 +465,28 @@ type TrialMetrics struct {
 	Metrics      JSONObj    `db:"metrics" json:"metrics"`
 }
 
+// Represent order of active states (Queued -> Pending -> Running).
+var allocationStateIndex = map[experimentv1.State]int{
```
stoksc, line 469: maybe this? this function already exists for allocation states
```suggestion
var experimentStateIndex = map[experimentv1.State]int{
```

stoksc (commented):
looks pretty good, last round of feedback, probably.

master/internal/command/command.go
stoksc: this is the only thing i'm really concerned about, maybe we can chat offline about a timeline for cleaning up the states that are now dead as far as the API is concerned.

docs/release-notes/4420-substates-of-active
```diff
@@ -0,0 +1,11 @@
+:orphan:
+
+**API Changes**
+
+- The ``STATE_ACTIVE`` state for Experiments and Trials is divided into four sub-states:
+ ``STATE_QUEUED``, ``STATE_PULLING``, ``STATE_STARTING``, and ``STATE_RUNNING``.
+ Queries to ``GetExperimentsRequest`` which filter by state continue to use
```
stoksc, line 7: i think this points out a sort of interesting gap in the current implementation. Being able to filter by ACTIVE is still great, but would be a "nice to have" to allow filtering by the substates too. Maybe ticket it?

docs/release-notes/4420-substates-of-active
```diff
@@ -0,0 +1,11 @@
...
+ ``STATE_ACTIVE``.
+
+- The state of Tasks is adjusted to match Experiments and Trials. Previous
+  ``STATE_PENDING`` and ``STATE_ASSIGNED`` become ``STATE_QUEUED``.
```
stoksc, line 11: Just a note, for context, assigned in k8s means pod created (not necessarily scheduled), in determined agents it means the task has acknowledged it received assignment (which actually means scheduled). Queued is sufficient since in determined agents nothing should stay in that start longer than an master:allocation<->agent:containers actor round trip, but just wanted to note.

stoksc: Could be a place for us, later, to internally improve the meaning of states too. So people dont have to know every RM implementation to manage.

master/internal/api_experiment.go
```diff
@@ -51,6 +53,65 @@ import (
...
+	// filter allocations by JobIDs on this page of experiments
+	jobFilter := make([]string, 0, len(experiments))
+	for _, exp := range experiments {
+		jobFilter = append(jobFilter, exp.JobId)
+	}
+
+	// get active experiments by JobID
+	tasks := []experimentAllocation{}
+	err := a.m.db.Query(
+		"get_active_allocations_by_job",
```
stoksc, line 76: i would rename these queries to `aggregate_allocation_state_by_job` or something to better express its cost.

master/internal/api_experiment.go
```diff
@@ -840,7 +914,13 @@ func (a *apiServer) PatchExperiment(
 		}
 	}
 
-	return &apiv1.PatchExperimentResponse{Experiment: exp}, nil
+	// include queued / pulling / starting / running state
+	expListForm, err := a.enrichExperimentState([]*experimentv1.Experiment{exp})
```
stoksc, line 918: since this takes a pointer and mutates the experiment anyway, you could just not make it 'appear' functional and also make it take a variadic number of experiments, solving this weird slice pass/slice get thing

master/internal/api_trials.go
```diff
@@ -93,6 +93,61 @@ type TrialLogBackend interface {
...
+	// Active trials converted to Queued, Pulling, Starting, or Running
+	for _, trial := range trials {
+		if trial.State == experimentv1.State_STATE_ACTIVE {
+			if setState, ok := byTaskID[model.TaskID(trial.TaskId)]; ok {
+				trial.State = setState
+			} else {
+				trial.State = experimentv1.State_STATE_QUEUED
+			}
+		}
+	}
```
stoksc, line 146: if you wanted to make this logic generic, probably the easiest approach would be to make `trialAllocation`/`experimentAllocation` implement some `aggregateAllocationStateWithID` interface and then make this a func that takes that.

master/pkg/model/experiment.go
```diff
@@ -465,6 +465,29 @@ type TrialMetrics struct {
...
 }
 
+// Represent order of active states (Queued -> Pulling -> Starting -> Running).
+var experimentStateIndex = map[experimentv1.State]int{
+	experimentv1.State_STATE_UNSPECIFIED: -2,
+	experimentv1.State_STATE_ACTIVE:      -1,
+	experimentv1.State_STATE_QUEUED:      0,
+	experimentv1.State_STATE_PULLING:     1,
+	experimentv1.State_STATE_STARTING:    2,
+	experimentv1.State_STATE_RUNNING:     3,
```
stoksc, line 475: go ahead and list all possible states in order?

proto/src/determined/task/v1/task.proto
stoksc: i think i changed my mind here, just go ahead and remove them. if their CLI is not up to date (the most likely thing to happen) they will get a warning anyway and it is in the release notes.

docs/release-notes/4420-substates-of-active
mapmeld: Created https://determinedai.atlassian.net/browse/DET-8333

master/internal/api_experiment.go
mapmeld: on this and `getExperiment` (lowercase) I changed it to pass by reference and not use `expListForm`, seems to work

master/internal/api_experiment.go
```diff
@@ -51,6 +53,65 @@ import (
...
 
+// Catches information on active running experiments.
+type experimentAllocation struct {
+	Job      model.JobID
+	Pulling  bool
+	Running  bool
+	Starting bool
+}
+
+func (a *apiServer) enrichExperimentState(experiments []*experimentv1.Experiment) (
```
stoksc, line 64: this can just return an error now. and what i meant was to make this
```suggestion
func (a *apiServer) enrichExperimentState(experiments ...*experimentv1.Experiment) (
```

master/internal/api_trials.go
```diff
@@ -93,6 +93,61 @@ type TrialLogBackend interface {
...
 
+// Catches information on active running trials.
+type trialAllocation struct {
+	Pulling  bool
+	Running  bool
+	Starting bool
+	Task     model.TaskID
+}
+
+func (a *apiServer) enrichTrialState(trials []*trialv1.Trial) ([]*trialv1.Trial, error) {
```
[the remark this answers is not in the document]
mapmeld: ✅

master/internal/api_experiment.go
```diff
@@ -134,7 +198,11 @@ func (a *apiServer) GetExperiment(
 		Experiment: exp,
 	}
 
-	if model.StateFromProto(exp.State) != model.ActiveState {
+	if !slices.Contains([]experimentv1.State{
+		experimentv1.State_STATE_ACTIVE,
+		experimentv1.State_STATE_PULLING, experimentv1.State_STATE_QUEUED,
+		experimentv1.State_STATE_RUNNING, experimentv1.State_STATE_STARTING,
```
stoksc, line 204: can you make a helper for this?

stoksc (approved):
lgtm. though I do think in the near future we need to materialize theses 'virtual' states on the task and jobs table.

master/internal/api_experiment.go
mapmeld: ✅

PR #4423, opened 2022-06-24
chore: strict yaml parsing warning for master and error for templates [DET-7639, DET-7640]

NicholasBlaskey:
## Description

Strict yaml parsing warning for master and error for templates. Templates already give an error but when you try and use the templates not when you create them.

## Test Plan

Templates e2e tests.

Master yaml
```
db: {}
db:
  name: test
  name: abcd
```

Should print
```
WARN[2022-06-24T11:46:53-04:00] master configuration parsing warning (will soon be error) error converting YAML to JSON: yaml: unmarshal errors:
  line 4: key "name" already set in map
  line 3: key "db" already set in map 
```

Extra keys
```
extra: {}
```

Should print
```
WARN[2022-06-24T11:47:35-04:00] master configuration parsing warning (will soon be error) error unmarshaling JSON: while decoding JSON: json: unknown field "extra" 
```

Extra key with a value should have already given an error 
```
extra=true
```

```
ERRO[2022-06-24T11:48:25-04:00] error unmarshaling JSON: while decoding JSON: json: unknown field "extra"
```

## Checklist

### DIFF
--- a/master/cmd/determined-master/root.go
+++ b/master/cmd/determined-master/root.go
@@ -86,7 +86,7 @@ func initializeConfig() error {
 	// not_a_real_key: {}
 	var c *config.Config
 	if err = yaml.UnmarshalStrict(bs, &c, yaml.DisallowUnknownFields); err != nil {
-		log.Warnf("master configuration parsing warning (will soon be error) %s", err)
+		log.Warnf("master configuration parsing warning: %s", err.Error())
 		// TODO uncomment to turn strict parsing warnings into errors next breaking change.
 		// return err
 	}
--- a/master/cmd/determined-master/root.go
+++ b/master/cmd/determined-master/root.go
@@ -79,6 +79,18 @@ func initializeConfig() error {
 	if err != nil {
 		return err
 	}
+
+	// Validate warnings for extraneous and duplicate keys before
+	// Viper merges configuration. Viper does some weird merging logic
+	// that appears to dedupe some values along with ignoring config like
+	// not_a_real_key: {}
+	var c *config.Config
+	if err := yaml.UnmarshalStrict(bs, &c, yaml.DisallowUnknownFields); err != nil {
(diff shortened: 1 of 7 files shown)


gh-determined-ai (approved):

master/cmd/determined-master/root.go
```diff
@@ -73,6 +73,18 @@ func initializeConfig() error {
...
 		return err
 	}
+
+	// Validate warnings for extraneous and duplicate keys before
+	// Viper merges configuration. Viper does some weird merging logic
+	// that appears to dedupe some values along with ignoring config like
+	// not_a_real_key: {}
+	var c *config.Config
+	if err = yaml.UnmarshalStrict(bs, &c, yaml.DisallowUnknownFields); err != nil {
+		log.Warnf("master configuration parsing warning (will soon be error) %s", err)
```
eecsliu, line 83: nit: wording feels weird here - are you saying that it will become an error in the future? Or that it will error later in the code? If the former, I'd just call it a warning and `TODO` the comment change, the latter I'd just call it an error.

eecsliu (approved):
nice work! Apologies for the late

master/internal/template/api.go
```diff
@@ -44,7 +45,10 @@ func (m *manager) put(c echo.Context) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := yaml.Unmarshal(body, make(map[interface{}]interface{})); err != nil {
+
+	// Validate template.
+	var t expconf.ExperimentConfig
+	if err := yaml.UnmarshalStrict(body, &t, yaml.DisallowUnknownFields); err != nil {
```
NicholasBlaskey, line 51: This isn't actually OK since templates can be used for command config too. 

Busy with other work now but will come back to this eventually to figure a workaround

NicholasBlaskey:
More complicated than this since checking templates needs to check for both expconf and command config

PR #4031, opened 2022-04-28
feat: rolling upgrades v0

ioga:
## Test Plan
Configuration:
- Agent reconnect attempts and backoff interval should bumped up to accommodate for longer master downtimes, e.g.: 
```
agent_reconnect_attempts: 25
agent_reconnect_backoff: 5
```
- Set `agent_reattach_enabled: true` on compute resource pools.
- `agent_reconnect_wait` in master must be higher than `agent_reconnect_backoff` to give the agents time after master startup to reconnect. Default is 25 seconds, which should be okay.

Update/upgrade procedure:
1. Launch an experiment
2. Restart master (with new config or new version)
3. Experiment should continue running
4. Restart agent to upgrade, if necessary.

Limitations:
- Only trials are currently can be recovered with zero downtime. No support for notebooks and other task types yet.
- Agent RM only.

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

PR #4424, opened 2022-06-24
feat: custom searcher

mpkouznetsov:
## Description

Feature PR for Custom Searcher.

Adding @rb-determined-ai to reviewers specifically for style and consistency of APIs.

## Checklist


harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,250 @@
...
+        self.last_event_id = d.get("lastEventId")
+        self.experiment_id = d.get("experimentId")
+        self.experiment_completed = d.get("experimentCompleted", False)
+
+
+class ExitedReason(Enum):
+    ERRORED = "ERRORED"
+    USER_CANCELED = "USER_CANCELED"
+    INVALID_HP = "INVALID_HP"
+    INIT_INVALID_HP = "INIT_INVALID_HP"
```
rb-determined-ai, line 55: is there a good reason this needs to be distinct from INVALID_HP?

My memory tells me that they were distinct due to architectural reasons at the time, but I don't think that's the case anymore.

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,243 @@
...
+                        [SearchRunner._searcher_event_as_dict(e) for e in events.searcherEvents]
+                    )
+                )
+                # the first event is an event we have already processed and told master about it
+                # however, we may not have saved the state after that event if we crashed
+                # after POSTing operations but before saving state
+                last_event_id = self.search_method.searcher_state.last_event_id
+                first_event = True
+                for event in events.searcherEvents:
+                    assert event.id is not None
```
rb-determined-ai, line 48: is this an issue with the rest API? If all events should have an ID, why is it a nullable type at all?

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,243 @@
...
+    def run_experiment(self, experiment_id: int) -> None:
+        assert client._determined is not None
+        session = client._determined._session
+
+        experiment_is_active = True
+
+        try:
+            while experiment_is_active:
+                time.sleep(5)
+                events = bindings.get_GetSearcherEvents(session, experimentId=experiment_id)
```
rb-determined-ai, line 34: maybe long polling instead?

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,243 @@
...
+                        break
+                    elif event.trialProgress:
+                        logging.debug(
+                            f"trialProgress({event.trialProgress.requestId}, "
+                            f"{event.trialProgress.partialUnits})"
+                        )
+                        request_id = uuid.UUID(event.trialProgress.requestId)
+                        self.search_method.searcher_state.trial_progress[request_id] = float(
+                            event.trialProgress.partialUnits
+                        )
```
rb-determined-ai, line 131: Who owns the SearcherState?

It seems to have some mixed ownership, which is usually a messy experience.  There are some things that the user needs to be worried about some things that are populated automatically (like trial progress) which the user should consider read-only.

Is there a way to draw a clearer line between what is owned by the system vs what is owned by the user code?

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,243 @@
...
+                        body = bindings.v1PostSearcherOperationsRequest(
+                            experimentId=self.search_method.searcher_state.experiment_id,
+                            searcherOperations=[op._to_searcher_operation() for op in operations],
+                            triggeredByEvent=event,
+                        )
+                        bindings.post_PostSearcherOperations(
+                            session,
+                            body=body,
+                            experimentId=experiment_id,
+                        )
```
rb-determined-ai, line 149: I think, since this API is designed for very advanced users, we should consider being more flexible in the programming paradigms we allow.

Right now, the rules for user code are basically: 
1. you must write event-based programming
2. it has to fit into an event loop that we write that is based on us polling the master
3. you are only allowed to emit new operations in response to master events.

But what I see from the REST api is that technically the searcher could post new operations at any time.  Which means 3. isn't a requirement we actually need to enforce on the user.  Event 1. isn't _technically_ a requirement, though I don't know how a user could do anything useful without it.  And because I assume the users of this API will be quite advanced, 2. seems undesirable to enforce.  Users who decide to come up with new events at some random time (maybe after some computationally heavy process, or due to some external event) should be able to post new events without waiting for the next event from the master, it seems like.

We should talk more about this in a meeting, I'm thinking.

harness/determined/searcher/search_runner.py
mpkouznetsov: Searcher owns it but we want to make it available to the search method. As we discussed offline, we could emphasize that the user should not modify it by passing it to event handlers as a parameter.

harness/determined/searcher/search_runner.py
mpkouznetsov: We will try to fix this.

harness/determined/layers/_workload_sequencer.py
```diff
@@ -9,6 +9,7 @@
 WorkloadStreamElem = Tuple[workload.Workload, workload.ResponseFunc]
 
 WorkloadGenerator = Generator[WorkloadStreamElem, None, None]
+logger = logging.getLogger("determined.core")
```
rb-determined-ai, line 12: use `determined.layers` if you like, but this is not part of the core api.

harness/determined/layers/_workload_sequencer.py
```diff
@@ -374,6 +375,7 @@ def __iter__(self) -> workload.Stream:
                 yield from self.validate(None)
 
             for op in self.core_context.searcher.operations(core.SearcherMode.ChiefOnly):
+                logger.warning(f"SearcherOperation length={op.length}")
```
rb-determined-ai, line 378: warning-level?  Is this a leftover debug print?

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+import pickle
+import time
+from typing import Any, Dict, List, Optional, Tuple
+
+import determined as det
+from determined.common.api.bindings import (
+    determinedexperimentv1State,
+    get_GetExperiment,
+    post_PauseExperiment,
+)
```
rb-determined-ai, line 12: [google style guide for python imports](https://google.github.io/styleguide/pyguide.html#22-imports)

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+from determined.experimental import client
+from determined.searcher.search_method import Operation, SearchMethod
+from determined.searcher.search_runner import SearchRunner
```
rb-determined-ai, line 15: ditto

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
+import logging
```
rb-determined-ai, line 1: The file layout needs reconsideration to be suitable for external users and internal developers alike.

The file layout we use in newly added code is laid out in [this old rfc](https://github.com/rb-determined-ai/pedl/blob/9bd2168ec23fbba866194f3d7d3cfc243d267100/python_code_layout.md).

Basically, what should the user actually be importing in, and what should they be thinking about?

Should they need to think about `det.searcher`?  (my guess is probably)

Should they need to think about `det.searcher.core_search_runner`?  (my guess is probably not)

That means that this file probably needs to be named `_core_search_runner`, and `searcher/__init__.py` probably needs to expose some symbols from inside this file.

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+from determined.searcher.search_runner import SearchRunner
+
+
+class CoreSearchRunner(SearchRunner):
+    def __init__(self, search_method: SearchMethod, context: det.core.Context) -> None:
+        super().__init__(search_method)
+        self.context = context
+        self.info = det.get_cluster_info()
+
+        assert self.info is not None, "CoreSearchRunner only runs on-cluster"
```
rb-determined-ai, line 24: You are saving an `Optional[ClusterInfo]` object as `self.info`, which is why you have to have other assertions in other places.

Instead, assert not None before saving to `self.info`, and mypy will know self.info is of type `ClusterInfo`:

```python
info = det.get_cluster_info()
assert info is not None, "..."
self.info = info
```

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+        self.latest_checkpoint = self.info.latest_checkpoint
+
+    def run(
+        self,
+        exp_config: Dict[str, Any],
+        context_dir: Optional[str] = None,
+    ) -> int:
+        logging.info("CoreSearchRunner.run")
```
rb-determined-ai, line 32: probably worth using a logger named `det.searcher` for anything in this module.

Eventually we'll move the whole codebase to look like that.

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+        client._require_singleton(lambda: None)()
```
rb-determined-ai, line 33: I recommend a comment explaining what this is for.

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+        metadata = {"steps_completed": steps_completed}
+        with self.context.checkpoint.store_path(metadata) as (path, storage_id):
+            self.search_method.save(path, experiment_id=experiment_id)
+            with path.joinpath("ops").open("wb") as ops_file:
+                pickle.dump(operations, ops_file)
+
+    def maybe_preempt(self, session: client.Session, multitrial_experiment_id: int) -> bool:
+        if self.context.preempt.should_preempt():
+            # if searcher is preempted, then pause multitrial experiment
+            self._pause_and_wait(session, multitrial_experiment_id)
```
rb-determined-ai, line 76: Question: how do we become unpreempted?

IIUC, we can get preempted either by scheduler decision or user action.

In either case, it looks like we pause the experiment.  So if we got preemtped by the scheduler, we're going to push the experiment into a PAUSED state, then we put ourselves into a PAUSED state, then we won't ever wake back up.

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,121 @@
...
+def _is_experiment_active(state: determinedexperimentv1State) -> bool:
+    if (
+        state == determinedexperimentv1State.STATE_ACTIVE
+        or state == determinedexperimentv1State.STATE_QUEUED
+        or state == determinedexperimentv1State.STATE_RUNNING
+        or state == determinedexperimentv1State.STATE_STARTING
+        or state == determinedexperimentv1State.STATE_PULLING
+    ):
+        return True
+    return False
```
rb-determined-ai, line 121: simplify this: `return state in [...]`

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,245 @@
...
+            bindings_exited_reason
+            == bindings.v1TrialExitedEarlyExitedReason.EXITED_REASON_INVALID_HP
+        ):
+            return cls.INVALID_HP
+        if (
+            bindings_exited_reason
+            == bindings.v1TrialExitedEarlyExitedReason.EXITED_REASON_USER_REQUESTED_STOP
+        ):
+            return cls.USER_CANCELED
+        raise RuntimeError(f"Invalid exited reason: {bindings_exited_reason}")
```
rb-determined-ai, line 70: Can this get rewritten to look more like the [python sdk enums](https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/experiment.py#L12)?

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,245 @@
...
+            searcherProgress=bindings.v1SearcherProgressOperation(progress=self.progress)
+        )
+
+
+class Shutdown(Operation):
+    def __init__(self) -> None:
+        super().__init__()
+
+    def _to_searcher_operation(self) -> bindings.v1SearcherOperation:
+        return bindings.v1SearcherOperation(shutdown=bindings.v1ShutdownOperation())
```
rb-determined-ai, line 121: question: why have a shutdown?  Why not just exit 0 and let that be your shutdown?

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,245 @@
...
+        operation.
+        """
+        pass
+
+    @abstractmethod
+    def progress(self) -> float:
+        """
+        progress returns experiment progress as a float between 0 and 1.
+        """
+        pass
```
rb-determined-ai, line 190: Do we synthesize progress events in the master when operations are completed?

If we don't... report_progress() is actually an optional call so I want to make sure that report_progress() isn't the only way this gets called.

Separately, other events can still emit Progress operations, right?  Why not just let Progress be a normal operation that can be emitted anytime, and the progress event just returns normal operations as well?

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,245 @@
+import dataclasses
+import json
+import pathlib
+import uuid
+from abc import abstractmethod
+from enum import Enum
+from typing import Any, Dict, List, Optional, Set
+
+from determined.common.api import bindings
+from determined.common.experimental import Checkpoint
```
rb-determined-ai, line 10: google style guide for python imports

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,328 @@
...
+import uuid
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Sequence, Tuple
+
+from determined.common.api import bindings
+from determined.common.api.bindings import (
+    determinedexperimentv1State,
+    v1SearcherEvent,
+    v1TrialExitedEarlyExitedReason,
+)
```
rb-determined-ai, line 15: google style guide for python imports.

Also, in this case you've actually already imported `bindings`, so just use it.

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,328 @@
...
+from determined.experimental import client
+from determined.searcher.search_method import ExitedReason, Operation, Progress, SearchMethod
```
rb-determined-ai, line 17: google style guide for python imports

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,328 @@
...
+
+EXPERIMENT_ID_FILE = "experiment_id"
```
rb-determined-ai, line 19: what do you think about `"experiment_id.txt"`?

I don't mind text files in unix without the .txt extension myself.  Just a thought.

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,328 @@
...
+STATE_FILE = "state"
```
rb-determined-ai, line 20: appears unused in this file.

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,328 @@
...
+                        and last_event_id != 0
+                        and last_event_id >= event.id >= 0
+                        and prior_operations is not None
+                    ):
+                        logging.info(f"Resubmitting operations for event.id={event.id}")
+                        operations = prior_operations
+                    else:
+                        try:
+                            operations = self._get_operations(event)
+                        except _ExperimentInactiveException as ex:
```
rb-determined-ai, line 148: Just detect `event.experimentInactive` outside of `self._get_operations()`.



Using exceptions like this is a bit of a hack I think.  It's like changing your `_get_operations` signature to secretly be:

```python
def _get_operations(event) -> Union[List[Operations], _ExperimentInactiveException]:
```

because it actually has additional, expected behaviors other than returning a list of operations.  But when you use exceptions for those extra behaviors, it isn't part of the typing of the function.

schemas/expconf/v0/searcher.json
```diff
@@ -80,21 +84,10 @@
...
-            ],
-            "default": null
-        },
-        "source_checkpoint_uuid": {
-            "type": [
-                "string",
-                "null"
-            ],
-            "default": null
-        },
```
rb-determined-ai, line 96: why relax these validations?

If you relax these validations here because it's not supported by custom searcher (why would that be the case, by the way?) then you'll need to add the validations to the non-custom searchers.

harness/determined/searcher/core_search_runner.py
mpkouznetsov: @aciborowska , would you mind looking at this one?

harness/determined/searcher/search_method.py
mpkouznetsov: The `bindings.v1TrialExitedEarlyExitedReason` are internal whereas `ExitedReason` values are user-facing and more user-friendly. I realize we missed `EXITED_REASON_UNSPECIFIED`, and I am going to add it. However, I think it is better to have short descriptive values like "ERRORED" as opposed to "EXITED_REASON_UNSPECIFIED".

harness/determined/searcher/search_method.py
mpkouznetsov: Then the experiment will never end, it will just stay there, supposedly "ACTIVE", without doing anything. Maybe I am missing something but master already had the concept of a shutdown operation, and we decided to use it the way, I think, it was intended.

harness/determined/searcher/search_method.py
mpkouznetsov: > Do we synthesize progress events in the master when operations are completed?

Not sure what you mean by synthesizing. `report_progress` _is_ the only way this gets called. I don't see a problem with that.

Progress operations are not user-facing (maybe I should actually rename `Progress -> _Progress`), they have been introduced by us here solely as a vehicle to convey experiment progress back to master after another trial reports progress to master and master tells the SearchRunner about it. The user should not be aware that Progress is an operation and they should not mix Progress into a list of operations returned by other event handlers.

We can discuss this.

harness/determined/searcher/search_runner.py
mpkouznetsov: hmm, it actually happens to be a binary:

harness/determined/searcher/search_method.py
mpkouznetsov: This will be addressed in https://github.com/determined-ai/determined/pull/5158 which I will merge once the build is successful.

docs/training/hyperparameter/search-methods/hp-custom.rst
```diff
@@ -0,0 +1,219 @@
...
+     smaller_is_better: true
+     unit: batches
+
+******************************
+ Search Method Implementation
+******************************
+
+Subclass :class:`~determined.searcher.SearchMethod`. Below is a starting template:
+
+.. code:: python
```
rb-determined-ai, line 60: I'm not crazy about duplicating our api definitions and documentation here in a code block.

Is there information in these comments which is different than what is in the docstrings of the actual interface?  If so, then it's either duplicate information, or the docstrings need improvement (at which point it will all be duplicated).

Also, there's no way to keep this up-to-date with the real interface.  However... the sphinx-generated docs for the real interface are already always up-to-date with the real interface.

Basically, I think the copy/paste benefit of this snippet is not worth the added complexity to our docs or our development process.

docs/training/hyperparameter/search-methods/hp-custom.rst
```diff
@@ -0,0 +1,219 @@
...
+       model_context_dir = "experiment_files"
+
+       # Path to the .yaml file with the multi-trial experiment configuration.
+       model_config = "experiment_files/config.yaml"
+
+       with det.core.init() as core_context:
+
+           info = det.get_cluster_info()
+           assert info is not None
+           args = AttrDict(info.trial.hparams)
```
rb-determined-ai, line 212: I'd prefer we leave `AttrDict` out of this, because it's not really relevant to what's happening.  Also it's coming from a package that isn't imported in this script, and adding imports to the script would further distract from the purpose of the example.

e2e_tests/tests/experiment/experiment.py
```diff
@@ -72,6 +72,84 @@ def cancel_trial(trial_id: int) -> None:
...
+                pytest.fail(
+                    f"Multiple experiments with name={experiment_name}, "
+                    f"expected only 1 experiment."
+                )
+            experiment = response[0]
+            experiment_id = experiment.id
+        # Ignore network errors while polling for experiment state to avoid a
+        # single network flake to cause a test suite failure. If the master is
+        # unreachable multiple times, this test will fail after max_wait_secs.
+        except api.errors.MasterNotFoundException:
```
rb-determined-ai, line 99: why is this an acceptable error in e2e tests?

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
+import logging
+import subprocess
+import tempfile
+import time
+from pathlib import Path
```
rb-determined-ai, line 5: [google style guide for imports](https://google.github.io/styleguide/pyguide.html#22-imports)

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
...
+from typing import List, Optional
+
+import pytest
+import yaml
+from urllib3.connectionpool import HTTPConnectionPool, MaxRetryError
```
rb-determined-ai, line 10: ditto

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
...
+
+from determined import searcher
+from determined.common.api import bindings
+from determined.common.api.bindings import determinedexperimentv1State
```
rb-determined-ai, line 14: ditto

e2e_tests/tests/fixtures/custom_searcher/core_api_searcher_random.yaml
```diff
@@ -0,0 +1,9 @@
+name: core-api-searcher-random
+entrypoint: python3 core_api_custom_searcher.py --searcher random --max-trials 5 --max-concurrent-trials 2 --max-length 500
+searcher:
+  metric: validation_error
+  smaller_is_better: true
+  name: single
+  max_length:
+    batches: 100
+max_restarts: 0
```
rb-determined-ai, line 9: need line ending

e2e_tests/tests/fixtures/custom_searcher/model_noop.py
```diff
@@ -0,0 +1,259 @@
...
+import random
+import sys
+import time
+from typing import Any, Dict, Optional, Type
+
+import numpy as np
+
+import determined as det
+from determined import layers, tensorboard, util, workload
+from determined.common import check
```
rb-determined-ai, line 16: Do not use `check` module in any new code.

It's a terrible antipattern; it raises a meaningless error type, it doesn't work with mypy, and even though it's specifically intended to always deliver usable error messages, about half of our usages of it result in actually useless error messages, like:

```python
check.true(os.path.isfile("startup-hook-ran"), "File should exists.")
```

Which will print `CheckFailed: expected true but got false, file should exist`... What file?

e2e_tests/tests/fixtures/custom_searcher/model_noop.py
```diff
@@ -0,0 +1,259 @@
...
+        )
+
+    def get_per_slot_batch_size(self) -> int:
+        return self._per_slot_batch_size
+
+    def get_global_batch_size(self) -> int:
+        return self._global_batch_size
+
+
+class NoOpTrialController(det.TrialController):
```
rb-determined-ai, line 39: This model_noop.py has way too much cruft to just copy/paste it.

The real NoOpTrialController is used in like 30 different tests to obtain like 30 different behaviors.  It's also one of two existing non-`determined` TrialController clases, which are overall going away.  All changes made to the to the TrialController classes have to get copied into the existing NoOpTrialController and the MetricMaker TrialController.  This is a pain in the butt, and copy/pasting NoOpTrialController compounds this pain.

Instead, can you write a core api training loop?  I think it will amount to far less code and it shouldn't increase our developer burden any.

e2e_tests/tests/fixtures/custom_searcher/searchers.py
```diff
@@ -0,0 +1,607 @@
+import dataclasses
+import json
+import logging
+import pickle
+import random
+import sys
+import uuid
+from pathlib import Path
```
rb-determined-ai, line 8: google style guide for python imports

e2e_tests/tests/fixtures/custom_searcher/searchers.py
```diff
@@ -0,0 +1,607 @@
...
+from typing import Dict, List, Optional, Set
+
+from urllib3.connectionpool import HTTPConnectionPool, MaxRetryError
```
rb-determined-ai, line 11: ditto

e2e_tests/tests/fixtures/custom_searcher/searchers.py
```diff
@@ -0,0 +1,607 @@
...
+        exception_points: Optional[List[str]] = None,
+    ) -> None:
+        self.max_trials = max_trials
+        self.max_concurrent_trials = max_concurrent_trials
+        self.max_length = max_length
+
+        self.test_type = test_type
+        self.exception_points = exception_points
+
+        # TODO remove created_trials and closed_trials before merging the feature branch
```
rb-determined-ai, line 83: just calling this out so it isn't forgotten

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
...
+    assert response.experiment.numTrials == 5
+    assert search_method.created_trials == 5
+    assert search_method.pending_trials == 0
+    assert search_method.closed_trials == 5
+    assert len(search_runner.state.trials_created) == search_method.created_trials
+    assert len(search_runner.state.trials_closed) == search_method.closed_trials
+
+
+@pytest.mark.e2e_cpu_2a
+@pytest.mark.parametrize(
```
rb-determined-ai, line 80: question: how long does this set of e2e tests take to run?

I like the failure testing, but I'm a bit concerned about e2e test costs (developer time more than cloud costs)

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
...
+    trials = bindings.get_GetExperimentTrials(
+        exp.determined_test_session(), experimentId=experiment.id
+    ).trials
+    for trial in trials:
+        assert trial.state == bindings.determinedexperimentv1State.STATE_COMPLETED
+        assert trial.totalBatchesProcessed == 500
+
+
+@pytest.mark.e2e_cpu_2a
+@pytest.mark.parametrize(
```
rb-determined-ai, line 236: ditto

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,580 @@
...
+    failures = logs.count("Max retries exceeded with url: http://dummyurl (Caused by None)")
+    assert failures == len(exception_points)
+
+    # check for resubmitting operations
+    resubmissions = logs.count("determined.searcher: Resubmitting operations for event.id=")
+    assert resubmissions == sum([x == "after_save" for x in exception_points])
+
+
+@pytest.mark.e2e_cpu
+@pytest.mark.parametrize(
```
rb-determined-ai, line 457: you guys went nuts on these tests :tada: 

I like that a lot, especially for a feature with complex distributed state behaviors.

e2e_tests/tests/experiment/experiment.py
mpkouznetsov: This is not new. This logic was introduced by Yoav [here](https://github.com/determined-ai/determined/pull/571). If you think this exception handling should be ripped out, we should do it across the board (and look if the number of flakes/spurious test failures increases).

e2e_tests/tests/fixtures/custom_searcher/searchers.py
mpkouznetsov: Actually, those are test classes, so I am removing the comment but keeping the attributes.

harness/determined/searcher/search_method.py
rb-determined-ai: EXITED_REASON_UNSPECIFIED probably shouldn't be part of it... that's sort of the "not present" value in protobuf, and we should probably just raise an error rather than hand the user something invalid.

e2e_tests/tests/experiment/experiment.py
rb-determined-ai: Anything based on the `Session` (which includes all of `bindings.*`) already has retries built in, after Ilia added automatic retries [here](https://github.com/determined-ai/determined/pull/4031).

I don't mind ripping out automatic retries in tests across the board.  Yoav's change was so long ago I can't remember what the context was.  Today I would probably say "we should diagnose why network failures are so common we're noticing them in our tests".

But in any case, we don't need duplicate layers of retry logic in tests using the `bindings` module.

e2e_tests/tests/fixtures/custom_searcher/model_noop.py
rb-determined-ai: Yes please!

rb-determined-ai: (I haven't read that file in years... I didn't know it said that)

e2e_tests/tests/experiment/experiment.py
```diff
@@ -72,6 +72,84 @@ def cancel_trial(trial_id: int) -> None:
...
+
+        time.sleep(1)
+
+    else:
+        pytest.fail(f"Experiment {experiment_name} did not start any trial {max_wait_secs} seconds")
+
+
+def _is_experiment_active(exp_state: determinedexperimentv1State) -> bool:
+    if (
+        exp_state == determinedexperimentv1State.STATE_ACTIVE
```
rb-determined-ai, line 143: simplify this as `if exp_state in (...):`

rb-determined-ai: oh wait... `return exp_state in (...)`, skip the `if thing_is_true(): return True` nonsense

master/pkg/searcher/search_method.go
```diff
@@ -34,11 +35,19 @@ type SearchMethod interface {
...
 		ctx context, requestID model.RequestID, exitedReason model.ExitedReason,
 	) ([]Operation, error)
+
 	// TODO: refactor as model.Snapshotter interface or something
 	model.Snapshotter
 	expconf.InUnits
 }
 
+// CustomSearchMethod is the interface for the custom search method.
+type CustomSearchMethod interface {
```
dzhu, line 45: Hang on, does this actually need to be an interface? I know I previously suggested switching this to a separate interface when it was previously just part of the `SearchMethod` interface, but do we actually need it at all?

harness/determined/searcher/_search_runner.py
```diff
@@ -0,0 +1,351 @@
...
+        session: client.Session,
+        prior_operations: Optional[List[searcher.Operation]],
+    ) -> None:
+        experiment_is_active = True
+
+        try:
+            while experiment_is_active:
+                time.sleep(
+                    2
+                )  # we don't want to call long polling API more often than every 2 seconds.
```
rb-determined-ai, line 130: > `we don't want to call long polling API more often than every 2 seconds.`

Why not?  This feels like we're making the searcher less responsive on purpose.

harness/determined/searcher/_search_runner.py
```diff
@@ -0,0 +1,351 @@
...
+                                event.experimentInactive.experimentState
+                                == bindings.determinedexperimentv1State.STATE_PAUSED
+                            ):
+                                self._show_experiment_paused_msg()
+                            else:
+                                experiment_is_active = False
+                            break
+
+                        else:
+                            operations = self._get_operations(event)
```
rb-determined-ai, line 171: small nit: since the previous if statement ends in a break, eliminate the `else` and unindent `operations = ...`

It's confusing that:
```
operations = self._get_operations(event)
```
and
```
# save state
self.state.last_event_id = event.id
self.save_state(experiment_id, operations)
```
appear at different levels of indentation, but are logically in the same block of code that always runs as a unit.

harness/determined/searcher/_search_runner.py
```diff
@@ -0,0 +1,351 @@
...
+        if event.trialClosed:
+            d["trialClosed"] = event.trialClosed.to_json()
+        if event.trialCreated:
+            d["trialCreated"] = event.trialCreated.to_json()
+        if event.initialOperations:
+            d["initialOperations"] = event.initialOperations.to_json()
+        if event.experimentInactive:
+            d["experimentInactive"] = event.experimentInactive.to_json()
+        d["id"] = event.id
+        return d
```
rb-determined-ai, line 234: is this function materially different than:

```python
return {k: v for k, v in event.to_json() if v is not None}
```

It might be nice to lean on the generated bindings here for their type safety.

rb-determined-ai (approved):
Great work, all!

harness/determined/searcher/_search_runner.py
mpkouznetsov: Wait, that would completely alter the logic!

mpkouznetsov: 👍

dzhu (approved):

PR #4425, opened 2022-06-24
Send Docker image scans to the infrastructure engineering channel instead

mackrorysd:
Just reflecting some changes to internal Slack channels


loksonarius (approved):

mackrorysd:
I've filed JIRAs for the test failures, as they are obviously flakes.

PR #4427, opened 2022-06-26
chore: webui changes for slurm

stoksc:
## Description
For Slurm, agents can be shared between resource pools. This PR fixes some of the fallout from this novelty.

## Test Plan
- [x] tested against a slot cluster, sanity checked k8s/agentRM

## Commentary (optional)
yay .ts..

## Checklist


stoksc:
Effect of changes:
Screen Shot 2022-06-26 at 9 00 41 AM

webui/react/src/pages/Cluster/ClusterOverallStats.tsx
```diff
@@ -7,8 +7,6 @@ import { useStore } from 'contexts/Store';
 import { ShirtSize } from 'themes';
 import { ResourceType } from 'types';
 
-import { maxPoolSlotCapacity } from '../Clusters/ClustersOverview';
-
```
hkang1, line 11: **question:** I noticed that `maxPoolSlotCapacity` is no longer used in the `ClusterOverallStats` and `ClusterOverallBar` display. There's two more places that reference that function.
1. `ClusterOverview` - we talked about a dedicated allocation endpoint in the future, but for now will the function `clusterStatusText` be incorrect and also double count agents/slots since it is still relying on `maxPoolSlotCapacity`?
2. `ResourcePoolCardLight` - this one might be ok since it's showing stats on a per pool basis, but should it use the new overview agents data instead also?

webui/react/src/pages/Cluster/ClusterOverallBar.tsx
```diff
@@ -26,46 +24,27 @@ export const ClusterOverallBar: React.FC = () => {
...
-          break;
-        default:
-          index = undefined;
-      }
-      if (index === undefined) return acc;
-      acc[index] += maxPoolSlotCapacity(pool);
-      return acc;
-    }, [ 0, 0 ]);
-  }, [ resourcePools ]);
-
```
hkang1, line 47: nice, so much cleaner

hkang1 (commented):
Mostly looks good, wondering if there are additional areas we should fix up using the new `resourcePools` field? Added a few comments and questions throughout.

webui/react/src/types.ts
```diff
@@ -111,7 +111,7 @@ export interface Resource {
 export interface Agent {
   id: string;
   registeredTime: number;
-  resourcePool: string;
+  resourcePools: string[];
```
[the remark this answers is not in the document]
stoksc: I think pulling the bandaid off is probably ok, so long as it isn't used anywhere, so long as when people upgrade they refresh to get the new UI?

webui/react/src/pages/Cluster/ClusterOverallStats.tsx
stoksc: 1. yeah.. I think it would be incorrect. good catch. let me give this some thought and see what I come up with.
2. yeah, I intentionally didn't update it, but after more thought I agree, it should. will update.

stoksc: I guess the change I made has issues when using dynamic agents. this may get a bit more tricky than I first thought.

stoksc: ok, so.
1. I corrected the calculation for `maxClusterSlotCapacity` for all cases and pulled in into a function and reused it for `clusterStatusText`
2. I misread this question - since it is per pool I think it is fine still.

hkang1 (approved):
LGTM, I haven't been able to test it on the UI preview side, but the code and calculations look good.

webui/react/src/types.ts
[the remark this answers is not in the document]
hkang1: yeah the only time it would break is if the cluster version that is release after this change attempts to talk to a cluster before this change, then it wouldn't work...

webui/react/src/components/NavigationTabbar.tsx
```diff
@@ -42,7 +42,7 @@ const ToolbarItem: React.FC<ToolbarItemProps> = ({ path, status, ...props }: Too
 };
 
 const NavigationTabbar: React.FC = () => {
-  const { auth, cluster: overview, ui, resourcePools, info, pinnedWorkspaces } = useStore();
+  const { auth, cluster: overview, ui, resourcePools, info, pinnedWorkspaces, agents } = useStore();
```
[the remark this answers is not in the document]
stoksc: done

PR #4428, opened 2022-06-27
docs: clarify count requirement for grid search

garrett361:
## Description

Minor word fix to clarify that providing a `count` key is mandatory for `double`, `int`, and `log` data types when performing a `grid` search.


gh-determined-ai (approved):
lgtm

PR #4432, opened 2022-06-28
chore: cleanup subtree commits

trentwatt:
## Description

`subtree push` and `pull` depend on it having access to the full commit history that its commands generate.

In the initial subtree migration, these commits were squashed, breaking `subtree push` and `pull`. 

This is a branch that recreates the subtree (with 0 changed files) according to the following process:

`git checkout master`
`git subtree split --prefix webui/react/src/shared --branch core --rejoin`

Note that there are three commits in the history for this PR. the first two are ones generated by the `split` command to reconstruct the history for the shared directory. the third is the marker commit for the adding of the subtree.

those first two are shared-specific versions of the two commits that have touched shared since shared was added back to the repo proper, so they reflect the entire history of `/shared` within the determined repo according to git.

if you look at [the history for this branch](https://github.com/trentwatt/determined/commits/subtree-cleanup), you can see the shared-specific versions of said commits alongside the actual commits to the main repo.

## Test Plan

Test that pushing and pulling from remote works:

checkout branch. this branch will be playing the role of `master`

`git checkout subtree-cleanup`

create a new branch on shared to test with

`git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web my-test-branch`

create a new local branch for making changes

`git checkout -b subtree-test`

make changes

`touch webui/react/src/shared/thing`

commit changes

`git add . && git commit -m "make changes"`

push changes to shared remote

`git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web my-test-branch`

switch back to "`master`" branch

`git checkout subtree-cleanup`

pull in changes from shared

`git subtree pull --prefix=webui/react/src/shared shared-web my-test-branch`

## Commentary

both squash merge as well as rebase merge alter the commits and their SHAs that `git subtree` relies on to keep repos in sync.

As a result, it would be necessary to push this branch directly to master, as opposed to merging the PR.


trentwatt:
pushed directly to master

PR #4434, opened 2022-06-28
fix: hide continue trial for experiments with 0 or more than 1 trial

hkang1:
## Description

The `Continue Trial` button shows up for single trial experiments when the experiment hasn't had the chance to actually start the 1 trial, so the `numTrials` end up being 0. This leads to a blank modal since there isn't any trial info available to continue off of.

Screen Shot 2022-06-28 at 9 40 27 AM

## Test Plan

- [ ] Start a single trial experiment, then cancel it before it has a chance to run the trial. Confirm that the `Continue Trial` doesn't show up for that experiment.
- [ ] Test that a single trial experiment with a trial attempt shows the `Continue Trial` option exists.
- [ ] Ensure that `Continue Trial` shows up for trials under multi-trial experiment.

## Checklist

### DIFF
--- a/webui/react/src/utils/experiment.ts
+++ b/webui/react/src/utils/experiment.ts
@@ -115,8 +115,8 @@ export const canExperimentContinueTrial = (
   experiment: ProjectExperiment,
   user?: DetailedUser,
   trial?: TrialDetails,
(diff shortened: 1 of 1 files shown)


hkang1:
Not adding a release note since it's an extension of another previous PR that addresses this area prior to a release.

PR #4436, opened 2022-06-28
feat: add spinner to show data fetching

thiagodallacqua-hpe:
## Description

Original ticket: [DET-7180]

Some of the tabs from the `ExperimentSingleTrialTabs` component were showing a "No data" message and icon while the component fetches the experiment/trial details in the background.

So, to better visualize the initial data fetching, I've bound the `trial` prop (which is `undefined` while it fetches) to a spinner conditional rendering on the affected component tabs (Overview, Hyperparameters and Logs).

## Test Plan

To test this implementation, go to any experiment overview page `/experiments/<experiment_id>/overview`, upon reaching the page, the spinner should appear to highlight data fetching, instead of the "No Data" background.

The same behavior is expected for each experiment tab on this page.

For more context, check the gifs linked below.

## Commentary (optional)

Here's some gifs to better visualize the changes:

https://user-images.githubusercontent.com/104855841/176259925-e6d8d9d1-9ae7-47a4-a5c2-0b93f1a6d689.mp4

https://user-images.githubusercontent.com/104855841/176259941-8736fe88-6920-4908-baa8-5ba36b765038.mp4

https://user-images.githubusercontent.com/104855841/176259949-9d9532a2-612f-45c4-b11f-dc17ed26a0d1.mp4

## Checklist

[DET-7180]: https://determinedai.atlassian.net/browse/DET-7180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/LogViewer/LogViewer.tsx
+++ b/webui/react/src/components/LogViewer/LogViewer.tsx
@@ -115,7 +115,7 @@ const LogViewer: React.FC<Props> = ({
   const baseRef = useRef<HTMLDivElement>(null);
   const logsRef = useRef<HTMLDivElement>(null);
   const listRef = useRef<VariableSizeList>(null);
-  const [ isFetching, setIsFectching ] = useState(false);
+  const [ isFetching, setIsFetching ] = useState(false);
   const local = useRef(clone(defaultLocal));
   const [ canceler ] = useState(new AbortController());
   const [ fetchDirection, setFetchDirection ] = useState(FetchDirection.Older);
@@ -178,7 +178,8 @@ const LogViewer: React.FC<Props> = ({
 
     const buffer: Log[] = [];
 
-    setIsFectching(true);
+    setIsFetching(true);
+    local.current.isFetching = true;
 
     await readStream(
       onFetch({ limit: PAGE_LIMIT, ...config } as FetchConfig, type),
@@ -188,7 +189,8 @@ const LogViewer: React.FC<Props> = ({
       },
     );
 
-    setIsFectching(false);
+    setIsFetching(false);
+    local.current.isFetching = false;
 
     return processLogs(buffer);
   }, [ decoder, fetchDirection, onFetch, processLogs ]);
@@ -542,16 +544,6 @@ const LogViewer: React.FC<Props> = ({
     />
   ), [ dateTimeWidth ]);
 
-  const getChildrenWhenEmpty = (): JSX.Element | undefined => {
-    // TODO: improve the usage of the prop onFetch...
-    if (isFetching) return <Spinner className={css.empty} spinning />;
-    if (logs.length === 0) return (
-      <div className={css.empty}>
-        <Message title="No logs to show." type={MessageType.Empty} />
-      </div>
-    );
-  };
-
   return (
     <Section
       bodyNoPadding
@@ -575,7 +567,17 @@ const LogViewer: React.FC<Props> = ({
               {LogViewerRow}
             </VariableSizeList>
           </div>
-          {getChildrenWhenEmpty()}
+          {
+            <Spinner className={css.empty} conditionalRender spinning={isFetching}>
+              {
+                (logs.length === 0) && (
+                  <div className={css.empty}>
+                    <Message title="No logs to show." type={MessageType.Empty} />
+                  </div>
+                )
+              }
+            </Spinner>
+          }
         </div>
         <div className={css.buttons} style={{ display: showButtons ? 'flex' : 'none' }}>
           <Tooltip placement="left" title={ARIA_LABEL_SCROLL_TO_OLDEST}>
--- a/webui/react/src/components/LogViewer/LogViewer.tsx
+++ b/webui/react/src/components/LogViewer/LogViewer.tsx
@@ -15,6 +15,7 @@ import { FetchArgs } from 'services/api-ts-sdk';
 import { readStream } from 'services/utils';
 import Icon from 'shared/components/Icon/Icon';
 import Message, { MessageType } from 'shared/components/Message';
+import Spinner from 'shared/components/Spinner';
 import { clone } from 'shared/utils/data';
 import { formatDatetime } from 'shared/utils/datetime';
 import { copyToClipboard } from 'shared/utils/dom';
@@ -114,6 +115,7 @@ const LogViewer: React.FC<Props> = ({
   const baseRef = useRef<HTMLDivElement>(null);
   const logsRef = useRef<HTMLDivElement>(null);
   const listRef = useRef<VariableSizeList>(null);
+  const [ isFetching, setIsFectching ] = useState(false);
   const local = useRef(clone(defaultLocal));
   const [ canceler ] = useState(new AbortController());
   const [ fetchDirection, setFetchDirection ] = useState(FetchDirection.Older);
@@ -176,7 +178,7 @@ const LogViewer: React.FC<Props> = ({
 
     const buffer: Log[] = [];
 
-    local.current.isFetching = true;
+    setIsFectching(true);
 
     await readStream(
       onFetch({ limit: PAGE_LIMIT, ...config } as FetchConfig, type),
@@ -186,7 +188,7 @@ const LogViewer: React.FC<Props> = ({
       },
     );
 
-    local.current.isFetching = false;
+    setIsFectching(false);
 
     return processLogs(buffer);
   }, [ decoder, fetchDirection, onFetch, processLogs ]);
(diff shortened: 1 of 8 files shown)


webui/react/src/components/LogViewer/LogViewer.tsx
```diff
@@ -176,7 +178,7 @@ const LogViewer: React.FC<Props> = ({
 
     const buffer: Log[] = [];
 
-    local.current.isFetching = true;
+    setIsFectching(true);
 
     await readStream(
       onFetch({ limit: PAGE_LIMIT, ...config } as FetchConfig, type),
```
thiagodallacqua-hpe, line 184: I think that, maybe, we have to re-write some parts of this component... the `onFetch` is `?` and, the only place that it actually isn't being passed (as far as I could find) is the `TrialDetailsLogs` component, because the `trial` is undefined while initial fetching... also, as it is, it's kind of confusing imho.

webui/react/src/components/LogViewer/LogViewer.tsx
```diff
@@ -186,7 +188,7 @@ const LogViewer: React.FC<Props> = ({
       },
     );
 
-    local.current.isFetching = false;
```
thiagodallacqua-hpe, line 189: better to use state as the ref wasn't reflect in the UI.

webui/react/src/components/ResponsiveTable.tsx
```diff
@@ -86,14 +86,17 @@ const ResponsiveTable: ResponsiveTable = ({
...
+        spinning ? <Spinner spinning={true} />
+          : (
+            <Table
+              bordered
+              scroll={tableScroll}
+              tableLayout="auto"
+              {...props}
+            />
+          )
+      }
```
trentwatt, line 99: question: why is this change needed? 

Overall the preferred pattern in our codebase is to wrap things in `<Spinner></Spinner>` the way it is done before the changes.

webui/react/src/components/LogViewer/LogViewer.tsx
trentwatt: looks like `local.current.isFetching` is still being used elsewhere, so we should probably not delete the parts where we are updating it.

probably best for now to add a `useState` to handle the render logic, but leave the `local.current.isFetching` code untouched.

webui/react/src/components/LogViewer/LogViewer.tsx
```diff
@@ -176,7 +178,7 @@ const LogViewer: React.FC<Props> = ({
 
     const buffer: Log[] = [];
 
-    local.current.isFetching = true;
+    setIsFectching(true);
```
trentwatt, line 181: `setIsFectching` -> `setIsFetching`

webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
```diff
@@ -57,18 +58,23 @@ const TrialDetailsHyperparameters: React.FC<Props> = ({ trial, pageRef }: Props)
...
+            rowClassName={defaultRowClassName({ clickable: false })}
+            rowKey="hyperparameter"
+            settings={settings as InteractiveTableSettings}
+            showSorterTooltip={false}
+            size="small"
+            updateSettings={updateSettings as UpdateSettings<InteractiveTableSettings>}
+          />
+        )
+          : <Spinner spinning />
+      }
```
trentwatt, line 77: chore: wrap in Spinner instead of using conditional JSX

webui/react/src/pages/TrialDetails/TrialDetailsLogs.tsx
```diff
@@ -171,12 +172,17 @@ const TrialDetailsLogs: React.FC<Props> = ({ experiment, trial }: Props) => {
...
+        trial ? (
+          <LogViewer
+            decoder={mapV1LogsResponse}
+            title={logFilters}
+            onDownload={handleDownloadLogs}
+            onFetch={handleFetch}
+          />
+        )
+          : <Spinner spinning />
+      }
```
trentwatt, line 185: chore: wrap in spinner instead of using conditional JSX

trentwatt (commented):
good work. some comments about keeping the style consistent with the rest of the code base- for example avoiding helper functions that return JSX.

also probably best to leave `local.current` in `LogViewer` alone for purposes of this PR.

it looks like there is a common pattern of having
```
<Spinner>
    {data ? <ChildComponent /> : <NoDataMessage>}
</Spinner>

```
as a part of [DET-7496](https://determinedai.atlassian.net/browse/DET-7496)  we are trying to discover and implement patterns that will help us to have consistent loading behavior across the pages.

based on your work, do you have any ideas about waht those patterns should be and how we could implement them.



can you think of any ways we could re

webui/react/src/pages/TrialDetails/TrialDetailsHyperparameters.tsx
```diff
@@ -57,18 +58,20 @@ const TrialDetailsHyperparameters: React.FC<Props> = ({ trial, pageRef }: Props)
...
-        dataSource={dataSource}
-        pagination={false}
-        rowClassName={defaultRowClassName({ clickable: false })}
-        rowKey="hyperparameter"
-        settings={settings as InteractiveTableSettings}
-        showSorterTooltip={false}
-        size="small"
-        updateSettings={updateSettings as UpdateSettings<InteractiveTableSettings>}
-      />
+      <Spinner conditionalRender spinning={!trial}>
```
trentwatt, line 61: The `conditionalRender` keeps the "No Data" message from showing underneath the spinner-- that is good.

But it also makes it so where the layout changes more once the data is loaded, and it is not obvious that the thing that is being loaded is a table-- less than ideal.

It would be nice if we could still have the table skeleton stay in place underneath the loading spinner, but not say "No Data". It looks like we actually have a `SkeletonTable` that is not used... might be worth looking in to.

webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -110,9 +111,15 @@ const TrialChart: React.FC<Props> = ({
   return (
     <Section bodyBorder options={options} title="Metrics">
       <div className={css.base}>
-        {chartData[0].length === 0 ?
-          <Empty description="No data to plot." image={Empty.PRESENTED_IMAGE_SIMPLE} /> :
-          <UPlotChart data={chartData} options={chartOptions} />}
+        {
+          <Spinner className={css.spinner} conditionalRender spinning={!trialId}>
```
trentwatt, line 115: similar thing, doing the conditional render makes it less obvious what the page is going to be while it is loading.

would be good to have like a skeleton chart or something we could fall back to.

webui/react/src/pages/TrialDetails/TrialDetailsLogs.tsx
```diff
@@ -171,12 +172,14 @@ const TrialDetailsLogs: React.FC<Props> = ({ experiment, trial }: Props) => {
 
   return (
     <div className={css.base}>
-      <LogViewer
-        decoder={mapV1LogsResponse}
-        title={logFilters}
-        onDownload={handleDownloadLogs}
-        onFetch={trial && handleFetch}
-      />
+      <Spinner conditionalRender spinning={!trial}>
```
trentwatt, line 175: same thing here:

i think leaving out the `conditionalRender` is better for now, because it makes it more obvious what it actually is that is loading.

webui/react/src/components/ResponsiveTable.tsx
```diff
@@ -86,7 +86,7 @@ const ResponsiveTable: ResponsiveTable = ({
 
   return (
     <div ref={tableRef}>
-      <Spinner spinning={spinning}>
+      <Spinner conditionalRender spinning={spinning}>
```
trentwatt, line 89: think it would be better to show a `SkeletonTable` type thing than to do `conditionalRender`

trentwatt (approved):
Looking Good!

I would prefer *not* doing the `conditionalRender` in most places though: it obscures what the different elements on the page will be, whereas doing just the spinner makes it obvious both what the element is, and that the element is loading (even though there is a faint "No Data" message in the background).

ideally we could have different skeleton elements that make for good fallbacks. Then passing a `loading` prop to e.g. `InteractiveTable`, `UPlotChart`, or `LogViewer` would result in the component displaying the corresponding skeleton wrapped in a spinner.

PR #4437, opened 2022-06-28
fix: default max length properly to undefined

hkang1:
## Description

PROBLEM:
For `Continue Trial`, the current form will submit and create an experiment even when the `max length` field is left empty (despite being required).

Screen Shot 2022-06-28 at 1 40 36 PM

## Test Plan

- [ ] Navigate to single trial experiment with exactly 1 trial, and click `Continue Trial`, leave max length empty and attempt to submit. The modal should show that max length is required and prevent the user from moving on and persist the modal.
- [ ] Try the same test above, but for a multi-trial experiment trial.
- [ ] Test for entering an invalid max length value.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
+++ b/webui/react/src/hooks/useModal/useModalExperimentCreate.tsx
@@ -121,6 +121,13 @@ const useModalExperimentCreate = (props: Props): ModalHooks => {
     options: { rawCancel: true },
   });
 
+  const handleFieldsChange = useCallback(() => {
+    setModalState(prev => {
+      if (prev.error) return { ...prev, error: undefined };
+      return prev;
+    });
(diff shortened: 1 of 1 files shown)


gt2345 (approved):
Works great for preventing max length being empty, but one minor thing I came cross, if first set max length to an invalid value, like 0 or negative number, after submit there is an error message and modal persists, then change max length to a valid number and submit again, error message does not go away and also not able to submit.  
Screen Shot 2022-06-28 at 2 55 13 PM

hkang1:
> h to an invalid value, like 0 or negative number, after submit there is an error message and modal persists, then change max length to a valid number and submit again, error message does not go away and also not able to submit.

Oh dang, good catch. I'll get that fixed, thanks!

PR #4443, opened 2022-06-29
fix: various workspaces fixes

EmilyBonar:
## Description

- Adding empty state to workspaces list
- Don't have autocomplete on Delete Workspace modal
- Don't show Project picker in Move Experiment modal if Uncategorized is selected
- Fix Experiment Details overflow actions
- Dismiss Move Experiment notification on click
- Remove redundant Move Experiment confirmation

## Checklist

### DIFF
--- a/webui/react/src/pages/WorkspaceList.tsx
+++ b/webui/react/src/pages/WorkspaceList.tsx
@@ -19,6 +19,7 @@ import useSettings, { UpdateSettings } from 'hooks/useSettings';
 import { paths } from 'routes/utils';
 import { getWorkspaces } from 'services/api';
 import { V1GetWorkspacesRequestSortBy } from 'services/api-ts-sdk';
+import Icon from 'shared/components/Icon';
 import Message, { MessageType } from 'shared/components/Message';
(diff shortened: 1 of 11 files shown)


webui/react/src/pages/WorkspaceList.tsx
```diff
@@ -322,12 +324,24 @@ const WorkspaceList: React.FC = () => {
...
         {workspaces.length !== 0 ? (
           workspacesList
-        ) : (
-          <Message
-            title="No workspaces matching the current filters"
-            type={MessageType.Empty}
-          />
-        )}
+        ) :
+          (workspaceFilter === WorkspaceFilters.All && settings.archived && !isLoading) ?
```
trentwatt, line 328: if we generally expect people to have the archived toggle off (it is off by default) then this wouldn't really show up.

i guess to actually know would require an additional API call or some additional data being passed in the API though

trentwatt (approved):
LGTM, 

would be nice if we can find a way to get the "Create a workspace" message to show up when user first goes to that page

PR #4451, opened 2022-06-29
fix: trim workspace and project name whitespace [DET-7390] [DET-7747]

EmilyBonar:
## Description
This PR trims whitespace from workspace and project names, as well as re-expanding long pinned workspace names after the navbar is collapsed and re-expanded [DET-7390] [DET-7747].

## Checklist

[DET-7390]: https://determinedai.atlassian.net/browse/DET-7390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DET-7747]: https://determinedai.atlassian.net/browse/DET-7747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
+++ b/webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
@@ -39,7 +39,7 @@ const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
       <div className={css.base}>
         <div>
           <label className={css.label} htmlFor="name">Name</label>
-          <Input id="name" value={name} onChange={handleNameInput} />
+          <Input id="name" maxLength={80} value={name} onChange={handleNameInput} />
         </div>
         <div>
           <label className={css.label} htmlFor="description">Description</label>
@@ -51,7 +51,11 @@ const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
 
   const handleOk = useCallback(async () => {
     try {
-      const response = await createProject({ description, name, workspaceId });
+      const response = await createProject({
+        description: description.trim(),
+        name: name.trim(),
+        workspaceId,
+      });
       routeToReactUrl(paths.projectDetails(response.id));
     } catch (e) {
       handleError(e, {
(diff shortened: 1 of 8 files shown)


webui/react/src/hooks/useModal/Workspace/useModalWorkspaceCreate.tsx
```diff
@@ -57,7 +57,7 @@ const useModalWorkspaceCreate = ({ onClose }: Props): ModalHooks => {
       closable: true,
       content: modalContent,
       icon: null,
-      okButtonProps: { disabled: name.length === 0 },
+      okButtonProps: { disabled: name.trim().length === 0 || name.trim().length > 80 },
```
hamidzr, line 60: nit: share the logic for what a valid name is between workspace create and edit

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
```diff
@@ -51,7 +51,11 @@ const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
 
   const handleOk = useCallback(async () => {
     try {
-      const response = await createProject({ description, name, workspaceId });
+      const response = await createProject({
+        description: description.trim(),
```
hamidzr, line 55: if we always want this, it could be something that `createProject` does.

hamidzr: or better probably a backend treatment but that might take longer

trentwatt (approved):
Looks fine, agree with @hamidzr about handling it a layer or two down.

Overall I woudn't see this as release blocking, because somebody has to actually go and type the trailing space, which seems unlikely.

But the sidebar fix we should definitely get in, and that looks good.

garrett361 (approved):
I also think Hamid's suggestions should be considered. Otherwise, it looks good to me.  

I verified that all leading and trailing white space is trimmed when created in the WebUI and that subsequent experiment submission to the trimmed Workspace/Project names is successful.

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
garrett361: Backend would be nice. One can still create Workspaces/Projects with leading or trailing whitespace via the CLI, but you'd really have to go out of your way to do so.

PR #4452, opened 2022-06-29
chore: add helpers for working with shared-code

hamidzr:
## Description

- fix a tester error regarding an implicit submodule update from get-deps messing with the final code
- add cli arg to set repo-hash for the destination repo
- symlink /web to /webui/react (we like this route we can the webui/react to here later)
- add helpers for working with shared code

## Checklist

### DIFF
--- a/web
+++ b/web
@@ -1,0 +1,1 @@
+webui/react
--- a/webui/react/Makefile
+++ b/webui/react/Makefile
@@ -90,4 +90,4 @@ push-shared: test-shared
 	fi
 	# fail if there are uncommitted changes.
 	git diff --exit-code || (echo "uncommitted changes, please commit or stash" ; false)
-	cd ../../; git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web main
+	cd $(git rev-parse --show-toplevel) && git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web main
--- a/webui/react/Makefile
+++ b/webui/react/Makefile
@@ -82,7 +82,12 @@ test-shared:
 	cd src/shared; ./bin/test.py --repos saas --test-local
 
 .PHONY: push
-push-shared: test-shared-code
-	# CHECK should this be allowed only in core/master branch?
-	# TODO make sure changes are committed
+push-shared: test-shared
+	# fail if the branch name is not master
+	if [ "$(git rev-parse --abbrev-ref HEAD)" != "master" ]; then \
+		echo "You must be on the master branch to push to shared-web"; \
+		false; \
+	fi
+	# fail if there are uncommitted changes.
+	git diff --exit-code || (echo "uncommitted changes, please commit or stash" ; false)
 	cd ../../; git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web main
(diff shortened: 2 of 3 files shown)


hamidzr:
@trentwatt  
> maybe drop a brief note in the README.md that this script exists?

hamidzr:
https://app.circleci.com/pipelines/github/determined-ai/determined/24082/workflows/46ac17b7-e0cb-4bb0-bd3c-78402b18923c/jobs/777334 odd test failure

hamidzr:
@hkang1 I pushed in the fix to the tester error in here

web
```diff
@@ -0,0 +1 @@
+webui/react
```
hkang1, line 1: should we just move `webui/react` over to `web`?

hkang1: 💯

webui/react/src/shared/bin/test.py
```diff
@@ -80,7 +84,7 @@ def run_webui_tests(web_dir: pathlib.Path):
     """
     runs webui tests. requires web_dir to be set up
     """
-    for target in ['get-deps', 'check', 'build', 'test']:
+    for target in ['check', 'build', 'test']:
```
hkang1, line 87: 👍

hamidzr: mainy needed because it triggers a submodule update which can unwind previous changes from the `*local*` function

web
trentwatt: yeah i would really like for that to happen as well.

only thing is just to merge `webui/Makefile` and `webui/react/Makefile`

webui/react/Makefile
```diff
@@ -74,3 +74,20 @@ test:
...
+.PHONY: push
+push-shared: test-shared
+	# fail if the branch name is not master
+	if [ "$(git rev-parse --abbrev-ref HEAD)" != "master" ]; then \
+		echo "You must be on the master branch to push to shared-web"; \
+		false; \
+	fi
+	# fail if there are uncommitted changes.
+	git diff --exit-code || (echo "uncommitted changes, please commit or stash" ; false)
+	cd $(git rev-parse --show-toplevel) && git subtree push --prefix=webui/react/src/shared https://github.com/determined-ai/shared-web main
```
trentwatt, line 93: i guess this one isn't able to make use of symlink?

trentwatt (approved):
LGTM, how do you feel about putting `push-shared` in the `.circleci/config.yml`? 

since it has `test-shared` as a prereq, I guess we need to work out the auth stuff first. other than that, do you think it's good to go though?

webui/react/Makefile
hamidzr: wdy mean? I switched an absolute `cd ../../` to  `git rev-parse...` to remove the dependency on where this folder is located.

ahh are you talking about the prefix? yeah that might break things.. I rather steer off of it and get the tester updated for now

hamidzr:
> how do you feel about putting push-shared in the .circleci/config.yml?

I don't know.. automatic pushing seems always scary

hamidzr:
for some reason the branch check cmd doesn't work through make but works through `sh -c ''` which is how I tested it.. 

this works instead 

```
	# fail if we're not on master branch
	git rev-parse --abbrev-ref HEAD | grep -qE "^master$\" || (echo "must be on master branch" ; false)
```

hamidzr:
I got some more relavant code changes here. https://github.com/determined-ai/shared-web/pull/20 . I'll try to get both of these changes added through another pr to core

hamidzr:
followup https://github.com/determined-ai/determined/pull/4504

PR #4463, opened 2022-06-30
fix: sync agent actor init on master restart [DET-7746]

ioga:
## Description

Wait until agents actor initializes agent actors and they add themselves to a resource pool to avoid a race condition in restoring tasks on master restart. 
This is a race condition which happens 100% of the time when manually cycling `det deploy aws` clusters, but not in CircleCI or local tests.

## Test Plan

1. Make a `det deploy aws` cluster
2. ssh into the cluster and set `agent_reattach_enabled: true` for both resource pools in the master config in `/usr/local/determined/etc/master.yaml`.
3. Start a jupyter notebook
4. `sudo docker restart determined-master`
5. The notebook should still be running and accessible (jupyterlab will show a disconnection message).

## Checklist

### DIFF
--- a/master/internal/resourcemanagers/agent/agent.go
+++ b/master/internal/resourcemanagers/agent/agent.go
@@ -117,7 +117,8 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
 			a.socketDisconnected(ctx)
 			// TODO(ilia): Adding restored agent here will overcount AgentStarts by maximum
 			// agentReconnectWait if it never reconnects.
-			ctx.Tell(a.resourcePool, sproto.AddAgent{Agent: ctx.Self(), Label: a.agentState.Label})
+			// Ensure RP is aware of the agent.
+			ctx.Ask(a.resourcePool, sproto.AddAgent{Agent: ctx.Self(), Label: a.agentState.Label}).Get()
 		}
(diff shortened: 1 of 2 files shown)


master/internal/resourcemanagers/agent/agent.go
```diff
@@ -117,7 +117,8 @@ func (a *agent) receive(ctx *actor.Context, msg interface{}) error {
 			a.socketDisconnected(ctx)
 			// TODO(ilia): Adding restored agent here will overcount AgentStarts by maximum
 			// agentReconnectWait if it never reconnects.
-			ctx.Tell(a.resourcePool, sproto.AddAgent{Agent: ctx.Self(), Label: a.agentState.Label})
+			// Ensure RP is aware of the agent.
+			ctx.Ask(a.resourcePool, sproto.AddAgent{Agent: ctx.Self(), Label: a.agentState.Label}).Get()
```
stoksc, line 121: There are no asks in the other direction that I can remember. Is that right?

stoksc (approved):
1 q

master/internal/resourcemanagers/agent/agent.go
ioga: There're but they should only happen once allocations start, and these should not happen until agents are initialized.

ioga:
Going to admin-merge over some react issues

PR #4467, opened 2022-07-01
fix: allocation errors on finished hp search [DET-7724]

eecsliu:
## Description
Sometimes a trial will report errored even though it has exited with code 0. This is because even if it has exited successfully, the daemon process might still need to be killed. This change make sure that we don't report an error after slaying the daemon. 

## Test Plan
Run some tensorflow or deepspeed hp search jobs with a `slots_per_trial` value greater than the slots in one node. Make sure that no trials error with exit code 0.

## Checklist

### DIFF
--- a/master/internal/task/allocation.go
+++ b/master/internal/task/allocation.go
@@ -88,6 +88,7 @@ type (
 	AllocationExited struct {
 		// userRequestedStop is when a container unexpectedly exits with 0.
 		UserRequestedStop bool
+		DaemonsKilled     bool
 		Err               error
 		FinalState        AllocationState
 	}
@@ -688,7 +689,7 @@ func (a *Allocation) Exit(ctx *actor.Context, reason string) (exited bool) {
 		return true
 	case a.allNonDaemonsExited():
 		a.killedDaemons = true
-		if strings.Contains(reason, normalLogSubstr) {
+		if a.exitWithoutErr() {
 			a.killedDaemonsNominal = true
 		}
 		a.kill(ctx, reason)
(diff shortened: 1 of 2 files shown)


ioga (approved):

master/internal/task/allocation.go
```diff
@@ -734,6 +740,16 @@ func (a *Allocation) allNonDaemonsExited() bool {
 	return true
 }
 
+func (a *Allocation) exitWithoutErr() bool {
```
ioga, line 743:
```suggestion
func (a *Allocation) exitedWithoutErr() bool {
```

master/internal/task/allocation.go
```diff
@@ -86,6 +88,7 @@ type (
 	AllocationExited struct {
 		// userRequestedStop is when a container unexpectedly exits with 0.
 		UserRequestedStop bool
+		DaemonsKilled     bool
```
ioga, line 91: so that's unused?

master/internal/task/allocation.go
```diff
@@ -49,6 +49,8 @@ type (
 		killedWhileRunning bool
 		// Marks that the trial exited successfully, but we killed some daemon containers.
 		killedDaemons bool
+		// Marks that we killed some daemon containers but after a nonzero exit.
```
ioga, line 52: hmmm these comments look odd
e.g. `killedDaemons` doesn't mean the "trial exited successfully"

and `killedDaemonsNominal` looks like a zero exit, not a "nonzero".

ioga: also maybe rename `killedDaemonsNominal` -> `killedDaemonsGracefully` or `killedDaemonsWithoutErr`? 
(I would struggle figuring out the meaning of the word "nominal" in such context)

eecsliu: nonzero is a typo - should be zero. Will change to `gracefully`

master/internal/task/allocation.go
eecsliu: forgot to delete this - is vestigial from removing stuff from trial.

PR #4378, opened 2022-06-17
style: theme tuning [DET-7362, DET-7363, DET-7364, DET-7365, DET-7366, DET-7383, DET-7425, DET-7498, DET-7551]

hkang1:
## Description

Fix list of [recorded list of issues](https://determinedai.atlassian.net/browse/DET-7358).

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

PR #4468, opened 2022-07-01
refactor: update modals to better support contexts [DET-6297]

hkang1:
## Description

Key areas of improvement:
- Centralize `contextHolder` generation to one place (`useModal.tsx`) and inject at the hook usage areas.
- Avoid calling modal hooks within other modal hooks, instead use event handlers to call the next modal hook in the parent component. The nested modal hooks can cause the context to be lost when the top-level modal hook is dismissed. This also makes the modal hooks easier to manage and test.
- Update `useModal` to prevent re-render loops.
- Move some modals around and rename them to be consistent.
- Extract reusable model and model version delete modal hooks.

## Checklist


hkang1:
@EmilyBonar @johnkim-det @julian-determined-ai Please take a look at the following two commits in this PR regarding `contextHolder` and let me know if that makes sense:
- [refactor: update base useModal hook to auto generate contextHolder](https://github.com/determined-ai/determined/pull/4468/commits/21fafa9a786b0b0105421cf356843424cb8e3c91)
- [refactor: update modal hooks to pass contextHolder through](https://github.com/determined-ai/determined/pull/4468/commits/39ddc6435d180820a7c60d8465e0682ac498fafe)

webui/react/src/hooks/useModal/useModal.tsx
```diff
@@ -51,18 +51,28 @@ const useModal = (config: {
...
+
   const modalOpen = useCallback((props: ModalFuncProps = {}) => {
     setModalProps(props);
   }, []);
 
-  const modalClose = useCallback((reason?: ModalCloseReason) => {
-    if (!modalRef.current) return;
-    modalRef.current.destroy();
-    modalRef.current = undefined;
-    if (reason) config.onClose?.(reason);
```
hkang1, line 61: if a reason is not provided, the `onClose` event handler that is passed in doesn't get called

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
```diff
@@ -131,7 +92,7 @@ const useRegisterCheckpointModal = (onClose?: (checkpointUuid?: string) => void)
         modelName: selectedModelName,
       });
       if (!response) return;
-      closeModal();
+      modalClose();
```
hamidzr: not really helpful but +1 for closeModal => less chars to type to filter down the list of suggestions in a linear var name matching situation

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
```diff
@@ -152,10 +113,10 @@ const useRegisterCheckpointModal = (onClose?: (checkpointUuid?: string) => void)
         type: ErrorType.Api,
       });
     }
-  }, [ closeModal, selectedModelNumVersions ]);
+  }, [ modalClose, selectedModelNumVersions ]);
 
   const handleOk = useCallback(async (state: ModalState) => {
-    if (!modalRef.current) return Promise.reject();
+    // if (!modalRef.current) return Promise.reject();
```
hamidzr, line 120: .

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
```diff
@@ -188,12 +145,46 @@ const useRegisterCheckpointModal = (onClose?: (checkpointUuid?: string) => void)
...
+          V1GetModelsRequestSortBy,
+          V1GetModelsRequestSortBy.LASTUPDATEDTIME,
+        ),
+      }, { signal: canceler.signal });
+      setModels(prev => {
+        if (isEqual(prev, response.models)) return prev;
+        return response.models;
+      });
+    } catch(e) {
+      handleError(e, {
```
hamidzr, line 168: I've had this question with useModal and antd modals. do we want these crucial calls to actually throw instead? eg
add the information you have to the error and [augment it](https://github.com/determined-ai/saas/blob/49e3a3d9713c2a222f0db857622f8d92521748f2/web/src/utils/error.ts#L20) and let the modal open call throw. either useModal or antd should then handle that as a critical failure

hamidzr: also not the goal of this pr so not blocking

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
```diff
@@ -283,30 +276,26 @@ const useRegisterCheckpointModal = (onClose?: (checkpointUuid?: string) => void)
...
+    return () => canceler.abort();
+  }, [ canceler ]);
 
-  // When the component using the hook unmounts, remove the modal automatically.
+  /**
+   * When modal props changes are detected, such as modal content
+   * title, and buttons, update the modal.
+   */
   useEffect(() => {
-    return () => {
```
hamidzr, line 302: not needed anymore with the context shift? or is this to change the behavior

webui/react/src/pages/ModelRegistry.tsx
```diff
@@ -500,40 +500,40 @@ const ModelRegistry: React.FC = () => {
...
-              <Icon name="model" size="mega" />
-            </div>
-            <h4>No Models Registered</h4>
-            <p className={css.description}>
-              Track important checkpoints and versions from your experiments.&nbsp;
-              <Link external path={paths.docs('/post-training/model-registry.html')}>
-                Learn more
-              </Link>
-            </p>
+      {(models.length === 0 && !isLoading && filterCount === 0) ? (
```
hamidzr, line 486: nit avoid to avoid introducing the changes

webui/react/src/hooks/useModal/UserSettings/useModalNameChange.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  );
+  const user = userEvent.setup();
+  await user.click(await screen.findByText(OPEN_MODAL_TEXT));
+  await user.click(await screen.findByText(NAME_CHANGE_TEXT));
+
+  return { user, view };
+};
+
+describe('useModalNameChange', () => {
+  it('opens modal with correct values', async () => {
```
hamidzr, line 81: it would have been valuable if we could get jest to use a different testing engine in our local machines for testing the tests. I've written tests before without having checked that they fail. if you can see tests running (slowed down and in a browser, while you see the description "it should ...." you can can some of those cases.

webui/react/src/pages/ModelDetails.tsx
```diff
@@ -299,25 +291,15 @@ const ModelDetails: React.FC = () => {
     }
   }, [ model?.model.archived, modelName ]);
 
-  const deleteCurrentModel = useCallback(() => {
-    deleteModel({ modelName });
```
hamidzr, line 303: 👍🏼

webui/react/src/pages/ModelDetails/ModelVersionActionDropdown.tsx
```diff
@@ -102,7 +67,8 @@ const ModelVersionActionDropdown: React.FC<Props> = (
         onVisibleChange={onVisibleChange}>
         {children}
       </Dropdown>
-      {contextHolder}
+      {modalModelDownloadContextHolder}
```
hamidzr, line 70: do we want these to be at the parent level since we could have many action dropdowns rendering in one page at once?

hamidzr (commented):
this looks good. I'll do a visual review as well. so far I went commit by commit so I might have commented on some outdated changes..

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.tsx
```diff
@@ -108,61 +127,71 @@ const useModalCheckpoint = ({ checkpoint, config, title, ...props }: Props): Mod
...
+          {renderRow('Total Size', totalSize)}
+          {resources.length !== 0 && renderRow(
+            'Resources', (
+              <div className={css.resources}>
+                {resources.map(resource => renderResource(resource.name, resource.size))}
+              </div>
+            ),
+          )}
+        </div>
+      </>
```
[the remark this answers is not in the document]
hkang1: oh great catch, I think I removed an element or something and then forgot to remove the fragment

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
hkang1: now that it is composing from `useModal`, this behavior is inherited from there.

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
hkang1: I may need to chat with you about this async to get a better idea of this approach.

webui/react/src/pages/ModelDetails/ModelVersionActionDropdown.tsx
hkang1: https://github.com/determined-ai/determined/pull/4468/commits/d85716d3df77a0c6e12ba99dd91e9496f71a06bb

webui/react/src/pages/ModelRegistry.tsx
hkang1: I'll revert this. I'm hoping to find some more JSX eslint rules to auto arrange these to have less variance.

hamidzr:
in the jupyterlab modal switching to advanced mode causes the modal to flash white (on black theme) momentarily. not present in latest master

hamidzr:
leaving the modal by clicking the bg in advanced mode (jupyterlab) then navigating in the sidebar the modal pops up seemingly randomly

hamidzr:
after I've cancelled a delete model modal (you can't delete that model only admin..) next few times I try deleting no modal shows up

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
hamidzr: I can't tell how much work that is but feel free to leave out.

hamidzr:
these were tested on `92c141ec8b29b6bbf6df4698fdd18c0eeda2fc57`

hkang1:
> in the jupyterlab modal switching to advanced mode causes the modal to flash white (on black theme) momentarily. not present in latest master

This one should be addressed in the [theme-tuning PR](https://github.com/determined-ai/determined/pull/4378)

hkang1:
> leaving the modal by clicking the bg in advanced mode (jupyterlab) then navigating in the sidebar the modal pops up seemingly randomly

Great find, especially isolating it to the bg in advance mode case (so specific and yet you were still able to figure out the repro)!
Here's the [fix](https://github.com/determined-ai/determined/pull/4468/commits/9b761718660d937aa4774003f701f25594647110)

hkang1:
> after I've cancelled a delete model modal (you can't delete that model only admin..) next few times I try deleting no modal shows up

Another great find! The problem was that the default "you can't delete this item, only admin..." modal props was being passed in as the same reference, so it was not triggering a new modal (which relies on modal props reference changes).
Here's the [fix](https://github.com/determined-ai/determined/pull/4468/commits/3d95738e1850d0f367b07e9388ff0e78a44f1925).

hamidzr (approved):
lgtm. great work. thanks for splitting the fixes and the commits nicely

hkang1:
> lgtm. great work. thanks for splitting the fixes and the commits nicely

thanks for the thorough review as always!

PR #4470, opened 2022-07-01
fix: disable pointer events on log viewer spinner block

hkang1:
## Description

Follow up to [this PR](https://github.com/determined-ai/determined/pull/4436).

PROBLEM:
For cluster master logs, after the logs load the spinner container remains as an absolutely positioned element that sits on top of the logs interface, blocking any user interactions.

SOLUTION:
Enable spinner container to pass through pointer events for log viewer.

## Checklist


thiagodallacqua-hpe (approved):
I thought that I had changed that, but thanks for catching that 😅

PR #4476, opened 2022-07-05
fix: notebook ignores --template CLI param and notebooks still launch if --preview param is set [DET-7632]

NicholasBlaskey:
## Description

Notebook now correctly uses the ```--template``` CLI parameter. In addition notebooks now won't launch if ```--preview``` is passed in.

## Test Plan

Create the following template with the file below
```
environment:
  environment_variables:
    - "TEST=true"
```

Using the command
```
det tpl set test <path to above.yaml>
```

Then start a notebook with the option
```
det notebook start --template test
```

Then in the notebook run this in a cell and verify that the environment variable set above is actually set
```
import os
os.environ["TEST"]
```

## Checklist


eecsliu (approved):
lgtm!

PR #4478, opened 2022-07-05
feat: Add two endpoints (POST and GET) in the backend to post Searcherops and receive Searcherevents [DET-7715]

nrajanee:
## Description

feat: Add two endpoints (POST and GET) in the backend to post Searcherops and receive Searcherevents [DET-7715] 

## Test Plan

I tested this using bindings.py to ensure that you can call these API endpoints. I've added e2e_tests for this.  

[DET-7715]: https://determinedai.atlassian.net/browse/DET-7715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


proto/src/determined/api/v1/experiment.proto
```diff
@@ -575,3 +575,35 @@ message MoveExperimentRequest {
...
+  // to master which sent them successfully to the experiment actor. 
+  string last_triggering_event = 3; 
+}
+
+// Request for sending operations from a custom user search method.
+message PostSearcherOperationsRequest {
+  // The experiment id
+  int32 experiment_id = 1;
+  // list of operations in the method.  
+  repeated determined.experiment.v1.SearcherOperation search_operations = 2;
```
mpkouznetsov, line 602:
```suggestion
  repeated determined.experiment.v1.SearcherOperation searcher_operations = 2;
```

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -28,9 +28,49 @@ message SearcherOperation {
   oneof union {
     // ValidateAfter is issued to tell a trial to train some then validate.
     ValidateAfterOperation validate_after = 1;
+    // CreateTrial is issued to create trial. TODO: Might need to create a
```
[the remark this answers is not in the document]
nrajanee: @dzhu mentioned it's better to make these a message type similar to ValidateAfterOperation (even if it's empty) to keep it consistent. 
Sep Q for liam, shouldn't creating a trial come with a hyperparameter set?  -> will post in same convo above.

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -0,0 +1,29 @@
...
+
+@pytest.mark.e2e_cpu
+def test_post_searcher_ops() -> None:
+    exp_id = exp.run_basic_test(
+        conf.fixtures_path("no_op/single.yaml"), conf.fixtures_path("no_op"), 1
+    )
+    sess = exp.determined_test_session()
+    op1 = bindings.v1SearcherOperation(
+        createTrial="createTrial"
+    )  # what do we need to include for an operation such as createTrial?
```
mpkouznetsov, line 25: I would assume you need to include everything to instantiate a master-side operation: https://github.com/determined-ai/determined/blob/97d3e6029eed98cc4ab073097753df5eff32a693/master/pkg/searcher/operations.go#L106

mpkouznetsov: I see that Liam answered your question: only Hparams need to be communicated by the client; the rest is available on master.

nrajanee: Yes making that change! Thanks Maksim.

mpkouznetsov (approved):
Let's go with this as a starting point to unblock the client side

proto/src/determined/api/v1/experiment.proto
```diff
@@ -575,3 +575,34 @@ message MoveExperimentRequest {
...
+  int32 experiment_id = 1;
+}
+
+// Response to GetSearcherEventsRequest.
+message GetSearcherEventsResponse {
+  // The requested list of searcher events.
+  // str value can be: InitialOperations, TrialCreated, ValidationCreated,
+  // TrialExitedEarly, TrialClosed appended with a counter id to represent
+  // order.
+  repeated string searcher_event = 1;
```
dzhu, line 591: Doesn't this need to be a message type with other fields in addition to the event type?

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -22,12 +22,64 @@ message CompleteValidateAfterOperation {
...
+    // Constant hyperparameter.
+    ConstantHyperparameter constant_hyperparam = 1;
+  }
+}
+
+// Create a trial with given hyperparameters.
+message CreateTrialOperation {
+  // The key refers to which part of the model this hyperparameters are for.
+  // An example in this map would be: {"optimizer": {"learning_rate": 0.01}}
+  map<string, Hyperparameter> hyperparams = 1;
```
dzhu, line 45: We support nested hyperparameters, which this doesn't cover. I suppose that can wait, though. But also the comment above seems to indicate a nested hyperparameter.

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -22,12 +22,64 @@ message CompleteValidateAfterOperation {
...
 message SearcherOperation {
   // A searcher operation is one of the following operations.
   oneof union {
     // ValidateAfter is issued to tell a trial to train some then validate.
     ValidateAfterOperation validate_after = 1;
+    // CreateTrial is issued to create trial.
+    CreateTrialOperation create_trial = 2;
+    // CloseTrial is issued to close trial.
+    CloseTrialOperation close_trial = 3;
+    // GetProgressOperation is issued to get the progress of the custom searcher
```
dzhu, line 78: These are messages that go from the custom searcher to the master, so I feel like this one doesn't belong here. Unless it's supposed to be a _report_ of progress, in which case the message type should contain a progress value.

proto/src/determined/api/v1/experiment.proto
```diff
@@ -575,3 +575,34 @@ message MoveExperimentRequest {
...
+  // The last event that triggered client which sent the appropriate events
+  // to master which sent them successfully to the experiment actor.
+  string last_triggering_event = 3;
```
mpkouznetsov, line 594: should this be = 2?

nrajanee: yes 2 thank you

proto/src/determined/experiment/v1/searcher.proto
nrajanee: Yup will add this later. Will remove the comment.

PR #4484, opened 2022-07-06
docs: restructure content

gh-determined-ai:
## Checklist


gh-determined-ai:
@garrett361 
- As far as landing page cards, I think 6 is about the right number. They cover the basic learning modes, usage patterns, and main topics.
- Card titles almost universally just say what it is, although, I do like the imperative mood in general and use it for doc titles once you get into the docs. (When I get there, I will.)
- Model Hub had its own section previously because it doesn't seem to belong anywhere. At the last meeting someone said put it with Examples, which kind of made sense. Low priority and we can always change it if it becomes a real issue.
- Agree on Model Hub API Reference suggestion.
- Good catch on Basic Workflow link. That means I need to move the page to after Load Model Data.

gh-determined-ai:
The **Reproducibility** section looks like a combination of architectural principles and best practices, which shouldn't be in the same document. **Reproducibility** -> **Coding Guidelines** & -> **Deterministic Floating Point on CPUs** looks like a best practices.

gh-determined-ai:
@shiyuann @garrett361 Implemented most of the recommended changes. For the Training->Best Practices collective, I moved most of those topics into the **Introduction to Distributed Training** section, because they were all at about the same superficial level. That's a good, centralized location from which to restructure the Training content; a good step forward in identifying the doc issues.
The updated Netlify location is: https://fastidious-blini-c64f40.netlify.app/index.html

shiyuann:
The "Load Model Data" page includes three options for where data should be stored and how data should be downloaded for different data storage but not loading data into training. Loading data into training is part of the Training APIs. So the title "Load Model Data" cannot cover what the contents are. This page is similar to "Prepare Container Environment" in that it is a "how to" and works for all training APIs. The title should be renamed as "Prepare Data", or a better name that can cover its contents.

garrett361:
I would suggest re-ordering some of the Model Developer Guide subsections to put the things that users care most about first. A suggested order:

1. Introduction to Distributed Training
2. Training API Guides
3. Prepare Data
4. Submit Experiment
5. Hyperparameter Tuning
6. Model Management
7. How to Debug Models
8. Prepare Container Environment
9. Best Practices

I also think Workspace and Projects would be more relevant to the Model Developer than the Administrator. It could go after number 7 above, say.

garrett361:
Model Hub still doesn't have a very natural home in the side bar. I'm not sure we really would want people to get started there, but it's also not quite a Tool or Integration.

I am also generally still left with the impression that there is a big disparity between the importance of the various links we display in the side bar.  

For instance, in the Model Developer Guide section, knowing how to Prepare Data, understanding a Training API and how to Submit Experiment are absolutely crucial to getting anything done, whereas all of the other links you could get by without knowing about for a while (though we do still want to highlight Hyperparameter Tuning, since it's one of our points of strength).  It would seem cleaner and more useful to me if we only displayed section numbers 1-5 from my previous comment, plus one more link which captures everything else.

I'm not as familiar with the relative importances of the sections under the Administrative Guide and whether they might benefit from a similar re-structuring.

garrett361:
Do you think it would be useful to split the current Reference section into two different sections: API Reference and Configuration Reference? I would find that helpful and think it's a natural and useful distinction to make.

I think that might also help surface some of the important reference pages which are also currently a little buried. For instance, I'm always having to reference the Experiment Configuration page, but it's currently a little hidden under Training Reference.  Similarly, I also imagine that most Model Developers will be frequently referencing the API Reference pages  (Core/PyTorch/etc) which are currently under Training Reference, in which case it would be more helpful to surface them.

gh-determined-ai:
@garrett361 The reference section was previously organized the way you suggested but someone wanted experiment config included with training.
Bookmarks might address your 2nd, surfacing point.

gh-determined-ai:
Latest site available for preview at: [https://dancing-taiyaki-b94c7f.netlify.app/](https://dancing-taiyaki-b94c7f.netlify.app/)

shiyuann (approved):

shiyuann:
Approved it. Please merge after addressing the redirecting links.

PR #4485, opened 2022-07-06
feat: environment_variables in task_container_defaults [DET-7638]

NicholasBlaskey:
## Description

Environment variables are now in task_container_defaults.

## Test Plan

Setup master config with task_container_defaults
```yaml
        task_container_defaults:
          environment_variables:
            - test_a=task_container_defaults
            - test_b=task_container_defaults
            - test_c=task_container_defaults
```

Then create a file named template.yaml
```yaml
environment:
  environment_variables:
    - test_b=template
    - test_c=template
```

Create the template
```
det tpl set tpl ../template.yaml
```

Run this command
```bash
det cmd run --template=test --config="environment.environment_variables=['test_c=expconf']" "printenv | grep test"
```

and verify the output is
```
[2022-07-06T16:30:29.240274Z] f8065628 || test_a=task_container_defaults
[2022-07-06T16:30:29.240279Z] f8065628 || test_b=template
[2022-07-06T16:30:29.240280Z] f8065628 || test_c=expconf
```

Also since experiments appear to take a different code path
```bash
det e create --template=test --config="environment.environment_variables=['test_c=expconf']" examples/tutorials/mnist_pytorch/const.yaml examples/tutorials/mnist_pytorch
```

Find ID with ```docker ps```
```bash
docker exec -it 1bdcbd393121 printenv | grep test
```

and should see same output as above

## Checklist

### DIFF
--- a/docs/release-notes/task-container-env-defaults.txt
+++ b/docs/release-notes/task-container-env-defaults.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**New Features**
+
+-  Configuration: Add support for ``task_container_defaults.environment_variables`` in master
+   config.
--- a/docs/release-notes/task-container-env-defaults.txt
+++ b/docs/release-notes/task-container-env-defaults.txt
@@ -3,4 +3,5 @@
 **New Features**
 
 -  Configuration: Add support for ``task_container_defaults.environment_variables`` in master
-   config.
+   config. Users can now specify a list of environment variables that will be set in all task
+   containers.
--- a/docs/sysadmin-basics/cluster-config.txt
+++ b/docs/sysadmin-basics/cluster-config.txt
@@ -241,7 +241,7 @@ The master supports the following configuration settings:
       -  ``determinedai/environments:rocm-4.2-pytorch-1.9-tf-2.5-rocm-0.18.2`` for ROCm.
       -  ``determinedai/environments:py-3.8-pytorch-1.10-lightning-1.5-tf-2.8-cpu-0.18.2`` for CPUs.
 
-   -  ``environment_variables``: A list of environment variables that will be set in every trial
+   -  ``environment_variables``: A list of environment variables that will be set in every task
       container. Each element of the list should be a string of the form ``NAME=VALUE``. See
       :ref:`environment-variables` for more details. Environment variables specified in experiment
       config will override default values specified here. Users can customize environment variables
--- a/docs/sysadmin-basics/cluster-config.txt
+++ b/docs/sysadmin-basics/cluster-config.txt
@@ -241,6 +241,13 @@ The master supports the following configuration settings:
       -  ``determinedai/environments:rocm-4.2-pytorch-1.9-tf-2.5-rocm-0.18.2`` for ROCm.
       -  ``determinedai/environments:py-3.8-pytorch-1.10-lightning-1.5-tf-2.8-cpu-0.18.2`` for CPUs.
 
+   -  ``environment_variables``: A list of environment variables that will be set in every trial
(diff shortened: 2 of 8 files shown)


docs/release-notes/task-container-env-defaults.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**New Features**
+
+-  Configuration: Add support for ``task_container_defaults.environment_variables`` in master
+   config.
```
ioga, line 6:
```suggestion
-  Configuration: Add support for ``task_container_defaults.environment_variables`` in master
   config. Users can now specify a list of environment variables that will be set in all task containers.
```

docs/sysadmin-basics/cluster-config.txt
```diff
@@ -241,6 +241,13 @@ The master supports the following configuration settings:
       -  ``determinedai/environments:rocm-4.2-pytorch-1.9-tf-2.5-rocm-0.18.2`` for ROCm.
       -  ``determinedai/environments:py-3.8-pytorch-1.10-lightning-1.5-tf-2.8-cpu-0.18.2`` for CPUs.
 
+   -  ``environment_variables``: A list of environment variables that will be set in every trial
```
ioga, line 244:
```suggestion
   -  ``environment_variables``: A list of environment variables that will be set in every task
```

master/pkg/model/experiment_config_test.go
```diff
@@ -116,6 +116,22 @@ func TestOverrideMasterConfigRegistryAuth(t *testing.T) {
...
+			CPU: []string{"a=from_master", "b=from_master"},
+		},
+	}
+	actual := DefaultEnvConfig(masterDefault)
+	assert.NilError(t, json.Unmarshal([]byte(`{
+    "environment_variables": ["a=from_exp", "c=from_master"]
+}`), &actual))
+	assert.DeepEqual(t, actual.EnvironmentVariables.CPU, []string{
+		"a=from_master", "b=from_master",
+		"a=from_exp", "c=from_master", // Exp config overwriters master config by being later.
```
ioga, line 131: can you remind me, where does this happen? not in our code, but at runtime by the shell of the task container itself?

ioga (approved):
lgtm

ioga:
by the way,

>Find ID with docker ps
>docker exec -it 1bdcbd393121 printenv | grep test

you could've run `printenv` in experiment's `startup-hook.sh` instead, then you wouldn't need to exec.

master/pkg/model/experiment_config_test.go
NicholasBlaskey: Not sure if I fully understand the question but

In k8s it is our code converting the env list to a map taking precedence to later elements

In docker we pass the full list unprocessed into the docker runtime 

https://github.com/opencontainers/runtime-spec/blob/main/config.md#process

which states the env array has the same semantics as this document

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_01

Which technically says this behavior is undefined "There is no meaning associated with the order of strings in the environment. If more than one string in an environment of a process has the same name, the consequences are undefined."

Tried digging through Moby code to see if I could find where this behavior is defined but couldn't find it but I imagine somewhere it has this latest overrides earlier behavior

I don't know what do you think should we update the code to only pass in the latest variables to Docker?

ioga: no worries, it's okay, I was just trying to remember if we do anything here or it's all external systems.

ioga:
lgtm

PR #4488, opened 2022-07-06
feat: shiftvit port example

garrett361:
## Description

Porting exercise sanity checking. Not planning to merge.

## Checklist


examples/computer_vision/shiftvit_pytorch/README.md
```diff
@@ -0,0 +1,66 @@
...
+| `const_cifar10.yaml`       | 81.7         |
+| `distributed_cifar10.yaml` | 84.4         |
+
+### ImageNet Timing
+
+The graphic below demonstrates the timing results for training the ShiftVit model on ImageNet data
+using 1, 2, 4, and 8 GPUs.  Each model was trained on a fixed number of records and the speed-up
+due to distributed training is evident. The durations are measured relative to the beginning of
+of each experiment.
+![ShiftVit Timing Tests](ShiftVitTimingTest.png)
```
liamcli, line 66: is this showing how much time it takes to train on a given number of records for 1, 2, 4, and 8 GPUs? if so, i expect the curves to be more linear than power law scaling.

examples/computer_vision/shiftvit_pytorch/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Files
+
+### Python
+
+* `model_def.py`: core code containing the Determined AI `PyTorchTrial` subclass.
+* `data.py`: handles the
+[MNIST](http://pytorch.org/vision/main/generated/torchvision.datasets.MNIST.html#torchvision.datasets.MNIST),
+[CIFAR10](http://pytorch.org/vision/main/generated/torchvision.datasets.CIFAR10.html),
+and [ImageNet](https://www.image-net.org) datasets.
+* `generate_blob_list.py`: scripts to pre-process GCS blobs (increases ImageNet speed)
```
liamcli, line 15: why can't the blob list be gathered and directly stored in memory at dataset creation instead of saving it to a file and then rereading it back in when creating the dataset?

examples/computer_vision/shiftvit_pytorch/model_def.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # Use timm's create_xxx factories for the optimizer and scheduler.
+        optimizer = create_optimizer(self.optimizer_config, self.model)
+        self.optimizer = self.context.wrap_optimizer(optimizer)
+
+        # timm's scheduler expects to be stepped at the end of each epoch and to be passed the
+        # epoch_idx as an arg. Passing epoch_idx requires using MANUAL_STEP mode.
+        scheduler, _ = create_scheduler(self.scheduler_config, self.optimizer)
+        self.scheduler = self.context.wrap_lr_scheduler(
+            scheduler, step_mode=LRScheduler.StepMode.MANUAL_STEP
```
liamcli, line 58: is there a reason we are using manual step instead of step every epoch for the lr_scheduler?

examples/computer_vision/shiftvit_pytorch/README.md
garrett361: It's apparently much faster this way (per Corey), as getting the blob-list itself takes a while. So, just doing it once and storing the result is a helpful pre-processing step if you're going to use ImageNet more than once.

examples/computer_vision/shiftvit_pytorch/README.md
garrett361: Correct, and this was something I wanted to ask you about.

The plot shows the timing from the very start of the trial, so the initial portion of the curve shows delays from initializing everything, but it looks like the curves all turn linear(-ish) by the end.  These tests don't really run for all that long, so I was wondering if you're seeing like a caching-effect in the middle of the curve.  I could run the tests for longer, if you think it's worth it?

examples/computer_vision/shiftvit_pytorch/model_def.py
garrett361: I'm using `timm`'s LR scheduler (as the ShiftVit paper does) which requires you to pass in the `epoch_idx` as an argument to `.step()`, and we only support such argument passing in `MANUAL_STEP` mode.

This is also why I was waiting to merge in the `epoch_idx` sync PR first; previously the `epoch_idx`'s became out of sync across workers.

garrett361:
Closing as this was never intended to be actually merged. Used to learn the typical workflow.

PR #4493, opened 2022-07-07
fix: give a reasonable name to model def download

rb-determined-ai:
A tarball named "4" is not a reasonable name.


mapmeld (approved):
:+1:

PR #4494, opened 2022-07-07
chore: test improvement for _pytorch_context and _pytorch_trial [DET-7761] [DET-7763]

drh-determined-ai:
## Description

Improve/add automated tests of code in and related to `_pytorch_context.py` and `_pytorch_trial.py`, especially that which utilizes GPUs, Apex, and/or AMP; also especially with concern for expanding code coverage. [DET-7761] [DET-7763]

## Test Plan

Automated tests should pass, of course.

Code coverage is expected to improve overall, albeit modestly. Some code has been removed or marked to be skipped in coverage reports.

## Commentary (optional)

There is much more work to be done on these parts of the code, but this particular PR is being closed because it has become stale.

## Checklist

[DET-7761]: https://determinedai.atlassian.net/browse/DET-7761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-7763]: https://determinedai.atlassian.net/browse/DET-7763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -63,7 +63,7 @@ def read_git_metadata(model_def_path: pathlib.Path) -> Tuple[str, str, str, str]
     """
     try:
         from git import Repo
-    except ImportError as e:
+    except ImportError as e:  # pragma: no cover
         print("Error: Please verify that git is installed correctly: {}".format(e))
         sys.exit(1)
 
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -639,7 +639,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
                         gradient=gradient, retain_graph=retain_graph, create_graph=create_graph
                     )
 
-    def _average_gradients(self, parameters: Any, divisor: int) -> None:
+    @staticmethod
+    def _average_gradients(parameters: Any, divisor: int) -> None:
         check.gt_eq(divisor, 1)
         if divisor == 1:
             return
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -14,7 +14,8 @@ from determined.tensorboard import get_base_path
 # Apex is included only for GPU trials.
 try:
     import apex
-except ImportError:
+except ImportError:  # pragma: no cover
+    apex = None
     if torch.cuda.is_available():
         logging.warning("Failed to import apex.")
     pass
@@ -24,7 +25,8 @@ try:
     import torch.cuda.amp as amp
 
     amp_import_error = False
-except ImportError:
+except ImportError:  # pragma: no cover
+    amp = None
     amp_import_error = True
     if torch.cuda.is_available():
         logging.warning("PyTorch AMP is unavailable.")
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -22,6 +22,7 @@ except ImportError:  # pragma: no cover
 # AMP is only available in PyTorch 1.6+
 try:
     import torch.cuda.amp as amp
+
     HAVE_AMP = True
 except ImportError:  # pragma: no cover
     HAVE_AMP = False
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -76,9 +76,9 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
         # by torch DDP and apex to initialize in the correct order
         self._wrapped_models = {}  # type: Dict[nn.Module, nn.Module]
 
-        # Use a main model to contain all of the models because when using horovod
+        # Use a main model to contain all the models because when using horovod
         # to broadcast the states of models we want to avoid name conflicts for these
-        # states so we set all the models to be sub-module of the main model with
+        # states, so we set all the models to be submodule of the main model with
         # different names using __setattr__ and use the state_dict of the main model
         # for broadcasting. Note that broadcast_parameters only accepts state_dict()
         # although its doc says it also accepts named_parameters()
@@ -264,7 +264,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             if step_mode != pytorch.LRScheduler.StepMode.MANUAL_STEP:
                 raise det.errors.InvalidExperimentException(
                     "detected that context.wrap_lr_scheduler() was called with an instance of "
-                    "torch.optim.lr_scheduer.ReduceLROnPlateau as the lr_scheduler.  This lr "
+                    "torch.optim.lr_scheduler.ReduceLROnPlateau as the lr_scheduler.  This lr "
                     "scheduler class does not have the usual step() parameters, and so it can "
                     "only be used with step_mode=MANUAL_STEP.\n"
                     "\n"
@@ -318,7 +318,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
     def _filter_named_parameters(self, optimizer: torch.optim.Optimizer) -> List:
         """_filter_named_parameters filters the named parameters of a specified optimizer out
         of all the named parameters from a specified model. We need this function because
-        a ``torch.optim.Optimizer`` doesn't store parameter names and we need the names of
+        a ``torch.optim.Optimizer`` doesn't store parameter names, and we need the names of
         the parameters when mapping parameters to each ``horovod.DistributedOptimizer``.
         """
         opt_params = {p for group in optimizer.param_groups for p in group.get("params", [])}
@@ -556,8 +556,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             When using distributed training, we don't support manual gradient accumulation.
             That means the gradient on each parameter can only be calculated once on each batch.
             If a parameter is associated with multiple losses, you can either choose to call
-            backward on only one of those losses or you could set ``require_grads`` flag of a
-            parameter or module to false to avoid manual gradient accumulation on that parameter.
+            ``backward'' on only one of those losses, or you can set the ``require_grads`` flag of
+            a parameter or module to false to avoid manual gradient accumulation on that parameter.
             However, you can do gradient accumulation across batches by setting
             :ref:`optimizations.aggregation_frequency <config-aggregation-frequency>` in the
             experiment configuration to be greater than 1.
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -15,7 +15,6 @@ from determined.tensorboard import get_base_path
 try:
     import apex
 except ImportError:  # pragma: no cover
-    apex = None
     if torch.cuda.is_available():
         logging.warning("Failed to import apex.")
     pass
@@ -23,11 +22,9 @@ except ImportError:  # pragma: no cover
 # AMP is only available in PyTorch 1.6+
 try:
     import torch.cuda.amp as amp
-
-    amp_import_error = False
+    HAVE_AMP = True
 except ImportError:  # pragma: no cover
-    amp = None
-    amp_import_error = True
+    HAVE_AMP = False
     if torch.cuda.is_available():
         logging.warning("PyTorch AMP is unavailable.")
     pass
@@ -370,7 +367,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             The scaler. It may be wrapped to add additional functionality for use in Determined.
         """
 
-        check.false(amp_import_error, "Failed to import torch.cuda.amp. PyTorch >= 1.6 required.")
+        check.true(HAVE_AMP, "Failed to import torch.cuda.amp. PyTorch >= 1.6 required.")
 
         check.false(self._use_apex, "Do not mix APEX with PyTorch AMP.")
 
--- a/harness/tests/experiment/fixtures/pytorch_xor_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_xor_model.py
@@ -405,4 +405,3 @@ class XORTrialGradClipping(XORTrial):
             self.context.step_optimizer(self.optimizer)
 
         return {"loss": loss}
-
--- a/harness/tests/experiment/fixtures/pytorch_xor_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_xor_model.py
@@ -406,10 +406,3 @@ class XORTrialGradClipping(XORTrial):
 
         return {"loss": loss}
 
-
-class XORTrialWithApexAmp(XORTrial):
-    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
-        super().__init__(context)
-        self.model, self.optimizer = self.context.configure_apex_amp(
-            models=self.model, optimizers=self.optimizer
-        )
--- a/harness/tests/experiment/fixtures/pytorch_xor_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_xor_model.py
@@ -101,7 +101,7 @@ class BaseXORTrial(pytorch.PyTorchTrial):
     """
     Models a lightweight neural network model with one hidden layer to
     learn a binary XOR function. See Deep Learning Book, chapter 6.1 for
-    the solution with a hidden size of 2, and a MSE loss function.
+    the solution with a hidden size of 2, and an MSE loss function.
 
     This model has only one output node "loss".
     """
--- a/harness/tests/experiment/fixtures/pytorch_xor_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_xor_model.py
@@ -405,3 +405,11 @@ class XORTrialGradClipping(XORTrial):
             self.context.step_optimizer(self.optimizer)
 
         return {"loss": loss}
+
+
+class XORTrialWithApexAmp(XORTrial):
+    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
+        super().__init__(context)
+        self.model, self.optimizer = self.context.configure_apex_amp(
+            models=self.model, optimizers=self.optimizer
+        )
--- a/harness/tests/experiment/fixtures/pytorch_xor_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_xor_model.py
@@ -1,6 +1,6 @@
 # type: ignore
 import logging
-from typing import Any, Dict, cast
+from typing import Any, Dict, List, cast
 
 import numpy as np
 import torch
@@ -81,7 +81,7 @@ class XORNetMulti(XORNet):
 
 
 class StepableLRSchedule(torch.optim.lr_scheduler._LRScheduler):
-    def get_lr(self) -> float:
+    def get_lr(self) -> List[float]:
         return [self._step_count for _ in self.base_lrs]
 
 
@@ -90,7 +90,7 @@ class ModifyableLRSchedule(torch.optim.lr_scheduler._LRScheduler):
         self.lr = float(0)
         super().__init__(*args, **kwargs)
 
-    def get_lr(self) -> float:
+    def get_lr(self) -> List[float]:
         return [self.lr for _ in self.base_lrs]
 
     def set_lr(self, lr: float) -> None:
--- a/harness/tests/experiment/keras/test_tf_keras_trial.py
+++ b/harness/tests/experiment/keras/test_tf_keras_trial.py
@@ -41,7 +41,7 @@ def test_executing_eagerly():
 )
 def xor_trial_controller(request):
     """
-    This fixture will provide a function that takes a hyperparameters dictionary as input and
+    This fixture provides a function that takes a hyperparameter's dictionary as input and
     returns a trial controller. It is parameterized over different implementations, so that any test
     that uses it may test a full set of implementations.
     """
--- a/harness/tests/experiment/keras/test_tf_keras_trial.py
+++ b/harness/tests/experiment/keras/test_tf_keras_trial.py
@@ -41,7 +41,7 @@ def test_executing_eagerly():
 )
 def xor_trial_controller(request):
     """
-    This fixture provides a function that takes a hyperparameter's dictionary as input and
+    This fixture provides a function that takes a hyperparameter dictionary as input and
     returns a trial controller. It is parameterized over different implementations, so that any test
     that uses it may test a full set of implementations.
     """
--- a/harness/tests/experiment/pytorch/test_pytorch_context.py
+++ b/harness/tests/experiment/pytorch/test_pytorch_context.py
@@ -1,4 +1,4 @@
-import unittest
+import pytest
 
 import torch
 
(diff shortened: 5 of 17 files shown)


harness/tests/experiment/fixtures/pytorch_xor_model.py
```diff
@@ -405,3 +405,11 @@ def train_batch(
             self.context.step_optimizer(self.optimizer)
 
         return {"loss": loss}
+
+
+class XORTrialWithApexAmp(XORTrial):
```
rb-determined-ai, line 410: Let's avoid any new XOR-based tests.  Please use the linear `pytorch_onevar_model` instead.

The reason is: XOR-based tests only test that something runs; they're much harder to test if e.g. the gradients were updated as we would expect them to be.

harness/tests/experiment/keras/test_tf_keras_trial.py
```diff
@@ -41,7 +41,7 @@ def test_executing_eagerly():
 )
 def xor_trial_controller(request):
     """
-    This fixture will provide a function that takes a hyperparameters dictionary as input and
+    This fixture provides a function that takes a hyperparameter's dictionary as input and
```
rb-determined-ai, line 44: pretty sure the apostrophe is wrong.  `hyperparameter dictionary` might be better, but `hyperparameter's dictionary` is definitely not right.

harness/tests/experiment/pytorch/test_pytorch_context.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+
+from determined import pytorch
+from determined.common.check import CheckFailedError
+from determined.errors import InternalException
+from tests.experiment.fixtures import pytorch_onevar_model
+
+
+class TestPyTorchContext(unittest.TestCase):
+    def setUp(self) -> None:
```
rb-determined-ai, line 12: I'm assuming that `setUp` gets called automatically by pytest somehow?  I couldn't find any docs about it, so I wanted to confirm.

harness/tests/experiment/pytorch/test_pytorch_context.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_training_not_started(self) -> None:
+        self.assertRaises(InternalException, self.context.is_epoch_start)
+        self.assertRaises(InternalException, self.context.is_epoch_end)
+        self.assertRaises(InternalException, self.context.current_train_batch)
+        self.assertRaises(InternalException, self.context.current_train_epoch)
+        self.context.env.managed_training = True
+        self.assertRaises(InternalException, self.context._should_communicate_and_update)
+
+    def test_wrap_scalar(self) -> None:
+        scaler = 1
```
rb-determined-ai, line 45: I'd recommend using a scaler object instead of an int here.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -583,6 +610,42 @@ def make_workloads() -> workload.Stream:
         )
         controller.run()
 
+    # At the time of writing this test, supports_mixed_precision and
+    # supports_averaging_training_metrics are class methods, but we instantiate a
+    # PyTorchTrialController here in case that ever changes. We expect that a controller
+    # instantiated in the following way should always provide these supports.
+    def test_supports(self) -> None:
```
rb-determined-ai, line 617: I don't think we call `supports_mixed_precision` anywhere anymore.  We used to, to ensure somebody didn't configure mixed precision for a trial that didn't support it.

But now we are moving to Trainer APIs, and these sorts of things will get set in a parameter of the framework-specific Trainer API, and if somebody passes `mixed_precision=True` to a non-mixed-precision-capable trainer api, they'll get an error about illegal parameter names.

So I wouldn't add this test.  I would delete `supports_mixed_precision` and I would even consider deleting `supports_averaging_training_metrics` (though probably I'd hold off).

But ultimately, getters that return booleans don't need tests.  There are bigger fish to fry.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -583,6 +610,42 @@ def make_workloads() -> workload.Stream:
...
+            hparams=self.hparams,
+            workloads=make_workloads(),
+            trial_seed=self.trial_seed,
+        )
+        assert controller.supports_mixed_precision()
+        assert controller.supports_averaging_training_metrics()
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="no gpu available")
+    @pytest.mark.gpu
+    def test_apex_amp(self) -> None:
```
rb-determined-ai, line 633: after you convert this to the onevar model, I would like to see this test actually verify that the AMP is working, by feeding in an input that causes overflow or underflow and confirming that the scaling behavior is working.

Additionally, if you can confirm that the scaling behavior is properly preserved after loading from checkpoints, that would help us confirm/deny the existence of a long-standing bug report.

rb-determined-ai (commented):
If you batch up all of the grammar fixes into a separate PR, I can stamp it and we can get those out of the way.

harness/tests/experiment/keras/test_tf_keras_trial.py
drh-determined-ai: Ack, you're right. The meaning is "a dictionary of hyperparameters".

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: I believe it does, depending on `TestPyTorchContext` inheriting from `unittest.TestCase`. I know our other other test classes don't do this, but I wanted to take advantage of the `assertRaises` method.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I'm going to defer this development to a fresh branch.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -414,19 +421,41 @@ def make_workloads2() -> workload.Stream:
...
         controller = utils.make_trial_controller_from_trial_implementation(
             trial_class=pytorch_xor_model.XORTrialAccessContext,
-            hparams=self.hparams,
+            hparams=hparams,
             workloads=make_workloads(),
             trial_seed=self.trial_seed,
         )
         controller.run()
 
+    @pytest.mark.skip("WIP")  # FIXME
```
azhou-determined, line 446: i'd say this test probably isn't worth investing time into, since i imagine we'd need a whole suite of unit tests to test the behavior of the workload sequencer/generator. we plan on deprecating the workload sequencer in the near future anyway, so e2e tests are probably enough for this part of the codebase.

harness/determined/cli/experiment.py
```diff
@@ -63,7 +63,7 @@ def read_git_metadata(model_def_path: pathlib.Path) -> Tuple[str, str, str, str]
     """
     try:
         from git import Repo
-    except ImportError as e:
+    except ImportError as e:  # pragma: no cover
```
rb-determined-ai, line 66: We actually don't have a policy around this yet, but I would like to make one.

I propose that a `#pragma: no cover` with no explanation is just as harmful as a line with no coverage in the first place.  If coverage is not needed, then it should be explained when the `#pragma: no cover` is added.

What do you think?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -24,7 +25,8 @@
     import torch.cuda.amp as amp
 
     amp_import_error = False
-except ImportError:
+except ImportError:  # pragma: no cover
+    amp = None
```
rb-determined-ai, line 29: what is the purpose of `amp = None` here?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -14,7 +14,8 @@
 # Apex is included only for GPU trials.
 try:
     import apex
-except ImportError:
+except ImportError:  # pragma: no cover
+    apex = None
```
rb-determined-ai, line 18: what is the purpose of `apex = None` here?

harness/tests/experiment/fixtures/pytorch_xor_model.py
```diff
@@ -405,3 +405,4 @@ def train_batch(
             self.context.step_optimizer(self.optimizer)
 
         return {"loss": loss}
+
```
rb-determined-ai, line 408: extraneous newline

harness/tests/experiment/pytorch/test_pytorch_context.py
rb-determined-ai: Ah, I missed the TestCase subclassing part.  That explains my confusion.

Can you use `with pytest.raises` instead?  It has the same functionality but it matches the rest of our test code.

harness/tests/experiment/pytorch/test_pytorch_context.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+
+import torch
+
+from determined import pytorch
+from determined.common.check import CheckFailedError
+from determined.errors import InternalException
```
rb-determined-ai, line 7: We use the [google style guide for python imports](https://google.github.io/styleguide/pyguide.html#22-imports) (just for python imports, not the whole style guide).

So can you change these to 
```python
from determined.common import check
...
check.CheckFailedError
```
and
```python
from determined import errors
...
errors.InternalException
```

Though, in the case of `det.errors` or `det.util`, I think it's ok to cheat the style guide a bit and just use the full `det.errors.InternalException`.  However, importing individual symbols should be avoided.

harness/tests/experiment/pytorch/test_pytorch_context.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_average_gradients(self) -> None:
+        self.assertRaises(CheckFailedError, self.context._average_gradients, None, 0)
+        self.assertIsNone(self.context._average_gradients(None, 1))
+
+    def test_training_not_started(self) -> None:
+        self.assertRaises(InternalException, self.context.is_epoch_start)
+        self.assertRaises(InternalException, self.context.is_epoch_end)
+        self.assertRaises(InternalException, self.context.current_train_batch)
+        self.assertRaises(InternalException, self.context.current_train_epoch)
```
rb-determined-ai, line 42: Whoa, we should not be raising an InternalException here.  I never realized this was happening.  InternalException should be reserved for logic bugs, not "user used a public method before we wanted them to".

Testing for these exceptions is fine, but can you change them to be normal python `RuntimeErrors` instead of det.errors.InternalExceptions?

harness/tests/experiment/pytorch/test_pytorch_context.py
rb-determined-ai: A scaler object would be like [this one](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler).

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: agreed, the entire `_workload_sequencer.py` file is pure legacy, please don't waste any time testing it.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -583,6 +612,42 @@ def make_workloads() -> workload.Stream:
...
+        controller = utils.make_trial_controller_from_trial_implementation(
+            trial_class=pytorch_xor_model.XORTrialAccessContext,
+            hparams=self.hparams,
+            workloads=make_workloads(),
+            trial_seed=self.trial_seed,
+        )
+        assert controller.supports_mixed_precision()
+        assert controller.supports_averaging_training_metrics()
+
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="no gpu available")
```
rb-determined-ai, line 633: maybe also skip if `apex is None`?

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Ok, I'll scrap it.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: It's to fix a complaint from PyCharm that does seem reasonable. Without the line, you don't know if a variable `amp` will be in this scope at runtime or not, a possibility made relevant by the try-except block.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Ditto on your comment about `amp = None`.

harness/tests/experiment/fixtures/pytorch_xor_model.py
drh-determined-ai: Oops! I'll run `make check` and/or the like, don't worry.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: Understood and fixed! I think I leveled up while reading this. Good style!

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Good call.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: It would have to be `Exception`, because that's what `errors.InternalException` inherits from.

drh-determined-ai: I also don't think using `errors.InternalException` is wrong: after all, it is our code, and we should be as specific as possible.

harness/determined/cli/experiment.py
drh-determined-ai: I think the pattern is self-explanatory.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: One second thought, I think I'll keep the code but mark it as expected to fail. It can be blown away later with the other legacy code it touches.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: This makes sense, but I note that `wrap_scaler` has the type hinting `scaler: Any`, which is very general. Anyway, my using `1` is wrong and was due to my misunderstanding it as a "`scalar`".

harness/determined/cli/experiment.py
rb-determined-ai: Ok, that's fair.

There may come a time where somebody thinks something is self-explanatory when others don't think it is... but I suppose we can deal with that problem when/if it arises.

harness/tests/experiment/pytorch/test_pytorch_context.py
rb-determined-ai: Why would it have to be an Exception?

Returning an `InternalException` as a result of a user using an API when they are not allowed to is definitely wrong.  InternalException is like an HTTP 500: it means you've done nothing wrong, but you've uncovered a bug in somebody else's code.

In these cases, the user is just not allowed to call those methods during `Trial.__init__` and it indicates no internal error if the user tries to call them at the wrong time.

harness/tests/experiment/pytorch/test_pytorch_context.py
rb-determined-ai: The reason it is an `Any` is because it's tricky to have it be more precise, since the actual object type doesn't exist in some of the earlier forms of pytorch that we support.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: @drh-determined-ai, you still haven't responded to this.

Adding tests to test tech debt is basically just adding tech debt, and I really think you should not add this test to the source code.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: Please do not do this.  Adding code whose only purpose is to be deleted later just makes the codebase harder to read and harder to understand.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: If you feel the need to fix that error, that's ok, but please fix it the right way: make `amp` into an `Optional[module]` (that is, either the real amp module or `None`), and change the if statements in the code to check if `amp` is None, instead of checking if amp_import_error is False.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: ditto

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Ok, I will remove this code now, but I did not write this code in the first place with the purpose to be deleted later: I wrote it with purpose for my assignment to "write tests; improve coverage".
I made an educated suggestion to mark the code because it is easer to remove code than write it.

harness/determined/cli/experiment.py
drh-determined-ai: I absolutely agree. That's why I didn't mark this as resolved until we reached an agreement. I will mark this as resolved now.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: I'm not arguing about `InternalException` being the wrong thing for these methods to raise. I'm saying that that's the exception that these methods raise, and `InternalException` inherits from `Exception`.
Using `self.assertRaises(RuntimeError, blah)` does not work.

azhou-determined: actually, should we _ever_ throw `InternalException`? in all current uses, it's more like an invalid input/usage error. philosophically, does it even make sense? when would we catch known Runtime errors and throw up a 500? 

maybe we should just change that exception to something else, like InvalidUsageException or something. @rb-determined-ai

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: I agree with what you're asking for, but unfortunately it doesn't seem possible right now. [Here's](https://adamj.eu/tech/2021/12/29/python-type-hints-optional-imports/) a good summary of the situation.
It's much more important that we make MyPy happy than we do PyCharm, so I will revert to the flag solution here and in similar places.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Sorry. I lost track of this thread somehow. I have removed the test.

harness/tests/experiment/pytorch/test_pytorch_context.py
```diff
@@ -0,0 +1,50 @@
...
+        train_ds = trial.build_training_data_loader()
+        for epoch_idx in range(3):
+            for batch_idx, batch in enumerate(train_ds):
+                metrics = trial.train_batch(batch, epoch_idx, batch_idx)
+                # Verify the training is correct.
+                pytorch_onevar_model.OneVarTrial.check_batch_metrics(metrics, batch_idx)
+
+        eval_ds = trial.build_validation_data_loader()
+        for batch in eval_ds:
+            metrics = trial.evaluate_batch(batch)
```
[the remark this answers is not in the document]
drh-determined-ai: I fixed this a while ago. I have unresolved the conversation so you can rereview it.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: [DET-7885](https://determinedai.atlassian.net/browse/DET-7885)

harness/tests/experiment/pytorch/test_pytorch_context.py
rb-determined-ai: > I'm saying that that's the exception that these methods raise

And I am saying that it's clearly wrong for a public API to raise an "InternalError".  I'm asking you to change the functions, not just the test.

> maybe we should just change that exception to something else, like InvalidUsageException or something.

I'm fine with this too.  Just public methods should never raise an InternalError when used wrong by the user.  InternalError would mean that we had a bug, while InvalidUsageException or RuntimeError would indicate that the user has a bug (arguably, the former is clearer about that than the latter).

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: What about this:

```python
amp = None
try:
    import torch.cuda.amp

    amp = torch.cuda.amp
except ImportError:
    if torch.cuda.is_available():
        logging.warning("PyTorch AMP is unavailable.")
    pass
```

This seems to pass for me locally.  Not totally sure if it breaks mypy in other ways though.

rb-determined-ai: (also I don't have pycharm to test with pycharm)

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: Well sure, nobody writes code with the purpose of deleting.  I am sorry if I implied that you intended it to be that way; I did not think that was the case.

I have written plenty of code like that fits the category of "intended to be useful but turned out to not be useful right now".  And when that happens it's best to leave it out of the code.

Personally, I have a few branches laying around that I don't want to delete but which I don't want to land either, like this [alternate build system](https://github.com/rb-determined-ai/determined/tree/mkninja) and this [buildcache integration](https://github.com/rb-determined-ai/determined/tree/buildcache).  Both efforts were scrapped because I realized we were actively hiring infra people and they would probably rather write/maintain something entirely of their own creation.

So if you have code that's not useful right now, I'd recommend keeping it around until you know it won't be useful.  I just mean that it should not go into the codebase.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: That code might get us around the rule that we can't use a predeclared variable in an `import` statement, but it still won't satisfy PyCharm because `torch.cuda.amp` isn't defined in the `except` block.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: > And I am saying that it's clearly wrong for a public API to raise an "InternalError". I'm asking you to change the functions, not just the test.
I agree with the first sentence, and I'd be happy to make the change you're asking for, but that change is outside the scope of this PR and associated tickets.

azhou-determined: let's just change the exception in the code and tests to InvalidUsageException. we can do it in a separate PR, and leave this as is for now, since without changing the code, this test is technically accurate. agree? @rb-determined-ai @drh-determined-ai

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: got it

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: I agree with @azhou-determined and will gladly be the assignee for the task.

harness/tests/experiment/pytorch/test_pytorch_context.py
drh-determined-ai: This is done.

harness/tests/experiment/fixtures/pytorch_xor_model.py
drh-determined-ai: This is done.

rb-determined-ai (approved):
nice work!

PR #4495, opened 2022-07-07
test: fix editable metadata test flake [DET-7782]

hkang1:
## Description

UPDATE:
Ran this multiple times (about a dozen times on CircleCI) and did not encounter a failure during those runs.

PROBLEM:
There seems to be a timing issue where the dropdown option of `Delete Row` doesn't exist yet when `user.click()` is getting called. There is a slight animation of displaying the dropdown, so under high load, the animation could possibly be delayed.

SOLUTION:
Change the `getBy...` UI element finders to `findBy...` equivalent (which are basically combination of `waitFor` and `getBy...` that wait for element to exist).

## Test Plan

- [ ] check that the CI react unit tests passes. Could rerun this test over and over to see if any flakes occur.

## Checklist

### DIFF
--- a/webui/react/src/components/Metadata/EditableMetadata.test.tsx
+++ b/webui/react/src/components/Metadata/EditableMetadata.test.tsx
@@ -64,9 +64,8 @@ describe('EditableMetadata', () => {
     ));
     const { handleOnChange } = setup(initMetadata, true);
 
-    await user.click(screen.getAllByRole('button')[removalIndex]);
-    await user.click(screen.getByText('Delete Row'));
-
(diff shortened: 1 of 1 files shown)


hamidzr:
I looked to check whether getByText does not have the wait built in but I couldn't do that easilyt through the doc or src code 

https://github.com/testing-library/react-testing-library 
https://testing-library.com/docs/queries/bytext/

hamidzr (approved):
I saw the doc you shared about the queries. lg

hkang1:
> I saw the doc you shared about the queries. lg

Thanks @hamidzr !

PR #4496, opened 2022-07-07
docs: add ClusterInfo docs

rb-determined-ai:


rb-determined-ai:
@gh-determined-ai I wasn't quite sure where the reference docs belong, so I lumped them with the Core API.  It's not the core api technically, but non-Core-API users wouldn't need the ClusterInfo API, so it's related.  Anyway, it's a first-draft.

gh-determined-ai (approved):

PR #4499, opened 2022-07-07
fix: make actor system `GetOrElseTimeout` actually timeout.

ioga:
## Description

- `GetOrElseTimeout` was de-facto only returning once it's got the
  response, regardless of the timeout value. The unit test for it was
  testing it wrong.

## Checklist

### DIFF
--- a/master/pkg/actor/response.go
+++ b/master/pkg/actor/response.go
@@ -53,15 +53,20 @@ func (r *response) Source() *Ref {
 	return r.source
 }
 
-func (r *response) get() Message {
+func (r *response) get(cancel <-chan bool) Message {
 	if r.fetched {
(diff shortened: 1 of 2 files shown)


master/pkg/actor/response.go
```diff
@@ -53,15 +53,20 @@ func (r *response) Source() *Ref {
...
 		return r.result
 	}
 	r.lock.Lock()
 	defer r.lock.Unlock()
 	r.fetched = true
-	r.result = <-r.future
-	return r.result
+	select {
+	case result := <-r.future:
+		r.result = result
```
NicholasBlaskey, line 65: Can this be
```suggestion
	case r.result = <-r.future:
```

PR #4500, opened 2022-07-08
feat: add new PyTorchCallbacks [DET-7760]

garrett361:
## Description

Add two new `PyTorchCallback`s,  `on_training_workload_end` and `on_checkpoint_upload_end`, available for use with both `PyTorchTrial` and `DeepSpeedTrial`.

- `on_training_workload_end`: gives users access to the `avg_metrics` and `batch_metrics`  collected during each training workload
- `on_checkpoint_upload_end`: allows users to trigger actions after checkpoint has been uploaded to external storage

## Test Plan

Combination of manual e2e testing and new unit tests. Expected behaviors:

- `on_training_workload_end`:  should be called after each training workload has finished and should give access to both the individual `batch_metrics` and the processed `avg_metrics` from the batches in the workload.
- `on_checkpoint_upload_end`: should be called after checkpoint upload has completed and should give access to the `uuid` used as the `storage_id` in the training loop.

The above behaviors were verified by manual inspection of training logs for both single- and multi-slot training.

## Checklist


rb-determined-ai:
I don't think we can safely move the existing callback, because right now, it's a valid place to store extra information into the directory (we pass the directory to the user as a parameter) and expect it to be uploaded.  It would be a pretty bad breakage for that to suddenly become invalid.  That is why I originally proposed the (awkward) name of `on_checkpoint_upload_end()`.

rb-determined-ai:
Oh... you are right.  Moving it inside of that function seems very reasonable to me.

harness/determined/pytorch/_callback.py
```diff
@@ -126,6 +135,16 @@ def on_training_epoch_end(self, epoch_idx: int) -> None:
         """
         pass
 
+    def on_training_workload_end(self, metrics: Dict[str, Any]) -> None:
```
rb-determined-ai, line 138: is this `metrics` going to contain an extra layer of dictionaries, one for averaged metrics and one for batch metrics?

e2e_tests/tests/experiment/test_pytorch.py
```diff
@@ -280,3 +280,30 @@ def test_epoch_sync(num_workers: int, global_batch_size: int, dataset_len: int)
             assert exp.check_if_string_present_in_trial_logs(
                 t_id, f"rank {rank} finished batch {batch_idx} in epoch {epoch_idx}"
             )
+
+
+@pytest.mark.e2e_cpu
```
rb-determined-ai, line 285: can you write these as unit tests instead of e2e tests?

That will be far more performant, and more useful to developers.

e2e_tests/tests/fixtures/pytorch_no_op/model_def.py
```diff
@@ -6,6 +6,7 @@
 from torch import nn
 
 from determined import pytorch
+from determined.pytorch import PyTorchCallback
```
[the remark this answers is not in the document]
garrett361: Ah right, forgot about that one, thanks!

harness/determined/pytorch/_callback.py
garrett361: Right now it just the `metrics` arg just receives a single dictionary which has `batch_metrics` and `avg_metrics` as sub-dictionaries, but those could be split out into two args, which is probably clearer. I'll do that.

harness/determined/pytorch/_callback.py
```diff
@@ -126,6 +135,16 @@ def on_training_epoch_end(self, epoch_idx: int) -> None:
...
+        """
+        Run on end of a training workload. Each workload contains a maximum of ``scheduling_unit``
+        batches, as defined in the experiment config.
+
+        .. warning::
+            This callback only executes on the chief GPU when doing distributed training.
```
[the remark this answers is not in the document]
garrett361: Right now only the chief has access to the `storage_id` that gets passed into this callback, but I guess we could broadcast it to all workers if needed.

However, I was working under the impression that this callback would be used to kick off some external process related to the uploaded checkpoint, in which case you would probably only want a single worker to use the callback, no?  I think that was the most probable use-case we came up with, in which case you probably only want `on_checkpoint_upload_end` being called once.

harness/determined/pytorch/_callback.py
```diff
@@ -126,6 +135,16 @@ def on_training_epoch_end(self, epoch_idx: int) -> None:
         """
         pass
 
+    def on_training_workload_end(self, metrics: Dict[str, Any]) -> None:
+        """
+        Run on end of a training workload. Each workload contains a maximum of ``scheduling_unit``
+        batches, as defined in the experiment config.
```
rb-determined-ai, line 141: I recommend softening this language to make it easier for us to change it with trainer API and true epoch-based training.

"The exact length of a training workload is not guaranteed, though in the current implementation of PyTorchTrial, each workload contains a maximum ..."

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
rb-determined-ai: hm?  This comment is about `on_training_workload_end`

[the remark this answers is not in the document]
garrett361: Oops, misread where you were commenting on. I think we want both of these callbacks to only fire on the chief worker, though, since also for `on_training_workload_end` only the chief has access to the relevant metrics.  We could again broadcast them to all workers, but if the user wants this, they could just do the broadcasting themselves in the callback.

e2e_tests/tests/experiment/test_pytorch.py
garrett361: I thought about that for a while, and looked at the existing unit tests for inspiration, but given how little these callbacks do, the primary thing to check seems to be ensuring they fire where you expect them to.  I will look into doing this as a unit test somehow some more, but I wasn't able to see a way of implementing a useful unit test up to this point.

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
rb-determined-ai: actually they _can't_ do the broadcasting themselves, because the master would hang waiting for the broadcast to be received by the workers, while the workers were hanging waiting for the master to finish its `on_train_workload_end`.

> only the chief has access to the relevant metrics
Yeah, I was kind of afraid of this.  Do you think this is going to be a problem for users?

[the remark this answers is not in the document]
garrett361: Ah right, since `on_train_workload_end` will only currently be called on the chief.

So, I could refactor so that the broadcasting works and all workers have access to the training metrics, but if we think the most-likely use-case for `on_train_workload_end` is to pass the training metrics to some other visualization or logging backend (as indicated by the OSS user who last asked about this), then you would only want one worker to be calling `on_train_workload_end`.

In the spirit of maximum flexibility and minimal assumptions, though, I will try to refactor both `on_train_workload_end` and `on_checkpoint_upload_end` so that they're called on every worker. If the user only wants them to fire on the chief, they can write their callbacks to have a reference to their own `PyTorchTrial` instances, so they can have `if self.trial.is_chief ...` conditionals as needed.

e2e_tests/tests/experiment/test_pytorch.py
garrett361: Rewritten as unit tests, now that I understand our unit tests better.

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
rb-determined-ai: I like this, but there's still one more complication, and that's that the training metrics are still just worker 0's metrics, unless the user specifically set `average_training_metrics` to true in their config.

This is a bit of a hot take, but what do you think about making that value true either:
- by by default, and only call `on_training_workload_end` from all workers when it is set to true, or
- always, and always call `on_training_workload_end` with the correct metrics on every worker

Right now, it seems like a bit of a gotcha that the training metrics we broadcast to every worker are just worker 0's metrics by default.  The reason the system is like that right now is that the first dev to write our dtrain support thought training metrics were pretty dumb, and the first customer who wanted our dtrain support literally only rated us on wall clock performance.  (not even model accuracy!)

harness/determined/pytorch/_callback.py
```diff
@@ -88,13 +88,19 @@ def on_checkpoint_save_start(self, checkpoint: Dict[str, Any]) -> None:
 
     def on_checkpoint_end(self, checkpoint_dir: str) -> None:
```
rb-determined-ai, line 89: optional, but what do you think about renaming this to `on_checkpoint_write_end()`, and leave a deprecated alias for it called `on_checkpoint_end()`?

harness/determined/pytorch/_callback.py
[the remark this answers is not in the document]
garrett361: Ugh, yep, totally forgot that `average_training_metrics` is `False` by default.  Thought I had manually tested for this, but I must have have it set to `True` or not tested what I thought I did; my bad.

I actually wrote a Jira ticket a while ago ([DET-7240]) to benchmark the cost of changing the default to `True`, since it seemed strange to me that it wasn't already the default and the cost of making the switch seemed low. So, my opinion is clear.  I haven't managed to revisit the ticket since opening it, though.

harness/determined/pytorch/_callback.py
garrett361: Agree entirely. I expanded the `on_checkpoint_end` doc string because I thought the method name would be extra confusing after the addition of `on_checkpoint_upload_end`

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -38,6 +38,12 @@ def __init__(self, trial_inst: det.Trial, *args: Any, **kwargs: Any) -> None:
         if torch.cuda.is_available():
             self.prof._set_sync_device(self._sync_device)
         self.callbacks = self.trial.build_callbacks()
+        for callback in self.callbacks.values():
+            if util.is_overridden(callback.on_checkpoint_end, pytorch.PyTorchCallback):
+                logging.warning(
+                    "The on_checkpoint_end callback is deprecated, please use on"
+                    "checkpoint_write_end instead"
+                )
```
rb-determined-ai, line 46: can you use a FutureWarning, which automatically guarantees that the line only prints once?

There's a good example of a FutureWarning [here](https://github.com/determined-ai/determined/blob/master/harness/determined/common/experimental/checkpoint/_checkpoint.py#L216-L224).

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -274,14 +280,30 @@ def _run(self) -> None:
...
+                            batch_metrics=metrics["batch_metrics"],
+                        )
+                        if (
+                            self.context.distributed.size > 1
+                            and not self.context._average_training_metrics
+                        ):
+                            logging.warning(
+                                "Only the chief worker's training metrics are being reported, due "
+                                "to setting average_training_metrics to False."
+                            )
```
rb-determined-ai, line 305: Let's not log this:
- more than once per python process, or
- if the user never defined an on_training_workload_end

harness/determined/pytorch/_pytorch_trial.py
garrett361: Done.

harness/determined/pytorch/_pytorch_trial.py
garrett361: Ah, this one was actually meant for [4646](https://github.com/determined-ai/determined/pull/4646), just to bring extra attention to the consequence of having `average_training_metrics` false.  Removed here and added there.  It's also intended to be shown regardless of what callbacks are used, since its importance is independent of callbacks.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -38,6 +39,13 @@ def __init__(self, trial_inst: det.Trial, *args: Any, **kwargs: Any) -> None:
         if torch.cuda.is_available():
             self.prof._set_sync_device(self._sync_device)
         self.callbacks = self.trial.build_callbacks()
+        for callback in self.callbacks.values():
+            if util.is_overridden(callback.on_checkpoint_end, pytorch.PyTorchCallback):
+                warnings.warn(
+                    "The on_checkpoint_end callback is deprecated, please use on"
+                    "checkpoint_write_end instead",
```
rb-determined-ai, line 46: you are missing an underscore from `oncheckpoint_write_end`

rb-determined-ai (commented):
Just realized that this doesn't look like it's being applied to deepspeed trial yet, even though deepspeed trial uses the same callback structure.

gh-determined-ai (approved):

docs/release-notes/4500-new-pytorchtrial-callbacks.txt
```diff
@@ -0,0 +1,8 @@
+:orphan:
+
+**Improvements**
+
+-  API: new ``on_training_workload_end`` and ``on_checkpoint_upload_end`` callbacks available for
+   ``PyTorchTrial``.
```
rb-determined-ai, line 6: and DeepSpeedTrial, right?

rb-determined-ai (approved):
Great work!

garrett361:
Thank you @rb-determined-ai !

PR #4503, opened 2022-07-08
test: fix local time tests [DET-7449]

hkang1:
## Description

PROBLEM:
tests dealing with local time is failing because the local time conversion doesn't take Daylight Savings Time difference into account.

SOLUTION:
use native `dayjs.local()` to automatically handle DST

## Test Plan

- [ ] check that the datetime tests pass locally
- [ ] check that the datetime tests pass in CI

## Checklist

### DIFF
--- a/webui/react/src/shared/utils/datetime.test.ts
+++ b/webui/react/src/shared/utils/datetime.test.ts
@@ -4,7 +4,6 @@ import * as utils from './datetime';
 
 describe('Datetime Utilities', () => {
   describe('formatDatetime', () => {
-    const SYSTEM_UTC_OFFSET = dayjs().utcOffset();
     const DATE = [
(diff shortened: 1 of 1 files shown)


hamidzr:
hmm, so it was an issue with the test and not our utility?

hamidzr (approved):
thanks for taking this on. tests pass locally for me. as long as [this is a correct assumption](https://github.com/determined-ai/determined/pull/4503#issuecomment-1179341494) ship it! 🚀

gt2345 (approved):
Works great

hkang1:
> hmm, so it was an issue with the test and not our utility?

I wrote this test a while back when writing a bunch of utility tests, but I'm not sure why this is coming up now. DST being the issue and failing locally, this should have failed a while ago. 🤷

PR #4507, opened 2022-07-11
feat: new searcher name custom in experiment config

mpkouznetsov:
## Description

Add `name` as authorized value for searcher.

## Checklist

### DIFF
--- a/schemas/expconf/v0/searcher-custom.json
+++ b/schemas/expconf/v0/searcher-custom.json
@@ -1,0 +1,15 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://determined.ai/schemas/expconf/v0/searcher-custom.json",
+    "title": "CustomConfig",
+    "type": "object",
+    "additionalProperties": false,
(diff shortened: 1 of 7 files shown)


schemas/expconf/v0/searcher-custom.json
```diff
@@ -0,0 +1,15 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://determined.ai/schemas/expconf/v0/searcher-custom.json",
+    "title": "CustomConfig",
+    "type": "object",
+    "additionalProperties": false,
+    "required": [
```
nrajanee, line 7: looks good to me. Do we need anything else here? Or we can add it later if we can think of anything.

nrajanee (approved):

schemas/expconf/v0/searcher-custom.json
mpkouznetsov: exactly - I cannot think of anything right now but we are free to update this while the feature branch is there.

PR #4515, opened 2022-07-11
fix: minor bugs in test config creation

pjanowski:
## Description

Proposing a fix for a minor bug when Determined creates a test config. Currently `slots_per_trial` are not be set to 1 as intended.

## Test Plan
Tested locally.

## Commentary (optional)

None

## Checklist

## Title

fix: set slots_per_trial in test experiment correctly


gh-determined-ai (approved):

pjanowski:
@loksonarius Dan, I don't know much about Determined's linters/unit tests so let me know if there's anything more I can do here.

loksonarius:
I'm re-running e2e test preemptively, but I haven't actually looked at that failure yet.
As for the `lint` check failing, there's an error in the `lint-docs` job that mentions the formatting of the changelog entry:
image
I'll double check what the expected output is manually and let ya know.

docs/release-notes/4515-fix-test-submit.txt
```diff
@@ -0,0 +1,5 @@
+:orphan:
+
+**Bug Fix**
+
+-  When creating a `test` experiment the number of slots should be set to 1 but a bug kept it at the original number. This fix ensures that the number of slots is configured to 1.
```
loksonarius, line 5:
```suggestion
:orphan:

**Bug Fix**

-  When creating a `test` experiment the number of slots should be set to 1 but a bug kept it at the
   original number. This fix ensures that the number of slots is configured to 1.
```

loksonarius:
`test-e2e` is moving along now -- I'm not sure why it failed previously to run `build-go` with an "Unauthorized" error, but it passed that know. Waiting on `test-det-deploy-local` and a re-run of `test-e2e-cpu`.

pjanowski:
@loksonarius Thanks Dan, I committed the doc formatting change (I had see the error in CI but didn't see any indication of what the actual problem was, I guess it was max line length).

pjanowski:
Yeah, linting passes now but `tests-e2e` still fail

loksonarius:
So I manually kicked off the `test-e2e` job that was marked as unauthorized and made a PR for us that should prevent that `Unauthorized` issue from popping up again in the future.
There's just two tests left running right now for the `e2e_workflow`: https://app.circleci.com/pipelines/github/determined-ai/determined/24723/workflows/149b36c3-2ee2-4ca7-8e40-eaf864b01939

ioga:
also cc @rb-determined-ai for sanity check

harness/determined/common/api/experiment.py
```diff
@@ -217,11 +217,11 @@ def make_test_experiment_config(config: Dict[str, Any]) -> Dict[str, Any]:
                 "max_length": {"batches": 1},
             },
             "hyperparameters": generate_random_hparam_values(config.get("hyperparameters", {})),
-            "resources": {**config_test.get("resources", {"slots_per_trial": 1})},
```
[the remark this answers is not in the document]
pjanowski: Apologies for the slow response but how does that fail since just after I added this:

```config_test["resources"]["slots_per_trial"] = 1```

Wouldn't that ensure that `config["resources"` always exists?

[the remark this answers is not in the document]
Comment (ioga) [reply harness/determined/common/api/experiment.py:220]: no, if config_test["resources"] is not a dict already this would fail. for example:

```
In [1]: config_test = {}

In [2]: config_test["resources"]["slots_per_trial"] = 1
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-586ebcd75583> in <module>
----> 1 config_test["resources"]["slots_per_trial"] = 1

KeyError: 'resources'
```

that may happen when the input experiment config doesn't have the `resources` section.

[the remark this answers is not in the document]
Comment (pjanowski) [reply harness/determined/common/api/experiment.py:220]: duh, of course, thanks, let me put in a fix

File: harness/determined/common/api/experiment.py [hunk 3]
```diff
@@ -217,11 +217,11 @@ def make_test_experiment_config(config: Dict[str, Any]) -> Dict[str, Any]:
...
             "max_restarts": 0,
         }
     )
-    config.setdefault(
```
rb-determined-ai, line 224: ouch, good catch.

harness/determined/common/api/experiment.py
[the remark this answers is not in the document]
rb-determined-ai: Caveat: I'm not convinced we should change the current behavior, but if we do...

`config_test["resources"]["slots_per_trial"] = 1` is exactly the line that would fail if `config_test["resources"]` is None.

But rather than doing any more fancy oneliners, which this PR demonstrates can be the source of bugs, let's just do it simple-style:

```python
# handle resources not present, or resources is None
if not config_test.get("resources"):
    config_test["resources"] = {}
config_test["resources"]["slots_per_trial"] = 1
```

rb-determined-ai (commented):
Question: why not let --test mode work on multiple slots?

I for one didn't realize that it wasn't supposed to, as you can tell from how I wrote [this](https://docs.determined.ai/latest/training-debug/index.html#step-7-verify-that-cluster-test-mode-works-with-slots-per-trial-set-to-1) model debugging step.

I'm not sure I see the point in enforcing single-slot --test mode.

pjanowski:
> Question: why not let --test mode work on multiple slots?
> 
> I for one didn't realize that it wasn't supposed to, as you can tell from how I wrote [this](https://docs.determined.ai/latest/training-debug/index.html#step-7-verify-that-cluster-test-mode-works-with-slots-per-trial-set-to-1) model debugging step.
> 
> I'm not sure I see the point in enforcing single-slot --test mode.

Main reason would be because of `make_test_experiment_config`'s docstring which led me to expect that each job would run on a single slot. As a user I have no strong feelings as long as the docs and behavior are consistent. Let me know what you prefer.

pjanowski:
Also, got tests to pass but now the linter is failing again. If you can tell me where the issue I'm happy to update.

rb-determined-ai:
I agree they should match. I would prefer a docstring change.  I don't see any linting failures...

rb-determined-ai:
oh, actually, I do think the code should get updated to leave resources alone.  It's clearly _trying_ to set slots_per_trial to one, it's just not working, and I think that happens to be better.

pjanowski:
> I agree they should match. I would prefer a docstring change. I don't see any linting failures...

@rb-determined-ai Ryan, thanks, see latest commit which should be what you are asking for here. As for linting I set up `black` locally...

docs/release-notes/4515-fix-test-submit.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Bug Fix**
+
+-  When creating a `test` experiment the container storage path was not being set correctly.
+   Also fixes the docstring which incorrectly claimed that only one slot would be used.
```
rb-determined-ai, line 6: nit: Since the docstring was not published in our docs, I would consider this detail about the docstring not important enough for release notes.

Definitely worth fixing though, just not for release notes.

harness/determined/common/api/experiment.py
```diff
@@ -216,12 +215,15 @@ def make_test_experiment_config(config: Dict[str, Any]) -> Dict[str, Any]:
                 "metric": config_test["searcher"]["metric"],
                 "max_length": {"batches": 1},
             },
-            "hyperparameters": generate_random_hparam_values(config.get("hyperparameters", {})),
-            "resources": {**config_test.get("resources", {"slots_per_trial": 1})},
+            "hyperparameters": generate_random_hparam_values(
+                config_test.get("hyperparameters", {})
+            ),
+            "resources": config_test.get("resources", {"slots_per_trial": 1}),
```
rb-determined-ai, line 221: can we just drop this line entirely?  I don't see its purpose anymore.

pjanowski: Without this line the test will fail if the config does not already have a `resources` key. The test expects a `resources` key to be present. That was the first failing test we ran into. FWIW, I personally also like that it's clear that it will run on either the provided slots or 1 by default.

docs/release-notes/4515-fix-test-submit.txt
pjanowski: ok, done

harness/determined/common/api/experiment.py
rb-determined-ai: ah, this is because most of the system depends on the master filling in default values, except the tests.

Ok. cool.

rb-determined-ai (approved):
Thank you, great work!

PR #4517, opened 2022-07-11
fix: skip gzip only for proxy web assets [DET-7802]

azhou-determined:
## Description
gzip compression shortens loading time for web assets, but newer versions of tensorboard don't support it. workaround is to only skip gzip compression for proxy paths.

## Checklist

### DIFF
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -79,7 +79,7 @@ var staticWebDirectoryPaths = map[string]bool{
 	"/docs/rest-api": true,
 }
 
-// gzipSkipPaths are locations of paths to be skipped by GZIP compression
+// gzipSkipPaths are locations of paths to be skipped by GZIP compression.
 var gzipSkipPaths = map[string]bool{
 	"/proxy": true,
 }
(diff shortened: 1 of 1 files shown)


master/internal/core.go
```diff
@@ -79,6 +79,11 @@ var staticWebDirectoryPaths = map[string]bool{
 	"/docs/rest-api": true,
 }
 
+// gzipSkipPaths are locations of paths to be skipped by GZIP compression
```
hkang1, line 82: **nit:** dang, strict lint! just needs a period at the end it looks like

hkang1 (commented):
LGTM, I would rely on @stoksc to evaluate if this is good on the Go side. I'll give this a shot locally but not sure if that would enable gzip properly in the first place.

hkang1:
this was only tested locally, but when locally deployed the payload size seems to still be uncompressed?
Screen Shot 2022-07-11 at 5 05 26 PM
This chunk.js file for example, is still being served as a 1.4 MB file. The bottom number should have shown something much smaller (in the KBs).

hkang1 (approved):
ok tested it again locally with @julian-determined-ai and can confirm it is working! Thanks for enabling this!

PR #4519, opened 2022-07-11
chore: User Groups SQL

maxrussell:
## Description

Implements user groups SQL as described in [DET-7803].

## Test Plan

To test this, I ran a local Postgres and ElasticSearch (OpenSearch), then ran the new integration tests for this code like so (from the `master/internal/usergroups` directory; replace `[user]` and `[password]` with your Postgres user and password):

```
DET_INTEGRATION_ES_HOST=localhost DET_INTEGRATION_ES_PORT=9200 DET_INTEGRATION_POSTGRES_URL="postgres://[user]:[password]@127.0.0.1:5432/determined" go test -tags=integration -run 'TestUserGroups'
```

[DET-7803]: https://determinedai.atlassian.net/browse/DET-7803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	res, err := tx.NewDelete().Table("user_group_membership").
+		Where("group_id = ?", gid).
+		Where("user_id IN (?)", bun.In(uids)).
+		Exec(ctx)
+	if foundErr := mustHaveAffectedRows(res, err); foundErr != nil {
```
NicholasBlaskey, line 137: I believe this won't throw an error if I call it with a user in the group and a user not in the group since we are only checking if ```affectedRows``` is nonzero compared to the function comment above implying we need to check if ```affectedRows``` is ```len(uids)```? 

Adding users to groups doesn't have this issue because it protected by the foreign key / duplicate key pg I think?

master/internal/db/group_intg_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+func TestUserGroups(t *testing.T) {
```
NicholasBlaskey, line 16: Nice integration tests! Good work.

NicholasBlaskey (commented):
Great work! I'm going to let @ioga review / approve

master/static/migrations/20220706152654_add-user-groups.tx.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE groups (
+    id serial PRIMARY KEY,
```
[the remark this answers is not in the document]
maxrussell: Good catch!

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return g, matchSentinelError(err)
+}
+
+// SearchGroups searches the database for groups. userBelongsTo is "optional"
+// in that if a value < 1 is passed in, the parameter is ignored. SearchGroups
+// does not return an error if no groups are found, as that is considered a
+// successful search.
+func (db *PgDB) SearchGroups(ctx context.Context,
+	userBelongsTo model.UserID) ([]model.Group, error) {
```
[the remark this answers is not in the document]
maxrussell: If that's a strong preference, that seems okay with me.

I tend to like to avoid extraneous pointers when I can, and I think calling it with a non-pointer signature is a little less of a hassle. For example:

```go
g, err := db.SearchGroups(ctx, 12)
g, err := db.SearchGroups(ctx, user.ID)
g, err := db.SearchGroups(ctx, 0)
```
vs
```go
var id model.UserID = 12
g, err := db.SearchGroups(ctx, &id)
g, err := db.SearchGroups(ctx, &user.ID)
g, err := db.SearchGroups(ctx, nil)
```

Though I totally recognize the oddness of passing in an out-of-band value as a sentinel to ignore a parameter.

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// DeleteGroup deletes a group from the database. Returns ErrNotFound if the
+// group doesn't exist.
+func (db *PgDB) DeleteGroup(ctx context.Context, gid int) error {
+	tx, err := Bun().BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = tx.Rollback() }()
```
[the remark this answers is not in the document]
maxrussell: Definitely can. Good idea.

master/internal/db/group_intg_test.go
```diff
@@ -0,0 +1,235 @@
...
+		foundGroup, err := pgDB.GroupByID(ctx, testGroup.ID)
+		require.NoError(t, err, "failed to find group by id")
+		require.Equal(t, testGroup.Name, foundGroup.Name, "Expected found group to have the same name as the one we created")
+	})
+
+	t.Run("update group", func(t *testing.T) {
+		// Put it back the way it was when we're done
+		defer func(name string) {
+			testGroup.Name = name
+			pgDB.UpdateGroup(ctx, testGroup)
```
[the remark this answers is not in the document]
maxrussell: Good idea!

master/internal/db/postgres_groups.go
maxrussell: You're absolutely right about both. Time to add to the test cases.

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return err
+}
+
+// mustHaveAffectedRows checks if bun has affected rows in a table or not.
+// Returns ErrNotFound if no rows were affected and returns the provided error otherwise.
+func mustHaveAffectedRows(result sql.Result, err error) error {
+	if err == nil {
+		rowsAffected, affectedErr := result.RowsAffected()
+		if affectedErr == nil && rowsAffected == 0 {
```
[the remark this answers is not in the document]
maxrussell: Great question. AFAIK we should only get an error here when the driver doesn't support this operation. But on further consideration, since calling code might depend on this all-or-nothing behavior, I'm leaning toward returning `affectedErr`. I'll update the code to do so, and if anyone disagrees, please say so!

master/static/migrations/20220706152654_add-user-groups.tx.up.sql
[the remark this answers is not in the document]
ioga: I also did not know serial is bad now. cool.
however I believe this old transaction code was not handwritten but generated from some wizard/tool.

probably if a real human was writing this after checking on postgres docs, it'd be 
```
     id integer PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
```
without extra alter table step

master/internal/db/group_intg_test.go
```diff
@@ -0,0 +1,246 @@
...
+		require.NoError(t, err, "failed to create group")
+	})
+
+	t.Run("search groups", func(t *testing.T) {
+		groups, err := pgDB.SearchGroups(ctx, 0)
+		require.NoError(t, err, "failed to search for groups")
+
+		index := groupsContain(groups, testGroup.ID)
+		foundGroup := groups[index]
+		require.NotEqual(t, -1, index, "Expected groups to contain the new one")
```
ioga, line 35: if it's -1 first line will probably fail with some panic because go doesn't have negative indices
```suggestion
		require.NotEqual(t, -1, index, "Expected groups to contain the new one")
		foundGroup := groups[index]
```

master/internal/db/postgres_groups.go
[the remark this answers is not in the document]
ioga: I also somewhat prefer nil pointers over special errors

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	err := query.Scan(ctx)
+
+	return groups, err
+}
+
+// DeleteGroup deletes a group from the database. Returns ErrNotFound if the
+// group doesn't exist.
+func (db *PgDB) DeleteGroup(ctx context.Context, gid int) error {
```
ioga, line 52: can we have `ON DELETE CASCADE` for foreign keys in `user_group_membership` in SQL instead? then this won't have to be a transaction.

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,218 @@
...
+		rbErr := tx.Rollback()
+		if rbErr != sql.ErrTxDone && rbErr != nil {
+			log.WithError(rbErr).
+				WithField("groupID", gid).
+				WithField("userIDs", uids).
+				Error("error rolling back transaction in AddUsersToGroup")
+		}
+	}()
+
+	res, err := tx.NewInsert().Model(&groupMem).Exec(ctx)
```
ioga, line 125: hmm why does this func need a Tx? looks like it runs a single insert?

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,218 @@
...
+		rbErr := tx.Rollback()
+		if rbErr != sql.ErrTxDone && rbErr != nil {
+			log.WithError(rbErr).
+				WithField("groupID", gid).
+				WithField("userIDs", uids).
+				Error("error rolling back transaction in RemoveUsersFromGroup")
+		}
+	}()
+
+	res, err := tx.NewDelete().Table("user_group_membership").
```
ioga, line 156: similar to prev func - why does this need a Tx?

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,218 @@
...
+	var users []model.User
+	err := Bun().NewSelect().Table("users").Model(&users).
+		Join("INNER JOIN user_group_membership AS ugm ON users.id=ugm.user_id").
+		Where("ugm.group_id = ?", gid).
+		Scan(ctx)
+
+	return users, err
+}
+
+func matchSentinelError(err error) error {
```
ioga, line 181: nice. we could keep these helpers in `db` package for reuse

master/static/migrations/20220706152654_add-user-groups.tx.up.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE groups (
+    id integer NOT NULL,
+    group_name text unique NOT NULL,
+    user_id integer REFERENCES users (id) NULL
```
ioga, line 4:
```suggestion
    user_id integer REFERENCES users (id) ON DELETE CASCADE NULL
```

master/static/migrations/20220706152654_add-user-groups.tx.up.sql
```diff
@@ -0,0 +1,21 @@
...
+    START WITH 1
+    INCREMENT BY 1
+    NO MINVALUE
+    NO MAXVALUE
+    CACHE 1
+);
+
+CREATE TABLE user_group_membership (
+    user_id integer REFERENCES users (id),
+    group_id integer REFERENCES groups (id),
```
ioga, line 18:
```suggestion
    user_id integer REFERENCES users (id) ON DELETE CASCADE,
    group_id integer REFERENCES groups (id) ON DELETE CASCADE,
```

master/internal/db/postgres_groups.go
```diff
@@ -0,0 +1,218 @@
...
+	tx, err := Bun().BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		rbErr := tx.Rollback()
+		if rbErr != sql.ErrTxDone && rbErr != nil {
+			log.WithError(rbErr).
+				WithField("groupID", gid).
+				WithField("userIDs", uids).
```
ioga, line 120: nitpick: looking at `git grep 'WithField'` we mostly do `field-names-with-dashes`, e.g. `trial-id` or `allocation-id`
```suggestion
				WithField("group-id", gid).
				WithField("user-ids", uids).
```

ioga (approved):
lgtm

master/internal/db/group_intg_test.go
maxrussell: Oh, good catch. Line order should be reversed here.

master/internal/db/postgres_groups.go
maxrussell: Thanks! Yeah, I thought about that briefly. But I can sorta imagine it being possible different packages need to respond to codes differently. So for the time being decided to leave this specific to this package and then if we discover we're doing the same thing in several spots we can collect it into a centralized place. Totally willing to change my mind on this if y'all disagree that this seems prudent.

PR #4521, opened 2022-07-12
fix: custom searcher schema

mpkouznetsov:
## Description

Attempting to create an experiment with "searcher": { "name": "custom" } fails because of validation failures.

## Checklist


harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -2637,13 +2637,30 @@
     "$id": "http://determined.ai/schemas/expconf/v0/searcher-custom.json",
     "title": "CustomConfig",
     "type": "object",
-    "additionalProperties": false,
+    "additionalProperties": true,
```
mpkouznetsov, line 2640: this is probably the most potentially controversial part of this change. I have to do it because "custom" does not support `source_trial_id` and `source_checkpoint_uuid`. I am happy to revisit this later.

mpkouznetsov:
I am going to merge this to unblock the client but the corresponding commit should be reviewed on the feature branch.

PR #4524, opened 2022-07-12
feat: Reject experiments which take too many slots to run [DET-7722]

mapmeld:
## Description

A known problem in our queuing system is that if an experiment's slotsPerTrial is larger than the largest agent, it will be queued forever and never run ( https://determinedai.atlassian.net/browse/DET-7722 ).  I'm working on flagging these experiments in the UI (#4420) but the consensus was that we should raise an error and not create these experiments in the first place.

This code returns an error from `parseCreateExperiment` because it's used by all paths to create an experiment (CLI, forks, new API, old API, wherever).

I've added `force_queue` as an experiment config option to override this check.

## Test Plan

We already had a test that creates an infinite-queue experiment. I modified the test so we expect an error, then add `force_queue` to the experiment config and resume the original experiment.

## Checklist

### DIFF
--- a/master/internal/core_experiment.go
+++ b/master/internal/core_experiment.go
@@ -367,6 +367,27 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams, user *mod
 		projectID = *params.ProjectID
 	}
 
+	if !resources.ForceQueue() {
+		resourceManager := sproto.GetCurrentRM(m.system)
+		agReq := m.system.Ask(resourceManager, &apiv1.GetAgentsRequest{})
+		if agReq.Error() != nil {
+			return nil, false, nil, agReq.Error()
+		}
+		getAgentsResponse := agReq.Get().(*apiv1.GetAgentsResponse)
+		foundAgent := false
+		for _, agent := range getAgentsResponse.Agents {
+			if len(agent.Slots) >= resources.SlotsPerTrial() {
+				foundAgent = true
+				break
+			}
+		}
+		if !foundAgent {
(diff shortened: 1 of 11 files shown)


ioga:
>if an experiment's slotsPerTrial is larger than the largest agent, it will be queued forever and never run

that is not true for experiments: `slots_per_trial: 16` will need two 8-slot agents, or four 4-slot agents, etc. that's just our distributed training. it is true for NTSC (notebooks, tensorboards, ...) however because these cannot be "distributed".

master/internal/core_experiment.go
```diff
@@ -367,6 +367,24 @@ func (m *Master) parseCreateExperiment(params *CreateExperimentParams, user *mod
 		projectID = *params.ProjectID
 	}
 
+	if !resources.ForceQueue() {
+		resourceManager := sproto.GetCurrentRM(m.system)
+		agReq := m.system.Ask(resourceManager, &apiv1.GetAgentsRequest{})
+		if agReq.Error() != nil {
+			return nil, false, nil, agReq.Error()
+		}
+		getAgentsResponse := agReq.Get().(*apiv1.GetAgentsResponse)
```
hamidzr, line 376: missing a check to see if the cast was successful

hamidzr:
also would we need to look at the configured resource pools as well as the agents to determine this?

mapmeld:
Closing with these issues:
- currently uses the wrong limit for experiments
- not applicable when resource manager can provision more slots
- likely to cause problems during temporary agent outages or changes on a cluster

PR #4533, opened 2022-07-13
fix: race condition on experiment config for 'det e create --paused' [DET-7789]

NicholasBlaskey:
## Description

Race condition found through ```--race``` on experiments creation paused. The race is that we don't wait for the experiment actor's ```PreStart``` to finish before returning the experiment config which ```PreStart``` modifies it. 

This wasn't an issue when the experiment was set to activate since we send the activation message to the experiment actor causing synchronization to occur. 

Interestingly enough ```--race``` detected even when the experiment was set to activate since it was smart enough to know that that branch didn't always activate. Which kinda blew my mind. Could be interesting to add a way to report race conditions in devcluster in automated tests.

## Test Plan

Run master built with ```make build-race```

then run any experiment and verify no ```--race``` output occurred. 

## Checklist


stoksc (commented):
works for this, but also, this is just a general issue, tbh. we don't do a great job at deep copying all actor messages, so this probably isn't the only place this happens (setWeight can be sent, too, so if someone sends it immediately after activating, I think this still happens).

there is this cool language, pony, that has reference capabilities with actors just for this (or rusts borrow checker).. but in go we just need to remember either deep copy the config I think (rip .clone()) or remove this from the response.. I can't think of anywhere it is used (would be cool if you could `e, err := newExperiment(m, move dbExp, move taskSpec)` in go so you could prevent this sort of bug a bit easier).

NicholasBlaskey:
Pony looks cool and the story is interesting
  
The most recent changed I assumed we care about getting the updated config, maybe that is a wrong assumption.

stoksc (commented):
lgtm

stoksc (approved):

stoksc:
> The most recent changed I assumed we care about getting the updated config, maybe that is a wrong assumption.

this code is kinda confusing, I don't think it actually gets updated in this case, I think `e.setPriority` just always updates then tells the RM (even if called from PreStart without an update). It is returned because the updates are interesting, but the updates that are interesting are the ones that happen when we merge in defaults/master config settings before it is created.

PR #4538, opened 2022-07-13
fix: scroll bar overlap

keita-determined:
## Description

[DET-7815]

### Before
Screen Shot 2022-07-13 at 4 18 46 PM

### After
Screen Shot 2022-07-13 at 4 42 16 PM

## Test Plan
Verify scroll bar is not on text in the Modal page

## Checklist

[DET-7815]: https://determinedai.atlassian.net/browse/DET-7815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.module.scss
+++ b/webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.module.scss
@@ -3,6 +3,7 @@
   flex-direction: column;
   max-height: 400px;
   overflow: auto;
+  padding-right: 16px;
 
   .row {
     display: flex;


webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.module.scss
```diff
@@ -3,6 +3,7 @@
   flex-direction: column;
   max-height: 400px;
   overflow: auto;
+  padding-right: 1rem;
```
trentwatt, line 6: I think we pretty much only use px in our codebase. Not sure on details, maybe @hkang1 can confirm/elaborate?

keita-determined: Hmm. I usually use `rem` over `px` cuz it is scalable. For example, if user wanna change the text bigger `rem` can do that very easily. but I'll follow the codebase. but I wanna know the @hkang1 opinion too.

hkang1: So historically I have used `rem` with `:root/html` of font-size set to `62.5%` so all the rem units are a factor of 10 compared to px (1rem = 10px). The reason in the past was for accessibility reasons, where if you used `px` units and you change the browser default font size, it didn't change the text size. rem came in the picture to save the day, but with an added cost of dev complexities.

Modern browsers today support zoom (`cmd` + `+/-`) really well, where it will take font-size px values and dynamically adjust them based on zoom level. So even thought px still remains static relative to default font-size, the zoom override provides a way forward for accessibility. So with modern browsers, I typically don't use rem anymore unless accessibility is one of the top primary concerns (generally mobile focused apps that caters to folks that have partial vision impairment).

Additional notes on rem application, there are two rough school of thoughts:
- applying them to all size units
- applying them ONLY to text size units
each has their own implications and they differ widely on results.

keita-determined: That makes sense. `cmd` + `+/-` is actually well implemented. I think applying rem to text is more make sense, but depends on context ofc...
Anyways, I'll change rem to px.

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.module.scss
```diff
@@ -3,6 +3,7 @@
   flex-direction: column;
   max-height: 400px;
   overflow: auto;
+  padding-right: 16px;
```
keita-determined, line 6: used `16px` cuz 1rem = 16px in this environment

trentwatt (approved):
LGTM.

PR #4543, opened 2022-07-14
WIP: Query Trials API

julian-determined-ai:
## Checklist


master/trials.tx.up.sql
```diff
@@ -0,0 +1,76 @@
+-- keeping out of the migration directory for now to avoid having it run
```
trentwatt, line 1: going to try to get rid of the aggs in the view, then do a proper migrations file

master/internal/db/postgres.go
```diff
@@ -82,7 +82,9 @@ func PaginateBun(
 		direction = SortDirectionAsc
 	}
 	orderExp := fmt.Sprintf("%s %s", orderColumn, direction)
-	query = query.Order(orderExp)
+
+	// needed this in order to order by e.g. metrics->>'loss'
+	query = query.OrderExpr(orderExp)
```
trentwatt, line 87: this is unsafe, input would need to be sanitized, maybe should add a boolean flag and have safe be default?

master/trials.tx.up.sql
```diff
@@ -0,0 +1,76 @@
...
+  STYPE = jsonb,
+  INITCOND = '{}'
+);
+
+
+-- aggregations should be mostly redundant.the only way they 
+-- would get applied is when there are multiple trial_run_ids for a single 
+-- (trial_id, total_batches) on steps or validations. in which case,
+-- which trial_run_id is it appropriate to take the metrics from?
+-- there have been situations where they differ.
```
trentwatt, line 37: basically in order to get rid of the aggregation,  I am blocked on the question "Which `trial_run_id` is the proper source of truth for metrics."

I have seen situations where for the same `(trial_id, total_batches)`, run 2 has `{"metric_a", "0.33", "metric_b": 0.66}` and run 3 has {`"metric_a": "0.33"}` (which means I cant just take the first or the last).

master/internal/db/postgres_trial.go
```diff
@@ -441,3 +450,288 @@ WHERE t.id = $1;
...
+		q.Where("? BETWEEN ? AND ?", column, startTime.AsTime(), endTime.AsTime())
+	case startTime != nil:
+		q.Where("? > ?", column, startTime.AsTime())
+	case endTime != nil:
+		q.Where("? < ?", column, endTime.AsTime())
+	}
+}
+
+// FilterTrials queries for Trials matching the supplied TrialFilters.
+func (db *PgDB) FilterTrials(q *bun.SelectQuery,
```
ioga, line 615: I think for readability it may be useful to distinguish database funcs which build and run a query and return the result vs the ones which just build a query for use elsewhere. I am not sure we already have a precedents for this case. so perhaps rename this function to one of:
- `BuildFilterTrialsQuery`
- `MakeFilterTrialsQuery`
- `BuildFilterTrials`
- `MakeFilterTrials`
- `FilterTrialsQuery`
or similar

master/internal/api_trials.go
```diff
@@ -368,6 +370,292 @@ func (a *apiServer) GetTrialCheckpoints(
...
+	req *apiv1.QueryTrialsRequest) (*apiv1.QueryTrialsResponse, error) {
+	err := checkTrialFiltersEmpty(req.Filters)
+	if err != nil {
+		return nil, fmt.Errorf("error querying tags for trials %w", err)
+	}
+
+	trials := []db.TrialsAugmented{}
+
+	q := db.Bun().NewSelect().Model(&trials)
+	q, err = a.m.db.FilterTrials(q, req.Filters, true)
```
ioga, line 414: how about moving `q := db.Bun().NewSelect().Model(&trials)` *inside* `BuildFilterTrialsQuery` 

i.e. in `BuildFilterTrialsQuery` do
```
q := db.Bun().NewSelect().Model((*db.TrialsAugmented)(nil))
```
and here do
```
q := BuildFilterTrialsQuery(req.Filters, true)
...
...
trials := []db.TrialsAugmented{}
q.Scan(context.TODO(), &trials)
```

this way the caller won't have to do this "Bun-NewSelect-Model" preamble for each `FilterTrials` call.

master/internal/api_trials.go
```diff
@@ -368,6 +370,292 @@ func (a *apiServer) GetTrialCheckpoints(
...
+	if err != nil {
+		return nil, fmt.Errorf("couldnt patch trials %w", err)
+	}
+
+	// check user is authorized for modifying project? after RBAC?
+	// in that case we will want to make sure len(req.Filters.ProjectID) == 1
+	// right now only option is adding/removing tags, pretty low stakes
+
+	q := db.Bun().NewUpdate().Table("trials")
+	q, err = a.m.db.ApplyTrialPatch(q, req.Patch)
```
ioga, line 468: can we move this line inside `ApplyTrialPatch` similar to the previous comment? maybe the function is not "applying" anymore, and could be `BuildTrialPatchQuery`

master/internal/db/postgres.go
```diff
@@ -37,8 +37,8 @@ func initTheOneBun(db *sql.DB) {
 	// This will print every query that runs.
 	// theOneBun.AddQueryHook(bundebug.NewQueryHook(bundebug.WithVerbose(true)))
 
-	// This will print only the failed queries.
-	theOneBun.AddQueryHook(bundebug.NewQueryHook())
+	// Change this back before merge
```
ioga, line 40: 👀

master/internal/db/postgres_trial.go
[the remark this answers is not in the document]
ioga: (current `postgres_trial` will move there eventually)

master/static/srv/proto_get_trials_plus.sql
```diff
@@ -42,6 +42,8 @@ trial_validations AS (
     ) IS NOT NULL
 ),
 best_validation AS (
+  -- is the window function needed here?
```
ioga, line 45: I believe it'd still be a nested query with group by. we could benchmark it though.

master/trials.tx.up.sql
```diff
@@ -0,0 +1,76 @@
...
+);
+
+-- make analogous modifications to those found in reindex-steps pr
+-- because we are joining validations on (trial_id, total_batches)
+CREATE INDEX validations_archived ON raw_validations(archived);
+ALTER TABLE raw_validations DROP CONSTRAINT validations_trial_id_run_id_total_batches_unique;
+CREATE UNIQUE INDEX validations_trial_id_total_batches_run_id_unique ON raw_validations(trial_id, total_batches, trial_run_id);
+
+
+CREATE AGGREGATE jsonb_collect(jsonb) (
```
ioga, line 26: cool

trentwatt:
@ioga thanks for review. would it make sense to just put all this stuff (And any new bun stuff going forward) into a `master/internal/trials`

ioga:
@trentwatt yes, bun stuff and these new API views can go into `master/internal/trials`.

master/internal/api_trials.go
trentwatt: yeah, this is a vestige from when `FilterTrials` took a `bun.QueryBuilder` (to accommodate `bun.UpdateQuery`). that ended up not working because `ExpRank` was doing a window expression in a `WHERE` clause, which apparently is not legal. so ended up doing `WHERE IN ({trial_id_subquery})` instead.

on the other hand the [update query](https://github.com/determined-ai/determined/pull/4543/files#diff-22336866e3c77178639f72a957328a898ce19c360034317e30275b87fb8869b6R482) is only using the`ids` from the filtered subquery, whereas  `q := db.Bun().NewSelect().Model(&trials)` would select them all. pretty sure the query compiler wouldnt bother, but i guess we would have to do an extra `q.select(id from ?', subQ)` [here](https://github.com/determined-ai/determined/pull/4543/files#diff-22336866e3c77178639f72a957328a898ce19c360034317e30275b87fb8869b6R482-R487). which to me is a little goofy, but maybe there is a better way..?

ioga: won't this work?

```
q := BuildFilterTrialsQuery(req.Filters, true)
q = q.Column("trial_id")
```

trentwatt: maybe so? will check. seems like bun generally gets confused when you specify things in two different ways (in this case, specifying the columns via both model and `.Column()`.

trentwatt: yeah looks like that works.

 couple questions.

- Is there a name for this type of construct: `*db.TrialsAugmented)(nil)` and/or what does it do?
- in any case it looks like it avoids scanning into a model? but in the case of the select query we actually do want to scan. in which case we should just do a `.Scan(&here)` at the end?
- can you help me understand a little more the thinking behind where the boundaries should be between these functions? the initial thought was to have separate functions for different clauses: one handles `WHERE`, one handles `SET` etc,... although that also broke down around the `ExpRank` functionality....

master/internal/api_trials.go
trentwatt: ok, by this line you mean `q := db.Bun().NewUpdate().Table("trials")`, right?

master/internal/db/postgres_trial.go
trentwatt: just pick one for me lol 🙃

trentwatt: in seriousness though, im partial to `BuildTrialFiltersQuery` (not on the list but much in keeping)

master/trials.tx.up.sql
```diff
@@ -0,0 +1,76 @@
...
+  LEFT JOIN validations v ON t.id = v.trial_id AND v.id = t.best_validation_id
+  LEFT JOIN steps s on t.id = s.trial_id AND v.total_batches = s.total_batches
+  LEFT JOIN b on t.id = b.trial_id 
+  group by t.id
+  -- find other subquery way to do this??
+
+
+-- is the assumption here valid? will we always have the row in steps for a corresponding row in validations?s
+-- does avg_metrics correspond to the actual state at that batch?
+-- or an average over previous batches?
```
trentwatt, line 76: note here: "this assumption" refers to the assumption that we could use (paraphrasing) `batch = validations.batch where validation.id = trial.best_validation id` and basically get all the validation and training metrics from that same batch. 

the goal being for each metric to associate a value to a trial.

This assumption remains valid, [for now](https://determinedai.atlassian.net/browse/DET-4292?focusedCommentId=16689), but It is worth calling attention to the fact that this is _not ideal_. For example: searcher metric might be best at one point, but another validation metric is better at another point. 

Given the alternatives though:
- scanning raw_validations/steps
- doing a different "index" in some form or fashion for each validation metric

this seemed the least bad option. In any case searcher metric is probably a decent proxy for the other metrics, and they can drill into more how the various metrics evolved on the charts

trentwatt: apparently`JSONB` indexes on fields with numeric values [can be used to efficiently query this kind of thing](https://stackoverflow.com/a/30122313/9778768), but it would be a pretty massive index probably.

master/internal/db/postgres_trial.go
```diff
@@ -441,3 +450,288 @@ WHERE t.id = $1;
...
+	}
+
+	if len(filters.Tags) > 0 {
+		tagKeys := []string{}
+		for _, tag := range filters.Tags {
+			tagKeys = append(tagKeys, tag.Key)
+		}
+		// bun please ignore the first question mark,
+		// it is an operator, not a placeholder
+		q.Where("tags ?| ?", "?", pgdialect.Array(tagKeys))
```
trentwatt, line 667: This needs some parens

master/internal/api_trials.go
ioga: yes

master/internal/api_trials.go
ioga: 1. it's a typed nil pointer. it's typed so bun can reflect the type and figure out which model/table to query
2. yes
3. I'd break it down based on how you reuse and compose them around the code. we could come up with some naming scheme, like `Build...` creates a new query while `Apply...` modifies a query passed via parameter.

proto/src/determined/api/v1/trial.proto
```diff
@@ -44,6 +45,319 @@ message WorkloadContainer {
   }
 }
 
+// A filter structure to allow for range filters on arbitrary keys
+message NumberRangeFilter {
```
ioga, line 49: maybe `DoubleRangeFilter`? Because one day we may need `Int32RangeFilter`.

master/determined.code-workspace
```diff
@@ -1,13 +1,23 @@
...
-			"path": "."
-		},
-		{
-			"path": "../agent"
-		},
-		{
-			"path": "../proto"
-		}
-	]
+  "folders": [
```
stoksc, line 2: this definitely was accidentally added in the checkpoints work, but if we wanna keep it around as a shared one, cool. (it was my pr, but it is not my workspace so idk how it got there, a few people worked on the pr)

trentwatt: Yeah will remove

master/internal/trials/api_trials.go
```diff
@@ -0,0 +1,323 @@
...
+	if user == nil {
+		return nil, errors.New("user not found")
+	}
+	collections := []*TrialsCollection{}
+
+	q := db.Bun().
+		NewSelect().
+		Model(&collections)
+
+	// Where("user_id = ?", user.ID)
```
ioga, line 186: Remove?

master/internal/trials/api_trials.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if req.ProjectId != 0 {
+		q = q.Where("project_id = ?", req.ProjectId)
+	}
+
+	err = q.Scan(context.TODO())
+
+	if err != nil {
+		return nil, fmt.Errorf("couldnt get trials collections %w", err)
```
ioga, line 195:
```suggestion
		return nil, fmt.Errorf("failed to get trials collections %w", err)
```

master/static/migrations/20220907143410_trials-collection.tx.down.sql
```diff
@@ -0,0 +1,17 @@
+DROP INDEX validations_trial_id_total_batches_run_id_unique;
+
+DROP INDEX validations_archived;
+ALTER TABLE raw_validations ADD CONSTRAINT validations_trial_id_run_id_total_batches_unique UNIQUE (trial_id, trial_run_id, total_batches);
+
+drop view if exists public.trials_augmented_view;
```
ioga, line 6:
```suggestion
DROP VIEW public.trials_augmented_view;
```

master/static/migrations/20220907143410_trials-collection.tx.down.sql
```diff
@@ -0,0 +1,17 @@
...
+
+DROP AGGREGATE jsonb_collect(jsonb);
+
+DROP TABLE trials_collections;
+
+
+
+drop index trials_tags_index;
+
+alter table trials drop column if exists tags;
```
ioga, line 17: we don't auto-lint these at the moment, but could you please fix blank lines and capitalization?

master/trials.tx.up.sql
ioga, line 1: this file should be removed?

proto/src/determined/api/v1/trial.proto
```diff
@@ -44,6 +45,324 @@ message WorkloadContainer {
   }
 }
 
+// A filter structure to allow for range filters on arbitrary keys
+message DoubleRangeFilter {
```
ioga, line 49: there's a parallel discussion about filters here: https://github.com/determined-ai/determined/pull/4925#discussion_r966373082

I think this is hitting the same question: we need a unified way for standard filters of fields. This min/max solution is basically `__gte` / `__lte`, in addition to `_in` / `_notin` in Keita's PR.

proto/src/determined/api/v1/trial.proto
```diff
@@ -44,6 +45,324 @@ message WorkloadContainer {
...
+  google.protobuf.DoubleValue max = 3;
+}
+
+// A filter structure for filtering times within a given interval.
+message TimeRangeFilter {
+  // The start of the interval. optional.
+  google.protobuf.Timestamp interval_start = 1;
+  // The end of the interval. optional.
+  google.protobuf.Timestamp interval_end = 2;
+}
```
ioga, line 64: this is also another `gte` / `lte` filter.

ClaireNeveu: Okay, looks like protobuf has built-in support for multi-stage decoding now in the form of `Any`.

What about just something like this:
```
type Filter =
 | Eq(Any)
 | GT(Any)
 | GTE(Any)
 | LT(Any)
 | LTE(Any)
 | In(Repeated<Any>)
 | NotIn(Repeated<Any>)
 | InRange(Any, Any)
```
Using `OneOf` to make a sum type. We can drop that in some common proto file. Feels pretty straightforward.

master/internal/db/bun_filters.go
```diff
@@ -19,30 +20,46 @@ type FilterComparison[T any] struct {
 	Lte *T
 }
 
-func applyFieldFilterComparison[T any](
+func applyFieldFilterComparison[T any, U string | schema.Ident](
```
ClaireNeveu, line 23: I tried to make this `SelectExpr | Ident` but it doesn't _seem_ to work. It's possible the error was in where I was constructing the expression but Bun was blowing the stack when I tried to interpolate a SelectExpr.

e2e_tests/tests/cluster/test_trial_collections.py
```diff
@@ -0,0 +1,282 @@
...
+) -> None:
+    target: Dict[str, Any] = (
+        {"trial": {"ids": trials}} if isinstance(trials, list) else {"filters": trials}
+    )
+    patch = {"patch": {"addTag": addTags, "removeTag": removeTags}}
+    body = bindings.v1UpdateTrialTagsRequest.from_json({**target, **patch})
+    bindings.patch_UpdateTrialTags(sess, body=body)
+
+
+def assert_same_ids(a: List["SupportsLessThanT"], b: List["SupportsLessThanT"]) -> None:
```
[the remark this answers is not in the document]
ClaireNeveu: This is the correct typing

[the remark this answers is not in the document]
ioga: it seems the current version of typeshed is doing `Iterable[SupportsRichComparisonT]`: https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L1646
the difference between `SupportsRichComparisonT`, `SupportsLessThanT` or pure `List` won't move the needle here, mypy doesn't require it, so I'd recommend going for the simpler solution.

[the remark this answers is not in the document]
ClaireNeveu: We can easily change it when we upgrade

proto/src/determined/trial/v1/trial.proto
```diff
@@ -30,6 +30,30 @@ message CheckpointWorkload {
   google.protobuf.Struct metadata = 9;
 }
 
+// The current state of the trial. see \dT+ trial_state in db
+enum State {
```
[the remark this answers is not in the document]
trentwatt:
```
 determined> \dT+ trial_state
 
 PAUSED             
 STOPPING_CANCELED  
 STOPPING_KILLED    
 STOPPING_COMPLETED 
 STOPPING_ERROR     
 CANCELED           
 COMPLETED          
 ERROR              


determined> \dT+ experiment_state

PAUSED             
STOPPING_CANCELED  
STOPPING_KILLED    
STOPPING_COMPLETED 
STOPPING_ERROR     
DELETING           
CANCELED           
COMPLETED          
ERROR              
DELETE_FAILED      
```

so change `Trial`? 

related: can we override/remove the rule about prefixing with enum with `STATE_`? seems to just make things more error-prone.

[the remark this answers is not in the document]
ioga: >so change `Trial`?

that seems right a good way to go, can you give it a try? I am not sure if that'd need some dramatic changes. On JSON API level it shouldn't affect anything.

>related: can we override/remove the rule about prefixing with enum with STATE_? seems to just make things more error-prone.

it is controversial, but 
- we do it everywhere so that be awkward for consistency
- if we change it here it'd be a breaking change for a bunch of Trial APIs.

e2e_tests/tests/api_utils.py
```diff
@@ -0,0 +1,10 @@
+from determined.common import api
+from determined.common.api import authentication, certs
+from tests import config as conf
+
+
+def determined_test_session() -> api.Session:
```
ioga, line 6: heads up for a rebase conflict: I updated this method in https://github.com/determined-ai/determined/pull/5055/files#diff-66c234e5c973d0546d1f6713c5b496deca22124eea269b1f8c5a3fb0c1193c81R275

proto/src/determined/trial/v1/trial.proto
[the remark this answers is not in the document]
trentwatt: can this be a follow up?

master/internal/trials/api_trials.go
```diff
@@ -0,0 +1,321 @@
...
+// of trial ids, or a set of filters, according to the provided patch.
+func (a *TrialsAPIServer) UpdateTrialTags(ctx context.Context,
+	req *apiv1.UpdateTrialTagsRequest,
+) (*apiv1.UpdateTrialTagsResponse, error) {
+	_, _, err := grpcutil.GetUser(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldnt patch trials %w", err)
+	}
+
+	// check user is authorized for modifying project? after RBAC?
```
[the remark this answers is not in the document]
trentwatt: so this is okay for now, but do a fast follow up? or do it now?

> it must be done before RBAC ships though, i.e. soon.

what is the target? next release? one after?

master/internal/api_experiment.go
```diff
@@ -1433,6 +1434,7 @@ func (a *apiServer) fetchTrialSample(trialID int32, metricName string, metricTyp
 	return &trial, nil
 }
 
+// DEPRECATED -- do not use.
```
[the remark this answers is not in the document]
trentwatt: get rid of `ExpCompareTrialsSample`

proto/src/determined/api/v1/api.proto
```diff
@@ -500,16 +500,86 @@ service Determined {
...
+  // Update tags for trials by the trial ID
+  rpc PatchTrials(PatchTrialsRequest) returns (PatchTrialsResponse) {
+    option (google.api.http) = {
+      patch: "/api/v1/trials"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: [ "Trial Comparison" ]
+    };
+  }
```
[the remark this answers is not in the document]
trentwatt: > Switch existing /api/v1/trials/compare to /api/v1/trial-comparison/compare as a tech debt item.

"tech debt item" = separate ticket?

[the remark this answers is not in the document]
trentwatt: > Looking at this from an API user perspective, if I discovered an API called "PatchTrials" I'd not be expecting it to do what it does in this PR. So I'd suggest to change to a different, more specific name, e.g. UpdateTrialTags

The thought was to potentially allow for different updates in the future, but this makes sense for now.

master/static/migrations/20220914115713_trials-collection.tx.down.sql
```diff
@@ -0,0 +1,17 @@
+DROP INDEX validations_trial_id_total_batches_run_id_unique;
+
+DROP INDEX validations_archived;
+ALTER TABLE raw_validations ADD CONSTRAINT validations_trial_id_run_id_total_batches_unique UNIQUE (trial_id, trial_run_id, total_batches);
+
+drop view if exists public.trials_augmented_view;
```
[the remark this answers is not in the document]
trentwatt: ok, i am assuming this applies for columns as well?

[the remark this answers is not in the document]
ioga: you mean `alter table trials drop column if exists tags;`? yes. 
I can't think of any practical application for `if exists` in our app.

proto/src/determined/api/v1/api.proto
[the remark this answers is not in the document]
ioga: >"tech debt item" = separate ticket?

If you could change it in this PR it'd be stellar. Otherwise yes let's please have a tracking ticket for that.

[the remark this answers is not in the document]
trentwatt: it would require frontend changes as well since that API is already being used by unrelated work. 

I would rather keep this scoped to just backend changes if that is okay.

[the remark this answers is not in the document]
ioga: sounds good.

proto/src/determined/trial/v1/trial.proto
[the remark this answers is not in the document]
trentwatt: (this would require frontend changes as well)

master/internal/trials/api_trials.go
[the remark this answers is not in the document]
ioga: Fast follow up is ok

proto/src/determined/trial/v1/trial.proto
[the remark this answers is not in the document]
ioga: Fast follow up is ok

ioga (approved):

PR #4546, opened 2022-07-14
ci(scan-docker-images): migrate to gha

loksonarius:
## Description

Looking to migrate our CircleCI workflow that scans our docker images weekly against a locally-deployed anchor engine into a GHA workflow that uses another anchore tool: [Grype](https://github.com/anchore/grype#specifying-matches-to-ignore). Having read over the script we currently run, it seems like we can have the same functionality (including ignoring specific CVEs) using the grype action.

## Test Plan

Doing some testing locally with `act`, but ultimately, I'll be trusting what the actual output of the job is when run from an actual runner. Before merging in this change, I'll be removing the `pull_request` trigger from the workflow definition.

## Checklist

### DIFF
--- a/.github/workflows/scan-docker-images.yml
+++ b/.github/workflows/scan-docker-images.yml
@@ -1,0 +1,50 @@
+---
+name: "Scan Docker images"
+
+on:
+  schedule:
+    - cron: "15 1 * * 0"
+
+jobs:
+  list-images:
+    runs-on: ubuntu-latest
+    outputs:
+      images: ${{ steps.get-images.outputs.result }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Parse image list
+        id: get-images
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq -o=json '[with_entries(select(.key | test("_hashed$"))).[].new]' tools/scripts/bumpenvs.yaml
+  scan-images:
+    needs: list-images
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        image: ${{ fromJSON(needs.list-images.outputs.images) }}
+    steps:
+      - name: Free up some space
+        if: contains(matrix.image, 'cuda-11') || contains(matrix.image, 'rocm')
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: true
+      - name: Scan ${{ matrix.image }}
+        id: scan
+        continue-on-error: true
+        uses: anchore/scan-action@v3
+        with:
+          image: ${{ matrix.image }}
+          acs-report-enable: true
+          fail-build: true
+          severity-cutoff: high
+      - name: Print SARIF report for ${{ matrix.image }}
+        run: cat ${{ steps.scan.outputs.sarif }}
+      - name: Upload SARIF report for ${{ matrix.image }}
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: ${{ steps.scan.outputs.sarif }}
+      - name: Fail job if scan failed
+        if: steps.scan.outcome == 'failure'
+        run:  exit 1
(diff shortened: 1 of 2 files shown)


loksonarius:
Sweet, the job definition works just fine now with the `free-disk-space` action running before the `cuda-11` and `rocm` image scans.
I'll be deleting the definition of this workflow from our CircleCI pipeline in a new commit and removing the test changes I made for the first commit.

Sample successful run: https://github.com/determined-ai/determined/runs/7347289994?check_suite_focus=true

loksonarius:
Going over the reports, they're accurate. One thing worth noting is that a scan failure only reports that there were vulnerabilities outside of our ignore list that are of severity `high` or higher, the SARIF report includes _all_ vulnerabilities regardless of our ignore list as its output is separate to the output of `grype`. Using `grype` locally from the repo root against any of the images listed will give a report that excludes ignored vulnerabilities.

I'm removing the WIP label now that I'm good with the state of the workflow and removing the `pull_request` trigger and `if: 'false'` protection around SARIF report uploading. PR's ready for review now.

ioga:
@loksonarius when working on images, I'd sometimes build and scan a new image locally as a test. Can we have docs/readme on how exactly someone can run the scan on a single local image (specified by docker tag) with `act`?
also I suppose this PR can delete `tools/scripts/scan-docker-images.sh`

dannysauer:
I'd probably be inclined to delete the old scan separately after letting them run in parallel for a couple of iterations.

.github/workflows/scan-docker-images.yml
```diff
@@ -0,0 +1,50 @@
+---
+name: "Scan Docker images"
+
+on:
+  schedule:
+    - cron: "15 1 * * 0"
```
dannysauer, line 6: I'm sorely disapointed that github doesn't just support `@weekly` so they can load balance.

dannysauer (approved):
I'm cool with this.  Though, if you're deleting all the code which runs the old check, it probably does make sense to delete tools/scripts/scan-docker-images.sh like @ioga said.

.github/workflows/scan-docker-images.yml
loksonarius: 🤷 I felt the same way, but I guess I wouldn't care to deal with the ramifications of GitHub-wide managed `weekly` load balancing of tasks. If I were on the GHA team, I'd prolly just be salty having to implement that instead of letting users just write their own unique schedules.

loksonarius:
> @loksonarius when working on images, I'd sometimes build and scan a new image locally as a test. Can we have docs/readme on how exactly someone can run the scan on a single local image (specified by docker tag) with `act`?

Do we iterate on images locally from the mono repo or from the environments repo? If the latter, I figure the tooling and docs to do that would just be a `make` target in that repo.

Presuming we need to be able to do that from _this_ repo, I'd recommend users install `grype` and just use that locally -- it's _drastically_ faster than `act` and doesn't require us maintaining a GitHub workflow that _just_ generates a `grype` report and isn't used in any automated workflow. I can add a quick `make` target that:

- runs `grype ${image-name}` from the root of the repo if the CLI is installed locally
- runs `docker run --rm -v $PWD/.grype.yaml:/tmp/.grype.yaml anchore/grype:latest '${image-name}`

This should allow ez local scanning without demanding everyone download a new tool.

> also I suppose this PR can delete `tools/scripts/scan-docker-images.sh`

>Though, if you're deleting all the code which runs the old check, it probably does make sense to delete tools/scripts/scan-docker-images.sh like @ioga said.

I'm cool keeping both runs around for a bit -- one only generates noise for us currently. I'm sure I'll be looking at this repo again on Monday after the new workflow runs to see what kind of notifications it generates for us when the report is uploaded.
If nothing's too bad, I'll delete the CircleCI workflow in two weeks. That sound cool?

dannysauer:
If running the scan locally provides significant value, perhaps it'd make sense to include the filesystem scan in the upcoming precommit action instead of as a make target?

ioga:
the scan takes a ton of time, so in practice I'd only do it on-demand when I REALLY want it.

loksonarius:
@cla-bot[bot] check

PR #4549, opened 2022-07-14
feat: Custom searcher client-side API

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/DET-7585,
https://determinedai.atlassian.net/browse/DET-7620

Initial implementation of the python-side API.
This works for a client-side (custom search) equivalent of Single search, implemented in `e2e_tests/tests/experiment/test_custom_searcher.py::test_run_custom_searcher_experiment`.

## Test Plan

Run `e2e_tests/tests/experiment/test_custom_searcher.py::test_run_custom_searcher_experiment`.

## Checklist


master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -35,7 +35,6 @@ func (q *SearcherEventQueue) GetEvents() []*experimentv1.SearcherEvent {
 func (q *SearcherEventQueue) RemoveUpTo(eventID int) error {
 	for i, v := range q.events {
 		if v.Id == int32(eventID) {
-			q.eventCount -= -v.Id
```
mpkouznetsov, line 38: We decided that q.eventCount is the total event count. It is incremented with each event and never decremented.

master/internal/api_experiment.go
```diff
@@ -75,28 +75,56 @@ func (a *apiServer) getExperiment(experimentID int) (*experimentv1.Experiment, e
...
 func (a *apiServer) GetSearcherEvents(_ context.Context, req *apiv1.GetSearcherEventsRequest) (
-	*apiv1.GetSearcherEventsResponse, error) {
+	resp *apiv1.GetSearcherEventsResponse, err error) {
 	exp, err := a.getExperiment(int(req.ExperimentId))
 	if err != nil {
-		return nil, errors.Wrap(err, "fetching experiment from db")
+		return nil, err
+	}
+	if exp.State == experimentv1.State_STATE_COMPLETED {
+		eventCount := 1000 // TODO get event count from experiment
```
mpkouznetsov, line 84: This is a problem we have to address, either in this PR or later. Since the experiment is gone we don't have access to the searcher or the search method or the queue. The only thing I can think of is save the number of events in the experiment in the DB. Suggestions are welcome.

nrajanee: I can address it in this ticket: https://determinedai.atlassian.net/browse/DET-7717.

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -111,17 +134,23 @@ message Hyperparameter {
   // A user can provide one of the hyperparameter types for
   // the custom searcher.
   oneof union {
-    // Constant hyperparameter.
-    ConstantHyperparameter constant_hyperparam = 1;
+    // Double hyperparameter.
+    DoubleHyperparameter double_hyperparam = 1;
```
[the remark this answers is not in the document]
mpkouznetsov: No, they are still essentially constant, not ranges. I ran into type validation problems and had to introduce types specific to the wrapped value types but those are single values, not intervals.

master/internal/api_experiment.go
nrajanee: Right so I was thinking of returning a list like this: [ExperimentInactive] then it'll look similar to the events list in the SearcherQueue.

proto/src/determined/experiment/v1/searcher.proto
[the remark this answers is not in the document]
nrajanee: Oh okay got it thank you!

nrajanee (approved):

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -74,14 +81,18 @@ message SearcherEvent {
...
+    // experiment is inactive
+    ExperimentInactive experiment_inactive = 8;
   }
 }
 
 // ValidateAfterOperation means the trial should train and validate after
 // training the given length.
 message ValidateAfterOperation {
+  // trial_id is the id of the trial to close.
+  string trial_id = 1;
```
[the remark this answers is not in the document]
mpkouznetsov: Maybe @dzhu has an idea. We seem to be using those somewhat interchangeably. Calling it `trial_id` here indicates to the user that it identifies the trial. `RequestID` seems to be more of an internal usage. In other words, within the master code it identifies a searcher request but for the user it identifies the trial.

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,183 @@
...
+    @abstractmethod
+    def on_trial_created(self, trial_id: uuid.UUID) -> List[Operation]:
+        """
+        on_trial_created informs the searcher that a trial has been created
+        as a result of Create operation.
+        """
+        pass
+
+    @abstractmethod
+    def on_validation_completed(self, metric: float) -> List[Operation]:
```
liamcli, line 135: we need trial_id and operation here as well.  as an example, for multi-fidelity bayesian methods, we need to know the hparams as well as the training length corresponding to the metric in order to update the model

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,112 @@
...
+                            f" {e.validationCompleted.metric})"
+                        )
+                        if e.validationCompleted.metric is None:
+                            raise RuntimeError(
+                                "validationCompleted event is invalid without a metric"
+                            )
+                        operations = self.search_method.on_validation_completed(
+                            e.validationCompleted.metric
+                        )
+                    elif e.experimentInactive:
```
liamcli, line 93: what would generate the event master side?

harness/determined/searcher/search_method.py
mpkouznetsov: `trial_id` is omitted by mistake, I will add it.
I see `op` in the PRD as well but I am realizing I don't understand the purpose; also which operation are we talking about?

harness/determined/searcher/search_runner.py
mpkouznetsov: This relates to both of your questions.

Looking at the master logic, I see that Searcher.ValidationCompleted does take an operation, of `ValidateAfter` type. It is used to confirm that the SearchMethod is only notified once about the completion of `ValidateAfter` for a specific trial at a specific length. It is not passed on to the SearchMethod.
This event is generated here https://github.com/determined-ai/determined/blob/fb4dba510677e9a6472e92de2e2a4b7da98e516f/master/internal/api_trials.go#L901
and that is called from https://github.com/determined-ai/determined/blob/fb4dba510677e9a6472e92de2e2a4b7da98e516f/harness/determined/core/_searcher.py#L85

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -1,36 +1,89 @@
...
+    def on_trial_exited_early(
+        self, trial_id: uuid.UUID, exit_reason: ExitedReason
+    ) -> List[Operation]:
+        logging.warning(f"Trial {trial_id} exited early: {exit_reason}")
+        return [Shutdown()]
+
+    def initial_operations(self) -> List[Operation]:
+        logging.info("initial_operations")
+        values: List[HparamValue] = []
+        for name, value in self.hyperparameters.items():
```
liamcli, line 70: since we are not relying on the master to sample hyperparameters, i feel like we can assume HparamValue is a nested dictionary of const hyperparameters not deal with the other types that we have master side.

master/internal/experiment.go
```diff
@@ -142,6 +141,12 @@ func newExperiment(master *Master, expModel *model.Experiment, taskSpec *tasks.T
 		telemetry.ReportExperimentCreated(master.system, expModel)
 	}
 
+	method := searcher.NewSearchMethod(conf.Searcher())
```
dzhu, line 144: Any particular reason to move this?

harness/determined/searcher/search_runner.py
```diff
@@ -0,0 +1,112 @@
...
+                        logging.info(
+                            f"experiment {experiment_id} is inactive"
+                            f" state={e.experimentInactive.experimentState}"
+                        )
+                        experiment_is_active = False
+                        break
+                    else:
+                        raise RuntimeError(f"Unsupported event {e}")
+                    bindings.post_PostSearcherOperations(
```
dzhu, line 102: It doesn't matter a whole lot, but I was angling for the operations from multiple events to be batched together, and it sounded like that was fine, but it's not happening here.

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,183 @@
...
+            bindings_exited_reason
+            == bindings.v1TrialExitedEarlyExitedReason.EXITED_REASON_INIT_INVALID_HP
+        ):
+            return cls.INIT_INVALID_HP
+        raise RuntimeError(f"Invalid exited reason: {bindings_exited_reason}")
+
+
+class Operation:
+    def __init__(self) -> None:
+        pass
```
dzhu, line 46: You can just leave this out.

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def _to_searcher_operation(self) -> bindings.v1SearcherOperation:
```
dzhu, line 48: `@abstractmethod`?

harness/determined/searcher/search_method.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+class Shutdown(Operation):
+    def __init__(self) -> None:
+        super().__init__()
+
+    def _to_searcher_operation(self) -> bindings.v1SearcherOperation:
+        return bindings.v1SearcherOperation(
+            shutdown=bindings.v1ShutdownOperation(
+                emptyField=0,
```
dzhu, line 85: You can leave this out (same for any other dummy fields).

harness/determined/searcher/search_method.py
liamcli: the op we pass is the ValidateAfter op that generated the metric which contains the length field that is needed to know how long the trial was trained for to generate the metric.

harness/determined/searcher/search_runner.py
liamcli: i am asking about the `experimentInactive` event

mpkouznetsov: when client asks master for events, the first thing master does is check the status of the experiment. If the experiment is completed the master responds with the list containing a single event, and that is event is experimentInactive. I am not sure this exactly the right thing to do - maybe it is better to return 404 not found. We will figure it out outside this PR.

e2e_tests/tests/experiment/test_custom_searcher.py
mpkouznetsov: 👍

PR #4550, opened 2022-07-14
refactor: move user settings to page [DET-7795]

hkang1:
## Description

Setup work to prep for RBAC user management. Move user settings into a dedicated page from a modal to support advanced RBAC management interfaces.

BEFORE
Screen Shot 2022-07-14 at 2 41 45 PM

AFTER
Screen Shot 2022-07-14 at 2 34 02 PM

## Checklist


gt2345 (approved):
Functionality works great. 
One nit that potentially may change in future is the `Change Name` button is very far right at the page. 
Screen Shot 2022-07-15 at 11 41 35 AM

johnkim-det (approved):
overall looks good, just a few nits:

when there are multiple validation errors for a password, they display horizontally instead of vertically.

minor suggestion to make the folder in `useModal` match the name of the component -- both either `UserSettings` or `SettingsAccount`.

super nit: call it `handleOk` instead of `handleOkay` (i've been wanting to go back and change this to just call `form.validateFields` for a while so i'm glad you were able to address that here lol)

hkang1:
Thanks for the review, everything you suggested makes sense!

> overall looks good, just a few nits:
> 
> when there are multiple validation errors for a password, they display horizontally instead of vertically.
> 
Ah good catch, I've applied the css to the global form errors to be to `block`s

> minor suggestion to make the folder in `useModal` match the name of the component -- both either `UserSettings` or `SettingsAccount`.
>
makes sense
 
> super nit: call it `handleOk` instead of `handleOkay` (i've been wanting to go back and change this to just call `form.validateFields` for a while so i'm glad you were able to address that here lol)

good call

webui/react/src/hooks/useModal/UserSettings/useModalPasswordChange.tsx
```diff
@@ -1,146 +1,139 @@
...
+      required
+      rules={[
+        {
+          message: INCORRECT_PASSWORD_MESSAGE,
+          validator: async (rule, value) => {
+            await login({ password: value ?? '', username });
+          },
+        },
+      ]}
+      validateTrigger={[ 'onSubmit' ]}>
```
johnkim-det, line 51: realized this may not be necessary since we're not using the submit action for the form anymore.

PR #4551, opened 2022-07-14
fix: hide description placeholder when archived

keita-determined:
## Description
[DET-7556]

## Test Plan
- Check description placeholder is blank when the item is archived.
- Also check description placeholder is visible when the item is NOT archived.
You can find these placeholders in `Uncategorized` tab, experience in `Workspace` tab and so on...

## Commentary (optional)
I assumed when disable variable is passed to `InlineEditor` component, all placeholder should be blank.
Is it right? @trentwatt 

## Checklist

[DET-7556]: https://determinedai.atlassian.net/browse/DET-7556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -145,7 +145,7 @@ const InlineEditor: React.FC<Props> = ({
         <textarea
           cols={1}
           maxLength={maxLength}
-          placeholder={disabled ? 'Archived' : placeholder}
+          placeholder={placeholder}
           readOnly={!isEditable}
           ref={textareaRef}
           rows={1}
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -145,7 +145,7 @@ const InlineEditor: React.FC<Props> = ({
         <textarea
           cols={1}
           maxLength={maxLength}
-          placeholder={placeholder}
+          placeholder={disabled ? '' : placeholder}
           readOnly={!isEditable}
           ref={textareaRef}
           rows={1}
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -145,7 +145,7 @@ const InlineEditor: React.FC<Props> = ({
         <textarea
           cols={1}
           maxLength={maxLength}
-          placeholder={disabled ? '' : placeholder}
+          placeholder={disabled ? 'No description' : placeholder}
           readOnly={!isEditable}
           ref={textareaRef}
           rows={1}
--- a/webui/react/src/components/InlineEditor.tsx
+++ b/webui/react/src/components/InlineEditor.tsx
@@ -145,7 +145,7 @@ const InlineEditor: React.FC<Props> = ({
         <textarea
           cols={1}
           maxLength={maxLength}
-          placeholder={disabled ? 'No description' : placeholder}
+          placeholder={disabled ? 'Archived' : placeholder}
           readOnly={!isEditable}
           ref={textareaRef}
           rows={1}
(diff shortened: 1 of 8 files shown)


trentwatt:
> I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
Is it right? @trentwatt

That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?

hkang1:
> > I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
> > Is it right? @trentwatt
> 
> That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?

hmmm it's probably ok. I don't have a good solution but showing a blank (because the border only shows when it's activated?) is a bit misleading too. Wonder if there is a better way to show an empty but disabled state for the inline editor. @alexjdetermined ? :D

alexjdetermined:
> > > I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
> > > Is it right? @trentwatt
> > 
> > 
> > That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?
> 
> hmmm it's probably ok. I don't have a good solution but showing a blank (because the border only shows when it's activated?) is a bit misleading too. Wonder if there is a better way to show an empty but disabled state for the inline editor. @alexjdetermined ? :D

Can we replace the placeholder when in achived state and null to "No description" from "Add description"?

keita-determined:
> > > > I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
> > > > Is it right? @trentwatt
> > > 
> > > 
> > > That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?
> > 
> > 
> > hmmm it's probably ok. I don't have a good solution but showing a blank (because the border only shows when it's activated?) is a bit misleading too. Wonder if there is a better way to show an empty but disabled state for the inline editor. @alexjdetermined ? :D
> 
> Can we replace the placeholder when in achived state and null to "No description" from "Add description"?

image
Something like this?

alexjdetermined:
> > > > > I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
> > > > > Is it right? @trentwatt
> > > > 
> > > > 
> > > > That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?
> > > 
> > > 
> > > hmmm it's probably ok. I don't have a good solution but showing a blank (because the border only shows when it's activated?) is a bit misleading too. Wonder if there is a better way to show an empty but disabled state for the inline editor. @alexjdetermined ? :D
> > 
> > 
> > Can we replace the placeholder when in achived state and null to "No description" from "Add description"?
> 
> image
> 
> Something like this?

Hmm yeah, doesnt work as well there. Maybe using "Archived" is a clearer empth description placeholder message. So users can better understand why they cant edit it. I think its better than "Blank" because users might not immedietly know when they cant edit the description in a blank world.

keita-determined:
> > > > > > I assumed when disable variable is passed to InlineEditor component, all placeholder should be blank.
> > > > > > Is it right? @trentwatt
> > > > > 
> > > > > 
> > > > > That didn't occur to me but I think it is a pretty good assumption. Thoughts @hkang1 ?
> > > > 
> > > > 
> > > > hmmm it's probably ok. I don't have a good solution but showing a blank (because the border only shows when it's activated?) is a bit misleading too. Wonder if there is a better way to show an empty but disabled state for the inline editor. @alexjdetermined ? :D
> > > 
> > > 
> > > Can we replace the placeholder when in achived state and null to "No description" from "Add description"?
> > 
> > 
> > image
> > Something like this?
> 
> Hmm yeah, doesnt work as well there. Maybe using "Archived" is a clearer empth description placeholder message. So users can better understand why they cant edit it. I think its better than "Blank" because users might not immedietly know when they cant edit the description in a blank world.

Probably adding a tooltip on a disabled description would be helpful?

trentwatt:
> Maybe using "Archived" is a clearer empty description placeholder message.

+1

keita-determined:
image
Fixed!

trentwatt (approved):
Yeah this duplicates things a bit, but better I think than encoding the assumption that disabled is always equivalent to archived at the component level.

PR #4560, opened 2022-07-15
feat: add custom search method and event queue [DET-7716]

nrajanee:
## Description
feat: add custom search method and event queue [DET-7716]

## Test Plan

Tested in e2e_tests. 

[DET-7716]: https://determinedai.atlassian.net/browse/DET-7716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -61,25 +53,27 @@ message TrialExitedEarly {
...
 // custom searcher informing it of relevant changes in the state
 // of an experiment
 message SearcherEvent {
+  // incremental id of the event
+  int32 id = 1;
   // The concrete event
   oneof event {
     // an experiment has just been created
-    InitialOperations initial_operations = 1;
+    InitialOpsEvent initial_ops_event = 3;
```
mpkouznetsov, line 68: I know I okayed this but I think we should be consistent: either we add Event suffix to all events or we don't. I don't care much which way we go.

nrajanee: I might suffix to all of them. Will think about if there's better way.

master/internal/experiment.go
```diff
@@ -395,6 +395,34 @@ func (e *experiment) Receive(ctx *actor.Context) error {
...
+	case *apiv1.PostSearcherOperationsRequest:
+		queue, err := e.searcher.GetCustomSearcherEventQueue()
+		if err != nil {
+			ctx.Log().WithError(err).Errorf("Custom Searcher Event Queue was not retrieved.")
+		}
+		// Ensuring that the lastProcessedEvent is the same as what the
+		// (triggered) client processed and the sent the appropriate opertations.
+		// if queue.GetLastProcessedEventId() == -1 then the
+		// first event initial operations triggered the client.
+		if msg.TriggeredByEvent.Id == queue.GetLastProcessedEventID() ||
```
mpkouznetsov, line 407: Wouldn't `msg.TriggeredByEvent.Id` rather be the first _unprocessed_ event? I.e. the event immediately _after_ the last processed event?

nrajanee: oh yes that's true. Thanks for catching that.

master/pkg/searcher/asha.go
```diff
@@ -27,6 +27,7 @@ type (
 	}
 
 	asyncHalvingSearch struct {
+		defaultSearchMethod
```
mpkouznetsov, line 30: I would defer to @nrajanee  and @dzhu and @stoksc on the better design or experiment actor and search method but just want to point out:
if you make the search event queue an attribute of the experiment actor, instead of a search method, you will not have to add `defaultSearchMethod` parent to every existing search method.

master/pkg/searcher/custom_search.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (s *customSearch) GetSearcherEventQueue(context) *SearcherEventQueue {
+	return s.SearcherEventQueue
+}
+
+func (s *customSearch) trialCreated(context, model.RequestID) ([]Operation, error) {
+	s.eventCount++
+	event := experimentv1.SearcherEvent_TrialCreated{
+		TrialCreated: &experimentv1.TrialCreated{},
```
dzhu, line 61: The request IDs and other fields should be filled in for all of these events.

master/pkg/searcher/asha.go
nrajanee: Yes, I agree that having it on the experiment actor would be odd because this is a property that is very specific to the custom searcher. I also did think about maybe having it as a searcher instead. We can talk about this more and I'll fix it in my next ticket. Does that sounds okay? defaultSearchMethod was used in randomSearch for example for default implementation of a few methods so I thought that would be appropriate in this scenario too.

master/pkg/searcher/custom_search.go
nrajanee: I'm going to do this as part of the next ticket.

master/pkg/searcher/asha.go
nrajanee: Leaving this as is, will revisit in next PR.

mpkouznetsov (approved):
To me it looks good as a baseline.

master/pkg/searcher/custom_search.go
```diff
@@ -0,0 +1,133 @@
...
+
+func createSearcherEventQueue() *SearcherEventQueue {
+	return newSearcherEventQueue()
+}
+func (s *customSearch) initialOperations(ctx context) ([]Operation, error) {
+	event := experimentv1.SearcherEvent_InitialOperations{
+		InitialOperations: &experimentv1.InitialOperations{},
+	}
+	searcherEvent := experimentv1.SearcherEvent{
+		Id:    -1,
```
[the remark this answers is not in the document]
dzhu: Oh right, I thought of that issue but then forgot about it, sorry. Yeah, so it's awkward because the interface type of the `Event` field is not exported in the generated protobuf code, for some unknown reason. (I thought it was to ensure that no external code could implement the interface for an unexpected type, but I believe that's still guaranteed if the interface type is exported and the method it contains is not.) I was imagining `Enqueue` would just take `any`, but now I think that actually doesn't work, since it would still need to cast the argument to the unnameable interface type internally. _Sigh_, Go and/or protobuf.

Anyway, that was a long way of explaining the conclusion that I think the caller _does_ need to construct the `experimentv1.SearcherEvent` itself, incredibly annoyingly. But you can just not put anything for the `Id
` field (which will set it to 0).

Also `Enqueue` doesn't need to return an error anymore.

PR #4564, opened 2022-07-16
fix: missing `key` for rows

keita-determined:
## Description

[DET-7827]

## Test Plan
- Check if there is no key warning in  project list in workspace tab

## Commentary (optional)
When I tried to a make duplicate name project, I got an error that says I can't make duplicate name. I believe name is a good candidate for key

## Checklist

[DET-7827]: https://determinedai.atlassian.net/browse/DET-7827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/WorkspaceDetails.tsx
+++ b/webui/react/src/pages/WorkspaceDetails.tsx
@@ -314,6 +314,7 @@ const WorkspaceDetails: React.FC = () => {
               limit: settings.tableLimit,
               offset: settings.tableOffset,
             }, total)}
+            rowKey="id"
             settings={settings}
             size="small"
             updateSettings={updateSettings as UpdateSettings<InteractiveTableSettings>}


trentwatt (commented):
I agree that we don't allow duplicate names, but `project.id` seems a more natural fit.

keita-determined:
> I agree that we don't allow duplicate names, but `project.id` seems a more natural fit.

I think `project.id` isn't available for `InteractiveTable` in this case.

trentwatt:
using `project.id` would correspond to passing `rowKey="id"` since the individual row records are the projects themselves. does that not work?

keita-determined:
It worked with `id`. id wasnt in the columns, so I assumed I cant use it...

PR #4566, opened 2022-07-18
feat: Show active experiment and task count on resources page [DET-7680]

mapmeld:
## Description

Adds a row of numbers for Active Experiments, TensorBoards, JupyterLabs, etc. to the Clusters page. This was previously on the Experiments page but was removed when that became a Project page.

For consistency with other stats on this page, I've put them into the shared Store.

I'm waiting for UI approval, but in the meantime it can go through a technical review.

Screen Shot 2022-07-18 at 10 12 51 AM

## Test Plan

When running experiments and notebooks, this row of mini-cards should appear on the Clusters page. 

When nothing is running, none of these mini-cards should be visible.

## Checklist

### DIFF
--- a/docs/attributions.txt
+++ b/docs/attributions.txt
@@ -1,0 +1,4966 @@
+######################
+ Open Source Licenses
+######################
+
+The following sets forth attribution notices for third-party software
+that may be contained in Determined. We thank the open-source community
+for all of their contributions.
+
+*******
+ WebUI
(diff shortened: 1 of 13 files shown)


trentwatt:
Also think button to start shell disappeared?

Maybe it could go here as well?

gt2345 (approved):
Works great~

PR #4570, opened 2022-07-18
fix: helm option "defaultPassword" caused the deployment to hang [DET-7814]

NicholasBlaskey:
## Description

API changes had caused this option to cause any deployment with a non empty ```defaultPassword``` to hang.

## Test Plan

On a GKE cluster run a deployment with (any recent determined version should be fine) 
```
helm install determined helm/charts/determined/ --set maxSlotsPerPod=2 --set defaultPassword=secret --set detVersion=0.18.4
``` 
Above command can take a minute but it should not hang

Then running
```
det user login admin
det user login determined
```

should both work when ```secret``` is used as the password.

## Commentary (optional)

Moves the python code out of this repo https://github.com/determined-ai/k8s-utility and we should ideally replace the image to something we maintain actively (possible dev image we are making with all determined dependencies?) 

## Checklist

### DIFF
--- a/helm/charts/determined/templates/change-password.yaml
+++ b/helm/charts/determined/templates/change-password.yaml
@@ -28,7 +28,6 @@ spec:
         args:
           - "-c"
           - >-
-            sleep 5 &&
             echo -e {{ .Files.Get "scripts/k8s-password-change.py" | quote }} > change-pw.py &&
             python3 change-pw.py
             {{ .Release.Namespace | quote }} \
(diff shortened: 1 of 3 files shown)


helm/charts/determined/templates/change-password.yaml
```diff
@@ -24,12 +24,16 @@ spec:
...
         args:
-          - "opt/determined/change-password.py"
-          - {{ .Release.Namespace }}
-          - {{ .Release.Name }}
-          - {{ .Values.masterPort | quote }}
-          - {{ .Values.useNodePortForMaster | quote }}
-          - {{ .Values.defaultPassword | default ""}}
+          - "-c"
+          - >-
+            sleep 5 &&
```
eecsliu, line 31: non-blocking - why sleep here?

eecsliu (approved):
Looks good! thanks for the additional code polish.

gh-determined-ai (approved):

PR #4571, opened 2022-07-18
chore: share useModal hook [DET-7781]

hamidzr:
## Description

move the updated useModal hook to shared-web

## Commentary (optional)

saas follow up https://github.com/determined-ai/saas/pull/229

## Checklist


trentwatt (approved):

webui/react/src/shared/hooks/useModal/useModal.tsx
```diff
@@ -20,6 +20,12 @@ export interface ModalHooks {
   modalRef: React.MutableRefObject<ReturnType<ModalFunc> | undefined>;
 }
 
+export type ModalOpen<T> = (modalProps?: ModalFuncProps & { openProps?: T }) => void;
```
hamidzr, line 23: 👀

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpoint.test.tsx
```diff
@@ -6,7 +6,7 @@ import React, { PropsWithChildren, useCallback } from 'react';
 import StoreProvider from 'contexts/Store';
 import { generateTestExperimentData } from 'storybook/shared/generateTestExperiments';
 
-import { ModalCloseReason } from '../useModal';
+import { ModalCloseReason } from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 9: **nit:** wonder why this one didn't take for the `shared/hooks` prefix?

webui/react/src/hooks/useModal/Model/useModalModelDelete.tsx
```diff
@@ -12,7 +12,7 @@ import handleError from 'utils/error';
 
 import useModal, {
   CANNOT_DELETE_MODAL_PROPS, ModalHooks as Hooks, ModalCloseReason,
-} from '../useModal';
+} from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 15: ditto

webui/react/src/hooks/useModal/Model/useModalModelVersionDelete.tsx
```diff
@@ -12,7 +12,7 @@ import handleError from 'utils/error';
 
 import useModal, {
   CANNOT_DELETE_MODAL_PROPS, ModalHooks as Hooks, ModalCloseReason,
-} from '../useModal';
+} from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 15: ditto

webui/react/src/hooks/useModal/UserSettings/useModalNameChange.tsx
```diff
@@ -5,7 +5,7 @@ import { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
 import { patchUser } from 'services/api';
 import handleError from 'utils/error';
 
-import useModal, { ModalHooks } from '../useModal';
+import useModal, { ModalHooks } from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 8: ditto

webui/react/src/hooks/useModal/UserSettings/useModalPasswordChange.tsx
```diff
@@ -5,7 +5,7 @@ import { useStore } from 'contexts/Store';
 import { login, setUserPassword } from 'services/api';
 import handleError from 'utils/error';
 
-import useModal, { ModalHooks } from '../useModal';
+import useModal, { ModalHooks } from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 8: ditto

webui/react/src/hooks/useModal/UserSettings/useModalUserSettings.tsx
```diff
@@ -6,7 +6,7 @@ import { useStore } from 'contexts/Store';
 import useModalNameChange from 'hooks/useModal/UserSettings/useModalNameChange';
 import useModalPasswordChange from 'hooks/useModal/UserSettings/useModalPasswordChange';
 
-import useModal, { ModalHooks } from '../useModal';
+import useModal, { ModalHooks } from '../../../shared/hooks/useModal/useModal';
```
hkang1, line 9: ditto

webui/react/src/hooks/useSettings.ts
```diff
@@ -11,10 +11,10 @@ import { clone, hasObjectKeys, isBoolean, isEqual, isNumber,
 import { Storage } from 'shared/utils/storage';
 import handleError from 'utils/error';
 
+import usePrevious from '../shared/hooks/usePrevious';
 import { Primitive, RecordKey } from '../shared/types';
```
hkang1, line 15: ditto

webui/react/src/shared/hooks/useModal/useModal.tsx
hkang1: I thought about this during the modal refactor and I didn't take the time to do this then, but what do you think about having `modalOpen` just take in a 2nd param of custom props/context?
```
modalOpen: (modalProps?: ModalFuncProps = {}, context?: RecordUnknown) => void;
```

hkang1 (commented):
changes look good! Added a comment for `modalOpen` and making the hook open more flexible up front

webui/react/src/hooks/useSettings.ts
hamidzr: https://github.com/determined-ai/determined/pull/4581 let's land that and then I make sure this pr doesn't readd those

webui/react/src/shared/hooks/useModal/useModal.tsx
hamidzr: and you're thinking the `context` would be used the same way openProps is used? should we still ask for the user to defien the type instead of `RecordUnknown` we could default to RecordUnknown instead of just `unknown`

hamidzr: if so that works but I wouldn't see much of a difference. looping in @johnkim-det as well since we had some discussion on this [here](https://github.com/determined-ai/saas/pull/205)

johnkim-det: yea we considered having two params. personally i preferred not having to pass undefined or empty object if there are no 'modalProps' being used.

hkang1: hmmm good points. My thought is to achieve the following:
1. Preserve the modalProps as something that continuously gets exposed during composition.
2. allow additional context to be passed in for the newly composed modal hook.

What about something something like this?
```
interface ModalProps extends ModalFuncProps {
  context?: RecordUnknown;
}

...
modalOpen: (modalProps: ModalProps = {}) => void;
...
```

hamidzr: sounds good to me. 
to clarify the changes:
this is the same as we have here with the differences being:
- renaming the key to `context` instead of `openProps`
- apply the change to the original `ModalHooks` instead of adding a new type

hkang1: yeah do those two changes sound ok to you two?

PR #4573, opened 2022-07-18
feat: add sort and order to users api [DET-7828]

gt2345:
## Description

Modify get users API: 
- add support for `sortBy`
- add support for `orderBy` (asc / desc)
- add support for `offset` and `limit`

## Checklist


ioga (approved):
lgtm

jfyi I have a catch-all ticket for pagination improvements, e.g. to address pagination that is done after SQL (instead of in SQL): https://determinedai.atlassian.net/browse/DET-7818

PR #4575, opened 2022-07-19
ci(test-cli): migrate to gha workflow

loksonarius:
## Description

Migrating test-cli CircleCI workflow into an independent GHA workflow as
part of general migration out of CircleCI. The steps are pretty much the
same with the following differences from the current CircleCI workflow:

- test-cli will now run all tests across both Windows and Ubuntu for
  each Python version
- the test environment is entirely setup with just Python and pip -- no
  conda installation of anything

## Test Plan

Run workflow in both CircleCI and GHA with a PR and double-check output for differences.

## Checklist

### DIFF
--- a/.github/workflows/test-cli.yml
+++ b/.github/workflows/test-cli.yml
@@ -1,0 +1,54 @@
+---
+name: "Test det CLI"
+
+on:
+  pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'docs/*'
+  push:
+    branches:
+      - 'main'
+      - 'releases/**'
+
+jobs:
+  test-cli:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.7", "3.8", "3.9"]
+        os: ["ubuntu-latest", "windows-latest"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install det dependencies
+        run: |
+          python --version
+          python -m pip install --upgrade pip
+          pip --version
+          # Windows needs special treatment of cryptography for unknown reasons.
+          # LMDB also has special requirements that only apply to windows (patch-ng).
+          pip install wheel cryptography patch-ng
+          cd harness; python setup.py bdist_wheel -d ../build
+      - name: Install determined
+        run: |
+          pip install --find-links build determined==0.18.5.dev0
+          # Ensure Determined cli can run without installing cli test requirements
+          det --help
+          pip freeze --all
+          # Allow this to fail, but it is useful for debugging.
+          pip check || true
+      - name: Install test dependencies
+        run: |
+          pip install setuptools_scm
+          pip install -r harness/tests/requirements/requirements-cli.txt
+          pip freeze --all
+          # Allow this to fail, but it is useful for debugging.
+          pip check || true
+      - name: Run tests
+        run: |
+          pytest harness/tests/cli
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -20,6 +20,8 @@ values =
 
 [bumpversion:file:.circleci/config.yml]
 
+[bumpversion:file:.github/workflows/test-cli.yml]
+
 [bumpversion:glob:*/setup.py]
 
 [bumpversion:glob:*/*/__version__.py]


loksonarius:
So comparing the output of the two runs for Windows@Py3.7, it seems the CircleCI workflow is hiding some warnings that show up in GHA -- the same amount of tests still pass and skip though.
Ref Circle: https://app.circleci.com/pipelines/github/determined-ai/determined/24556/workflows/1785af75-9a15-4075-9799-0c85f03b4d6a/jobs/795328
Ref GHA: https://github.com/determined-ai/determined/runs/7400922371?check_suite_focus=true

Output for the Ubuntu runners was the same, and run time was comparable across both platforms (sub-minute at worst).
Windows run times were comparable as well within a minute of each other for Python 3.7.

.github/workflows/test-cli.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Install det dependencies
+        run: |
+          python --version
+          python -m pip install --upgrade pip
+          pip --version
+          pip install wheel cryptography patch-ng
+          cd harness; python setup.py bdist_wheel -d ../build
+      - name: Install determined
+        run: |
+          pip install --find-links build determined==0.18.5.dev0
```
loksonarius, line 37: Working right now on figuring out the source of truth for this value -- if anyone knows, please comment 🙏

rb-determined-ai: you mean the version?  We maintain it with bumpversion, see `.bumpversion.cfg` I think.

loksonarius: So I should just use the value of `current_version` there?

.github/workflows/test-cli.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install det dependencies
+        run: |
+          python --version
+          python -m pip install --upgrade pip
+          pip --version
+          pip install wheel cryptography patch-ng
```
rb-determined-ai, line 33: there used to be context as to why these two lines were here... maybe keep the comment that explained them?

If you want to install the extra deps on linux too, that's fine.  But leave the explanations please.

.github/workflows/test-cli.yml
```diff
@@ -0,0 +1,49 @@
...
+          cd harness; python setup.py bdist_wheel -d ../build
+      - name: Install determined
+        run: |
+          pip install --find-links build determined==0.18.5.dev0
+          det --help
```
rb-determined-ai, line 38: preserve comment please

.github/workflows/test-cli.yml
```diff
@@ -0,0 +1,49 @@
...
+          pip freeze --all
+          pip check || true
```
rb-determined-ai, line 40: ditto

.github/workflows/test-cli.yml
rb-determined-ai: I mean, I think it you configure bumpversion to update this file, then it should be safe to have the literal `0.18.5.dev0` written into the file.

loksonarius: Ah, that's an idea as well. While I don't like the idea of potentially adding noise when version bumping, it seems much more reasonable to make the relationship between the files explicit by adding it to that config file. Thanks for the suggestion!

dannysauer: (quietly mentions that having a versioned version file is kinda silly when the repo metadata already tracks versions)

rb-determined-ai (approved):

dannysauer (approved):
I tend to think it's the base image maintainer's problem to keep pip up to date, and not something we should spend action cycles on (re: `pip install --upgrade pip`).  But that's just something to consider for future updates; I'm assuming this is a near-direct copy of existing code for the first pass.  So we can leave it.  This LGTM otherwise.  Thanks!

rb-determined-ai:
> I tend to think it's the base image maintainer's problem to keep pip up to date, and not something we should spend action cycles on (re: pip install --upgrade pip)

I would agree with you in principle, but maybe the only useful coverage these tests have ever given us is early warning when upstream dependencies have broken.

Upgrading pip itself hasn't empirically proven to be useful, but in the same vein as always updating the latest python dependencies every test run, I think this is a decent solution.

I'll admit that there's no particular reason why the dependency install testing needs to be part of test-cli, and it wouldn't hurt my feelings to move it to a dedicated test, and leave the pip installing in the ci base image.

PR #4274, opened 2022-06-08
fix: jupyterlab modal stale values, not sending values to api

trentwatt:
## Description

[DET-7478] [DET-7389]

fixes some staleness with jupyterlab modal in both the ui and values that get sent to API 

also does some refactoring/cleanup of the adjacent code (mainly the older parts of the code, from before the recent refactor)

## Before

jupyterlab modal has stale values for compute pool (clicking in dropdown doesnt update value):

https://user-images.githubusercontent.com/28383080/172557596-60342111-e605-4063-ad30-f9ee647d37e7.mov

and slots (pressing enter or clicking away causes the current input value to be clobbered by previous):

https://user-images.githubusercontent.com/28383080/172557934-b56badca-dbfa-47b8-b2c4-efa1d496e4b8.mov

and sends empty/stale values to api:

https://user-images.githubusercontent.com/28383080/172558702-7bd90a52-d4b2-4974-bdad-54a54090b8a1.mov

there is a warning logo on the top left, which isn't suitable.

also there was some accumulated messy code (going way back before recent refactor)

`useJupyterlabModal` calls `useJupyterLab` and `useJupyterLabForm`

`useJupyterLab` is actually just some util functions so i moved them to `utils/jupyter`

`useJupyterLabForm` is a wrapper around a reducer that does `useStorage` to manage state. we prefer `useSettings` now to `useStorage`

## After

these are hopefully all now resolved.

https://user-images.githubusercontent.com/28383080/172563126-3d525796-819a-461f-8806-3f254a9bcaa4.mov

## Test Plan

Test creation of jupyterlab with all the various options, verify that the POST payloads are correct.

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](../docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.

[DET-7478]: https://determinedai.atlassian.net/browse/DET-7478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[...]

PR #4582, opened 2022-07-19
fix: allow zero slots for JupyterLab in modal

trentwatt:
## Description

we were restoring the default when slots input was 0. this fixes that.

## Test Plan

do `devcluster`, go to JupyterLab modal, click down arrow on slots, should allow a selection of zero. 

## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
+++ b/webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
@@ -360,7 +360,7 @@ const JupyterLabForm: React.FC<FormProps> = ({ updateFields, fields }: FormProps
             <InputNumber
               defaultValue={fields.slots !== undefined ? fields.slots : DEFAULT_SLOT_COUNT}
               max={resourceInfo.maxSlots === -1 ? Number.MAX_SAFE_INTEGER : resourceInfo.maxSlots}
-              min={resourceInfo.hasAux ? 0 : 1}
+              min={0}
               value={fields.slots}
               onChange={(value) => updateFields?.({ slots: value })}
             />
--- a/webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
+++ b/webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
@@ -277,7 +277,7 @@ const JupyterLabForm: React.FC<FormProps> = ({ updateFields, fields }: FormProps
     const hasSlotsPerAgent = maxSlots !== 0;
     const hasComputeCapacity = hasSlots || hasSlotsPerAgent;
     if (hasAuxCapacity && !hasComputeCapacity) updateFields?.({ slots: 0 });
-    if (hasComputeCapacity && !fields.slots) updateFields?.({ slots: DEFAULT_SLOT_COUNT });
+    if (hasComputeCapacity && fields.slots == null) updateFields?.({ slots: DEFAULT_SLOT_COUNT });
 
     return {
       hasAux: hasAuxCapacity,
@@ -360,7 +360,7 @@ const JupyterLabForm: React.FC<FormProps> = ({ updateFields, fields }: FormProps
             <InputNumber
               defaultValue={fields.slots !== undefined ? fields.slots : DEFAULT_SLOT_COUNT}
               max={resourceInfo.maxSlots === -1 ? Number.MAX_SAFE_INTEGER : resourceInfo.maxSlots}
-              min={0}
+              min={resourceInfo.hasAux ? 0 : 1}
               value={fields.slots}
               onChange={(value) => updateFields?.({ slots: value })}
             />


webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
```diff
@@ -360,7 +360,7 @@ const JupyterLabForm: React.FC<FormProps> = ({ updateFields, fields }: FormProps
             <InputNumber
               defaultValue={fields.slots !== undefined ? fields.slots : DEFAULT_SLOT_COUNT}
               max={resourceInfo.maxSlots === -1 ? Number.MAX_SAFE_INTEGER : resourceInfo.maxSlots}
-              min={resourceInfo.hasAux ? 0 : 1}
```
hamidzr, line 363: is this because hasAux is inaccurate? if not what's the issue here. 
I should re-read what hasAux advertises. but if it means that the RP has no support for auxiliary tasks then this won't be able to run 🤔 maybe it's still fine to allow this

hamidzr: is there a problem description or jira?

eecsliu: So the main issue is that kubernetes doesn't have resource pools, so it's forcing users to use a slot in kubernetes for notebooks. Also on latest-master, if you select the compute pool, it locks you into using at least one slot (which might be the correct behavior, but shouldn't we still allow users to run zero slot tasks?)

eecsliu: @NicholasBlaskey mentioned that the problem isn't with hasAux, so this might not be the correct solution.

trentwatt: ah okay, looks like the issue was not with `min`, but that we were restoring the default when `slots` input was `0`

@hamidzr you were right, that logic is fine. nicholas and eric were reporting an issue were they couldnt give a notebook zero slots.

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
```diff
@@ -277,7 +277,7 @@ const JupyterLabForm: React.FC<FormProps> = ({ updateFields, fields }: FormProps
     const hasSlotsPerAgent = maxSlots !== 0;
     const hasComputeCapacity = hasSlots || hasSlotsPerAgent;
     if (hasAuxCapacity && !hasComputeCapacity) updateFields?.({ slots: 0 });
-    if (hasComputeCapacity && !fields.slots) updateFields?.({ slots: DEFAULT_SLOT_COUNT });
```
hamidzr, line 280: ahh this `!` meant so that `0` slots was being updated to the default count?

trentwatt: yes 🤪

hamidzr: we've had issues with the logic around this a couple of times IIRC. maybe there is a simpler way of writing this so it's less error-prone. we probably wanna leave that off for a separate work.

hamidzr (approved):
the change makes sense to me. if you add a repro I can help verify that it's fixed

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
hamidzr: [this change](https://github.com/determined-ai/determined/pull/4582/commits/4484471792da857951494d50205574846d481194) would have the described behavior 🚀

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
trentwatt: [this one is on me](https://github.com/determined-ai/determined/pull/4274/files#diff-07881315d579e2a52d39780a88700aac54a27651c9d743de0acef882f3243d10R264-R265), it just didn't occur to me that `0` was a valid number for slots at the time.

hamidzr: could a linter force us to not use `! or !!` on non-booleans?

trentwatt: root cause here was a bad assumption about slots, not a misuse of `!` operator but ¯\_(ツ)_/¯  i guess it might've helped anyway

trentwatt: there are a lot of settings, e.g. `workspaceId` and `projectId` where `0` is treated as a null value, so being able to do `if (!project.id)` is really nice for those.

PR #4586, opened 2022-07-19
fix: zero slot tasks k8s using wrong image and exposing all GPUs [DET-7808]

NicholasBlaskey:
## Description

Zero slot tasks would use the GPU image before this fix. 

In addition zero slot tasks would end up getting exposed all GPUs (not sure if all in the cluster or on the node but still a problem either way).

## Test Plan

Create a EKS cluster with a GPU

then run and verify that we can see the GPUs (and the image used has ```-gpu-``` in the name)
```
det cmd run --config="resources.slots=1" nvidia-smi
```

now run a zero slot command and verify we do not see any GPUs (should get command not found with image with ```-cpu-``` in the name)
```
det cmd run --config="resources.slots=0" nvidia-smi
```

Then start a notebook verify the image has ```-cpu-``` in the name
```
det notebook start --config="resources.slots=0"
```

Then run the following snippet in a notebook and verify that the output is ```0``` 
```python
import torch
torch.cuda.device_count()
```

## Checklist

### DIFF
--- a/docs/release-notes/zero-slot-k8s.txt
+++ b/docs/release-notes/zero-slot-k8s.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  Kubernetes: Fix an issue where zero slot trials would use the GPU image instead of the CPU image.
+-  Kubernetes: Fix an issue where zero slot trials would incorrectly be exposed all GPUs.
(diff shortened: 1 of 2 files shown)


gh-determined-ai (approved):

ioga (approved):
awesome

PR #4587, opened 2022-07-19
ci(test-e2e): move optional jobs into own workflow

loksonarius:
## Description

Currently, all long-running jobs are only manually triggerable. We still
want this to be the case, but because of a lack of integration between
GH status checks and CircleCI pipeline reporting, we can't specify what
subset of the `test-e2e` job is okay not to run.

As a workaround, we're moving those manually-runnable jobs into a
separate workflow that'll be initiated the same as the current
`test-e2e` workflow, but will be marked as entirely optional for status
checks on PRs to the monorepo. This allows us to not have to circumvent
branch protections for PRs that don't require running those manual jobs.

## Test Plan

Make a PR and verify that both `test-e2e` and `test-e2e-longrunning` are correctly configured by CircleCI.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2277,129 +2277,6 @@ workflows:
               mark: ["e2e_gpu"]
               slack-mentions: ["${SLACK_USER_ID}"]
 
-      - request-dev-deploy:
-          type: approval
-          filters: *upstream-feature-branch
-
-      - deploy:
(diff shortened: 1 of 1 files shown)


loksonarius:
What a beautiful day for a CircleCI outage 😩

ioga (approved):

PR #4589, opened 2022-07-20
fix: ignore order for Set isEqual [DET-7822]

trentwatt:
## Description
[DET-7822]

`Sets` are the same if they have the same size and their elements are equal.

Note: with the logic in this PR, 
```
isEqual(new Set([[1]]), new Set([[1]])) === false
``` 

This is in keeping with how `Sets` see things though: `Sets` see different arrays containing the same elements as distinct, e.g.
```
> new Set([[1], [1]]).size 
2
```
so that seemed the appropriate way to handle it. 

It would not be possible to implement custom `isEquals` for individual set elements while maintaining performance characteristics one would expect for an `isEqual(a: Set, b: Set)` function (caveat: runtime for `Set` is not part of the JS spec).

## Checklist

[DET-7822]: https://determinedai.atlassian.net/browse/DET-7822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/shared/utils/data.test.ts
+++ b/webui/react/src/shared/utils/data.test.ts
@@ -185,6 +185,20 @@ describe('Data Utilities', () => {
         ],
         output: false,
       },
+      /**
+       * for now isEquals for set elements respects how the set views
+       * equality of its members, as opposed to how isEquals views them
+       * otherwise you would have the following:
+       * > a.forEach(x => b.delete(x)); isEquals(a, b)
+       * true
+       */ 
+      {
+        input: [
+          new Set([[ 1 ]]),
+          new Set([[ 1 ]]),
+        ],
+        output: false,
+      },
     ];
     tests.forEach((test) => {
       const [ a, b ] = test.input;
--- a/webui/react/src/shared/utils/data.test.ts
+++ b/webui/react/src/shared/utils/data.test.ts
@@ -143,8 +143,8 @@ describe('Data Utilities', () => {
       { input: [ Symbol('bit'), 123 ], output: false },
       {
         input: [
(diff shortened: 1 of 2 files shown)


webui/react/src/shared/utils/data.test.ts
```diff
@@ -143,8 +143,8 @@ describe('Data Utilities', () => {
       { input: [ Symbol('bit'), 123 ], output: false },
       {
         input: [
-          new Set([ 'abc', 123, Symbol('xyz') ]),
-          new Set([ 'abc', 123, Symbol('xyz') ]),
```
trentwatt, line 147: this tested ended up failing because `Symbol('xyz') !== Symbol('xyz')`. the whole point of Symbol is that separate instantiations are not considered equal, so it seems appropriate.

on a related note, it might make more sense for `isEqual(Symbol('xyz'), Symbol('xyz')) === false`

hamidzr (commented):
thanks for taking this on.

question about what we want isEqual to do. 
```
if (isEqual([1], [1]))

then we expect ?

isEqual(Set([[1]]) Set([[1]])) === true
```

either way, could you add a test for the following so we can put the expectation in code and make sure it doesn't change without intention?
`isEqual(new Set([[1]]), new Set([[1]])) === false`

hamidzr (approved):
it looks good to me. we had some chat on another channel we're not super clear on what the exact behavior and usage of isEqual but this is a good improvement as is :shipit: 
let's just add that test

PR #4594, opened 2022-07-20
fix: Drop rendezvous interface warning (FOUNDENG-139)

jerryharrow:
## Description

On Slurm we attempt prioritize an eth* interface for rendezvous to make sure we have a reachable address for the master.  It appears common for systems to not have an eth* interface, so drop the warning to avoid the noise.

If the eth* interface did not have a valid address, the fallback to hostname would not have been applied, so fixed
that as well.

## Test Plan

No behavior change, just changed warning logs.

## Checklist

### DIFF
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -88,16 +88,13 @@ def do_rendezvous_slurm(
     assert num_peers_str, "Unable to complete rendezvous without SLURM_NPROCS"
     num_peers = int(num_peers_str)
 
-    rendezvous_ip, resolution_error = None, None
+    rendezvous_ip = socket.gethostbyname(socket.gethostname())
     for rendezvous_iface in rendezvous_ifaces():
         try:
             rendezvous_ip = get_ip_from_interface(rendezvous_iface)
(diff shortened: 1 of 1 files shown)


harness/determined/exec/prep_container.py
```diff
@@ -88,16 +88,13 @@ def do_rendezvous_slurm(
...
     for rendezvous_iface in rendezvous_ifaces():
         try:
             rendezvous_ip = get_ip_from_interface(rendezvous_iface)
             break
         except ValueError as e:
-            resolution_error = e
-    else:
-        logging.warning(f"falling back to naive ip resolution after:\n\t{resolution_error}")
-        rendezvous_ip = socket.gethostbyname(socket.gethostname())
+            logging.warning(f"Unable to resolve ip for {rendezvous_iface}: {str(e)}")
```
stoksc, line 97: does this not always print, still? I think it would

stoksc: wait no, my brain failed to read a unified diff ..

stoksc (approved):
lgtm

PR #4597, opened 2022-07-20
feat: allow DET_PASS and DET_USER to skip need for login [DET-7025]

NicholasBlaskey:
## Description

Allows for setting ```DET_PASS``` and ```DET_USER``` to skip needing to login.

## Test Plan

e2e tests + 

Should print admin 
```
DET_USER=admin DET_PASS= det user whoami
```

Should still work since it uses the cached token
```
DET_USER=admin DET_PASS=bad det user whoami
```

Should print determined
```
DET_USER=admin DET_PASS= det -u determined user whoami
```

## Checklist


gh-determined-ai (approved):

stoksc:
isn't the usual approach here something like `cat passfile | det user login determined --password-stdin` to avoid something like `DET_PASS=xyz det user login determined` showing up in CI as easily?

e2e_tests/tests/cluster/test_users.py
```diff
@@ -367,6 +367,25 @@ def test_login_as_non_existent_user(clean_auth: None) -> None:
     assert child.exitstatus != 0
 
 
+@pytest.mark.e2e_cpu
+def test_login_det_pass_environment(clean_auth: None) -> None:
+    creds = create_test_user(ADMIN_CREDENTIALS, True)
+
+    try:
+        os.environ["DET_PASS"] = creds.password
```
rb-determined-ai, line 375: nit: if this were to fail, your finally would also fail.  So technically, it should be:
```python
# setup before try
os.environ["DET_PASS"] = creds.password
try:
    # do stuff
    ...
finally:
    # cleanup in finally
    del os.environ["DET_PASS"]
```

stoksc (approved):
lgtm, but I'll let @rb-determined-ai look too

stoksc:
talked offline, let's just do both one day

harness/determined/cli/user.py
```diff
@@ -69,10 +70,14 @@ def log_in_user(parsed_args: Namespace) -> None:
     else:
         username = parsed_args.username
 
-    message = "Password for user '{}': ".format(username)
-
+    if "DET_PASS" in os.environ:
+        print("Using the enviornment variable DET_PASS for a user's password")
+        password = os.environ["DET_PASS"]
+    else:
+        message = "Password for user '{}': ".format(username)
```
rb-determined-ai, line 77: another nit: when you touch old `.format()` coded, can you update it to use f-strings?

rb-determined-ai (commented):
Implementation looks good to me.

But overall, is it a gotcha that you still have to call `det user login` at all, when `DET_PASS` is set?  For example, we also have a `DET_USER` setting that we honor in a lot of places.  Why isn't it enough to set DET_USER and DET_PASS and not login?

NicholasBlaskey:
The needing to call ```det login``` was a really good point. It was a case of being so zoomed in that you don't think of the bigger picture for me.

So now ```det login``` can't take any environments variables (like before this change) to give a way to override the ```DET_USER``` and ```DET_PASS``` variables that will just work

gh-determined-ai (approved):
lgtm

harness/determined/common/api/authentication.py
```diff
@@ -61,6 +61,7 @@ def _init_session(
     ) -> Session:
         session_user = (
             requested_user
+            or util.get_det_username_from_env()
             or self.token_store.get_active_user()
```
rb-determined-ai, line 65: oh man, I have to think so hard when I get close to this code.

I _think_ that this ordering differs from what you said in your PR comments.  I think this is going to cause the det login user (set in token_store) to never get used, making the `det user login` override you talked about ineffective.

harness/determined/common/util.py
```diff
@@ -58,6 +58,14 @@ def get_container_user_token() -> Optional[str]:
     return os.environ.get("DET_USER_TOKEN")
 
 
+def get_det_username_from_env() -> Optional[str]:
+    return os.environ.get("DET_USERNAME")
```
rb-determined-ai, line 62: question: why use `DET_USERNAME` here when we already use `DET_USER` in other places?

docs/interfaces/commands-and-shells.rst
```diff
@@ -107,6 +107,17 @@ Each noun also provides a ``help`` verb that describes the possible verbs for th
...
 
+Environment Variables
+=====================
+
+-  ``DET_MASTER``: The network address of the master of the Determined installation. The value can
+   be overridden using the ``-m`` flag.
+-  ``DET_USERNAME``: Specifies the current Determined user. The value can be overridden using the ``-u``
+   flag.
+-  ``DET_PASSWORD``: Specifies the password of the current Determined user. This value can be overriden
+   by running ``det login <username>`` then entering the different password in the prompt.
```
rb-determined-ai, line 118: Can this be rephrased to make it clearer that `det user login` is the preferred mechanism, but DET_USERNAME and DET_PASSWORD can be used in a pinch where non-interactive behavior is required?

harness/determined/common/util.py
NicholasBlaskey: I switched it to ```DET_USER``` and ```DET_PASS```

I made it ```DET_USERNAME``` to avoid conflicting with ```DET_USER``` but it seems that it is much more sensible to just also use ```DET_USER```

harness/determined/common/api/authentication.py
NicholasBlaskey: I simplified the logic and made it more sane. 

Now both ```DET_USER``` and ```DET_PASS``` must be set to take effect with ```-u``` overriding the environment variables.

docs/release-notes/det-pass.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Improvements**
+
+-  CLI: Commands now support setting the environment variables ``DET_USER`` and ``DET_PASS`` to skip
```
rb-determined-ai, line 5: gosh, this is the tiniest nit, but... can you make this say "The ``det`` CLI command now supports" instead?

My reason is that "commands" refers to a special thing in our system, so upon first reading it, I skimmed a bit too fast and thought you were talking about `det cmd run` type of "commands".

rb-determined-ai (approved):
great work!

Especially in a messy part of the system.

PR #4604, opened 2022-07-21
ci: test downstream

trentwatt:
## Description

adds `intg-test` that checks out saas and tests saas web

test pr against this branch:
- [#4616 upstream branch with shared changes](https://github.com/determined-ai/determined/pull/4616): should fail because it breaks saas tests [yep](https://app.circleci.com/pipelines/github/determined-ai/determined/24792/workflows/5f6feea2-8fb0-4c25-99f4-22a89b7b286c/jobs/805118) ✅
- [#4617 fork branch with shared changes](https://github.com/determined-ai/determined/pull/4617): should fail due to lack of CI permissions-- requires PR from upstream branch [yep](https://app.circleci.com/pipelines/github/determined-ai/determined/24791/workflows/d47eb68e-1e00-493b-a6a9-69fc7a091339/jobs/805081) ✅
- [#4618 fork branch w/o shared changes](https://github.com/determined-ai/determined/pull/4618): should pass because the downstream test is skipped [yep](https://app.circleci.com/pipelines/github/determined-ai/determined/24790/workflows/89a9dadb-dd38-4a9b-962b-9255384b218f/jobs/805064) ✅

for tests, i had to make a temporary switch to simulate it being `upstream/master`.  i will switch it back for merge, so it might trigger subsequent checks on the linked PRs with erroneous results. the `yeps` link to the actual workflow runs for future reference.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ commands:
       - run:
           name: check for shared changes
           command: |
-            MERGE_BASE="$(git merge-base upstream/test-downstream HEAD)"
+            MERGE_BASE="$(git merge-base upstream/master HEAD)"
             DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE":webui/react/src HEAD:webui/react/src)"
             if [ "$DIFF_DIRS" = "shared" ] ; then
               echo "shared code changed, running downstream test"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1268,6 +1268,14 @@ jobs:
       - store_artifacts:
           path: webui/react/coverage/lcov-report
 
+    test-unit-downstream:
+    docker:
+      - image: cimg/node:16.14
+    steps:
+      - checkout-with-sm
+      - install-python
+      - run: cd webui/react/src/shared && ./bin/test.py --test-local --repos saas 
+
   build-storybook:
     docker:
       - image: cimg/node:16.14.2
@@ -2077,6 +2085,7 @@ workflows:
       - test-unit-go
       - test-unit-react
       - test-unit-harness-cpu
+      - test-unit-downstream
       # Exclude gpu tests for now since we do not have access to those instance types on circleci.
       #- test-unit-harness-gpu
       - test-unit-harness-tf2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,7 @@ commands:
           name: check for shared changes
           command: |
             MERGE_BASE="$(git merge-base upstream/test-downstream HEAD)"
-            DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE":webui/react/src HEAD:webui/react/src")"
+            DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE":webui/react/src HEAD:webui/react/src)"
             if [ "$DIFF_DIRS" = "shared" ] ; then
               echo "shared code changed, running downstream test"
             else
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ commands:
       - run:
           name: check for shared changes # must be run after skip-if-docs-only
           command: |
-            MERGE_BASE="$(git merge-base upstream/master HEAD)"
+            MERGE_BASE="$(git merge-base upstream/test-downstream HEAD)"
             DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE":webui/react/src HEAD:webui/react/src")
             if [ "$DIFF_DIRS" = "shared" ] ; then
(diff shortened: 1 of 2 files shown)


webui/react/src/shared/bin/test.py
```diff
@@ -7,7 +7,7 @@
 parser = argparse.ArgumentParser()
 
 SHARED_WEB_REPO='https://github.com/determined-ai/shared-web'
-SAAS_REPO='https://github.com/determined-ai/saas'
+SAAS_REPO = "ssh://git@github.com/determined-ai/saas"
```
trentwatt, line 10: don't know if this change was actually needed. will check.

hamidzr: probably needed as it's a private repo. could we keep the spacing and quotations the same?

.circleci/config.yml
```diff
@@ -1268,6 +1268,15 @@ jobs:
       - store_artifacts:
           path: webui/react/coverage/lcov-report
 
+  test-unit-downstream:
```
hamidzr, line 1271: hmm, would it fit better if it's categorized as an integration test? 

wdy about limiting this to run only on prs opened from upstream. if that's possible

trentwatt: > hmm, would it fit better if it's categorized as an integration test?


yeah good call


> wdy about limiting this to run only on prs opened from upstream. if that's possible
i would rather have it run on forks as well, that way it "fails safe".  

if it automatically fails when it is not able to run, then it is easier to make sure that it is run, and we don't need any additional mechanisms (codeowners, checklist, PSAs)

webui/react/src/shared/bin/test.py
trentwatt: leave `ssh` but change quotation marks?

or change to `https` and quotation marks and see if it works?

.circleci/config.yml
```diff
@@ -1268,6 +1268,15 @@ jobs:
...
+    docker:
+      - image: cimg/node:16.14
+    steps:
+      - checkout-with-sm
+      - setup-python-venv:
+          executor: <<pipeline.parameters.machine-image>>
+      - run: cd webui/react/src/shared && ./bin/test.py --test-local --repos saas
```
hamidzr, line 1278: use test-shared target?

.circleci/config.yml
hamidzr: wouldn't that mean that there'd be a failing test for all prs originating from forks?

trentwatt: can make it so where it skips for things that don't touch shared.

hamidzr: perfect

trentwatt: ok it's up, will make some test PRs then put them in the description

webui/react/src/shared/bin/test.py
hamidzr: if as is, it also works locally for people as well then either one sounds good.

.circleci/config.yml
```diff
@@ -125,6 +125,20 @@ commands:
               echo "detected non-webui-only changes: $DIFF_DIRS"
             fi
 
+  skip-if-not-shared:
+    steps:
+      - run: git fetch upstream
+      - run:
+          name: check for shared changes (must be run after skip-if-docs-only)
```
hamidzr, line 132: nit: is this "must be run after.." needed ? 
maybe could be a comment 🤔 the name shows up in the ui

hamidzr (approved):
the changes lgtm/ thanks for setting this up. I asked for a review from @determined-ai/infrastructure just in case they notice something

.circleci/config.yml
trentwatt: hmm unfortunately it would be hard to test the skip because of how we get what's changed

`MERGE_BASE="$(git merge-base upstream/master HEAD)"`

emphasis on `upstream/master`

trentwatt: i will hard code it to `upstream/test-downstream` for testing purposes and switch back before merge.

.circleci/config.yml
trentwatt: ok i moved to comment.

the situation was they both do `git fetch upstream` and if you try to add an existing remote it errors. 

so just made it where you add the remote in one and then reuse it in the other. feels gross but w/e

trentwatt: could clean it up a bit, but maybe next time around lol.

loksonarius (approved):

.circleci/config.yml
trentwatt: it turns out putting it into `test-intg` made it no longer have access to remote.

@hamidzr do you think it's okay to do 
```
git remote add upstream https://github.com/determined-ai/determined 2>/dev/null
git fetch upstream
```

that way adding error for existing remote is supressed (but fetch fail will still cause error)

and we can get rid of dependency between `skip-webui` and `skip-docs`, making things simpler

PR #4606, opened 2022-07-21
chore: remove getProjectExperiments and merge into getExperiments

mapmeld:
## Description

Separated out from the experiment.state work in #4420 , this removes the separate `getProjectExperiments` API endpoint and combines it with `getExperiments`.  If we're going to change the API it makes sense to package that change on its own and roll it out sooner rather than later.

## Test Plan

Project pages and tests should run normally

## Checklist


docs/release-notes/4606-get-experiments-consolidation.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Breaking Changes**
+
+-  API: the `GetProjectExperiments` endpoint (`/projects/:id/experiments`) is removed and replaced
```
gh-determined-ai, line 5: I believe we're using double backticks (``), not single (`) in our docs.

gh-determined-ai (approved):

docs/release-notes/4606-get-experiments-consolidation.txt
mapmeld: Pushed new version with ``

EmilyBonar (approved):
LGTM! It's probably outside the scope of this PR, but would it be possible for `getExperiments` to also return the experiment configs (i.e. returning `ExperimentBase` vs `ProjectExperiment` or `ExperimentItem`)? That would allow us to have HP Search as an option in the context menu on the Project Details page.

ioga (approved):

mapmeld:
@ioga - added **config** object to GetExperiments and moved it from a separate attribute to part of the experiment in GetExperiment; lemme know if this raises issues.

ioga (approved):
approved, but let's try to call next version 0.19.0

PR #4614, opened 2022-07-21
feat: custom searcher unit tests, add/change minor things [DET-7717]

nrajanee:
## Description

feat: custom unit tests, add/change a few other things [DET-7717]

## Test Plan

No testing required. 

[DET-7717]: https://determinedai.atlassian.net/browse/DET-7717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/master/pkg/searcher/custom_search.go
+++ b/master/pkg/searcher/custom_search.go
@@ -80,6 +80,7 @@ func (s *customSearch) validationCompleted(
 	event := experimentv1.SearcherEvent_ValidationCompleted{
 		ValidationCompleted: &experimentv1.ValidationCompleted{
 			RequestId: requestID.String(),
+			Op:        op.ToProto(),
 			Metric:    metric,
 		},
 	}
@@ -94,9 +95,11 @@ func (s *customSearch) validationCompleted(
 func (s *customSearch) trialExitedEarly(
 	ctx context, requestID model.RequestID, exitedReason model.ExitedReason,
 ) ([]Operation, error) {
+
 	event := experimentv1.SearcherEvent_TrialExitedEarly{
 		TrialExitedEarly: &experimentv1.TrialExitedEarly{
-			RequestId: requestID.String(),
+			RequestId:    requestID.String(),
+			ExitedReason: model.ExitedReasonToProto(exitedReason),
 		},
 	}
 	searcherEvent := experimentv1.SearcherEvent{
--- a/master/pkg/searcher/custom_search.go
+++ b/master/pkg/searcher/custom_search.go
@@ -75,7 +75,7 @@ func (s *customSearch) progress(
 }
 
 func (s *customSearch) validationCompleted(
-	ctx context, requestID model.RequestID, metric float64,
+	ctx context, requestID model.RequestID, metric float64, op ValidateAfter,
 ) ([]Operation, error) {
 	event := experimentv1.SearcherEvent_ValidationCompleted{
 		ValidationCompleted: &experimentv1.ValidationCompleted{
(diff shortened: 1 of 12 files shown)


nrajanee:
I haven't added changes to use another interface for this method: getSearcherEventQueue() *SearcherEventQueue. Because first I wanted to discuss another way of doing it with @dzhu. It should be small change so I thought I'd create the PR anyway.

master/pkg/searcher/custom_search_test.go
```diff
@@ -0,0 +1,76 @@
...
+			Op:        validateAfterOp.ToProto(),
+		},
+	}
+	expEventCount += 1
+	searcherEvent3 := experimentv1.SearcherEvent{
+		Event: &validationCompletedEvent,
+		Id:    expEventCount,
+	}
+	expEvents = append(expEvents, &searcherEvent3)
+	require.Equal(t, expEvents, queue.GetEvents())
```
mpkouznetsov, line 75: while you are at it do you want to include `RemoveUpTo` and confirm that the queue is as expected after that?

nrajanee: oops, yes I had intended to add that initially. Forgot about it. Thanks Maksim!

proto/src/determined/experiment/v1/searcher.proto
```diff
@@ -26,6 +26,8 @@ message ValidationCompleted {
   string request_id = 1;
   // value of the validation metric used to direct the search
   double metric = 2;
+  // Need the length from ValidateAfterOperation.
```
mpkouznetsov, line 29: I was thinking about how this comment could be improved, and that prompted me to ask @liamcli the question in Slack about whether we would rather just include the length in the message and not the whole operation. Including the operation feels like we are annoying the user with implementation details. Let's see what his answer is.

mpkouznetsov (approved):
lgtm

master/pkg/searcher/custom_search.go
```diff
@@ -75,12 +75,13 @@ func (s *customSearch) progress(
 }
 
 func (s *customSearch) validationCompleted(
-	ctx context, requestID model.RequestID, metric float64,
+	ctx context, requestID model.RequestID, metric float64, op ValidateAfter,
```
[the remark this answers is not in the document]
nrajanee: I thought this way it would be clear that the length is coming from validateAfter op in the validationCompleted event.

gh-determined-ai (approved):
Make sure to provide a release note for this if it hasn't already been done.

master/pkg/searcher/custom_search_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	customSearchMethod := NewSearchMethod(config)
+	rand := nprand.New(0)
+	ctx := context{rand: rand}
+
+	var queue *SearcherEventQueue
+	if csMethod, ok := customSearchMethod.(CustomSearchMethod); ok {
+		queue = csMethod.getSearcherEventQueue()
+	}
```
dzhu, line 31: You can just do `queue := customSearchMethod.(CustomSearchMethod).getSearcherEventQueue()`, since you know this will work (and if it somehow didn't, we would want it to fail immediately anyway).

master/pkg/searcher/custom_search_test.go
```diff
@@ -0,0 +1,91 @@
...
+	expEvents := make([]*experimentv1.SearcherEvent, 0)
+	require.Equal(t, 0, len(queue.events))
+	// Add initialOperations
+	_, err := customSearchMethod.initialOperations(ctx)
+	require.NoError(t, err)
+	var expEventCount int32
+	expEventCount = 0
```
dzhu, line 38: You can drop this, since it's a big thing in Go that variables are always zero-initialized.

dzhu (approved):

PR #4619, opened 2022-07-21
chore: migrate `react-router-dom` from v5 to v6

keita-determined:
## Description

[DET-7851]

## Test Plan
- Check all pages if router works as expected
- Check code if there's something I miss

## Commentary (optional)

I followed [this instruction](https://github.com/remix-run/react-router/discussions/8753) and [this](https://reactrouter.com/docs/en/v6/upgrading/v5)

## Checklist

[DET-7851]: https://determinedai.atlassian.net/browse/DET-7851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/components/NotesCard.tsx
```diff
@@ -135,7 +135,7 @@ const NotesCard: React.FC<Props> = (
           onClick={handleNotesClick}
         />
       </Spinner>
-      <Prompt
+      {/* <Prompt
```
keita-determined, line 138: https://github.com/remix-run/react-router/issues/8139
`Prompt` is not available in v6

webui/react/src/pages/InteractiveTask.tsx
```diff
@@ -85,25 +84,25 @@ export const InteractiveTask: React.FC = () => {
         <div className={css.barContainer}>
           <TaskBar
             handleViewLogsClick={() => setPageView(PageView.TASK_LOGS)}
-            id={taskId}
-            name={taskName}
-            resourcePool={taskResourcePool}
-            type={taskType}
+            id={taskId!}
```
keita-determined, line 87: im thinking to `assert` these cuz using `??` is not appropriate. Any other better options?

webui/react/src/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { Router } from 'react-router-dom';
+import { unstable_HistoryRouter as HistoryRouter } from 'react-router-dom';
```
keita-determined, line 3: This code should be safe because I set `history` lib to be the same version as `react-router-dom`
https://reactrouter.com/docs/en/v6/routers/history-router
https://github.com/remix-run/react-router/blob/f16c5490dfa75f15dcfb86d2a981a7c58a9d1a33/package.json#L44

webui/react/package.json
```diff
@@ -28,7 +28,7 @@
     "fp-ts": "^2.11.2",
     "fuse.js": "^6.4.6",
     "hermes-parallel-coordinates": "^0.5.18",
-    "history": "^4.10.1",
+    "history": "^5.2.0",
```
keita-determined, line 31: https://github.com/remix-run/history/releases/tag/v5.0.0
there is breaking changes but In my investigation, theres no affects.

trentwatt:
From the instructions you linked:
>Instead of upgrading and updating all of your code at once (which is incredibly difficult and prone to bugs), this strategy enables you to upgrade one component, one hook, and one route at a time by running both v5 and v6 in parallel. Any code you haven't touched is still running the very same code it was before. Once all components are exclusively using the v6 APIs, your app no longer needs the compatibility package and is running on v6.

Where are we in this process? Does this PR just lay the groundwork for us to be able to upgrade components one by one? Or does it update some (or all??) of the components?

trentwatt:
@hkang1 

This one (or at least the migration PRs going forward) probably needs a pretty comprehensive test plan. Can we put together some summary spec of the existing (intended) behavior for navigation?

e.g. from project details, click on experiment. then click a tab. existing behavior in this case is that you go back to project details. is this desired?

also: `useSettings` has `shouldPush` param, which to my understanding means that the url-encoded setting gets it own entry in the navigation history-- is that right? different test cases involving url-encoding of settings without `shouldPush` probably also need to be incorporated as well.

there have also occasionally been issues around navigating e.g, `/det/resourcepool/compute-pool` and /det/resourcepool/aux-pools. 

also redirecting from tensorboard, opening jupyterlab in new window. navigating back from embedded view.

anything else come to mind?

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,6 +1,7 @@
 import queryString from 'query-string';
 import { useCallback, useEffect, useMemo, useState } from 'react';
-import { useHistory, useLocation } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
```
hkang1, line 4: **nit:** combine into one line

webui/react/src/index.tsx
hkang1: **suggestion:** can you add this comment above this line just so people reading this part of the code can lookup why it's called `unstable_HistoryRouter`
```
/**
 * It's considered unstable until `react-router-dom` can detect
 * history version mismatches when supplying your own history.
 * https://reactrouter.com/docs/en/v6/routers/history-router
 */
```

webui/react/src/pages/ExperimentDetails/ExperimentSingleTrialTabs.tsx
```diff
@@ -1,6 +1,7 @@
 import { Tabs } from 'antd';
 import React, { useCallback, useEffect, useState } from 'react';
-import { useHistory, useParams } from 'react-router';
+import { useParams } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
```
hkang1, line 4: **nit:** one line, maybe IDE needs adjustment to do this?

webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
```diff
@@ -1,6 +1,7 @@
 import { Alert, Tabs } from 'antd';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { useHistory, useLocation } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
```
hkang1, line 4: ditto

keita-determined:
> From the instructions you linked:
> 
> > Instead of upgrading and updating all of your code at once (which is incredibly difficult and prone to bugs), this strategy enables you to upgrade one component, one hook, and one route at a time by running both v5 and v6 in parallel. Any code you haven't touched is still running the very same code it was before. Once all components are exclusively using the v6 APIs, your app no longer needs the compatibility package and is running on v6.
> 
> Where are we in this process? Does this PR just lay the groundwork for us to be able to upgrade components one by one? Or does it update some (or all??) of the components?

Done with the entire process, so now on v6 completely. I tried to to update it little by little but our website relies on `history` library a lot instead of using `NavLink`. Although we don't need to have `history` library since `react-router-dom v6` have it as a dependency, we rely on `history` a lot, so I needed to handle differently which unables to co-exist both v5 and v6.
For example, [This](https://github.com/determined-ai/determined/pull/4619/files#diff-d211b226e853bf7f7d109f740456cb93344b177185a22a35c095610e4b066f72R2) syntax is changed. [New HistoryRouter](https://github.com/determined-ai/determined/pull/4619/files#diff-7ce3dfc51c01b4d5a4972c9d0770cfbe49a60d20e1c440cf755b13001843a53cR14) requires new version of `history`.
That's why I updated all from v5 and v6 at once. However, as @trentwatt said, we need to have comprehensive test cases to upgrade safely. I think I (or someone else) should write tests first, and then test this PR would be good.

webui/react/src/pages/InteractiveTask.tsx
hkang1: **question:** correct me if I'm wrong, but there's no guarantee that `taskId` is always defined right? you can still go to routes like `/wait/notebook/` (instead of `/wait/notebook/<some-id>`. Variables coming through URL params should probably be treated as conditional definition. So `??` seems appropriate, or not showing the respective component if it requires the prop to be defined...

webui/react/src/pages/InteractiveTask.tsx
```diff
@@ -9,7 +9,7 @@ import { CommandState, CommandType } from 'types';
 import css from './InteractiveTask.module.scss';
 import TaskLogs from './TaskLogs';
 
-interface Params {
+type Params = {
```
hkang1, line 12: **question:** related to the question written below, should `taskId`, `taskName` and `taskResourcePool` be written with the `?` suffix (e.g. `taskId?: string;`)?

much like `ProjectDetails.tsx Params`

webui/react/src/pages/TaskLogs.tsx
```diff
@@ -16,7 +16,7 @@ import { CommandType } from 'types';
 
 import css from './TaskLogs.module.scss';
 
-interface Params {
+type Params = {
```
hkang1, line 19: ditto: conditional params?

webui/react/src/pages/TrialDetails.tsx
```diff
@@ -1,6 +1,7 @@
 import { Tabs } from 'antd';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
-import { useHistory, useParams } from 'react-router';
+import { useParams } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
```
hkang1, line 4: ditto

webui/react/src/pages/Wait.tsx
```diff
@@ -17,7 +17,7 @@ import { WaitStatus } from 'wait';
 
 import css from './Wait.module.scss';
 
-interface Params {
+type Params = {
```
hkang1, line 20: ditto: conditional params? unless the router is preventing this page from rendering if the params are not provided?

webui/react/src/pages/WorkspaceList.tsx
```diff
@@ -248,6 +248,7 @@ const WorkspaceList: React.FC = () => {
               limit: settings.tableLimit,
               offset: settings.tableOffset,
             }, total)}
+            rowKey="id"
```
hkang1, line 251: 👍

webui/react/src/routes/routes.ts
```diff
@@ -113,21 +125,41 @@ const routes: RouteConfig[] = [
...
+    path: '/cluster/:tab',
     redirect: '/clusters',
     title: 'Cluster',
   },
   {
     icon: 'cluster',
     id: 'clusters',
     needAuth: true,
-    path: '/clusters/:tab?',
+    path: '/clusters',
```
hkang1, line 144: **question:** huh, is this something that changed with the version bump or was it always like this? I remember if we didn't put the more specific path before the general path in the array, it would resolve to the general before the more specific route, but playing with the preview, it doesn't seem to be the case?

webui/react/src/shared/routes/history.ts
```diff
@@ -1,2 +1,2 @@
 import { createBrowserHistory } from 'history';
-export default createBrowserHistory({ basename: process.env.PUBLIC_URL });
+export default createBrowserHistory();
```
hkang1, line 2: **question:** this is ok to do? same question with the changes in `stripUrl`

webui/react/src/pages/ExperimentDetails/ExperimentSingleTrialTabs.tsx
keita-determined: I forgot to combine them when I replace `react-router-dom-v5-compat` with `react-router-dom`. Nice catch!

webui/react/src/shared/routes/history.ts
keita-determined: The latest `history` library doesnt provide `basename` anymore, and we need to update `history` as well as `react-router-dom`

webui/react/src/components/NotesCard.tsx
```diff
@@ -135,18 +133,6 @@ const NotesCard: React.FC<Props> = (
...
-          const isSameExperiment = location.pathname.split('/')[0] === 'experiment' &&
-             newLocation.pathname.startsWith(location.pathname.split('/').slice(0, -1).join('/'));
-          return (
-            isSameExperiment ?
-              true :
-              'You have unsaved notes, are you sure you want to leave? Unsaved notes will be lost.'
-          );
-        }}
-        when={editedNotes !== notes}
-      />
```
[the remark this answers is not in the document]
keita-determined: I made a comment [here](https://github.com/determined-ai/determined/pull/4619#discussion_r927176853). 
It is already buggy before the update. I think we should find the alternative or wait for the new version which supports this.

webui/react/src/routes/routes.ts
keita-determined: It comes from [this](https://github.com/remix-run/react-router/discussions/8753)
image
`react-router-dom v6` doesnt support this syntax so I changed it

webui/react/src/pages/InteractiveTask.tsx
keita-determined: in `react-router-dom v6`, `useParam` returns always `string | undefined` even when we don't put `?`. 
In v5, `useParam` returns only string if we dont put `?`, so it used be like [this](https://github.com/determined-ai/determined/blob/2da2ed2a1f11a87bd5999200561671149bd10cc9/webui/react/src/pages/InteractiveTask.tsx#L12). I just followed the previous style which doesnt have `?`. 
Also, I replaced `interface` with `type` due to type error in a new version.

webui/react/src/pages/InteractiveTask.tsx
keita-determined: It used be like [this](https://github.com/determined-ai/determined/blob/2da2ed2a1f11a87bd5999200561671149bd10cc9/webui/react/src/pages/InteractiveTask.tsx#L12). I just followed the previous style which doesn't have `?`.

webui/react/src/routes/routes.ts
hkang1: it's a bit more verbose but it's cleaner IMO

webui/react/src/shared/utils/routes.ts
```diff
@@ -48,9 +48,6 @@ export const parseUrl = (url: string): URL => {
 const stripUrl = (aUrl: string): string => {
   const url = parseUrl(aUrl);
   const rest = url.href.replace(url.origin, '');
-  if (rest.startsWith(process.env.PUBLIC_URL)) {
```
hamidzr, line 51: why were we using this 🤔

keita-determined: not 100% sure, but its for netlify and production mode. In local, there's `/det` but on netlify there isn't.
@trentwatt might know about it more.

hamidzr: most likely I wrote this.. do we need to change or remove it?

keita-determined: the `history.ts` is changed, so the basename is not included anymore. That's why I removed `replace()` part.

webui/react/src/shared/routes/history.ts
hamidzr: some use cases that rely on PUBLIC_URL are:
- serve ui from any base path: eg `/det` `/` and `/cat`

webui/react/src/shared/utils/routes.ts
hamidzr: thanks. I think if [these](https://github.com/determined-ai/determined/pull/4619/files#r929350828) use cases still work we should be good here

webui/react/src/shared/routes/history.ts
keita-determined: I moved base name to [here](https://github.com/determined-ai/determined/pull/4619/files#diff-fe9fa55e9afd85c5adb046f05a63357a7ed70bca08fae9622936c71c5a6f39b6R23). I tested most of pages manually, but we should write tests before this PR is merged.

webui/react/src/components/Router.tsx
```diff
@@ -32,25 +33,23 @@ const Router: React.FC<Props> = (props: Props) => {
...
-                    state: { loginRedirect: filterOutLoginLocation(location) },
-                  }}
+              element={(
+                <Navigate
+                  state={filterOutLoginLocation(location)}
+                  to={paths.login()}
                 />
               )}
+              key={route.id}
+              {...route}
```
trentwatt, line 52: move to top?

webui/react/src/pages/Wait.tsx
keita-determined: i use `assertIsDefined` to check if `undefined` or not. Even if the variable is undefined, the page will redirect to default page, and `assertIsDefined` doesnt actually throw error

keita-determined:
@hkang1 @hamidzr @trentwatt 

## New Changes Since the Last Review
- Found the alternative of `Prompt`
- Fix minor bugs
  - Order of props in `Router.tsx`
- Added `assertIsDefined` to check if params are defined (it is mostly for typescript type checking)
  - Even though the code has `throw Error`, it actually doesnt run since if the params are missing, page is redirected to the `paroject/1` or if the params do not exist, the page says `id xxx doesnt exist` or something similar 

## Concern
- In the current behavior with old `react-router-dom` and `history`, it automatically redirect to the default path if the path doesn't exist, such as `http://localhost:3000/` -> `http://localhost:3000/det`. However, due to the change of both `react-router-dom` and `history` (`history` doesnt take `basename` param anymore), we have to manually add `/det` if we run it locally.

trentwatt (commented):
I think redirecting `serverAddress` -> `serverAddress/det` is a behavior that needs to be preserved with any PR, unless we can confirm otherwise.

as of now, if i go to `localhost:3000` I get a blank page. I could see a situation where customers have an autocompleted  address that they go to, and if that ends up being a blank page, it would not be a great thing.

also i would prefer that we use typescript type assertions instead of writing our own `assertIsDefined` that throws and error (and crashes the page on my local)

Also couple bugs as discussed:
- clicking on project card results in blank page
- forking experiment results in blank page
- creating new project results in blank page
- creating new workspace results in blank page

probably some other things? can you verify all of the different pages and functionalities?

keita-determined:
> ## Concern
> * In the current behavior with old `react-router-dom` and `history`, it automatically redirect to the default path if the path doesn't exist, such as `http://localhost:3000/` -> `http://localhost:3000/det`. However, due to the change of both `react-router-dom` and `history` (`history` doesnt take `basename` param anymore), we have to manually add `/det` if we run it locally.

This concern is fixed!!

trentwatt (commented):
Couple more regressions: 

https://user-images.githubusercontent.com/28383080/184429853-fd693eae-5ecc-4f51-b568-a7048d536111.mov



- changing the password results in blank screen
- another time after changing password and going to `localhost:3000` I kept getting redirected to blank screen
- another time the page crashed immediately after login.

At this point I would prefer that we actually follow the advice from the react router team if at all possible:
>Instead of upgrading and updating all of your code at once (which is incredibly difficult and prone to bugs), this strategy enables you to upgrade one component, one hook, and one route at a time by running both v5 and v6 in parallel. Any code you haven't touched is still running the very same code it was before. Once all components are exclusively using the v6 APIs, your app no longer needs the compatibility package and is running on v6.


It was unclear to me from your [comment above](https://github.com/determined-ai/determined/pull/4619#issuecomment-1194191120) whether that was impossible, or just somehow difficult.

Side question: what is the motivation for the upgrade in the first place? is v5 nearing end of support?

keita-determined:
I'll fix the issue around password. Thanks for finding

> It was unclear to me from your https://github.com/determined-ai/determined/pull/4619#issuecomment-1194191120 whether that was impossible, or just somehow difficult.

I followed the instruction step by step, but at some point, project got broken. We heavily use `history`, so I assume it might be the reason why it is difficult to upgrade step by step. I won't say it's impossible but IMO it is hard to gradually upgrade.

> Side question: what is the motivation for the upgrade in the first place? is v5 nearing end of support?

The motivation is that lots of libraries around React is not updated for several months. The longer we wait to update libraries, the more difficult it will be in the future. Especially, React 18 and React-Router-Dom v6 have some breaking changes.

keita-determined:
I close this PR to start over.

keita-determined:
I close this PR to start over.

PR #4620, opened 2022-07-22
chore: RBAC user groups crud

maxrussell:
## Description

For [DET-7854], implementation of user groups CRUD API.

## Test Plan

The plan is to end-to-end test this with the CLI when that's implemented, but we tested this by starting up `devcluster` and going to e.g. http://localhost:8080/docs/rest-api/#/Internal/CreateGroup, clicking "Try it out," and sending HTTP requests to the API.

[DET-7854]: https://determinedai.atlassian.net/browse/DET-7854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+		err = mapAndFilterErrors(err)
+	}()
+
+	groups, err := SearchGroups(ctx, req.Name, model.UserID(req.UserId), int(req.Offset),
+		int(req.Limit))
+	if err != nil {
+		return nil, err
+	}
+
+	return &apiv1.GroupSearchResponse{
```
NicholasBlaskey, line 67: ```GroupSearchResponse.pagination``` is always nil here

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+		BaseModel: bun.BaseModel{},
+		ID:        int(req.GroupId),
+		Name:      newName,
+		OwnerID:   oldGroup.OwnerID,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(req.AddUsers) > 0 {
```
NicholasBlaskey, line 127: Also not atomic here

proto/src/determined/api/v1/group.proto
```diff
@@ -0,0 +1,67 @@
...
+
+message GetGroupRequest {
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+        json_schema: { required: ["group_id"] }
+    };
+    int32 group_id = 1;
+}
+
+message GetGroupResponse {
+    determined.group.v1.GroupDetails group = 1;
```
NicholasBlaskey, line 19: required here

proto/src/determined/api/v1/group.proto
```diff
@@ -0,0 +1,67 @@
...
+    int32 group_id = 1;
+    string name = 2;
+    // The user ids of users to add to the group
+    repeated int32 add_users = 3;
+    // The user ids of users to delete from the group
+    repeated int32 remove_users = 4;
+} // returns GroupWriteResponse
+
+message GroupWriteResponse {
+    determined.group.v1.GroupDetails group = 1;
```
NicholasBlaskey, line 54: required here

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (a *ApiServer) DeleteGroup(ctx context.Context, req *apiv1.DeleteGroupRequest,
+) (resp *apiv1.DeleteGroupResponse, err error) {
+	// Detect whether we're returning special errors and convert to gRPC error
+	defer func() {
+		err = mapAndFilterErrors(err)
+	}()
+
+	err = DeleteGroup(ctx, int(req.GroupId))
+	if err != nil {
```
NicholasBlaskey, line 173: nit: could do this ```if err := DeleteGroup(ctx, int(req.GroupId)); err != nil``` syntax and in some other places too

NicholasBlaskey (commented):
Nice work!

A general comment is some of the error messages are somewhat unfriendly currently like trying to create a group with a name that already exists you get

```
{"error":{"code":6,"reason":"AlreadyExists","error":"Duplicate record"}}
```

or worse update group gives the same error
```
{"error":{"code":5,"reason":"NotFound","error":"Not found"}}
```
if the group is not found or trying to add a user who doesn't exist

master/internal/usergroup/postgres_groups.go
```diff
@@ -14,11 +15,53 @@ import (
 
 // AddGroup adds a group to the database. Returns ErrDuplicateRow if a
 // group already exists with the same name or ID.
-func AddGroup(ctx context.Context, group Group) (Group, error) {
-	_, err := db.Bun().NewInsert().Model(&group).Exec(ctx)
+func AddGroup(ctx context.Context, idb bun.IDB, group Group) (Group, error) {
```
ioga, line 18: how about we introduce a rule that db functions that take (context, idb, ...) have `Tx` suffix (i.e. `AddGroupTx`, `AddUsersToGroupTx`, etc.)
```suggestion
func AddGroupTx(ctx context.Context, idb bun.IDB, group Group) (Group, error) {
```

it was item 3 in an early proposal here: https://github.com/determined-ai/ci-shortcuts/pull/2#issuecomment-1070221693

ioga (approved):

master/internal/db/postgres.go
```diff
@@ -52,6 +52,27 @@ func Bun() *bun.DB {
 	return theOneBun
 }
 
+// PaginateBun adds sorting and pagination to the provided bun query, defaulting to certain values
+// if they are not specified. By default, we order by ascending on the id column, with no limit.
+func PaginateBun(query *bun.SelectQuery, orderColumn, direction string, offset, limit int) *bun.SelectQuery {
```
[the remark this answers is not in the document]
maxrussell: Done!

master/internal/usergroup/api_groups.go
maxrussell: I'm aware that's a commonly used syntax in Go, but I prefer not to use it because I don't like how much it buries the main action being done (e.g. `DeleteGroup`). I'm willing to change, though, if others have a strong preference here.

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+		Name: req.Name,
+	}
+
+	createdGroup, err := AddGroup(ctx, group)
+	if err != nil {
+		return nil, err
+	}
+
+	uids := intsToUserIDs(req.AddUsers)
+	err = AddUsersToGroup(ctx, createdGroup.ID, uids...)
```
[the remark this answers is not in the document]
maxrussell: Good catch on the atomicity. Fixed!

I'm not quite understanding what the particular problem is with the error messages. Does the CLI merely print the exact error message returned by the API or something?

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	switch err {
+	case db.ErrNotFound:
+		return errNotFound
+	case db.ErrDuplicateRecord:
+		return errDuplicateRecord
+	}
+
+	return errInternal
```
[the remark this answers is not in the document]
maxrussell: Yeah, not exposing too much was explicitly the point. Maybe we should log it for debugging locally.

master/internal/usergroup/postgres_groups.go
maxrussell: No objection here.

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (a *ApiServer) GetGroups(ctx context.Context, req *apiv1.GroupSearchRequest,
+) (resp *apiv1.GroupSearchResponse, err error) {
+	// Detect whether we're returning special errors and convert to gRPC error
+	defer func() {
+		err = mapAndFilterErrors(err)
+	}()
+
+	groups, err := SearchGroups(ctx, req.Name, model.UserID(req.UserId), int(req.Offset),
```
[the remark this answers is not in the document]
maxrussell: This is a really interesting point. I'm most accustomed to designing APIs for SaaS sorts of products, and I think in that context pagination should always be mandatory. But we're open source software, so I figured users would be running their own clusters and such and wouldn't deliberately slow down the server by sending requests designed to artificially generate load. But your comment made me realize that we *are* going to offer this in a SaaS context. So yeah, totally agree that page size should be limited. What do we think as a max? 500?

master/internal/usergroup/api_groups.go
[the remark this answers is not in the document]
NicholasBlaskey: The CLI for the most part prints the exact message returned by the API. 

However regardless the API is exposed to end users and personally if I am consuming an API would much rather see something like ```could not create group: failed to add users with ID 9998,9999 not found```. 

If I had never used Determined the error message currently returned would make me wonder if I even have the right endpoint

master/internal/usergroup/api_groups.go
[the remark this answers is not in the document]
NicholasBlaskey: A lot of other places in the API code we return the error unfiltered (even before user group routes I think the token lookup or something will throw an error unfiltered if the database is down).

So for consistency sake, I think we should decide with the team and communicate one preferred approach so we don't eventually end up with some routes giving information that others don't.

master/internal/usergroup/api_groups.go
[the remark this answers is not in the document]
eecsliu: @NicholasBlaskey Great points and we changed some of the error handling to allow us to make more descriptive error messages. There is a technical limitation on returning more descriptive information on bulk operations because we won't know for sure which specific item failed without adding some more expensive code, so we added a more generic message for those operations (batch add/remove users). Hopefully we get good mileage out of what we have now though.

loksonarius:
Uh... I think some rebase went a little weird. Is the commit history intended?

master/internal/usergroup/api_groups.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return ids
+}
+
+const (
+	maxLimit = 500
+)
+
+var (
+	errBadRequest   = status.Error(codes.InvalidArgument, "Bad request")
```
NicholasBlaskey, line 181: nit: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

NicholasBlaskey (approved):
Great work, LGTM

master/internal/usergroup/postgres_groups.go
```diff
@@ -6,71 +6,176 @@ import (
...
+// addGroup adds a group to the database. Returns ErrDuplicateRow if a
+// group already exists with the same name or ID. Will use db.Bun() if
+// passed nil for idb.
+func addGroup(ctx context.Context, idb bun.IDB, group Group) (Group, error) {
+	if idb == nil {
+		idb = db.Bun()
+	}
+
+	_, err := idb.NewInsert().Model(&group).Exec(ctx)
+	return group, errors.Wrapf(matchSentinelError(err), "Error creating group %d", group.ID)
```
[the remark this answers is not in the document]
maxrussell: Great point!

master/internal/usergroup/api_groups.go
maxrussell: Hm, yeah, that's interesting. Since this is what's sent to HTTP clients, the goal was to mimic the standard library in places like [this](https://cs.opensource.google/go/go/+/refs/tags/go1.18.5:src/net/http/server.go;l=1931), but it seems like I picked an awkward not-quite-consistent middle ground. I'll go with lowercase.

proto/src/determined/api/v1/group.proto
```diff
@@ -0,0 +1,117 @@
...
+    json_schema: { required: [] }
+  };
+  // The id of the user to use to find groups to which the user belongs.
+  int32 user_id = 1;
+  // The group name to use when searching.
+  string name = 2;
+  // Skip the number of groups before returning results. Negative values
+  // denote number of groups to skip from the end before returning results.
+  int32 offset = 3;
+  // Limit the number of groups. A value of 0 denotes no limit.
```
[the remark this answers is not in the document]
maxrussell: Ah, yeah, I'll update to reflect its new required nature.

Here 0 is interpreted as "no limit," which is by definition greater than our maximum of 500. Totally agree that we should make this consistent across the API. I don't have a *super* strong opinion, but mine is that limit should be required and be explicit with no default.

PR #4622, opened 2022-07-22
fix: gpt_neox deepspeed example

liamcli:
## Description
fix a minor issue with the example:
- parsing args failed if `overwrite_values` is not provided in the determined experiment config -> now we will return an empty dict

Note that we do not have a dataloader resume issue since we are relying on determined's dataloader to skip the correct number of datapoints when resuming.  If we were using gpt-neox's dataloader directly, we'd need to make sure to set start_iter after loading a checkpoint.  

## Test Plan
Tested manually

## Checklist

### DIFF
--- a/examples/deepspeed/gpt_neox/det_utils.py
+++ b/examples/deepspeed/gpt_neox/det_utils.py
@@ -17,7 +17,7 @@ def get_neox_args(context):
     exp_config = context.get_experiment_config()
 
     # Gather overrides.
-    overwrite_values = args.pop("overwrite_values")
+    overwrite_values = args.pop("overwrite_values", {})
     # We are going to overwrite certain neox_args with determined config values
     # from the experiment config to ensure consistency.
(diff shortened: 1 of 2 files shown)


coreystatendet:
Question that might be obvious: where is `start_iter` used?  I see it's used on the first iteration of NeoX's `DistributedBatchSampler`, but if I'm interpreting correctly we're using Determined's `DistributedBatchSampler` here which doesn't reference it.

liamcli:
> Question that might be obvious: where is `start_iter` used? I see it's used on the first iteration of NeoX's `DistributedBatchSampler`, but if I'm interpreting correctly we're using Determined's `DistributedBatchSampler` here which doesn't reference it.

good point, I did think about this when i first wrote it and the dataloader resume was handled correctly by determined's dataloader.  we don't need this change here.

coreystatendet (approved):

PR #4625, opened 2022-07-22
docs: complete switch from `txt` to `rst` extensions

dzhu:
## Description

Most files were renamed in the recent Great Restructure, but some things were missed. Part of the effect was that `make fmt` no longer did anything, so a bunch of files are being reformatted now.

## Test Plan

- [x] build docs


gh-determined-ai (approved):
Good catch!

PR #4626, opened 2022-07-22
feat: pluggable authorization for RBAC.

ioga:
## Description

We'd like to enable pluggable authorization so:
1. it can be added and enabled in EE in an easier, additive-only fashion to
   minimize EE issues.
2. it's modular: no single interface, authorization can instead be implemented
   for modules (i.e. `internal/*` subdirectories).

See `AuthZProviderType` in `internal/authz/authz_provider_type.go` for the core implementation.

As an example, the "can user change someone's password" APIs were modified to use it:
- `internal/user/authz_iface.go` contains an interface that all AuthZ types will need to implement and "the registry" (`AuthZProvider`) 
- `internal/user/authz_basic_impl.go` implements this interface for our existing authorization (users can change their own password, admins can change anyones)
- `internal/user/authz_unrestricted_impl.go` is a toy implementation which allows anyone to do anything. this only serves as an example and should be removed before we merge to master.
- see changes in `internal/api_user.go` and `internal/user/service.go` for interface usage examples.

When implementing authorization for new module (e.g. usergroups), one would need to:
- add `internal/usergroups/authz_iface.go` with an interface (`UserGroupAuthZ`) describing all permission `Can*` or `Filter*` functions, and `var AuthZProvider authz.AuthZProviderType[UserGroupAuthZ]`
- use the interface's methods in API handlers
- add `internal/usergroups/authz_basic_impl.go` with interface implementation for basic OSS auth
- [IN EE ONLY] add `internal/usergroups/authz_rbac_impl.go` with RBAC authz.

In configuration, `security.authz` contains required `Type` field and optional `Fallback`:
- `Type` is the selected authz implementation
- `FallbackType` is an optional backup to be used when the selected implementation is not available *for some modules*, allowing us to partially implement RBAC authz on per-module basis, falling back to basic in the interim. It is likely this can also be removed at merge.

## Test Plan

Given admin user A, and normal users U1 and U2:
- with default config (or security -> authz -> type set to "basic")
    - A can change A, U1, and U2 passwords
    - U1 can change their own password
    - U1 cannot change A or U2 password
- with security -> authz -> type set to "unrestricted":
    - anyone can change anyone's password. yolo.

## Commentary (optional)

Some people call this "dependency injection" or "inversion of control".

## Checklist

### DIFF
--- a/master/internal/authz/authz_provider_type.go
+++ b/master/internal/authz/authz_provider_type.go
@@ -1,0 +1,63 @@
+package authz
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"sync"
+
+	"github.com/determined-ai/determined/master/internal/config"
+	"golang.org/x/exp/maps"
+)
+
+type AuthZProviderType[T any] struct {
+	registry map[string]T
+	once     sync.Once
+}
+
+func (p *AuthZProviderType[T]) Register(authZType string, impl T) {
+	p.once.Do(func() {
+		p.registry = make(map[string]T)
+	})
+	// TODO(ilia): keep registry here or global in internal/authz/authz_basic.go
+	config.RegisterAuthZType(authZType)
+	if _, ok := p.registry[authZType]; ok {
+		panic(fmt.Errorf("can't do double register of type %s for: %s", authZType, p.string()))
+	}
+	p.registry[authZType] = impl
+}
+
+func (p *AuthZProviderType[T]) string() string {
+	return reflect.TypeOf((*T)(nil)).Elem().String()
+}
+
+func (p *AuthZProviderType[T]) Get() T {
+	if len(p.registry) == 0 {
+		panic(fmt.Errorf("empty registry for: %s", p.string()))
+	}
+
+	authZConfig := config.GetAuthZConfig()
+
+	res, ok := p.registry[authZConfig.Type]
+	if ok {
+		return res
+	}
+
+	okTypes := strings.Join(maps.Keys(p.registry), ", ")
+
+	if authZConfig.FallbackType == nil {
+		panic(fmt.Errorf(
+			"failed to find authz type %s in %s for: %s",
+			authZConfig.Type, okTypes, p.string()))
+	}
+
+	res, ok = p.registry[*authZConfig.FallbackType]
+	if ok {
+		return res
+	}
+
+	panic(fmt.Errorf(
+		"failed to find authz types %s, %s in %s for: %s",
+		authZConfig.Type, *authZConfig.FallbackType, okTypes, p.string()))
+
+}
--- a/master/internal/authz/authz_provider_type.go
+++ b/master/internal/authz/authz_provider_type.go
@@ -6,15 +6,18 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/determined-ai/determined/master/internal/config"
 	"golang.org/x/exp/maps"
+
+	"github.com/determined-ai/determined/master/internal/config"
 )
 
+// AuthZProviderType is a per-module registry for authz implementations.
 type AuthZProviderType[T any] struct {
 	registry map[string]T
 	once     sync.Once
 }
 
+// Register adds new implementation.
 func (p *AuthZProviderType[T]) Register(authZType string, impl T) {
 	p.once.Do(func() {
 		p.registry = make(map[string]T)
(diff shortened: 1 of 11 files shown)


master/internal/authz/authz_provider_type.go
```diff
@@ -0,0 +1,63 @@
...
+	// TODO(ilia): keep registry here or global in internal/authz/authz_basic.go
+	config.RegisterAuthZType(authZType)
+	if _, ok := p.registry[authZType]; ok {
+		panic(fmt.Errorf("can't do double register of type %s for: %s", authZType, p.string()))
+	}
+	p.registry[authZType] = impl
+}
+
+func (p *AuthZProviderType[T]) string() string {
+	return reflect.TypeOf((*T)(nil)).Elem().String()
```
NicholasBlaskey, line 31: nit: Does ```fmt.Sprintf("%T", p.registry["anykey"])``` work here?

I mean they are both pretty gross tbh

master/internal/authz/authz_provider_type.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (p *AuthZProviderType[T]) Get() T {
```
NicholasBlaskey, line 34: Non blocking since this is something we will need to see how plays it out

But in the case of someone testing out RBAC they are going to run it permissive but still wanting original OSS controls (especially for things like normal users being unable to change someones else's password)

This might require us to define a new interface that basically just calls the other two interfaces which might actually be the best solution here.

master/internal/user/authz_unrestricted_impl.go
```diff
@@ -0,0 +1,17 @@
+package user
+
+import (
+	"github.com/determined-ai/determined/master/pkg/model"
+)
+
+// TODO XXX: this is for demo purposes only. Remove before merging to master branch.
+type UserAuthZUnrestricted struct {
```
NicholasBlaskey, line 8: docker registry's call this silly auth lol

NicholasBlaskey (approved):
Looks really good and super flexible to handle any weird cases that we need to auth for

master/internal/authz/authz_provider_type.go
ioga: for non-existing key it prints `<nil>`. we could `fmt.Sprintf("%T", p.registry)` but that'd be prefixed with `map[string]`...

master/internal/user/authz_basic_impl.go
```diff
@@ -0,0 +1,22 @@
...
+// CanSetUserPassword for basic authz.
+func (a *UserAuthZBasic) CanSetUserPassword(currentUser model.User, targetUser model.User) error {
+	if !currentUser.Admin && currentUser.ID != targetUser.ID {
+		return fmt.Errorf("non-admin users can only change their own password")
+	}
+	return nil
+}
+
+func init() {
+	AuthZProvider.Register("basic", &UserAuthZBasic{})
```
stoksc, line 21: how does this init pattern work if you need the config in initialization?

stoksc: and that we end up with a registry when there should only be one valid `UserAuthZ` is interesting.

ioga: do you mean you want to initialize something inside `UserAuthZBasic`?
I think you could add `sync.Once` to `UserAuthZBasic` and then call a helper that uses it to initialize once lazily on first use (in all the implemented methods).

stoksc: I'm just reconciling this with my RM implementation, where there is a `Setup` method that must be called and it initializes a `type ResourceManager interface` singleton in the package. These things are different, but it'd be nice to me if they all followed a similar-ish pattern.

ioga: >and that we end up with a registry when there should only be one valid UserAuthZ is interesting.

one could implement UserAuthZ which instantiates and calls two other UserAuthZ implementations, e.g. when we want "permissive rbac with audit, enforce basic" strategy.

stoksc: > I think you could add sync.Once to UserAuthZBasic and then call a helper that uses it to initialize once lazily on first use (in all the implemented methods).

yeah, that just seems less than ideal, but I suppose it would work.

stoksc: > one could implement UserAuthZ which instantiates and calls two other UserAuthZ implementations, e.g. when we want "permissive rbac with audit, enforce basic" strategy.

ah yeah, I would['ve just thought to register a new type in this case but I see.

stoksc: once initialization can fail, failing in initialization on the first call kinda sucks, though

PR #4631, opened 2022-07-23
feat: log TaskInfo line with various IDs [DET-7832]

varlogtim:
Adds various IDs to all task logs.

Useful for knowing exactly where a task container is running and with what resources.

Test plan. Run a trial and run some STNCs, check the logs for `TaskInfo:`

### DIFF
--- a/agent/internal/container.go
+++ b/agent/internal/container.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


agent/internal/container.go
```diff
@@ -146,6 +148,14 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 		// TODO(ilia): When do we need to start a checker for these containers?
 
 	case containerReady:
+		taskInfo := c.makeTaskInfo()
+		ctx.Tell(ctx.Self().Parent(), aproto.ContainerLog{
```
stoksc, line 152: can't really do it in the agent, or it doesn't work for k8s/slurm

stoksc: could probably do it in the container itself or in the allocation actor

varlogtim:
I am not sure if this was done another way. Closing because it is stale.

PR #4637, opened 2022-07-25
feat: user group CLI and RBAC feature flag [DET-7889,DET-8210]

NicholasBlaskey:
## Description

CLI implementation of user groups. 

## Test Plan

e2e tests

Along with following ```det user-group -h``` and manually verifying the commands work as expected and that the help makes sense.

## Commentary (optional)

Needs #4620 to be merged.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_groups.py
+++ b/e2e_tests/tests/cluster/test_groups.py
@@ -9,15 +9,17 @@ from tests import config as conf
 from .test_users import get_random_string
 
 
-def det_cmd(cmd: List[str]) -> subprocess.CompletedProcess:
+def det_cmd(cmd: List[str], **kwargs: Any) -> subprocess.CompletedProcess:
     return subprocess.run(
-        ["det", "-m", conf.make_master_url()] + cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        ["det", "-m", conf.make_master_url()] + cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        **kwargs,
     )
 
 
 def det_cmd_json(cmd: List[str]) -> Any:
-    res = det_cmd(cmd)
-    assert res.returncode == 0
+    res = det_cmd(cmd, check=True)
     return json.loads(res.stdout)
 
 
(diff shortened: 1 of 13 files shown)


harness/determined/cli/user_groups.py
```diff
@@ -0,0 +1,229 @@
...
+)
+
+v1GroupHeaders = namedtuple(
+    "v1GroupHeaders",
+    ["groupId", "name"],
+)
+
+
+@authentication.required
+def create_group(args: Namespace) -> None:
```
eecsliu, line 27: I think we should check if group name is empty here

harness/determined/cli/user_groups.py
```diff
@@ -0,0 +1,229 @@
...
+            f"could not find users for usernames {', '.join(missing_users)}"
+        )
+    return user_ids
+
+
+def group_name_to_group_id(session: session.Session, group_name: str) -> int:
+    body = bindings.v1GroupSearchRequest(name=group_name)
+    resp = bindings.post_GetGroups(session, body=body)
+    groups = resp.groups
+    if groups is None or len(groups) != 1 or groups[0].groupId is None:
```
eecsliu, line 147: Maybe also check if groups == "" in here

harness/determined/cli/user_groups.py
```diff
@@ -0,0 +1,229 @@
...
+    resp = bindings.put_UpdateGroup(setup_session(args), groupId=group_id, body=body)
+
+    print(f"user group with ID {resp.group.groupId} name {resp.group.name}")
+    for user_id, username in zip(user_ids, usernames):
+        print(f"user removed from the group with username {username} and ID {user_id}")
+
+
+@authentication.required
+def change_group_name(args: Namespace) -> None:
+    session = setup_session(args)
```
eecsliu, line 104: check that `new_group_name` can't be empty

eecsliu (commented):
Great work! Couple of items that _should_ be covered by the db error handling, but you could address here.

harness/determined/cli/user_groups.py
NicholasBlaskey: Since group name is a positional arg here it can't be empty right?

harness/determined/cli/user_groups.py
NicholasBlaskey: ```new_group_name``` is also a positional arg

harness/determined/cli/user_groups.py
eecsliu: I'm not actually super sure about positional arg behavior, but it looks like `None` is provided as a default arg value, which would cause an error further down because the name field can't be null?

master/internal/config/config.go
```diff
@@ -290,6 +291,11 @@ func (t *TLSConfig) ReadCertificate() (*tls.Certificate, error) {
 	return &cert, err
 }
 
+// RBACConfig is the configuration for RBAC.
+type RBACConfig struct {
+	Enabled bool `json:"enabled"`
```
ioga, line 296: Maybe have it as an optional explicit field on AuthZConfig?

```

type AuthZConfig struct {
   ...
   RBACUIEnabled *bool `json:"rbac_ui_enabled"`
}

func (a *AuthZConfig) IsRBACUIEnabled() bool {
  if a.RBACUIEnabled != nil {
    return a.RBACUIEnabled
  }

  return a.Type != BasicAuthZType
}
```

this leaves a flexibility of explicitly enabling it but also adds sensible default

e2e_tests/tests/cluster/test_groups.py
```diff
@@ -0,0 +1,132 @@
...
+    group_name = get_random_string()
+    create_group_cmd = ["user-group", "create", group_name]
+    for add_user in add_users:
+        create_group_cmd += ["--add-user", add_user]
+
+    assert det_cmd(create_group_cmd).returncode == 0
+
+    # Can view through list.
+    group_list = det_cmd_json(["user-group", "list", "--json"])
+    assert sum([group["group"]["name"] == group_name for group in group_list["groups"]]) == 1
```
ioga, line 42: it's not exactly the same but 
```suggestion
    assert any([group["group"]["name"] == group_name for group in group_list["groups"]])
```

e2e_tests/tests/cluster/test_groups.py
```diff
@@ -0,0 +1,132 @@
...
+        assert sum([group["group"]["name"] == group_name for group in group_list["groups"]]) == 1
+
+    # Can describe properly.
+    group_desc = det_cmd_json(["user-group", "describe", group_name, "--json"])
+    assert group_desc["name"] == group_name
+    for add_user in add_users:
+        assert sum([u["username"] == add_user for u in group_desc["users"]]) == 1
+
+    # Can delete.
+    assert det_cmd(["user-group", "delete", group_name, "--yes"]).returncode == 0
```
ioga, line 58: will that print the error output when it fails? if no then maybe

```
def det_cmd(cmd: List[str], **kwargs) -> subprocess.CompletedProcess:
    return subprocess.run(
        ["det", "-m", conf.make_master_url()] + cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs
    )
```

and here

```suggestion
    det_cmd(["user-group", "delete", group_name, "--yes"], check=True)
```

e2e_tests/tests/cluster/test_groups.py
ioga, line 42: and if you want to check if it's really "only one" I'd think `len` is few characters longer but I think reflects the purpose better?
```suggestion
    assert len([group for group in group_list["groups"] if group["group"]["name"] == group_name]) == 1
```

harness/determined/cli/_util.py
```diff
@@ -84,3 +86,16 @@ def setup_session(args: argparse.Namespace) -> api.Session:
     cert = certs.default_load(master_url)
 
     return api.Session(master_url, args.user, authentication.cli_auth, cert)
+
+
+def require_feature_flag(feature_flag: str, error_message: str) -> Callable[..., Any]:
+    def decorator(function: Callable[..., Any]) -> Callable[..., Any]:
+        def wrapper(args: argparse.Namespace) -> None:
+            resp = bindings.get_GetMaster(setup_session(args))
+            if not resp.to_json()["rbacEnabled"]:
```
ioga, line 95:
```suggestion
            if not resp.to_json().get(featureFlag):
```

harness/determined/cli/user_groups.py
```diff
@@ -0,0 +1,262 @@
...
+        "user-group",
+        None,
+        "manage user groups",
+        [
+            Cmd(
+                "create",
+                create_group,
+                "creates a user group",
+                [
+                    Arg("group_name", default=None, help="name of user group to be created"),
```
ioga, line 190: since it's required, maybe shouldn't default to None?
```suggestion
                    Arg("group_name", help="name of user group to be created"),
```

(this should be applied to all required positional arguments here)

ioga (approved):

PR #4638, opened 2022-07-25
feat: task specific actions to job overflow menu

keita-determined:
## Description
[DET-7077]

## Test Plan

- Check if there are `view log` and `kill` buttons in the `notebooks`, `tensor boards`, `commands`, and `shells` job types and check the functionalities
- Check if there are `Cancel` and `kill` buttons in the `experiment` job type and check the functionalities

## Checklist

[DET-7077]: https://determinedai.atlassian.net/browse/DET-7077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/pages/JobQueue/JobQueue.tsx
```diff
@@ -98,30 +101,34 @@ const JobQueue: React.FC<Props> = ({ bodyNoPadding, selectedRp, jobState }) => {
 
   usePolling(fetchAll, { rerunOnNewFn: true });
 
+  const getCommandType = (jobType: Determinedjobv1Type): CommandType | null => {
```
keita-determined, line 104: Should I move this func to utility or leave it here?

hamidzr: nit: do we have a utils/job? let's put it there for discoverability.

webui/react/src/pages/JobQueue/JobQueue.tsx
```diff
@@ -97,30 +100,34 @@ const JobQueue: React.FC<Props> = ({ bodyNoPadding, selectedRp, jobState }) => {
...
-    };
+    const triggers: Triggers<JobAction> = {};
+    const commandType = getCommandType(job.type);
+
+    if (commandType) {
+      triggers[JobAction.Kill] = () => {
+        killTask({ id: job.entityId, type: commandType });
+      };
+      triggers[JobAction.ViewLog] = () => {
+        history.push(paths.taskLogs({ id: job.entityId, name: job.name, type: commandType }));
```
hamidzr, line 127: use `routeToReactUrl` util?

webui/react/src/routes/utils.ts
```diff
@@ -138,7 +138,7 @@ export const paths = {
   taskList: (): string => {
     return '/tasks';
   },
-  taskLogs: (task: CommandTask): string => {
+  taskLogs: (task: Pick<CommandTask, 'id' | 'name' | 'type'>): string => {
```
hamidzr, line 141: 👍🏼

webui/react/src/types.ts
```diff
@@ -697,6 +697,7 @@ export enum JobAction {
   Kill = 'Kill',
   ManageJob = 'Manage Job',
   MoveToTop = 'Move To Top',
+  ViewLog = 'View Task Logs',
```
hamidzr, line 700: just view Log?

hamidzr (commented):
this looks good. I'm a bit curious if we could reuse tasks or experiment action dropdowns or if it's not a good fit and means a lot of work

high-level actions for a job:
- actions generic to a job
- actions dependent on if the job is an experiment or NTbCS
  - in this case can we determine which one the job is, mold the job into the minimum needed for `taskActionDropDown` and ExperimentActionDropdown and reuse those?

webui/react/src/pages/JobQueue/JobQueue.tsx
keita-determined: the same function is already in utils/job!

keita-determined:
> this looks good. I'm a bit curious if we could reuse tasks or experiment action dropdowns or if it's not a good fit and means a lot of work
> 
> high-level actions for a job:
> 
> * actions generic to a job
> * actions dependent on if the job is an experiment or NTbCS
>   
>   * in this case can we determine which one the job is, mold the job into the minimum needed for `taskActionDropDown` and ExperimentActionDropdown and reuse those?

Technically we can unify these dropdown into one dropdown, but IMO it is not a good idea because each dropdown is already complex enough and if we wanna add something similar but slightly different in the future, the unified dropdown would be very complex. I think some redundancy is important for readability and maintainability.

hamidzr (approved):
LGTM. 
wdy think about using the same confirmation wording between kill/cancel via jobs view and our other viewsl? eg canceling an experiment through experiment list doesn't seem to ask for confirmation. also, the wording for killing an experiment seems different. "cancel" "abort" "kill" "ok" button labels might get confusing.

keita-determined:
> LGTM.
> wdy think about using the same confirmation wording between kill/cancel via jobs view and our other viewsl? eg canceling and experiment through experiment list doesn't seem to ask for confirmation. also, the wording for killing an experiment seems different. "cancel" "abort" "kill" "ok" button labels might get confusing.

Partially using the same component like confirmation is a good idea cuz its a smaller change than make everything generic

PR #4639, opened 2022-07-25
ci(lint-python): migrate to gha workflow

loksonarius:
## Description

Replicating the current lint-python CircleCI job as a GHA workflow. One
difference is that the GHA one runs all check tasks simultaneously.

Part of this also involves updating the bumpversion config so that it
uses a glob for all GHA workflows instead of just the single test-cli
workflow file.

## Test Plan

Make a PR and compare output of CircleCI and GHA workflows.

## Checklist


dannysauer (approved):
It'd be nice if 0.18.5.dev0 was a variable instead of being repeated in the run commands, but direct copy, etc, fine.

PR #4640, opened 2022-07-25
fix: model deital bugs

keita-determined:
## Description
[DET-7888] [DET-7848]

## Test Plan
- Check if we can update name, node, metadata, and description. Even after refresh the page, the updated data should be there.  Also, we can route to the version page property

## Checklist

[DET-7888]: https://determinedai.atlassian.net/browse/DET-7888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-7848]: https://determinedai.atlassian.net/browse/DET-7848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/Table.tsx
+++ b/webui/react/src/components/Table.tsx
@@ -165,7 +165,7 @@ export const modelNameRenderer = (value: string, record: ModelItem): React.React
 );
 
 export const modelVersionNameRenderer = (value: string, record: ModelVersion): React.ReactNode => (
-  <Link path={paths.modelVersionDetails(record.model.name, record.version)}>
+  <Link path={paths.modelVersionDetails(record.model.name, record.id)}>
     {value ? value : 'Version ' + record.version}
   </Link>
 );
--- a/webui/react/src/components/Table.tsx
+++ b/webui/react/src/components/Table.tsx
@@ -160,12 +160,12 @@ export const modelNameRenderer = (value: string, record: ModelItem): React.React
     <div style={{ paddingInline: 4 }}>
       <Icon name="model" size="medium" />
     </div>
-    <Link path={paths.modelDetails(String(record.id))}>{value}</Link>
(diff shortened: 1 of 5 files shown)


mapmeld:
I think that these are good changes, but they only update links... did we find and fix the original ticket problem?

mapmeld:
OK on review we actually do want the model paths to use `record.id` and not the name, because it's a more permanent link (name can be changed).

keita-determined:
> I think that these are good changes, but they only update links... did we find and fix the original ticket problem?

[patchModel](https://github.com/determined-ai/determined/blob/master/webui/react/src/pages/ModelDetails.tsx#L282) requires modelName as a parameter. Currently, link is used ID instead of name. That's why its not working with the ticket problem. But I'll find a way to keep the permalink

mapmeld:
`patchModel` should be able to see the ID is a number and adjust accordingly (`a.ModelFromIdentifier`)

keita-determined:
> `patchModel` should be able to see the ID is a number and adjust accordingly on the server (`a.ModelFromIdentifier`)

But somehow it doesn't work right now, and when I used model name, it worked

keita-determined:
It is not related but [this line](https://github.com/determined-ai/determined/pull/4640/commits/6a8018043c445e709e4ee11bcf32ffa840bd4539#diff-84ae63ef8b3a9d76b21ba552e425e32d39ed953241892423070ef0c13e721438R179) should be `id` instead of `name`?

mapmeld (approved):
Thanks for going through the changes, LGTM :+1:

PR #4646, opened 2022-07-26
fix: change average_training_metrics default to true [DET-7240]

garrett361:
## Description

BREAKING CHANGE: Change the default value of `average_training_metrics` in the `optimizations` section of the experiment config to `true`.  The current `false` default leads to only the chief's training metrics being reported, which is surprising behavior.

## Test Plan

Manual testing:
- Verify that changing the default doesn't break selected experiments in `examples/`
- Perform modest benchmarking to verify that this default does not degrade performance to an appreciable degree.

Also verify that changing this default does not break existing tests.

## Checklist

### DIFF
--- a/docs/reference/reference-training/experiment-config-reference.rst
+++ b/docs/reference/reference-training/experiment-config-reference.rst
@@ -1105,7 +1105,7 @@ experiment.
    For multi-GPU training, whether to average the training metrics across GPUs instead of only using
    metrics from the chief GPU. This impacts the metrics shown in the Determined UI and TensorBoard,
    but does not impact the outcome of training or hyperparameter search. This option is currently
-   only supported in PyTorch. Defaults to ``false``.
+   only supported in PyTorch. Defaults to ``true``.
 
 ``gradient_compression``
    Whether to compress gradients when they are exchanged during :ref:`multi-gpu-training`.
--- a/docs/reference/reference-training/experiment-config-reference.rst
+++ b/docs/reference/reference-training/experiment-config-reference.rst
@@ -1105,7 +1105,7 @@ experiment.
    For multi-GPU training, whether to average the training metrics across GPUs instead of only using
    metrics from the chief GPU. This impacts the metrics shown in the Determined UI and TensorBoard,
    but does not impact the outcome of training or hyperparameter search. This option is currently
-   only supported in PyTorch. Defaults to ``true``.
+   supported for ``PyTorchTrial`` and ``TFKerasTrial`` instances. Defaults to ``true``.
 
 ``gradient_compression``
    Whether to compress gradients when they are exchanged during :ref:`multi-gpu-training`.
--- a/docs/release-notes/4646-default-average_training_metrics-true.txt
+++ b/docs/release-notes/4646-default-average_training_metrics-true.txt
@@ -3,4 +3,8 @@
 **Breaking Changes**
 
 -  Experiment: the ``average_training_metrics`` experiment configuration default has been changed to
-   ``True``.
+   ``True``.  This change only affects distributed training.  The previous default of ``False``
+   leads to only the chief worker's training metrics being reported. Setting this configuration to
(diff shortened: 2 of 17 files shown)


docs/reference/reference-training/experiment-config-reference.rst
```diff
@@ -1105,7 +1105,7 @@ experiment.
    For multi-GPU training, whether to average the training metrics across GPUs instead of only using
    metrics from the chief GPU. This impacts the metrics shown in the Determined UI and TensorBoard,
    but does not impact the outcome of training or hyperparameter search. This option is currently
-   only supported in PyTorch. Defaults to ``false``.
+   only supported in PyTorch. Defaults to ``true``.
```
rb-determined-ai, line 1108: it's actually supported in PyTorchTrial and TFKerasTrial.

Also, now that we have Core API, referring to PyTorchTrial as "PyTorch" isn't really specific enough anymore.

garrett361: It's also marked at supported for DeepSpeedTrials, but possibly not accurately.

garrett361: I changed this to "PytorchTrial and TFKerasTrial instances" for now. Left DeepSpeed out of it, provisionally.  There are also other instances of bare "PyTorch" that could be replaced, but I left them for a larger docs clean-up PR.

harness/determined/estimator/_estimator_trial.py
```diff
@@ -430,6 +431,12 @@ def __init__(
 
         self._init_model()
 
+    @classmethod
+    def supports_average_training_metrics(cls: Type["EstimatorTrialController"]) -> bool:
+        # Return True with a warning in order to pass tests.
+        warnings.warn("EstimatorTrialController does not support average_training_metrics = True")
+        return True
```
rb-determined-ai, line 438: can you just make the change in the `_check_if_trial_supports_configurations()` instead?

This pattern of "everything is a trial" is going away, so the value of that check is approaching zero.  Part of the Trainer API is to move training-loop settings like average_training_metrics out of the expconf and into the Trainer APIs, where individual framework APIs can handle it themselves.

garrett361: Oh, do you just want me to remove `_check_if_trial_supports_configurations` entirely? `experiment_config.average_training_metrics_enabled` is the only thing it checks.  I can also remove `supports_average_training_metrics` everywhere, if we do that

garrett361: Doing the above.

docs/release-notes/4646-default-average_training_metrics-true.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Breaking Changes**
+
+-  Experiment: the ``average_training_metrics`` experiment configuration default has been changed to
+   ``True``.
```
rb-determined-ai, line 6: Please elaborate a bit more here.  If I'm the user, what action do I need to take as a result of this "breakage"?  Or how can I decide if I need to take action at all?

rb-determined-ai (approved):

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -282,6 +279,15 @@ def _run(self) -> None:
...
                         "stop_requested": self.context.get_stop_requested(),
                     }  # type: workload.Response
+                    if (
+                        self.context.distributed.size > 1
+                        and not self.context._average_training_metrics
+                    ):
+                        warnings.warn(
+                            "Only the chief worker's training metrics are being reported, due "
+                            "to setting average_training_metrics to False.",
+                            FutureWarning,
```
rb-determined-ai, line 289: this shouldn't be a future warning.  Maybe make it a `UserWarning` or a `RuntimeWarning`.  Docs [here](https://docs.python.org/3/library/warnings.html#warning-categories)

rb-determined-ai (approved):

PR #4650, opened 2022-07-27
test: add test cases for `AvatarCard`

keita-determined:
## Description
[DET-7896]

## Test Plan
- Check test cases are covered enough

## Commentary (optional)
I added storybook as well.

## Checklist

[DET-7896]: https://determinedai.atlassian.net/browse/DET-7896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hamidzr (approved):
LGTM. thansk for testing these. this might need to be on upstream to pass the integration test.

webui/react/src/shared/components/Avatar/Avatar.tsx
```diff
@@ -50,7 +50,7 @@ const Avatar: React.FC<Props> = (
   if (large) classes.push(css.large);
   if (className) classes.push(className);
   const avatar = (
-    <div className={classes.join(' ')} id="avatar" style={style}>
+    <div className={classes.join(' ')} data-testid="avatar-element" id="avatar" style={style}>
```
hamidzr, line 53: nit: is the change just to facilitate the tests? I'd guess we had added `id` for the same purpose. maybe we don't need both?

keita-determined: oh yeah. I couldn't see the `id` 😄

hamidzr: 👍🏼  I don't know which one is better to keep/use.

keita-determined: I think `id` is better in this case? https://stackoverflow.com/questions/55151142/react-testing-library-why-use-test-id

PR #4656, opened 2022-07-27
fix: description overflowing table cell

trentwatt:
## Before:

image

image

## After: 

image

image

## Test Plan
Check light and dark mode, sorting by different things, change description text to various long things, etc.

## Commentary (optional)

dark mode color for sorted column does not look ideal IMO. cc @alexjdetermined 

had to manually specific background colors for the table cells. the existing background colors for table are not using the theme architecture pattern so it is not clear what the colors should be. 

## Checklist


alexjdetermined (approved):
LGTM

trentwatt:
@hkang1 any concerns about the css? the `!important` is there to override ant table styles that are more specific

hkang1:
> @hkang1 any concerns about the css? the `!important` is there to override ant table styles that are more specific

Is there any way of writing the rules to be more specific? Another option is to apply this at the `antd.scss` level. I generally use `!important` as a last resort after an exhaustive set of approaches, just because the hijacking nature of !important usage.

webui/react/src/shared/styles/antd.scss
```diff
@@ -217,9 +217,12 @@ html {
     opacity: 1;
     user-select: auto;
   }
-  .ant-table .ant-table-tbody > tr > td[class*=ant-table-cell-fix-] {
```
trentwatt, line 220: is there any reason why the overflow menu should have a different background color than the other cells?

webui/react/src/shared/styles/antd.scss
```diff
@@ -217,9 +217,12 @@ html {
...
+  .ant-table .ant-table-tbody > tr > td.ant-table-cell {
     background-color: var(--theme-stage-weak);
   }
+  .ant-table .ant-table-tbody > tr > td.ant-table-cell.ant-table-column-sort {
+    filter: invert(2%);
```
trentwatt, line 224: not sure if `filter` is the preferred way to do this, but this at least preserves the visual of:
- highlighted row + sorted column is stronger color than either of the two individually (initial PR did not)

image

webui/react/src/shared/styles/antd.scss
```diff
@@ -217,9 +217,12 @@ html {
     opacity: 1;
     user-select: auto;
   }
-  .ant-table .ant-table-tbody > tr > td[class*=ant-table-cell-fix-] {
+  .ant-table .ant-table-tbody > tr > td.ant-table-cell {
     background-color: var(--theme-stage-weak);
```
trentwatt, line 221: according to theme arch doc, i would think this would be `theme-surface`, as you mentioned in chat.

however i have kept this PR to just target the description overflow.

and we can revisit table color separately if desired.

trentwatt:
> > @hkang1 any concerns about the css? the `!important` is there to override ant table styles that are more specific
> 
> Is there any way of writing the rules to be more specific? Another option is to apply this at the `antd.scss` level. I generally use `!important` as a last resort after an exhaustive set of approaches, just because the hijacking nature of !important usage.

@hkang1 

OK, I have done this. Thanks for bringing up `antd.scss`, it hadn't occured to me that was an option.

PR #4658, opened 2022-07-28
fix: Count only active tasks in cluster info board

mapmeld:
## Description

Fixing a bug for the release - previously any TensorBoard / JupyterLab / Command / Shell which would be visible on the Tasks page was being counted, but we really only want the task.state == Active ones.

Alternatives which didn't work:
- the GetTensorboardsRequest and other request objects can be filtered by user, but not by state
- I was unable to write a count function that accepts all types. The proto objects cannot be made subclasses of a universal Task type. I also tried converting the response's `[]*tensorboardv1.Tensorboard` into a new `[]*universalTask` type requiring only a State, but type conversion might require looping over each element in the slice anyway.

## Test Plan

- Start a TensorBoard or JupyterLab.
- Observe that it is counted as Active on the Cluster page.
- Kill the task on the Tasks page. Observe that it is still listed as Terminated on the Tasks page.
- Return to the Cluster page; it is not counted.

## Checklist


gt2345 (approved):
LGTM, but I recommend having someone more familiar with backend to confirm.

master/internal/api_tasks.go
```diff
@@ -143,35 +144,56 @@ func (a *apiServer) TaskLogs(
 func (a *apiServer) GetTasksCount(
 	_ context.Context, req *apiv1.GetTasksCountRequest,
 ) (resp *apiv1.GetTasksCountResponse, err error) {
-	finalResp := &apiv1.GetTasksCountResponse{}
+	finalResp := &apiv1.GetTasksCountResponse{
```
dzhu, line 147: These don't need to be explicitly initialized, and I'd say it's more idiomatic to not do so.

master/internal/api_tasks.go
```diff
@@ -143,35 +144,56 @@ func (a *apiServer) TaskLogs(
 func (a *apiServer) GetTasksCount(
```
dzhu, line 144: Could you rename everything from `GetTasks` to `GetActiveTasks`?

(Or maybe this should return both the total task count and the active task count, or allow for filtering to only active tasks. Depends on how it's intended to be used, which I'm not familiar with.)

master/internal/api_tasks.go
```diff
@@ -143,35 +144,56 @@ func (a *apiServer) TaskLogs(
...
+		Tensorboards: 0,
+	}
 
 	req1 := &apiv1.GetNotebooksRequest{}
 	resp1 := &apiv1.GetNotebooksResponse{}
 	if err = a.ask(notebooksAddr, req1, &resp1); err != nil {
 		return nil, err
 	}
-	finalResp.Notebooks = int32(len(resp1.Notebooks))
+	for _, n := range resp1.Notebooks {
```
dzhu, line 159: Just as a general note, all the loops could be written a bit more nicely with generics, but no need to get fancy here, probably.

master/internal/api_tasks.go
mapmeld: ✅

PR #4662, opened 2022-07-28
fix: persist task state update in interactive task view [DET-7929]

julian-determined-ai:
## Description

fix: persist task state update in interactive task view [DET-7929]

The previous implementation of tracking the `Command` state for the `InteractiveTask` did not fully solve the problem since the `Wait` page ultimately gets redirected and the `Command` state polling is stopped. Since the polling would stop users would not be notified of any `Command` state changes. 

The initial implementation was chosen as a way to avoid additional polling for the command state but this PR does add additional polling within the `InteractiveTask` component. The decision comes with two advantages: 

1. Now the `InteractiveTask` page will always receive the most up to date information about the appropriate task, and no longer depend on the `Wait` component,  which will be beneficial as more features are added to the page itself. 
2. The `Wait` component does eventually stop polling once a task is either `ready` or `terminated` so the dual polling only occurs for a short period of time meaning resources are not wasted for identical API calls.

  

## Test Plan
 
1. Start a task ensure that during all tasks states the page title does not change while the tab in the browser is active.
2. Start a task and open another tab in the same browser, the page title for the original task tab should update to the most recent task state.
    - Ensure that when the task is `Running` the page title is the normal title "Tasks - Determined" 
    - Ensure once a task is terminated the page title is "Terminated"
Note: The task state is only updated every 2 seconds so if a state changes very rapidly you may not see the state change in the tab which is expected.

## Commentary (optional)

Polling is configured to only occur ever `2` seconds. If we find this is not fast enough for the user we can update the configuration.

## Checklist

[DET-7929]: https://determinedai.atlassian.net/browse/DET-7929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/InteractiveTask.tsx
+++ b/webui/react/src/pages/InteractiveTask.tsx
@@ -4,7 +4,9 @@ import { useParams } from 'react-router-dom';
 
 import TaskBar from 'components/TaskBar';
 import { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
+import { getTask } from 'services/api';
 import { CommandState, CommandType } from 'types';
+import handleError from 'utils/error';
 
 import css from './InteractiveTask.module.scss';
 import TaskLogs from './TaskLogs';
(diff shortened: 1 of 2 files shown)


mapmeld (approved):
LGTM

PR #4667, opened 2022-07-28
test: add test cases for `Image.tsx`

keita-determined:
## Description
[DET-7898]

## Test Plan
- Check if all test cases are meaningful, and not miss anything for test cases

## Checklist

[DET-7898]: https://determinedai.atlassian.net/browse/DET-7898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/shared/components/Image/Image.test.tsx
+++ b/webui/react/src/shared/components/Image/Image.test.tsx
@@ -1,0 +1,125 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+
+import { DarkLight } from 'shared/themes';
+
+import { ImageAlert, ImageEmpty, ImageWarning } from './Image';
+import type { Props } from './Image';
+
+const setupImageAlert = (props?: Props) => {
+  const view = render(<ImageAlert {...props} />);
+  return { view };
+};
+
+const setupImageEmpty = () => {
+  const view = render(<ImageEmpty />);
+  return { view };
+};
+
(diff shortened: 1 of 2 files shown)


webui/react/src/shared/components/Image/Image.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+import React from 'react';
+
+import { DarkLight } from 'shared/themes';
+
+import { ImageAlert, ImageEmpty, ImageWarning } from './Image';
+import type { Props } from './Image';
+
+const setupImageAlert = (props?: Props) => {
+  const view = render(<ImageAlert {...props} />);
+  return { view };
```
hkang1, line 11: **nit:** totally up to you, but I don't foresee any future properties getting added and returned besides `view`. It could probably just return it straight up instead of wrapped in an object.

hkang1 (approved):
LGTM! Nice work with the tests!

webui/react/src/shared/components/Image/Image.test.tsx
keita-determined: I agree that any properties wouldn't be added in the future, but if it happens, that would be painful to change everything. I just wanted to make it flexible, so I'll leave it

PR #4668, opened 2022-07-28
fix: shutdown operation [DET-7920]

nrajanee:
## Description

fix: shutdown operation [DET_7920]

## Test Plan 

No tests required.

### DIFF
--- a/master/pkg/searcher/searcher.go
+++ b/master/pkg/searcher/searcher.go
@@ -110,6 +110,7 @@ func (s *Searcher) TrialExitedEarly(
 	}
 	s.Exits[requestID] = true
 
+	s.Record(operations)
 	_, isCustomSMethod := s.method.(CustomSearchMethod)
 	// for non custom-search methods you can make the assumption
 	// that trials will be created immediately.
@@ -118,7 +119,6 @@ func (s *Searcher) TrialExitedEarly(
 		operations = append(operations, shutdown)
 	}
 
-	s.Record(operations)
 	return operations, nil
 }
 
@@ -155,6 +155,7 @@ func (s *Searcher) TrialClosed(requestID model.RequestID) ([]Operation, error) {
 		return nil, errors.Wrapf(err, "error while handling a trial closed event: %s", requestID)
 	}
 
+	s.Record(operations)
 	_, isCustomSMethod := s.method.(CustomSearchMethod)
 	// for non custom-search methods you can make the assumption
 	// that trials will be created immediately.
--- a/master/pkg/searcher/searcher.go
+++ b/master/pkg/searcher/searcher.go
@@ -111,7 +111,8 @@ func (s *Searcher) TrialExitedEarly(
 	s.Exits[requestID] = true
 
 	_, isCustomSMethod := s.method.(CustomSearchMethod)
-	// for non custom-search methods you can make the assumption that trials will be created immediately.
+	// for non custom-search methods you can make the assumption
+	// that trials will be created immediately.
 	if s.TrialsRequested == len(s.TrialsClosed) && !isCustomSMethod {
 		shutdown := Shutdown{Failure: len(s.Failures) >= s.TrialsRequested}
 		operations = append(operations, shutdown)
@@ -155,7 +156,8 @@ func (s *Searcher) TrialClosed(requestID model.RequestID) ([]Operation, error) {
 	}
 
 	_, isCustomSMethod := s.method.(CustomSearchMethod)
-	// for non custom-search methods you can make the assumption that trials will be created immediately.
+	// for non custom-search methods you can make the assumption
+	// that trials will be created immediately.
 	if s.TrialsRequested == len(s.TrialsClosed) && !isCustomSMethod { // if not custom search method.
 		shutdown := Shutdown{Failure: len(s.Failures) >= s.TrialsRequested}
 		s.Record([]Operation{shutdown})
--- a/master/pkg/searcher/searcher.go
+++ b/master/pkg/searcher/searcher.go
@@ -109,6 +109,13 @@ func (s *Searcher) TrialExitedEarly(
 		return nil, errors.Wrapf(err, "error relaying trial exited early to trial %d", requestID)
 	}
 	s.Exits[requestID] = true
+
+	_, isCustomSMethod := s.method.(CustomSearchMethod)
+	if s.TrialsRequested == len(s.TrialsClosed) && !isCustomSMethod {
+		shutdown := Shutdown{Failure: len(s.Failures) >= s.TrialsRequested}
+		operations = append(operations, shutdown)
+	}
+
 	s.Record(operations)
 	return operations, nil
 }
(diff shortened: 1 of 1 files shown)


master/pkg/searcher/random.go
```diff
@@ -126,6 +127,9 @@ func (s *randomSearch) trialClosed(ctx context, requestID model.RequestID) ([]Op
 		ops = append(ops, NewClose(create.RequestID))
 		s.CreatedTrials++
 		s.PendingTrials++
+	} else if s.PendingTrials == 0 { // even if the number of created trials == MaxTrials there still might be active trials.
+		shutdown := Shutdown{Failure: len(s.Failures) >= s.TrialsRequested}
```
aciborowska, line 131: I think we should add this code in `trialExitedEarly`.

Let's say we are running last trial and it ends with `trialExitedEarly`. We have two options:
* if exitedReason is related to HP then we go into `if` and create a new trial
* if exitedReason is not related to HP, then we return `nil, nil`. If I am not missing anything `nil, nil` goes to searcher.go and then to searcher.record function, which expects Create or Shutdown operation, so (I assume) `nil` will be ignored. Since this was our last trial, we will never get trial closed and (probably) never shut down the experiment.

master/pkg/searcher/random.go
```diff
@@ -25,6 +25,7 @@ type (
 	// randomSearch corresponds to the standard random search method. Each random trial configuration
 	// is trained for the specified number of steps, and then validation metrics are computed.
 	randomSearch struct {
+		SearcherState
```
mpkouznetsov, line 28: 1. Not sure it is a good idea to make all search methods be aware of the SearcherState.
2. More importantly, if we do this we have to use a pointer here to the same SearcherState that is used by Searcher.

master/pkg/searcher/asha.go
```diff
@@ -170,7 +171,30 @@ func (s *asyncHalvingSearch) trialClosed(
 ) ([]Operation, error) {
 	s.TrialsCompleted++
 	s.ClosedTrials[requestID] = true
-	return nil, nil
+	var ops []Operation
+	if s.sendShutdownOperation() {
```
nrajanee, line 175: Please note we don't need do this in trialExitedEarly in ASHA because if the last trial did not exit because of a HP related reason then it will be marked as a EarlyExitedTrial and it will call s.promoteAsync where this trial will be closed. Please see: https://github.com/determined-ai/determined/blob/d42ada95bc3d6a334f3322b3c9c6755085844b8a/master/pkg/searcher/asha.go#L345.

master/pkg/searcher/asha.go
```diff
@@ -170,7 +172,30 @@ func (s *asyncHalvingSearch) trialClosed(
...
+	}
+
+	return ops, nil
+}
+
+func (s *asyncHalvingSearch) sendShutdownOperation() bool {
+	shutdown := true
+	allValidTrials := len(s.Rungs[0].Metrics)
+	allCreatedTrials := len(s.Rungs[0].Metrics) + s.InvalidTrials
+	if s.PendingTrials > 0 || allCreatedTrials > s.TrialsCompleted || allValidTrials < s.MaxTrials() {
```
nrajanee, line 190: s.PendingTrials > 0  and allValidTrials < s.MaxTrials() conditions are similar to random and grid search. 

The check allCreatedTrials > s.TrialsCompleted has to be made to ensure that all the Close Operations are processed (even for Invalid Trials) otherwise there might be some left over Close operations that aren't processed and searcher will shutdown prematurely. This is different from others because in ASHA NewClose is not created right after NewCreate as you'll see in random and grid. 

OutStandingTrials for each rung counts any trial created in the bottom rung or promoted to other rungs. The summation of OutstandingTrials in all rungs should always be <= s.PendingTrials. So, I think we can remove this.  Any thoughts?

master/pkg/searcher/asha.go
```diff
@@ -170,7 +172,30 @@ func (s *asyncHalvingSearch) trialClosed(
...
+		return false
+	}
+	for _, rung := range s.Rungs {
```
nrajanee, line 193: OutStandingTrials for each rung counts any trial created in the bottom rung or promoted to other rungs. The summation of OutstandingTrials in all rungs should always be <= s.PendingTrials. So, I think we can remove this.  Any thoughts?

nrajanee, line 193: This check ensures that there are no more OutstandingTrials (trials created in the bottom rung or trials promoted to the next rung) in each rung.

mpkouznetsov (approved):
LGTM

master/pkg/searcher/tournament.go
```diff
@@ -139,11 +141,28 @@ func (s *tournamentSearch) Unit() expconf.Unit {
...
+		case Shutdown:
+			s.ShutdownSubSearches[subSearchID] = operation
+		default:
+			res = append(res, operation)
 		}
 	}
-	return operations
+	if len(s.ShutdownSubSearches) == len(s.subSearches) {
+		// does tournament fail iff any subsearch fails
+		// or does it fail iff all subsearches fail?
```
aciborowska, line 158: The answer from Liam is `if any of the sub searchers failed`, so the current implementation is correct.

aciborowska (approved):
LGTM

master/pkg/searcher/tournament.go
mpkouznetsov: in this case, @nrajanee, would you mind removing the comments and adding `break` after `failed = true`.

mpkouznetsov (approved):

aciborowska (approved):

master/pkg/searcher/searcher.go
```diff
@@ -109,6 +109,15 @@ func (s *Searcher) TrialExitedEarly(
...
+
+	_, isCustomSMethod := s.method.(CustomSearchMethod)
+	// for non custom-search methods you can make the assumption
+	// that trials will be created immediately.
+	if s.TrialsRequested == len(s.TrialsClosed) && !isCustomSMethod {
+		shutdown := Shutdown{Failure: len(s.Failures) >= s.TrialsRequested}
+		operations = append(operations, shutdown)
+	}
+
 	s.Record(operations)
```
aciborowska, line 121: In the [fix made by Danny](https://github.com/determined-ai/determined/commit/2c575fd946a38f5e6e75eb2657b6d81b47d5b068) `s.Record(operations)` is invoked before checking for shutdown to ensure all operations are recorded. I think we should follow the same logic here.

dzhu: Yep, both places should have the logic from that fix.

dzhu (approved):

dzhu:
Whoops, I left in some old comments from the previous version; ignore those. Just look at the one about the `s.Record` calls.

gh-determined-ai (approved):
Is there a release note for this?

mpkouznetsov:
> Is there a release note for this?

There will be a release note for the whole feature of custom searcher, as well as some updates to the documentation. We were planning to involve you, @gh-determined-ai,  soon, as design solidifies.

PR #4670, opened 2022-07-28
refactor: remove unused code in model version detail page

keita-determined:
## Description
I found a minor bug in ModelVersionDeital.tsx which is about router issue.
`/det` is a env variable so removed it. However, I noticed the code is not used at all, so removed. 
Also, after deleting the version, it originally routes to the model list instead of model detail. I think it's unnatural so change it to model detail after deletion

## Test Plan
Check if it would route to the model detail page after deleting the version

## Checklist


mapmeld (approved):
LGTM. Actions tested: deleting a model from the model-version details page

PR #4675, opened 2022-07-29
feat: Async deleting workspaces and projects

mapmeld:
## Description

- Removes UI which hides the delete feature
- Deleting an empty project/workspace continues to work immediately (added new `completed: true` attribute to response)
- Deleting used projects starts an async process to delete experiments (until one fails), then deletes the project
- Deleting used workspace starts an async process to delete projects (until one fails) with the same project-deletion function, then deletes the workspace

WIP: We're discussing how this can use state badges, notifications, (?) to make the process and errors more transparent to the user. 

## Test Plan

Currently outputs go to the log

- Create and delete a workspace without adding anything
- Create a workspace, add empty projects, and delete the workspace
- Create a workspaces, add projects, add experiments inside some of the projects (`det e create sample/const.yaml sample/ --project_id 248`)

## Checklist


webui/react/src/hooks/useModal/Project/useModalProjectDelete.tsx
```diff
@@ -39,8 +39,10 @@ const useModalProjectDelete = ({ onClose, project }: Props): ModalHooks => {
 
   const handleOk = useCallback(async () => {
     try {
-      await deleteProject({ id: project.id });
-      routeToReactUrl(paths.workspaceDetails(project.workspaceId));
+      const response = await deleteProject({ id: project.id });
+      if (response.completed) {
```
EmilyBonar, line 43: question: is it correct to do nothing if `response.completed` is false?

webui/react/src/hooks/useModal/Workspace/useModalWorkspaceDelete.tsx
```diff
@@ -41,8 +43,10 @@ const useModalWorkspaceDelete = ({ onClose, workspace }: Props): ModalHooks => {
 
   const handleOk = useCallback(async () => {
     try {
-      await deleteWorkspace({ id: workspace.id });
-      routeToReactUrl(paths.workspaceList());
+      const response = await deleteWorkspace({ id: workspace.id });
+      if (response.completed) {
```
EmilyBonar, line 47: same question

EmilyBonar (commented):
I think this generally looks good, but I'm not sure I'm able to fully sign off on the backend code. This should probably have a release note since it's a user-facing change.

webui/react/src/hooks/useModal/Project/useModalProjectDelete.tsx
mapmeld: I'm open to `routeToReactUrl(paths.workspaceDetails(project.workspaceId));` in both cases. As a tester it was helpful to stay on the page and see that the experiments were getting deleted.

harness/determined/cli/project.py
```diff
@@ -134,20 +134,17 @@ def describe_project(args: Namespace) -> None:
...
+        "For a recoverable alternative, see the 'archive' command. Do you still \n"
         "wish to proceed?"
     ):
-        bindings.delete_DeleteProject(sess, id=p.id)
-        print(f"Successfully deleted project {args.project_name}.")
+        resp = bindings.delete_DeleteProject(sess, id=p.id)
+        if resp.completed:
+            print(f"Successfully deleted project {args.project_name}.")
+        else:
+            print(f"Started deletion of project {args.project_name}...")
```
ioga, line 147: from CLI UX perspective I'd say the CLI should be able to wait for completion.
- if waiting, report an error if it occurs.
- waiting could be optionally disabled through a flag such as `--no-wait`.

webui/react/src/hooks/useModal/Project/useModalProjectDelete.tsx
EmilyBonar: My main concern is what happens once all of the experiments are deleted? Does the page just error out? Your call, ulitmately.

mapmeld: OK, I will remove the change to URL behavior on both modals. The only changes to the modals will be saying that it deletes all experiments. The delete option is appearing in WebUI only for workspaces/projects without experiments, so it will just be the wording changing.

ioga (approved):

PR #4676, opened 2022-07-29
refactor: authz provider implementation and authz users basic implementation  

NicholasBlaskey:
## Description

Refactor the user api both echo and grpc to use the new ```authz``` pattern.

## Test Plan

existing e2e tests + new integration tests

## Checklist


master/internal/api_user.go
```diff
@@ -207,10 +239,20 @@ func (a *apiServer) PatchUser(
...
+		return nil, err
 	}
+	targetUser := targetFullUser.ToUser()
+	if err = user.AuthZProvider.Get().CanSetUsersDisplayName(*curUser, targetUser); err != nil {
+		if !user.AuthZProvider.Get().CanGetUser(*curUser, targetUser) {
+			return nil, errUserNotFound
+		}
+		return nil, errors.Wrap(grpcutil.ErrPermissionDenied, err.Error())
+	}
+
```
eecsliu, line 255: no check for if the user can patch other users?

master/internal/config/authz_config.go
```diff
@@ -0,0 +1,80 @@
...
+
+func initAuthZTypes() {
+	authZConfigMutex.Lock()
+	defer authZConfigMutex.Unlock()
+
+	if knownAuthZTypes != nil {
+		return
+	}
+
+	knownAuthZTypes = make(map[string]bool)
```
eecsliu, line 63: It seems like `knownAuthZTypes` can be remade if this is called more than once? Should be locked down with a once like in the provider

eecsliu (commented):
Looks good! Couple of questions below

master/internal/config/authz_config.go
NicholasBlaskey: There is a nil check for ```knownAuthZTypes``` above -- which I missed for like a couple minutes wondering how it even worked lol

eecsliu: oh lmao I'm just blind

eecsliu:
Is this dependent on Ilia's authz work? Would you want to merge this into `rbac-user-groups` branch on top of his changes before merging all of the user groups stuff into master together?

NicholasBlaskey:
> Is this dependent on Ilia's authz work? Would you want to merge this into `rbac-user-groups` branch on top of his changes before merging all of the user groups stuff into master together?

Ilia's authz work is already in this PR and I think I would rather to keep commit history cleaner keeping this change separate?  

Also I think I changed my mind about ```CanGetUser``` not returning an error. I am thinking it would be pretty bad UX for a user to like try to do an operation on another user and get not found when they should get a database failure error. I'm going to make ```CanGetUser``` return ```(canViewUser bool, serverError error)``` to try and make it clear that an error shouldn't be returned for denying viewing a user.

eecsliu (approved):

PR #4685, opened 2022-08-01
test: add samlauth tests [DET-7946]

hkang1:
## Description

Add unit test coverage SAML Auth.

## Checklist

### DIFF
--- a/webui/react/src/ee/SamlAuth.test.ts
+++ b/webui/react/src/ee/SamlAuth.test.ts
@@ -1,0 +1,64 @@
+import history from 'shared/routes/history';
+
+import * as utils from './SamlAuth';
+
+describe('SamlAuth', () => {
+  describe('samlUrl', () => {
+    const BASE_PATHS = [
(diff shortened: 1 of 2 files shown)


webui/react/src/ee/SamlAuth.test.ts
```diff
@@ -0,0 +1,64 @@
+import history from 'shared/routes/history';
+
+import * as utils from './SamlAuth';
```
hamidzr, line 3: nit: we do name these like this in many of out tests but why not make it specific to what's being imported. eg 'saml' instead of utils

hamidzr (approved):

webui/react/src/ee/SamlAuth.test.ts
hkang1: maybe `utils` is not the best name as it's only applicable to actual utils. Maybe `target` or `subject`? The thinking is that no matter what test you are looking at, the scope of everything you want to test is under this common phrase/name...

PR #4691, opened 2022-08-01
fix: cursor in modal text field jumps to end of input

keita-determined:
## Description
[DET-7392] [DET-7225]

## Test Plan
- Check if cursor won't jump to the end of text in `new Model modal`, `new project modal` in workspace and `new Workspace modal`
- Check if warning will emerge when there is a duplicate name in models
- Check if we can create model in model registry with name substring of existing model

## Commentary (optional)
- newly added `react-hooks-form` to fix this issue and for future form use. This library is better than other libraries like formik, react form and more in terms of performance and bundle size. What do you think?

## Checklist

[DET-7392]: https://determinedai.atlassian.net/browse/DET-7392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DET-7225]: https://determinedai.atlassian.net/browse/DET-7225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/hooks/useModal/Model/useModalModelCreate.module.scss
```diff
@@ -7,7 +7,7 @@
     margin: 0;
   }
   .uniqueWarning {
-    color: var(--theme-colors-danger-normal);
```
keita-determined, line 10: This variable doesnt work anymore

webui/react/package.json
```diff
@@ -45,6 +45,7 @@
     "react-dom": "^17.0.2",
     "react-draggable": "^4.4.4",
     "react-helmet-async": "^1.1.2",
+    "react-hook-form": "^7.34.0",
```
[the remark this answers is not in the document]
keita-determined: [antd Form](https://ant.design/components/form/) doesn't work in this case because it is rendered every time I hit a key. Good thing about react hooks form is that it only renders once so its efficient. However, I agree not to bring new libraries especially there is a similar library. 
I'll find other way without react hooks form.

webui/react/src/hooks/useModal/Model/useModalModelCreate.tsx
```diff
@@ -155,14 +155,14 @@ const useModalModelCreate = (): ModalHooks => {
         </p>
         <div>
           <h2>Model name</h2>
-          <Input value={modelName} onChange={updateModelName} />
```
keita-determined, line 158: I found a solution without react hook form.
All I need to do is to remove the value from props

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
```diff
@@ -20,12 +21,22 @@ interface Props {
 const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
   const [ name, setName ] = useState('');
   const [ description, setDescription ] = useState('');
+  const [ isNameUnique, setIsNameUnique ] = useState<boolean>(true);
 
   const { modalOpen: openOrUpdate, modalRef, ...modalHook } = useModal({ onClose });
 
+  const findIsNameUnique = useCallback(async (name: string) => {
+    const projectList = await getWorkspaceProjects({ id: workspaceId });
```
hkang1, line 29: **chore:** API calls, we'll want to wrap in a try/catch block because it can cause errors to be thrown.

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
```diff
@@ -20,12 +21,22 @@ interface Props {
 const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
   const [ name, setName ] = useState('');
   const [ description, setDescription ] = useState('');
+  const [ isNameUnique, setIsNameUnique ] = useState<boolean>(true);
 
   const { modalOpen: openOrUpdate, modalRef, ...modalHook } = useModal({ onClose });
 
+  const findIsNameUnique = useCallback(async (name: string) => {
```
hkang1, line 28: **question:** does the API endpoint of creating the project return an error if a project with the same name has been provided? If so, it might be better to just let the creation of the project catch this problem instead of proactively trying to fetch the full list of project names each time and compare it to the input.

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
hkang1: just noting that this won't be needed if we let the backend handle the not-a-unique-name catching

hkang1 (commented):
The cursor issue seems to work great on the preview! There are some questions specific to the workspace project name uniqueness logic.

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
```diff
@@ -36,15 +47,18 @@ const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
       <div className={css.base}>
         <div>
           <label className={css.label} htmlFor="name">Name</label>
-          <Input id="name" maxLength={80} value={name} onChange={handleNameInput} />
+          <Input id="name" maxLength={80} onChange={handleNameInput} />
```
hkang1, line 50: **suggestion:** this is a required field I believe? A good form pattern to follow (using `Form.Item`) can be `useModalPasswordChange.tsx`

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
keita-determined: Yes, API catches the error. I implemented cuz useModalModelCreate.tsx has one but not others. I'll remove it from all

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
keita-determined: it is actually checked [here](https://github.com/determined-ai/determined/pull/4691/files#diff-565663c0bf42a622416ed00decc158f8f7074bd7e4f19c84f95d85f823a650daR83), but I'll consider using it

webui/react/src/hooks/useModal/Project/useModalProjectCreate.module.scss
```diff
@@ -7,3 +7,7 @@
     color: var(--theme-float-on);
   }
 }
+.uniqueWarning {
+  color: var(--theme-status-warning);
+  margin: 0;
+}
```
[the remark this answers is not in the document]
keita-determined: If we remove `findIsNameUnique`, we dont need it anymore. I'll remove it

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
keita-determined: In this case, I think using Form is not appropriate because we handle modal with `getModalProps`

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
```diff
@@ -36,28 +36,40 @@ const useModalProjectCreate = ({ onClose, workspaceId }: Props): ModalHooks => {
...
+        });
+      } else {
+        handleError(e, {
+          level: ErrorLevel.Error,
+          publicMessage: 'Please try again later.',
+          publicSubject: 'Unable to create project.',
+          silent: false,
+          type: ErrorType.Server,
+        });
+      }
```
keita-determined, line 72: Better error message than before
It is a bit wordy but i think the new error message is more informative

### Before

image


### After

image

webui/react/src/hooks/useModal/Project/useModalProjectCreate.tsx
keita-determined: Added better error message  as a replacement of this https://github.com/determined-ai/determined/pull/4691#discussion_r936965850

hkang1 (approved):
Thanks for updating the PR. The code is looking good and the behavior of checking the unique name is great.

One tweak I would suggest:
If there already exists a duplicate name, rather than showing that error as a toast/notification and then dismissing the modal, IMO it would be more of a time saver to keep the user still on the modal and show them on the name field that their name is not unique. Otherwise the user has to reopen and type in a new name.

keita-determined:
> One tweak I would suggest:
> If there already exists a duplicate name, rather than showing that error as a toast/notification and then dismissing the modal, IMO it would be more of a time saver to keep the user still on the modal and show them on the name field that their name is not unique. Otherwise the user has to reopen and type in a new name.

I agree, but it looks a bit difficult due to `handleOk`. When `handleOk` is triggered, model is closed no matter what. I'll merge this PR and create a ticket for it.

hkang1:
> I agree, but it looks a bit difficult due to handleOk. When handleOk is triggered, model is closed no matter what. I'll merge this PR and create a ticket for it.

With `antd` modal, (check out `useModal`) patterns, if you throw an error, it should block the modal from getting dismissed.

PR #4693, opened 2022-08-01
fix: allocation count should be source of truth for trial run id [DET-7953]

stoksc:
## Description
Allocation count must be the source of truth for trial run id since trial run id cannot be written, as is, until after it would be required.

## Test Plan
- [x] run an experiment that will never schedule, restart the cluster. allocation shouldn't fail on restart.

## Commentary (optional)
I would just write the trial row first, but that impacts UX a lot, so I wanted to avoid making such a change so quickly.

Going to followup to remove trial run ID altogether shortly

## Checklist


stoksc:
gonna go for a better fix that will take longer, with a patch just in case for now

PR #4695, opened 2022-08-02
fix: make allocation saves idempotent [DET-7953]

stoksc:
## Description
We had a bug where, because allocations and trials are on different threads and the trial table rows are written lazily, trial run id and rows in the allocation table could get out of sync, which would cause us to attempt to rewrite allocation rows on restart that already existed. The change just makes the DB write a bit more lenient, for now.

## Test Plan
- [x] run an experiment that will never schedule, restart the cluster. allocation shouldn't fail on restart.

## Commentary (optional)
This is a little non-ideal but much simpler than the alternative for now. In a follow up, I'm going to probably remove trial run ID altogether and make trial table saves not lazy.

## Checklist

### DIFF
--- a/e2e_tests/tests/cluster/test_agent_restart.py
+++ b/e2e_tests/tests/cluster/test_agent_restart.py
@@ -1,6 +1,7 @@
 import json
 import subprocess
 import time
+import uuid
 from pathlib import Path
 from typing import Any, Dict, Iterator, Tuple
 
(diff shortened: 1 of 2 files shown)


e2e_tests/tests/cluster/test_agent_restart.py
```diff
@@ -309,3 +311,32 @@ def test_agent_reconnect_trigger_schedule(
...
+    past_marker = False
+    with open(DEVCLUSTER_MASTER_LOG_PATH) as lf:
+        for line in lf.readlines():
+            if not past_marker:
+                if log_marker in line:
+                    past_marker = True
+                continue
+
+            if err in line:
+                pytest.fail(f"allocation id save failure after restart: {line}")
```
ioga, line 342: nice

ioga (approved):

PR #4696, opened 2022-08-02
fix: Unable to fetch data after sorting or filtering items or filter

keita-determined:
## Description
[DET-7930]

## Test Plan
- Check if we can able to fetch data after clicking sort many times in few seconds (ideally click 10 times in a second and table should have at least 50 items) Check out Jira ticket for the reproducing video

## Commentary (optional)
We use polling every 5 sec so I removed fetching experiments when sorting or filtering are triggered because it causes the issue.
I tried using `throttle` and `debouce` but it's not working well due to useEffect. Is there any better approach besides removing the useEffect?

## Checklist

[DET-7930]: https://determinedai.atlassian.net/browse/DET-7930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ProjectDetails.tsx
+++ b/webui/react/src/pages/ProjectDetails.tsx
@@ -4,6 +4,7 @@ import type { MenuProps } from 'antd';
 import { FilterDropdownProps } from 'antd/lib/table/interface';
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useParams } from 'react-router-dom';
+import { debounce } from 'throttle-debounce';
 
 import Badge, { BadgeType } from 'components/Badge';
 import ExperimentActionDropdown from 'components/ExperimentActionDropdown';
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -757,26 +757,6 @@ const ProjectDetails: React.FC = () => {
...
-    settings.sortDesc,
-    settings.sortKey,
-    settings.state,
-    settings.tableLimit,
-    settings.tableOffset,
-    settings.user,
-  ]);
-
   useEffect(() => {
     return () => canceler.abort();
```
hkang1, line 761: I don't know this for sure, but this suspiciously seems like API calls are being made continuously but previous calls are not cancelled properly and allowed to linger until completion. So with browser HTTP calls, there is a rough limit of about 6-8 per host and once it hits that limit it puts all subsequent calls in Pending. So I **think** the problem is that the canceler needs to be created for each call appropriately.

keita-determined: That's a great perspective!

keita-determined:
this behavior is resolved, close!!

PR #4697, opened 2022-08-02
fix: word break in description

keita-determined:
## Description
I found that sentences in description are sometimes invisible when width is narrow and sentences are long.
Sentences should be always visible 

### Before

https://user-images.githubusercontent.com/109113616/182296754-9b0c0136-1909-4903-860d-3d43cc76d876.mov

### After

https://user-images.githubusercontent.com/109113616/182296648-db2b52db-377c-4efa-895b-829aceb29481.mov

## Test Plan
- Check if sentences are visible in any length and type of sentences and description width

## Commentary (optional)
Description in experiment detail has 500 letters limit but the one in experiment list doesnt. Should we add one?

## Checklist

### DIFF
--- a/webui/react/src/components/InlineEditor.module.scss
+++ b/webui/react/src/components/InlineEditor.module.scss
@@ -13,6 +13,7 @@
      */
     display: grid;
     position: relative;
+    word-break: break-word;
   }
   .growWrap::after {
     /* Extra space needed to prevent jumpy behavior. */


trentwatt:
can you add screen shots of experiment details and experiment list pages before and after?

> Description in experiment detail has 500 letters limit but the one in experiment list doesnt. Should we add one? 
sgtm.

keita-determined:
> can you add screen shots of experiment details and experiment list pages before and after?

I added them in description

trentwatt (approved):
Looks like this just affects when single word is really long?  looks good to me.

keita-determined:
> Looks like this just affects when single word is really long? looks good to me.

Yes. Someone might put a link in description, so i think we should handle it

PR #4698, opened 2022-08-02
fix: remove the default theme from initialization

thiagodallacqua-hpe:
## Description

Original ticket: DET-7101.

While reaching the login page with SaaS login, the login button had a UI flicker (passing from the default theme to the HPE theme). That was happening due to the fact that we load our initial state with the DeterminedAI branding set in the theme, then, after fetching the `info`, we switch the theme properly.

The initial state for the `branding` and `theme` has been changed as a proposed fix.

 SaaS login after the fix:

https://user-images.githubusercontent.com/104855841/182389179-70f22f54-cbbe-4d22-8e84-d4cfbcdff7d5.mp4

Default login after the fix:

https://user-images.githubusercontent.com/104855841/182389220-79b45c79-6261-4e16-9d3c-1d7a744b935a.mp4

## Test Plan

Use the SaaS login flow, you might have to set your server address to the `gcloud` env, then reach the login page.

The expected behavior is to not have any flickering in the UI.

## Checklist

### DIFF
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -5,7 +5,7 @@ import { DarkLight, Mode, Theme } from 'shared/themes';
 import { clone, isEqual } from 'shared/utils/data';
 import { percent } from 'shared/utils/number';
 import {
-  Agent, Auth, BrandingType, ClusterOverview, ClusterOverviewResource,
+  Agent, Auth, ClusterOverview, ClusterOverviewResource,
   DetailedUser, DeterminedInfo, PoolOverview, ResourcePool, ResourceType, Workspace,
 } from 'types';
 
@@ -138,7 +138,7 @@ const initClusterOverview: ClusterOverview = {
   [ResourceType.UNSPECIFIED]: clone(initResourceTally),
 };
 const initInfo = {
-  branding: '' as BrandingType,
+  branding: undefined,
   checked: false,
   clusterId: '',
   clusterName: '',
--- a/webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
+++ b/webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
@@ -94,7 +94,7 @@ const useModalCreateUser = ({ onClose }: ModalProps): ModalHooks => {
     openOrUpdate({
       closable: true,
       // passing a default brandind due to changes on the initial state
-      content: <ModalForm branding={info.branding || 'determined'} form={form} />,
+      content: <ModalForm branding={info.branding || BrandingType.Determined} form={form} />,
       icon: null,
       okText: 'Create User',
       onCancel: handleCancel,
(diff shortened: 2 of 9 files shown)


webui/react/src/contexts/Store.tsx
```diff
@@ -139,7 +138,7 @@ const initClusterOverview: ClusterOverview = {
   [ResourceType.UNSPECIFIED]: clone(initResourceTally),
 };
 const initInfo = {
-  branding: BrandingType.Determined,
+  branding: '' as BrandingType,
```
hkang1, line 141: **suggestion:** this makes me a little bit nervous because it's forcing TypeScript to acknowledge an empty string as a branding type. If there are areas in the code using this value and expecting HPE/Determined values but uses the empty string, things can fail poorly if uncaught.

My recommendation is to leave it as `BrandingType.Determined` and checking the `checked` field (may require some tweaking to ensure it is set to `true` properly). Then in areas referencing the branding type, before rendering, checking against that bit to make sure it is true before rendering the appropriate UI.

webui/react/src/hooks/useTheme.ts
```diff
@@ -71,7 +71,7 @@ export const useTheme = (): void => {
     if (!event.matches) setSystemMode(getSystemMode());
   }, []);
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
hkang1, line 74: nice! yeah this makes total sense to use this instead.

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -93,7 +93,8 @@ const useModalCreateUser = ({ onClose }: ModalProps): ModalHooks => {
   const modalOpen = useCallback(() => {
     openOrUpdate({
       closable: true,
-      content: <ModalForm branding={info.branding} form={form} />,
+      // passing a default brandind due to changes on the initial state
+      content: <ModalForm branding={info.branding || 'determined'} form={form} />,
```
hkang1, line 97: **suggestion:** use `BrandingType.Determined` instead of the string `determined`

webui/react/src/contexts/Store.tsx
hkang1: Another alternative approach would be to use `branding?: BrandingType;` (allow it to be `undefined` and checking that the value is set). The downside is that this might require a lot of changes across the codebase (basically anywhere branding is used)

webui/react/src/contexts/Store.tsx
```diff
@@ -155,7 +154,7 @@ const initUI = {
   omnibar: { isShowing: false },
   showChrome: true,
   showSpinner: false,
-  theme: themes[BrandingType.Determined][DarkLight.Light],
+  theme: {} as Theme,
```
hkang1, line 157: based on the 1st suggestion above, this would just remain as is (prior to this change)

hkang1 (commented):
The behavior looks great on the videos! Take a look at some alternative approaches to checking branding in the comments.

webui/react/src/contexts/Store.tsx
hkang1: does this cause any issues having this as an empty object? Similar question as the branding type as empty string.

hkang1 (approved):
Thanks for making the changes. I think code structure it should be good, but let's keep an eye on this for any weird artifacts that may pop up from the nature of things being "undefined" for a brief moment (until cluster responds), particularly the releases.

PR #4700, opened 2022-08-02
fix: WebUI workspace pagination [DET-7927]

gt2345:
## Description

Fix workspace table view pagination. Bug details here: https://determinedai.atlassian.net/browse/DET-7927

## Test Plan

Verify that interactive table pagination works as expected. (i.e. workspace list table, user list table)

## Checklist

### DIFF
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -363,17 +363,20 @@ const InteractiveTable: InteractiveTable = ({
 
   const handleChange = useCallback(
     (tablePagination, tableFilters, tableSorter): void => {
+
       if (Array.isArray(tableSorter)) return;
 
-      const { columnKey, order } = tableSorter as SorterResult<unknown>;
-      if (!columnKey || !settings.columns.find((col) => columnDefs[col]?.key === columnKey)) return;
-
-      const newSettings = {
-        sortDesc: order === 'descend',
-        sortKey: columnKey,
+      const newSettings: Partial<InteractiveTableSettings> = {
         tableLimit: tablePagination.pageSize,
         tableOffset: (tablePagination.current - 1) * tablePagination.pageSize,
       };
+
+      const { columnKey, order } = tableSorter as SorterResult<unknown>;
+      if (columnKey && settings.columns.find((col) => columnDefs[col]?.key === columnKey)) {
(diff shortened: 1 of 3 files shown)


webui/react/src/components/InteractiveTable.tsx
```diff
@@ -363,17 +363,20 @@ const InteractiveTable: InteractiveTable = ({
...
-      const newSettings = {
-        sortDesc: order === 'descend',
-        sortKey: columnKey,
+      const newSettings: Partial<InteractiveTableSettings> = {
         tableLimit: tablePagination.pageSize,
         tableOffset: (tablePagination.current - 1) * tablePagination.pageSize,
       };
+
+      const { columnKey, order } = tableSorter as SorterResult<unknown>;
+      if (columnKey && settings.columns.find((col) => columnDefs[col]?.key === columnKey)) {
```
trentwatt, line 375: `settings.columns` is the list of displayed columns, which might not include all columns.

so if we were sorting on a column, and then hide that column, this code would cause it to no longer sort by that column. 

this PR doesn't change the functionality, but it at least got me wondering it that is the right thing. what do you think?

trentwatt (approved):
LGTM. I did notice something. 

If there's exactly 10 workspaces and the page size is 10, you get an option to go to the next page, but then it's blank. 

I don't think it's directly related to this change because the same thing does not happen for experiments.

https://user-images.githubusercontent.com/28383080/182912389-6847ca97-36f4-4e7e-b987-7993095310de.mov

webui/react/src/pages/WorkspaceList.tsx
```diff
@@ -64,7 +64,7 @@ const WorkspaceList: React.FC = () => {
         sortBy: validateDetApiEnum(V1GetWorkspacesRequestSortBy, settings.sortKey),
         users: settings.user,
       }, { signal: canceler.signal });
-      setTotal(response.pagination.total ?? 0);
+      setTotal((response.pagination.total ?? 1) - 1); // -1 because we do not display immutable ws
```
trentwatt, line 67: 👍

webui/react/src/components/InteractiveTable.tsx
gt2345: I think what makes most sense to me is to keep sorted on selected column, even that column got hidden.

PR #4702, opened 2022-08-02
chore: test AMP autocast and gradient scaling [DET-7885]

drh-determined-ai:
## Description

Add automated tests of Automatic Mixed Precision and Gradient Scaling (AMP) in Determined in three different ways:
- Determined's API for internally using Apex,
- Determined's API for internally using PyTorch's native AMP,
- manually using PyTorch's native AMP in a user-defined training method.

## Test Plan

Run unit tests with GPU(s) and with `apex` installed.

## Commentary (optional)

Because these tests rely on GPU(s), **they are not currently being run by CircleCI**.

## Checklist


harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -601,17 +601,31 @@ def make_workloads() -> workload.Stream:
...
+            pytorch_onevar_model.OneVarAutoAMPTrial,
+            pytorch_onevar_model.OneVarManualAMPTrial,
+        ],
+        ids=[
+            "apex",
+            "autocast",
+            "manual",
+        ],
+    )
+    def test_amp(self, trial_class) -> None:
```
azhou-determined, line 619: do we assert / somewhere that the losses are what we expect after AMP/scaling? am i missing something?

drh-determined-ai: You're not missing anything. The code is. Stay tuned!

azhou-determined:
i can't comment on the file directly, but remember to remove harness/tests/experiment/fixtures/.pytorch_onevar_model.py.swp

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -182,33 +185,33 @@ def make_workloads_1() -> workload.Stream:
             latest_checkpoint = interceptor.metrics_result()["uuid"]
             steps_completed = trainer.get_steps_completed()
 
-        controller1 = utils.make_trial_controller_from_trial_implementation(
+        controller_1 = utils.make_trial_controller_from_trial_implementation(
```
azhou-determined, line 188: nitpick: not a huge fan of using numbers in variable names. i know you didn't make this change, it was already there, but maybe we can change the name to something more descriptive, like invalid_controller, invalid_, etc?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -1010,7 +1010,7 @@ def build_callbacks(self) -> Dict[str, pytorch.PyTorchCallback]:
         """
         return {}
 
-    def evaluate_batch(self, batch: pytorch.TorchData, batch_idx: int) -> Dict[str, Any]:
+    def evaluate_batch(self, batch: pytorch.TorchData) -> Dict[str, Any]:
```
rb-determined-ai: This is a public API, please don't make causal changes to it as part of other changes.

If there is a change you want to make to it, you probably should propose it in a public channel first.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -172,7 +180,8 @@ def evaluate_batch(self, batch: pytorch.TorchData) -> Dict[str, Any]:
         self.cls_reducer.update(sum(label), None)
         self.fn_reducer.update((sum(label), None))
 
-        loss = self.loss_fn(self.model(data), label)
+        output = self.model(data)
+        loss = self.loss_fn(output, label)
         return {"val_loss": loss}
```
rb-determined-ai, line 185: Either return "output" as a metric, as with `train_batch`, or leave this code alone.

In general, changes which have no functional benefit make it harder to understand the history of the codebase.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: I think there's still some assertions missing here.

The assert_output_float16 assertion seems good.

All the math is written in the top of pytorch_onevar_model.py to calculate what the exact expected gradient updates would be.  An example of what I have in mind is how `TestPyTorchTrial::test_onevar_single()` tests each gradient step.

Another good example is `e2e/tests/experiment/test_pytorch_distributed.py::test_gradient_aggregation()`, which also calculates all of the expected gradient updates and makes sure the actual updates are within computational error of the expected ones.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I totally agree with this nitpick. Done!

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -601,19 +605,36 @@ def make_workloads() -> workload.Stream:
...
+        ],
+    )
+    def test_amp(self, trial_class) -> None:
+        if trial_class is pytorch_onevar_model.OneVarApexAMPTrial and not HAVE_APEX:
+            pytest.skip("Apex not available")
 
-        def make_workloads() -> workload.Stream:
-            trainer = utils.TrainAndValidate()
-            yield from trainer.send(steps=1, validation_freq=1, scheduling_unit=1)
+        if trial_class is pytorch_onevar_model.OneVarApexAMPTrial:
```
rb-determined-ai, line 627: You can parameterize multiple values at once to avoid this awkward else/if'ing:

```
@parameterize(
    "class,output_float16,scale_changed",
    [(OneVarTrialWithApexAmp,False,True), ...],
)
```

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: You're absolutely right. This was an erroenous change I made while experimenting. I've reverted it already.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: We can definitely do this (and I have the feeling we might as well), but the reason I made the `assert_*` flags be directly derived from the trial class is because - when either/both are False - it's a choice made out of necessary, not, say, preference for typical use-cases. Ideally we'd have checks for casting and changes of scale for all three trial classes.

That might not be clear, but I think this is an ok change anyway. A plus is that what exactly is being checked in the three test variations is made clear when running them (without looking at the code).

drh-determined-ai: Actually, the plus I mentioned isn't quite true since we have the "apex", "autocast", and "manual" ids.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: I was returning `output` as a "metric" at one point, and this was leftover. I've reverted it as well.

drh-determined-ai:
Anyone know anything about why the `package-and-push-system-local` job in `test-e2e` is failing? I haven't encountered any problems with this check before, and I don't know what I could have changed to cause it...

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +654,36 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+            # If the scale changed, it's because an inf or NaN was encountered, so the
+            #  model couldn't have changed this iteration.
+            continue
+        assert newer["loss"] <= older["loss"]
+        assert newer["loss"].dtype is np.dtype("float32")
+        # Check the accuracy of the gradient change.
+        trial_class.check_batch_metrics(newer, idx, epsilon=1e-3)
+        if idx == 0:
+            trial_class.check_batch_metrics(older, idx, epsilon=1e-3)
+    assert scale_ever_decreased or scale_ever_increased
```
drh-determined-ai, line 689: While this code works and does what it's supposed to, it could be tightened up quite a bit.

drh-determined-ai:
Curiously, `TestPyTorchTrial::test_custom_eval` is failing on my GCP machine (but not on my laptop or with CircleCI). The assertion `original["loss"] == custom_eval["loss"]` is incorrect at the order of `1e-7`. 
We _could_ replace statements like this with `isclose` or similar calls, but without understanding the issue better I think that'd be very premature.
Since things are ok on CircleCI I'm ok with putting a pin in this. I wanted to make note of it now, though.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -221,13 +235,92 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
...
+        self.scaler = context.wrap_scaler(GradScaler())
+        super().__init__(context)
+
+    def train_batch(
+        self, batch: pytorch.TorchData, epoch_idx: int, batch_idx: int
+    ) -> Dict[str, torch.Tensor]:
+        data, label = batch
+
+        self.cls_reducer.update(sum(label), batch_idx)
+        self.fn_reducer.update((sum(label), batch_idx))
```
rb-determined-ai, line 280: the custom reducers here are from an orthogonal feature.  Does it work if you just don't update them, leaving them totally out of the subclass definition?

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -148,23 +155,30 @@ def train_batch(
...
     @staticmethod
-    def check_batch_metrics(metrics: Dict[str, Any], batch_idx: int) -> None:
+    def check_batch_metrics(
+        metrics: Dict[str, Any],
+        batch_idx: int,
+        metric_keyname_pairs: Iterable[Tuple[str, str]] = (
+            ("loss", "loss_exp"),
+            ("w_after", "w_exp"),
+        ),
+        **kwargs,
```
rb-determined-ai, line 169: Let's avoid using kwargs like this, because it hurts the readability of the test.  If I read the function prototype of check_batch_metrics, I don't even know what kwargs is meant to get passed along to.  If you called it `atol` instead of `**kwargs`, it would accomplish the same purpose, only it would be more obvious how it was meant to be used, from reading the function definition.

Using `kwargs` makes a lot of sense when you are wrapping a function transparently or something, where the args the function takes are opaque to the wrapper.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -64,6 +65,8 @@ def test_require_global_batch_size(self) -> None:
         utils.ensure_requires_global_batch_size(pytorch_onevar_model.OneVarTrial, self.hparams)
 
     def test_onevar_single(self) -> None:
+        """Assert that the training loss and validation error decrease monotonically."""
+
```
rb-determined-ai, line 69: This is a duplicate of what the comment says.

If you want to put this here, probably delete the comment.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: If you think there's a lot of tightening up to do, now is the time to do it.

I'd be interested to be a bit more precise about when the scale overflows or underflows.  Right now, which batch index is the underflow/overflow it happening on?  The OnesDataset seems to me like it's going to have a pretty consistent and manageable gradient.  I thought you might change the dataset to emit some particular record of 100000 or something to force it to overflow just when you wanted it to.

rb-determined-ai (commented):
Overall, looks really good!

> The thing I was missing is that the weights will not be changed during a batch iteration in which the scale changed (due to a NaN or inf), because the optimizer step is skipped

I don't see where this case you mentioned is being handled?

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -34,15 +34,21 @@
 TODO(DET-1597): migrate the all pytorch XOR trial unit tests to variations of the OneVarTrial.
 """
 
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import numpy as np
 import torch
 import yaml
+from torch.cuda.amp import GradScaler, autocast
```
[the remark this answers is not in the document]
drh-determined-ai: Thanks! i needed that reminder.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: You're absolutely right.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Yeah, doc-string > comment

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: Cool. Gone!

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -148,23 +155,30 @@ def train_batch(
...
 
     @staticmethod
-    def check_batch_metrics(metrics: Dict[str, Any], batch_idx: int) -> None:
+    def check_batch_metrics(
+        metrics: Dict[str, Any],
+        batch_idx: int,
+        metric_keyname_pairs: Iterable[Tuple[str, str]] = (
+            ("loss", "loss_exp"),
+            ("w_after", "w_exp"),
+        ),
```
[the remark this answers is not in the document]
drh-determined-ai: Done.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: The scaler factor is initiailly 2e16 by default, so the ones data is already "large", which causes the rescaling and the skipping of the optimizer. If you were to set, say, `init_scale=8` in the `GradScaler` constructor and rerun the manual test it would fail.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: This is done.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: I'd still like to see the test be a little more precise about what the exact losses are for every batch.

But perhaps, my request might need a bit of explanation as to what is coming down the line for our PyTorchTrial unit tests.

Eventually I want unit tests for (gradient aggregation + amp), (checkpointing + amp), and (dtrain + amp).  Those tests will need to be very precise about what exact losses and scale factors are expected at each batch index.  I'm thinking of a function like:

```python
def expected_loss_scale_weights(num_batches):
    out = []
    for i in range(num_batches):
        loss = ...
        scale = ...
        weight = ...
        out.append((loss, scale, weight))
    return out
```

In the current test, this is dangerously close to just testing AMP, but if you mix amp with dtrain, gradient aggregation, or checkpointing, where we have our hands all over various states, having an source of truth which is absolute (based on mathematical principles, not based on the previous output of the train_batch function) will mean that the one source of truth works for all of those tests.

Also, from a purely testing perspective, I trust test-against-ground-truth much more than test-by-inductive-reasoning.  I told the same thing to Maksim when he was writing the `test_gradient_aggregation` test.

azhou-determined: ^ i think we'll also need this to verify the path where a loss underflow/overflows for a particular batch and the scale changes but the loss doesn't step. not sure if you're planning on adding that in this PR, but we should probably check for that in a comprehensive AMP test.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,71 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+        loss = metrics["loss"].item()
+        scale_before = metrics["scale_before"]
+        scaled_loss = loss * scale_before
+        if "scaled_loss" in metrics:
+            assert scaled_loss == metrics["scaled_loss"]
+        growth_countdown -= 1
+        scale = metrics["scale"]
+        if scale != scale_before:
+            if scale < scale_before:
+                assert scaled_loss > 32758, (
```
azhou-determined, line 709: can we maybe make this magic number a variable? or at least add a comment so it's less magical?

drh-determined-ai: We definitely should, but it's exact meaning is mysterious to me. I found through some trial-and-error that - in the native AMP case (dunno about Apex) - this is the maximum value of the scaled loss for which the scaler is not reduced. Why it's this particular value I don't know. I had expected `65504/2 = 32752` (half the largest normal FP16 number) or `65536/2 = 32768`.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -92,6 +97,8 @@ def triangle_label_sum(updates: List) -> Any:
 
 
 class OneVarTrial(pytorch.PyTorchTrial):
+    _dataset_cls: Type[torch.utils.data.Dataset] = OnesDataset
+    _dataset_args = ()
```
rb-determined-ai, line 101: Let's avoid this sort of magic for the dataset.  The reason we should avoid it is because magic generally makes things harder to reason about.

Full disclosure, it was I who added the `_searcher_metric`.  It is intentionally breaking the Trial abstraction in a way that makes writing unit tests a lot easier, because sometimes the unit testing framework needs to know this information as part of testing.  It's a hack, but it's a useful hack.

The difference is that nobody outside of the Trial class needs to know what the `_dataset_cls` is.  It adds an extra layer of indirection without any real gain.

Why not just subclass the `build_training_data_loader()` instead?

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I was slightly wrong about the inequality, but I've confirmed that I have it correct now (and that it's the same for Apex), and I also gave a name to the magic number.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -221,13 +227,165 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
...
+        stage = self._get_stage(index)
+        return self._get_stage_item(stage)
+
+    def _get_stage_item(self, stage) -> Tuple:
+        for _stage, x in self.STAGES.items():
+            if _stage.lower() == stage.lower():
+                y = x
+                return torch.Tensor([float(x)]), torch.Tensor([float(y)])
+        else:
+            raise ValueError(f"Unrecognized {self.__class__.__name__} stage {stage}")
```
rb-determined-ai, line 258: This is a bit verbose.  I think since this is test code, you can make stronger assumptions about the behavior the calling code (which is all written by us) and you can make the result simpler to read and understand.

So if you wrote:

```python
def __getitem__(self, index: int) -> Tuple:
    n = self.STAGES[self.stages[index]]
    return torch.Tensor([x]), torch.Tensor([x])
```

so no need for `_get_stage()` and `_get_stage_item()` become unnecessary.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: I'm fine with what you suggest if I can ignore the code block inside `elif self.hparams["dataloader_type"] == "torch":`. If I can't, that's a lot of repeated code that I'd be sad about. ;'(

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: I totally agree. I want to say that there used to be a better reason for all this extra stuff before I made some other changes, but I dunno. Anyway, this is done!

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -221,13 +227,165 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
...
+            + 1 * ["zero"]
+            + 4 * ["one"]
+            + []
+        ),
+    )
+    _init_scale = None
+    _growth_interval = None
+
+    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
+        context._per_slot_batch_size, context._global_batch_size = 1, 1
```
azhou-determined, line 267: these could potentially be passed into the `make_trial_controller_from_trial_implementation` test method. would avoid setting technically protected variables on the context object. but this is just a test, so not a big deal.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
rb-determined-ai: Let me flip the question on you: can you find what code would use that, and tell me if it's safe to ignore?

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -525,7 +531,7 @@ def reducer_fn(_):
             return 1.0
 
         # Inject an unnamed metric which returns a non-dict (which is not allowed).
-        controller.context.wrap_reducer(reducer_fn)
+        controller.context.wrap_reducer(reducer_fn, name=None)
```
rb-determined-ai, line 534: this feels unnecessary

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -221,12 +227,152 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
...
+        w_exp = w_before + 2 * self.lr * data[0] * (label[0] - (data[0] * w_before))
+
+        with torch.cuda.amp.autocast():
+            output = self.model(data)
+            loss = self.loss_fn(output, label)
+
+        scaled_loss = self.scaler.scale(loss)
+        self.context.backward(scaled_loss)
+        self.context.step_optimizer(self.opt, scaler=self.scaler)
+        self.scaler.update()
```
rb-determined-ai, line 351: can you write the test without this?  It's technically not allowed in the PyTorchTrial API, and so the times we call scaler.update() are going to cause the control loop to run at the wrong rate.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+    yield from trainer.send(steps=20, validation_freq=1, scheduling_unit=1)
+    training_metrics, _ = trainer.result()
+
+    for metrics, next_metrics in zip(training_metrics[:-1], training_metrics[1:]):
+        scale = next_metrics["scale_before"]
+        if "scale" in metrics:
+            assert metrics["scale"] == scale
+        else:
+            metrics["scale"] = scale
+        assert "scale" in metrics
```
azhou-determined, line 685: this assert probably isn't necessary?

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+
+def make_amp_workloads(
+    trial_class,
+    assert_output_float16=False,
+) -> workload.Stream:
+    trainer = utils.TrainAndValidate()
+    yield from trainer.send(steps=20, validation_freq=1, scheduling_unit=1)
+    training_metrics, _ = trainer.result()
+
+    for metrics, next_metrics in zip(training_metrics[:-1], training_metrics[1:]):
```
azhou-determined, line 679: could this for loop be combined with the below one? where the below one could check for next index?

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+            assert metrics["scale"] == scale
+        else:
+            metrics["scale"] = scale
+        assert "scale" in metrics
+    loss_prev = None
+    GROWTH_INTERVAL = trial_class._growth_interval
+    growth_countdown = GROWTH_INTERVAL
+    # Only attempt assertions up to and including the penultimate batch, because
+    #  we may not have the updated scale from the final batch.
+    for idx, metrics in enumerate(training_metrics[:-1]):
```
azhou-determined, line 691: does it make sense/is there a good way to break up the checks inside this for loop into their own functions? like separate methods for checking dtype, scale, scale_loss, etc. it's fine as is, but would make it easier to read if there was some way to organize the logic.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
         assert isinstance(trial, torch.nn.Module), type(trial)
+
+
+def make_amp_workloads(
+    trial_class,
+    assert_output_float16=False,
+) -> workload.Stream:
+    trainer = utils.TrainAndValidate()
+    yield from trainer.send(steps=20, validation_freq=1, scheduling_unit=1)
+    training_metrics, _ = trainer.result()
```
rb-determined-ai, line 677: This is definitely a nit.

But it is possible to write this in a way that the assertions happen in the body of the test instead of in the workloads function.  It's not materially different, just slightly cleaner I think.  You would need to use the `nonlocal` to manually bind an external variable into the function definition so you can write to it inside the bound function and read from it after the bound function is complete.  It would look like this:

```
def test_amp():
    
    training_metrics = None

    def make_workloads() -> workload.Stream:
        trainer = utils.TrainAndValidate()
        yield from trainer.send(steps=20, scheduling_unit=1)
        
        nonlocal training_metrics
        training_metrics, _ = trainer.result()

    controller = ...
    controller.run()

    # by now training_metrics will be defined
    assert training_metrics

File: harness/tests/experiment/pytorch/test_pytorch_trial.py (hunk 19 above)
Comment (rb-determined-ai) [inline harness/tests/experiment/pytorch/test_pytorch_trial.py:679]: I think what you are doing here is the right strategy, but still it is tricky enough to warrant a comment explaining why the scale after each step is defined in the next_metrics.

File: harness/tests/experiment/fixtures/pytorch_onevar_model.py (hunk 17 above)
Comment (azhou-determined) [reply harness/tests/experiment/fixtures/pytorch_onevar_model.py:351]: i think this is the manual case, in which the auto_amp scaler.update() doesn't actually get called, so we do need this.

Comment (rb-determined-ai) [reply harness/tests/experiment/fixtures/pytorch_onevar_model.py:351]: oh... duh.  I missed the `or not self.context.experimental._auto_amp` in `_should_update_scaler`.  Thank you.

File: harness/tests/experiment/pytorch/test_pytorch_trial.py [hunk 22]
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+
+    for metrics, next_metrics in zip(training_metrics[:-1], training_metrics[1:]):
+        scale = next_metrics["scale_before"]
+        if "scale" in metrics:
```
rb-determined-ai, line 681: I think this is another case where it's safe to make stronger assumptions about the behavior of the code we're dealing with, since we are writing the tests.

Basically, just write the test it so that the metrics you want are always present in all cases, and skip the if statements and complex fallback logic.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+            # For the latter case, see the hook end_f16
+            #   defined in PyTorchTrialContext.autocast_forward_pass
+            assert metrics["output"].dtype is np.dtype("float32"), (
+                f"output has dtype {metrics['output'].dtype} but should be float32"
+            )
+        loss = metrics["loss"].item()
+        scale_before = metrics["scale_before"]
+        scaled_loss = loss * scale_before
+        if "scaled_loss" in metrics:
+            assert scaled_loss == metrics["scaled_loss"]
```
rb-determined-ai, line 708: this assertion strikes me as not testing any code path in PyTorchTrial, since the only time `scaled_loss` is set is in the manual case.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -636,3 +666,77 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+            else:
+                assert growth_countdown == 0, (
+                    f"scale grew from {scale_before} to {scale}, "
+                    f"but not when expected (growth_countdown={growth_countdown:d})"
+                )
+            growth_countdown = GROWTH_INTERVAL
+        else:
+            assert (
+                growth_countdown != 0
+            ), f"scale was expected to grow (growth_countdown={growth_countdown}) but did not"
```
rb-determined-ai, line 728: The logic here seems backwards from what I would expect.

I would expect something logic to flow from what is known to what is expected:
```python
if growth_countdown == 0:
    assert scale > scale_before
else:
    if scaled_loss >= MIN_SCALED_LOSS_TO_REDUCE_SCALE:
        assert scale < scale_before
    else:
        ...
```

(not saying that my pseudo code is what you should write, but that it looks at known quantities to figure out what assertions to make, rather than looking at results and making assertions about known quantities.)

To me, that's part of the pattern of "calculate a ground truth from first principles and compare results to the ground truth".

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -601,19 +611,39 @@ def make_workloads() -> workload.Stream:
...
-
-        def make_workloads() -> workload.Stream:
-            trainer = utils.TrainAndValidate()
-            yield from trainer.send(steps=1, validation_freq=1, scheduling_unit=1)
+    @pytest.mark.parametrize(
+        "trial_class,assert_output_float16",
+        [
+            (pytorch_onevar_model.OneVarApexAMPTrial, False),
+            (pytorch_onevar_model.OneVarAutoAMPTrial, False),
+            (pytorch_onevar_model.OneVarManualAMPTrial, True),
```
rb-determined-ai, line 621: since the only case that produces `assert_output_float16` seems to be the manual case, it seems like that assertion isn't very meaningful.  You could simplify the test by removing that assertion with no loss of test coverage on the product I think.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Yep, removed in 7e6ac53.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Yep, this was leftover from some debugging. Removed in 7e6ac53.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: You're right. I removed this and other related code in 7e6ac53.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Fine with me! Removed in 7e6ac53.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I think this is as strong as it can be. Is the `"scale" in metrics` statement what's bothering you?

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I would love to do this, but we should do it for all `PyTorchTrial` tests. If it's not too much of an abuse of the task scope, I'd be happy to do it now, but I don't want to make that call on my own. We can always make a new task, etc., but this would be a pretty quick change.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Done in 73a6008.

drh-determined-ai: Done in 73a6008.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Good call! This is done.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I think we were right on the edge of whether or not it'd be worth making more methods in order to reduce the body of another. After combining the loops as you suggested (73a6008) and reversing the assertion logic (8004a18), I think this is less of a concern (still iffy, though).

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: Good call. This is done.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: Methinks so: the custom (i.e. non-Determined) dataloader is only used in `test_require_disable_dataset_reproducibility` and `test_custom_dataloader`.
I'm gonna go ahead with your suggestion.

drh-determined-ai:
I want to highlight the use of Apex's [optimization level](https://nvidia.github.io/apex/amp.html#opt-levels-and-properties) `O2`. Although `O1` is recommended for typical use, it is not appropriate in these kinds of tests because it patches `torch` functions, which screws up later tests. Thank goodness for version control!

I also want to remind us that these (and other) tests that depend on GPU(s) are **not being run** by CircleCI. I let the bug just mentioned slip by for so long because I had forgotten this!

azhou-determined (approved):
lgtm! nice job, i think this is a needed test to add to our test suite.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: yes, why would "scale" not be in metrics?

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: no need to grow the scope of this change beyond the code you are writing.

But the question of asserts inside the workloads function is distinct from having `make_amp_workloads` as a standalone function (defined outside the scope of `test_amp()`).  This _is_ different than all the other tests in `test_pytorch_trial.py`.

harness/determined/pytorch/_experimental.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, parent: Any) -> None:
         self._data_repro_checks_disabled = False
         self._auto_to_device = True
 
-    def use_amp(self) -> None:
+    def use_amp(self, gradscaler_kwargs: Optional[Dict] = None) -> None:
```
rb-determined-ai, line 22: I'm sorry I missed this earlier.

I don't think this is an acceptable change to the public API, because `use_amp()` is supposed to be the fully automatic API, so adding kwargs to it is in opposition of the goals of the API.

Instead, for the purposes of your test, I would say you should just hack it, like this:
```python
self.context.experimental.use_amp()
# HACK: overwrite the scaler with a manually configured one, which
# is not something we don't actually allow with the use_amp() API.
self.context._scaler = amp.GradScaler(...)
self.scaler = self.context._scaler
```

rb-determined-ai: Also, in general, if you are going to make any change at all to the public api, please advertise it loudly.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: If it's unknown, which it is in the Apex case? We could write `metrics["scale"] = None` I suppose. Then of course the logic statement becomes `metrics["scale"] is None`. 

I see the advantage I think you have mind: that way, `metrics` has the same set of keys no matter what.

The disadvantage I've had in mind depends on how strict one wants to be about the (ab)use of `None`. The scale isn't nothing, it's just not known by the code that creates the `metrics` object.

I'm willing to skirt the semantics debate and go with your suggestion (assuming I understand you correctly). ;-p

harness/determined/pytorch/_experimental.py
drh-determined-ai: Thanks for catching this! This is done.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: How about we revisit this possibility when I get to [DET-1597](https://determinedai.atlassian.net/browse/DET-1597)?

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Oops, writing `metrics["scale"] = None` in `_train_batch` causes a `ZeroDivisionError` on the line `metrics = det.util.make_metrics(num_inputs, per_batch_metrics)` in `PyTorchTrialController._train_for_step`. Numpy is trying to take the mean of an empty list. I'm not sure why this causes an error instead of just giving `nan`. :-(

rb-determined-ai: > If it's unknown, which it is in the Apex case

Oh, I didn't realize that.  If it's not known in the apex case, then I'm happy with what you had.

drh-determined-ai: In defense of the offended code, bunching the scale and stage into `metrics` _is_ an abuse, since those aren't metrics per se but are rather state variables.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: This was a nit-pick, and not a big deal.  Land it how you like...

> How about we revisit this possibility

... but let's not land something we plan on changing right away.  I'm fine leaving it as-written indefinitely.  If you are ok with that, let's plan on that.

rb-determined-ai (approved):
Great work!

This is a very technical test, and I think you did a good job implementing it.

PR #4705, opened 2022-08-02
refactor: solidify rm interface [DET-7852, DET-7984]

stoksc:
## Description
The purpose of this change is to get the RM a proper interface. Previously the interface to
the resource manager was an assortment of messages, scattered over a few packages,
handled in the receive block of the resource managers. The lack of strictness of
implementation of such an interface has lead to some bugs, made it difficult to reason
about or design feature parity, and inherits all the issues with actors. This change adds a
plain Go interface as the resource manager interface, and for most messages shims the
implementation to the actor implementation. The path forward after this change is roughly
* separate the rms into their own packages
* _maybe_ move to a package level singleton rm, maybe make it more plugin-y
* shrink the interface where possible, it is non-orthogonal in a lot of places and is a
   ridiculously massive interface, clarify things that should be feature flagged and arent
   (so they just don't work well)
* then the plan is for the resource manager interface to become just a manager of pools,
   and to add an interface like this RM interface for pools, pushing all that is pool-related
   down into that layer. this should set us up nicely for resource manager multi-tenancy or
   whatever that project is.
* probably evaluate more simplifications after splitting out pool.
* then we can start to de-actorify this interface if it seems worth it

## Test Plan
- [x] run existing tests, if I write a bug I'll write a regression test. also a lot of manual testing.

## Commentary (optional)
🤷 (sorry a bunch of this is just renames from circular imports)

## Checklist


NicholasBlaskey (approved):

stoksc:
ran the full test suite here: https://github.com/determined-ai/determined/runs/7664214039

ioga (approved):

PR #4706, opened 2022-08-02
test: test cases for `shared/utils/routes.ts` [DET-7902]

keita-determined:
## Description
[DET-7902]

## Test Plan
- Check if test cases are valid 

## Commentary (optional)
I change `isFullPath` and `isAbsolutePath` for more strict and accurate validation

## Checklist

[DET-7902]: https://determinedai.atlassian.net/browse/DET-7902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/shared/utils/routes.ts
```diff
@@ -4,9 +4,18 @@ import history from '../routes/history';
 
 import { clone } from './data';
 
-export const isFullPath = (url: string): boolean => url.startsWith('http');
+export const isFullPath = (url: string): boolean => {
+  try {
+    return url.startsWith('http') && Boolean(new URL(url));
```
hkang1, line 9: **nit suggestion:** could also use the double bang instead of `Boolean()`.
```
&& !!(new URL(url));
```

webui/react/src/shared/utils/routes.test.ts
```diff
@@ -0,0 +1,363 @@
...
+      writable: true,
+    },
+  );
+  const location: Location = window.location;
+  return { location };
+};
+
+beforeEach(() => {
+  setup();
+});
```
hkang1, line 38: **question:** should there be a `afterEach()` that clears the `window.location` property? I'm guessing it normally is `undefined` in this jest test environment, but would it persist outside of `routes.test.ts` (aka other tests?)

webui/react/src/shared/utils/routes.test.ts
```diff
@@ -0,0 +1,363 @@
+import React from 'react';
+
+import { paths } from 'routes/utils';
+
+import history from '../routes/history';
```
hkang1, line 5: **nit:** `from 'shared/routes/history';` ?

hkang1 (approved):
Beautiful unit tests! Thanks for taking this on.

webui/react/src/shared/utils/routes.test.ts
keita-determined: I'm not sure... `setup()` function create `window.location` in `routes.test.ts`. I thought I don't need `afterEach` because `beforeEach` initialize it every time. Let me investigate.

trentwatt: i am 99% sure changes to `window` don't persist.

keita-determined: Hm I'll add `afterEach` then

webui/react/src/shared/utils/routes.test.ts
```diff
@@ -36,6 +38,11 @@ beforeEach(() => {
   setup();
 });
 
+afterEach(() => {
+  // eslint-disable-next-line no-global-assign
+  window = originWindow;
+});
+
```
keita-determined, line 45: @trentwatt Do you think its good?

PR #4711, opened 2022-08-03
chore: share sort utility [DET-7970]

hamidzr:
## Commentary (optional)

most of the changes here are generated/made by ide or our linter. I'd suggest reviewing by looking at the commits

## Checklist


trentwatt:
What do you think of moving remaining things in `src/utils/sort.ts` to other  files like e.g. like moving `metricNameSorter` to `utils/metric.ts` and getting rid of `src/utils/sort.ts`?

That way we dont have two `sort.ts`?

hamidzr:
> What do you think of moving remaining things in `src/utils/sort.ts` to other files like e.g. like moving `metricNameSorter` to `utils/metric.ts` and getting rid of `src/utils/sort.ts`?
> 
> That way we dont have two `sort.ts`?

hmm I don't have a strong preference, IMO the original motivation behind putting these in a utility module still holds. the point I see is that not moving these out make the job of sharing pieces easier and avoids having to think about and discuss where things go.

hamidzr:
@thiagodallacqua-hpe  is out today and maybe tomorrow? I'll unassign him

trentwatt:
> IMO the original motivation behind putting these in a utility module still holds

what was the original motivation?

seems like `utils/metric.ts` is still a utility module? and would be more appropriate for `metric`-specific code like `metricNameSorter`? likewise for states, etc.

>avoids having to think about and discuss where things go.
to me having `utils/sort.ts` and `shared/utils/sort.ts` is the main thing that would make me have to think about where things go.

what about just moving the entire `sort.ts` to shared, even if if there are parts you wouldn't use?


My thinking overall is that if things are generic, they can go in a generically-named module in shared. 

and if they are not shareable because they are particular to a specific aspect of core, then they should go in a module pertaining to that specific aspect.

hamidzr:
> what about just moving the entire `sort.ts` to shared, even if if there are parts you wouldn't use?

that'd be nice but the dependencies missing from the shared is why I'm avoiding that. the remaining sorters are specific to core.


good call. I  moved the sorters around.

trentwatt (approved):
LGTM.

Thanks for making the change 🫶

PR #4712, opened 2022-08-03
fix: DET-7973 client progress

mpkouznetsov:
## Description

After master implemented progress reporting by creating a trialProgress event, client needs to catch up by adding an event handler and returning a pseudo-operation Progress.

Note: I will rebase on custom-searcher after Nikita's branch is merged.

## Checklist


harness/determined/searcher/search_method.py
```diff
@@ -3,15 +3,25 @@
...
 
 from determined.common.api import bindings
 from determined.common.experimental import Checkpoint
 
 
 @dataclasses.dataclass
-class SearchState:
-    checkpoint_id: Optional[str]
+class SearcherState:
+    failures: Set[uuid.UUID]
```
liamcli, line 14: I looked at the master side state and saw we have Exits, Failures, and TrialsClosed which have the following relationship:
- All requests passed to TrialExitedEarly event will be stored in Exits but only those that Errored will be stored in Failures.
- We expect all Create operations to have a corresponding Close operation so the ShutDown decision is dependent on the number of TrialsCreated being the same as TrialsClosed.  Therefore, we expect a Close operation to be returned for TrialExitedEarly events.

I agree with excluding Exits client side since it seems a little redundant with TrialsClosed except Exits will be updated before TrialsClosed.

liamcli: We should double check and also make sure to document for the user that a Close operation needs to be created for InvalidHP early exits but not for others types of early exits since they correspond to actor states/events for which the experiment actor would auto close.

harness/determined/searcher/search_runner.py
```diff
@@ -98,6 +114,13 @@ def run(
                         )
                         experiment_is_active = False
                         break
+                    elif e.trialProgress:
```
liamcli, line 117: looksl ike we need to have the trial_progress state updated when handling a TrialProgress event

mpkouznetsov: 👍

harness/determined/searcher/search_method.py
mpkouznetsov: 1. Do you suggest I actually change anything here?
2. We probably have to write a little guide on implementing search methods with an example, like our Random implementation.

mpkouznetsov: Actually, @liamcli , looking at, for example, the random search implementation, we return a triplet of Create/ValidateAfter/Close for trials that exited early due to InvalidHP, and do nothing for the other trials.

aciborowska (approved):

PR #4713, opened 2022-08-03
fix: resolve issues around `InlineEditor` [DET-7914]

keita-determined:
## Description
- [DET-7914]
-  Cannot ignore keydown event until the IME is confirmed
- Description in uncategorized table has no limit but the one in experiment detail has 500 letters limit
- `isDarkOn` is deprecated in `InlineEditor` 

## Test Plan
- Check if `InlineEditor` works as extected
- Check if description in uncategorized table has 500 letters limit
- Check if display name in settings has 32 letters limit
- Check if display name is ellipt when its too long
- Check if IME ignores keydown event until the IME is confirmed

## Checklist

[DET-7914]: https://determinedai.atlassian.net/browse/DET-7914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/InlineEditor.test.tsx
+++ b/webui/react/src/components/InlineEditor.test.tsx
@@ -1,4 +1,5 @@
 import {
+  fireEvent,
   render,
   screen,
   waitForElementToBeRemoved,
@@ -8,13 +9,12 @@ import React from 'react';
 
 import InlineEditor from './InlineEditor';
 
-const user = userEvent.setup();
-
 const setup = ({ disabled, onSaveReturnsError, value } = {
   disabled: false,
   onSaveReturnsError: false,
   value: 'before',
 }) => {
+  const user = userEvent.setup();
   const onSave = onSaveReturnsError
     ? jest.fn(() => Promise.resolve(new Error()))
     : jest.fn(() => Promise.resolve());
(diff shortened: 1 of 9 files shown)


webui/react/src/components/InlineEditor.test.tsx
```diff
@@ -89,9 +89,28 @@ describe('InlineEditor', () => {
...
+    const textbox = screen.getByRole('textbox');
+    await user.click(textbox);
+    await user.clear(textbox);
+    await user.type(textbox, 'こんにちは');
+    fireEvent.keyDown(textbox, {
+      charCode: 229,
+      code: 'Enter',
+      key: 'Enter',
+      keyCode: 229,
+    });
```
keita-determined, line 109: I dont wanna use `fireEvent` here but seems like userEvent doesnt have option to specify keyCode

webui/react/src/components/InlineEditor.tsx
```diff
@@ -114,20 +111,21 @@ const InlineEditor: React.FC<Props> = ({
       e.preventDefault();
       return;
     }
-    if (e.code === CODE_ENTER) {
-      if (!allowNewline || !e.shiftKey) e.preventDefault();
+    if (e.which === 229) {
+      // Ignore keydown event until IME is confirmed like Japanese and Chinese
+      return;
+    }
```
hkang1, line 117: ah interesting, learned something new today!

webui/react/src/components/InlineEditor.test.tsx
hkang1: yeah, totally get it. I have encountered something like this where fireEvent still had to be used

hkang1 (approved):
Thanks for catching this issue and fixing it! Not to mention testing it via IME keyboard too.

PR #4716, opened 2022-08-04
chore: add regex in `InlineEditor` [DET-7518]

keita-determined:
## Description
[DET-7518]

## Screenshot

https://user-images.githubusercontent.com/109113616/182768603-4b33765a-742e-45dd-a44f-5c92fffb384c.mov

## Test Plan
- Check if InlineEditor works correctly
- Check if display name editor works correctly 
  - should shake when its empty or non alphanumeric is used, and previous value should be set
  - should save data when input is alphanumeric

## Commentary (optional)
Jira ticket says 
> When a user edits their displayName to be empty, it should get set to null

but decided that empty value is invalid

## Checklist

[DET-7518]: https://determinedai.atlassian.net/browse/DET-7518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/components/InlineEditor.module.scss
```diff
@@ -111,3 +111,26 @@
 .base.disabled .growWrap:hover > textarea {
   border-color: transparent;
 }
+.shakeAnimation {
```
hkang1, line 114: ha! nice little animation, love it

webui/react/src/components/InlineEditor.module.scss
```diff
@@ -111,3 +111,26 @@
...
+  animation: shake 0.8s cubic-bezier(0.36, 0.07, 0.2, 0.95) both;
```
hkang1, line 115: **question:** would it be possible (and do you think it'll be a good idea) to change the border to `var(--theme-status-critical)` as it's shaking and back to the regular border color to show that it's a problem?

webui/react/src/pages/Settings/SettingsAccount.tsx
```diff
@@ -57,6 +57,7 @@ const SettingsAccount: React.FC = () => {
       <div className={css.row}>
         <label>Display Name</label>
         <InlineEditor
+          pattern={new RegExp('^[a-zA-Z]+[a-zA-Z0-9]*$')}
```
hkang1, line 60: idk if there is a significant performance difference, but could also be written as:
```
new RegExp('^[a-z]+[a-z0-9]*$', 'i')
```
OR (this one is succinct but may not be as typescript friendly? not sure)
```
/^[a-z]+[a-z0-9]*$/i
```

hkang1 (approved):
This looks great! I added some minor suggestions and comments, but nothing blocking.

webui/react/src/pages/Settings/SettingsAccount.tsx
keita-determined: That's a good suggestion. Enabling insensitivity improves the performance a lot in my experience

webui/react/src/components/InlineEditor.module.scss
keita-determined: I think its possible and a good idea. My concern is that if the entered value is invalid, the previous value is set immediately. Ideally setting the previous value after shaking would be better with the `critical border`. Let me try

keita-determined: I fixed it. It was actually not hard to do

https://user-images.githubusercontent.com/109113616/183133976-92a3f6bc-c25d-485d-b5e1-c519040ba77c.mov

webui/react/src/pages/Settings/SettingsAccount.tsx
```diff
@@ -58,6 +58,7 @@ const SettingsAccount: React.FC = () => {
         <label>Display Name</label>
         <InlineEditor
           maxLength={32}
+          pattern={new RegExp('^[a-z][a-z0-9\\s]*$', 'i')}
```
keita-determined, line 61: Forgot to add whitespace, so added

PR #4719, opened 2022-08-04
fix: record operations at the right places around shutting down

dzhu:
## Description

The `s.Record` call was previously moved after the `if` block so that
the `Shutdown` operation was correctly recorded. However, this meant
that, running, e.g., random search with `max_concurrent_trials = 1`
would shut down after one trial, since the request to create the next
trial would not be counted in the check. So we simply have to record the
other operations first and then the `Shutdown` separately.

## Test Plan

- [x] test manually
- [x] add test

### DIFF
--- a/master/pkg/searcher/random_test.go
+++ b/master/pkg/searcher/random_test.go
@@ -103,3 +103,18 @@ func TestSingleSearchMethod(t *testing.T) {
 
 	runValueSimulationTestCases(t, testCases)
 }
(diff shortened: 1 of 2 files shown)


stoksc (approved):
lgtm

mpkouznetsov:
This fix actually addressed this issue: https://determinedai.atlassian.net/browse/DET-6735

PR #4721, opened 2022-08-04
refactor: authz interface for projects and workspaces [DET-8002]

NicholasBlaskey:
## Description

Implements the new authz pattern for workspace and projects. The idea behind the interface is to make it easier to support different authorization schemes across Determined.

## Test Plan

existing e2e tests + added integration tests

## Checklist


master/internal/api_experiment.go
```diff
@@ -1576,9 +1576,13 @@ func (a *apiServer) GetModelDef(
...
 func (a *apiServer) MoveExperiment(
-	_ context.Context, req *apiv1.MoveExperimentRequest,
+	ctx context.Context, req *apiv1.MoveExperimentRequest,
 ) (*apiv1.MoveExperimentResponse, error) {
-	p, err := a.GetProjectByID(req.DestinationProjectId)
+	user, err := a.CurrentUser(ctx, &apiv1.CurrentUserRequest{})
+	if err != nil {
+		return nil, err
+	}
+	p, err := a.GetProjectByID(req.DestinationProjectId, model.UserFromProto(user.User))
```
mapmeld, line 1585: up to you, but you could get a model.User with `grpcutil.GetUser` instead of converting response from `a.CurrentUser`

master/internal/api_workspace.go
```diff
@@ -12,28 +12,58 @@ import (
...
+	if err := a.m.db.QueryProto("get_workspace", w, id, curUser.ID); errors.Is(err, db.ErrNotFound) {
+		return nil, notFoundErr
+	} else if err != nil {
+		return nil, errors.Wrapf(err, "error fetching workspace (%d) from database", id)
+	}
+
+	if ok, err := workspace.AuthZProvider.Get().CanGetWorkspace(curUser, w); err != nil {
+		return nil, err
+	} else if !ok {
+		return nil, notFoundErr
```
mapmeld, line 36: I'm on the fence with this (and CanGetProject) about whether we should specifically acknowledge it's an auth/access issue and not a not-found error. You made a good point earlier that they are sequential so it is not an issue with someone discovering a valid link.

master/internal/api_workspace.go
```diff
@@ -100,6 +129,16 @@ func (a *apiServer) GetWorkspaceProjects(ctx context.Context,
 		return nil, err
 	}
 
+	user, err := a.CurrentUser(ctx, &apiv1.CurrentUserRequest{})
+	if err != nil {
+		return nil, err
+	}
+	resp.Projects, err = workspace.AuthZProvider.Get().
```
mapmeld, line 136: as currently implemented, `FilterWorkspaceProjects` is letting everything through, no?  We should have a step before this which checks whether the user is allowed to read the Workspace.

master/internal/api_workspace.go
```diff
@@ -185,13 +237,8 @@ func (a *apiServer) PostWorkspace(
...
 ) (*apiv1.PatchWorkspaceResponse, error) {
-	user, err := a.CurrentUser(ctx, &apiv1.CurrentUserRequest{})
-	if err != nil {
-		return nil, err
-	}
-
-	// Verify current workspace exists and can be edited.
-	currWorkspace, err := a.GetWorkspaceByID(req.Id, user.User.Id, true)
+	currWorkspace, currUser, err := a.getWorkspaceAndCheckCanDoActions(ctx, req.Id, true,
+		workspace.AuthZProvider.Get().CanSetWorkspacesName)
```
mapmeld, line 241: in case we add more fields in the future, I'd prefer this to be more like the `PatchProject` function and check access inside of the `if req.Workspace.Name != nil ... { ... }` block

master/internal/api_experiment.go
```diff
@@ -1576,9 +1576,13 @@ func (a *apiServer) GetModelDef(
 }
 
 func (a *apiServer) MoveExperiment(
-	_ context.Context, req *apiv1.MoveExperimentRequest,
+	ctx context.Context, req *apiv1.MoveExperimentRequest,
```
mapmeld, line 1579: **MoveExperiment**
- MoveExperiment should require access (to edit? to specifically move experiments?) in the Source Project and Destination Project. 

**GetExperiments**
- If someone makes a GetExperimentsRequest with a project ID, we could check for project-read access

Future **GetExperiments** / likely notes for a future ticket:
- If someone makes a GetExperimentsRequest without a project ID, we should filter the experiments based on experiment-read / project-read access
- As a Paginated response, we are going to have some new ?s in how we return the count of access-filtered Experiments, page 2, etc. If access is down to the project level then we could maybe first get all project IDs where they have access.

master/internal/api_workspace.go
NicholasBlaskey: The argument for keeping it not found would be consistency and habits in the code base so we don't expose information in cases where it matters more (non sequential IDs).

If we can get to full isolation where a user never sees a project that they don't have access to then a user shouldn't run into it very often? 

I suppose a user could send another user a link to a project like ```127.0.0.1:8080/det/projects/5``` and it might be somewhat surprising to see not found instead of not having access.  

I don't know I'm not really sure which is better to be honest.

master/internal/api_workspace.go
NicholasBlaskey: Yep that is a good point added a check since it doesn't make sense to view workspace projects if you can't view the workspace.

master/internal/api_experiment.go
NicholasBlaskey: The plan is to tackle ```MoveExperiment``` and ```GetExperiments``` in the next PR which is applying this same interface to experiments if that is okay. 

Those are interesting points for the future. Counting might be especially tough to do in a performant way.

master/pkg/model/user.go
```diff
@@ -62,6 +64,18 @@ func (u FullUser) ToUser() User {
 	}
 }
 
+// UserFromProto converts a proto user to a model user.
```
mapmeld, line 67: you're not using `UserFromProto` anymore, so it can be cut

master/internal/api_workspace.go
```diff
@@ -159,23 +202,39 @@ func (a *apiServer) GetWorkspaces(
 	if err != nil {
 		return nil, err
 	}
+
+	curUser, _, err := grpcutil.GetUser(ctx, a.m.db, &a.m.config.InternalConfig.ExternalSessions)
```
mapmeld, line 206: since `GetWorkspaces` already gets the current user for the get_workspaces query earlier (for pinned workspaces), you could move this `grpcutil.GetUser` up earlier , or keep and use your new `UserFromProto` function

mapmeld (approved):

harness/determined/common/api/request.py
```diff
@@ -142,7 +142,12 @@ def do_request(
 
     def _get_message(r: requests.models.Response) -> str:
```
ioga, line 143: `_get_error_message` or `_get_error_str` would probably be more appropriate name for this func

master/internal/project/authz_basic_impl.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// ProjectAuthZBasic is classic OSS Determined authentication for projects.
+type ProjectAuthZBasic struct{}
+
+// CanGetProject always return true and a nil error for basic auth.
+func (a *ProjectAuthZBasic) CanGetProject(
+	curUser model.User, project *projectv1.Project,
+) (canGetProject bool, serverError error) {
+	return true, nil
```
ioga, line 22: why does this return a pair of bool and error, and not just an error like the methods after?

master/internal/workspace/authz_basic_impl.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// WorkspaceAuthZBasic is classic OSS Determined authentication for workspaces.
+type WorkspaceAuthZBasic struct{}
+
+// CanGetWorkspace always return true and a nil error.
+func (a *WorkspaceAuthZBasic) CanGetWorkspace(
+	curUser model.User, workspace *workspacev1.Workspace,
+) (canGetWorkspace bool, serverError error) {
+	return true, nil
```
ioga, line 18: same as previous q - why return a pair

ioga (approved):
looks good, a few minor points

master/internal/api_project_intg_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/determined-ai/determined/master/internal/mocks"
+	"github.com/determined-ai/determined/master/internal/project"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/proto/pkg/apiv1"
+	"github.com/determined-ai/determined/proto/pkg/projectv1"
+)
+
+var projectAuthZ *mocks.ProjectAuthZ
```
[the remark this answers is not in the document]
NicholasBlaskey: I don't see much value in testing at the basic auth interface like ```CanGetProject``` methods since they should be mostly covered by already existing e2e tests.

I do think that will change for RBAC in EE.

I personally feel that the integration tests with this mock interface is useful to make sure that the interface is being called correctly.

master/internal/project/authz_basic_impl.go
NicholasBlaskey: The pair is to allow us to prevent information leakage (the existence of a resource). The error is for some db error that should be bubbled up to the user and the boolean is to indicate that the user should not be able to view the resource. 

Whether this should be the case, ties into the question of system clarity vs security which I will raise a discussion about.

PR #4740, opened 2022-08-08
fix: hardcode pathname instead of using `paths`

keita-determined:
## Description
- fix typo of test

## Test Plan
- All tests should pass

## Checklist

### DIFF
--- a/webui/react/src/shared/utils/routes.test.ts
+++ b/webui/react/src/shared/utils/routes.test.ts
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import { paths } from 'routes/utils';
 import history from 'shared/routes/history';
 
 import type { AnyMouseEvent } from './routes';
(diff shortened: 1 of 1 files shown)


webui/react/src/shared/utils/routes.test.ts
```diff
@@ -310,7 +309,7 @@ describe('Routes Utilities', () => {
 
     it('should route to react URL', () => {
       expect(history.push).not.toHaveBeenCalled();
-      routes.routeToReactUrl(paths.cluster());
+      routes.routeToReactUrl('/clusters');
```
hamidzr, line 312: suggestion: ensure this paths and what you test for (a few lines below) don't go out of sync using a const var

hamidzr (approved):

hamidzr:
sharing these here ftr 
some thoughts/options I can think of:

- this wouldn't be a shared test
- avoid importing paths => hardcode /clusters. IMO this is better compared to useing `paths`
- test in a way that it doesn't assume certain pages exist in all the downstream apps
- officially share a limited subset of paths  and use that

PR #4745, opened 2022-08-09
fix: python sdk can parse master output again [DET-8021]

rb-determined-ai:
## Description

The GET experiments endpoint changed, but the python sdk did not.

Follow-up work includes migrating all of the python sdk over to
use the generated bindings, but that was no way going to happen
in time for the 0.19.1 release.

### DIFF
--- a/e2e_tests/tests/test_sdk.py
+++ b/e2e_tests/tests/test_sdk.py
@@ -1,0 +1,168 @@
+import random
+import time
+
+import pytest
+
+from determined.common import yaml
+from determined.common.api import bindings, errors
+from determined.experimental import client as _client
+from tests import config as conf
+
+
+@pytest.fixture(scope="session")
+def client() -> _client.Determined:
+    """
+    Reduce logins by having one session-level fixture do the login.
+    """
+    return _client.Determined(conf.make_master_url())
+
+
+@pytest.mark.e2e_cpu
+def test_completed_experiment_and_checkpoint_lookups(client: _client.Determined) -> None:
+    with open(conf.fixtures_path("no_op/single-one-short-step.yaml")) as f:
+        config = yaml.safe_load(f)
+    exp = client.create_experiment(config, conf.fixtures_path("no_op"))
+    assert exp.wait() == _client.ExperimentState.COMPLETED
+
+    trials = exp.get_trials()
+    assert len(trials) == 1, trials
+    trial = trials[0]
+    assert client.get_trial(trial.id).id == trial.id
+
+    ckpt = trial.top_checkpoint()
+    assert trial.select_checkpoint(latest=True).uuid == ckpt.uuid
+    assert exp.top_checkpoint().uuid == ckpt.uuid
+    assert ckpt.uuid in (c.uuid for c in exp.top_n_checkpoints(100))
+    assert client.get_checkpoint(ckpt.uuid).uuid == ckpt.uuid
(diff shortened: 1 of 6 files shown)


harness/determined/common/experimental/experiment.py
```diff
@@ -131,7 +130,7 @@ def pause(self) -> None:
     def unarchive(self) -> None:
         self._session.post(f"/api/v1/experiments/{self.id}/unarchive")
 
-    def wait(self, interval: int = 5) -> ExperimentState:
+    def wait(self, interval: float = 0.5) -> ExperimentState:
```
azhou-determined, line 133: what's the reasoning behind this?

rb-determined-ai: oops, I meant to change it to a `float`, not to change the default value

harness/determined/common/experimental/model.py
```diff
@@ -300,9 +305,14 @@ def set_labels(self, labels: List[str]) -> None:
...
-            labels (List[string]): All labels to set on the model.
+            labels (List[str]): All labels to set on the model.
         """
-        self.labels = labels
+        if not isinstance(labels, Iterable):
+            raise ValueError(
+                f"set_labels() requires a list of strings as input but got: {labels}"
+            )
+
+        self.labels = list(labels)
```
azhou-determined, line 315: why is this list() call necessary?

rb-determined-ai: Previously we'd write whatever you put onto the wire.

I added a check, but not a check to make sure it was actually a `list`, just that we could iterate through it (which is more pythonic).

But this code actually caches the value to the object.  So 1) it shouldn't be the same list object, and 2) the value on the object should actually be a list, even if the user passes a tuple, or the object's typing will be wrong.

e2e_tests/tests/test_sdk.py
```diff
@@ -0,0 +1,166 @@
...
+    exp.delete()
+    deleting_exp = exp
+
+    # Create another experiment and kill it.
+    exp = _make_live_experiment(client)
+    exp.kill()
+    assert exp.wait() == _client.ExperimentState.CANCELED
+
+    # Test remaining methods
+    exp.archive()
```
azhou-determined, line 85: do we need asserts for these status changes?

e2e_tests/tests/test_sdk.py
```diff
@@ -0,0 +1,166 @@
...
+    exp.pause()
+    with pytest.raises(ValueError, match="Make sure the experiment is active"):
+        # Wait throws an error if the experiment is paused by a user.
+        exp.wait(interval=0.1)
+
+    exp.activate()
+
+    exp.cancel()
+    assert exp.wait() == _client.ExperimentState.CANCELED
+    exp.wait()
```
azhou-determined, line 71: is this additional .wait() after the assert necessary?

azhou-determined (approved):
lgtm

e2e_tests/tests/test_sdk.py
rb-determined-ai: nope, that's a carryover from some intermediate state of this code

e2e_tests/tests/test_sdk.py
rb-determined-ai: I can add those.

PR #4750, opened 2022-08-10
fix: Incompatible variable type fixed

luca-digrazia:
**"filename"**: "harness/determined/cli/command.py"
**"warning_type"**: "Incompatible variable type [9]",
**"warning_message"**: " Unable to unpack `str`, expected a tuple.",
**"warning_line"**: 270,
**"fix"**: remove type


ioga:
static code analysis

PR #4754, opened 2022-08-10
fix: use correct experiment list offset when deleting an experiment [DET-7880]

julian-determined-ai:
## Description

fix: use correct experiment list offset when deleting an experiment [DET-7880]

Currently, the table `offset` is not correctly updated when an experiment is deleted. We want to re-calculate the correct offset if the current offset is greater than or equal to the total number of items in the table. 

## Test Plan

1. Navigate to the experiments list and navigate to the last page of experiments.
2. Delete or Archive all the experiments on the last page. 
3. After deleting the last experiment on the last page, the table to automatically show the previously second to last page. 

## Checklist

[DET-7880]: https://determinedai.atlassian.net/browse/DET-7880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ProjectDetails.tsx
+++ b/webui/react/src/pages/ProjectDetails.tsx
@@ -754,9 +754,9 @@ const ProjectDetails: React.FC = () => {
 
   useEffect(() => {
     if (settings.tableOffset >= total && total){
-    const newTotal = settings.tableOffset > total ? total : total-1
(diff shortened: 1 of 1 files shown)


gt2345 (approved):
Not sure why I cannot delete experiment on netlify cluster, but archiving works great.

PR #4756, opened 2022-08-10
fix: job queue pagination

trentwatt:
## Description
[DET-7841] [DET-7842]

- adds filtering by state on the API side for pagination
- updates webui to use those filters
- modifies `useSettings` to take a parameter that makes them only apply on certain routes
- uses that parameter to make sure settings for different job queue table to not affect the others

## Test Plan

create a bunch of jobs at once by spamming `det e create`. this should result in a bunch of Queued jobs and a few active jobs.

- The jobs should all be displayed properly in their respective tabs
- the pagination in each tab should work correctly
- it should not be possible to mess anything up by navigating back and forth between tabs (except that the pagination will be reset)

## Commentary (optional)

see PR comments

## Checklist

[DET-7841]: https://determinedai.atlassian.net/browse/DET-7841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-7842]: https://determinedai.atlassian.net/browse/DET-7842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/master/internal/job/job.go
+++ b/master/internal/job/job.go
@@ -59,7 +59,11 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (sproto.AQu
 	return resp, nil
 }
 
-func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool, states []jobv1.State) ([]*jobv1.Job, error) {
+func (j *Jobs) getJobs(
+	ctx *actor.Context,
+	resourcePool string,
+	desc bool,
+	states []jobv1.State) ([]*jobv1.Job, error) {
 	jobQ, err := j.jobQSnapshot(ctx, resourcePool)
 	if err != nil {
 		return nil, err
@@ -138,7 +142,11 @@ func (j *Jobs) Receive(ctx *actor.Context) error {
 		delete(j.actorByID, msg.JobID)
 
 	case *apiv1.GetJobsRequest:
-		jobs, err := j.getJobs(ctx, msg.ResourcePool, msg.OrderBy == apiv1.OrderBy_ORDER_BY_DESC, msg.States)
+		jobs, err := j.getJobs(
+			ctx,
+			msg.ResourcePool,
+			msg.OrderBy == apiv1.OrderBy_ORDER_BY_DESC,
+			msg.States)
 		if err != nil {
 			ctx.Respond(err)
 			return nil
--- a/master/internal/job/job.go
+++ b/master/internal/job/job.go
@@ -5,6 +5,8 @@ import (
 	"sort"
 	"strings"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/pkg/errors"
 
 	"github.com/determined-ai/determined/master/internal/rm"
@@ -57,7 +59,7 @@ func (j *Jobs) jobQSnapshot(ctx *actor.Context, resourcePool string) (sproto.AQu
 	return resp, nil
 }
 
-func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool) ([]*jobv1.Job, error) {
+func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool, states []jobv1.State) ([]*jobv1.Job, error) {
 	jobQ, err := j.jobQSnapshot(ctx, resourcePool)
 	if err != nil {
 		return nil, err
@@ -82,8 +84,14 @@ func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool) ([]*j
 	for jID, jRMInfo := range jobQ {
 		v1Job, ok := jobs[jID]
 		if ok {
+			// interesting that the update is a side effect
+			// of the getJobs function. I am guessing that
+			// I should leave it, regardless of filters?
 			UpdateJobQInfo(v1Job, jRMInfo)
-			jobsInRM = append(jobsInRM, v1Job)
+
+			if states == nil || slices.Contains(states, v1Job.Summary.State) {
+				jobsInRM = append(jobsInRM, v1Job)
+			}
 		}
 	}
 
@@ -130,7 +138,7 @@ func (j *Jobs) Receive(ctx *actor.Context) error {
 		delete(j.actorByID, msg.JobID)
 
 	case *apiv1.GetJobsRequest:
-		jobs, err := j.getJobs(ctx, msg.ResourcePool, msg.OrderBy == apiv1.OrderBy_ORDER_BY_DESC)
+		jobs, err := j.getJobs(ctx, msg.ResourcePool, msg.OrderBy == apiv1.OrderBy_ORDER_BY_DESC, msg.States)
 		if err != nil {
 			ctx.Respond(err)
 			return nil
--- a/webui/react/src/hooks/useSettings.ts
+++ b/webui/react/src/hooks/useSettings.ts
@@ -56,6 +56,7 @@ export interface SettingsConfigProp {
 }
 
 export interface SettingsConfig {
+  applicableRoutespace?: string;
   settings: SettingsConfigProp[];
   storagePath: string;
 }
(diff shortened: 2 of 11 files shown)


master/internal/job/job.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"golang.org/x/exp/slices"
 
-	"github.com/pkg/errors"
+	"github.com/shopspring/decimal"
 
 	"github.com/determined-ai/determined/master/pkg/actor"
 	"github.com/determined-ai/determined/master/pkg/model"
 	"github.com/determined-ai/determined/proto/pkg/apiv1"
 	"github.com/determined-ai/determined/proto/pkg/jobv1"
+	"github.com/pkg/errors"
```
trentwatt, line 16: only thing new is the addition of standard library `slices` package, not sure why the rest got reordered. new gofumpt formatter?

master/internal/job/job.go
```diff
@@ -216,8 +217,14 @@ func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool) ([]*j
 	for jID, jRMInfo := range jobQ {
 		v1Job, ok := jobs[jID]
 		if ok {
+			// interesting that the update is a side effect
+			// of the getJobs function. I am guessing that
+			// I should leave it, regardless of filters?
 			UpdateJobQInfo(v1Job, jRMInfo)
-			jobsInRM = append(jobsInRM, v1Job)
+
+			if states == nil || slices.Contains(states, v1Job.Summary.State) {
```
trentwatt, line 225: `UpdateJobQInfo` makes sure `v1Job.Summary` is not null

webui/react/src/hooks/useSettings.ts
```diff
@@ -56,6 +56,7 @@ export interface SettingsConfigProp {
 }
 
 export interface SettingsConfig {
+  applicableRoutespace?: string;
```
trentwatt, line 59: basically `offset` exists on two settings at the same time, they are erroneously getting each others' settings from the url. 

giving them their own routespace prevents this.

webui/react/src/hooks/useSettings.ts
```diff
@@ -384,6 +390,9 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
   useEffect(() => {
     if (pathChange.type === PathChangeType.None) return;
 
+    // probably don't need this, we do need in updateSettings though
+    if (!location.pathname.includes(config.applicableRoutespace ?? '')) return;
```
trentwatt, line 394: leave it just to be safe though?

webui/react/src/hooks/useSettings.ts
```diff
@@ -358,6 +361,9 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
   useEffect(() => {
     if (location.search === prevSearch) return;
 
+    // probably don't need this, we do need in updateSettings though
+    if (!location.pathname.includes(config.applicableRoutespace ?? '')) return;
```
trentwatt, line 365: leave it just to be safe though?

webui/react/src/pages/ResourcepoolDetail.tsx
```diff
@@ -101,7 +101,7 @@ const ResourcepoolDetail: React.FC = () => {
     if (!pool) return;
     setTabKey(key);
     const basePath = paths.resourcePool(pool.name);
-    history.replace(key === DEFAULT_TAB_KEY ? basePath : `${basePath}/${key}`);
+    history.replace(`${basePath}/${key}`);
```
trentwatt, line 104: need the default tab to have it's own route to avoid conflicts

webui/react/src/services/api-ts-sdk/api.ts
```diff
@@ -14597,11 +14603,12 @@ export const InternalApiFactory = function (configuration?: Configuration, fetch
...
          * @param {*} [options] Override http request option.
          * @throws {RequiredError}
          */
-        getJobs(paginationOffset?: number, paginationLimit?: number, resourcePool?: string, orderBy?: 'ORDER_BY_UNSPECIFIED' | 'ORDER_BY_ASC' | 'ORDER_BY_DESC', options?: any) {
-            return InternalApiFp(configuration).getJobs(paginationOffset, paginationLimit, resourcePool, orderBy, options)(fetch, basePath);
+        getJobs(paginationOffset?: number, paginationLimit?: number, resourcePool?: string, orderBy?: 'ORDER_BY_UNSPECIFIED' | 'ORDER_BY_ASC' | 'ORDER_BY_DESC', states?: Array<'STATE_UNSPECIFIED' | 'STATE_QUEUED' | 'STATE_SCHEDULED' | 'STATE_SCHEDULED_BACKFILLED'>, options?: any) {
+            return InternalApiFp(configuration).getJobs(paginationOffset, paginationLimit, resourcePool, orderBy, states, options)(fetch, basePath);
```
trentwatt, line 14611: not sure why the bindings get generated as 
```
states?: Array<'STATE_UNSPECIFIED' | 'STATE_QUEUED' | 'STATE_SCHEDULED' | 'STATE_SCHEDULED_BACKFILLED'>
```

even though the thing its generated from is also present in the bindings as `determinedv1JobState`. do you know if there is a way to make it just use generated struct here cc @hamidzr ?

this is the reason for the hacky decoder code below.

webui/react/src/services/decoder.ts
```diff
@@ -697,3 +697,11 @@ export const mapV1Project = (data: Sdk.V1Project): types.Project => {
     workspaceName: data.workspaceName ?? '',
   };
 };
+
+export const decodeJobStates = (states?: Sdk.Determinedjobv1State[]): Array<
+'STATE_UNSPECIFIED' | 'STATE_QUEUED' | 'STATE_SCHEDULED' | 'STATE_SCHEDULED_BACKFILLED'
+> => {
+  return states as unknown as Array<
+  'STATE_UNSPECIFIED' | 'STATE_QUEUED' | 'STATE_SCHEDULED' | 'STATE_SCHEDULED_BACKFILLED'
+  >;
```
trentwatt, line 706: this is the hacky decoder code being referred to above

webui/react/src/services/api-ts-sdk/api.ts
hamidzr: yeah this is unfortunate but how the generator generates the parameters. it just behaes weird around enums and these I think come through in a sub optimal way through the generated open api spec as well. we added a hacky solution in our in house python generator to address this https://github.com/hamidzr/determined/blob/0d4fe2385f0777b650e8841cfce9c21c1c9276e1/bindings/generate_bindings.py#L506-L511

hamidzr:
there might be an issue with this. originally paginating jobs hadn't been in scope when the job endpoints and job queue experience was designed. thus some of the operations in the client depend on the client being able to compute update requests that rely on finding what job is currently occupying a current position in the queue. This could be addressed by probably changing the api interface or introducing new ones so that 

when clients wants to put job A in position N it can ask the backend:
- put job A in position N
- instead of : put job A before or ahead of job B (current state) where it needs to figure out what job is occupying position N

master/internal/job/job.go
```diff
@@ -216,8 +218,14 @@ func (j *Jobs) getJobs(ctx *actor.Context, resourcePool string, desc bool) ([]*j
 	for jID, jRMInfo := range jobQ {
 		v1Job, ok := jobs[jID]
 		if ok {
+			// interesting that the update is a side effect
+			// of the getJobs function. I am guessing that
+			// I should leave it, regardless of filters?
 			UpdateJobQInfo(v1Job, jRMInfo)
```
gt2345, line 224: I believe this is to make sure the job queue info is as up to date as possible

gt2345 (approved):
Status filter and the pagination are great (though the pagination was already there?) , one thing (probably not related) I was not sure is if the percentage is suppose to display this way, looks kinda off... 
Screen Shot 2022-08-11 at 5 57 03 PM

master/internal/job/job.go
trentwatt: https://user-images.githubusercontent.com/28383080/184258065-5bb6fdbb-a056-42ec-ac99-48bf2a3cfe17.mp4

this was the issue this PR is intended to resolve

trentwatt: that does look off though, i agree. is that a regression?

hamidzr: the pagination issue shown in the video. I see that in master as well

PR #4758, opened 2022-08-10
feat: persist custom searcher events [DET-7833]

nrajanee:
## Description

feat: persist custom searcher events [DET-7833]

[DET-7833]: https://determinedai.atlassian.net/browse/DET-7833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mpkouznetsov (approved):
Looks good to _me_. I would defer to @dzhu .

master/pkg/searcher/custom_search.go
```diff
@@ -3,17 +3,22 @@ package searcher
...
-		// store the operations
-		// store the events
-		SearchMethodType     SearchMethodType `json:"search_method_type"`
-		SearcherEventQueue   *SearcherEventQueue
+		SearchMethodType   SearchMethodType `json:"search_method_type"`
+		SearcherEventQueue *SearcherEventQueue
+		// a list of marshaled protobuf experimentv1.Events object.
+		// Each protobuf obj needs to be marshaled individually.
+		PBEventsJSON         []json.RawMessage
 		customSearchProgress float64
```
dzhu, line 22: I think the progress is also going to need to be exported and serialized, so that we have a value after restoration and before the next trial progress event comes in.

Though, after seeing this, I think I would propose doing something different altogether, where we have a separate type used only for serialization, and do a quick translation between the two. I think that would be nicer than of muddying up the main type with the field `PBEventsJSON` field.

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -41,3 +57,58 @@ func (q *SearcherEventQueue) RemoveUpTo(eventID int) error {
...
+func (q *SearcherEventQueue) MarshalJSON() ([]byte, error) {
+	marshaledPBEvents, err := marshalPBEvents(q.events)
+	if err != nil {
+		return nil, err
+	}
+
+	var searcherEQJSON searcherEventQueueJSON
+	searcherEQJSON.EventsJSON = marshaledPBEvents
+	searcherEQJSON.EventCount = q.eventCount
+	return json.Marshal(searcherEQJSON)
```
dzhu, line 71:
```suggestion
	return json.Marshal(searcherEventQueueJSON{
		EventsJSON: marshaledPBEvents,
		EventCount: q.eventCount,
	})
```

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -41,3 +57,58 @@ func (q *SearcherEventQueue) RemoveUpTo(eventID int) error {
...
+}
+
+// UnmarshalJSON unmarshals searcherEventQueueJSON.
+func (q *SearcherEventQueue) UnmarshalJSON(sJSON []byte) error {
+	var searcherEQJSON searcherEventQueueJSON
+	err := json.Unmarshal(sJSON, &searcherEQJSON)
+	if err != nil {
```
dzhu, line 78:
```suggestion
	if err := json.Unmarshal(sJSON, &searcherEQJSON); err != nil {
```

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -41,3 +57,58 @@ func (q *SearcherEventQueue) RemoveUpTo(eventID int) error {
...
+	}
+	return marshaledPBEvents, nil
+}
+
+func unmarshalPBEvents(mEvents []json.RawMessage) ([]*experimentv1.SearcherEvent, error) {
+	unmarshaledPBEvents := make([]*experimentv1.SearcherEvent, 0)
+	for _, mEvent := range mEvents {
+		var pbEvent experimentv1.SearcherEvent
+		err := protojson.Unmarshal(mEvent, &pbEvent)
+		if err != nil {
```
dzhu, line 107:
```suggestion
		if err := protojson.Unmarshal(mEvent, &pbEvent); err != nil {
```

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -31,6 +42,11 @@ func (q *SearcherEventQueue) GetEvents() []*experimentv1.SearcherEvent {
 	return q.events
 }
 
+// SetEvents sets the events.
```
dzhu, line 45: Remove this function.

dzhu (approved):

PR #4759, opened 2022-08-10
fix: cancel experiment's sole trial should cancel experiment [DET-6707]

stoksc:
## Description
Exactly what it says in the release note.

## Test Plan
- [x] wrote a test

## Commentary (optional)
I renamed user_canceled -> user_requested_stop internally for clarify. Also removed user_requested_stop from the protos, since it is unused (and shouldn't be used).

## Checklist


docs/release-notes/canceled-experiment-trial-fix.txt
```diff
@@ -0,0 +1,6 @@
+:orphan:
+
+**Fixes**
+
+-  Experiments with a single trial are now considered canceled when their trial is
```
ioga, line 5: will that also be now true for HP search experiments where a user kills all trials as they appear?

ioga (approved):

docs/release-notes/canceled-experiment-trial-fix.txt
stoksc: yes.

stoksc: but they would need to kill them all

PR #4762, opened 2022-08-10
feat: persisting state on client

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/DET-8011

## Checklist


harness/determined/searcher/search_method.py
```diff
@@ -182,18 +206,42 @@ def on_trial_exited_early(
...
         from the ``SearchMethod``
         """
+        searcher_state_file = path.joinpath(STATE_FILE)
+        self.save_method_state(path, event_id)
+        d = self.searcher_state.to_dict()
+        d["experimentId"] = experiment_id
+        with searcher_state_file.open("w") as f:
+            json.dump(d, f)
+
+    def save_method_state(self, path: pathlib.Path, event_id: int) -> None:
```
liamcli, line 222: do we need event_id here for `save_method_state` and `load_method_state`?  i feel like after your proposed change in handling event_id offset client side, we can probably remove it for the user facing api.

harness/determined/searcher/core_search_runner.py
```diff
@@ -0,0 +1,37 @@
...
+        self.latest_checkpoint = info.latest_checkpoint
+
+    def run(
+        self,
+        exp_config: Dict[str, Any],
+        context_dir: Optional[str] = None,
+    ) -> None:
+        logging.info("CoreSearchRunner.run")
+
+        if self.latest_checkpoint is not None:
```
liamcli, line 25: i slightly prefer moving out load as a separate method and calling it here in run.

harness/determined/searcher/search_runner.py
```diff
@@ -49,94 +34,189 @@ def run(
...
                         self.search_method.searcher_state.trial_progress[request_id] = float(
-                            e.trialProgress.partialUnits
+                            event.trialProgress.partialUnits
                         )
                         progress = self.search_method.progress()
                         operations = [Progress(progress)]
                     else:
-                        raise RuntimeError(f"Unsupported event {e}")
+                        raise RuntimeError(f"Unsupported event {event}")
                     bindings.post_PostSearcherOperations(
```
liamcli, line 119: i thought we'd check the last_event_id here and use that to decide whether we'd want to post operations or not.  or have some other way of skipping the first event if we already processed it client side.

harness/determined/searcher/search_runner.py
```diff
@@ -49,94 +34,189 @@ def run(
...
+                f"searcher_dir={self.searcher_dir} already exists and is not a directory"
+            )
+
+    def run(
+        self,
+        exp_config: Dict[str, Any],
+        context_dir: Optional[str] = None,
+    ) -> int:
+        """
+        Run custom search without an experiment id
```
aciborowska, line 183: Are we planning to allow users to resume a specific experiment by providing experiment_id? 
It looks like right now they can edit EXPERIMENT_ID_FILE to resume selected experiment anyway.

harness/determined/searcher/search_runner.py
```diff
@@ -49,94 +34,189 @@ def run(
...
-                        if e.validationCompleted.metric is None:
+                        request_id = uuid.UUID(event.validationCompleted.requestId)
+                        if event.validationCompleted.metric is None:
                             raise RuntimeError(
                                 "validationCompleted event is invalid without a metric"
                             )
                         operations = self.search_method.on_validation_completed(
                             request_id,
-                            e.validationCompleted.metric,
+                            event.validationCompleted.metric,
```
liamcli, line 97: i'm not sure if this is too late but i feel strongly that we need to be able to know the train length corresponding to the metric which is information available from the operation that generated this event. this is somewhat of a requirement for ease of use when it comes to multi-fidelity bayesian methods that suggest hparams with associated training durations and then updates the bayesian model over joint space of hparams and training duration.  it would also give us a way to bypass saving checkpoint for progress events by making sure we eventually update trialProgress for each validationCompleted event.

harness/determined/searcher/search_method.py
liamcli, line 222: thanks for making this api so much better!  much cleaner to handle shared searcher state completely and have the user just focus on the method state.  also user does not need to worry about event_id anymore as well.

harness/determined/searcher/search_runner.py
mpkouznetsov: Not sure I understand the question. Editing EXPERIMENT_ID_FILE is hacky / unsupported.

harness/determined/searcher/search_runner.py
mpkouznetsov: We can do it but let's leave it outside the scope of this PR.

harness/determined/searcher/search_runner.py
```diff
@@ -49,94 +34,210 @@ def run(
...
                         )
-                        if e.trialExitedEarly.exitedReason is None:
+                        if event.trialExitedEarly.exitedReason is None:
                             raise RuntimeError(
                                 "trialExitedEarly event is invalid without exitedReason"
                             )
-                        request_id = uuid.UUID(e.trialExitedEarly.requestId)
-                        if e.trialExitedEarly.exitedReason in (
+                        request_id = uuid.UUID(event.trialExitedEarly.requestId)
+                        if event.trialExitedEarly.exitedReason in (
```
liamcli, line 85: there are potentially some sharp edges around handling for invalidHP.  i'll think more about this but i think otherwise i'm on board with this implementation for fault handling.

liamcli: also, it's probably the case that users would prefer to apply invalidHP filters in python as part of search method instead of relying on training code to raise invalidHP once we have custom searcher so this may be a moot point.

PR #4763, opened 2022-08-10
perf: improve too many user api call

keita-determined:
## Description
[DET-7962]

## Test Plan
- Check if user api call is reduced (user api call less than 5 times before polling starts in network tab)
  - user api used to be called more than 5 times before polling starts
- Check if user data is fetched appropriately instead of no user data

## Checklist

[DET-7962]: https://determinedai.atlassian.net/browse/DET-7962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/components/UserAvatar.tsx
+++ b/webui/react/src/components/UserAvatar.tsx
@@ -1,7 +1,6 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 
 import { useStore } from 'contexts/Store';
-import { useFetchUsers } from 'hooks/useFetch';
 import Avatar, { Props as AvatarProps } from 'shared/components/Avatar';
 import { getDisplayName } from 'utils/user';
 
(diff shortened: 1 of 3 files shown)


keita-determined:
`test-cli ` is failed. I believe that it should fail cuz the code change is not related to what ci tests

webui/react/src/components/UserAvatar.tsx
```diff
@@ -10,16 +9,8 @@ export interface Props extends Omit<AvatarProps, 'darkLight' | 'displayName'> {
 }
 
 const UserAvatar: React.FC<Props> = ({ userId, ...rest }) => {
-  const [ displayName, setDisplayName ] = useState('');
   const { ui, users } = useStore();
-  const fetchUsers = useFetchUsers(new AbortController());
-
-  useEffect(() => {
-    if (!userId) return;
-    if (!users.length) fetchUsers();
```
keita-determined, line 19: When lots of user avatars is rendered at once `fetchUsers` is called many times, so `fetchUsers` is moved in `App.tsx` and call once. 
This implementation should be fine because lots of components is polling `fetchUsers` every few seconds

gt2345 (approved):
LGTM

PR #4764, opened 2022-08-10
fix: add unit to custom searcher config

mpkouznetsov:
## Description

Trials require the unit to be specified. All searchers have it included in MaxLength but custom searcher does not.

### Background
We noticed that after I removed max_length from the custom searcher configuration (because it is not supposed to be required), the custom searcher test started only running a single batch in each trial. Furthermore, the following message appeared in the trial logs:

```
root: The searcher configuration provided was configured without units, but the training loop you are using (one of the Trial APIs) requires a searcher configured with units.  Proceeding anyway, and assuming that the lengths configured in the searcher are in terms of epochs.
```

Trials quietly default to epochs as a unit.

## Checklist

### DIFF
--- a/harness/determined/core/_searcher.py
+++ b/harness/determined/core/_searcher.py
@@ -1,6 +1,6 @@
 import enum
 import logging
-from typing import Iterator, Optional, Any
+from typing import Any, Iterator, Optional
 
 import determined as det
 from determined import core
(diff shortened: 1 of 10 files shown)


harness/determined/core/_searcher.py
```diff
@@ -17,10 +17,17 @@ class Unit(enum.Enum):
 
 def _parse_searcher_units(experiment_config: dict) -> Optional[Unit]:
```
mpkouznetsov, line 18: @rb-determined-ai I would only ask you to look at this file,

aciborowska (approved):
LGTM

PR #4769, opened 2022-08-11
fix: Model table - tags filter [DET-8027]

mapmeld:
## Description

There's some logic error in how I set up tag-level filtering in `get_models.sql`. In this fixed version, you can choose tags and return any models with 1 or more of those tags. 

## Test Plan

- On the model registry page, create empty new models with no tags, common tags, and unique tags
- Verify that all model tags appear in the table's Tags column filter
- In the Tags column, use filters to limit table rows to models with the common tag, or with any of a few unique tags

## Checklist


EmilyBonar (approved):
LGTM! Had to look up postgres operators for this one haha

PR #4770, opened 2022-08-11
test: asha as custom searcher method [DET-8013]

aciborowska:
## Description
ASHA implemented as a Custom Searcher method (following asha.go).

## Test Plan

Test should perform 16 trials in 3 rungs with the max length of 3000 batches (187, 750 and 3000) batches. All trials should be marked as completed, same as experiment.

## Checklist

### DIFF
--- a/e2e_tests/tests/experiment/test_custom_searcher.py
+++ b/e2e_tests/tests/experiment/test_custom_searcher.py
@@ -256,8 +256,8 @@ def test_run_asha_batches_exp() -> None:
     assert search_method.completed_trials == 5
     assert len(search_method.searcher_state.trials_closed) == len(search_method.closed_trials)
 
-    response = bindings.get_GetExperimentTrials(session, experimentId=experiment_id)
-    for trial in response.trials:
+    response_trials = bindings.get_GetExperimentTrials(session, experimentId=experiment_id)
+    for trial in response_trials.trials:
         assert trial.state == bindings.determinedexperimentv1State.STATE_COMPLETED
         # number of batch for rung[0] = 750
         assert trial.totalBatchesProcessed >= 750
--- a/e2e_tests/tests/experiment/test_custom_searcher.py
+++ b/e2e_tests/tests/experiment/test_custom_searcher.py
@@ -6,7 +6,6 @@ import uuid
 from dataclasses import dataclass
 from typing import Dict, List, Set
 
-import numpy as np
 import pytest
 
 from determined.common.api import bindings
@@ -516,7 +515,7 @@ class ASHASearchMethod(SearchMethod):
     def sample_params(self) -> Dict[str, object]:
         hparams = {
             "global_batch_size": 10,
-            "metrics_base": 0.05 * (len(self.trial_rungs)+1),
+            "metrics_base": 0.05 * (len(self.trial_rungs) + 1),
             "metrics_progression": "constant",
         }
         logging.info(f"hparams={hparams}")
--- a/e2e_tests/tests/experiment/test_custom_searcher.py
+++ b/e2e_tests/tests/experiment/test_custom_searcher.py
@@ -1,9 +1,10 @@
 import dataclasses
 import logging
 import random
+import sys
 import uuid
 from dataclasses import dataclass
-from typing import Dict, List
+from typing import Dict, List, Set
 
 import numpy as np
 import pytest
@@ -283,17 +284,17 @@ class Rung:
         logging.info(f"Rung {self.idx}")
         logging.info(f"outstanding_trials {self.outstanding_trials}")
 
-        old_num_promote = int(len(self.metrics) / divisor)
-        num_promote = int((len(self.metrics) + 1) / divisor)
+        old_num_promote = len(self.metrics) // divisor
+        num_promote = (len(self.metrics) + 1) // divisor
 
         index = self._search_metric_index(metric)
-        promoteNow = index < num_promote
-        trial_metric = TrialMetric(request_id=request_id, metric=metric, promoted=promoteNow)
+        promote_now = index < num_promote
+        trial_metric = TrialMetric(request_id=request_id, metric=metric, promoted=promote_now)
         self.metrics.insert(index, trial_metric)
 
-        if promoteNow:
+        if promote_now:
             return [request_id]
-        elif num_promote != old_num_promote and not self.metrics[old_num_promote].promoted:
+        if num_promote != old_num_promote and not self.metrics[old_num_promote].promoted:
             self.metrics[old_num_promote].promoted = True
             return [self.metrics[old_num_promote].request_id]
 
@@ -319,7 +320,7 @@ class ASHASearchMethod(SearchMethod):
         max_trials: int,
         num_rungs: int,
         divisor: int = 2,
-        max_concurrent_trial: int = 0,
+        max_concurrent_trials: int = 0,
     ) -> None:
         super().__init__()
 
@@ -328,7 +329,7 @@ class ASHASearchMethod(SearchMethod):
         self.max_trials = max_trials
         self.num_rungs = num_rungs
         self.divisor = divisor
-        self.max_concurrent_trials = max_concurrent_trial
+        self.max_concurrent_trials = max_concurrent_trials
 
         self.is_smaller_better = True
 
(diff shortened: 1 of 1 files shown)


e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+        rung = self.rungs[rung_idx]
+        rung.outstanding_trials -= 1
+        added_train_workload = False
+
+        ops: List[Operation] = []
+
+        if rung_idx == self.num_rungs - 1:
+            rung.metrics.append(TrialMetric(request_id=request_id, metric=metric))
+
+            if request_id in self.early_exit_trials and self.early_exit_trials[request_id] is True:
```
mpkouznetsov, line 454: Line 206 in `asha.go` says
`if !s.EarlyExitTrials[requestID] {` which is the _opposite_ of your condition.

Also, you might consider using a set as a python equivalent of golang's `map[model.RequestID]bool`.

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+        self.max_trials = max_trials
+        self.num_rungs = num_rungs
+        self.divisor = divisor
+        self.max_concurrent_trials = max_concurrent_trial
+
+        self.is_smaller_better = True
+
+        # structs
+        self.rungs: List[Rung] = []
+        self.trial_rungs: Dict[uuid.UUID, int] = {}
```
mpkouznetsov, line 337: I think using a set here would result in simpler code, except for the serialization (sets are not json-serializable). Go just does not have a native set structure which is why we use maps to represent sets.

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+                    request_id not in self.early_exit_trials
+                    or not self.early_exit_trials[promoted_request_id]
+                ):
+                    logging.info(f"Promoted {promoted_request_id}")
+                    units_needed = max(next_rung.units_needed - rung.units_needed, 1)
+                    ops.append(ValidateAfter(promoted_request_id, units_needed))
+                    added_train_workload = True
+                    self.pending_trials += 1
+                else:
+                    return self.promote_async(promoted_request_id, 100.0)
```
mpkouznetsov, line 474: Should we use a number larger than 100.0 to be the equivalent of `math.MaxFloat64`? Maybe `sys.float_info.max` would do the trick?

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+    metrics: List[TrialMetric] = dataclasses.field(default_factory=list)
+    outstanding_trials: int = 0
+
+    def promotions_async(
+        self, request_id: uuid.UUID, metric: float, divisor: int
+    ) -> List[uuid.UUID]:
+        logging.info(f"Rung {self.idx}")
+        logging.info(f"outstanding_trials {self.outstanding_trials}")
+
+        old_num_promote = int(len(self.metrics) / divisor)
```
mpkouznetsov, line 286:
```suggestion
        old_num_promote = len(self.metrics) // divisor
```

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+        num_promote = int((len(self.metrics) + 1) / divisor)
+
+        index = self._search_metric_index(metric)
+        promoteNow = index < num_promote
```
mpkouznetsov, line 290:
```suggestion
        promote_now = index < num_promote
```

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +225,314 @@ def sample_params(self) -> Dict[str, int]:
...
+        trial_metric = TrialMetric(request_id=request_id, metric=metric, promoted=promoteNow)
+        self.metrics.insert(index, trial_metric)
+
+        if promoteNow:
+            return [request_id]
+        elif num_promote != old_num_promote and not self.metrics[old_num_promote].promoted:
```
mpkouznetsov, line 296:
```suggestion
        if num_promote != old_num_promote and not self.metrics[old_num_promote].promoted:
```

mpkouznetsov (commented):
A few suggestions.

e2e_tests/tests/experiment/test_custom_searcher.py
aciborowska: Good catch. thanks!
Changing from dictionaries to sets.

e2e_tests/tests/experiment/test_custom_searcher.py
aciborowska: While early_exit_trials and closed_trials should be sets in Python, trial_rungs actually maps a trial to a rung that the trial is currently in.

e2e_tests/tests/experiment/test_custom_searcher.py
aciborowska: My bad. Should be max float.

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +226,305 @@ def sample_params(self) -> Dict[str, int]:
...
+        "name": "custom",
+        "metric": "validation_error",
+        "smaller_is_better": True,
+        "unit": "batches",
+    }
+    config["name"] = "asha"
+    config["description"] = "custom searcher"
+
+    max_length = 3000
+    max_trials = 5
```
liamcli, line 244: with 3 rungs, we'd need at least 16 trials to see a trial on the top rung, so i suggest either 2 rungs with 8 trials or 3 rungs with 16 trials.  let's also see if we can get deterministic metrics so that we know exactly which and how many trials would be promoted to each rung and have checks to make sure things are as expected.

liamcli: we can get this behavior by configuring the no_op experiment config to have 0 std and a given metric_progression: https://github.com/determined-ai/determined/blob/custom-searcher/e2e_tests/tests/fixtures/no_op/model_def.py#L147

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +226,305 @@ def sample_params(self) -> Dict[str, int]:
...
+        ops: List[Operation] = []
+
+        for rung in self.rungs:
+            if rung.outstanding_trials > 0:
+                break
+            for trial_metric in rung.metrics:
+                if not trial_metric.promoted and trial_metric.request_id not in self.closed_trials:
+                    if trial_metric.request_id not in self.early_exit_trials:
+                        logging.info(f"Closing trial {trial_metric.request_id}")
+                        ops.append(Close(trial_metric.request_id))
```
liamcli, line 506: want to note that it's definitely not ideal that users have to create "Close" operations for invalid HP early exits but not other types of early exists since it results in inconsistent behavior.

e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -219,3 +226,305 @@ def sample_params(self) -> Dict[str, int]:
...
+        }
+        logging.info(f"hparams={hparams}")
+        return hparams
+
+    def progress(self) -> float:
+        if 0 < self.max_concurrent_trials < self.pending_trials:
+            raise RuntimeError("Pending trial is greater than max concurrent trials")
+
+        all_trials = len(self.rungs[0].metrics)
+        progress = all_trials / (1.2 * self.max_trials)
```
liamcli, line 524: this was intended to be a lower bound estimate on progress and there may be better alternatives. let me know if you have suggestions.

e2e_tests/tests/experiment/test_custom_searcher.py
aciborowska: Updated to 16 trials, 3 rungs, divisor 4. Added checks for number of batches processed by each trial to ensure promotion happens.

PR #4771, opened 2022-08-11
refactor: reduce unneeded api calls [DET-7451]

EmilyBonar:
## Test Plan
Go to the experiment details page with the network inspector open. There shouldn't be a call to `/api/v1/models?...` until and unless you open the Checkpoint modal and then the Register Checkpoint modal. Register Checkpoint modal should still work as expected.

## Checklist


mapmeld (approved):
LGTM, checked network tab and tested registering checkpoints -> model versions

PR #4774, opened 2022-08-11
fix: lint-python ci failure

liamcli:
## Description
mypy tests for model hub started failing when run with harness installed via `pip install -e harness` from the requirements.txt file.  installing in editable mode prevented mypy from finding determined stubs unless we modified mypy path to include the harness directory.  this will make sure `make check` tests will pass when developing locally and modifying `harness` code in editable mode. 

i haven't fixed the underlying issue which is we reinstall determined in editable mode into the python venv used for ci after installing determined from wheel.  there is also the issue of duplicated github actions tests and our standard circleci workflows.  

## Checklist

### DIFF
--- a/e2e_tests/mypy.ini
+++ b/e2e_tests/mypy.ini
@@ -1,4 +1,5 @@
 [mypy]
+mypy_path = ../harness
 python_version = 3.7
 follow_imports = silent
 ignore_missing_imports = True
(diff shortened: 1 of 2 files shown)


rb-determined-ai (approved):

dannysauer (approved):
Looks like a reasonable workaround.

PR #4783, opened 2022-08-12
feat: HF trainer callback

aciborowska:
## Description

Determined callback for HF Trainer API

## Checklist


examples/computer_vision/vision_transformer/image_classification.py
```diff
@@ -0,0 +1,397 @@
...
+    if training_args.push_to_hub:
+        trainer.push_to_hub(**kwargs)
+    else:
+        trainer.create_model_card(**kwargs)
+
+
+if __name__ == "__main__":
+    distributed = det.core.DistributedContext.from_torch_distributed()
+    with det.core.init(distributed=distributed) as core_context:
+        main(core_context)
```
aciborowska, line 397: In any other integration (MLFlow, WandB) this should happen in the callback init, like so:
```
def __init__(self, sth):
    assert is_determined_available(), "DetCallback requires determined to be installed. Run `pip install determined`."
    import determined
    self._det = determined

[...]
def setup():
    distributed = self._det.core.DistributedContext.from_torch_distributed()
    self.core_context = self._det.core.init(distributed=distributed)
```
This will work, but as soon as you try to add preemption, it will stop working since context was not properly initialized... also our docs are specifically saying not to do this. If we ever want to upstream to HF, that would have to be solved.

aciborowska, line 397: In any other integration (MLFlow, WandB) this should happen in the callback init, like so:
```
def __init__(self, sth):
    assert is_determined_available(), "DetCallback requires determined to be installed. Run `pip install determined`."
    import determined
    self._det = determined

[...]
def setup():
    distributed = self._det.core.DistributedContext.from_torch_distributed()
    self.core_context = self._det.core.init(distributed=distributed)
```
This will work, but as soon as you try to add preemption, it will stop working since context was not properly initialized... also our docs are specifically saying not to do this. If we ever want to upstream to HF, that would have to be solved.

examples/computer_vision/vision_transformer/image_classification.py
```diff
@@ -0,0 +1,397 @@
...
+        args=training_args,
+        train_dataset=dataset["train"] if training_args.do_train else None,
+        eval_dataset=dataset["validation"] if training_args.do_eval else None,
+        compute_metrics=compute_metrics,
+        tokenizer=feature_extractor,
+        data_collator=collate_fn,
+    )
+
+    det_callback = DetCallback(core_context, training_args, filter_metrics=['loss', 'accuracy'],
+                               checkpoint_metadata={'feature_extractor': feature_extractor})
```
aciborowska, line 358: This callback is mixing saving and loading. While saving makes sense as this is part of the training loop, loading is not. The main disadvantage of this setting is that data uploaded through context.checkpoint.upload() will not be downloaded without creating det callback. Another problem: what happens if a user tries to download data while not being in core context?

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,115 @@
...
+        if self.core_context.preempt.should_preempt():
+            return
+
+
+def override_training_args(training_args):
+    hparams = det.get_cluster_info().trial.hparams
+    for k, v in hparams.items():
+        if hasattr(training_args, k):
+            training_args.k = v
+    return training_args
```
aciborowska, line 115: To fully support Trainer API integration we need to add Hp scheduler (e.g., [Ray](https://github.com/huggingface/transformers/blob/main/src/transformers/integrations.py#L171))

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+
+        assert is_determined_available(), "DetCallback requires determined to be installed. Run `pip install determined`."
+        import determined
+        self._det = determined
+        distributed = self._det.core.DistributedContext.from_torch_distributed()
+        self.core_context = self._det.core.init(distributed=distributed)
+        self.core_context.__enter__()
+        weakref.finalize(self, exit_context, self.core_context)
+
+        self.filter_metrics = filter_metrics
```
aciborowska, line 33: Not required but useful to remove some of HF metrics

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+            if self.tokenizer is not None:
+                self.tokenizer.save_pretrained(os.path.join(save_path, "tokenizer"))
+
+            if self.user_checkpoint_metadata is not None:
+                self._on_save_user_data(save_path)
+
+            self.core_context.checkpoint.upload(save_path, det_checkpoint_metadata)
+
+        if self.core_context.preempt.should_preempt():
+            raise Exception("Process preempted / killed")
```
aciborowska, line 87: AFAIK, that's the safest place to do preemption, since Trainer completed saving its data, and determined uploaded directory with Trainer data, tokenizer and other user-specific objects to persistent storage.

In Trainer API loop on_epoch_end/on_step end happens before on_save, so we don't want to preempt there.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+        weakref.finalize(self, exit_context, self.core_context)
+
+        self.filter_metrics = filter_metrics
+        self.user_checkpoint_metadata = checkpoint_metadata
+        self.tokenizer = tokenizer
+        self.tokenizer_options = tokenizer_options
+        self.load_last_checkpoint(args)
+
+        self.last_eval_metrics = None
+        self.searcher_metric = self._det.get_cluster_info().trial._config['searcher']['metric']
```
aciborowska, line 40: Is there a better way to get searcher metric?

examples/computer_vision/vision_transformer/image_classification.py
```diff
@@ -0,0 +1,394 @@
...
+        args=training_args,
+        train_dataset=dataset["train"] if training_args.do_train else None,
+        eval_dataset=dataset["validation"] if training_args.do_eval else None,
+        compute_metrics=compute_metrics,
+        tokenizer=feature_extractor,
+        data_collator=collate_fn,
+    )
+
+    det_callback = DetCallback(training_args, filter_metrics=['loss', 'accuracy'], tokenizer=feature_extractor)
+    trainer.add_callback(det_callback)
```
aciborowska, line 358: The only change compared to the original script from HF.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+        self.searcher_ops = self.core_context.searcher.operations()
+        self.current_op = next(self.searcher_ops)
+
+    def on_log(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, model=None, logs=None,
+               **kwargs):
+        if state.is_world_process_zero:
+            metrics = self.process_log(logs)
+            if 'loss' in metrics:
```
coreystatendet, line 48: Is `loss` or `eval_loss` always present?  Might be more resilient to check some other way whether we're in train or val -- looks like maybe `state.epoch is None` when we're in val?

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+
+        if self.core_context.preempt.should_preempt():
+            raise Exception("Process preempted / killed")
+
+    def _on_save_user_data(self, save_path) -> None:
+        '''
+        User-defined saving of objects from self.checkpoint_metadata under save_path.
+        After objects are saved, Determined handles uploading and downloading objects to/from selected storage.
+        '''
+        logging.warning("No implementation for _on_save_user_data. "
```
coreystatendet, line 94: nit: if we're publishing this, I'd make this a `raise NotImplementedException` if the metadata isn't `None`, just in case.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+            metrics = {}
+            for k, v in logs.items():
+                if any(m in k for m in self.filter_metrics) is True:
+                    metrics[k] = v
+
+        return metrics
+
+    def on_save(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, **kwargs):
+        info = self._det.get_cluster_info()
+        assert info is not None
```
coreystatendet, line 70: The `info is None` case should probably skip saving checkpoints to the store instead of erroring, to support local mode in the future.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,150 @@
...
+        '''
+        User-defined saving of objects from self.checkpoint_metadata under save_path.
+        After objects are saved, Determined handles uploading and downloading objects to/from selected storage.
+        '''
+        logging.warning("No implementation for _on_save_user_data. "
+                        "Objects passed to the callback via checkpoint_metadata will not be saved.")
+
+    def load_last_checkpoint(self, args: typing.Dict) -> None:
+        info = self._det.get_cluster_info()
+        assert info is not None
```
coreystatendet, line 99: Similar here, skip if `info is None`?

examples/computer_vision/vision_transformer/det_callback.py
aciborowska: Good point.

[`Loss` is always present in the log](https://github.com/huggingface/transformers/blob/a123eee9df11b5a116452c3f539f2ff0427324d2/src/transformers/trainer.py#L2038) and refers to the loss during training. After HF Trainer finishes training, it also reports `train_loss` (average loss), which may cause DB error, if we try to report `train_loss` with the same step as previously reported `loss`. I decided to omit `train_loss`, but I am open to suggestions. Be aware that it's a single point and looks suspicious in our WebUI.

`eval_loss` is more tricky. For the evaluation step, we would get `eval_loss` [assuming loss is not None.](https://github.com/huggingface/transformers/blob/a123eee9df11b5a116452c3f539f2ff0427324d2/src/transformers/trainer.py#L3059). 
I see that WandB callback and TensorBoard callback use [re-write function](https://github.com/huggingface/transformers/blob/a123eee9df11b5a116452c3f539f2ff0427324d2/src/transformers/integrations.py#L461), that checks for prefixes, such as `eval_` or `test_`. I will follow the same approach and check for prefix since it seems safer than an exact check for `eval_loss`.

aciborowska: `state.epoch` is not None when we run validation during training. I decided to do sth similar to the [re-write function](https://github.com/huggingface/transformers/blob/a123eee9df11b5a116452c3f539f2ff0427324d2/src/transformers/integrations.py#L461).

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,187 @@
...
+
+import os
+import determined as det
+import logging
+
+import weakref
+import importlib.util
+
+
+class DetCallback(TrainerCallback):
```
liamcli, line 18: instead of having this as an example, i wonder if we should have module in harness with integrations via callbacks.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,187 @@
...
+        self.tokenizer = tokenizer
+        self.tokenizer_options = tokenizer_options
+        self.load_last_checkpoint(args)
+
+        self.last_metrics = {}
+        self.searcher_metric = self._det.get_cluster_info().trial._config['searcher']['metric']
+        self.searcher_ops = self.core_context.searcher.operations()
+        self.current_op = next(self.searcher_ops)
+
+    def on_log(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, model=None, logs=None,
```
liamcli, line 44: model argument doesn't seem to be used here

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,187 @@
...
+            if self.tokenizer_options is not None:
+                det_checkpoint_metadata['tokenizer_options'] = self.tokenizer_options
+
+            if self.tokenizer is not None:
+                self.tokenizer.save_pretrained(os.path.join(save_path, "tokenizer"))
+
+            if self.user_checkpoint_metadata is not None:
+                self._on_save_user_data(save_path)
+
+            self.core_context.checkpoint.upload(save_path, det_checkpoint_metadata)
```
liamcli, line 88: this could fail if a user is using the deepspeed backend with sharded checkpoint that need to be take across multiple nodes in which case only files on the rank 0 node would get uploaded.

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,187 @@
...
+        if latest_checkpoint is not None:
+            metadata = self.core_context.checkpoint.get_metadata(latest_checkpoint)
+            prev_trial_id = metadata["trial_id"]
+            trial_id = info.trial.trial_id
+            if trial_id != prev_trial_id:
+                resume_step = 0
+            else:
+                resume_step = metadata['steps_completed']
+            checkpoint_path = os.path.join(args.output_dir, f"checkpoint-{resume_step}")
+            self.core_context.checkpoint.download(latest_checkpoint, checkpoint_path)
```
liamcli, line 119: would all ranks perform this download? we should probably limit it to local rank 0 processes on each node so we do not duplicate multiple times per node

examples/computer_vision/vision_transformer/det_callback.py
```diff
@@ -0,0 +1,187 @@
...
+
+            args.resume_from_checkpoint = checkpoint_path
+
+    def on_step_end(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, **kwargs):
+        if self.core_context.preempt.should_preempt():
+            control.should_save = True
+
+    def on_epoch_end(self, args: TrainingArguments, state: TrainerState, control: TrainerControl, **kwargs):
```
liamcli, line 127: i like how you handled getting searcher ops

liamcli (commented):
looking pretty good.  just a few comments for distributed checkpointing and a question for us on where we want to put the callback.

e2e_tests/tests/nightly/test_distributed.py
```diff
@@ -172,6 +172,16 @@ def test_byol_pytorch_distributed() -> None:
     exp.run_basic_test_with_temp_config(config, conf.cv_examples_path("byol_pytorch"), 1)
 
 
+@pytest.mark.distributed
+@pytest.mark.gpu_required
+def test_hf_trainer_api_intergration() -> None:
```
liamcli, line 177: intergration -> integration

examples/integrations/hf_trainer_api/README.md
```diff
@@ -0,0 +1,58 @@
...
+This example is adapted from the 
+[VisionTransformer example in the Hugging Face examples](https://github.com/huggingface/transformers/tree/main/examples/pytorch/image-classification). 
+It is intended to demonstrate how to use Determined callback with Hugging Face Trainer API to
+enable Determined's distributed training, fault tolerance, checkpointing and metrics reporting.
+
+## Files
+* **image_classification.py**: The code from Hugging Face that (1) loads Vision Transformer from Model Hub; (2)
+configure Trainer; (3) uses Determined callback.
+
+The key portion of the code is providing Determined callback to the Trainer in line 413:
```
liamcli, line 11: The key portion of the code passing the Determined callback to the Transformers Trainer is in line 413

examples/integrations/hf_trainer_api/README.md
```diff
@@ -0,0 +1,58 @@
...
+* **ds_config_stage_1.json**: Optimizer state partitioning (ZeRO stage 1).
+* **ds_config_stage_2.json**: Gradient partitioning (ZeRO stage 2).
+* **ds_config_stage_2_cpu_offload.json**: Gradient partitioning and CPU offloading (ZeRO stage 2).
+* **ds_config_stage_3.json**: Parameter partitioning (ZeRO stage 3).
+
+To learn more about DeepSpeed, see [DeepSpeed docs](https://deepspeed.readthedocs.io/en/latest/) and 
+[HF DeepSpeed integration](https://huggingface.co/docs/transformers/main_classes/deepspeed).
+
+## Data
+This example uses [beans dataset](https://huggingface.co/datasets/beans).
```
liamcli, line 34: ^ the beans dataset

examples/integrations/hf_trainer_api/adaptive.yaml
```diff
@@ -0,0 +1,49 @@
+name: vision transformer adaptive
```
liamcli, line 1: do we actually find a configuration that outperforms the default hyperparameter setting with asha?

examples/integrations/hf_trainer_api/adaptive.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - NCCL_DEBUG=INFO
+    # You may need to modify this to match your network configuration.
+    - NCCL_SOCKET_IFNAME=ens,eth,ib
+resources:
+  slots_per_trial: 2
+records_per_epoch: 1000
+searcher:
+  name: adaptive_asha
+  max_length:
+    epochs: 5
```
liamcli, line 14: we should switch to batches so asha can be more aggressive with downsampling e.g. have max_rungs of 5 if max_length is >=256 batches

examples/integrations/hf_trainer_api/deepspeed.yaml
```diff
@@ -0,0 +1,36 @@
...
+debug: false
+environment:
+    environment_variables:
+        - NCCL_DEBUG=INFO
+        # You may need to modify this to match your network configuration.
+        - NCCL_SOCKET_IFNAME=ens,eth,ib
+    image:
+      gpu: determinedai/environments:cuda-11.3-pytorch-1.10-tf-2.8-deepspeed-0.7.0-gpu-9119094
+resources:
+  slots_per_trial: 4
```
liamcli, line 11: did you test with nodes with 2 gpus each?  otherwise not sure if setting would test model shards split across multiple nodes

examples/computer_vision/vision_transformer/det_callback.py
aciborowska: Added uploading checkpoints from multiple nodes.

examples/integrations/hf_trainer_api/adaptive.yaml
aciborowska: Not really. The reason is that the key hyperparameter here is the learning rate, and it is already set to the optimal 2e-5 value in the HF example. To make our example more convincing, I changed it to 1e-5 so that adaptive may claim the win.

examples/integrations/hf_trainer_api/adaptive.yaml
aciborowska: Switched to batches, and increased the number of trials.

examples/integrations/hf_trainer_api/deepspeed.yaml
aciborowska: Short answer: I run the deepspeed config for stage 3 on 2 nodes x 4 GPUs, and I see all files being stored in the checkpoint, and deepspeed resumes without any warnings.

Long answer: before saving the data from all nodes, deepspeed also "worked" but complained in the logs that some states (e.g., model/optimizer) could not be loaded, and then it just moved forward as if nothing happened. After saving sharded data, the warning disappeared, and when running with 4 nodes x 1 GPU, all 4 nodes said they loaded sharded data. 
Now, with 2 nodes x 4 GPUs setting, I see that local_rank=0/4 loads itself and happily announces loading 8 zero partition checkpoints (4 for model and 4 for optimizer). Also, deepspeed does not show any warnings, and logs look fine. Imo works.

examples/integrations/hf_trainer_api/image_classification.py
```diff
@@ -0,0 +1,452 @@
...
+    if len(sys.argv) == 2 and sys.argv[1].endswith(".json"):
+        # If we pass only one argument to the script and it's the path to a json file,
+        # let's parse it to get our arguments.
+        model_args, data_args, training_args = parser.parse_json_file(
+            json_file=os.path.abspath(sys.argv[1])
+        )
+    else:
+        model_args, data_args, training_args = parser.parse_args_into_dataclasses()
+
+    set_hyperparameters(training_args)
```
aciborowska, line 213: Here is where training arguments are updated with hyperparameters from the searcher. I decided to use a separate function (and not the callback) since training arguments are passed to different objects and should be updated ASAP to make sure all objects see the updated arguments. While we can do it in the callback, it would require users to initialize the callback here, which they may not do. So, a dedicated function seemed to me like a better idea, but I am open to changing it.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+        control: TrainerControl,
+        logs=None,
+        **kwargs,
+    ):
+        if state.is_world_process_zero:
+            metrics, metric_type = self._get_metrics(logs)
+            if metric_type == TRAIN:
+                if (
+                    "train_step" not in self.last_metrics
+                    or self.last_metrics["eval_step"] != state.global_step
```
aciborowska, line 76: This is done to avoid reporting training/validation metrics twice for the same global_step. AFAIK, this only happens with the **average** training/validation metrics, which, depending on the settings, may be computed for the same global step as regular (not average) metrics leading to DB errors on duplicate key.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+        import determined
+
+        self._det = determined
+        if args.deepspeed:
+            distributed = self._det.core.DistributedContext.from_deepspeed()
+        else:
+            distributed = self._det.core.DistributedContext.from_torch_distributed()
+        self.core_context = self._det.core.init(distributed=distributed)
+        self.core_context.__enter__()
+        weakref.finalize(self, exit_context, self.core_context)
```
rb-determined-ai, line 44: I think this is a bad idea.

It _does_ guarantee that core_context.close() is called, eventually.  Though if it happens during interpreter shutdown you can get some crazy confusing stacktraces (which we've seen before](https://determinedai.atlassian.net/browse/DET-2302)).

But there are other useful things that core_context does, like catching InvalidHP exceptions.  And some day, I'd like to add asynchronous metrics and/or profiling and/or tensorboard logging, in which case we'll need a finish() or flush() call.  And that finish() or flush() call will not be safe to try to run during interpreter shutdown.

I think a cleaner API would be to require the user to do something like:

```python

def my_main(core_context):
    trainer.Train(..., callbacks=[DetCallback(core_context)])

with det.core.init() as core_context:
    my_main(core_context)
```

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+    ):
+        info = self._det.get_cluster_info()
+        if info is None:
+            # TODO: modify to support local mode
+            logging.warning("ClusterInfo is None: not running in a task. Skip saving.")
+            return
+
+        local_path = os.path.join(args.output_dir, f"checkpoint-{state.global_step}")
+
+        storage_manager = self.core_context.checkpoint._storage_manager
```
rb-determined-ai, line 124: We can't publish an example that includes using internal APIs.  They are internal, so teaching users to use them is asking for trouble.

The `._storage_manager` is an internal attribute.

Can this example be written against purely public APIs?  If not, what additional APIs would need to exist for that to be possible?

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+
+def exit_context(context: det.core.Context) -> None:
+    context.__exit__(None, None, None)
+
+
+def set_hyperparameters(training_args: TrainingArguments):
+    hparams = det.get_cluster_info().trial.hparams
+    for k, v in hparams.items():
+        if hasattr(training_args, k):
+            setattr(training_args, k, v)
```
rb-determined-ai, line 330: This `hasattr()`/`setattr()` strategy is a bit too magic.  It will silently use or ignore hyperparameters based on if the name matches a TrainingArguments attribute, not based on if the user intended for that to happen.

Additionally, there's a ton of stuff that TrainingArguments does during `__init__` (and its custom `__post_init__()` method), and `setattr()` is going to mutate the TrainingArguments after all that logic is completed.

How about we take a different approach, where the user sets some hyperparameters that are explicitly named for TrainingArguments:

```yaml
hyperparameters:
  my_val: 1
  ...
  training_arguments:
    learning_rate: 0.001
    per_device_train_batch_size:
      type: int
      minval: 32
      maxval: 128
```

Then in the example the user can do something like:

```python
tr_args = TrainingArgs(fp16=true, **hparams["training_arguments"])
```

This way, if they name something wrong, it will throw an error, and it should be obvious which hyperparameters apply to TrainingArguments when looking at the config.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+            f"while HF Trainer uses {trainer_units}={trainer_len}. "
+            f"Continuing this run may cause Searcher not to behave correctly. "
+            f"Make sure to match the units between HF Trainer and Searcher: "
+            f"use (--num_train_epochs and searcher.max_length.epochs) OR "
+            f"(--max_steps and searcher.max_length.batches)."
+        )
+
+
+def is_determined_available() -> bool:
+    return importlib.util.find_spec("determined") is not None
```
rb-determined-ai, line 319: why not just 
```python
try:
    import determined as det
    have_det = True
except:
    have_det = False
```
?

aciborowska: No particular reason. I was following other Trainer callback integrations. Anyway, that will not be needed anymore after core_context is initialized in the main script.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,348 @@
...
+            shutil.copytree(local_path, path, dirs_exist_ok=True)
+
+        if self.core_context.distributed.rank == 0:
+            if self.tokenizer is not None:
+                self.tokenizer.save_pretrained(os.path.join(path, "tokenizer"))
+
+            if self.user_checkpoint_metadata is not None:
+                self._on_save_user_data(path)
+
+    def _on_save_user_data(self, save_path) -> None:
```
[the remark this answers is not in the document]
aciborowska: Good point. The plan for this callback was to (1) make it a part of public API dai integrations and (2) upstream it to HF Trainer integrations. Right now, I want to keep it as an example, and once more integrations are there, we can make the experience of integrating with eg. HF, PLT, Mosaic coherent. 
Long story short, I agree with your comment and will update the comment accordingly. We can decide on how to best deal with user data across all the integrations once we have them.

examples/integrations/hf_trainer_api/image_classification.py
```diff
@@ -0,0 +1,461 @@
...
+        model=model,
+        args=training_args,
+        train_dataset=dataset["train"] if training_args.do_train else None,
+        eval_dataset=dataset["validation"] if training_args.do_eval else None,
+        compute_metrics=compute_metrics,
+        tokenizer=feature_extractor,
+        data_collator=collate_fn,
+    )
+
+    trainer.add_callback(det_callback)
```
aciborowska, line 412: This is new compared to the original example.

examples/integrations/hf_trainer_api/image_classification.py
```diff
@@ -0,0 +1,461 @@
...
+        args = json.loads(Path(args_fname).read_text())
+        args = args.update(train_hps)
+        model_args, data_args, training_args = parser.parse_dict(args)
+    # 2. If your arguments are in a sys.argv:
+    else:
+        args = sys.argv[1:]
+        args.extend(dict2args(train_hps))
+        model_args, data_args, training_args = parser.parse_args_into_dataclasses(args)
+
+    return model_args, data_args, training_args
```
aciborowska, line 220: This is new compared to the original example.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,243 @@
...
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
+        info = det.get_cluster_info()
+        assert info
+
+        # local_path is where HF Trainer saves model and tokenizer.
+        local_path = os.path.join(args.output_dir, f"checkpoint-{state.global_step}")
```
liamcli, line 97: are there output files in the top level `output_dir` that are always generated?  would a user want to persist those in the checkpoint as well?  i.e. would we want to save any file in the top level dir along with the files in the checkpoint folder for a particular `global_step`?

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,243 @@
...
+
+        if state.is_world_process_zero:
+            if self.user_checkpoint_metadata is not None:
```
liamcli, line 100: can you provide an example usage of `user_checkpoint_metadata`?  with an overwritten `_on_save_user_data`?

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -0,0 +1,243 @@
...
+                self._on_save_user_data(local_path)
+
+        det_checkpoint_metadata = {
+            "steps_completed": state.global_step,
+            "trial_id": info.trial.trial_id,
+        }
+
+        self.core_context.checkpoint.upload(
```
liamcli, line 108: i know it's more natural to use `upload` with the hf trainer since it assumes a fixed `output_dir`.  i'm a little concerned about double copy for large models and shared_fs though.  what would using `store_path/restore_path` require and  is it worth it to use it instead of `upload/download`?

aciborowska: Let's see if I understand your question correctly. Trainer saves all files to `output_dir`. When we do `upload` for shared_fs, under the hood, we do:
```
def upload(src, dst):
    copytree(src, os.path.join(self._base_path, dst), dirs_exist_ok=True)
```

Let's assume I switch it to `store_path` which gives me the path under which I should save my files. But Trainer already saved my files locally, which means I would need to copy files from `output_dir` to the path:
```
with self.core_context.checkpoint.store_path(metadata) as (path, storage_id):
        copytree(output_dir, path, dirs_exist_ok=True)
```
The only thing to avoid that copying would be to force Trainer to save the files already under the path (not in the output_dir) via args.output_dir = path. 
I looked through Trainer code and I don't see a clean way of modifying args.output_dir dynamically with new paths via callback, especially that we would need to do in the checkpoint context to make sure it works for cloud storage managers.

examples/integrations/hf_trainer_api/det_callback.py
aciborowska: good point. `output_dir` contains two types of dirs:
* `checkpoint-step` with all files necessary for resumption
* `runs\date_ip\events.out.tfevents...`  which are events that can be visualized in the Tensorboard. It could be useful to some users and we should save it.

This means saving top-level `output_dir`. One issue is that `output_dir` contains all checkpoints which means its size will be gradually growing. Users can solve that problem for us by using `--save_total_limit N`, which limits number of checkpoints in `output_dir` to N. We cannot really solve that problem for users, since upload/store_path does not have a selector, which means we upload all content of the directory. A possible solution would be to use `store_path` and copy `runs` and selected `checkpoint_dir` manually, but I think I would rather rely on `--save_total_limit N` and copy full `output_dir` than to show users incorrect usage of `store_path` with manual copying.

coreystatendet: You can use a trick like https://github.com/determined-ai/determined/blob/42e91783665ba18746bc10bdcc02809e46912cca/examples/pytorch_lightning/core_api/integration.py#L178 to `upload` a selection of files without an additional copy by making symbolic links.

That said, maybe we should set `TensorboardMode` to `MANUAL` and upload the tensorboard artifacts that way?  That way they show up as expected in the Determined Tensorboard.  https://github.com/determined-ai/determined/blob/master/harness/determined/core/_tensorboard_mode.py

aciborowska: That's a helpful trick I haven't thought about, thanks Corey! 
Still, I am looking into adding a selector to `upload` to avoid doing that, but if I cannot make it happen within a reasonable timeline, I am going to follow your suggestion.

examples/integrations/hf_trainer_api/det_callback.py
coreystatendet: Maybe add a comment here about this issue, in case we revisit in the future?

coreystatendet (approved):
Modulo minor comments, looks good to me!

examples/integrations/hf_trainer_api/det_callback.py
aciborowska: I don't want to mud the waters in the example. I am working on a document describing closer training loop integration with HF Trainer and will include the above discussion there.

examples/integrations/hf_trainer_api/det_callback.py
```diff
@@ -176,9 +189,17 @@ def on_epoch_end(
                 self.current_op.report_progress(state.epoch)
 
             if round(state.epoch) >= self.current_op.length:
```
coreystatendet, line 191: Modulo looking into this `round` call as we discussed, looks good to me.  Nice catch.

PR #4786, opened 2022-08-15
fix: use jupyter icon in navigation sidebar [DET-7097] 

gt2345:
## Description

Use jupyter icon instead of plus icon in navigation side bar when collapse.

Before:
Screen Shot 2022-08-15 at 12 06 07 PM

After:
Screen Shot 2022-08-15 at 12 06 15 PM

## Checklist


trentwatt (approved):
jupyter

PR #4787, opened 2022-08-15
chore: promote Session to a first-class citizen

rb-determined-ai:
Originally, Session was part of the python sdk, and not used anywhere
else.  Slowly it has come to be preferred for REST API interactions,
especially now that it is used with the new generated bindings.

Most things need bindings to work, and this requirement will grow
stronger as we port more and more code to use the generated bindings.
Bindings needs Session to work.  Session needs api.requests and
api.authentication to work.  This is all a big mess unless the importing
of Session is carefully controlled.

This PR promotes Session to be part of determined.common.api, with a
controlled import order to prevent failed circular imports.  Python
SDK-related import paths just point at the real definition now.

Additionally, this renames the old api.authentication.Session class to
UserTokenPair, to reduce confusion.

### DIFF
--- a/harness/determined/common/__init__.py
+++ b/harness/determined/common/__init__.py
@@ -4,5 +4,6 @@ except ModuleNotFoundError:
     # Inexplicably, sometimes ruamel.yaml is pacakged as ruamel_yaml instead.
     import ruamel_yaml as yaml  # type: ignore
 
-from . import api, check, constants, context, requests, storage, util
-from ._logging import set_logger
+from determined.common import util
+from determined.common import api, check, constants, context, requests, storage
+from determined.common._logging import set_logger
--- a/harness/determined/common/api/_session.py
+++ b/harness/determined/common/api/_session.py
@@ -1,5 +1,3 @@
-import json
-import os
 from typing import Any, Dict, Optional
 
 import requests
@@ -7,34 +5,6 @@ import urllib3
 
 from determined.common import util
 from determined.common.api import authentication, certs, request
-
-
-def get_max_retries_config() -> urllib3.util.retry.Retry:
-    # Allow overriding retry settings when necessary.
-    # `DET_RETRY_CONFIG` env variable can contain `urllib3` `Retry` parameters,
-    # encoded as JSON.
-    # For example:
-    #  - disable retries: {"total":0}
-    #  - shorten the wait times {"total":10,"backoff_factor":0.5,"method_whitelist":false}
-
-    config_data = os.environ.get("DET_RETRY_CONFIG")
-    if config_data is not None:
-        config = json.loads(config_data)
-        return urllib3.util.retry.Retry(**config)
-
-    # Defaults.
-    try:
-        return urllib3.util.retry.Retry(
-            total=20,
-            backoff_factor=0.5,
-            allowed_methods=False,
-        )
-    except TypeError:  # Support urllib3 prior to 1.26
-        return urllib3.util.retry.Retry(
-            total=20,
-            backoff_factor=0.5,
-            method_whitelist=False,  # type: ignore
-        )
 
 
 class Session:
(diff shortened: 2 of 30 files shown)


harness/determined/common/api/_session.py
```diff
@@ -9,34 +7,6 @@
 from determined.common.api import authentication, certs, request
 
 
-def get_max_retries_config() -> urllib3.util.retry.Retry:
```
rb-determined-ai, line 12: moved to common.util

harness/determined/common/util.py
```diff
@@ -141,3 +144,25 @@ def get_config_path() -> pathlib.Path:
...
+        config = json.loads(config_data)
+        return urllib3.util.retry.Retry(**config)
+
+    # Defaults.
+    try:
+        kwargs = {"allowed_methods": False}
+        return urllib3.util.retry.Retry(total=20, backoff_factor=0.5, **kwargs)
+    except TypeError:  # Support urllib3 prior to 1.26
+        kwargs = {"method_whitelist": False}
+        return urllib3.util.retry.Retry(total=20, backoff_factor=0.5, **kwargs)
```
rb-determined-ai, line 168: rewritten from before to use kwargs, so as to avoid possibility of `unused #type: ignore comments`

harness/determined/common/__init__.py
```diff
@@ -4,5 +4,6 @@
     # Inexplicably, sometimes ruamel.yaml is pacakged as ruamel_yaml instead.
     import ruamel_yaml as yaml  # type: ignore
 
-from . import api, check, constants, context, requests, storage, util
-from ._logging import set_logger
+from determined.common import util
```
mpkouznetsov, line 7: is this separate from the line below to make sure it is imported before all the other things (`api`, `check`, `constants` etc.) ?

harness/determined/common/api/authentication.py
```diff
@@ -32,7 +32,7 @@ def salt_and_hash(password: str) -> str:
         return password
 
 
-class Session:
+class UsernameTokenPair:
```
mpkouznetsov, line 35: 👍

mpkouznetsov (approved):

harness/determined/common/__init__.py
rb-determined-ai: yes, since `common.api._session` imports `common.util`.

Long-term, I want to redo our internal code layout to reflect the more opinionated public api layout that we have.  I'm thinking something like:

```python
determined.internal                 # no symbols here, maybe?
determined.internal.bindings
determined.internal.constants       # only one constants module!
determined.internal.util            # only one util module!
determined.internal.request         # unroll common.api
determined.internal.authentication  # ...

# OR maybe don't unroll common.api, but eliminate the submodules from import paths
determined.internal.api._bindings   # exposed as api.*
determined.internal.api._request    # exposed as api.*
# ...
```

This would also make it crystal clear to users what counts as internal and our own import paths would be a lot less confusing.

harness/determined/common/experimental/_session.py
```diff
@@ -0,0 +1,7 @@
+# link for backwards compatibility, since this was the original place that Session was exposed.
```
rb-determined-ai, line 1: note to self: this should not have been renamed to `_session`, that would be a breaking change.

rb-determined-ai: Actually, I will do this in a separate PR.  If I do it in this PR, it will get squished into one commit and git will decide that I rewrote this file instead of moved it elsewhere and replaced with a new one, which would break `git blame` on that file.

harness/determined/common/util.py
rb-determined-ai: (this didn't pass linting in CI so I undid it)

PR #4790, opened 2022-08-15
perf: reduce user settings api call

keita-determined:
## Description

- [DET-8051]
- 139 calls -> 8 calls in `Uncategorized` page on my local (94% reduction) (It depends on the page)

## Test Plan

- Check if user setting api call is reduced, especially in `Uncategorized` page
- Check if user setting works as expected

## Checklist

[DET-8051]: https://determinedai.atlassian.net/browse/DET-8051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/hooks/useSettings.ts
+++ b/webui/react/src/hooks/useSettings.ts
@@ -207,6 +207,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions): 
   const storage = useStorage(options?.storagePath || config.storagePath);
   const { auth: { user } } = useStore();
   const prevSearch = usePrevious(location.search, undefined);
+  const prevUser = usePrevious(user, user);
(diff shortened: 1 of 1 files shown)


webui/react/src/hooks/useSettings.ts
```diff
@@ -315,7 +316,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
   }, [ config.settings, updateSettings ]);
 
   const fetchUserSetting = useCallback(async () => {
-    if (!user) return;
+    if (!user || user === prevUser) return;
```
keita-determined, line 319: `getUserSetting` is called based on `user.id`, so check if `user` is the same or not. If its the same user, return early in function.

gt2345 (approved):
Looks good, thank you for fixing it.

PR #4794, opened 2022-08-16
fix: grid view on Workspaces and Projects pages show all items [DET-8031]

EmilyBonar:
## Test Plan
On the Workspaces List page all workspaces should be shown in grid view, even if the table limit in list view is smaller than the total number. Same for the Projects List/Workspace Details page.

## Checklist


johnkim-det (commented):
limit fix looks good, i think there's still an issue when you switch back to Grid View after choosing page >1 on List View. assuming it just needs the same change for the `offset` field.

EmilyBonar:
> limit fix looks good, i think there's still an issue when you switch back to Grid View after choosing page >1 on List View. assuming it just needs the same change for the `offset` field.

Good catch, should be fixed now.

johnkim-det (approved):
lgtm

PR #4796, opened 2022-08-16
fix: get trial datapoints from trial comparison/summarization endpoint

mapmeld:
## Description

Use the trial-comparison / trial-summarization endpoint to get downsampled metrics from an experimental run. As intended back in #4300 and missed in #4703 .

## Test Plan

Single-trial experiments should have an experiment graph with metrics added/removed

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails/TrialChart.tsx
+++ b/webui/react/src/pages/TrialDetails/TrialChart.tsx
@@ -10,9 +10,10 @@ import UPlotChart, { Options } from 'components/UPlot/UPlotChart';
 import { tooltipsPlugin } from 'components/UPlot/UPlotChart/tooltipsPlugin';
 import { trackAxis } from 'components/UPlot/UPlotChart/trackAxis';
 import css from 'pages/TrialDetails/TrialChart.module.scss';
+import { compareTrials } from 'services/api';
 import Spinner from 'shared/components/Spinner';
 import { glasbeyColor } from 'shared/utils/color';
-import { MetricName, MetricType, Scale, WorkloadGroup } from 'types';
+import { MetricContainer, MetricName, Scale } from 'types';
 
 interface Props {
   defaultMetricNames: MetricName[];
@@ -21,7 +22,6 @@ interface Props {
   metrics: MetricName[];
   onMetricChange: (value: MetricName[]) => void;
   trialId?: number;
-  workloads?: WorkloadGroup[];
 }
 
 const getChartMetricLabel = (metric: MetricName): string => {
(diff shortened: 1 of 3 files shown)


webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -35,31 +37,53 @@ const TrialChart: React.FC<Props> = ({
...
-  workloads,
   trialId,
+  trialTerminated,
 }: Props) => {
   const [ scale, setScale ] = useState<Scale>(Scale.Linear);
+  const [ trialSumm, setTrialSummary ] = useState<MetricContainer[]>([]);
+
+  const fetchTrialSummary = useCallback(async () => {
+    if (trialId) {
+      const summ = await compareTrials({
```
trentwatt, line 48: question: why this instead of summarizeTrials?

mapmeld: Your call as part of the trial-comparison team.  Selected compareTrials in the original code in case you're doing multi-trial comparison here in the future

webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -35,31 +37,53 @@ const TrialChart: React.FC<Props> = ({
...
+  }, [ metricNames, scale, trialId ]);
+
+  const { stopPolling } = usePolling(fetchTrialSummary, { interval: 2000, rerunOnNewFn: true });
+  useEffect(() => {
+    if (trialTerminated) {
+      stopPolling();
+    }
+  }, [ trialTerminated, stopPolling ]);
+  if (trialTerminated) {
+    stopPolling();
```
trentwatt, line 66: so i guess this extra one is needed in case you start polling when the trial is in terminal state?

trentwatt, line 66: so i guess this one is needed in case you start polling when the trial is terminated?

webui/react/src/pages/TrialDetails/TrialChart.tsx
trentwatt, line 48: question: why this instead of `summarizeTrials`?

I guess maybe `summarizeTrials` is redundant and we could get rid of it?

trentwatt (approved):

webui/react/src/pages/TrialDetails/TrialChart.tsx
```diff
@@ -35,31 +37,53 @@ const TrialChart: React.FC<Props> = ({
...
+  }
 
   const chartData: AlignedData = useMemo(() => {
     const xValues: number[] = [];
     const yValues: Record<string, Record<string, number | null>> = {};
```
trentwatt, line 71: i feel like this could use some type asiases

PR #4797, opened 2022-08-16
fix: job queue experiment restore

hamidzr:
## Description

the k8 resource manager's actor was missing a handler for RecoverJobPosition msg

## Test Plan

reproduced the issue through minikube and manually setting q_position to -1
  - the job positions does get defaulted to 0 somewhere I'm not sure where. in the customer's case they might have ran an older migration where it had translated into -1
  - `for i in $(seq 1 4); do det e create ./examples/gan/gan_mnist_pytorch/const.yaml ./examples/gan/gan_mnist_pytorch/&; done`
  - `update jobs set q_position=-1;`
repeated the steps in this branch and the resource manager did not crash

DeepinScreenshot_Alacritty_20220823130720

## Checklist

### DIFF
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -214,7 +214,7 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 				return err
 			}
 
-			if j.QPos.GreaterThanOrEqual(decimal.Zero) {
+			if j.QPos.GreaterThan(decimal.Zero) {
 				e.rm.RecoverJobPosition(ctx, sproto.RecoverJobPosition{
 					JobID:        e.JobID,
 					JobPosition:  j.QPos,
--- a/master/internal/experiment.go
+++ b/master/internal/experiment.go
@@ -214,7 +214,7 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 				return err
 			}
 
-			if !j.QPos.Equals(decimal.Zero) {
+			if j.QPos.GreaterThanOrEqual(decimal.Zero) {
 				e.rm.RecoverJobPosition(ctx, sproto.RecoverJobPosition{
 					JobID:        e.JobID,
 					JobPosition:  j.QPos,
--- a/docs/release-notes/experiment-restore.txt
+++ b/docs/release-notes/experiment-restore.txt
@@ -1,0 +1,6 @@
+:orphan:
+
+**Bug Fixes**
+
+-  Fix an issue where restoring a job in into a Kubernetes set up could crash the
+   resource manager.
--- a/master/internal/rm/job.go
+++ b/master/internal/rm/job.go
@@ -12,6 +12,8 @@ import (
 	"github.com/determined-ai/determined/proto/pkg/jobv1"
 )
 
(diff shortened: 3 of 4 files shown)


master/internal/experiment.go
```diff
@@ -214,7 +214,7 @@ func (e *experiment) Receive(ctx *actor.Context) error {
 				return err
 			}
 
-			if !j.QPos.Equals(decimal.Zero) {
+			if j.QPos.GreaterThanOrEqual(decimal.Zero) {
```
hamidzr, line 217: the jobs already get the zero position by default so maybe we don't need to send this msg if the position is at `0`

ioga: yes I'd say this should be strictly greater than 0

hamidzr: I was thinking what if the receiving actor hooks onto this message for whatever other reason or if more information ends up being carried with this message later? wouldn't it be more expected for the msg to be sent considering that we won't get 0 q position as a default value?

hamidzr:
this should fix the issue but I need to repro the problem then verify that this fixes it and for that I need a k8 setup which will  take me some time. should I be able to use `det deploy` and google cloud credentials to sping one up?

ioga:
- there's `det deploy gke-experimental` but it's quite experimental
- you can run it locally on minikube using scripts and helm chart patch from @stoksc

hamidzr:
I've got determined running in minikube using cpu slots but I haven't figured 
- how do I simulate master going down and coming back up while keeping the db data persisted using minikube
  - using a persisted volume?
- how do I deploy an unpublished version

ioga:
1. you could kubectl exec into master pod and kill -9 master process, or kill the pod.
2. minikube pulls images from local docker registry. you can `make package` in master, and then specify this version in the helm chart.

hamidzr:
@eecsliu any thoughts on treating 0 as special value vs not?
also something  I noticied  when testing this was that the jobs end up getting defaulted to 0 instead of -1 maybe as a result of changes to the job db model?

ioga (approved):

eecsliu:
> @eecsliu any thoughts on treating 0 as special value vs not? also, something I noticed when testing this was that the jobs end up getting defaulted to 0 instead of -1 maybe as a result of changes to the job db model? -1 wold have only happened if through our db handler we didnt' provide a position which could have happened for earlier versions of job management? I have to dig into the code history to figure that out

Do we mean 0 vs -1? I don't think the treatment of position 0 would differ from position -1 since we're strictly looking for > 0 positions. I don't have an issue switching, as long as we have a way of representing invalid or position-less values

hamidzr:
> since we're strictly looking for > 0 positions

hmm, that check used to be != 0 if you're referring to the one that's changed in this pr.
does that still sound good?

eecsliu:
Looking at the k8s scheduler code again, it looks like that uses 0 as the magic value. I think we should keep it 0 for now and ticket it for next release? I'm not sure about changing the k8s scheduler code so close to release.

eecsliu (approved):
lgtm!

PR #4801, opened 2022-08-16
chore: fix ListValue types in swagger spec

rb-determined-ai:
Protobuf google.protobuf.ListValue types come through swagger as:

    {"type":"array", "items":{"type":"object"} }

which would convert to `Sequence[Dict[str, Any]]`, which is a pretty
strange field type for v1PatchModel.labels.

This PR adds tweaks via the proto/patches/api.json to force the right
type in the swagger.


hamidzr (commented):
~I'm worried about what happens when the endpoint or the proto definitions change :thinking_face:  would people realize that their changes didn’t result in generating the change they expected?~

update: the JSON patch seems to so precisely target the type of the label that I’m not worried

```
hbp ➜ determined git:(master) jq '.definitions.v1PatchExperiment' proto/build/swagger/determined/api/v1/api.swagger.json > /tmp/before.json
hbp ➜ determined git:(master) colordiff /tmp/before.json /tmp/after.json
16c16
<         "type": "object"
---
>         "type": "string"
hbp ➜ determined git:(master)
```

looking at the rest of changes

hamidzr (approved):
lgtm

PR #4802, opened 2022-08-17
chore: proto build should fail on first error.

ioga:
## Description

Small QoL improvement to check errors in the loop and fail faster.

## Test Plan

Successful builds should not be affected; failed builds should have shorter and more readable error output.

## Commentary (optional)

I won't be surprised there's some bash or makefile magic to achieve this more efficiently.

## Checklist

### DIFF
--- a/proto/Makefile
+++ b/proto/Makefile
@@ -41,7 +41,7 @@ build/proto.stamp: $(source_files)
 	rm -rf build/proto pkg
 	mkdir -p build/proto
 	# Protobuf generation.
-	for source in $(source_files) ; do protoc -I src "$$source" --go_out=plugins=grpc:build/proto ; test $$? -eq 0 || exit 1; done
+	for source in $(source_files) ; do protoc -I src "$$source" --go_out=plugins=grpc:build/proto || exit 1; done
 	# GRPC generation.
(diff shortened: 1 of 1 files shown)


proto/Makefile
```diff
@@ -41,7 +41,7 @@ build/proto.stamp: $(source_files)
 	rm -rf build/proto pkg
 	mkdir -p build/proto
 	# Protobuf generation.
-	for source in $(source_files) ; do protoc -I src "$$source" --go_out=plugins=grpc:build/proto ; done
+	for source in $(source_files) ; do protoc -I src "$$source" --go_out=plugins=grpc:build/proto ; test $$? -eq 0 || exit 1; done
```
rb-determined-ai, line 44: why not:
```
 	... do protoc -I src "$$source" --go_out=plugins=grpc:build/proto || exit 1; done
```
instead of
```
 	... do protoc -I src "$$source" --go_out=plugins=grpc:build/proto; test $$? -eq 0 || exit 1; done
```

rb-determined-ai (approved):

proto/Makefile
ioga: that's exactly the kind of code review this PR needed 🙇

PR #4806, opened 2022-08-17
chore: add authz on moving experiments between projects [DET-7750]

mapmeld:
## Description

My goal is to add permissions on moving experiments and incorporate it into the authz interface

- [Already required] Destination project must be visible to the user (through `GetProjectByID`)
- [Already required] Destination project cannot be archived
- Moved experiment must be visible to the user (runs through `getExperiment`, which doesn't use authz yet, but will)
- Experiment's source project must be visible to the user (through `GetProjectByID`)
- Call new `CanMoveProjectExperiments` authz

Conditions to allow `CanMoveProjectExperiments`:
- admin user
- immutable source project [not quite right, but tl;dr I don't want to block users from withdrawing experiments from their default Uncategorized project]
- user with ownership of both source and destination projects [this doesn't seem right, maybe it should be user owns experiment, and we already have restrictions on viewable source and destination?]
 
## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/determined-ai/determined/master/internal/project"
 	"github.com/determined-ai/determined/master/internal/prom"
 	"github.com/determined-ai/determined/master/internal/sproto"
 
@@ -1583,13 +1584,31 @@ func (a *apiServer) MoveExperiment(
 	if err != nil {
 		return nil, err
 	}
-	p, err := a.GetProjectByID(req.DestinationProjectId, *curUser)
+
+	// check that user can view destination project
+	destProject, err := a.GetProjectByID(req.DestinationProjectId, *curUser)
 	if err != nil {
 		return nil, err
 	}
-	if p.Archived {
+	if destProject.Archived {
 		return nil, errors.Errorf("project (%v) is archived and cannot add new experiments.",
 			req.DestinationProjectId)
+	}
+
+	// get experiment info
+	exp, err := a.getExperiment(int(req.ExperimentId))
+	if err != nil {
+		return nil, err
+	}
+
+	// check that user can view source project
+	srcProject, err := a.GetProjectByID(exp.ProjectId, *curUser)
+	if err != nil {
+		return nil, err
+	}
+	if err = project.AuthZProvider.Get().CanMoveProjectExperiments(*curUser, srcProject,
+		destProject); err != nil {
+		return nil, status.Error(codes.PermissionDenied, err.Error())
 	}
 
 	holder := &experimentv1.Experiment{}
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -1614,7 +1614,7 @@ func (a *apiServer) MoveExperiment(
 			srcProject.Id)
 	}
 
-	if err = project.AuthZProvider.Get().CanMoveProjectExperiments(*curUser, srcProject,
+	if err = project.AuthZProvider.Get().CanMoveProjectExperiments(*curUser, exp, srcProject,
 		destProject); err != nil {
 		return nil, status.Error(codes.PermissionDenied, err.Error())
 	}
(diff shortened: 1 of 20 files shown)


master/internal/project/authz_basic_impl.go
```diff
@@ -96,6 +96,17 @@ func (a *ProjectAuthZBasic) CanMoveProject(
 	return nil
 }
 
+// CanMoveProjectExperiments returns an error if the user isn't a admin or owner of a project.
+func (a *ProjectAuthZBasic) CanMoveProjectExperiments(
+	curUser model.User, from, to *projectv1.Project,
+) error {
+	if !curUser.Admin && !from.Immutable && ((curUser.ID != model.UserID(from.UserId)) ||
```
NicholasBlaskey, line 103: This ```!from.Immutable``` makes it so any user can move from the default project right?

That makes sense to me just want to make sure I understand

NicholasBlaskey (approved):
Nice work looks good to me!

master/internal/project/authz_basic_impl.go
mapmeld: yes - I am asking in SaaS how they implement the default project to see whether this is the right way to reference it

master/internal/api_experiment.go
```diff
@@ -1583,15 +1584,38 @@ func (a *apiServer) MoveExperiment(
...
+	srcProject, err := a.GetProjectByID(exp.ProjectId, *curUser)
+	if err != nil {
+		return nil, err
+	}
+	if srcProject.Archived {
+		return nil, errors.Errorf("project (%v) is archived and cannot have experiments moved from it.",
+			srcProject.Id)
+	}
+
+	if err = project.AuthZProvider.Get().CanMoveProjectExperiments(*curUser, srcProject,
```
gt2345, line 1614: nit: can we handle the check experiment, check project logic inside of `CanMoveProjectExperiments`

master/internal/project/authz_basic_impl.go
```diff
@@ -96,6 +96,17 @@ func (a *ProjectAuthZBasic) CanMoveProject(
...
+		(curUser.ID != model.UserID(to.UserId))) {
```
NicholasBlaskey, line 104: Allowing someone who only owns the destination project allows anyone to make a new destination project and "steal" someones experiment from any project

Though I think the current protection was nothing so this might be fine?

NicholasBlaskey (approved):

master/internal/project/authz_basic_impl.go
mapmeld: I will rewrite this in a second since the logic is confusing in the negative; I do mean to require from and to projects to be from the current UserId

master/internal/api_experiment.go
mapmeld: I had the same question when I was looking at authz, so I based it on `CanMoveProject`, where all checks except happen in api_project.go except for the admin/user-id

gt2345 (approved):
WebUI part looks great to me

master/internal/project/authz_basic_impl.go
```diff
@@ -96,6 +96,17 @@ func (a *ProjectAuthZBasic) CanMoveProject(
 	return nil
 }
 
+// CanMoveProjectExperiments returns an error if the user isn't a admin or owner of a project.
+func (a *ProjectAuthZBasic) CanMoveProjectExperiments(
+	curUser model.User, from, to *projectv1.Project,
+) error {
+	if curUser.Admin || from.Immutable || ((model.UserID(from.UserId) == curUser.ID) &&
+		(model.UserID(to.UserId) == curUser.ID)) {
```
ioga, line 104: >user with ownership of both source and destination projects [this doesn't seem right, maybe it should be user owns experiment, and we already have restrictions on viewable source and destination?]

for me it seems that "user is admin or owns experiment" would be a better check for our OSS setup, because it mirrors what we do e.g. for experiment deletion.
maybe something worth running by product team?

mapmeld: talked with product; have added a restriction here (webui) and in backend that the experiment can be moved under one of these two conditions:
- admin user
- experiment owner who is moving from the Uncategorized project to their project, or between two of their own projects

PR #4808, opened 2022-08-17
fix: remove under the given event id for correct resumption in client.

nrajanee:
## Description

fix: remove under the given event id for correct resumption in client.


mpkouznetsov (approved):
LGTM

PR #4813, opened 2022-08-17
feat: GetExperiments to bun

NicholasBlaskey:
## Description

Converts ```GetExperiments``` to a bun query keeping the same behavior.

## Test Plan

Integration tests (didn't see the existing integration for ```get_experiments.sql``` test until the end but I think the new integration test should cover all old cases)

Manual check of ```det e list``` and options along with viewing webui page for listing experiments

## Commentary (optional)

Benchmark results

existing code benchmark
```
goos: linux
goarch: amd64
pkg: github.com/determined-ai/determined/master/internal
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkGetExeriments50-8           183           6489682 ns/op
BenchmarkGetExeriments250-8           40          28504759 ns/op
BenchmarkGetExeriments500-8           20          55002471 ns/op
BenchmarkGetExeriments2500-8           4         269186241 ns/op
PASS
ok      github.com/determined-ai/determined/master/internal     16.378s
```

After bun migration benchmark
```
goos: linux
goarch: amd64
pkg: github.com/determined-ai/determined/master/internal
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkGetExeriments50-8           320           3664846 ns/op
BenchmarkGetExeriments250-8           62          16916158 ns/op
BenchmarkGetExeriments500-8           36          31944340 ns/op
BenchmarkGetExeriments2500-8           7         151448986 ns/op
PASS
ok      github.com/determined-ai/determined/master/internal     11.858s
```

After using bun models
```
goos: linux
goarch: amd64
pkg: github.com/determined-ai/determined/master/internal
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkGetExeriments50-8            51          20587772 ns/op
BenchmarkGetExeriments250-8           12          93092094 ns/op
BenchmarkGetExeriments500-8            6         185193340 ns/op
BenchmarkGetExeriments2500-8           2         929198592 ns/op
PASS
ok      github.com/determined-ai/determined/master/internal     13.666s
```

## Checklist


master/internal/api_experiment.go
```diff
@@ -235,26 +237,54 @@ func (a *apiServer) deleteExperiment(exp *model.Experiment, user *model.User) er
...
+		ColumnExpr(protoStateDBCaseString(experimentv1.State_value, "e.state", "state", "STATE_")).
+		Column("e.archived").
+		ColumnExpr(
+			"(SELECT COUNT(*) FROM trials t WHERE e.id = t.experiment_id) AS num_trials").
+		// Intentionally not sending trial_ids due to performance.
+		ColumnExpr("COALESCE(u.display_name, u.username) as display_name").
+		ColumnExpr("e.owner_id as user_id").
+		Column("u.username").
+		ColumnExpr("e.config->'resources'->>'resource_pool' AS resource_pool").
+		ColumnExpr("e.config->'searcher'->>'name' AS searcher_type").
```
NicholasBlaskey, line 271: This was ```->'name'``` in the original query but got back names quoted so switched to ```->>'name'```

master/internal/api_experiment_intg_test.go
```diff
@@ -0,0 +1,374 @@
...
+		RawSingleConfig: &expconf.SingleConfig{
+			&expconf.Length{Units: 10, Unit: "batches"},
+		},
+	},
+}
+
+func TestGetExperiments(t *testing.T) {
+	// Setup.
+	api, _, ctx := SetupAPITest(t)
+	_, err := db.Bun().NewTruncateTable().Table("experiments").Cascade().Exec(ctx)
```
NicholasBlaskey, line 54: This will delete all of your experiments when running an integration test which may or may not be okay?

Not sure how to get around this besides like always filtering on some random field

Tried renaming ```experiments``` then copying it but that ran into a bunch of other issues like the IDs not generating then not being able to create trials to test with

master/internal/api_experiment.go
```diff
@@ -235,26 +237,54 @@ func (a *apiServer) deleteExperiment(exp *model.Experiment, user *model.User) er
...
+		Column("e.id").
+		ColumnExpr("e.config->>'description' AS description").
+		ColumnExpr("e.config->>'labels' AS labels").
+		ColumnExpr("proto_time(e.start_time) AS start_time").
+		ColumnExpr("proto_time(e.end_time) AS end_time").
+		ColumnExpr(protoStateDBCaseString(experimentv1.State_value, "e.state", "state", "STATE_")).
+		Column("e.archived").
+		ColumnExpr(
+			"(SELECT COUNT(*) FROM trials t WHERE e.id = t.experiment_id) AS num_trials").
+		// Intentionally not sending trial_ids due to performance.
```
stoksc, line 266: i'm not sure how trial ids would kill us in perf but `SELECT COUNT(*) FROM trials t WHERE e.id = t.experiment_id) AS num_trials"` would not.

master/internal/api_experiment.go
```diff
@@ -235,26 +237,54 @@ func (a *apiServer) deleteExperiment(exp *model.Experiment, user *model.User) er
...
+	resp := &apiv1.GetExperimentsResponse{Experiments: []*experimentv1.Experiment{}}
+	query := db.Bun().NewSelect().
+		Model(&resp.Experiments).
+		ModelTableExpr("experiments as e").
+		Column("e.id").
+		ColumnExpr("e.config->>'description' AS description").
+		ColumnExpr("e.config->>'labels' AS labels").
+		ColumnExpr("proto_time(e.start_time) AS start_time").
+		ColumnExpr("proto_time(e.end_time) AS end_time").
+		ColumnExpr(protoStateDBCaseString(experimentv1.State_value, "e.state", "state", "STATE_")).
```
stoksc, line 262: thoughts on converting this to bun meaning: make a better `type Experiment struct` bun model type and a `.ToProto` for a lot of this to protoing in the query. as opposed to just using bun as a query builder?

master/internal/api_experiment.go
NicholasBlaskey: The one difference I can think of is that sending trial IDs could potentially be a significant amount of IDs we need to serialize / deserialize / send over the network. 

I would be worried that it could cause someones setup to hang. Though I am honestly not sure.

stoksc: for even the largest hp searches i think this would be ~8kb max. so i guess i was thinking that the join cost would dwarf the network io. maybe not though.

NicholasBlaskey: It would be cool to have some level of performance testing for situations like this (maybe expected SLAs too?).

Though I honestly have no idea how to in the "bun" ORM way to actually get a count of something without actually getting it so I added trial IDs.

master/internal/api_experiment.go
NicholasBlaskey: Ok tried converting to using more of the Bun ORM features. Conflicted if it is even better.

For experiment config specifically we need to serialize into ```expconf``` then we deserialize into JSON bytes to then serialize into a proto ```struct``` which we then serialize to json where the webui / CLI will deserialize it again.

It is significantly slower than going directly to proto (~80% for 25 experiments ~500% for 2500 experiments). Which might be fine if it provides enough value. 

Beyond that it is just a little awkward at times. For ordering by trial ID count I need to basically query trial IDs once for bun and once for the order statement since Bun splits getting Trial IDs into its own query.

master/internal/api_experiment.go
```diff
@@ -289,23 +285,109 @@ func (a *apiServer) GetExperiments(
...
+	if offset > total || offset < -total {
+		startIndex = total
+	} else if offset < 0 {
+		startIndex = total + offset
+	}
+
+	endIndex := startIndex + limit
+	switch {
+	case limit == -2:
+		endIndex = startIndex
```
ioga, line 368: where is this thing implemented in the current code? why do we even need it?

ioga: thanks 👍 
it uses it to get the number of active experiments per cluster https://github.com/determined-ai/determined/blob/master/webui/react/src/hooks/useFetch.ts#L25

master/internal/api_experiment.go
stoksc: > (~450% for 25 experiments ~500% for 2500 experiments)

what precisely is this comparing (response times for the entire endpoint?) and it is 20 -> 100 or like 100 -> 500 ms?

NicholasBlaskey: response time for the proto function (so ignoring the proto -> json API layer / network) 

50 experiments -> 4ms vs 21ms
250 experiments -> 17ms vs 93ms
2500 experiments -> 151ms vs 929ms

I don't want to like prematurely optimize here but just somewhat concerned that this change could cause a noticeable slowdown on someones cluster (especially since the benchmark I did had no trials and a relatively minimal experiment config)

Going directly proto just using bun as a query builder avoids this but does repeat a lot of the mistakes that ```QueryProtof``` has so I don't know.

NicholasBlaskey: Talked offline to take the approach of using bun just a query builder for this case

Follow up ticket
https://determinedai.atlassian.net/browse/DET-8203

master/internal/api_experiment.go
stoksc: totally agree, we should add this as a tech debt item (it even came up this release, too)

master/internal/api_experiment.go
```diff
@@ -290,23 +321,100 @@ func (a *apiServer) GetExperiments(
...
+		query = query.Where("e.config->>'name' ILIKE ('%%' || ? || '%%')", req.Name)
+	}
+	if len(req.Labels) > 0 {
+		// In the event labels were removed, if all were removed we insert null,
+		// which previously broke this query.
+		query = query.Where(`string_to_array(?, ',') <@ ARRAY(SELECT jsonb_array_elements_text(
+				CASE WHEN e.config->'labels'::text = 'null'
+				THEN NULL
+				ELSE e.config->'labels' END
+			))`, strings.Join(req.Labels, ",")) // Trying bun.In doesn't work.
```
stoksc, line 340: what was up with bun.In?

master/internal/api_experiment.go
```diff
@@ -290,23 +321,100 @@ func (a *apiServer) GetExperiments(
...
+		if total < endIndex {
+			endIndex = total
+		}
+	case startIndex+limit > total:
+		endIndex = total
+	}
+
+	// Add start and end index to query.
+	query.Offset(startIndex)
+	query.Limit(endIndex - startIndex)
```
stoksc, line 409: maybe just move the 'maybe bun bug with limit' handling to here. i kinda prefer not mutating the bun query in here, maybe call it `apply...` if we do.

stoksc (approved):
lgtm, just a few random comments

master/internal/api_experiment.go
NicholasBlaskey: Was hoping you could do something like ```? <@ ARRAY``` but I needed to have this strings.Join instead

PR #4814, opened 2022-08-17
fix: use fluent version 1.9.3 everywhere by default.

ioga:
## Description

fluent 1.9.3 was used as a default in agent config itself (making it a default for static agent deployments), but master-side provisioner and k8s configs still use 1.6. 
Rectify this situation by avoiding copypasta and moving the default to a variable.

## Commentary (optional)

We've been using 1.9.3 for static agents for a while now, but [this OSS message](https://determined-community.slack.com/archives/CV3MTNZ6U/p1660742892630999?thread_ts=1654782582.474989&cid=CV3MTNZ6U) is slightly disconcerting.

## Checklist


agent/cmd/determined-agent/run.go
```diff
@@ -145,7 +145,7 @@ func newRunCmd() *cobra.Command {
 		"Addresses that the agent's containers should not proxy")
 
 	// Logging flags.
-	cmd.Flags().StringVar(&opts.Fluent.Image, "fluent-image", "fluent/fluent-bit:1.9.3",
+	cmd.Flags().StringVar(&opts.Fluent.Image, "fluent-image", aproto.FluentImage,
```
NicholasBlaskey, line 148: Just wondering why historically we have ```fluent-image``` as a per image config rather than as a master config option? Is there a use case for different ```fluent-image``` per different agent?

NicholasBlaskey (approved):
Looks good to me!

agent/cmd/determined-agent/run.go
ioga: static agents don't get this value from master but from their own config. master option is just for the agents spun up by provisioner.
in retrospect this could be driven by master, but I'd rather move fluent into task containers (which is a tech debt project) rather than rework this.

PR #4815, opened 2022-08-17
fix: add charactor limit for name and description fields

keita-determined:
## Description

[DET-8073]

## Test Plan

- Check if the char limit validation works in the fields

## Commentary (optional)

- If data already surpass the limit, the surpassed characters are not visible but still in db.  That would be nice to know if users have long names or descriptions in general. If so, we can increase the limit

## Checklist

[DET-8073]: https://determinedai.atlassian.net/browse/DET-8073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/constants/values.ts
+++ b/webui/react/src/constants/values.ts
@@ -1,0 +1,5 @@
+export const CharLength = {
+  Limit32: 32,
+  Limit64: 64,
+  Limit512: 512,
+} as const;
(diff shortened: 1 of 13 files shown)


keita-determined:
This solution is simplistic solution so close

PR #4816, opened 2022-08-18
chore: deprecate mmdetection example in favor of model-hub version

liamcli:
## Description
This PR removes the mmdetection example under the examples directory and points users to the model-hub implementation instead.  We've had users go to the deprecated example implementation a few times now that encountered issues that would not have arisen with the model-hub implementation.
## Checklist


model_hub/examples/mmdetection/maskrcnn.yaml
```diff
@@ -0,0 +1,50 @@
...
+searcher:
+  name: single
+  metric: bbox_mAP
+  max_length:
+    batches: 87960
+  smaller_is_better: false
+max_restarts: 5
+environment:
+  image:
+    gpu: determinedai/model-hub-mmdetection:0.17.15
```
varlogtim, line 43: This is hardcoded to a version. Needs an update.

varlogtim (approved):
Approve after the hardcoded version is fixed.

interactivetech (approved):
lgtm!

PR #4818, opened 2022-08-18
fix: correctly display nested categorical hyperparameters [DET-8074]

EmilyBonar:
## Description
From ticket:

> To repro, run a hp search experiment where one of parameters is categorical, e.g. as suggested in the thread:
> ```
> scheduler:
>       type: categorical
>       vals:
>         - _type: "WarmupDecayLR"
>           params:
>             warmup_min_lr: 0
>             warmup_num_steps: 100
>             total_num_steps: 1841
>             warmup_type: "linear"
>         - _type: "WarmupLR"
>           params:
>             warmup_min_lr: 0
>             warmup_num_steps: 100
>             warmup_type: "linear"
> ```
> Issue:
- >  In visualization → learning curve UI the parameter value is displayed as “undefined”.
- >  In visualization → hp parallel coordinates it is displayed as “[object Object]”

## Test Plan
Try to repro, parameter values should show up as stringified JSON instead of `undefined` or `[object Object]`

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
@@ -86,6 +86,7 @@ const ExperimentVisualization: React.FC<Props> = ({
     name: experiment.config.searcher.metric,
     type: MetricType.Validation,
   });
+  console.log(experiment.hyperparameters);
   const fullHParams = useRef<string[]>(
     (Object.keys(experiment.hyperparameters || {}).filter((key) => {
       // Constant hyperparameters are not useful for visualizations.
--- a/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentVisualization.tsx
@@ -86,7 +86,6 @@ const ExperimentVisualization: React.FC<Props> = ({
     name: experiment.config.searcher.metric,
     type: MetricType.Validation,
   });
-  console.log(experiment.hyperparameters);
   const fullHParams = useRef<string[]>(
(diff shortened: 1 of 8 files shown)


trentwatt (approved):
Overall, I would like it a lot better if we could find a way to make it so that nested categorical hps were displayed separately in the table with the `outer.inner` notation.

the json stringification makes it hard to read, and to understand whats different. also if there is more than one you get some of this going on 
image

Also [here](https://deploy-preview-4818--determined-ui.netlify.app/experiments/126/visualization/hp-parallel-coordinates) it looks like 2 of the trials have `WarmupDecayLR` scheduler, but the chart shows them all as being `WarmupLR`.

image


another point is that by default we display only the hyperparameters that differ across trials. we do that by checking if it has `Type==Constant`. but with unnested, you can have the nested hps structure as a whole differ but some of the inner ones are constant, in which case it may be better to hide those, which requires manual bookkeeping.

there some logic to do this in the `ExperimentComparison` pages, check the logic involving `hpValsMap` if you are interested.

Overall i think that the `hyperparameters` specific code in this part of the codebase is a lot more convoluted than it needs to be, and it makes dealing with this kind of thing a bit of a pain. 

We have some work in flight in trials comparison to simplify things a lot, it might be better to wait till that lands and build off that.  so i think probably just json stringifying is good enough for now. your call.

trentwatt:
Hopefully maybe we can see about this though
>Also [here](https://deploy-preview-4818--determined-ui.netlify.app/experiments/126/visualization/hp-parallel-coordinates) it looks like 2 of the trials have WarmupDecayLR scheduler, but the chart shows them all as being WarmupLR.

trentwatt:
also this: `[Object object]`

https://deploy-preview-4818--determined-ui.netlify.app/experiments/126/trials/351/hyperparameters

PR #4746, opened 2022-08-09
fix: reconcile metrics proto, move `det trial describe` to the new API. [DET-7617]

ioga:
## Description

- `det trial describe --json --metrics` is used to extract detailed info about trial for post-processing. It's been for large trials due to postgres json marshalling limits. This PR moves it to the new, paginated API.
- New API format is different. This is a breaking change for `det trial describe --json` output.
- Move `MetricsWorkload` and `TrialMetrics` protos to use `checkpointv1.Metrics` object for metrics instead of custom disjointed fields. This is a breaking change for REST APIs, but user-facing Core APIs for `report_{training,validation}_metrics` stay the same.
- Allow users to request per-batch metrics from `GetTrialWorkloads` API using `include_batch_metrics` query flag.

## Test Plan

- Tabulated `det trial describe [--metrics]` format should stay the same before previous and new version.
- `det trial describe --json [--metrics]` format will change, but should contain the same volume of information.

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](https://github.com/determined-ai/determined/blob/master/docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.
- [ ] If modifying `/webui/react/src/shared/` verify `make -C webui/react test-shared` passes.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4821, opened 2022-08-18
fix: handle avgMetrics response on trials of multi-trial experiment

mapmeld:
## Description

When viewing an individual trial on a multi-trial experiment, the endpoint returns `metrics: { avgMetrics: {...}, batchMetrics: [ ... ] }` instead of a single `metrics` object. On [latest-master](http://latest-master.determined.ai:8080/det/experiments/1292/trials/3709?filter=All&metric=validation%7Cvalidation_loss&metric=training%7Closs) the result is a workloads table with no numbers.

This modifies `filterNonScalarMetrics` to get metrics out of `metrics.avgMetrics` if it exists.

Edit: this also fixes `bestValidationMetric` and `latestValidationMetric` in the trial comparison table.

## Test Plan

- Single-trial experiment has numbers in workloads table.
- While running a multi-trial experiment, numbers appear in `bestValidationMetric` and `latestValidationMetric` columns.
- After running a multi-trial experiment, select one trial, and confirm numbers appear in the workloads table

## Checklist

### DIFF
--- a/webui/react/src/services/decoder.ts
+++ b/webui/react/src/services/decoder.ts
@@ -430,6 +430,9 @@ export const mapV1ExperimentList = (data: Sdk.V1Experiment[]): types.ExperimentI
 
 const filterNonScalarMetrics = (metrics: RawJson): RawJson | undefined => {
   if (!isObject(metrics)) return undefined;
+  if (metrics.avgMetrics) {
+    return filterNonScalarMetrics(metrics.avgMetrics);
+  }
   const scalarMetrics: RawJson = {};
   for (const key in metrics) {
     if ([ 'Infinity', '-Infinity', 'NaN' ].includes(metrics[key])) {


ioga:
oh. that's my fault #4746 
since all of this is now using autogenerated bindings, I wish these issues were caught by the type system.

mapmeld:
> I wish these issues were caught by the type system.

I was thinking that this (or a future PR) should have a test

ioga (approved):

webui/react/src/services/decoder.ts
```diff
@@ -430,6 +430,9 @@ export const mapV1ExperimentList = (data: Sdk.V1Experiment[]): types.ExperimentI
 
 const filterNonScalarMetrics = (metrics: RawJson): RawJson | undefined => {
```
ioga, line 431: I think this can be typed more strictly as `metrics: Sdk.V1Metrics`

PR #4822, opened 2022-08-18
chore: preserve failed action's error msg

hamidzr:
## Checklist


mapmeld (approved):
I'm not familiar with `wrapPublicMessage` but this is consistent with how it's used in `src/utils/job.ts`

hamidzr:
> I'm not familiar with `wrapPublicMessage` but this is consistent with how it's used in `src/utils/job.ts`

all it does is it wraps and exposes the underlying publicMessage if any instead of overwriting it.

PR #4829, opened 2022-08-19
feat: rolling upgrades support for `det deploy aws` and latest-master [DET-7853].

ioga:
## Description

Add support for agent_reattach option aka rolling upgrades and enable it on latest-master, with the goal of 
a) testing this feature more aggressively day-to-day
b) making latest-master updates less disruptive

- Connect the piping for agent connection options (`agent_reconnect_attempts`, `agent_reconnect_backoff`) to be passed via provisoner config.
- Connect the piping to pass `agent_reattach_enabled` and agent connection options in `det deploy aws`.
- Enable rolling upgrades with reasonable settings for `latest-master`.
- Also take out so-called "standalone mode" from agent because it's useless, as discussed in backend-eng.

## Test Plan

We'll see how it works on latest-master in practice. 

For the underlying feature itself, [substitute cluster-id and keypair for your own]
1. Make a cluster: `det deploy aws up --cluster-id ilia-test-2 --keypair ilia --det-version c27f9076f88e85329e5542ce5a808bf1123817be --agent-reattach-enabled true --agent-reconnect-attempts 26 --deployment-type efs --no-update-terminate-agents`
2. Start a test task, e.g. a shell: `det -m CLUSTER_URL shell start --config resources.slots=0`
3. ^D it; check it's still up in `det -m CLUSTER_URL shell list`
4. Redeploy a cluster. we're just going to change the agent reconnect option so the cluster is actually updated. `det deploy aws up --cluster-id ilia-test-2 --keypair ilia --det-version c27f9076f88e85329e5542ce5a808bf1123817be --agent-reattach-enabled true --agent-reconnect-attempts 24 --deployment-type efs --no-update-terminate-agents`
5. check the shell is still up in `det -m CLUSTER_URL shell list`. connect back to it using `det -m CLUSTER_URL shell open SHELL_ID`.

## Checklist


harness/determined/deploy/aws/templates/vpc.yaml
```diff
@@ -504,7 +521,7 @@ Resources:
   LogGroup:
     Type: AWS::Logs::LogGroup
     Properties:
-      LogGroupName: !Sub /determined/det-${AWS::StackName}
+      LogGroupName: !Sub /${LogGroupPrefix}/${AWS::StackName}
```
dannysauer, line 524: This seems like a minor but unrelated change / bugfix.

dannysauer (approved):
I'm fine with the CI part and the rest looks generally OK to me (though I do wonder about adding an extra parameter to python methods without a backwards compatability default value). So I'm adding my approval, but I'd prefer to see an additional approval by someone more familiar with the code before merging. :)

harness/determined/deploy/aws/templates/vpc.yaml
ioga: it happens to be a fix for https://determined-ai.slack.com/archives/C7E6HPZ5F/p1660787284400649

NicholasBlaskey (approved):

PR #4831, opened 2022-08-19
chore: rename a file to prevent API breakages

rb-determined-ai:
No user should be using determined.experimental.common.session directly,
but there's no point in possibly breaking any users, since this wasn't
clearly communicated when it was originally released.

This is separate from 726fe804 (#4787) because if it were combined into
a single commit, then the git history of the
determined/common/api/_session.py file would have been destroyed.


mpkouznetsov (approved):

mpkouznetsov:
My favorite fixes are one-liners. This is even better.

PR #4833, opened 2022-08-19
fix: begin standardizing API pagination behavior in CLI.

ioga:
## Description

- Jobs listing API/CLI was using a distinct pagination behavior. This PR unifies it with the rest of the code, at the expense of breaking it.
- Add support for fetching a single data page using `--pages=1`.
- QoL: add `make install` in `harness` for ease of installing harness from source in development.

## Test Plan

`det job list` should now list all jobs. `det job list --offset=50 --pages=1` should be replicating the old behavior.

## Checklist


proto/src/determined/api/v1/job.proto
```diff
@@ -12,8 +12,10 @@ message GetJobsRequest {
   option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
     json_schema: { required: [] }
   };
-  // Request to paginate the resposne.
-  PaginationRequest pagination = 1;
+  // Pagination offset.
+  int32 offset = 10;
+  // Pagination limit.
+  int32 limit = 11;
```
stoksc, line 18: what is the philosophical reason for this change when a `PaginationRequest` could enforce consistency of the API?

stoksc: better ux?

harness/determined/cli/util.py
```diff
@@ -26,30 +27,26 @@
...
-    ),
-    Arg(
-        "--reverse",
-        default=False,
-        action="store_true",
-        help="Reverse the requested order of results.",
-    ),
-]
+
+class PageOpts(str, Enum):
```
stoksc, line 31: these options are pretty limiting, but i guess that is fine.

proto/src/determined/api/v1/job.proto
ioga: all of the API is using top-level limit/offset, except for this one `GetJobs` API call. 
we can debate if nested `PaginationRequest` object is good for API v2. for v1 I'd rather normalize to the more popular and widespread option.

harness/determined/cli/util.py
ioga: For unlimited custom pagination I'd rather just use the bindings directly.

webui/react/src/services/api-ts-sdk/api.ts
```diff
@@ -12840,15 +12820,15 @@ export const InternalApiFetchParamCreator = function (configuration?: Configurat
         /**
          * 
          * @summary Get a list of jobs in queue.
-         * @param {number} [paginationOffset] The number of records to skip before returning results.
-         * @param {number} [paginationLimit] The amount of records limited in the results.
+         * @param {number} [offset] Pagination offset.
```
hamidzr, line 12823: nit: the previous comment seems more exact as to how it's implemented. keep it?

hamidzr (approved):
👍🏼  this shouldn't change how the webui uses pagination for job queues. I like the idea of sharing pagination param definitions (proto messages) across our apis so that the descriptions and naming has to stay the same but whichever direction is more consistent today sgtm

stoksc (approved):
lgtm

proto/src/determined/api/v1/job.proto
hamidzr: we seem to use the remaining `Pagination` message in a lot of our APIs. We seem to use them about the same time?

```
hrz ➜ proto git:(master) ag 'Pagination.*=' | wc -l
19
hrz ➜ proto git:(master) ag 'limit =' | wc -l
22
```
currently `[Pagination](https://github.com/hamidzr/determined/blob/60b29812654202a38dacdab0ef3f945ebb02d434/proto/src/determined/api/v1/pagination.proto#L16)` in concept extends `PaginationRequest` . an alternative could be to breakup `Pagination` so that we properly link and enforece PaginationRequest

ioga: proto3 doesn't have any sort of inheritance, "extend" was a thing in proto2 but it's not that useful. I don't see how we can "reuse" the top level fields. 
member message like `PaginationRequest` is a solution but it'd break everything. I don't think it's worth doing that in v1.

webui/react/src/services/api-ts-sdk/api.ts
ioga: it seems redundant to me to explain what "pagination offset" is. it's standard limit-offset pagination, if it doesn't work like one and needs a verbose explanation then we're doing something wrong.

PR #4843, opened 2022-08-23
feat: add security.default_task and openshift host options to helm chart [DPS-204]

asnelling:
## Description

add `security.default_task` and openshift host options to helm chart [DPS-204]

## Test Plan

Install Determined using helm chart as in [this guide](https://docs.determined.ai/latest/cluster-setup-guide/deploy-cluster/sysadmin-deploy-on-k8s/install-on-kubernetes.html):
- without changing `values.yaml`
- with `security.defaultTask` configured in `values.yaml` as:
   ```yaml
   security:
     defaultTask:
       user: det-nobody
       uid: 65533
       group: det-nobody
       gid: 65533
   ```
   Verify changes propagated to `master.yaml`:
   ```Shell
   $ det -u admin master config | grep -e det-nobody -e 65533
       gid: 65533
       group: det-nobody
       uid: 65533
       user det-nobody
   ```

## Checklist

[DPS-204]: https://determinedai.atlassian.net/browse/DPS-204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helm/charts/determined/templates/NOTES.txt
+++ b/helm/charts/determined/templates/NOTES.txt
@@ -19,12 +19,14 @@ kubectl get service determined-master-service-{{ .Release.Name }} -n {{ .Release
 {{ end -}}
 
 {{ $httpOrHttps := "http" }}
+{{- if .Values.security -}}
 {{- if .Values.security.tlsSecret -}}
   {{ $httpOrHttps = "https" }}
   {{ $secret := (lookup "v1" "Secret" .Release.Namespace (.Values.security.tlsSecret | toString) ) }}
   {{- if eq (len $secret) 0 }}
 ERROR: Secret {{ .Values.security.tlsSecret }} does not exist in namespace: {{ .Release.Namespace }}. Please update ".Values.security.tlsSecret".
   {{ end }}
+{{ end -}}
 {{ end -}}
 
 {{- if (and (not .Values.detVersion) (contains "dev" .Chart.AppVersion)) }}
--- a/helm/charts/determined/templates/NOTES.txt
+++ b/helm/charts/determined/templates/NOTES.txt
@@ -19,11 +19,11 @@ kubectl get service determined-master-service-{{ .Release.Name }} -n {{ .Release
 {{ end -}}
 
 {{ $httpOrHttps := "http" }}
-{{- if .Values.security.tlsSecret -}}
+{{- if .Values.tlsSecret -}}
   {{ $httpOrHttps = "https" }}
-  {{ $secret := (lookup "v1" "Secret" .Release.Namespace (.Values.security.tlsSecret | toString) ) }}
+  {{ $secret := (lookup "v1" "Secret" .Release.Namespace (.Values.tlsSecret | toString) ) }}
   {{- if eq (len $secret) 0 }}
-ERROR: Secret {{ .Values.security.tlsSecret }} does not exist in namespace: {{ .Release.Namespace }}. Please update ".Values.security.tlsSecret".
+ERROR: Secret {{ .Values.tlsSecret }} does not exist in namespace: {{ .Release.Namespace }}. Please update ".Values.tlsSecret".
   {{ end }}
 {{ end -}}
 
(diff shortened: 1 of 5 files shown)


helm/charts/determined/templates/master-deployment.yaml
```diff
@@ -40,11 +40,13 @@ spec:
...
             mountPath: /etc/determined/
             readOnly: true
-          {{- if .Values.tlsSecret }}
+          {{- if .Values.security }}
+          {{- if .Values.security.tlsSecret }}
           - name: tls-secret
             mountPath: {{ include "determined.secretPath" . }}
             readOnly: true
           {{ end }}
+          {{ end }}
```
dannysauer, line 49: We're oddly inconsistent with `{{ end }}`, `{{- end }}`, and `{{ end -}}` i this chart.  Maybe not really something that belongs in this PR, but probably something worth cleaning up.

helm/charts/determined/values.yaml
```diff
@@ -22,14 +22,26 @@ useNodePortForMaster: false
...
+
+  # defaultTask sets the user and group that tasks will run as. For convenience, the default Determined
+  # environments contain an unprivileged user named det-nobody, which does have a writable HOME
+  # directory. The det-nobody user is a suitable default user when using the default Determined
+  # environment images and when running containers as root is not desired.
+  # defaultTask:
+  #   user: det-nobody
+  #   uid: 65533
+  #   group: det-nobody
+  #   gid: 65533
```
dannysauer, line 44: I forget - is there a good reason to not use `-3` here so it'll overflow properly on both systems with 32-bit IDs and older systems with 16-bit IDs?  Is `65533` hard-coded elsewhere in the containers?

dannysauer (approved):
I don't see any technical issues, though I did point out a couple of _potential_ nits.

It also would likely be a good idea to at least include a changelog entry indicating that the hierarchical location of the `tlsSecret` parameter has changed.  I'm not sure how we typically do that.  I'm marking approved anyway; I'll let you figure out if those things need further changes or not. :)

helm/charts/determined/templates/NOTES.txt
```diff
@@ -19,13 +19,15 @@ kubectl get service determined-master-service-{{ .Release.Name }} -n {{ .Release
 {{ end -}}
 
 {{ $httpOrHttps := "http" }}
```
varlogtim, line 21: Per the thread, if they make no changes and don't read the docs, this silently fails to set their secret. Include an error message if they still have `.Values.tlsSecret` set.

varlogtim (changes requested):
Per slack thread, please handle the backwards compatibility case. Preferably by printing a warning/error message indicating that the location of the `tlsSecret` has moved.

helm/charts/determined/values.yaml
varlogtim: 65533 is hardcoded in our containers as `det-nobody`, I believe.

helm/charts/determined/templates/NOTES.txt
dannysauer: I was looking for an error function, but of course, Helm (and go templates) are never what anyone really wants. :D

My current best idea is to use [required](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function) in a random template file to depend on the new value if the old is set; maybe something along the lines of this untested blob of horror:
```plain
{{- if .Values.tlsSecret }} 
{{- required "tlsSecret has moved to security.tslSecret" .Values.security.tlsSecret }}
{{- end }}
```
I don't know for sure if that works in a conditional, though, so :shrug:

dannysauer: Oh, sprig has a fail function.  So maybe `{{- fail "tlsSecret has moved to security.tslSecret" }}` would be a similar valid option to always fail if the old value is provided  I kinda like continuing to work if the new one is also added.  Dealer's choice, IMO.

asnelling: The decision was made to move `tlsSecret` back to it's original home and avoid the situation altogether.

varlogtim (approved):

dannysauer (approved):
This LGTM

dannysauer:
Also, I'd say just go ahead and admin merge this. You could rebase and then run the teats again to get the python lint error to pass, but since there's no python change anyway, that seems pointless.

PR #4844, opened 2022-08-23
chore: migrate python sdk to generated bindings [DET-8005]

rb-determined-ai:
## Test Plan

All code touched should either be tested by unit tests (where possible) or by e2e tests (since we need a master running for most of the sdk).

## Commentary (optional)

The nullability information for proto objects is awful.  If you were to go proto -> json -> proto it wouldn't be as bad, because proto doesn't distinguish zero values and missing values typically, but json does.

However, our proto spec generates swagger, which is in json, so the concept of zero values vs missing values is significant.  The process of finding out which values might be _actually_ missing after passing an SQL query, the go code for a REST API, and proto marshaling is difficult, and had to be done on many fields.


harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -16,23 +17,40 @@ class ModelFramework(enum.Enum):
...
+    ERROR = bindings.determinedcheckpointv1State.STATE_ERROR.value
+    DELETED = bindings.determinedcheckpointv1State.STATE_DELETED.value
 
 
 @dataclasses.dataclass
 class CheckpointTrainingMetadata:
     experiment_config: Dict[str, Any]
-    experiment_id: str
-    trial_id: str
+    experiment_id: int
```
mpkouznetsov, line 30: How come those were ever `str`?

harness/determined/common/api/_unpaginate.py
```diff
@@ -0,0 +1,27 @@
...
+    bindings.v1GetExperimentCheckpointsResponse,
+    bindings.v1GetTrialCheckpointsResponse,
+    bindings.v1GetModelsResponse,
+    bindings.v1GetModelVersionsResponse,
+]
+
+T = TypeVar("T", bound=Unpaginatable)
+
+
+def unpaginate(get_with_offset: Callable[[int], T]) -> Iterator[T]:
```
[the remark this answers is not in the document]
rb-determined-ai: yes, I planned on talking about that with you.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -49,8 +67,8 @@ class Checkpoint(object):
     def __init__(
         self,
         session: api.Session,
-        task_id: str,
-        allocation_id: str,
+        task_id: Optional[str],
+        allocation_id: Optional[str],
```
mpkouznetsov, line 71: Are these conceptually (semantically) optional?

harness/determined/common/experimental/checkpoint/_checkpoint.py
rb-determined-ai: Pretty sure the proto always had them ints, and this type annotation was just wrong (and would have been wrong at runtime).

from_json() didn't let mypy find this error before.

harness/determined/common/experimental/checkpoint/_checkpoint.py
rb-determined-ai: proto just has no way to express that this entire struct is either None, or all the fields are present.

But that's conceptually what the system guarantees (either the TrainingMetadata isn't there or it's fully populated), so we promise that here in the sdk.

harness/determined/common/experimental/determined.py
```diff
@@ -155,12 +155,15 @@ def create_model(
             description (string, optional): A description of the model.
             metadata (dict, optional): Dictionary of metadata to add to the model.
         """
-        r = self._session.post(
-            "/api/v1/models",
-            json={"description": description, "metadata": metadata, "name": name, "labels": labels},
+
+        # TODO: add notes param to create_model()
```
mpkouznetsov, line 159: this is out of scope for this PR?

rb-determined-ai: no public api changes in this pr, migrating all to the generated bindings is plenty.

mpkouznetsov (approved):

harness/determined/common/api/_unpaginate.py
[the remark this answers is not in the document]
rb-determined-ai: ok, #4859 merges the feature sets of `limit_offset_paginator` and `unpaginate`, and refactored existing users of limit_offset_paginator.

harness/determined/common/experimental/checkpoint/_checkpoint.py
rb-determined-ai: Oh!  I was talking about something else when I answered.

task_id and allocation_id are optional because we want to support importing checkpoints into the system which are not related to any task or allocation.

PR #4845, opened 2022-08-23
chore: document reasons for scaler.update()

rb-determined-ai:
## Commentary (optional)

The code was written correctly the first time, but this information was nearly lost and we almost chose to break it.

This is a great time to add a comment.

### DIFF
--- a/harness/determined/pytorch/README.md
+++ b/harness/determined/pytorch/README.md
@@ -131,6 +131,9 @@ step the optimizer.
 Call the dumb control loop: if there was an overflow, decrease the scale
 factor.  If there was an underflow, increase it.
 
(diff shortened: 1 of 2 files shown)


drh-determined-ai:
Other than `lint` complaining about `TrialContext` not having any abstract methods, this looks good to me.

drh-determined-ai (approved):
Other than `lint` complaining about `TrialContext` not having any abstract methods, this looks good to me!

rb-determined-ai:
> Other than lint complaining about TrialContext not having any abstract methods, this looks good to me!

Yeah... the linter got smarter when we weren't looking.

PR #4857, opened 2022-08-24
fix: remove workloadCount from trial responses; single-trial view fix

mapmeld:
## Description

`workloadCount` was adding complexity to our queries with limited utility.  The PR removes `workloadCount` and instead makes a request for one trial workload (paginated workloads API).

If the trial finishes with 0 workloads then `trialNeverData` gets set to `true` and the TensorBoard button is hidden.

## Test Plan

On a multi-trial experiment, view a single trial. It should have a TensorBoard button in the top right.

## Checklist


master/test/integration/api/api_trials_test.go
```diff
@@ -124,9 +124,8 @@ func trialDetailAPITests(
 			ctx, _ := context.WithTimeout(creds, 10*time.Second)
 			req := apiv1.GetTrialRequest{TrialId: int32(trial.ID)}
 
-			tlCl, err := cl.GetTrial(ctx, &req)
+			_, err = cl.GetTrial(ctx, &req)
 			assert.NilError(t, err, "failed to fetch api details")
-			assert.Equal(t, tlCl.Trial.WorkloadCount, int32(1), "mismatching workload length")
```
hamidzr, line 129: missing out on this test

webui/react/src/pages/TrialDetails/TrialDetailsHeader.tsx
```diff
@@ -53,10 +50,23 @@ const TrialDetailsHeader: React.FC<Props> = ({
...
+      });
+      setTrialNeverData(wl.workloads.length === 0);
+    }
+  }, [ trial ]);
+
   const headerOptions = useMemo<Option[]>(() => {
     const options: Option[] = [];
 
-    if (!trialWillNeverHaveData(trial)) {
+    if (!trialNeverData) {
```
hamidzr, line 69: nit keep the same name? will avoid adding extra changes as well

webui/react/src/pages/TrialDetails/TrialDetailsHeader.tsx
```diff
@@ -32,6 +28,7 @@ const TrialDetailsHeader: React.FC<Props> = ({
   trial,
 }: Props) => {
   const [ isRunningTensorBoard, setIsRunningTensorBoard ] = useState<boolean>(false);
+  const [ trialNeverData, setTrialNeverData ] = useState<boolean>(false);
```
hamidzr, line 31: do we need this? could the usememo just return the value you want? const trialNeverData = useMemo ...

ioga (approved):

webui/react/src/pages/TrialDetails/TrialDetailsHeader.tsx
hamidzr: ah it's async so we do. maybe we then don't want it to be useMemo? that could be a useEffect I think

hamidzr (commented):
I reviewed all the changes but the sql one. mostly looks good.  I haven't visually tested it yet

webui/react/src/pages/TrialDetails/TrialDetailsHeader.tsx
mapmeld: I wasn't sure the right way to get the async value, and have it be usable inside `headerOptions`

master/test/integration/api/api_trials_test.go
mapmeld: the original test prior was looking for `len(tlCl.Trial.Workloads) == 1`, which was replaced by `tlCl.Trial.WorkloadCount == 1`, but now has no equivalent

PR #4859, opened 2022-08-24
chore: some cli cleanup

rb-determined-ai:
- rename limit_offset_paginator -> read_paginated, make it
  mypy-friendly, and move to common so it can be used in the python sdk
- collapse cli.util and cli.session into the toplevel cli namespace
- make a small dent in the cli non-compliance with google style guide
  for imports

# Test Plan:
- [x] change det e describe limit to 1 and describe an experiment with lots of steps and many trials.  Actually this looks broken, but also seems broken on tip-of-master.
- [x] try det e list with various pagination args
- [x] try det e lt on an experiment with many trials, with various pagination args
- [x] try det j list with various pagination args while many experiments are queued
- [x] try det t describe with various pagination args on a trial with many workloads


harness/determined/common/api/_util.py
```diff
@@ -0,0 +1,51 @@
...
+    pages: PageOpts = PageOpts.all,
+) -> Iterator[T]:
+    while True:
+        resp = get_with_offset(offset)
+        pagination = resp.pagination
+        assert pagination
+        yield resp
+        if pagination.endIndex == pagination.total or pages == PageOpts.single:
+            break
+        offset = resp.endIndex + 1
```
mapmeld, line 51: should be         `pagination.endIndex + 1`

rb-determined-ai: yeah, I hit this in manual testing.

Also, experimentally I find that it should be `pagination.endIndex`, without the `+ 1`.  Do you know why that is?  Otherwise I seem to miss one entry after every page (which is obvious when `--limit=1`).

mapmeld (approved):
lgtm now :shipit:

harness/determined/common/api/_util.py
mapmeld: I don't know enough about where endIndex comes from, but if you received one result you would want to OFFSET 1 and not OFFSET (1+1), so maybe it makes sense

harness/determined/cli/_util.py
```diff
@@ -1,67 +1,62 @@
-from enum import Enum
-from typing import Any, Callable, Dict, List, Optional
+import argparse
+from typing import Dict, List
 
-from determined.common import api
-from determined.common.declarative_argparse import Arg
```
ioga, line 5: nitpick: we import `declarative_argparse` members like this everywhere in cli.

harness/determined/common/api/_util.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class PageOpts(str, enum.Enum):
+    single = "1"
+    all = "all"
+
+
+# All paginated response types which want read_paginated() to work with.
+# A Paginated union member must have a .pagination attribute.
+# hint: grep 'pagination: .*v1Pagination.*' bindings.py -B 7 | sed -E -n -e 's/class (.*):/\1/p'
```
ioga, line 14: now I want the bindings generator to automatically produce this list: I am sure developers will be annoyed with having to add their new APIs here manually.

ioga (approved):

harness/determined/common/api/_util.py
rb-determined-ai: I am sure these developers can either add it themselves or get over it.

rb-determined-ai: Eh, I suppose it could be added to bindings.py without too much effort.

harness/determined/cli/_util.py
rb-determined-ai: The cli is very old code, and it was written before we adopted the [google style guide for python imports](https://google.github.io/styleguide/pyguide.html#22-imports), which is observed in most all of the rest of the harness code.

New code in the CLI has typically followed the existing style rather than the new standard, but that's a bug not a feature IMO.

harness/determined/common/api/_util.py
rb-determined-ai: @stoksc would you also double-check this?  I'm a bit scared to land it without confirmation of this detail.

harness/determined/cli/_util.py
ioga: do you mean we've banned python imports of functions/classes/constants from modules, and we only can import whole modules?

harness/determined/common/api/_util.py
rb-determined-ai: done, added as a second commit.

ioga: awesome

harness/determined/cli/_util.py
rb-determined-ai: Yes, a long time ago.

harness/determined/common/api/_util.py
rb-determined-ai: confirmed with @stoksc, `endIndex` has the same "one after the end" mechanics as python slice indexing.

PR #4862, opened 2022-08-24
feat: list existing tags

keita-determined:
## Description
[DET-8133]

## Test Plan
Check if there are list of existing tags in project, project detail, model, model version page
Check if we are able to select by cursor and pointer, and able to add tag by blur and enter key

## Checklist

[DET-8133]: https://determinedai.atlassian.net/browse/DET-8133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/master/internal/api_model.go
+++ b/master/internal/api_model.go
@@ -384,6 +384,15 @@ func (a *apiServer) GetModelVersions(
 	return resp, a.paginate(&resp.Pagination, &resp.ModelVersions, req.Offset, req.Limit)
 }
 
+func (a *apiServer) GetModelVersionLabels(_ context.Context, req *apiv1.GetModelVersionLabelsRequest) (*apiv1.GetModelVersionLabelsResponse, error) {
+	resp := &apiv1.GetModelVersionLabelsResponse{}
+	err := a.m.db.QueryProto("get_model_version_labels", resp)
+	if err != nil {
+		return nil, err
+	}
+	return resp, errors.Wrapf(err, "error getting model version labels")
+}
+
 func (a *apiServer) PostModelVersion(
 	ctx context.Context, req *apiv1.PostModelVersionRequest,
 ) (*apiv1.PostModelVersionResponse, error) {
--- a/master/internal/api_model.go
+++ b/master/internal/api_model.go
@@ -384,7 +384,9 @@ func (a *apiServer) GetModelVersions(
 	return resp, a.paginate(&resp.Pagination, &resp.ModelVersions, req.Offset, req.Limit)
 }
 
-func (a *apiServer) GetModelVersionLabels(_ context.Context, req *apiv1.GetModelVersionLabelsRequest) (*apiv1.GetModelVersionLabelsResponse, error) {
+func (a *apiServer) GetModelVersionLabels(
+	_ context.Context, req *apiv1.GetModelVersionLabelsRequest,
+) (*apiv1.GetModelVersionLabelsResponse, error) {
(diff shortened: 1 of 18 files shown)


master/internal/api_model.go
```diff
@@ -384,6 +384,17 @@ func (a *apiServer) GetModelVersions(
...
 
+func (a *apiServer) GetModelVersionLabels(
+	_ context.Context, req *apiv1.GetModelVersionLabelsRequest,
+) (*apiv1.GetModelVersionLabelsResponse, error) {
+	resp := &apiv1.GetModelVersionLabelsResponse{}
+	err := a.m.db.QueryProto("get_model_version_labels", resp)
+	if err != nil {
+		return nil, err
+	}
+	return resp, errors.Wrapf(err, "error getting model version labels")
```
trentwatt, line 395: something like 
```
var labels string[]
err := db.Bun().NewSelect()
    ...
    .Scan(&labels)

...
resp  := apiv1.GetModelVersionLabelsResponse{labels: &labels}
```

master/static/srv/get_model_version_labels.sql
```diff
@@ -0,0 +1,11 @@
+WITH sorted_labels AS (
+  SELECT label
+  FROM (
+    SELECT id, UNNEST(labels) AS label
+    FROM model_versions
```
ClaireNeveu, line 5: Oof, so it seems we don't have a table for tags and we need to scan the whole table of taggable objects to get the list of them? Do we have an index on labels? If not we could use an index to precompute this list on write so we don't have to scan the whole table on read.

keita-determined: we dont have a table for tags or index for labels. There are 2 types of index for array type. I'll check them out

ClaireNeveu: It's been a while since I've done this kind of thing but in addition to postgres specific stuff you might look into loose index scanning to just get the unique values from an index instead of all the values.

trentwatt: related:

https://github.com/determined-ai/determined/blob/master/master/static/srv/get_experiment_labels.sql

btw [...] has 2 million experiments.

we are going with a different data model for tags as a part of trials comparison, due to this and race condition issues with the current approach.

basically just using a `jsonb` dict where keys are present if the tag exists for the object.

using GIN `jsonb_path_ops` index allows for efficient lookup based on keys.

I think if you did an agg function just using the [index compatible operators](https://www.postgresql.org/docs/current/gin-builtin-opclasses.html), it would be able to use the index for getting distinct tags as well. Have not validated yet though.

trentwatt: the plan (based on team discussions) was to try the new data model/approach with trials, and then convert everywhere else we use tags after validating.

trentwatt: i updated [this ticket](https://determinedai.atlassian.net/browse/DET-8185) to reflect the overall context

ClaireNeveu: I think this ticket is blocked on that more efficient tag representation. Let's table DET-8133 for now and revisit once we can fix the tag lookup.

ClaireNeveu: I created DET-8243 to track the issue of tag representation and I'm gonna pull tag typeahed out of the sprint for the time being.

PR #4863, opened 2022-08-25
fix: det e describe with multiple trials

rb-determined-ai:
Since 0.17.12, det e describe has mangled the output of all workloads
from all contained trials together.

This was likely not noticed due to either:
- users not frequently using det e describe, or
- 90% of experiments are single-trials, so maybe those who do use
  det e describe didn't happen to need it on a multi-trial experiment.

## Test Plan

- [x] run an experiment with many trials, make sure `det e describe` works.


rb-determined-ai:
@mapmeld, just the second commit needs review.

mapmeld (approved):
LGTM

PR #4864, opened 2022-08-25
chore: remove init_invalid_hp [DET-8234]

nrajanee:
## Description

chore: remove init_invalid_hp [DET-8234]

[DET-8234]: https://determinedai.atlassian.net/browse/DET-8234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mpkouznetsov (commented):
1. It looks go linter is not happy. You probably have to run `make -C proto gen-buf-image` and add the generated file to the change list.

master/pkg/model/experiment.go
```diff
@@ -786,9 +786,6 @@ func ExitedReasonToProto(r ExitedReason) experimentv1.TrialExitedEarly_ExitedRea
 
 	case UserCanceled:
 		return *experimentv1.TrialExitedEarly_EXITED_REASON_USER_REQUESTED_STOP.Enum()
-
-	case InitInvalidHP:
-		return *experimentv1.TrialExitedEarly_EXITED_REASON_INIT_INVALID_HP.Enum()
```
mpkouznetsov, line 791: Eyeballing the code, it seems that this `case` will no longer occur and therefore we should never land in the `default` clause with `panic`. We can probably leave it like this for now. Alternatively, we could also map `case InitInvalidHP` to `*experimentv1.TrialExitedEarly_EXITED_REASON_INVALID_HP.Enum()`

nrajanee: Yup this case won't occur in this function for sure. Because this is only for custom-searcher.

mpkouznetsov (approved):

PR #4865, opened 2022-08-25
ci: fix flakes in test_max_concurrent_trials

rb-determined-ai:
There was a race condition where two trials might not get created at the
exact same time, creating the possibility for false negatives

### DIFF
--- a/e2e_tests/tests/experiment/test_core.py
+++ b/e2e_tests/tests/experiment/test_core.py
@@ -421,7 +421,7 @@ def test_core_api_distributed_tutorial() -> None:
                     "batches": 3000,
                 },
                 "max_trials": 8,
-                "max_concurrent_trials": 2,
+                "max_concurrent_trials": 1,
             },
         ),
         (
@@ -433,7 +433,7 @@ def test_core_api_distributed_tutorial() -> None:
                     "batches": 3000,
                 },
                 "max_trials": 8,
-                "max_concurrent_trials": 2,
+                "max_concurrent_trials": 1,
             },
         ),
     ],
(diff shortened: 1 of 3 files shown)


stoksc (approved):

e2e_tests/tests/experiment/test_core.py
```diff
@@ -467,8 +473,11 @@ def test_max_concurrent_trials(name: str, searcher_cfg: str) -> None:
         # the first two trials.
         exp.cancel_single(experiment_id)
 
+        trials = exp.wait_for_at_least_n_trials(experiment_id, 4)
+        assert len(trials) == 4, trials
+        time.sleep(3)
```
stoksc, line 478: i was trying so hard to avoid this. if we're going to do this, we prob can remove line 474

rb-determined-ai:
@stoksc I actually have a couple more questions about this test.

First question is: why are 2 new trials created _after_ the experiment is canceled?

Second question is: does this test run with enough slots on the cluster to even produce more than two concurrent trials?  Like, is max_concurrent_trials=2 being implicitly enforced by the cluster size?

e2e_tests/tests/experiment/test_core.py
rb-determined-ai: Avoiding it to avoid a sleep?

Without the sleep, how can you actually know max_concurrent_trials is even working?

Really... without some synchronization signal, it's still only _usually_ going to fail when max_concurrent_trials isn't respected.

rb-determined-ai:
We talked.  Long-term, the plan is to migrate this into a go test after some of the master-side tech debt has been addressed.

We'll reduce the sleeps (and increase the possibility of false positives) and switch max_concurrent_trials to 1, then call that "good enough" until this becomes a unit test someday.

PR #4868, opened 2022-08-25
chore: generate python bindings with better types

rb-determined-ai:
The generated bindings are now generous about what they accept, but
strict in what they return (from a typing perspective).

In cases where an input type annotation (like Sequence) is converted to
an output type annotation (like List), the originally-provided value
will be used if it matches the output type, otherwise a new object of
the stricter type will be created.

## Commentary (optional)

@mpkouznetsov was curious why there were some extra calls to `list()` when using the generated bindings.  There were cases where a Sequence was being returned, which was not very helpful to the caller of the bindings.

### DIFF
--- a/bindings/generate_bindings.py
+++ b/bindings/generate_bindings.py
@@ -1,4 +1,5 @@
 import abc
+import enum
 import json
 import os
 import sys
@@ -12,7 +13,16 @@ TypeDefs = typing.Dict[str, typing.Optional["TypeDef"]]
 
 
 class TypeAnno:
-    def annotation(self, prequoted=False) -> Code:
+    def input_annotation(self, prequoted=False) -> Code:
+        """input_annotation is for parameters; things accepted from a user."""
+        raise NotImplementedError(type(self))
+
+    def output_annotation(self, prequoted=False) -> Code:
+        """output_annotation is for return types and attributes; things provided to a user."""
+        raise NotImplementedError(type(self))
+
+    def input_to_output(self, val: Code, optional: bool) -> Code:
+        """Convert an input_annotation-type value to an output_annotation-type value."""
         raise NotImplementedError(type(self))
 
     def need_parse(self) -> bool:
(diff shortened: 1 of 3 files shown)


bindings/generate_bindings.py
```diff
@@ -12,7 +13,16 @@
 
 
 class TypeAnno:
-    def annotation(self, prequoted=False) -> Code:
+    def input_annotation(self, prequoted=False) -> Code:
+        """input_annotation is for parameters; things accepted from a user."""
+        raise NotImplementedError(type(self))
+
+    def output_annotation(self, prequoted=False) -> Code:
+        """input_annotation is for return types and attributes; things provided to a user."""
```
mpkouznetsov, line 21: you mean output annotation?

bindings/generate_bindings.py
```diff
@@ -301,7 +351,9 @@ def gen_def(self) -> Code:
         for name in required + optional:
             out += ["    " + self.params[name].gen_function_param()]
         out += ["    ):"]
-        out += [f"        self.{k} = {k}" for k in self.params]
+        for name in required + optional:
+            p = self.params[name]
+            out += [f"        self.{name} = {p.type.input_to_output(name, not p.required)}"]
```
rb-determined-ai, line 356: Note that before, the order of class `__init__` parameters did not match the order in which parameters were saved to the `self` object, which was annoying when inspecting the generated code.

I changed it, but it also caused a lot more churn in the generated bindings, sorry.

rb-determined-ai: this is now the second two commits, to make reviewing easier.

hamidzr (approved):
lgtm

we could add this as demo for the change
```
diff --git a/harness/determined/cli/experiment.py b/harness/determined/cli/experiment.py
index b654d61be..113f1d32d 100644
--- a/harness/determined/cli/experiment.py
+++ b/harness/determined/cli/experiment.py
@@ -688,7 +688,7 @@ def add_label(args: Namespace) -> None:
     if exp_patch.labels is None:
         exp_patch.labels = []
     if args.label not in exp_patch.labels:
-        exp_patch.labels = list(exp_patch.labels) + [args.label]
+        exp_patch.labels = exp_patch.labels + [args.label]
         bindings.patch_PatchExperiment(session, body=exp_patch, experiment_id=args.experiment_id)
     print("Added label '{}' to experiment {}".format(args.label, args.experiment_id))
 
```

rb-determined-ai:
test failure is real, it'll take some investigation.

rb-determined-ai:
closing due to staleness

PR #4869, opened 2022-08-25
fix: race condition in agent `container` actor around missing `containerInfo`.

ioga:
## Description

Should fix the race condition in https://determined-ai.slack.com/archives/C02PV33GSN5/p1661376182724349

## Test Plan

N/A: no repro.

## Checklist

### DIFF
--- a/agent/internal/container.go
+++ b/agent/internal/container.go
@@ -95,8 +95,8 @@ func getBaseTaskLog(spec *cproto.Spec) model.TaskLog {
 func (c *containerActor) Receive(ctx *actor.Context) error {
 	switch msg := ctx.Message().(type) {
 	case actor.PreStart:
-		c.docker, _ = ctx.ActorOf("docker", &dockerActor{Client: c.client})
 		if !c.reattached {
+			c.docker, _ = ctx.ActorOf("docker", &dockerActor{Client: c.client})
 			taskLog := getBaseTaskLog(c.spec)
 			c.transition(ctx, cproto.Pulling)
 			pull := pullImage{
@@ -108,7 +108,12 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 			ctx.Tell(c.docker, pull)
 			c.baseTaskLog = taskLog
 		} else {
-			ctx.Tell(c.docker, reattachContainer{ID: c.Container.ID})
+			c.docker, _ = ctx.ActorOf(
+				"docker",
+				&dockerActor{
+					Client:              c.client,
+					reattachContainerID: &c.Container.ID,
+				})
 		}
 	case getContainerSummary:
 		ctx.Respond(c.Container)
--- a/agent/internal/container.go
+++ b/agent/internal/container.go
@@ -114,6 +114,7 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 					Client:              c.client,
 					reattachContainerID: &c.Container.ID,
 				})
+			ctx.Ask(c.docker, actor.Ping{}).Get()
 		}
 	case getContainerSummary:
 		ctx.Respond(c.Container)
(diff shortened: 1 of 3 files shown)


agent/internal/container.go
```diff
@@ -108,7 +108,12 @@ func (c *containerActor) Receive(ctx *actor.Context) error {
 			ctx.Tell(c.docker, pull)
 			c.baseTaskLog = taskLog
 		} else {
-			ctx.Tell(c.docker, reattachContainer{ID: c.Container.ID})
```
NicholasBlaskey, line 111: Am I understanding this race right? 

the container actor sends ```reattachContainer``` to the docker actor 
which can send ```containerReattached``` back to the container actor setting ```containerInfo```

Then there was no synchronization between using ```containerInfo``` and that ```containerReattached``` message arriving?

NicholasBlaskey (approved):
Looks good to me!

agent/internal/container.go
ioga: `aproto.SignalContainer` needs the `containerInfo`, so if it arrives before `containerReattached` it'd hit nil pointer. so instead we try to get that data more "synchronously", which is a mess in the actor system.
for example, it may happen when agent has disconnected and master had to buffer some messages and then resend them when agent is back.

ioga: writing this comment made me realized we should also ping the docker actor for good measure.

stoksc (approved):
lgtm, but i hate it that it has to be this way ha

agent/internal/container.go
stoksc: re-reading, i'm actually don't get how this helps us get containerInfo any better? if the container actor receives a signal before `containerReattached` is sent back after the docker actor prestart, don't we still panic in the container actor trying to access c.containerInfo.id?

stoksc: it seems this would make it less likely but not impossible, since it basically saves a round trip through the actors

stoksc: it we had actually synchronization (reattachContainer as an actor ask returning containerInfo) between starting the container and having it reattached with container info, then it would work?

ioga: I think we need one more ping: `containers` actor should ask reattached `container` actors.

agent/internal/containers.go
```diff
@@ -449,6 +449,7 @@ func (c *containerManager) reattachContainer(
 		}
 		return nil, errors.New(errorMsg)
 	}
+	ctx.Ask(containerRef, actor.Ping{}).Get()
```
stoksc, line 452: with this ping, do we even need the rest? since it essentially guarantees container prestart has run before we forward anything to it? then docker would also never receive the signal before it had an id? (could still ping the docker actor for good measure)

stoksc (commented):
lgtm, though i had a comment that maybe it can be simplified now

agent/internal/containers.go
stoksc: if we get

reattach{cid: 1} -> cm
signal{cid: 1} -> cm
cm receive reattach
create cid1
ping cid1
  cid1 prestart
  create docker
  ping docker
    docker prestart
    docker reattaches container, sends container reattach
    cid1 inbox container reattach, will always be handled before is forwarded [*]
  docker ack ping
  cid1 -> docker reattach{cid: 1}, will always be handled before is forwarded [*]
cid1 ack ping
cm -> cid1 signal [*], always handled after the last ping
  cid1 is ok

idk, think this is right

ioga: this description looks like what this code is doing now. which part of this PR are you proposing to get rid of?

stoksc: creating the docker client with a cid is not in this flow, and i think it is still correct

stoksc: but also fine with it now

PR #4870, opened 2022-08-25
feat: experiment checkpoint list [DET-8201] [DET-8129]

EmilyBonar:
## Description
Adds a Checkpoint tab to Experiment Details page.

## Test Plan
Check that checkpoints are listed correctly and that the checkpoint modal can be opened.

## Checklist

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
@@ -13,7 +13,7 @@ import useModalCheckpointRegister from 'hooks/useModal/Checkpoint/useModalCheckp
 import useModalModelCreate from 'hooks/useModal/Model/useModalModelCreate';
 import usePolling from 'hooks/usePolling';
 import useSettings, { UpdateSettings } from 'hooks/useSettings';
-import { getExperimentCheckpoints, openOrCreateTensorBoard } from 'services/api';
+import { getExperimentCheckpoints } from 'services/api';
 import { Determinedcheckpointv1State,
   V1GetExperimentCheckpointsRequestSortBy } from 'services/api-ts-sdk';
 import { encodeCheckpointState } from 'services/decoder';
@@ -22,11 +22,9 @@ import { ModalCloseReason } from 'shared/hooks/useModal/useModal';
 import { ErrorLevel, ErrorType } from 'shared/utils/error';
 import { validateDetApiEnum, validateDetApiEnumList } from 'shared/utils/service';
 import {
-  ExperimentAction as Action, CheckpointState,
-  CommandTask, CoreApiGenericCheckpoint, ExperimentBase,
+  CheckpointState, CoreApiGenericCheckpoint, ExperimentBase,
 } from 'types';
 import handleError from 'utils/error';
-import { openCommand } from 'utils/wait';
 
 import settingsConfig, { Settings } from './ExperimentCheckpoints.settings';
 import { columns as defaultColumns } from './ExperimentCheckpoints.table';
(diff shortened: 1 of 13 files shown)


webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+import settingsConfig, { Settings } from './ExperimentCheckpoints.settings';
+import { columns as defaultColumns } from './ExperimentCheckpoints.table';
+import css from './ExperimentTrials.module.scss';
+
+interface Props {
+  experiment: ExperimentBase;
+  pageRef: React.RefObject<HTMLElement>;
+}
+
+enum CheckpointAction {
```
ClaireNeveu, line 38: Just curious about this, aren't typescript enums sort of "soft deprecated" in favor or literal types?

ClaireNeveu: This would also let you use string literals like you want to below.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+    settings.sortDesc,
+    settings.sortKey,
+    settings.state,
+    settings.tableLimit,
+    settings.tableOffset,
+  ]);
+
+  const submitBatchAction = useCallback(async (action: CheckpointAction) => {
+    try {
+      // TODO: Actions
```
trentwatt, line 204: turn off row selection/ `TableBatch` for now?

image


batch registering of checkpoints would be pretty cool though IMO.

we could generate names for the model versions based on a user input + the different UUIDs or similar.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+      if (column.key === settings.sortKey) {
+        column.sortOrder = settings.sortDesc ? 'descend' : 'ascend';
+      }
+      return column;
+    });
+
+    return newColumns;
+  }, [ dropDownOnTrigger, experiment, settings.sortDesc, settings.sortKey, stateFilterDropdown ]);
+
+  const handleTableChange = useCallback((tablePagination, tableFilters, tableSorter) => {
```
trentwatt, line 148: question: why do we need to supply our own `handleTableChange` instead of just using the one already in `InteractiveTable`?

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+    const shouldPush = settings.tableOffset !== newSettings.tableOffset;
+    updateSettings(newSettings, shouldPush);
+  }, [ columns, settings.tableOffset, updateSettings ]);
+
+  const fetchExperimentCheckpoints = useCallback(async () => {
+    try {
+      const states = (settings.state ?? []).map((state) => (
+        encodeCheckpointState(state as CheckpointState)
+      ));
+      const response = await getExperimentCheckpoints(
```
trentwatt, line 170: looks like this API is a couple years old and this is our first time using it in webui.

apparently it does a full scan over the checkpoints for the filter and paginates the result set in memory. might be worth doing some performance validation/asking around.. ?

I think it's probably not the worst offender, but worth mentioning at least.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+    fetchExperimentCheckpoints,
+    settings.sortDesc,
+    settings.sortKey,
+    settings.state,
+    settings.tableLimit,
+    settings.tableOffset,
+  ]);
+
+  useEffect(() => {
+    if (terminalRunStates.has(experiment.state)) stopPolling({ terminateGracefully: true });
```
trentwatt, line 236: i think maybe checkpoints can be deleted (even automatically) after the experiment itself is in a terminal state?

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -0,0 +1,287 @@
...
+      } else if (column.key === 'actions') {
+        column.render = actionRenderer;
+      }
+      if (column.key === settings.sortKey) {
+        column.sortOrder = settings.sortDesc ? 'descend' : 'ascend';
+      }
+      return column;
+    });
+
+    return newColumns;
```
trentwatt, line 145: can we do maybe `searcherMetric` columns too? (and/or others?)

trentwatt (approved):
LGTM, apart from a couple small things,

deletion modal getting closed unexpectedly:
https://user-images.githubusercontent.com/28383080/186887808-1e771940-d78c-4333-9875-6f9f02f3ae68.mov

also theres no apparent effect when actually trying to delete the checkpoints. i dont know think it's an issue with this PR in particular (the delete request appears to go through ok), but would be good to figure out soon.

more thoughts below.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
EmilyBonar: I'm open to that. I used an enum here to mimic the pattern in the `ExperimentTrials` page

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
EmilyBonar: Again, copied from `ExperimentTrials`. Did some testing and you're right, it shouldn't be necessary.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
EmilyBonar: I disabled it for now, I can go back in a later PR and turn this on.

PR #4885, opened 2022-08-26
chore: allow OrderBy in GetExperimentCheckpoints for SortBy SearcherMetrics

NicholasBlaskey:
## Description

If OrderBy is unspecified for sortBy SearcherMetrics sort by best, otherwise follow ASC or DESC. 

## Test Plan

Call ```GetExperimentsCheckpoints``` with ```SearcherMetrics``` 

With ```OrderBy``` unspecified and make sure ```searcherMetrics``` in the response has "best" according to metric at the top

With ```OrderBy``` DESC and make sure ```searcherMetrics``` is in DESC order
With ```OrderBy``` ASC and make sure ```searcherMetrics``` is in ASC order

## Checklist

### DIFF
--- a/master/internal/api_experiment.go
+++ b/master/internal/api_experiment.go
@@ -642,15 +642,10 @@ func (a *apiServer) GetExperimentCheckpoints(
 		return nil, status.Errorf(codes.NotFound, "experiment %d not found", req.Id)
 	}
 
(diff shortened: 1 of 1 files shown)


EmilyBonar (approved):
LGTM!

PR #4894, opened 2022-08-29
fix: duplicate checkpoints returned by listing checkpoint routes

NicholasBlaskey:
## Description

Checkpoints listing can return duplicates. The reason for this is because in ```checkpoints_view``` there is a LEFT JOIN on ```raw_steps``` that assumes that for each ```trial_id```, ```total_batches``` we will have at most one unarchived value. 

```addValidationMetrics``` broke this by adding a dummy entry to ```raw_steps``` when there did not exist one and did not archive other older checkpoints. Solution was to remove this dummy entry when there was another checkpoint at total_batches that existed. 

## Test Plan

Integration test. 

Manual check that listing checkpoints isn't broken

Along with making sure GetExperimentCheckpoints ```1294``` at latest master returns no unique uuids

## Commentary (optional)

Hopefully the migration is decently fast, the select part feels so on latest master? Hopefully it can take advantage of the unique index. (Seems we are getting more index scans than not)

```
 Unique  (cost=1232434.89..1232841.87 rows=33413 width=12) (actual time=433.146..433.153 rows=18 loops=1)
   ->  Sort  (cost=1232434.89..1232536.64 rows=40698 width=12) (actual time=433.146..433.148 rows=18 loops=1)
         Sort Key: outer_steps.trial_id, outer_steps.total_batches, outer_steps.trial_run_id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using steps_archived on raw_steps outer_steps  (cost=0.55..1229318.92 rows=40698 width=12) (actual time=4.137..433.128 rows=18 loops=1)
               Index Cond: (archived = false)
               Filter: ((archived IS FALSE) AND (trial_run_id < (SubPlan 2)))
               Rows Removed by Filter: 120920
               SubPlan 2
                 ->  Result  (cost=8.45..8.46 rows=1 width=4) (actual time=0.003..0.003 rows=1 loops=120938)
                       InitPlan 1 (returns $2)
                         ->  Limit  (cost=0.42..8.45 rows=1 width=4) (actual time=0.002..0.002 rows=1 loops=120938)
                               ->  Index Only Scan Backward using steps_trial_id_total_batches_run_id_unique on raw_steps inner_steps  (cost=0.42..8.45 rows=1 width=4) (actual time=0.002..0.002 rows=1 loops=120938)
                                     Index Cond: ((trial_id = outer_steps.trial_id) AND (total_batches = outer_steps.total_batches) AND (trial_run_id IS NOT NULL))
                                     Heap Fetches: 6965
```

## Checklist

### CODE REVIEW

File: master/internal/db/postgres_trial.go [hunk 1]
```diff
@@ -261,6 +261,17 @@ VALUES
 func (db *PgDB) ensureStep(
 	ctx context.Context, tx *sqlx.Tx, trialID, trialRunID, stepsCompleted int,
 ) error {
+	var exists bool
+	if err := tx.QueryRowxContext(ctx, `
+SELECT EXISTS(
+	SELECT 1 FROM raw_steps WHERE trial_id = $1 AND total_batches = $2 AND archived IS NOT TRUE
```
stoksc, line 267: technically can just use the view but this is fine, too
```suggestion
	SELECT 1 FROM steps WHERE trial_id = $1 AND total_batches = $2
```

master/internal/db/postgres_trial_intg_test.go
```diff
@@ -62,3 +69,111 @@ func TestProtoGetTrial(t *testing.T) {
...
 	require.Equal(t, trResp.WallClockTime, float64(3), "wall clock time is wrong")
 }
+
+// Covers an issue where checkpoint_view returned multiple records per checkpoint
+// due to the LEFT JOIN raw_steps ON total_batches AND trial_id.
+// This is in this file because AddValidationMetrics broke the assumption the join uses.
+// That assumption being that for each trial_id and total_batches that there
+// is at most one unarchived result.
+func TestAddValidationMetricsDupeCheckpoints(t *testing.T) {
+	ctx := context.TODO()
```
stoksc, line 79: `context.Background()` is appropriate for tests (at least its documentation says so) and then this wont show up when i grep for `context.TODO()`

master/static/migrations/20220829162026_set-raw-metrics-archive.tx.up.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE raw_steps update_steps
+SET archived = TRUE
+FROM (SELECT DISTINCT trial_id, total_batches, trial_run_id FROM raw_steps outer_steps
+    WHERE archived IS FALSE AND 
+        trial_run_id < (SELECT MAX(inner_steps.trial_run_id) FROM raw_steps inner_steps WHERE
```
stoksc, line 5: in the bug, wasn't the "right" one the one with a lower run id?

also this feels like it would archive more than intended (would all distinct steps less with a run id less than the highest get archived?)

master/internal/db/postgres_trial_intg_test.go
NicholasBlaskey: That is a good point, I really didn't put any thought into doing ```context.TODO()``` vs ```context.Background```

master/static/migrations/20220829162026_set-raw-metrics-archive.tx.up.sql
NicholasBlaskey: Yep you are correct on both points.

The old query def would have updated more though I don't really understand what is wrong but no point in reasoning about incorrect code. Interesting on latest master there is only one record with this bug ```4099``` distinct UUIDs vs ```4100``` records. 

For the new query tried to use ```checkpoints_view``` to minimize the amount of ```raw_steps``` we have to touch

master/internal/db/postgres_trial.go
```diff
@@ -261,6 +261,16 @@ VALUES
...
+	var exists bool
+	if err := tx.QueryRowxContext(ctx, `
+SELECT EXISTS(SELECT 1 FROM steps WHERE trial_id = $1 AND total_batches = $2);`,
+		trialID, stepsCompleted).Scan(&exists); err != nil {
+		return err
+	}
+	if exists {
+		return nil
+	}
+
```
stoksc, line 273: style, up for debate, use a switch?

master/internal/db/postgres_trial_intg_test.go
```diff
@@ -62,3 +69,111 @@ func TestProtoGetTrial(t *testing.T) {
...
+
+	// Report training metrics.
+	require.NoError(t, db.AddTrainingMetrics(ctx, &trialv1.TrialMetrics{
+		TrialId:        int32(tr.ID),
+		TrialRunId:     0,
+		StepsCompleted: 50,
+		Metrics:        &commonv1.Metrics{AvgMetrics: trainMetrics},
+	}))
+
+	db.AddCheckpointMetadata(ctx, &model.CheckpointV2{
```
stoksc, line 117: require no error?

master/internal/db/postgres_trial_intg_test.go
```diff
@@ -62,3 +69,111 @@ func TestProtoGetTrial(t *testing.T) {
...
+	checkpoint2UUID := uuid.New()
+	valMetrics2, err := structpb.NewStruct(map[string]any{"loss": 1.5})
+	require.NoError(t, err)
+	require.NoError(t, db.AddValidationMetrics(ctx, &trialv1.TrialMetrics{
+		TrialId:        int32(tr.ID),
+		TrialRunId:     1,
+		StepsCompleted: 400,
+		Metrics:        &commonv1.Metrics{AvgMetrics: valMetrics2},
+	}))
+	db.AddCheckpointMetadata(ctx, &model.CheckpointV2{
```
stoksc, line 159: same

master/static/migrations/20220829162026_set-raw-metrics-archive.tx.up.sql
```diff
@@ -0,0 +1,16 @@
+UPDATE raw_steps update_steps
+SET archived = TRUE
+FROM ( -- Get max trial_run_id for each (trial_id, total_batches) pair
+    SELECT trial_id, total_batches, max(trial_run_id) AS trial_run_id FROM raw_steps
```
stoksc, line 4: i am slightly worried about the cost of this migration. can you dry run against latest master and get a plan/time estimate. the only thing worrying is the window over checkpoints view.

stoksc: if it is too expensive, we can just warn in the release note and not do it.

stoksc (commented):
final comments

master/internal/db/postgres_trial.go
NicholasBlaskey: Changed to a switch. 

I notice we use switches a lot where I would use if elses and I don't think I really understand the reason? Here I guess it keeps exists coupled with the query statement?

master/static/migrations/20220829162026_set-raw-metrics-archive.tx.up.sql
NicholasBlaskey: Dry run takes ~500ms on latest master with ~4000 checkpoints in the ```checkpoints_view```

master/internal/db/postgres_trial.go
stoksc: yeah, that's basically the reason for me. idk, "go style" (whatever) usually prefers switches to chained if else (if this was an else if instead, the linter would've said the same). part of it for me is just ease of reading code written in the same style.

master/static/migrations/20220829162026_set-raw-metrics-archive.tx.up.sql
stoksc: wonder when there are also a lot of metrics/vals since the checkpoints view.. sucks. but i'll take this i tihnk.

PR #4900, opened 2022-08-30
chore: change import path in generated bindings

rb-determined-ai:
Small missed detail in recent refactor.


hamidzr (approved):
from this change? 726fe804c8

PR #4901, opened 2022-08-30
feat: Mock of permissions API [DET-8215] [DET-8213]

mapmeld:
## Description

Adds three endpoints to the GRPC API:

~`/api/v1/permissions` [cluster-scope permissions]~
~`/api/v1/permissions/:workspace_id` [workspace-scope permissions]~
- `/api/v1/permissions/summary` [as described in RBAC ERD, return cluster roles and a list of workspace IDs in the assignments]

I added the first two for more granular requests. If it makes sense I can make it `/permissions/workspaces/:workspace_id` or something.

~The response as described in the ERD~

## Test Plan

Confirm that `/summary` gets a mocked response like this, matching DET-8213. I'm using `view_user_permissions` to help complete a future ticket.

## Checklist


mapmeld:
updated API response from `permissions/summary` based on the design shown in [DET-8213](https://determinedai.atlassian.net/browse/DET-8213)

I did change a part where the role name is used as an object key... this makes it difficult for us to return it from sql and type it in Go and Proto, so we just have the name inside each object in a list.

Screen Shot 2022-09-01 at 3 47 56 PM

gt2345 (commented):
When we have real data, should each element in `assignments` corresponds with an item in `roles`? Or how should we know what `Foo Editor` actually means

mapmeld:
> When we have real data, should each element in `assignments` corresponds with an item in `roles`?

Yes. I updated the SQL to return info about both roles in the roles and assignments sections, so it should make more sense now.

Screen Shot 2022-09-02 at 10 53 11 AM

mapmeld:
might be putting this on pause while we sort out https://github.com/determined-ai/determined-ee/pull/409

PR #4904, opened 2022-08-30
chore: test apex checkpointing [DET-7886]

drh-determined-ai:
## Description

Test checkpointing and restoring with a simple linear model in an Apex or PyTorch AMP context.

## Test Plan

Run unit tests with GPU(s) and with `apex` installed.

## Commentary (optional)

Because these tests rely on GPU(s), they are **not currently being run by CircleCI**.

## Checklist

### DIFF
--- a/harness/tests/experiment/fixtures/pytorch_onevar_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_onevar_model.py
@@ -238,6 +238,15 @@ class OneVarApexTrial(OneVarTrial):
             opt_level="O2",
         )
 
+    def train_batch(
+        self, batch: pytorch.TorchData, epoch_idx: int, batch_idx: int
+    ) -> Dict[str, torch.Tensor]:
+        scale_before = apex.amp.state_dict()["loss_scaler0"]["loss_scale"]
+        metrics = super().train_batch(batch, epoch_idx, batch_idx)
+        metrics["scale_before"] = scale_before
+        metrics["scale"] = apex.amp.state_dict()["loss_scaler0"]["loss_scale"]
+        return metrics
+
 
 class AMPTestDataset(OnesDataset):
     STAGE_DATUM = {
--- a/harness/tests/experiment/fixtures/pytorch_onevar_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_onevar_model.py
@@ -229,25 +229,6 @@ class OneVarTrial(pytorch.PyTorchTrial):
             raise ValueError(f"unknown dataloader_type: {self.hparams['dataloader_type']}")
 
 
-class OneVarApexTrial(OneVarTrial):
-    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
-        super().__init__(context)
-        self.model, self.optimizer = self.context.configure_apex_amp(
-            models=self.model,
-            optimizers=self.opt,
-            opt_level="O2",
-        )
-
-    def train_batch(
-        self, batch: pytorch.TorchData, epoch_idx: int, batch_idx: int
-    ) -> Dict[str, torch.Tensor]:
-        scale_before = apex.amp.state_dict()["loss_scaler0"]["loss_scale"]
-        metrics = super().train_batch(batch, epoch_idx, batch_idx)
-        metrics["scale_before"] = scale_before
-        metrics["scale"] = apex.amp.state_dict()["loss_scaler0"]["loss_scale"]
-        return metrics
-
-
 class AMPTestDataset(OnesDataset):
     STAGE_DATUM = {
         "one": 1.0,
--- a/harness/tests/experiment/fixtures/pytorch_onevar_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_onevar_model.py
@@ -229,6 +229,16 @@ class OneVarTrial(pytorch.PyTorchTrial):
             raise ValueError(f"unknown dataloader_type: {self.hparams['dataloader_type']}")
 
 
+class OneVarApexTrial(OneVarTrial):
+    def __init__(self, context: pytorch.PyTorchTrialContext) -> None:
+        super().__init__(context)
+        self.model, self.optimizer = self.context.configure_apex_amp(
+            models=self.model,
+            optimizers=self.opt,
+            opt_level="O2",
+        )
+
+
 class AMPTestDataset(OnesDataset):
     STAGE_DATUM = {
         "one": 1.0,
--- a/harness/tests/experiment/pytorch/test_pytorch_trial.py
+++ b/harness/tests/experiment/pytorch/test_pytorch_trial.py
@@ -148,35 +148,7 @@ class TestPyTorchTrial:
         )
         controller.run()
 
-    @pytest.mark.gpu
-    @pytest.mark.parametrize(
-        "trial_class",
-        [
-            pytorch_xor_model.XORTrialWithLRScheduler,
-            pytorch_onevar_model.OneVarApexAMPTrial,
-            pytorch_onevar_model.OneVarAutoAMPTrial,
-            pytorch_onevar_model.OneVarManualAMPTrial,
-        ],
-        ids=[
-            "default",
-            "apex",
-            "autocast",
-            "manual",
-        ],
-    )
-    def test_checkpointing_and_restoring(self, trial_class, tmp_path: pathlib.Path) -> None:
-        utils_test_kwargs = {}
-        testing_amp = issubclass(trial_class, pytorch_onevar_model.OneVarAMPBaseTrial)
-        if testing_amp:
-            if not torch.cuda.is_available():
-                pytest.skip("no gpu available")
-            utils_test_kwargs.update({
-                "steps": (2, 2),
-                "scheduling_unit": 1,
-            })
-        if trial_class is pytorch_onevar_model.OneVarApexAMPTrial and not HAVE_APEX:
-            pytest.skip("Apex not available")
-
+    def test_checkpointing_and_restoring(self, tmp_path: pathlib.Path) -> None:
         def make_trial_controller_fn(
             workloads: workload.Stream,
             checkpoint_dir: typing.Optional[str] = None,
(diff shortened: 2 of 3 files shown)


harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -229,6 +229,16 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
             raise ValueError(f"unknown dataloader_type: {self.hparams['dataloader_type']}")
 
 
+class OneVarApexTrial(OneVarTrial):
```
rb-determined-ai, line 232: Why is this distinct from `OneVarApexAMPTrial`?

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -657,6 +657,55 @@ def test_amp(self, trial_class) -> None:
...
+                expose_gpus=True,
+            )
+            controller.run()
+
+        for c in range(1, num_checkpoints):
+            prev_checkpoint_last_loss = training_metrics[c - 1][-1]["loss"]
+            this_checkpoint_first_loss = training_metrics[c][0]["loss"]
+            assert (
+                this_checkpoint_first_loss < prev_checkpoint_last_loss
+            ), f"loss did not decrease after checkpoint {c-1}"
```
rb-determined-ai, line 707: I don't think this is a strong enough assertion.

It is true that the bug we're investigating is about "loss spikes after checkpoint restoring" or something.  But due to the linear model, stable inputs, and simple SGD loss function, I would be surprised to _ever_ see the loss increase, even if we totally messed up saving the scaler state.

Can you instead write the test so that it directly measures that the state of the scaler was properly saved/restored?  I can think of two ways to do it, either:
- direct inspection of scaler state, or
- inferring scaler state by expecting precise loss values, as with the last test you wrote

In either case, you'll need to make sure that the scaler finishes each run in a different state than its initial state, otherwise there's no way to tell if things were preserved or not.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: I don't want the wacky data for this test as we had for the autocasting & scaling tests; I want to be able to expect the loss to decrease monotonically. The trial class names are not descript as they are right now.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Anda had a similar concern in one of our calls, I think. What I can do is check if the scale value changed, and if it did check that there is a good, apparent reason (similar to what we did in the AMP tests).

An obvious problem we'd catch is that the scaler value resets to whatever initial value upon loading the state.

drh-determined-ai: Actually, scratch all that. That's just doing the same test as before. What I'm writing now and will push soon is this: the scale value just before saving a checkpoint should be the same as just after loading that checkpoint.

drh-determined-ai:
Note that there is the existing test `TestPyTorchTrial.test_checkpointing_and_restoring`. An alternative approach to this task would be to parameterize the `OneVarTrial` subclass (similarly as is being done for `test_amp`).

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -657,6 +657,61 @@ def test_amp(self, trial_class) -> None:
...
+        checkpoint_dir = str(tmp_path.joinpath("apex_checkpoint"))
+        latest_checkpoint = None
+        steps_completed = 0
+
+        def make_workloads(checkpoint_idx: int = 0) -> workload.Stream:
+            # Train as usual
+            trainer = utils.TrainAndValidate()
+            # In order to expect the loss to decrease after loading the first checkpoint,
+            #  make sure the scaler has had enough steps to stop reducing so that training
+            #  can actually proceed.
```
rb-determined-ai, line 677: how about just simplify the test so it doesn't make the "monotonic decreasing" assertion?  Because this is a unit test, it is reasonable that it only focuses on exactly one detail (preservation of scaler state, for instance).

The other test you wrote provides very good coverage of the loss behavior, but no coverage of scaler preservation around checkpoints.  This test could very reasonably provide the scaler preservation coverage without having to think about loss behavior.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -657,6 +657,61 @@ def test_amp(self, trial_class) -> None:
...
+            yield from trainer.send(steps=10, validation_freq=10)
```
rb-determined-ai, line 678: nit: why even validate at all?  Validation should not affect the scaler at all.

drh-determined-ai: I would agree, but `validation_freq` isn't an optional parameter, so I used the value that minimizes the time spent validating without needing clarification (like, I didn't put `11` or `inf`).

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -657,6 +657,61 @@ def test_amp(self, trial_class) -> None:
         )
         controller.run()
 
+    @pytest.mark.skipif(not HAVE_APEX, reason="Apex not available")
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="no gpu available")
+    @pytest.mark.gpu
+    def test_apex_checkpointing(self, tmp_path: pathlib.Path) -> None:
+        # There will be num_checkpoints-1 assertions made at the end of this test
+        num_checkpoints = 5
```
rb-determined-ai, line 665: question: what does num_checkpoints=5 buy us that num_checkpoints=2 wouldn't buy us?

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: that sounds good.  Can you also include an assertion to guarantee that the scaler state change over time?  I.e. that the test isn't passing on accident due to the scaler state always being the initial value.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: huh, I didn't realize we couldn't not validate.  Makes sense then.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I could, but that'll make the test resemble the AMP autoscaling test.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: Because the scaler state changes during training, so reloading the same state over and over doesn't really prove anything

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: Cool, so let's pick to the minimum num_checkpoints to achieve full coverage.  Otherwise we're just wasting time in tests.

harness/tests/experiment/pytorch/test_pytorch_trial.py
rb-determined-ai: I think avoiding false positives in tests is more important than the DRY principle.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: This code is gone.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: We are now using the same assertion code as is being used in the AMP scaling tests.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -229,6 +229,25 @@ def build_validation_data_loader(self) -> torch.utils.data.DataLoader:
...
```
rb-determined-ai, line 232: OneVarApexTrial is no longer used, right?

drh-determined-ai: Ah, correct!

harness/tests/experiment/utils.py
```diff
@@ -348,7 +348,10 @@ def make_workloads(steps: int) -> workload.Stream:
 
 
 def checkpointing_and_restoring_test(
```
rb-determined-ai, line 350: I like the change to this function.  Very neat.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -148,7 +148,35 @@ def make_workloads() -> workload.Stream:
...
+            pytorch_onevar_model.OneVarManualAMPTrial,
+        ],
+        ids=[
+            "default",
+            "apex",
+            "autocast",
+            "manual",
+        ],
+    )
+    def test_checkpointing_and_restoring(self, trial_class, tmp_path: pathlib.Path) -> None:
```
rb-determined-ai, line 167: But this test is now very hard to read.  There are just too many if statements.  To me, that's a sign that the code paths are not actually similar enough to be combined.  Let's leave the original test alone and stick with a separate AMP test.

drh-determined-ai: I agree 100%. I was able to tell you my same feelings!

rb-determined-ai (approved):
Great, I love this new test!

azhou-determined (approved):
looks great. this adds a needed test to our repository and also confirms that we don't need to fix that long-standing AMP checkpointing bug. you killed 2 JIRAs with one PR. 🥇

PR #4905, opened 2022-08-30
refactor: AuthZ for experiments [DET-8003]

NicholasBlaskey:
## Description

Implements AuthZ pattern for experiments. 

## Test Plan

integration tests and existing e2e tests 

## Checklist


master/internal/api_experiment.go
```diff
@@ -28,46 +28,49 @@ import (
...
-		return nil, status.Errorf(codes.NotFound, "experiment not found: %d", experimentID)
-	case err != nil:
-		return nil, errors.Wrapf(err,
-			"error fetching experiment from database: %d", experimentID)
+	modelExp, err := model.ExperimentFromProto(exp)
+	if err != nil {
+		return nil, err
+	}
+	if ok, err := expauth.AuthZProvider.Get().
+		CanGetExperiment(curUser, modelExp); err != nil {
```
ioga, line 70: idea: maybe `CanGetExperiment` can take `model.ExperimentId` as the 2nd parameter. this allows to check the permission first, and if that succeeds fetch the experiment.

master/internal/api_experiment.go
```diff
@@ -1047,9 +1193,18 @@ func (a *apiServer) MetricBatches(req *apiv1.MetricBatchesRequest,
 		period = defaultMetricsStreamPeriod
 	}
 
+	var timeSinceLastAuth time.Time
 	seenBatches := make(map[int32]bool)
 	var startTime time.Time
 	for {
+		if time.Now().Sub(timeSinceLastAuth) >= recheckAuthPeriod {
```
ioga, line 1200: nice

master/internal/api_experiment.go
```diff
@@ -673,16 +764,25 @@ func (a *apiServer) UnarchiveExperiment(
...
+		return nil, err
+	}
+	modelExp, err := model.ExperimentFromProto(exp)
+	if err != nil {
+		return nil, err
 	}
 
 	madeChanges := false
 	if req.Experiment.Name != nil && exp.Name != req.Experiment.Name.Value {
+		if err = expauth.AuthZProvider.Get().CanSetExperimentsName(*curUser, modelExp); err != nil {
```
ioga, line 782: I am worried this is way too granular. what's the case for allowing setting experiment name, but not the notes? 

maybe have `CanEditExperiment` for writes on general metadata and `CanEditExperimentArtifacts` for writes on "sensitive" stuff like the code or checkpoints. that'd be sufficient to map to our initial roles. we can break it down later. that'd also keep the authz API surface smaller.

on second thought, RBAC AuthZ implementation could do the same simplification internally (e.g. all of `CanSetExperimentsName`, `CanPauseExperiment` all calling internal `CanEditExperiment`).

master/internal/api_experiment.go
```diff
@@ -1687,13 +1882,17 @@ func (a *apiServer) GetModelDef(
 func (a *apiServer) MoveExperiment(
 	ctx context.Context, req *apiv1.MoveExperimentRequest,
 ) (*apiv1.MoveExperimentResponse, error) {
-	curUser, _, err := grpcutil.GetUser(ctx, a.m.db, &a.m.config.InternalConfig.ExternalSessions)
+	// get experiment info
+	exp, curUser, err := a.getExperimentAndCheckCanDoActions(ctx, int(req.ExperimentId), false)
```
ioga, line 1886: maybe this needs `CanEditExperiment` (or similar) on the source experiment.

ioga (commented):
let's talk on how to find the right granularity level

master/internal/api_experiment.go
NicholasBlaskey: My viewpoint is that I want to avoid as much queries / stuff in the ```AuthZ``` implementations so that we don't need to implement the same getting experiment logic across both implementations.

master/internal/api_experiment.go
NicholasBlaskey: I agree it is way too excessive now, I'll merge a bunch of ```AuthZ``` methods together now. I think we could do the simplification internally but that seems like more work for no benefit. 

I do think for some cases like ```CanSetExperimentsWeight(curUser model.User, e *model.Experiment, weight float64)``` some people would be happy to see the feature to limit scheduling priority ranges to groups eventually

NicholasBlaskey: Ok let me know what you think of the new smaller interface.

harness/determined/exec/prep_container.py
```diff
@@ -13,18 +13,25 @@
...
     trial_info = det.TrialInfo._from_env()
     trial_info._to_file()
 
+    auth = authentication.Authentication(
+        info.master_url,
+        util.get_det_username_from_env(),
+        None,
+        True,
+        cert,
+    )
```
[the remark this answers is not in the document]
NicholasBlaskey: That makes sense to me. 

I think that session is still using ```x-allocation-token``` so I need to look more into this and if there any issues switching to a user token session

master/internal/api_experiment.go
ioga: looks good 👍 

my primary concern is the permissions in the world where we allow custom roles. if we go coarse today and need to add granularity later, we can do a SQL migration where coarse `get_experiment_artifacts` permission is split into `get_experiment_code` and `get_experiment_checkpoints`. this won't be breaking change.

master/internal/api_experiment.go
ioga: ok 👍

harness/determined/exec/prep_container.py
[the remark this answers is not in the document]
ioga: yes, once we have consistent way to auth it'd be easier to change it

ioga (approved):
lgtm

harness/determined/exec/prep_container.py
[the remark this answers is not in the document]
NicholasBlaskey: Making ```add_tokens_to)headers``` add a header for ```DET_USER_SESSION``` seems to make tests pass

Checkpoint GC I don't believe has a ```DET_USER_SESSION``` but those tests still seem to pass, skimming over the code it seems like checkpoint GC does not not even need to talk to the master API directly at all? Is that correct?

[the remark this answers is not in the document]
NicholasBlaskey: Also there is still a decent amount of code that if we decide we don't need any of the ```x-allocation-token``` we should be able to delete in a follow up PR

PR #4914, opened 2022-08-30
chore: clean up det e create cli [MLG-44]

rb-determined-ai:
## Description

The original intention of experiment.py was a place for reusable code to
interact with the REST API.  The generated bindings now serve that
purpose, and experiment.py is just tech debt.

The contents of experiment.py met with the following fates:

- create_experiment: replaced with bindings
- create_experiment_and_follow_logs: deleted, it served no purpose
- create_test_experiment_and_follow_logs: deleted, it served no purpose
- patch_experiment: moved to cli/experiment.py, until all of its callers
  are replaced with bindings
- follow_experiment_logs: moved to cli/experiment.py where it is used
- follow_test_experiment_logs:  moved to cli/experiment.py where it is used
- generate_random_hparam_values: moved to _execution.py where it belongs
- make_test_experiment_config: moved to _execution.py where it belongs

Additionally, this change brings the /api/v1 experiment creation to
feature parity with the old echo endpoint, and migrates the CLI to use
the /api/v1 endpoint.

After this change, it should be possible to delete the old API endpoint,
though it might be prudent to wait a month or two for slow-to-update cli
users.

## Test Plan

- [x] If there is one thing our automated tests have covered really well, I think it is experiment submission.


e2e_tests/tests/cluster/test_users_experiment_api.py
```diff
@@ -1,53 +1,25 @@
...
 
     determined_master = conf.make_master_url()
-    user_creds = create_test_user(add_password=True)
+    user_creds = test_users.create_test_user(add_password=True)
 
     try:
-        det_spawn(["user", "deactivate", "determined"])
-        certs.cli_cert = certs.default_load(
-            master_url=determined_master,
+        d = client.Determined(determined_master, user_creds.username, user_creds.password)
```
ioga, line 18: I think this test is supposed to be disabling the default `determined` user, and it got lost in refactoring

harness/determined/_execution.py
```diff
@@ -43,6 +45,75 @@ def _catch_sys_exit() -> Any:
...
+                # In this case we have a dictionary of nested hyperparameters.
+                return _generate_random_hparam_values(hparam)
+            elif hparam["type"] == "const":
+                return hparam["val"]
+            elif hparam["type"] == "int":
+                return random.randint(hparam["minval"], hparam["maxval"])
+            elif hparam["type"] == "double":
+                return random.uniform(hparam["minval"], hparam["maxval"])
+            elif hparam["type"] == "categorical":
+                return hparam["vals"][random.randint(0, len(hparam["vals"]) - 1)]
```
ioga, line 61:
```suggestion
                return random.choice(hparam["vals"])
```

proto/src/determined/api/v1/experiment.proto
```diff
@@ -308,7 +309,18 @@ message CreateExperimentRequest {
...
   int32 project_id = 6;
+  // Template to use for the experiment.
+  string template = 7;
+  // Git remote at submission time.
+  string git_remote = 8;
+  // Git commit at submission time.
+  string git_commit = 9;
+  // Git committer at submission time.
+  string git_committer = 10;
+  // Git commit date at submission time.
```
ioga, line 320: is it possible to make these all `optional`, and get rid of `emptyStrToNil` conversion?

e2e_tests/tests/cluster/test_users_experiment_api.py
rb-determined-ai: excellent catch, will fix

ioga (approved):

proto/src/determined/api/v1/experiment.proto
rb-determined-ai: is it?  I can do that

PR #4925, opened 2022-08-31
feat: pin experiments

keita-determined:
## Description
[DET-8135]

## Test Plan
- Check if experiment item can be pinned or unpinned
- Check if experiment items stick to the top of table
- Check if there are pinned items on each pagination page on top
- Check if pagination works well
- Check if pinned experiment should be removed after Move Action
- Check if pin experiments up to 5 in each project 

## Commentary (optional)
- Netlify doesn't work as expected cuz i changed backend 

## Checklist

[DET-8135]: https://determinedai.atlassian.net/browse/DET-8135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


proto/src/determined/api/v1/experiment.proto
```diff
@@ -43,6 +43,23 @@ message GetExperimentResponse {
 
 // Get a list of experiments.
 message GetExperimentsRequest {
+  // filtering by experiment ids
+  message ExperimentFilter {
```
keita-determined, line 47: I dont have good naming sense. Any better name?

webui/react/src/components/ExperimentActionDropdown.tsx
```diff
@@ -98,6 +104,17 @@ const ExperimentActionDropdown: React.FC<Props> = ({
...
         }
+        case Action.Pin: {
+          const pinList = [ ...settings.pinned ];
+          pinList.push(id);
+          updateSettings({ pinned: [ ...Array.from(new Set(pinList)) ] });
+          break;
+        }
+        case Action.Unpin: {
+          const pinList = settings.pinned.filter((p) => p !== id);
+          updateSettings({ pinned: [ ...Array.from(new Set(pinList)) ] });
```
EmilyBonar, line 115: perf: Since you're going to be converting it into a Set anyway to dedupe the Array, it would be faster to do something like this:
```
const pinList = new Set(settings.pinned);
pinList.delete(id);
updateSettings({ pinned: [ ...pinList ] });
```
Benchmarked here: https://jsben.ch/zXESV

webui/react/src/components/ExperimentActionDropdown.tsx
```diff
@@ -98,6 +104,17 @@ const ExperimentActionDropdown: React.FC<Props> = ({
           openCommand(tensorboard);
           break;
         }
+        case Action.Pin: {
+          const pinList = [ ...settings.pinned ];
+          pinList.push(id);
+          updateSettings({ pinned: [ ...Array.from(new Set(pinList)) ] });
```
EmilyBonar, line 110: perf: I didn't benchmark this one, and my gut says it's less important since you're not filtering the array like you are in unpin, but the following will still probably be faster:
```
const pinList = new Set(settings.pinned);
pinList.add(id);
updateSettings({ pinned: [ ...pinList ] });

File: webui/react/src/components/ExperimentActionDropdown.tsx [hunk 4]
```diff
@@ -157,9 +174,15 @@ const ExperimentActionDropdown: React.FC<Props> = ({
...
-    { danger: action === Action.Delete, key: action, label: action }
-  ));
+  const menuItems = getActionsForExperiment(experiment, dropdownActions, curUser)
+    .filter((action) => {
+      const isPinned = settings.pinned.includes(experiment.id);
+      if ((isPinned && action === Action.Pin) || (!isPinned && action === Action.Unpin)) {
+        return false;
+      } else {
+        return true;
+      }
```
EmilyBonar, line 184: perf: combining "pin" and "unpin" into "switch pin" would allow you to get rid of this loop

webui/react/src/components/ExperimentActionDropdown.tsx
```diff
@@ -189,7 +212,8 @@ const ExperimentActionDropdown: React.FC<Props> = ({
     <div className={css.base} title="Open actions menu" onClick={stopPropagation}>
       <Dropdown overlay={menu} placement="bottomRight" trigger={[ 'click' ]}>
         <button onClick={stopPropagation}>
-          <Icon name="overflow-vertical" />
+          {settings.pinned.includes(id) ?
+            <PushpinOutlined /> : <Icon name="overflow-vertical" />}
```
EmilyBonar, line 216: nit: Feels kinda weird to me that you click on the pin to bring up the overflow menu? If this went through alex then don't worry about it, just thought I'd mention

webui/react/src/pages/ProjectDetails.module.scss
```diff
@@ -16,6 +15,15 @@
...
 }
+.pinnedContainer {
+  align-items: center;
+  display: flex;
+  flex-wrap: nowrap;
+  gap: 12px;
+  overflow-x: scroll;
+  overflow-y: none;
+  padding: 20px 0;
+}
```
EmilyBonar, line 26: nit: as far as I can tell this is unused

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -162,42 +164,58 @@ const ProjectDetails: React.FC = () => {
       const states = (settings.state || []).map((state) => (
         encodeExperimentState(state as RunState)
       ));
-      const response = await getExperiments(
+      const experimentFilter: GetExperimentsParams = {
```
EmilyBonar, line 167: super nit: maybe name this something like `baseParams` instead

webui/react/src/pages/ProjectDetails.settings.ts
```diff
@@ -64,6 +65,12 @@ export interface ProjectDetailsSettings extends InteractiveTableSettings {
 
 const config: SettingsConfig = {
   settings: [
+    {
+      defaultValue: [],
+      key: 'pinned',
+      storageKey: 'pinned',
+      type: { baseType: BaseType.Integer, isArray: true },
+    },
```
EmilyBonar, line 73: nit: I think this should probably have `skipUrlEncoding: true,`

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -162,42 +164,58 @@ const ProjectDetails: React.FC = () => {
...
-          name: settings.search,
-          offset: settings.tableOffset,
-          orderBy: settings.sortDesc ? 'ORDER_BY_DESC' : 'ORDER_BY_ASC',
-          projectId: id,
-          sortBy: validateDetApiEnum(V1GetExperimentsRequestSortBy, settings.sortKey),
-          states: validateDetApiEnumList(Determinedexperimentv1State, states),
-          users: settings.user,
+          ...experimentFilter,
+          experimentFilter: { experimentIds: settings.pinned, restriction: 'RESTRICTION_INCLUDE' },
+          limit: 0,
```
EmilyBonar, line 181: fix: the backend team is planning on potentially removing support for `limit: 0`, so we probably shouldn't rely on it. More info: https://determined-ai.slack.com/archives/C02PV33GSN5/p1661546385122749

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -162,42 +164,58 @@ const ProjectDetails: React.FC = () => {
...
-      setTotal(response.pagination.total ?? 0);
-      setExperiments((prev) => {
-        if (isEqual(prev, response.experiments)) return prev;
-        return response.experiments;
-      });
+      const otherExperimentsRequest = getExperiments(
+        {
+          ...experimentFilter,
+          experimentFilter: { experimentIds: settings.pinned, restriction: 'RESTRICTION_EXCLUDE' },
+          limit: settings.tableLimit - settings.pinned.length,
```
EmilyBonar, line 190: fix: If you have more pinned experiments than your current table limit this goes negative, which doesn't have well defined behavior

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -162,42 +164,58 @@ const ProjectDetails: React.FC = () => {
...
+          offset: settings.tableOffset -
+            (settings.tableOffset / settings.tableLimit) * settings.pinned.length,
+        },
+        { signal: canceler.signal },
+      );
+      const allRequests = [ pinnedExperimentsRequest, otherExperimentsRequest ];
+      const [ pinnedExpResponse, otherExpResponse ] = await Promise.all(allRequests);
+      setTotal(otherExpResponse.pagination.total ?? 0);
```
EmilyBonar, line 198: fix: pagination gets messed up if you have too many experiments pinned compared to total experiments. On my local cluster I have 15 experiments and pinned 11 of them. I'm only able to see the 10 of the pinned experiments and I have no pagination controls or way to see any other experiments

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -883,6 +902,12 @@ const ProjectDetails: React.FC = () => {
             ContextMenu={ContextMenu}
             dataSource={experiments}
             loading={isLoading}
+            numOfPinned={
+              Array.from(intersection(
+                new Set(experiments.map((e) => e.id)),
+                new Set(settings.pinned),
+              )).length
+            }
```
EmilyBonar, line 910: question: why can't this just be `settings.pinned.length`?

EmilyBonar (changes requested):
Looks cool so far, but needs some more polish and testing when it comes to table and pagination issues

webui/react/src/utils/experiment.ts
```diff
@@ -172,6 +172,10 @@ const experimentCheckers: Record<ExperimentAction, ExperimentChecker> = {
   [ExperimentAction.ViewLogs]: alwaysTrueExperimentChecker,
 
   [ExperimentAction.CompareExperiments]: alwaysTrueExperimentChecker,
+
+  // cannot access to settings.pinned here, so manage visibility in each place
+  [ExperimentAction.Pin]: alwaysTrueExperimentChecker,
+  [ExperimentAction.Unpin]: alwaysTrueExperimentChecker,
```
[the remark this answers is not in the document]
keita-determined: I cannot call hooks in `ts` file.

webui/react/src/components/ExperimentActionDropdown.tsx
keita-determined: thats a nice website. 
I agree. I worked on it at night. My brain was dead :( 
Thanks for the pointing out

webui/react/src/components/ExperimentActionDropdown.tsx
keita-determined: Its just my idea. I wanted to change icon when its hovered but I forgot to do that. But let me ask him first

webui/react/src/pages/ProjectDetails.tsx
keita-determined: I'll add limit in pin api call

webui/react/src/pages/ProjectDetails.tsx
keita-determined: Because of the projects. I store all pinned items in the userSettings array. 
I originally wanted to use key value data type but userSettings doesn't support it. 
That's one of reasons why I wanna recreate userSettings

webui/react/src/utils/experiment.ts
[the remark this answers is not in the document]
keita-determined: That's a good idea. My concern is that how do users know which one is pinned or not. If we can solve the issue `Switch Pin` sounds better

webui/react/src/pages/ProjectDetails.settings.ts
keita-determined: can i ask you a reason why?

webui/react/src/pages/ProjectDetails.tsx
keita-determined: I thought I fixed the issue. Let me take a look

proto/src/determined/api/v1/experiment.proto
ioga, line 47: I have a question about REST API design here. 
this ends up generating URLs like this: `/api/v1/experiments?sortBy=SORT_BY_ID&orderBy=ORDER_BY_ASC&offset=0&limit=6&archived=false&projectId=104&experimentFilter.restriction=RESTRICTION_EXCLUDE&experimentFilter.experimentIds=125&experimentFilter.experimentIds=126&experimentFilter.experimentIds=34&experimentFilter.experimentIds=105`

it's a bit verbose. some clients / systems sometimes have an upper limit on the URL length (at ~2k characters IIRC). this URL format spends `30+len(experiment_id)` characters per experiment, so it'd hit a limit at 66 experiments which doesn't sound unrealistic to me.

if you were building this REST API without the luggage of grpc-gateway and its enums, how would you want this API URL to look like? 

one possible suggestion here is to do something like this: `/api/v1/experiments?sortBy=SORT_BY_ID&orderBy=ORDER_BY_ASC&offset=0&limit=6&archived=false&projectId=104&experimentIds__in=125,126,34,105`, i.e. "lookup field" and "lookup operation" in a single query parameter, separated by double underscore. 
- for "exclusion" it could look like `experimentIds__notin=125,126,34,105`
- this provides a basis for other operations like `gte` for greater or equal, `lt` for less than, etc. allowing a generic search operations such as `?experimentIds__lte=100&experimentIds__gt=50` for looking up `50 < experiment id <= 100`, as an example.

we currently do not have examples of this approach implemented in the code but it may be worth investing in for the sake of good API.

master/internal/api_experiment.go
```diff
@@ -390,6 +390,20 @@ func (a *apiServer) GetExperiments(
...
+		expRestriction := req.ExperimentFilter.Restriction.String()
+		experimentIds := make([]int32, 0)
+		if req.ExperimentFilter.ExperimentIds != nil {
+			experimentIds = append(experimentIds, req.ExperimentFilter.ExperimentIds...)
+		}
+		if expRestriction == apiv1.GetExperimentsRequest_ExperimentFilter_Restriction_name[1] {
+			query = query.Where("e.id = ANY (ARRAY[?]::integer[])", bun.In(experimentIds))
+		} else if expRestriction == apiv1.GetExperimentsRequest_ExperimentFilter_Restriction_name[2] &&
+			len(experimentIds) > 0 {
+			query = query.Where("NOT (e.id = ANY (ARRAY[?]::integer[]))", bun.In(experimentIds))
```
ioga, line 403:
```suggestion
		if expRestriction == apiv1.GetExperimentsRequest_ExperimentFilter_Restriction_name[1] {
			query = query.Where("e.id IN (?)", bun.In(experimentIds))
		} else if expRestriction == apiv1.GetExperimentsRequest_ExperimentFilter_Restriction_name[2] &&
			len(experimentIds) > 0 {
			query = query.Where("e.id  NOT IN (?)", bun.In(experimentIds))
```

keita-determined: I use `ANY (ARRAY[?]::integer[]` to handle empty list just in case. Or do i need to handle it in Go instead of SQL?

proto/src/determined/api/v1/experiment.proto
keita-determined: I originally wanted to have the one you suggested `ids=1,2,3`. 
Let me take a look at how REST API is generated

webui/react/src/pages/ProjectDetails.settings.ts
EmilyBonar: It makes the URL pretty long, but more importantly it would be very weird if someone sent you a link to a project that ended up overwriting your pinned experiments.

master/internal/api_experiment.go
ioga: perhaps skip doing `where` if it's empty?

keita-determined: ok I'll change Go code instead of SQL then

webui/react/src/pages/ProjectDetails.settings.ts
keita-determined: Ah I thought its for avoiding encoding the data. Its for avoiding preserving setting in the URL query param.
That makes sense.
Thanks

master/internal/api_experiment.go
keita-determined: note:
I just remember why I implemented this way. If there's no item in array with `INCLUDE`, return 0 experience.  If there's no item in array with `EXCLUDE`, return all experiences.
That was my intention. 
I'll change the code with `IN` but its just a note

ioga: is there a valid use case for having no items in include list? maybe that should just throw an error, or return an empty response early?
constructing and running a query for that seems unnecessary.

keita-determined: I can't come up with any valid use case but I just thought it's intuitive. 
> maybe that should just throw an error, or return an empty response early?

I'll go with this way

webui/react/src/components/ExperimentActionDropdown.tsx
keita-determined: I talked to alex and added border in the last item

webui/react/src/pages/ProjectDetails.tsx
keita-determined: I added limit for the number of pin items which is 5.

proto/src/determined/api/v1/experiment.proto
keita-determined: In [the documentation ](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#rules-for-http-mapping), 
> All other fields are passed via the URL query parameters, and the parameter name is the field path in the request message. A repeated field can be represented as multiple query parameters under the same name.

We might not able to change the style of param query in generated REST API.
I'll search more but if we cant change it all we can do with repeated type is to use shorter variable name or use post as get but its not ideal.

keita-determined:
@EmilyBonar 
- Add limit up to 5 pinned items <- that fixes pagination issue
- Replaced `Pin` and `Unpin` with `SwitchPin`
- Change data structure of `pinned` (KV type)
- Remove pin icon. Instead, add border in the end of pinned items
- Support `Move` action

master/internal/api_experiment.go
keita-determined: >I just remember why I implemented this way. If there's no item in array with INCLUDE, return 0 experience. If there's no item in array with EXCLUDE, return all experiences.
That was my intention. IN doesnt accept empty array, that's why I used it
I'll change the code with IN but its just a note

Just remembered (again) that why I wanted it to be behaved like this.
It's because some projects could be empty or there could be no pinned item in a project. In these cases, this behavior is useful. 
I rewrote the code to archive the behavior with `IN` instead of using `ANY` clause, but the code became a bit uglier

webui/react/src/components/ExperimentActionDropdown.tsx
```diff
@@ -97,6 +104,25 @@ const ExperimentActionDropdown: React.FC<Props> = ({
           openCommand(tensorboard);
           break;
         }
+        case Action.SwitchPin: {
```
EmilyBonar, line 107: fix: GitHub won't let me put this comment there, but at line 194, you should check whether

1. `action === Action.SwitchPin`
and if so,
2. `label: pinSet.has(id) ? 'Unpin' : 'Pin'`

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -167,42 +168,55 @@ const ProjectDetails: React.FC = () => {
...
+      const [ pinnedExpResponse, otherExpResponse ] = await Promise.all(allRequests);
+      setTotal(
+        (pinnedExpResponse.pagination.total ?? 0) + (otherExpResponse.pagination.total ?? 0),
       );
-      setTotal(response.pagination.total ?? 0);
-      setExperiments((prev) => {
-        if (isEqual(prev, response.experiments)) return prev;
-        return response.experiments;
-      });
+      setExperiments([ ...pinnedExpResponse.experiments, ...otherExpResponse.experiments ]);
```
EmilyBonar, line 200: fix: There's some sort of race condition when it comes to pinning/unpinning and switching pages

https://user-images.githubusercontent.com/15078396/189415696-dec0a744-38df-492c-905b-ea6d062d9de6.mov

EmilyBonar (approved):
Fix these last couple things and you should be good

webui/react/src/hooks/useModal/Experiment/useModalExperimentMove.tsx
```diff
@@ -260,6 +266,15 @@ const useModalExperimentMove = ({ onClose }: Props): ModalHooks => {
         ),
         message: 'Move Success',
       });
+      if (sourceProjectId) {
+        const newPinned = { ...projectSettings.pinned };
```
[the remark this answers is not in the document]
keita-determined: What I want here is to deep copy only top level cuz its more efficient. I'll keep it instead of using clone

webui/react/src/pages/ProjectDetails.tsx
keita-determined: I think its `useSettings` bug.  In DB, it is stored as expected but `useSettings` looks at the old data

proto/src/determined/api/v1/experiment.proto
ioga: push comes to shove, it can be a string in proto that is split by `,` on server side.

keita-determined:
Wait for the backend change to archive more complex filtering 
Ref: https://determined-ai.slack.com/archives/CKQ3EV0G6/p1662755531081139

proto/src/determined/api/v1/experiment.proto
```diff
@@ -122,7 +107,7 @@ message GetExperimentsRequest {
   // projects.
   int32 project_id = 12;
   // filtering by experiment ids
-  ExperimentFilter experiment_filter = 13;
+  determined.common.v1.Int32FieldFilter experiment_filter = 13;
```
ioga, line 110: maybe rename experiment_id_filter

ioga (approved):

PR #4931, opened 2022-09-01
feat: add batch register and deletion of checkpoints from experiment [DET-8130]

EmilyBonar:
## Description
Adds ability to batch Register or Delete checkpoints from Experiment Checkpoints table
batch checkpoints

## Test Plan
Go to experiment checkpoints table, attempt to batch register and batch delete checkpoints.

## Checklist

### DIFF
--- a/webui/react/src/components/InteractiveTable.tsx
+++ b/webui/react/src/components/InteractiveTable.tsx
@@ -58,9 +58,9 @@ const SOURCE_TYPE = 'DraggableColumn';
 type DndItem = {
   index?: number;
 }
-interface ContextMenuProps {
+export interface ContextMenuProps<RecordType> {
   onVisibleChange: (visible: boolean) => void;
-  record: Record<string, unknown>;
+  record: RecordType;
 }
 
 export interface ColumnDef<RecordType> extends ColumnType<RecordType> {
@@ -74,7 +74,7 @@ export type ColumnDefs<ColumnName extends string, RecordType> = Record<
 >;
 
 interface InteractiveTableProps<RecordType> extends TableProps<RecordType> {
-  ContextMenu?: React.FC<ContextMenuProps>;
+  ContextMenu?: React.FC<ContextMenuProps<RecordType>>;
   areRowsSelected?: boolean;
   columns: ColumnDef<RecordType>[],
   containerRef: MutableRefObject<HTMLElement | null>,
@@ -84,14 +84,15 @@ interface InteractiveTableProps<RecordType> extends TableProps<RecordType> {
 
 /* eslint-disable-next-line @typescript-eslint/ban-types */
 type InteractiveTable = <T extends object>(props: InteractiveTableProps<T>) => JSX.Element;
+type Row = <T extends object>(props: RowProps<T>) => JSX.Element;
 
 type DragState = 'draggingRight' | 'draggingLeft' | 'notDragging';
-interface RowProps {
-  ContextMenu?: React.FC<ContextMenuProps>;
+interface RowProps<RecordType> {
+  ContextMenu?: React.FC<ContextMenuProps<RecordType>>;
   areRowsSelected?: boolean;
   children?: React.ReactNode;
   className?: string;
-  record: Record<string, unknown>;
+  record: RecordType;
 }
 
 interface HeaderCellProps {
@@ -123,14 +124,14 @@ const getAdjustedColumnWidthSum = (columnsWidths: number[]) => {
 
 const RightClickableRowContext = createContext({});
 
-const Row = ({
+const Row: Row = ({
   className,
   children,
   record,
   ContextMenu,
   areRowsSelected,
   ...props
-}: RowProps) => {
+}) => {
   const classes = [ className, css.row ];
 
   const [ rowHovered, setRowHovered ] = useState(false);
@@ -301,7 +302,6 @@ const InteractiveTable: InteractiveTable = ({
   areRowsSelected,
   ...props
 }) => {
-
   const columnDefs = useMemo(
     () =>
       columns
--- a/webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointDelete.tsx
+++ b/webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointDelete.tsx
@@ -1,0 +1,66 @@
+import { ExclamationCircleOutlined } from '@ant-design/icons';
+import { ModalFuncProps } from 'antd';
+import React, { useCallback, useEffect, useMemo } from 'react';
+
+import { detApi } from 'services/apiConfig';
+import { readStream } from 'services/utils';
+import useModal, { ModalCloseReason, ModalHooks } from 'shared/hooks/useModal/useModal';
+import {
+  CheckpointWorkloadExtended,
+  CoreApiGenericCheckpoint,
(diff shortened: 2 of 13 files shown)


ClaireNeveu:
Could you include a screenshot?

EmilyBonar:
Added screenshot

webui/react/src/components/InteractiveTable.tsx
```diff
@@ -58,9 +58,9 @@ const SOURCE_TYPE = 'DraggableColumn';
 type DndItem = {
   index?: number;
 }
-interface ContextMenuProps {
+export interface ContextMenuProps<RecordType> {
   onVisibleChange: (visible: boolean) => void;
-  record: Record<string, unknown>;
+  record: RecordType;
```
trentwatt, line 63: 👍

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointDelete.tsx
```diff
@@ -0,0 +1,80 @@
...
+      checkpointUuids: Array.isArray(checkpoints) ?
+        checkpoints :
+        [ checkpoints ],
+    }));
+    onClose?.(ModalCloseReason.Ok);
+  }, [ checkpoints, onClose ]);
+
+  const modalProps: ModalFuncProps = useMemo(() => {
+    const content =
+      `Are you sure you want to request checkpoint deletion for all batches in
```
trentwatt, line 45: I think a checkpoint always corresponds to a single batch.

The text in the existing useModalCheckpoint file should read "checkpoint deletion for batch ${totalBatches}"

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -50,23 +54,38 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
     reason?: ModalCloseReason,
-    checkpointUuid?: string,
+    checkpoints?: string[],
   ) => {
-    if (checkpointUuid) openModalCreateModel({ checkpointUuid });
+    if (checkpoints) openModalCreateModel({ checkpoints });
   }, [ openModalCreateModel ]);
 
-  const {
+  // Has to use var to hoist openModalCheckpointRegister for use above
```
trentwatt, line 77: Question: it looks like we are able to avoid doing this in CheckpointModalTrigger. Would it be possible to do same here?

Alternatively, should useModalCheckpointRegister wrap the useModalModelCreate?

webui/react/src/shared/components/ActionDropdown/ActionDropdown.tsx
```diff
@@ -18,17 +18,23 @@ export type Confirmations<T extends string> =
...
+  children?: React.ReactNode;
   /**
    * whether to prompt the user to confirm the action before executing it
    * with options to customize the generated modal.
    */
   confirmations?: Confirmations<T>
+  /**
+   * whether the action is marked as dangerous or not.
+   */
+  danger?: DangerousActions<T>;
```
trentwatt, line 37: Not necessarily in this PR but I was thinking it might be nice to reuse the Action interface in TableBatch for this as well. 

That way we can just add confirmations, dangerous, and the like to the Action interface themselves instead of having these extra metadata props. 

What do you think?

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -92,24 +111,56 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
+  useCallback((
+    { record, onVisibleChange, children },
+  ) => {
+    return (
+      <ActionDropdown<CheckpointAction>
+        actionOrder={[
+          'Register',
+          'Delete',
+        ]}
+        danger={{ [checkpointAction.Delete]: true }}
```
trentwatt, line 140: Think we want to add disabled here based on the state of the checkpoint

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -92,24 +111,56 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
+  }, [ dropDownOnTrigger ]);
 
   const columns = useMemo(() => {
     const actionRenderer = (_: string, record: CoreApiGenericCheckpoint): React.ReactNode => (
       <ActionDropdown<CheckpointAction>
         actionOrder={[
           'Register',
+          'Delete',
         ]}
+        danger={{ [checkpointAction.Delete]: true }}
```
trentwatt, line 159: Ditto re: disabled

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -92,24 +111,56 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
 
-  const dropDownOnTrigger = useCallback((checkpoint: CoreApiGenericCheckpoint) => {
-    return { [checkpointAction.Register]: () => handleRegisterCheckpoint(checkpoint) };
-  }, [ handleRegisterCheckpoint ]);
+  const handleDeleteCheckpoint = useCallback((checkpoints: string[]) => {
+    openModalCheckpointDelete({ checkpoints });
+  }, [ openModalCheckpointDelete ]);
+
+  const dropDownOnTrigger = useCallback((checkpoints: string | string[]) => {
+    const checkpointsArr = Array.isArray(checkpoints) ? checkpoints : [ checkpoints ];
```
trentwatt, line 123: It looks like if API failed for one checkpoint, the others aren't processed, so we would probably want to at least filter based on state here (state == completed for registered, state in [completed, errored] for delete?) to have best chance of successful API calls.

webui/react/src/hooks/useModal/Checkpoint/useModalCheckpointRegister.tsx
```diff
@@ -78,40 +79,68 @@ const useModalCheckpointRegister = ({ onClose }: Props = {}): ModalHooks => {
...
+                View Model
+              </Link>
+            </div>),
+          message: '',
+        });
+      }
     } catch (e) {
       handleError(e, {
-        publicSubject: 'Unable to register checkpoint.',
+        publicSubject: `Unable to register ${pluralizer(checkpoints.length, 'checkpoint')}.`,
```
trentwatt, line 143: Minor nit: I am not seeing this getting hit for errors (which is another ticket), but I think this would be misleading in case of partial failures

trentwatt (approved):
Overall LGTM, nice work!

Main thing would just be to disable actions based on checkpoint state (called out in relevant parts below), because failure of one prevents the other actions from proceeding.

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
EmilyBonar: That works for me. Are you sure that those are the only valid states for registering and deleting checkpoints?

PR #4932, opened 2022-09-01
fix: remove prevUser constraint

keita-determined:
## Description
[DET-8290]

 related: #4790

## Test Plan
- Check if user settings is applied even in secret mode

## Commentary (optional)
My prev change caused this. I'm so dumb 🐛. But still there is a problem which api is called many times. Hopefully react-query will be installed here

## Checklist

[DET-8290]: https://determinedai.atlassian.net/browse/DET-8290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gt2345 (commented):
Thank you for looking into this, the data seems to sync to local storage alright, but changes not actually applied to table.
Screen Shot 2022-09-01 at 12 25 36 PM

gt2345 (approved):

PR #4934, opened 2022-09-01
chore: ensure isAuthError can see into wrapped exceptions

hamidzr:
## Description

- add assertion error type
- ensure isAuthError can work with wrapped auth errors
- always logs error messages to logger

## Commentary (optional)

started out here https://github.com/determined-ai/saas/pull/333

## Checklist


webui/react/src/pages/SignOut.tsx
```diff
@@ -3,9 +3,9 @@ import { useHistory, useLocation } from 'react-router-dom';
 
 import { StoreAction, useStore, useStoreDispatch } from 'contexts/Store';
 import { paths, routeAll } from 'routes/utils';
-import { logout } from 'services/api';
+import { isAuthFailure, logout } from 'services/api';
```
johnkim-det, line 6: nit question: why do we import and export in `services/api` instead of just importing directly from `shared/utils/service`?

johnkim-det (approved):

webui/react/src/pages/SignOut.tsx
hamidzr: and we should directly import from the `shared` dir if available. at times we've reexported shared utils in their old module to avoid updating all the references at the same time. it would have looked like this https://github.com/determined-ai/determined/pull/4934/commits/624f178ec6e1389af6c7d2ea47054d6df0eb28fe

PR #4939, opened 2022-09-01
docs: update required python to 3.7

rb-determined-ai:
We stopped supporting Python 3.6 when it went EOL.

I reported it in the release notes in #3377, but forgot to fix the rest
of the docs.

Note that setup.py has not been modified; there's no need to break
existing users who are succesfully using an usupported old version.

### DIFF
--- a/docs/training/setup-guide/custom-env.rst
+++ b/docs/training/setup-guide/custom-env.rst
@@ -198,7 +198,7 @@ environments using :ref:`custom images <custom-docker-images>`:
    FROM determinedai/environments:py-3.8-pytorch-1.10-tf-2.8-cpu-0.19.1
 
    # Create a virtual environment
-   RUN conda create -n myenv python=3.6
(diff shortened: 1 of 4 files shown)


stoksc (approved):

docs/training/setup-guide/custom-env.rst
```diff
@@ -198,7 +198,7 @@ environments using :ref:`custom images <custom-docker-images>`:
    FROM determinedai/environments:py-3.8-pytorch-1.10-tf-2.8-cpu-0.19.1
 
    # Create a virtual environment
-   RUN conda create -n myenv python=3.6
+   RUN conda create -n myenv python=3.8
```
stoksc, line 201: 3.8?

rb-determined-ai: batches the FROM image

PR #4949, opened 2022-09-02
fix: associate allocation sessions with users

NicholasBlaskey:
## Description

#4905 assumed that using user session would be ok. I wasn't considering user sessions expiring to be an issue so switched back to using allocation sessions. This change associates allocation sessions with users so that they can use ```grpcutil.GetUser``` which is needed for RBAC to determined access for users.

## Test Plan

Run an experiment and see no errors.

## Checklist

### DIFF
--- a/harness/determined/common/api/request.py
+++ b/harness/determined/common/api/request.py
@@ -11,7 +11,6 @@ import urllib3
 
 import determined as det
 import determined.common.requests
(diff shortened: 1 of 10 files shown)


ioga (approved):

mpkouznetsov (approved):
looks good

PR #4950, opened 2022-09-02
feat: Add no permissions warning page[DET-8227]

julian-determined-ai:
## Description

feat: Add no permissions warning page [DET-8227]

If a user has no permissions as determined by RBAC then we want to show them a special warning page instead of the home page. The current home page and `defaultRoute` is the `uncategorized` page. 

Please see the `Additional Commentary`. section for an explanation as to why the font files have changed. 

WIP:
  - Correctly determine the current user's permissions.

## Commentary (optional)

While implementing this feature it came to my attention that the `warning` icon does not render correctly. 

1A8A7DEA-B9FB-4244-AD25-F44281773364_4_5005_c

The top portion of the svg icon as well as some portion of either side is cut off. This is not due to any html, css, or js issue this is a problem with the icon itself. Interestingly enough when viewing the svg icon itself it looks fine. I tested other icons and they seem to render correctly, it looks like only the `warning` icon has this issue. 

I have updated our icon set with the new warning icon and updated our font files.

## Checklist

[DET-8227]: https://determinedai.atlassian.net/browse/DET-8227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mapmeld:
For the font files, do we need to do anything to the frontend to make current browsers load the new version (i.e. adding `?v=2`) or is it loaded by default / reloaded when the UI has this new glyph/icon?

julian-determined-ai:
> For the font files, do we need to do anything to the frontend to make current browsers load the new version (i.e. adding `?v=2`) or is it loaded by default / reloaded when the UI has this new glyph/icon?

Great callout, this should be handled in our current bundling configuration. Our font files do include unique identifiers.
Screen Shot 2022-09-07 at 11 26 00 AM

mapmeld (approved):
LGTM. Tested by removing oss_user permission in src/contexts/Store.ts

PR #4952, opened 2022-09-03
chore: reduce and log effective store state changes

hamidzr:
## Description

- bring more visibility into Store
- check and prevent detectable unnecessary ref changes

## Commentary (optional)
DeepinScreenshot_select-area_20220902171300

suggestion: review commit by commit

we seem to set a few things 2 to 3 times in a row. (app level rerender?)

## Checklist

### DIFF
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -380,7 +380,11 @@ export const useStoreDispatch = (): Dispatch<Action> => {
 };
 
 const StoreProvider: React.FC<Props> = ({ children }: Props) => {
-  const [ state, dispatch ] = useReducer(reducer, initState);
+  const [ state, dispatch ] = useReducer((state: State, action: Action) => {
+    const newState = reducer(state, action);
+    if (isEqual(state, newState)) return state; // CHECK: performance concerns?
+    return newState;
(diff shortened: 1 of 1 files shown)


hamidzr:
no noticeable impact from testing against latest master. 
I tried to exaggerate its usage:
```
diff --git a/webui/react/src/contexts/Store.tsx b/webui/react/src/contexts/Store.tsx
index 15d5ffcc1..9780726ca 100644
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -381,6 +381,9 @@ export const useStoreDispatch = (): Dispatch<Action> => {
 const StoreProvider: React.FC<Props> = ({ children }: Props) => {
   const [ state, dispatch ] = useReducer((state: State, action: Action) => {
     const newState = reducer(state, action);
+    for (let i = 0; i < 100; i++) {
+      isEqual(state, newState);
+    }
     if (isEqual(state, newState)) return state; // CHECK: performance concerns?
     logger.debug('store state updated', action.type);
     return newState;

```
DeepinScreenshot_select-area_20220908130819

assuming this doesn't cause any missed updates 🤔 the pure time spent in isEqual should be way less that rerender it could stop

PR #4954, opened 2022-09-04
feat: use task names for interactive task page titles.

ioga:
## Description

Interactive task pages (e.g. Jupyterlab pages) were using the default page title `Tasks - Determined` (in HTML sense). When you have multiple tabs running simultaneously, it becomes challenging to distinguish these pages between each other and the "normal" tasks page at `/det/tasks`. With this PR, well show the task name instead, e.g. `Jupyterlab (brightly-capable-corgi)`

## Test Plan

Run a couple jupyterlabs and see they have unique page titles names.

## Checklist


julian-determined-ai (approved):
This is a very helpful change! LGTM

PR #4963, opened 2022-09-06
feat: allow specifying Fluent Bit container UID/GID on Kubernetes [DET-8012]

dzhu:
## Description
As the title says.

## Test Plan
- [x] manually test with local Kubernetes, confirm that container runs as a specified user

### DIFF
--- a/helm/charts/determined/templates/master-config.yaml
+++ b/helm/charts/determined/templates/master-config.yaml
@@ -116,8 +116,8 @@ data:
       {{- end }}
       {{- if .Values.fluent }}
       fluent:
-        {{- if .Values.fluent.image }}
-        image: {{ .Values.fluent.image }}
+        {{- if .Values.fluent }}
+        {{- toYaml .Values.fluent | nindent 8}}
         {{- end }}
       {{- end }}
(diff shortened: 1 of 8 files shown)


helm/charts/determined/templates/master-config.yaml
```diff
@@ -118,6 +118,8 @@ data:
       fluent:
         {{- if .Values.fluent.image }}
         image: {{ .Values.fluent.image }}
+        uid: {{ .Values.fluent.uid }}
```
ioga, line 121: this is under `if .Values.fluent.image` which doesn't seem right because it won't allow setting uid/gid without image. 
since contents of Values.fluent are 1:1 with the master config, can we embed it here directly, without going over the fields individually? i.e.
```
{{- if .Values.fluent }}
  {{- toYaml .Values.fluent | nindent 4}}
{{ end }}
```

(I may be misjudging the nindent value)

dzhu: Whoops, good catch and good idea. Changed.

ioga (approved):

PR #4965, opened 2022-09-06
feat: Frontend uses permission store to clear actions [DET-8215]

mapmeld:
## Description

~`utils/role.ts`~ `hooks/usePermissions` will check whether a user is authorized to do actions (as a frontend feature, this is for the UI to reflect which actions will be authorized by the backend).

`{ userAssignments, userRoles } = useStore()` retrieves the last-updated assignments and roles. For now these keep the initialized state; a future PR will request roles from the server and periodically check for updates.

For experiments, projects, and workspaces, `relevantPermissions()` receives the workspace ID and returns a Set<string> of cluster- and workspace-level permissions. The mock role has an `oss_user` permission which follows our original rules (e.g. to delete an experiment, you should be the experiment owner, or an admin).  RBAC roles will look for this specific permission (`delete_experiment`).

```typescript
export const canDeleteExperiment = (
  experiment: ProjectExperiment,
  user: DetailedUser,
  userAssignments?: UserAssignment[],
  userRoles?: UserRole[],
): boolean => {
  const permitted = relevantPermissions(userAssignments, userRoles, experiment.workspaceId);
  return !!experiment && !!user &&
    permitted.has('oss_user') ? (user.isAdmin || user.id === experiment.userId)
    : permitted.has('delete_experiment');
};
```

Notes / Commentary:
- This PR covers any cases where OSS has restrictions based on user.isAdmin; more will be added in the future (i.e. creating a project, modifying a checkpoint)
- Models and Model Versions have cluster-level permissions
- Future OSS work will add Groups to this logic
 
## Test Plan

- Confirm you have ability to move or delete your own experiment
- Can rename your workspace
- Can delete projects in your own workspace
- Can delete your own model or model version

Use the CLI to create admin-owned objects which cannot be modified by the non-admin user:

```
det user login admin
det w create apple
det p create apple orange
[get project id number]
det e create path/to.yaml path/ --project_id ##
det m create MyModel
```

## Checklist


webui/react/src/utils/role.ts
```diff
@@ -0,0 +1,121 @@
...
+  userRoles.filter((r) => relevantAssigned.includes(r.name)).forEach((r) => {
+    // TODO: is it possible a role is assigned to this workspace,
+    // but not all of its permissions?
+    permissions = permissions.concat(r.permissions.filter((p) => p.globalOnly || workspaceId));
+  });
+  return new Set<string>(permissions.map((p) => p.name));
+};
+
+// Experiment actions
+export const canDeleteExperiment = (
```
ClaireNeveu, line 34: Could you make some kind of `usePermissions` hook so that users don't have to manually plumb data from the store (while also hiding the details of how you're storing permissions)?

ClaireNeveu: I imagine this would look like:
```typescript
const permissions = usePermissions()
permissions.canDeleteExperiment(experiment)
```

mapmeld: I just got this working in a few parts, let me know if it makes sense to use this style in the other permissions. For example WorkspaceDetailsHeader

https://github.com/determined-ai/determined/pull/4965/files#diff-42d7ff039378e1e54f016def9813d3815bd8631e4caa3ef152d295c7ab34339a

and it goes through hooks/usePermissions

https://github.com/determined-ai/determined/pull/4965/files#diff-50e6e78f6aa3d324124d4e1d75ce89404b99f8e6e2aff107f58348d05d52a7a5R43

ClaireNeveu: Personally I think the experiment.ts stuff used in WorkspaceDetailsHeader should be hooks too since it's always using the same global data. That better encapsulates the intended functionality and I don't think it's a big rewrite. That said I'll leave this up to you.

webui/react/src/hooks/usePermissions.ts
```diff
@@ -0,0 +1,59 @@
...
+} from 'utils/role';
+
+interface PermissionsConfig {
+  project?: Project;
+  workspace?: PermissionWorkspace;
+}
+
+type PermissionFn = (arg0: PermissionsConfig) => boolean;
+
+interface PermissionsHook {
```
ClaireNeveu, line 18: If you wanted to avoid the duplication and don't mind the class baggage you could use a class here.

webui/react/src/utils/role.ts
mapmeld: I added `usePermissions` for experiments, models, etc.
I removed several lines which pass the current user object from component to component, because we're now going to get the current user and check permission inside of `usePermissions`.

webui/react/src/hooks/usePermissions.ts
mapmeld: I ended up making a different PermissionsArgs type for the experiments, models, model-versions, projects, workspaces etc.

ClaireNeveu: I just meant having to type all the function names twice.

webui/react/src/hooks/useModal/Model/useModalModelDelete.tsx
```diff
@@ -55,7 +55,7 @@ const useModalModelDelete = ({ onClose }: Props = {}): ModalHooks => {
       onOk: handleOk,
       title: 'Confirm Delete',
     } : clone(CANNOT_DELETE_MODAL_PROPS);
-  }, [ user?.id, user?.isAdmin ]);
+  }, [ canDeleteModel ]);
```
ClaireNeveu, line 58: I think it's okay to add a `permissions.rawPermissions`that returns an array of permissions that can be used in places like this.

webui/react/src/hooks/useModal/Model/useModalModelVersionDelete.tsx
```diff
@@ -22,12 +22,13 @@ interface ModalHooks extends Omit<Hooks, 'modalOpen'> {
 }
 
 const useModalModelVersionDelete = ({ onClose }: Props = {}): ModalHooks => {
-  const { auth: { user } } = useStore();
   const [ modelVersion, setModelVersion ] = useState<ModelVersion>();
 
+  const { canDeleteModelVersion } = usePermissions();
+
   const isDeletable = useMemo(() => {
```
ClaireNeveu, line 29: Do we actually need `useMemo` here?  This is a very cheap function.

webui/react/src/pages/WorkspaceList/WorkspaceActionDropdown.tsx
```diff
@@ -44,9 +43,15 @@ const WorkspaceActionDropdown: React.FC<Props> = ({
     modalOpen: openWorkspaceEdit,
   } = useModalWorkspaceEdit({ onClose: onComplete, workspace });
 
-  const userHasPermissions = useMemo(() => {
-    return curUser?.isAdmin || curUser?.id === workspace.userId;
-  }, [ curUser?.id, curUser?.isAdmin, workspace.userId ]);
+  const { canDeleteWorkspace, canModifyWorkspace } = usePermissions();
+
+  const canDelete = useMemo(() => {
```
ClaireNeveu, line 48: As above I don't know that we really need these `useMemo`s since these are trivial computations.

webui/react/src/utils/role.ts
ClaireNeveu: If you can you should just make these functions private in the `usePermissions` hook file to reduce the surface area of the permissions "library".

webui/react/src/hooks/useModal/Model/useModalModelDelete.tsx
mapmeld: There's some difficulty because in RBAC we just need the `delete_model` permission, but in OSS the permissions here depends on context of who owns the model (+ future Groups logic)

mapmeld:
> @ClaireNeveu If you can you should just make these functions private in the usePermissions hook file to reduce the surface area of the permissions "library".

✅  Moved them into `usePermissions`

webui/react/src/hooks/useModal/Model/useModalModelDelete.tsx
ClaireNeveu: For OSS we can just create a client-side only oss_admin permission object that we return in `rawPermissions`that would preserve the old behavior here. Though do we ever have to actually invalid this callback? When are users becoming admins during interaction?

ClaireNeveu (approved):

PR #4967, opened 2022-09-07
fix: set auth cookie client side for external flows

hamidzr:
## Description
DET-8310

## Test Plan

I was able to run a patched version of master through Samuel and verified that jupyterlab auth is in place. details in the ticket

## Commentary (optional)

~I think this should work as a proof of concept but I want to do it a bit differently with more time. 
the backend sets auth cookie when we login through other methods should it also handle this case or stop setting cookies altogether and leave it to the client~

## Checklist

### DIFF
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -8,6 +8,7 @@ import rootLogger from 'shared/utils/Logger';
 import { percent } from 'shared/utils/number';
 import { Agent, Auth, ClusterOverview, ClusterOverviewResource, DetailedUser, DeterminedInfo,
   PoolOverview, ResourcePool, ResourceType, UserAssignment, UserRole, Workspace } from 'types';
+import { getCookie, setCookie } from 'utils/browser';
 
 const logger = rootLogger.extend('store');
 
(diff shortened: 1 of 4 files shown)


mackrorysd:
You make a good point that with the external tokens there's no sign-in call, we just try using the token if we have one, and if that fails we redirect. That's a good enough reason to me to set the cookie client-side. We could check if the cookie is set and set it whenever we get the token, but that is checking the token on A LOT of calls...

hamidzr:
I was writing this up to ask the team but! I realize in the saas case the client also needs to handle clearing the cookie on log out which I think pushes me to keep the cookie handling in the web side 🤔 

Hello team,

I wanted to get your thoughts on a change related to cookies and jupyter lab authentication through saas.

Whenever we go through a login the backend sets the auth cookie on the domain but when we're logging in through
saas this does not get set and jupyterlab authentication fails.

Problem: we need to set the cookie somewhere or stop relying on the cookie.

Some potential solutions:
- one could be special casing this in the UI so that it sets the cookie when the auth token is coming from saas
- having the backend set this up at some point so the cookie origin is the same for 


context:
- AFAIK it's not possible set the auth header on websocket connections through js on mainstream browsers
- currently, that's how core authenticated when opening jupyterlab through webui

https://determinedai.atlassian.net/browse/DET-8310

hamidzr:
I need to figure out how to test these through a saas deployment as close to production as possible

hamidzr:
react test failure https://determinedai.atlassian.net/browse/DET-8351

julian-determined-ai (approved):
The implementation looks good. 

Do we need to plan to revisit how we want to handle this cookie in the future? Or have we decided that the need/want to set this cookie value on the client side?

PR #4970, opened 2022-09-07
fix: reduce settings api calls [DET-8307]

EmilyBonar:
## Description
This should reduce the number of user settings API calls.

## Test Plan
Open up the network inspector in Firefox and enter `setting` into the filter box. On an Experiment List page with 10 experiments, number of initial calls should be reduced from ~30 to ~10.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useFetch.ts
+++ b/webui/react/src/hooks/useFetch.ts
@@ -9,6 +9,7 @@ import {
   getInfo,
   getResourcePools,
   getUsers,
+  getUserSetting,
   getWorkspaces,
 } from 'services/api';
 import { ErrorType } from 'shared/utils/error';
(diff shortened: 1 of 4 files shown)


webui/react/src/hooks/useFetch.ts
```diff
@@ -81,6 +82,17 @@ export const useFetchUsers = (canceler: AbortController): () => Promise<void> =>
   }, [ canceler, storeDispatch ]);
 };
 
+export const useFetchUserSettings = (canceler: AbortController): () => Promise<void> => {
+  const storeDispatch = useStoreDispatch();
+
+  return useCallback(async (): Promise<void> => {
+    try {
+      const userSettingResponse = await getUserSetting({}, { signal: canceler.signal });
```
ClaireNeveu, line 90: This will definitely be an improvement over what currently exists but I wanted to call out a few issues we should address in the future:
1. Right now this initial fetch is always made. We should avoid a fetch if the store is already up-to-date.
2. Maybe this is buried in the auto-generated gRPC code but we aren't de-duplicating requests. if another call to `useFetchUsersSettings(...)()` is made before the first one is complete, even if point 1 is addressed we still make another call.

ClaireNeveu (approved):

PR #4979, opened 2022-09-08
fix: correct overflow action buttons [DET-8322]

EmilyBonar:
## Description
`.optionsMainButton:nth-of-type(1)` doesn't select the first element with the class `.optionsMainButton`. Apparently it actually selects the element with the class `.optionsMainButton` that is the first of its type (button). Since the first sibling button didn't have the class `.optionsMainButton`, it didn't select anything. It's very confusing, here are some more details: https://stackoverflow.com/questions/30603786/nth-of-type2-is-targeting-the-first-of-type

## Test Plan
Go to an experiment details page and change the width of the window. The actions buttons should swap between the button row and the overflow menu smoothly.
image

## Checklist

### DIFF
--- a/webui/react/src/components/PageHeaderFoldable.module.scss
+++ b/webui/react/src/components/PageHeaderFoldable.module.scss
@@ -37,8 +37,13 @@
       justify-content: flex-end;
       max-width: fit-content;
 
-      .optionsMainButton {
-        display: none;
+      .optionsButtons {
+        display: flex;
+        gap: 8px;
+
+        .optionsMainButton {
+          display: none;
+        }
       }
     }
     .optionsDropdown {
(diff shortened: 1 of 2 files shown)


ClaireNeveu:
I think the appropriate fix here is to wrap the `optionsMainButton`s into one div so that they're guaranteed to be the the only items considered for `nth-of-type`. Right now you're just excluding an item that happens to be ahead of the group.

EmilyBonar:
> I think the appropriate fix here is to wrap the `optionsMainButton`s into one div so that they're guaranteed to be the the only items considered for `nth-of-type`. Right now you're just excluding an item that happens to be ahead of the group. That makes this fragile in the face of refactoring.
> 
> If you _really_ want to get proper, we should give that div a class and then use `first-child` so that it's clear that the div is significant.

That's a good point. I did it this way in order to make the minimal possible change and not introduce the possibility of spacing problems, but what you said is more semantically correct. I'll switch it around.

ClaireNeveu (approved):
Make sure you tested again after rearranging the CSS

PR #4984, opened 2022-09-08
feat: can see users' permissions if view_permissions enabled [DET-8222]

mapmeld:
## Description

When viewing a user profile in RBAC, their permissions are listed in an `antd.Table` component. 

For testing purposes, this is available to OSS admin users, and the mock response is a global `oss_user` permission.

Screen Shot 2022-09-08 at 2 38 42 PM

## Test Plan

[to test in OSS]
- While logged in as admin, go to `/det/settings/user-management/rbac-enabled`. Click "Edit User" (or view, if it's available). You should see the table and oss_user permission.
- Log in as determined / other user, and go to the same page. There is no table.

## Checklist

### DIFF
--- a/webui/react/src/services/apiConfig.ts
+++ b/webui/react/src/services/apiConfig.ts
@@ -198,6 +198,24 @@ export const resetUserSetting: DetApi<
   request: () => detApi.Users.resetUserSetting(),
 };
 
+export const getUserPermissions: DetApi<
+  Service.GetUserParams, number, Type.Permission[]
+> = {
+  name: 'getUserPermissions',
+  postProcess: (response) => {
+    const fillerPermission: Type.Permission = {
+      globalOnly: true,
+      id: response,
+      name: 'oss_user',
+      workspaceOnly: false,
+    };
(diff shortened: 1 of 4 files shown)


webui/react/src/services/apiConfig.ts
```diff
@@ -198,6 +198,24 @@ export const resetUserSetting: DetApi<
...
+    const fillerPermission: Type.Permission = {
+      globalOnly: true,
+      id: response,
+      name: 'oss_user',
+      workspaceOnly: false,
+    };
+    return [ fillerPermission ];
+  },
+  request: (params) => new Promise((resolve) => {
+    resolve(-1 * params.userId);
```
gt2345, line 215: Is there a reason why we use -1 here?

gt2345 (approved):
LGTM

webui/react/src/services/apiConfig.ts
mapmeld: oh, originally I was going to return -1 but then I needed to use `params`. I suppose it was not needed.

PR #4986, opened 2022-09-08
fix: don't attempt to remove zero checkpoints

phillip-gaisford:
## Description

When deleting an experiment, skip the checkout point deletion if the count is zero. An experiment can end up with zero checkpoints if the slurm launch is not successful, amongst other reasons.

## Test Plan

Tested manually.

## Checklist


jerryharrow (approved):
Are there any unit tests we can augment?

master/internal/api_experiment.go
```diff
@@ -230,16 +230,17 @@ func (a *apiServer) deleteExperiment(exp *model.Experiment, user *model.User) er
...
-		a.m.rm, a.m.db, a.m.taskLogger, taskID, exp.JobID, jobSubmissionTime, taskSpec, exp.ID,
-		conf, checkpoints, true, agentUserGroup, user, nil,
-	)
-	if gcErr := a.m.system.MustActorOf(addr, ckptGCTask).AwaitTermination(); gcErr != nil {
-		return errors.Wrapf(gcErr, "failed to gc checkpoints for experiment")
+	if len(checkpoints) > 0 {
+		addr := actor.Addr(fmt.Sprintf("delete-checkpoint-gc-%s", uuid.New().String()))
+		jobSubmissionTime := exp.StartTime
+		taskID := model.NewTaskID()
+		ckptGCTask := newCheckpointGCTask(
```
stoksc, line 237: it's fine to me if it is duplicated, but we should push this into `newCheckpointGCTask` (i guess the actor's prestart) otherwise this is just gonna happen again.

phillip-gaisford: Something like this (in checkpoint_gc.go):

```
	case actor.PreStart:
		if t.ToDelete == "" {
			t.completeTask(ctx)
			return nil
		}
```

Or is there more to it?

stoksc: ahh... deleting tensorboards is very good catch. we can never know upfront if there is tb data to gc without checking in the gc task.. i think we may need to modify the gc task itself/how it is launched to be ok with `--delete []` or something.

master/internal/checkpoint_gc.go
```diff
@@ -73,6 +73,11 @@ func newCheckpointGCTask(
 func (t *checkpointGCTask) Receive(ctx *actor.Context) error {
 	switch msg := ctx.Message().(type) {
 	case actor.PreStart:
+		if t.ToDelete == "" && !t.DeleteTensorboards {
+			// Early return as nothing to do
+			t.completeTask(ctx)
```
stoksc, line 78: no need for completeTask since `AddTask` hasn't even run yet. we should just call `ctx.Self().Stop()` and return nil.

master/pkg/tasks/task_gc.go
```diff
@@ -86,6 +86,10 @@ func (g GCCkptSpec) ToTaskSpec() TaskSpec {
 		"--delete",
 		g.ToDelete,
 	}
+	if g.ToDelete != "" {
+		res.Entrypoint = append(res.Entrypoint, "--delete")
```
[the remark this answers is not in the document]
phillip-gaisford: Yes; when tested I get a successful experiment delete

stoksc (approved):
lgtm.

stoksc:
@cla-bot check

master/pkg/tasks/task_gc.go
```diff
@@ -83,8 +83,10 @@ func (g GCCkptSpec) ToTaskSpec() TaskSpec {
 		strconv.Itoa(g.ExperimentID),
 		"--storage-config",
 		"/run/determined/checkpoint_gc/storage_config.json",
-		"--delete",
-		g.ToDelete,
+	}
+	if g.ToDelete != "" {
+		res.Entrypoint = append(res.Entrypoint, "--delete")
+		res.Entrypoint = append(res.Entrypoint, g.ToDelete)
```
stoksc, line 89: go style nit, just `res.Entrypoint = append(res.Entrypoint, "--delete", g.ToDelete`. but too late now.

phillip-gaisford: Good to know -- and thanks for the reviews!

PR #4987, opened 2022-09-08
chore: test PyTorch AMP with gradient aggregation [DET-6105]

drh-determined-ai:
## Description

chore: test PyTorch AMP with gradient aggregation [DET-6105]

## Test Plan

Run unit tests with a GPU and with `apex` installed.

## Commentary (optional)

Because these tests rely on a GPU, they are **not currently being run by CircleCI**.

## Checklist

### DIFF
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -456,10 +456,11 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             If  ``optimizers`` args were lists, the corresponding return value will
             also be a list.
         """
-        if not self.env.managed_training:
+        if not enabled or not self.env.managed_training:
             return models, optimizers
 
-        check.is_none(self._scaler, "Do not mix APEX with PyTorch AMP")
+        if self._scaler is not None:
+            check.false(self._scaler.is_enabled(), "Do not mix APEX with PyTorch AMP")
 
         check.false(self._use_apex, "Please only call configure_apex_amp once.")
         if self.distributed.size > 1:
@@ -477,6 +478,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
                 self._aggregation_frequency,
                 1,
                 "Mixed precision training (AMP) is not supported with "
+                "distributed training and "
                 "aggregation frequency > 1.",
             )
 
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -374,7 +374,9 @@ class PyTorchTrialController(det.TrialController):
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
             return False
-        return self.context._should_communicate_and_update()  # type: ignore
+        if self.context.distributed.size > 1 or self.context._aggregation_frequency > 1:
+            return self.context._should_communicate_and_update()  # type: ignore
+        return True
 
     def _train_for_step(
         self, step_id: int, num_batches: int, total_batches_processed: int
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -374,9 +374,7 @@ class PyTorchTrialController(det.TrialController):
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
             return False
-        if self.context.distributed.size > 1 or self.context._aggregation_frequency > 1:
-            return self.context._should_communicate_and_update()  # type: ignore
-        return True
+        return self.context._should_communicate_and_update()  # type: ignore
 
     def _train_for_step(
         self, step_id: int, num_batches: int, total_batches_processed: int
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -374,9 +374,7 @@ class PyTorchTrialController(det.TrialController):
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
             return False
-        if self.context.distributed.size > 1:
-            return self.context._should_communicate_and_update()  # type: ignore
-        return True
+        return self.context._should_communicate_and_update()  # type: ignore
 
     def _train_for_step(
         self, step_id: int, num_batches: int, total_batches_processed: int
--- a/harness/tests/experiment/fixtures/pytorch_onevar_model.py
+++ b/harness/tests/experiment/fixtures/pytorch_onevar_model.py
@@ -289,7 +289,7 @@ class OneVarApexAMPTrial(OneVarAMPBaseTrial):
         metrics = super().train_batch(batch, epoch_idx, batch_idx)
         metrics["scale_before"] = scale_before
         metrics["scale"] = apex.amp.state_dict()["loss_scaler0"]["loss_scale"]
-        metrics["stage"] = self._stages[batch_idx]
+        metrics["stage"] = self._stages[batch_idx // self._agg_freq]
         return metrics
 
 
@@ -319,7 +319,7 @@ class OneVarAutoAMPTrial(OneVarAMPBaseTrial):
         metrics = super().train_batch(batch, epoch_idx, batch_idx)
         metrics["scale_before"] = scale_before
         # self.scaler.update() gets called after this method returns
-        metrics["stage"] = self._stages[batch_idx]
+        metrics["stage"] = self._stages[batch_idx // self._agg_freq]
         return metrics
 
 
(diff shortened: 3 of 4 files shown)


harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -374,9 +374,7 @@ def _auto_step_lr_scheduler_per_batch(
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
             return False
-        if self.context.distributed.size > 1:
-            return self.context._should_communicate_and_update()  # type: ignore
-        return True
```
rb-determined-ai, line 379: So this looks like a breaking change, right?

Existing behavior: non-distributed training would always update the scaler, even if aggregation_frequency was > 1

Proposed behavior: scaler is updated only when the optimizer is stepped, which is based on aggregation_frequency

Is that correct?  It looks like the current behavior is just wrong, but we need to treat it a little specially if it's a breaking change.

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -734,16 +836,24 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
 
-def amp_metrics_test(trial_class, training_metrics):
+def amp_metrics_test(trial_class, training_metrics, agg_freq=1):
     loss_prev = None
     GROWTH_INTERVAL = trial_class._growth_interval
     MIN_SCALED_LOSS_TO_REDUCE_SCALE = 32760
     growth_countdown = GROWTH_INTERVAL
     # Only attempt assertions up to and including the penultimate batch, because
     #  we may not have the updated scale from the final batch.
-    for idx, (metrics, next_metrics) in enumerate(zip(training_metrics[:-1], training_metrics[1:])):
```
rb-determined-ai, line 744: did this line get changed or just reformatted?  If just reformatted, why bother?

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -734,16 +836,24 @@ def test_checkpoint_loading(ckpt: str, istrial: bool):
...
+    for idx, (metrics, next_metrics) in enumerate(
+        zip(
+            training_metrics[:-1],
+            training_metrics[1:],
+        )
+    ):
+        if (idx + 1) % agg_freq != 0:
+            # Don't test mini-batches
```
rb-determined-ai, line 853: I do think the terms "batch" and "mini-batch" are ambiguous in our field, but I don't think using "mini" batch in this way is more helpful, since the term mini-batch has often been used to describe what we think of as just a "batch".  See [here](https://machinelearningmastery.com/gentle-introduction-mini-batch-gradient-descent-configure-batch-size/) for such an example.

Better to be precise and literal here:

```python
# Only test batches which land on aggregation_frequency boundaries.
```

harness/tests/experiment/pytorch/test_pytorch_trial.py
```diff
@@ -715,6 +759,64 @@ def make_amp_workloads() -> workload.Stream:
 
         amp_metrics_test(trial_class, training_metrics)
 
+    @pytest.mark.skipif(not torch.cuda.is_available(), reason="no gpu available")
+    @pytest.mark.gpu
+    @pytest.mark.parallel
```
rb-determined-ai, line 764: some of these marks are not generic pytest things but things we've configured.  I think `@pytest.mark.parallel` is meaningful in the e2e tests but doesn't really mean anything in the harness tests.

Do you intend on this test working with multiple gpus?

harness/tests/experiment/fixtures/pytorch_onevar_model.py
```diff
@@ -237,14 +237,15 @@ class AMPTestDataset(OnesDataset):
...
+        self._agg_freq = aggregation_freq
 
     def __len__(self) -> int:
-        return len(self.stages)
+        return len(self.stages) * self._agg_freq
 
     def __getitem__(self, index: int) -> Tuple:
-        x = self.STAGE_DATUM[self.stages[index]]
+        x = self.STAGE_DATUM[self.stages[index // self._agg_freq]]
         return torch.Tensor([float(x)]), torch.Tensor([float(x)])
```
rb-determined-ai, line 249: What is the scope of this particular PR.  Do you aim to test distributed training too?  Or is that future work?

If you are aiming to test distributed training, I would challenge you to set up the dataset in a way that only one worker is hitting overflow events.

The failure case that concerns us is basically, if we mess up the gradient aggregation or the ordering between communicating gradients and stepping the optimizer, some scalers on some workers might overflow while other scalers on other workers might step normally.

So to write a test that ensures we don't do that, you will need a dataset where different workers see different data.

When all workers see identical data, the risk of a false positive test result is much higher.  But it definitely is a much harder test to write.

If you are ok leaving the distributed training work for another pr, I would be ok with that, and what you have here seems just fine to me.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: Ah, now that I have read the thread from the user, I understand this is the key point of the whole PR.

Great!

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Oops. this reformatting is leftover from some changes I undid before creating the PR. This line happens to be exactly 100 characters long.

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: Thanks for the clarification! This is done.

harness/tests/experiment/fixtures/pytorch_onevar_model.py
drh-determined-ai: I can envision such a distributed test, but it will take some work and time for me to get close enough to grokking how Determined does distributed training.
So let's keep a smaller scope for this particular PR, and let's work on creating a new task for a distributed version of this test (I think a call is in order).

harness/tests/experiment/pytorch/test_pytorch_trial.py
drh-determined-ai: I have removed the `parallel` mark. I think I had wrongly thought that Determined took non-trivial aggregation as implicative of distributed training. I know this isn't the case!
Of course, since we're doing AMP, we need _a_ GPU.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -374,9 +374,7 @@ def _auto_step_lr_scheduler_per_batch(
...
+        return self.context._should_communicate_and_update()  # type: ignore
```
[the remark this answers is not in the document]
drh-determined-ai: I think this alternative is better for being a more conservative change; also, the reason for it is easier to understand just by looking at the diff. 
If there's any possibility of `_should_communicate_and_update` causing an error (in the sense of misuse) if neither `distributed.size > 1` nor `_aggregation_frequency > 1` are `True`, this will prevent that.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -374,7 +374,7 @@ def _auto_step_lr_scheduler_per_batch(
     def _should_update_scaler(self) -> bool:
         if not self.context._scaler or not self.context.experimental._auto_amp:
             return False
-        if self.context.distributed.size > 1:
+        if self.context.distributed.size > 1 or self.context._aggregation_frequency > 1:
             return self.context._should_communicate_and_update()  # type: ignore
         return True
```
rb-determined-ai, line 379: Why not just return `self.context._should_communicate_and_update()` like you had it before?  It was simpler and it should be equivalent.

Semantically, this is exactly what `_should_communicate_and_update()` means.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -456,10 +456,11 @@ def configure_apex_amp(
             If  ``optimizers`` args were lists, the corresponding return value will
             also be a list.
         """
-        if not self.env.managed_training:
+        if not enabled or not self.env.managed_training:
             return models, optimizers
 
-        check.is_none(self._scaler, "Do not mix APEX with PyTorch AMP")
+        if self._scaler is not None:
+            check.false(self._scaler.is_enabled(), "Do not mix APEX with PyTorch AMP")
```
rb-determined-ai, line 463: The `check` module is legacy code.  Any time you touch it, please just replace it with an if statement.

Basically it is going to give low-quality error messages ("expected false but got true" or something), and it's going to raise a meaningless error type.

So please make this an if statement raising an exception of some sort.  Maybe InvalidExperimentException or whatever we call it.

rb-determined-ai (approved):
nice work

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: I'm glad to know that. I didn't like `check`. :-p

PR #4989, opened 2022-09-08
feat: checkpoint download through master

hanyucui:
## Description

Users do not always have access to checkpoint storage, and hence downloading checkpoints with the CLI does not always work. This feature overcomes the limitation by supporting checkpoint downloads through the master since t is often easier to grant access to the master than to individual users.  This feature only supports AWS S3 for now and relies on implicit IAM permissions on the master.  In the future, it can be extended to support other checkpoint storage types and take explicitly provided credentials.

A single endpoints is added:
- `/checkpoints/:checkpoint_uuid/`

The implementation downloads the checkpoint by making API calls to the S3 endpoint, packages it as a tgz or a zip file, and then sends the file to the client in the HTTP response. The whole process operates as streaming. No temporary files are stored or cached.

*Note:* One of the integration tests needs to run in the upstream repo to have access to cloud resources. Hence this PR is also pushed to https://github.com/determined-ai/determined/tree/hanyu/master-checkpoint-download

## Test Plan

- Use `devcluster` to start a dev cluster. Configure the `checkpoint_storage` section to use an S3 bucket and provide appropriate AWS credentials. For example:
```yaml
checkpoint_storage:
  type: s3
  bucket: <your test S3 bucket>
  prefix: /devcluster/checkpoint-download-test
  access_key: *** (redacted) ***
  secret_key: *** (redacted) ***
```
- Make sure the same or equivalent credentials are available in environment variables or `~/.aws/credentials`.
- Run `export DET_MASTER=http://localhost:8080`
- Run a simple experiment, e.g., the MNIST example. When the experiment completes, a checkpoint should have been created.
- Obtain the checkpoint UUID under that experiment using the CLI.
- Go to the S3 bucket in the AWS Console to verify the checkpoint with the same UUID.
- Obtain the bearer token for the cluster.
- Use cURL to call the tgz endpoint. For example:
  Substitute `$TOKEN` and `$UUID`, and run:  `curl -v -o out.det-curl.tgz --header "Authorization: Bearer $TOKEN" --header "Accept: application/gzip" http://localhost:8080/checkpoints/$UUID`
- Repeat the last step for the tgz endpoint. For example:
  Substitute `$TOKEN` and `$UUID`, and run:  `curl -v -o out.det-curl.tgz --header "Authorization: Bearer $TOKEN" --header "Accept: application/zip" http://localhost:8080/checkpoints/$UUID`
- Extract both files -- make sure you do not overwrite them. 
- Verify the contents extracted from the tgz and zip files are the same as those in S3.

## Checklist


master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,394 @@
...
+
+	case "application/zip":
+		zw := zip.NewWriter(dw)
+		closers = append(closers, zw)
+
+		return &zipArchiveWriter{zw, nil}, closers, nil
+
+	default:
+		return nil, nil, fmt.Errorf(
+			"MIME type must be applicatoin/gzip or application/zip but got " + mimeType)
```
mackrorysd, line 308: nit:typo (applicatoin)

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,394 @@
...
+			errs = append(errs, iter.Err())
+		}
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		// Return False to stop paging
+		return len(errs) == 0
+	}
+	prefix = "checkpoint_down"
```
hanyucui, line 262: Remove debug code

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,394 @@
...
+
+	checkpointUUID, err := uuid.Parse(args.CheckpointUUID)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest,
+			fmt.Sprintf("unable to parse checkpoint_uuid %s: %s",
+				args.CheckpointUUID, err))
+	}
+
+	// Assume we always have experiment configs for a checkpoint
+	expConfig, err := m.db.ExperimentConfigForCheckpoint(checkpointUUID)
```
hanyucui, line 329: Return a user-friendly error message when the UUID doesn't exist

mackrorysd (commented):
Haven't done a comprehensive review just yet, but on first-pass this is looking good.

master/internal/db/postgres_checkpoints.go
```diff
@@ -111,3 +114,23 @@ func (db *PgDB) GroupCheckpointUUIDsByExperimentID(checkpoints []uuid.UUID) (
 
 	return groupeIDcUUIDS, nil
 }
+
+// ExperimentConfigForCheckpoint looks up the experiment config of the checkpoint.
+func (db *PgDB) ExperimentConfigForCheckpoint(id uuid.UUID) (*expconf.ExperimentConfig, error) {
+	checkpoint, err := db.CheckpointByUUID(id)
```
hanyucui, line 120: Handle not found

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	return aw.tw.Write(p)
+}
+
+type zipArchiveWriter struct {
+	zw        *zip.Writer
+	zwContent io.Writer
+}
+
+func (aw *zipArchiveWriter) WriteFileHeader(fname string, size int64) error {
+	// Zip by default sets mode 0666 and 077 for files and folders respectively
```
hanyucui, line 59: `077` -> `0777`

master/internal/db/postgres_checkpoints.go
```diff
@@ -111,3 +114,27 @@ func (db *PgDB) GroupCheckpointUUIDsByExperimentID(checkpoints []uuid.UUID) (
...
+
+	bytes, err := json.Marshal(checkpoint.CheckpointTrainingMetadata.ExperimentConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	var config expconf.ExperimentConfig
+	err = json.Unmarshal(bytes, &config)
+	if err != nil {
+	}
```
rb-determined-ai, line 137: Any time you are reading a legacy experiment config (a config which was valid when the experiment ran, but the experiment is no longer running and there's no telling how old the config is), you must unmarshal it as an `expconf.LegacyConfig` instead of an `expconf.ExperimentConfig`.  `ExperimentConfig` is really just for experiments which are currently running on the cluster, and after that we have to be able to handle experiment configs which were valid but no longer are.  Check out `expconf.ParseLegacyConfigJSON` which does this for you.

Separately, since I totally forgot that Checkpoint objects have experiment config information already, I'd recommend just using `db.CheckpointByUUID`, parsing the LegacyConfig in `m.getCheckpoint()`, and not adding this function to the db package at all.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	closers := []io.Closer{dw}
+	switch mimeType {
+	case "application/gzip":
+		gz := gzip.NewWriter(dw)
+		closers = append(closers, gz)
+
+		tw := tar.NewWriter(gz)
+		closers = append(closers, tw)
+
+		return &tarArchiveWriter{tw}, closers, nil
```
rb-determined-ai, line 297: to make this more object oriented, can you had the list of closers to the archive writer, and add a `archiveWriter.Close()` method?

Otherwise this is sort of leaking the inner details of the pipeline contents out of the abstraction, by requiring the caller of buildWriterPipeline to keep track of the archive writer and the closers, then call Close on each closer in reverse order.

hanyucui: Yeah, that's a good idea!

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	w.buf = append(w.buf, p...)
+	if len(w.buf) < w.delayBytes {
+		return len(p), nil
+	}
+
+	n, err := w.next.Write(w.buf)
+	w.buf = nil
+	return n, err
```
rb-determined-ai, line 95: This implementation violates the io.Writer interface, since you can't return `n > len(p)`, but if you have a bunch of buffered bytes, the last `return n, err` will always violate that.

Also, this introduces an unnecessary data copy right when the buf exceeds the limit.

I think a better approach would be like:
```go
if w.buf != nil {
  if len(w.buf) + len(p) < w.delayBytes {
    // not enough bytes yet, just wait
    w.buf = append(w.buf, p...)
    return len(p), nil
  }
  // enough bytes, flush old data before writing current data
  _, err := w.next.Write(w.buf)
  w.buf = nil
  if err != nil {
    return 0, err
  }
}
return w.next.Write(p)
```

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	aw.zwContent = zwc
+	return nil
+}
+
+func (aw *zipArchiveWriter) Write(p []byte) (int, error) {
+	var w io.Writer
+	if aw.zwContent == nil {
+		return 0, nil
+	}
```
rb-determined-ai, line 72: What is this check for?

(Also it violates the io.Writer interface, which requires that you either return `n == len(p)` or you return a non-nil error)

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+) error {
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String(getAwsRegion()),
+	})
+	if err != nil {
+		return err
+	}
+	// We do not pass in credentials explicitly. Instead, we reply on
+	// the existing AWS credentials.
+	s3client := s3.New(sess)
```
rb-determined-ai, line 242: This is ok for saas, but can you make a ticket and leave a TODO to support explicitly configured s3 creds?

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+func newDelayWriter(w io.Writer, delayBytes int) *delayWriter {
+	return &delayWriter{
+		delayBytes: delayBytes,
+		buf:        make([]byte, 0, delayBytes),
+		next:       w,
+	}
+}
+
+// S3SeqWriterAt satisfies S3 APIs' io.WriterAt interface while staying sequential.
+// Ref: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#Downloader
```
rb-determined-ai, line 116: nit: can you explicitly mention that this only works when concurrent downloads is set to 1?  Otherwise an unsuspecting dev could think that s3SeqWriterAt is capable of a lot more than it is.

I think the comment where you set concurrent downloads is good, but I think you should have the comment in both places, since it's basically an implicit requirement that developers must follow for things to work right.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError,
+			fmt.Sprintf("unable to retrieve experiment config for checkpoint %s: %s",
+				args.CheckpointUUID, err.Error()))
+	}
+	if expConfig == nil {
+		return echo.NewHTTPError(http.StatusNotFound,
+			fmt.Sprintf("checkpoint %s does not exist", args.CheckpointUUID))
+	}
+	storageUnion := expConfig.CheckpointStorage()
```
rb-determined-ai, line 338: nit: just inline this later?
```go
switch storage := expconfig.CheckpointStorage().GetUnionMember().(type) {
```

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	region, ok := jsonObj["region"].(string)
+	if !ok {
+		return defaultRegion
+	}
+
+	return region
+}
+
+func s3DownloadCheckpoint(
+	c echo.Context, aw archiveWriter, bucket string, prefix string,
```
rb-determined-ai, line 232: is it possible to accept a generic `context.Context` here instead of the `echo.Context`?

I'm thinking accepting an echo.Context makes this way harder to unit test.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+func newS3SeqWriterAt(w io.Writer) *s3SeqWriterAt {
+	return &s3SeqWriterAt{next: w}
+}
+
+// WriteAt writes the content in buffer p.
+func (w *s3SeqWriterAt) WriteAt(p []byte, off int64) (int, error) {
+	if off != w.written {
+		return 0, fmt.Errorf(
+			"only supporting sequential writes,"+
+				" writing at offset %d while %d bytes have been written",
```
rb-determined-ai, line 131: I like the idea of a simple runtime check to make sure the s3 api sticks to its promises!

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+			off, w.written)
+	}
+	n, err := w.next.Write(p)
+	if err != nil {
+		return 0, err
+	}
+	w.written += int64(n)
```
rb-determined-ai, line 138: nit: add this above the error check?

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+	funcReadPage := func(output *s3.ListObjectsV2Output, lastPage bool) bool {
+		iter := newBatchDownloadIterator(aw, bucket, output.Contents)
+		// Download every bucket in this page
+		err = downloader.DownloadWithIterator(c.Request().Context(), iter)
+		if iter.Err() != nil {
+			errs = append(errs, iter.Err())
+		}
+		if err != nil {
+			errs = append(errs, err)
+		}
```
rb-determined-ai, line 257: Question about this code: if `iter.Err()` is set, is it possible for DownloadWithIterator to return any value other than that error?

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+
+		// Return False to stop paging
+		return len(errs) == 0
+	}
+	err = s3client.ListObjectsV2PagesWithContext(
+		c.Request().Context(),
+		&s3.ListObjectsV2Input{
+			Bucket: ptrs.Ptr(bucket),
+			Prefix: ptrs.Ptr(prefix),
```
rb-determined-ai, line 266: Just use `&bucket` and `&prefix`, `ptrs.Ptr` is for cases where the go language makes it difficult to get a pointer, like `ptrs.Ptr(true)` because `&true` is not allowed.

When you already have a `string` on the stack like `bucket`, `ptrs.Ptr(bucket)` is going to create a copy of it on the heap and pass a pointer to that.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,397 @@
...
+type archiveWriter interface {
+	WriteFileHeader(fname string, size int64) error
+	Write(b []byte) (int, error)
+}
+
+type tarArchiveWriter struct {
+	tw *tar.Writer
+}
+
+func (aw *tarArchiveWriter) WriteFileHeader(fname string, size int64) error {
```
rb-determined-ai, line 36: hey, it looks like you can pass all filenames through the same function, so maybe remove `File` from the `WriteFileHeader`?  If there's no `WriteDirHeader` needed then `WriteHeader` or `WriteEntryHeader` or something might make more sense.

master/internal/core_checkpoint.go
hanyucui: When `WriteFileHeader()` hasn't been called or has failed, `aw.zwContent` would be `nil`. Yeah, I need to return a non-nil error here.

master/internal/core_checkpoint.go
hanyucui: Sure.

master/internal/core_checkpoint.go
hanyucui: Makes sense!

master/internal/core_checkpoint.go
hanyucui: Seems doable. In a related question,  for unit tests, how much testing is preferred -- I will also add integration tests.?

master/internal/core_checkpoint.go
hanyucui: Good catch! This PR seems to have hit every scenario the documentation mentions...

master/internal/core_checkpoint.go
hanyucui: Yeah, `iter.Err()` isn't called in [DownloadWithIterator](https://github.com/aws/aws-sdk-go/blob/2ab2a7fda76a11d2df0a0d15e554de2139672d31/service/s3/s3manager/download.go#L259), hence I need to record the error here.

master/internal/core_checkpoint.go
hanyucui: Makes sense. Don't know what I was thinking...

master/internal/core_checkpoint.go
hanyucui: Yeah, that's clearer. Should I also replace `fname` with `path`?

master/internal/core_checkpoint.go
rb-determined-ai: It might be worth including a comment saying it's a guard just in case s3 calls this method even after the Next method returns false?

master/internal/core_checkpoint.go
rb-determined-ai: if it were me, I would only worry about testing against the real s3 api; I wouldn't bother with any fully self-contained unit tests.

master/internal/core_checkpoint.go
rb-determined-ai: lol I had to DM bradley to ask a couple questions about the io.Writer while reviewing

master/internal/core_checkpoint.go
rb-determined-ai: oh fantastic, I was curious where that implementation was.

That's also worth explaining in a comment, imo.

master/internal/core_checkpoint.go
hanyucui: Although we only use it with S3 currently, It doesn't need to be limited to S3. Hence I would mention S3 as an example. Now that I think about it, `s3SeqWriterAt` can well be just `seqWriterAt`.

master/internal/core_checkpoint.go
hanyucui: You can just click the link in my last comment.

master/internal/core_checkpoint_upstream_intg_test.go
```diff
@@ -0,0 +1,241 @@
...
+			return err
+		}, []any{mock.Anything, mock.Anything}},
+	}
+
+	for _, curCase := range cases {
+		require.NoError(t, curCase.IDToReqCall(id))
+	}
+}
+
+func mockExperimentS3(t *testing.T, pgDB *db.PgDB, user model.User, folderPath string) *model.Experiment {
```
hanyucui, line 199: This was copied from `RequireMockExperiment()`. Let me know if you prefer reusing it.

docs/release-notes/checkpoint-download-through-master.txt
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**New Features**
+
+-  API: Support downloading checkpoints through the master. It only supports AWS S3 for now and relies on implicit IAM permissions on the master. Two endpoints are added for downloading checkpoints in different file formats:
```
rb-determined-ai, line 5: Can we leave the release note out?  Generally we discourage users from using the REST API directly, since it has a habit of breaking changes frequently.

If we don't have specific customers asking for this feature, I don't think we should announce it.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,470 @@
...
+	if err != nil || checkpoint == nil {
+		return nil, err
+	}
+
+	bytes, err := json.Marshal(checkpoint.CheckpointTrainingMetadata.ExperimentConfig)
+	if err != nil {
+		return nil, err
+	}
+	spew.Dump(bytes)
+	spew.Dump(checkpoint.CheckpointTrainingMetadata.ExperimentConfig)
```
rb-determined-ai, line 386: is this leftover from some debugging?

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,470 @@
...
+
+	legacyConfig, err := expconf.ParseLegacyConfigJSON(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	return ptrs.Ptr(legacyConfig.CheckpointStorage()), nil
```
rb-determined-ai, line 393: Why bother with ptrs.Ptrs?  Why not just change the return type to `expconf.CheckpointStorageConfig`?

docs/release-notes/checkpoint-download-through-master.txt
hanyucui: Sure. I had no idea.

master/internal/core_checkpoint.go
hanyucui: Yeah, I though I deleted all of them. Let me remove them.

master/internal/core_checkpoint.go
hanyucui: So I can return `nil` in case of an error, which seems nicer than returning a full struct. I don't know what's idiomatic here.

master/internal/core_checkpoint_upstream_intg_test.go
```diff
@@ -0,0 +1,241 @@
+//go:build integration && upstream
+// +build integration,upstream
```
rb-determined-ai, line 2: Is it possible to just detect when the CI environment variables aren't present, and use t.Skip() or something?

I think all you'd have to change in the circleci config is which circleci security context was given to this test.  I think circleci would automatically not give any security context to non-upstream runs of the test.  There would never be overlapping circleci runs.  Then there wouldn't be changes to master/Makefile either I think.

master/internal/core_checkpoint_upstream_intg_test.go
```diff
@@ -0,0 +1,241 @@
...
+	return id.String(), err
+}
+
+func newEchoContext() (echo.Context, *httptest.ResponseRecorder) {
+	e := echo.New()
+	rec := httptest.NewRecorder()
+	return e.NewContext(nil, rec), rec
+}
+
+func TestGetCheckpointEcho(t *testing.T) {
```
rb-determined-ai, line 157: Is it possible to not run the whole echo server here?  That makes this test harder to parse.

Would it be easier to just call BuildWriterPipeline and NewDownloader directly?

Other than that, this test is really detailed and has great checks with it.  I like it a lot.

master/internal/core_checkpoint_upstream_intg_test.go
```diff
@@ -0,0 +1,241 @@
...
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break // End of archive
+		}
+		require.NoError(t, err, "failed to read record header")
+		pathname := strings.TrimPrefix(hdr.Name, prefix)
+		buf := &strings.Builder{}
+		if hdr.Size > 0 {
+			_, err := io.Copy(buf, tr)
+			_, err = io.Copy(os.Stdout, tr)
```
hanyucui, line 105: Remove this copy to stdout -- this is debugging only

master/internal/core_checkpoint_upstream_intg_test.go
hanyucui: I kind of prefer running the whole server. I am less experienced with streaming and would like to test it e2e. The code outside buidWriterPipeline and NewDownloader is also pretty important. E.g., the downloaded file would be truncated if Close() is not called. Are there other ways I can make the test easier to parse?

rb-determined-ai (commented):
Since we are almost 100% confident we'll be doing this for other storage backends, maybe we should just build out a `pkg/storage` module now?

I'm thinking:
- `pkg/storage/archive_writer.go`
- `pkg/storage/checkpoint_downloader.go`
- `pkg/storage/s3.go`
- delayWriter is pretty echo endpoint specific and probably stays where it's at.

I don't think it's the end of the world to leave it how it is, but huge portions of this code will get moved down the road, and it will make the git history harder to trace, which is mostly why I would recommend doing it.

Otherwise this looks great and I anticipate approving it on my next review.

master/internal/core_checkpoint_upstream_intg_test.go
rb-determined-ai: if you like it how it is, I don't feel strongly.

master/internal/core_checkpoint.go
rb-determined-ai: Using pointers where pointers are not necessary is an antipattern, IMO.  I'm not go expert, but that's been my experience in C.  Somebody like @stoksc would have more of an idea what direction we want to take for our codebase, if any.

hanyucui: I think pointers should also be used for large structs? Not sure about this particular case though.

hanyucui:
> Since we are almost 100% confident we'll be doing this for other storage backends, maybe we should just build out a `pkg/storage` module now?
> 
> I'm thinking:
> 
> * `pkg/storage/archive_writer.go`
> * `pkg/storage/checkpoint_downloader.go`
> * `pkg/storage/s3.go`
> * delayWriter is pretty echo endpoint specific and probably stays where it's at.
> 
> I don't think it's the end of the world to leave it how it is, but huge portions of this code will get moved down the road, and it will make the git history harder to trace, which is mostly why I would recommend doing it.
> 
> Otherwise this looks great and I anticipate approving it on my next review.

Thanks, Ryan. Having a storage module sounds good to me. Want to hear what @stoksc thinks.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,470 @@
...
+
+	c.Response().Header().Set(echo.HeaderContentType, mimeType)
+	writerPipe, err := buildWriterPipeline(c.Response(), mimeType)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+	}
+
+	downloader, err := newDownloader(storageConfig, writerPipe, args.CheckpointUUID)
+	err = downloader.download(c.Request().Context())
+	if err != nil {
```
stoksc, line 432: this error handling is kinda funky, maybe explicitly handle both errors?

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,470 @@
...
+		return echo.NewHTTPError(http.StatusInternalServerError,
+			fmt.Sprintf("unable to download checkpoint %s: %s", args.CheckpointUUID, err.Error()))
+	}
+
+	err = writerPipe.Close()
```
stoksc, line 437: deferring the close would be ideal, and i think just logging the close error is probably.. ok? wdyt?

stoksc (commented):
looks pretty good, mostly style stuff. i think overall, i would like to have `pkg/checkpoints` and `pkg/checkpoints/s3`. idk what you think. that way the s3 stuff is more internal to s3. then maybe decouple `getCheckpoint` from `echo.Context` (just give it a context and writer instead maybe?) and it is easy to test it without standing up an echo server.

master/internal/core_checkpoint_upstream_intg_test.go
hanyucui: Sounds good. How about skipping it if `CIRCLE_BRANCH` starts with `pull/`? This is the same logic used in `.circleci/config.yaml`

.circleci/config.yml
```diff
@@ -2172,7 +2176,13 @@ workflows:
...
     jobs:
       - test-intg-downstream
-      - test-intg-master
+      - test-intg-master:
+          name: test-intg-master
+      - test-intg-master:
+          name: test-intg-master-upstream-only
+          make_target: "test-intg-upstream"
+          context: storage-unit-tests
+          filters: *any-upstream
```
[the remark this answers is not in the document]
hanyucui: One of the new tests needs the AWS credentials that are only upstream. I can just skip it when not upstream like Ryan suggested rather than adding another target.

master/internal/core.go
```diff
@@ -1015,6 +1015,10 @@ func (m *Master) Run(ctx context.Context) error {
 	experimentsGroup.PATCH("/:experiment_id", api.Route(m.patchExperiment))
 	experimentsGroup.POST("", api.Route(m.postExperiment))
 
+	checkpointsGroup := m.echo.Group("/checkpoints", authFuncs...)
+	checkpointsGroup.GET("/:checkpoint_uuid/tgz", m.getCheckpointTgz)
+	checkpointsGroup.GET("/:checkpoint_uuid/zip", m.getCheckpointZip)
```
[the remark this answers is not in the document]
rb-determined-ai: can you annotate echo with an Accept-dependent content type?

That was my original reason to have these as separate endpoints.

master/internal/core_checkpoint.go
```diff
@@ -0,0 +1,470 @@
...
+	// calling Write(). The AWS SDK likely will not make this mistake but
+	// zipArchiveWriter is not just limited to being used with AWS.
+	if aw.zwContent == nil {
+		return 0, nil
+	}
+	w = aw.zwContent
+	return w.Write(p)
+}
+
+type delayWriter struct {
```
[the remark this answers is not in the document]
rb-determined-ai: I suppose NewWriterSize would meet the technical requirements, but it would also introduce an extra unnecessary data copy after the initial buffering period is over.  delayWriter only buffers the initial bytes and is transparent after that.

master/internal/core_checkpoint.go
rb-determined-ai: How does io.Closer work in go?  In C, when you close a writable file descriptor, an error there indicates that the final write didn't actually succeed.

master/internal/core_checkpoint.go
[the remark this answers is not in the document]
hanyucui: Yeah, essentially the same except `delayWriter` no long buffers after the buffer is full once. Yeah, I think I can use `bufio`.

master/internal/core_checkpoint.go
hanyucui: Yeah, I missed the one returned by `newDownloader()` if that's what you mean.

master/internal/core_checkpoint.go
hanyucui: Both zip and gzip require Close() to be called to finish writing the whole file. So I think Close() needs to be called before we call c.Response().Flush().

master/internal/core.go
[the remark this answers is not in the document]
hanyucui: I can use the Accept header. Can you explain what's the best way to do it here?

master/internal/core_checkpoint.go
rb-determined-ai: ah, so we do need to gather the error from Close, and it's definitely not something we can log and ignore.

We also need to guarantee that Close is called (in general).

I'd guess that `c.Response.Flush()` is unnecessary, because I imagine the response will get flushed by the echo framework eventually.

So I agree with the need to `defer`, only it'll be slightly complex because of the need to detect the error.  Might be time for a named return value.

hanyucui: When `Close()` is called, we likely have sent a `200`. Not sure what else we can do other than log it in this case. Of course, if a response has not been sent (committed), we can send the error in the response. 

What's the reason to defer here? It seems easier to handle the error if we don't defer. And if any error happens before we call `Close()`, we probably don't care about flushing (which is all `Close()` does).

rb-determined-ai: > When Close() is called, we likely have sent a 200. Not sure what else we can do other than log it in this case

I think it's worth finding out what the Echo framework expects to happen in cases like this.  IMO the echo api clearly expects an API endpoint function to return an error when an error is hit, and so honoring that seems like a bullet-proof solution, even without investigating the edge cases of the echo api.

> What's the reason to defer here?

Idiomatic usage of the interface, I think.  It might not leak resources to not call Close on these particular writers (gzip, tar, and zip writer), but I think it's always best to follow the expectations of the interface you are using.  And these objects have Close methods, which normally must be called in either success or failure cases.

This is an example of what I would call "local understandability".  If I see code which uses a Closer interface, I expect the Close method must be called.  If the close call is deferred, then I don't have to look at any other code to know that resources will be cleaned up.  If the code I'm looking at happens to know exactly which objects back that Closer interface, the code might decide Close doesn't need to be called.  But when reading the code, I can't understand that without reading a bunch more code so that I can verify all the assumptions that make it save to _not_ call Close.  In my opinion, writing code optimized for local understandability pays big dividends in the long run.  It's like the difference between code that is _correct_ vs code that is _obviously correct_.

hanyucui: Thanks for explaining the use of defer. That makes sense. I was on the fence whether to adhere to the convention or bend it to reduce complexity. It would be easier if I don't have to call Flush(). That works when I test it, but let me double-check Echo's code.

hanyucui: Haven't traced it all the way, but echo.ServeHTTP implements [http.Handler](https://pkg.go.dev/net/http#Handler), which does not require flushing or closing the response (doc does not mention closing or flushing at all). Returning from the handler is sufficient.

master/pkg/checkpoints/s3/s3checkpoints.go
```diff
@@ -0,0 +1,196 @@
...
+	pathname := strings.TrimPrefix(*i.objects[i.pos].Key, i.prefix)
+	err := i.aw.WriteHeader(pathname, *i.objects[i.pos].Size)
+	if err != nil {
+		i.err = err
+		return false
+	}
+	return true
+}
+
+// Err() eturns the error if any.
```
hanyucui, line 75: `eturns` -> `returns`

hanyucui:
@rb-determined-ai @stoksc I should have addressed all previous comments. Please take another look.

hanyucui:
I shall have the CLI PR up Thursday.

rb-determined-ai:
> I shall have the CLI PR up soon.

Please add me to that review.

hanyucui:
@rb-determined-ai @stoksc I've added you as reviewers on the [CLI PR](https://github.com/determined-ai/determined/pull/5083).

rb-determined-ai (commented):
This has my approval, but I'll let bradley have the final say here.

stoksc (approved):
this is looks great, thanks. all comments are negotiable, didn't find anything blocking.

master/pkg/checkpoints/s3/s3checkpoints.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	sess, err := session.NewSession(&aws.Config{
+		Region: &region,
+	})
+	if err != nil {
+		return err
+	}
+	// We do not pass in credentials explicitly. Instead, we reply on
+	// the existing AWS credentials.
+	s3client := s3.New(sess)
```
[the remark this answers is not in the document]
hanyucui: Downloading a checkpoint is a heavy-weight request and the cost of creating a client is relatively small. So there is perhaps not much benefit to reusing a client.

master/pkg/checkpoints/s3/s3checkpoints.go
```diff
@@ -0,0 +1,196 @@
...
+type batchDownloadIterator struct {
+	// The objects we are writing
+	objects []*s3.Object
+	// The output we are writing to
+	aw archive.ArchiveWriter
+	// Internal states
+	err    error
+	pos    int
+	bucket string
+	prefix string
```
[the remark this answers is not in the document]
hanyucui: Good catch. Will add another comment for them.

master/internal/core.go
```diff
@@ -1015,6 +1015,9 @@ func (m *Master) Run(ctx context.Context) error {
 	experimentsGroup.PATCH("/:experiment_id", api.Route(m.patchExperiment))
 	experimentsGroup.POST("", api.Route(m.postExperiment))
 
+	checkpointsGroup := m.echo.Group("/checkpoints", authFuncs...)
```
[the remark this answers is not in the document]
hanyucui: Talked to him. The refactor is paused, so I will merge first.

PR #4990, opened 2022-09-09
feat: add stub RBAC API

maxrussell:
## Description

Addresses the OSS portion of [DET-7867] and [DET-8000] by adding a stub version of the RBAC API and its types.

## Test Plan

Outside of EE, the RBAC API should be available, but making requests to its endpoints should result in the error message "Determined Enterprise Edition is required for this functionality." Additionally, creating a new group with an owner should add the owning user to the group regardless of if the user is passed in as one of the users to add to the group.

[DET-7867]: https://determinedai.atlassian.net/browse/DET-7867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-8000]: https://determinedai.atlassian.net/browse/DET-8000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/proto/src/determined/api/v1/api.proto
+++ b/proto/src/determined/api/v1/api.proto
@@ -1670,6 +1670,7 @@ service Determined {
     };
   }
 
+  // Get a set of roles by id
   rpc GetRolesByID(GetRolesByIDRequest) returns (GetRolesByIDResponse) {
     option (google.api.http) = {
       post: "/api/v1/roles/search/by-ids"
@@ -1681,6 +1682,7 @@ service Determined {
     };
   }
 
+  // Get the roles which are assigned to a user
   rpc GetRolesAssignedToUser(GetRolesAssignedToUserRequest)
       returns (GetRolesAssignedToUserResponse) {
     option (google.api.http) = {
@@ -1692,6 +1694,7 @@ service Determined {
     };
   }
 
+  // Get the roles which are assigned to a group
   rpc GetRolesAssignedToGroup(GetRolesAssignedToGroupRequest)
       returns (GetRolesAssignedToGroupResponse) {
     option (google.api.http) = {
@@ -1703,6 +1706,7 @@ service Determined {
     };
   }
 
+  // Search for roles assignable to a given scope
   rpc SearchRolesAssignableToScope(SearchRolesAssignableToScopeRequest)
       returns (SearchRolesAssignableToScopeResponse) {
     option (google.api.http) = {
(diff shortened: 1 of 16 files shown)


proto/src/determined/api/v1/api.proto
```diff
@@ -1668,4 +1669,84 @@ service Determined {
...
+    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "RBAC"
+    };
+  }
+
+  // AssignRoles adds a set of role assignments to the system.
+  rpc AssignRoles(AssignRolesRequest) returns (AssignRolesResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/roles/add-assignments"
```
NicholasBlaskey, line 1735:
```suggestion
      post: "/api/v1/roles/add-assignments"
      body: "*"
```

proto/src/determined/api/v1/api.proto
```diff
@@ -1668,4 +1669,84 @@ service Determined {
...
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "RBAC"
+    };
+  }
+
+  // RemoveAssignments removes a set of role assignments from the system.
+  rpc RemoveAssignments(RemoveAssignmentsRequest)
+      returns (RemoveAssignmentsResponse) {
+    option (google.api.http) = {
+      post: "/api/v1/roles/remove-assignments"
```
NicholasBlaskey, line 1746:
```suggestion
      post: "/api/v1/roles/remove-assignments"
      body: "*"
```

NicholasBlaskey (commented):
Without these the ```bindings.py``` doesn't seen to pickup the body

proto/src/determined/api/v1/api.proto
```diff
@@ -1668,4 +1669,84 @@ service Determined {
       tags: "Internal"
     };
   }
+
+  // Get a set of roles by id
```
mapmeld, line 1673: - Can this be clarified to be roles with the corresponding IDs (clarifying it is plural and is role.id and not a different object id)
- Nitpick, but there's a part of the generated TypeScript bindings which only picks up proto comments if they end with a period, so I'd request adding that and regenerating bindings

ioga (approved):
lgtm

PR #4991, opened 2022-09-09
perf: cache grpcutil.GetUser() result

NicholasBlaskey:
## Description

Motivation is the pattern we have now is we auth the request coming in and for RBAC we then call ```grpc.GetUser``` authing the user a second time. This makes it so for unary requests we just auth once. 

## Test Plan

Existing e2e tests.

## Commentary (optional)

This does technically change the behavior if you send both an allocation token and a user token

before it would check the allocation token then the user token 
but now it will check the user token then the allocation token 

but either way both shouldn't be sent so don't think it matters.

## Checklist

### DIFF
--- a/master/internal/grpcutil/auth.go
+++ b/master/internal/grpcutil/auth.go
@@ -32,6 +32,11 @@ const (
 	cookieName            = "auth"
 )
 
+type (
+	userContextKey        struct{}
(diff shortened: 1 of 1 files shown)


ioga (approved):
lgtm

PR #4996, opened 2022-09-09
fix: group together 4xx and 5xx HTTP responses for prometheus [DET-8336]

azhou-determined:
## Description

HTTP responses outside of /det and /api/v1 are not grouped, meaning 4xx/5xx responses will each create a new Prometheus metric label group and increase cardinality exponentially. 

Solution is to group and obscure all requests outside of /det and /api/v1 returning error codes. 

## Test Plan

- hit many nonexisting urls in Determined, and check /debug/prom/metrics. The urls should be grouped under "/**"
## Checklist


master/internal/core.go
```diff
@@ -1053,6 +1053,16 @@ func (m *Master) Run(ctx context.Context) error {
 
 	if m.config.Observability.EnablePrometheus {
 		p := prometheus.NewPrometheus("echo", nil)
+		// Group and obscure URLs returning 400 or 500 errors outside of /api/v1 and /det
+		p.RequestCounterURLLabelMappingFunc = func(c echo.Context) string {
+			if strings.HasPrefix(c.Path(), "/det/") || strings.HasPrefix(c.Path(), "/api/v1/") {
```
dannysauer, line 1058: Is there a DRY mechanism to get these prefixes from some other "known listening prefixes" list?

Either way, this only partially mitigates the problem; a motivated attacker could still just hit /det/nosuchpage1, for example, and cause the same cardinality explosion.  I'd be inclined to simply group all response codes regardless of prefix in the metrics. Diagnosing what individual URIs are causing an increase in error responses is more of a job for logs than metrics.

dannysauer: I guess I could see a coarse grouping (api v/s ui v/s unknown, perhaps).

azhou-determined: /det/* and /api/v1/* paths are already grouped to /det/* and /api/v1/* "buckets" so this wouldn't really happen. i agree though that ideally we could have some base paths that are sensible "buckets", but currently /det/ and /api/v1/ resolve to the same echo handlers and there's no easy way to come up with a base bucket path.

master/internal/core.go
```diff
@@ -1064,6 +1074,10 @@ func (m *Master) Run(ctx context.Context) error {
 	handler := m.system.AskAt(actor.Addr("proxy"), proxy.NewProxyHandler{ServiceID: "service"})
 	m.echo.Any("/proxy/:service/*", handler.Get().(echo.HandlerFunc))
 
+	m.echo.Any("/*", func (c echo.Context) error {
+		return echo.ErrNotFound
```
stoksc, line 1078: wait this works? don't we still need to set it on the ctx?

azhou-determined: this works, but i'm not entirely sure why. i think this "catchall" returns to echo and some handler downstream sets the error on the context.

stoksc: yeah, i guess if a handler explicitly returns an error it is set somewhere in echo internally? but if no handler matches, the error doesn't actually get set.

master/internal/core.go
stoksc, line 1058: can you add some really good comments explaining why all of this works the way it did (summarizing our digging).

stoksc (commented):
minor comments, otherwise looks good. but let's wait to land it until after a release, just incase.

master/internal/core.go
stoksc: e.g., how the API of RequestCounterURLLabelMappingFunc could be improved to avoid the issue we ran into.

stoksc (approved):

PR #4998, opened 2022-09-09
fix: .detignore interprets wildcards like .gitignore [DET-7094]

drh-determined-ai:
## Description

There is a bug in how Determined deals with wildcard patterns in `.detignore`.

## Test Plan

The new `test_cli.test_read_context_with_detignore_wildcard` should be run by CircleCI and should pass.

## Checklist

### DIFF
--- a/harness/determined/common/context.py
+++ b/harness/determined/common/context.py
@@ -133,8 +133,8 @@ def read_v1_context(
             file_path = pathlib.Path(parent).joinpath(file)
             file_rel_path = file_path.relative_to(root_path)
 
-            # If the file is the .detignore file, it matches one of the
-            #  paths specified in .detignore, or if its parent directory does, then ignore it.
+            # If the file is the .detignore file or matches one of the
+            # paths specified in .detignore, then ignore it.
             if file_rel_path.name == ".detignore":
                 continue
             if ignore_spec.match_file(str(file_rel_path)):
--- a/harness/determined/common/context.py
+++ b/harness/determined/common/context.py
@@ -125,7 +125,7 @@ def read_v1_context(
             entry_path = dir_rel_path.as_posix()
 
             add_item(v1File_from_local_dir(entry_path, dir_path))
-        # We can modify dirs in-place so that ignored directories are not recused into
+        # We can modify dirs in-place so that we do not recurse into ignored directories
         #  See https://docs.python.org/3/library/os.html#os.walk
         dirs[:] = keep_dirs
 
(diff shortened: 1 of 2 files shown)


harness/determined/common/context.py
```diff
@@ -126,13 +128,18 @@ def add_item(f: bindings.v1File) -> None:
         for file in files:
             file_path = pathlib.Path(parent).joinpath(file)
             file_rel_path = file_path.relative_to(root_path)
+            parent_rel_path = pathlib.Path(parent).relative_to(root_path)
 
-            # If the file is the .detignore file or matches one of the
-            # paths specified in .detignore, then ignore it.
+            # If the file is the .detignore file, it matches one of the
+            #  paths specified in .detignore, or if its parent directory does, then ignore it.
```
rb-determined-ai, line 134: hey, that's a tricky bug you are detecting here.  Nice catch.

Though I don't think you're fully solving it.  If `/bad/` is ignored, the `/bad/a/b/c/d` should also be ignored, though checking the direct parent isn't going to be enough.

Really what you need to do is avoid entering an ignored directory in the first place.

According to `os.walk()`'s [docs](https://docs.python.org/3/library/os.html#os.walk), that is possible by removing `directory` from `dirs:

> When topdown is True, the caller can modify the dirnames list in-place (perhaps using [del](https://docs.python.org/3/reference/simple_stmts.html#del) or slice assignment), and [walk()](https://docs.python.org/3/library/os.html#os.walk) will only recurse into the subdirectories whose names remain in dirnames; this can be used to prune the search

drh-determined-ai: Woah, I did not know about that feature of `walk()`. Pretty slick!

rb-determined-ai: cool, now let's undo changes to this comment, since you are no longer changing the code under the comment

drh-determined-ai: Oops! Good catch.

harness/determined/common/context.py
```diff
@@ -109,19 +109,25 @@ def add_item(f: bindings.v1File) -> None:
     # however, the Python documentation claims a warning that rglob may be
     # inefficient on large directory trees, so we use the older os.walk().
     for parent, dirs, files in os.walk(str(root_path)):
+        keep_dirs = []
```
rb-determined-ai, line 112: sorry, got distracted while reviewing.

This is like the tiniest nit, but I think it's easier to think about code when things are happening explicitly rather than implicitly.  Maybe convert this to `skip_dirs`, add to it right before each `continue`, and call `for directory in skip_dirs: dirs.remove(directory)`.

Same effect, but a little easier to think about.

I won't be offended if you totally ignore this comment either, ha.

Can you just do `dirs.remove(directory)` instead of this keep_dirs pattern?  Or does that break the iteration?

rb-determined-ai (approved):
great!

PR #5005, opened 2022-09-12
feat: add cache control headers to static web assets [DET-7450]

julian-determined-ai:
## Description

feat: add cache control headers to static web assets [DET-7450]

Currently we are not sending any cache-control headers with server responses for static assets for the WebUI. This change 
will add cache control headers to static assets. 
Assets that are given a unique filename per deployment we can cache "forever", one year according to [RFC 2616 14.21](https://www.ietf.org/rfc/rfc2616.txt). Assets that we may expect to change regularly such as image files we will only cache for a short period of 5 minutes.

Screen Shot 2022-09-12 at 12 44 30 PM

## Test Plan

1. Navigate to the  Determined website. Open up the network tab. The next few steps can be verified by looking at the `Headers` tab as shown above.
2. Ensure that all of our bundled and hashed files which are the `chunk.js or chunk.css`, `woff`, `woff2`, contain a `Cache-Control` header with a value of `max-age=31536000`
3. Ensure that all image files `png/jpeg/jpg/ico/gif` or any other WebUI static files contain a `Cache-Control` header with a value of `max-age=600`

## Checklist

[DET-7450]: https://determinedai.atlassian.net/browse/DET-7450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/master/internal/core.go
+++ b/master/internal/core.go
@@ -74,6 +74,7 @@ const (
 	maxConcurrentRestores = 10
 	defaultAskTimeout     = 2 * time.Second
 	webuiBaseRoute        = "/det"
+	
 )
 
 // staticWebDirectoryPaths are the locations of static files that comprise the webui.
@@ -722,6 +723,7 @@ func updateClusterHeartbeat(ctx context.Context, db *db.PgDB) {
 	}
 }
 
+
 func (m *Master) postTaskLogs(c echo.Context) (interface{}, error) {
 	var logs []*model.TaskLog
 	if err := json.NewDecoder(c.Request().Body).Decode(&logs); err != nil {
@@ -970,6 +972,7 @@ func (m *Master) Run(ctx context.Context) error {
 	webuiGroup := m.echo.Group(webuiBaseRoute)
 	webuiGroup.File("/", reactIndex)
 	webuiGroup.GET("/*", func(c echo.Context) error {
+
 		groupPath := strings.TrimPrefix(c.Request().URL.Path, webuiBaseRoute+"/")
 		requestedFile := filepath.Join(reactRoot, groupPath)
 		// We do a simple check against directory traversal attacks.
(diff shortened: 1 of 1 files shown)


master/internal/core.go
```diff
@@ -994,6 +994,18 @@ func (m *Master) Run(ctx context.Context) error {
...
+
+		// Files that receive a unique hash when bundled and deployed can be cached forever
+		// Other static files should only be cached for a short period of time
+		cacheFileLongTerm := regexp.MustCompile(`.(chunk\.(css|js)|woff2|woff)$`)
+		cacheFileShortTerm := regexp.MustCompile(`.(ico|png|jpe*g|gif|svg)$`)
+
+		if cacheFileLongTerm.MatchString(requestedFile) {
+			c.Response().Header().Set("cache-control", "max-age=31536000")
+		} else if cacheFileShortTerm.MatchString(requestedFile) {
+			c.Response().Header().Set("cache-control", "max-age=600")
```
ClaireNeveu, line 1006: This seems somewhat "dangerous" to blanket set this across all responses since we run a highly interactive website. Also given the small userbase not much of a win.

master/internal/core.go
```diff
@@ -994,6 +994,18 @@ func (m *Master) Run(ctx context.Context) error {
...
 			hasMatchingFile = !stat.IsDir()
 		}
+
+		// Files that receive a unique hash when bundled and deployed can be cached forever
+		// Other static files should only be cached for a short period of time
+		cacheFileLongTerm := regexp.MustCompile(`.(chunk\.(css|js)|woff2|woff)$`)
+		cacheFileShortTerm := regexp.MustCompile(`.(ico|png|jpe*g|gif|svg)$`)
+
+		if cacheFileLongTerm.MatchString(requestedFile) {
+			c.Response().Header().Set("cache-control", "max-age=31536000")
```
ClaireNeveu, line 1004: You can add the `public` directive here.

julian-determined-ai: Added!

gt2345 (approved):
Works out great. 
I wonder is there a way to actually see the cached content? or just verify the `Cache-Control` header existence is enough.
Also is there a more direct way to distinguish file types other than using regex

julian-determined-ai:
> Works out great. I wonder is there a way to actually see the cached content? or just verify the `Cache-Control` header existence is enough. Also is there a more direct way to distinguish file types other than using regex

Great questions. For the second question, in your browser you should be able to confirm in the network tab that the assets are being loaded from the cache. 

Screen Shot 2022-09-12 at 3 02 43 PM

You can note the `size` column says `memory cache` and the `time` to load the resources is `0ms` which is to be expected. 

The first question is also a good one with our current setup I did think this was the easiest way, especially since we have this custom function handler for webui files. I think something that could be better is if we served assets that we could cache forever from one directory, and all other assets from another. Afterwards, we could setup logic to add the correct headers solely depending on the directory. This could be a great future improvement,

PR #5008, opened 2022-09-12
feat: add echo authentication by default

eecsliu:
## Description
Reapplying the older echo-auth PR but with slightly changed endpoints. 

## Checklist


eecsliu:
@dzhu some of the endpoints has been removed and replaced with just `/api/v1/.*` - it seems like grpc does its own authentication and we therefore don't need to check in echo?

master/internal/user/service.go
```diff
@@ -108,20 +144,34 @@ func (s *Service) UserAndSessionFromRequest(
 	return UserByToken(token, s.extConfig)
 }
 
+// getAuthLevel returns what level of authentication a request needs.
+func (s *Service) getAuthLevel(c echo.Context) int {
+	switch {
+	case adminAuthPointsPattern.MatchString(c.Request().RequestURI):
+		return authAdmin
+	case unauthenticatedPointsPattern.MatchString(c.Path()):
+		return authNone
```
dzhu, line 153: Remove this. I have confirmed that the agents endpoint would be matched by `RequestURI`.

dzhu (approved):

master/internal/user/service.go
eecsliu: looks like this is necessary to pass integration/e2e tests. Will investigate a bit more.

PR #5009, opened 2022-09-12
feat: Add state wait to tensor board [DET-8273]

gt2345:
## Description

When starting a tensor board with no metrics available, state of the tensor board is wait rather than pending. 

## Test Plan

Start tensor board right after experiment starts, during the time when no metric is available, should be about to see the status waiting as shown below.  
Screen Shot 2022-09-12 at 2 53 55 PM

## Checklist


webui/react/src/pages/Wait.tsx
```diff
@@ -41,6 +41,10 @@ const Wait: React.FC = () => {
     message = 'Missing required parameters.';
   } else if (waitStatus && terminalCommandStates.has(waitStatus.state)) {
     message = `${capitalizedTaskType} has been terminated.`;
+  } else if (capitalizedTaskType === 'Tensor-board'
+  && waitStatus
+  && waitStatus?.state === CommandState.Waiting) {
+    message = `Waiting for ${capitalizedTaskType} metrics step to be completed.`;
```
varlogtim, line 47: So, technically we could be waiting for one of the following:

1. The `TensorboardFetcher` hasn't downloaded any files, which occurs when we haven't completed a step which produces metrics.
2. The `tensorboard` process is, for some reason, not starting or is otherwise producing a non-200 HTTP response.

99.99% of the time, it is going to be option 1. However, I wonder if we should change the verbiage here to say, "Waiting for tensorboard process to start. This delay will occur if we have not completed a metrics step." - Or some such thing?

I don't know though :shrug:

varlogtim:
Oops, I deleted my comment after reading your review in more detail. I see now that you place the `tensorboard` in the `waiting` state only if the trial has not actually completed a validation or training step. Nice!

master/internal/api_task.go
```diff
@@ -11,17 +11,31 @@ import (
...
 	case errors.Is(err, db.ErrNotFound):
 		return nil, status.Errorf(
 			codes.NotFound, "task %s not found", req.TaskId)
 	default:
+		// For tensor board task, put it in WAITING state if metrics is not ready.
+		// Use first allocation as an indicator.
+		if t.TaskType == "TENSORBOARD" && len(t.Allocations) > 0 && t.Allocations[0].State == taskv1.State_STATE_PENDING {
+			var tb *tensorboardv1.Tensorboard
+			if err = a.ask(tensorboardsAddr.Child(req.TaskId), &tensorboardv1.Tensorboard{}, &tb); err == nil {
+				if len(tb.ExperimentIds) > 0 {
```
gt2345, line 31: Can we assume experimentIds always exist when we want to check metric?

ioga:
since tensorboard is checking for the presence of the event files [in a loop](https://github.com/determined-ai/determined/blob/master/harness/determined/exec/tensorboard.py#L121), maybe it makes sense to report the "waiting" / ready state from that loop?

then we could easily materialize and store this new `WAITING` state in the database. this allows for automatic filtering using SQL. it'd also allow this state to properly show up in all calls which look at the task state, e.g. `GetTensorboards` or `det tensorboard list` CLI.

@varlogtim what do you think?

ioga:
yes, that's the solution I was thinking of. renaming `check_ready_logs.py` script is not strictly necessary. `/run/determined/check_ready_logs.py --ready-regex "$readiness_regex" --waiting-regex "$waiting_regex"` where the latter is a new optional parameter.

one extra concern that wasn't mentioned is that checking metrics in the db and what tensorboard is doing are not exactly identical. they're closely correlated, sure, but in some failure cases they may drift apart.

master/static/migrations/20220914185603_update-allocation-state.tx.up.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE public.allocations
```
ioga, line 1: @stoksc what's the recommended way to update enums? is this okay or is [this approach with temporary state_](https://github.com/determined-ai/determined/blob/master/master/static/migrations/20220609165106_update-job-type-jobs.tx.up.sql#L1) more preferable?

master/static/srv/check_ready_logs.py
```diff
@@ -24,22 +24,25 @@
...
         master_url,
-        f"/api/v1/allocations/{allocation_id}/ready",
+        f"/api/v1/allocations/{allocation_id}/{state}",
         {},
         cert=cert,
     )
 
 
-def main(ready: Pattern):
+def main(ready: Pattern, waiting: Pattern):
```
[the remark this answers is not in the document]
gt2345: Fixed

master/static/migrations/20220914185603_update-allocation-state.tx.up.sql
gt2345: I switched to using temporary state to stay consistent

master/static/srv/check_ready_logs.py
```diff
@@ -24,22 +24,25 @@
...
         master_url,
-        f"/api/v1/allocations/{allocation_id}/ready",
+        f"/api/v1/allocations/{allocation_id}/{state}",
         {},
         cert=cert,
     )
 
 
-def main(ready: Pattern):
+def main(ready: Pattern, waiting: Optional[Pattern]):
```
ioga, line 36:
```suggestion
def main(ready: Pattern, waiting: Optional[Pattern] = None):
```

ioga (approved):

master/static/srv/check_ready_logs.py
```diff
@@ -24,22 +24,25 @@
...
     master_url = str(os.environ["DET_MASTER"])
     cert = certs.default_load(master_url)
     allocation_id = str(os.environ["DET_ALLOCATION_ID"])
     for line in sys.stdin:
         if ready.match(line):
-            post_ready(master_url, cert, allocation_id)
+            post_ready(master_url, cert, allocation_id, "ready")
+            return
+        if waiting and waiting.match(line):
+            post_ready(master_url, cert, allocation_id, "waiting")
```
stoksc, line 45: random drive by comment, why do we return, shouldn't we be sending both ready and waiting for tb?

gt2345: Great catch! Thank you!

PR #5010, opened 2022-09-12
feat: PyTorch Trainer API

azhou-determined:
## Description

currently the training loop + dtrain implementation. features like profiling, AMP, checkpointing, etc still in progress.

## Checklist

### DIFF
--- a/harness/determined/__init__.py
+++ b/harness/determined/__init__.py
@@ -17,6 +17,7 @@ from determined._execution import (
     InvalidHP,
 )
 from determined import errors
+from determined import pytorch
 from determined import util
 
 # LOG_FORMAT is the standard format for use with the logging module, which is required for the
--- a/harness/determined/_trial_context.py
+++ b/harness/determined/_trial_context.py
@@ -1,6 +1,6 @@
 import logging
 import pathlib
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import determined as det
 from determined import core
@@ -14,10 +14,12 @@ class TrialContext:
     def __init__(
         self,
         core_context: core.Context,
-        env: det.EnvContext,
+        env: Optional[det.EnvContext] = None,
+        hparams: Optional[Dict] = None,
     ) -> None:
         self._core = core_context
         self.env = env
+        self.hparams = hparams
 
         self.distributed = self._core.distributed
         self._stop_requested = False
@@ -96,13 +98,13 @@ class TrialContext:
         """
         Return a dictionary of hyperparameter names to values.
         """
-        return self.env.hparams
+        return self.env and self.env.hparams or self.hparams
 
     def get_hparam(self, name: str) -> Any:
         """
         Return the current value of the hyperparameter with the given name.
         """
-        if name not in self.env.hparams:
+        if name not in self.get_hparams():
             raise ValueError(
                 "Could not find name '{}' in experiment "
                 "hyperparameters. Please check your experiment "
@@ -114,7 +116,7 @@ class TrialContext:
                 "`context.get_global_batch_size()` instead of accessing "
                 "`global_batch_size` directly."
             )
-        return self.env.hparams[name]
+        return self.get_hparams()[name]
 
     def get_stop_requested(self) -> bool:
         """
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -51,9 +51,18 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         det.TrialContext.__init__(self, *args, **kwargs)
         pytorch._PyTorchReducerContext.__init__(self, self.distributed.allgather)
+
+        self.cluster_info = det.get_cluster_info()
+        if self.cluster_info:
+            self.managed_training = True
+        else:
+            self.managed_training = False
+
+        self.experiment_config = self.cluster_info.trial._config
+
         self._per_slot_batch_size, self._global_batch_size = util.calculate_batch_sizes(
             self.get_hparams(),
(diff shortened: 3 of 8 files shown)


harness/determined/pytorch/trainer.py
```diff
@@ -0,0 +1,285 @@
...
+            dist.init_process_group(backend="gloo")  # type: ignore
+        return core.DistributedContext.from_torch_distributed()
+    else:
+        print(f"Backend {distributed_backend} not supported")
+
+
+@contextlib.contextmanager
+def init():
+    distributed_context = initialize_distributed_backend()
+    core_context = core.init(distributed=distributed_context)
```
rb-determined-ai, line 282: this needs to be `with core.init(...) as core_context`.

The way it's written right now, `core_context.close()` doesn't get ever called.

rb-determined-ai: I'd say at least 90% of the time, a contextmanager that doesn't have a `try: yield; finally: ...` or a `with X: yield` is not doing what you want, because if you are using a contextmanager it is so that you can do some cleanup step after execution continues from `yield`.

There's like 10% of the time where the steps you would complete afterwards should only be completed in success cases, and that's the only time you have neither a `try` nor a `with`.

harness/determined/__init__.py
```diff
@@ -17,6 +17,7 @@
     InvalidHP,
 )
 from determined import errors
+from determined import pytorch
```
rb-determined-ai, line 20: no adding mandatory frameworks to determined.  This would be very annoying to pure tensorflow users, or pure core api users who might be using something like scikit learn and neither pytorch nor tensorflow.

harness/determined/_trial_context.py
```diff
@@ -14,10 +14,12 @@ class TrialContext:
     def __init__(
         self,
         core_context: core.Context,
-        env: det.EnvContext,
+        env: Optional[det.EnvContext] = None,
```
rb-determined-ai, line 17: how do you feel about removing the `env` object entirely?

It's basically a pure tech debt item.  In local training mode, it's full of phony values that we make up to do local training.  I would like to see the `env` object not be found anywhere in our pytorch code when the pytorch trainer API lands.

I am not saying you need to do it now before we've finish iterating on the API, of course.  But I think it's important to do in the "implementation and automated testing" milestone, because I suspect there will be things in env that end up needing to be passed in by the user in `pytorch.init()` or something... which is important we don't release an API and figure that out later.

rb-determined-ai: oh, and definitely I'm not proposing any modifications to EstimatorTrialContext or TFKerasTrialContext.  IMO those can continue taking an `env` as part of their subclass definition.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -51,9 +51,18 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         det.TrialContext.__init__(self, *args, **kwargs)
         pytorch._PyTorchReducerContext.__init__(self, self.distributed.allgather)
+
+        self.cluster_info = det.get_cluster_info()
+        if self.cluster_info:
+            self.managed_training = True
+        else:
+            self.managed_training = False
```
rb-determined-ai, line 59: Is this accurate?  What if I want to do local training, but I happen to be in a determined notebook or shell?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -51,9 +51,18 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
...
+
+        self.experiment_config = self.cluster_info.trial._config
```
rb-determined-ai, line 61: Having a `self.experiment_config` on the PytorchTrialContext is sort of antithetical to the goal of supporting local training mode.

How hard/impossible is it to remove it entirely?  What would it take to get there?

harness/determined/_trial_context.py
rb-determined-ai: or... it looks like you already removed all env references from pytorchtrial and pytorchcontext?  If that's the case, then you're like 90% of the way there.

Last step is make sure there's not even a `self.env` that other developers might be tempted to use, ha

harness/determined/pytorch/trainer.py
```diff
@@ -0,0 +1,285 @@
...
+        val_loader = val_data.get_data_loader(
+            repeat=False, num_replicas=num_replicas, rank=rank
+        )
+        return val_loader
+
+    def train(
+        self,
+        max_epochs: Optional[int] = None,
+        # OR
+        max_batches: Optional[int] = None,
```
rb-determined-ai, line 70: why not something like:
`max_length=Epochs(10)` or `max_length=Batches(10000)`?

harness/determined/pytorch/trainer.py
```diff
@@ -0,0 +1,285 @@
...
+        average_aggregated_gradients: bool = True,
+        aggregation_frequency: int = 2,
+        searcher_metric="validation_loss",
+        profiling=True,
+        profiling_start=0,
+        profiling_end=10,
+        sync_timings=None,
+        checkpoint_policy="best|all|none"
+    ):
+        assert (max_epochs is None) ^ (max_batches is None), "Either max_batches or max_epochs must be defined"
```
rb-determined-ai, line 83: neat use of xor, I approve

harness/determined/pytorch/trainer.py
```diff
@@ -0,0 +1,285 @@
...
+            max_batches,
+            min_validation_period,
+            min_checkpoint_period,
+        )
+
+        train_loop.run()
+
+        return
+
+    def on_training_epoch_start(self, epoch_idx: int):
```
rb-determined-ai, line 118: no private helper functions with public-looking names, please use an underscore prefix.

PR #4692, opened 2022-08-01
fix: low contrast issue for dark theme [DET-7958]

hamidzr:
## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](https://github.com/determined-ai/determined/blob/master/docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.
- [ ] If modifying `/webui/react/src/shared/` verify `make -C webui/react test-shared` passes.

PR #5011, opened 2022-09-13
refactor: improve storybook [DET-8099]

EmilyBonar:
## Description
This PR updates and improves storybook, including enabling theme switching and converting (depricated) knobs to controls.

## Test Plan
Open up the Netlify storybook preview and click through the components, confirming they work as expected.

## Checklist


keita-determined:
I'm not a reviewer but let me comment as the one who introduced color theme before.
Dark mode isn't applied to some components 
image
image

EmilyBonar:
> I'm not a reviewer but let me comment as the one who introduced color theme before. Dark mode isn't applied to some components

That's weird. I can reproduce this, but it seems like a combination of refreshing and switching themes will sometimes 'shake it into place', so to speak.

ClaireNeveu:
Potential issues:
- The buttons I'm seeing in the app are mostly white but in storybook we only have blue is that a "primary" button?
- Switches sometimes have labels in the app, also do we use the big version anywhere?
- Should the tooltip be attached to something?
- None of the components under "Ant Design" obey dark mode
- What is `Theme`? It just looks like a button
- Does CheckpointModalTrigger have unique design elements? It just looks like a modal and a button
- How is GridListRadioGroup different than RadioGroup?
- Link disabled state is invisible, is it that way in the app?
- What's the difference between InteractiveTable and ResponsiveTable
- There are a couple of different Selects under different names.
- There is no story for Modal, just a bunch of different modals that are in the app.

EmilyBonar:
> Potential issues:
> 
>     * The buttons I'm seeing in the app are mostly white but in storybook we only have blue is that a "primary" button?
>     * Switches sometimes have labels in the app, also do we use the big version anywhere?
>     * Should the tooltip be attached to something?
>     * None of the components under "Ant Design" obey dark mode
>     * What is `Theme`? It just looks like a button
>     * Does CheckpointModalTrigger have unique design elements? It just looks like a modal and a button
>     * How is GridListRadioGroup different than RadioGroup?
>     * Link disabled state is invisible, is it that way in the app?
>     * What's the difference between InteractiveTable and ResponsiveTable
>     * There are a couple of different Selects under different names.
>     * There is no story for Modal, just a bunch of different modals that are in the app.

- Yes, the blue is a "primary" button. If you're looking at the Ant Design/Button story you can change the `type` control to "default" to see the usual buttons.
- I believe any labels on switches in the app are added separately, not part of the Switch component
- Not sure what you mean? In the Ant Design/Tooltip story it's attached to the `Trigger` text.
- I'll see what I can do about Ant Design and dark mode.
- Yeah, looks like Theme was created by the SaaS team, I'm guessing to test out whether theme switching was working? It's just a button. @hamidzr is it ok to delete `theme.stories.tsx`?
- No, CheckpointModalTrigger doesn't have any unique design elements other than the modal contents.
- GridListRadioGroup just implements RadioGroup with some fixed options. I can remove its story if you'd like.
- Hm, yeah, looks like our disabled Link sets the text color to `var(--theme-ix-inactive)`, which resolves to `rgba(245,245,245,1.0)`, very close to pure white. I'll look into the semantics of our theme system and see if I can't find a better option.
- ResponsiveTable is a light wrapper around antD's Table component that adds useSettings management and some better scrolling behavior. InteractiveTable also wraps antD's Table, but adds right click menus and the ability to resize and reorder columns. We've mostly moved from ResponsiveTable to InteractiveTable over the past couple months.
- You're right, there are. A lot of them have been implemented as separate components because they have special data fetching behavior and to prevent code duplication when we need the same thing in multiple places, but I can remove them from the Storybook if you think that's better.
- Yes, that's true. We're currently developing modals using our useModal hook, which makes it difficult to write a generic story. I can remove the current modals from Storybook if you'd like.

ClaireNeveu:
>     * I believe any labels on switches in the app are added separately, not part of the Switch component
They're actually part of the `Toggle` component.
>     * No, CheckpointModalTrigger doesn't have any unique design elements other than the modal contents.
Can we just do a modal story with the permuations of a modal?
>     * GridListRadioGroup just implements RadioGroup with some fixed options. I can remove its story if you'd like.
Yes, let's
>     * Hm, yeah, looks like our disabled Link sets the text color to `var(--theme-ix-inactive)`, which resolves to `rgba(245,245,245,1.0)`, very close to pure white. I'll look into the semantics of our theme system and see if I can't find a better option.
So I think it's actually the storybook that's wrong here. In the app the primary background color is white but we're using this off-gray to background elements in the storybook
>     * ResponsiveTable is a light wrapper around antD's Table component that adds useSettings management and some better scrolling behavior. InteractiveTable also wraps antD's Table, but adds right click menus and the ability to resize and reorder columns. We've mostly moved from ResponsiveTable to InteractiveTable over the past couple months.
Can we mark `ResponsiveTable` as deprecated then?
>     * You're right, there are. A lot of them have been implemented as separate components because they have special data fetching behavior and to prevent code duplication when we need the same thing in multiple places, but I can remove them from the Storybook if you think that's better.
Story book is just for the UI Kit which is not related to data fetching. We should have a story for the common design element.
>     * Yes, that's true. We're currently developing modals using our useModal hook, which makes it difficult to write a generic story. I can remove the current modals from Storybook if you'd like.
You should use the `useModal` hook for the story.

hamidzr:
> * Yeah, looks like Theme was created by the SaaS team, I'm guessing to test out whether theme switching was working? It's just a button. @hamidzr is it ok to delete `theme.stories.tsx`?

good by me I only added that to test https://github.com/determined-ai/determined/pull/4692 we can probably have that in the antd button one? 

the categorization looks cool

ClaireNeveu (approved):

keita-determined:
image
I still see that theme isn't applied to some components. This isn't a focus on this PR? or it's expected?

PR #5013, opened 2022-09-13
chore: split out ui store part 1 [DET-8218]

hamidzr:
## Commentary (optional)

- [ ] two semi-related todos in the code I rather leave out of this pr

next step: add an option to fetch UI state without importing the whole store but leaving ui accessible from the main useStore.

## Checklist

### DIFF
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -1,16 +1,24 @@
-import React, { Dispatch, useContext, useReducer } from 'react';
-
 import { globalStorage } from 'globalStorage';
+import React, { Dispatch, useContext, useReducer } from 'react';
 import { V1UserWebSetting } from 'services/api-ts-sdk';
-import { DarkLight, Mode, Theme } from 'shared/themes';
 import { clone, isEqual } from 'shared/utils/data';
 import rootLogger from 'shared/utils/Logger';
 import { percent } from 'shared/utils/number';
 import {
-  Agent, Auth, ClusterOverview, ClusterOverviewResource,
-  DetailedUser, DeterminedInfo, PoolOverview, ResourcePool, ResourceType,
-  UserAssignment, UserRole, Workspace,
+  Agent,
+  Auth,
+  ClusterOverview,
+  ClusterOverviewResource,
+  DetailedUser,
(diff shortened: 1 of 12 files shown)


webui/react/src/shared/contexts/UIStore.ts
```diff
@@ -0,0 +1,73 @@
...
+  chromeCollapsed: false,
+  darkLight: DarkLight.Light,
+  isPageHidden: false,
+  mode: Mode.System,
+  showChrome: true,
+  showSpinner: false,
+  theme: {} as Theme,
+};
+
+export enum StoreActionUI {
```
hamidzr, line 23: this is consistent with how we have these in Store but maybe we want a shorter name here

hamidzr:
this should be failing intg-downstream test. I suspect it's because we're adding files and the way we're looking for changes skip those

```
MERGE_BASE="$(git merge-base upstream/master HEAD)"
DIFF_DIRS="$(git diff-tree --no-commit-id --name-only "$MERGE_BASE":webui/react/src HEAD:webui/react/src)"
if [ "$DIFF_DIRS" = "shared" ] ; then
  echo "shared code changed, running downstream test"
else
  echo "shared code not affected, halting job now"
  circleci-agent step halt
fi
```

https://determinedai.atlassian.net/browse/DET-8034

cc @determined-ai/infrastructure

keita-determined (approved):
LGTM

webui/react/src/contexts/Store.tsx
```diff
@@ -48,7 +34,9 @@ export interface State {
   pinnedWorkspaces: Workspace[];
   pool: PoolOverview;
   resourcePools: ResourcePool[];
-  ui: UI;
+  ui: StateUI & {
+    omnibar: OmnibarState; // TODO: this should probably be out of ui
```
keita-determined, line 38: Does that mean omnibar will be completely removed?

hamidzr: by ui I mean not nested under ui. I'm not sure if it's the correct move or not yet

keita-determined: Hmm if its flexible and easy to change in the future, either way is fine?

PR #5015, opened 2022-09-13
chore: setup prettier

keita-determined:
## Description
[DET-8097]

## Test Plan
- Check if all rules are good for all
- Check if prettier works on local
- Check if CI catches perttier error 

## Commentary (optional)
- husky and lint-staged are not included in this PR
- I adjusted some rules to standard rules, such as space in array `[ 1, ,2, 3 ]` -> `[1, 2, 3]`
- I respect the current ESLint rules, I tried not to remove them as much as possible

## Checklist

[DET-8097]: https://determinedai.atlassian.net/browse/DET-8097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/.prettierignore
```diff
@@ -0,0 +1,3 @@
+build
+.vscode
+./src/services/api-ts-sdk/*
```
keita-determined, line 3: Ignore ProtoBuf generated files. What else do we need to ignore?

webui/react/package.json
```diff
@@ -11,9 +11,11 @@
     "postshrinkwrap": "npx npm-force-resolutions && npm install --ignore-scripts",
     "start": "craco start",
     "build": "craco build",
-    "lint": "npm run lint:js && npm run lint:css",
+    "lint": "npm run lint:js && npm run lint:css && npm run prettier:check",
     "lint:js": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
     "lint:css": "stylelint --config .stylelintrc.js \"src/**/*.{css,less,scss}\"",
+    "prettier": "prettier --write src/**/*.{js,jsx,ts,tsx,md,json,scss}",
+    "prettier:check": "prettier --check src/**/*.{js,jsx,ts,tsx,md,json,scss}",
```
keita-determined, line 18: added 2 new commands

ClaireNeveu:
Why is this marked "chore:"?

webui/react/.prettierrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "printWidth": 100,
```
ClaireNeveu, line 2: You only need to include options that are different than the default here which would make it easier to see which settings are custom.

keita-determined: Sometimes Prettier changes default such as https://prettier.io/docs/en/options.html#trailing-commas, so I wanna declarer some important ones explicitly

keita-determined:
> Why is this marked "chore:"?

Is it chore? just formatting

webui/react/src/components/Logo.tsx
```diff
@@ -53,7 +49,7 @@ const logos: Record<BrandingType, Record<Orientation, Record<DarkLight, string>>
 
 const Logo: React.FC<Props> = ({ branding, orientation }: Props) => {
   const { ui } = useStore();
-  const classes = [ css[branding], css[orientation] ];
```
hamidzr, line 56: we might have added an eslint rule before to do the opposite of this.

hamidzr: > I adjusted some rules to standard rules, such as space in array [ 1, ,2, 3 ] -> [1, 2, 3]

I should read the pr description first :)

hamidzr:
effectively I see at least two types of changes:

1. adding new rules and checks for cases we don't have any, assuming the tool auto formats on the rule, 
2. modifying existing rules that are already enforced by ci:

type 1 changes seem easy to add and sound pretty good to me
type 2 changes can cause a lot of very similar code changes affecting the git history, etc, and potentially break a previous assumption we had about it suiting us better. 

regarding type 2 changes: what is being changed here? [the changes to the config file](https://github.com/determined-ai/determined/pull/5015/files?show-viewed-files=true&file-filters%5B%5D=#diff-ccdba6f47a8c28ed527566dcc81dda571042435b30fe0bd1566c0b029735cd93)
- the bracket spacing
- what else?

webui/react/.prettierignore
hamidzr: maybe the same as eslintignore?

hamidzr:
please manually run the saas integ tests as well `make -C webui/react test-shared` the [ci job](https://app.circleci.com/pipelines/github/determined-ai/determined/26987/workflows/a3fd58fa-7ee1-4542-882e-ca3c2d0a1e48/jobs/898990) for it is broken in some cases  [ticket](https://determinedai.atlassian.net/browse/DET-8034)

keita-determined:
> effectively I see at least two types of changes:
> 
> 1. adding new rules and checks for cases we don't have any, assuming the tool auto formats on the rule,
> 2. modifying existing rules that are already enforced by ci:
> 
> type 1 changes seem easy to add and sound pretty good to me type 2 changes can cause a lot of very similar code changes affecting the git history, etc, and potentially break a previous assumption we had about it suiting us better.
> 
> regarding type 2 changes: what is being changed here? [the changes to the config file](https://github.com/determined-ai/determined/pull/5015/files?show-viewed-files=true&file-filters%5B%5D=#diff-ccdba6f47a8c28ed527566dcc81dda571042435b30fe0bd1566c0b029735cd93)
> 
> * the bracket spacing
> * what else?

I removed  `the bracket spacing` and `import grouping`, `react/jsx-wrap-multilines`, `comma-dangle`, `react/jsx-tag-spacing`

keita-determined:
> please manually run the saas integ tests as well make -C webui/react test-shared the [ci job](https://app.circleci.com/pipelines/github/determined-ai/determined/26987/workflows/a3fd58fa-7ee1-4542-882e-ca3c2d0a1e48/jobs/898990) for it is broken in some cases [ticket](https://determinedai.atlassian.net/browse/DET-8034)

It fails cuz SaaS doesnt have prettier in npm

webui/react/.prettierrc.json
ClaireNeveu: These changes are only with major versions and our prettier version will be pinned. Better to simplify this here and then if we need to change it when upgrading prettier we will.

ClaireNeveu: The reason for the default there change was because of a broader change in trailing comma handling in the JS community

hamidzr:
> > please manually run the saas integ tests as well make -C webui/react test-shared the [ci job](https://app.circleci.com/pipelines/github/determined-ai/determined/26987/workflows/a3fd58fa-7ee1-4542-882e-ca3c2d0a1e48/jobs/898990) for it is broken in some cases [ticket](https://determinedai.atlassian.net/browse/DET-8034)
> 
> It fails cuz SaaS doesnt have prettier in npm

thanks that's fine we can add it there as well after. we just need to account for the case where the linters from the two projects disagree on how the code should be formatted long term (ignoring the formatting issues in the shared module couldn't really be an approach because we want all the users of the shared code to be able to contribute new code to it.)

hamidzr:
> > effectively I see at least two types of changes:
> > 
> > 1. adding new rules and checks for cases we don't have any, assuming the tool auto formats on the rule,
> > 2. modifying existing rules that are already enforced by ci:
> > 
> > type 1 changes seem easy to add and sound pretty good to me type 2 changes can cause a lot of very similar code changes affecting the git history, etc, and potentially break a previous assumption we had about it suiting us better.
> > regarding type 2 changes: what is being changed here? [the changes to the config file](https://github.com/determined-ai/determined/pull/5015/files?show-viewed-files=true&file-filters%5B%5D=#diff-ccdba6f47a8c28ed527566dcc81dda571042435b30fe0bd1566c0b029735cd93)
> > 
> > * the bracket spacing
> > * what else?
> 
> I removed `the bracket spacing` and `import grouping`, `react/jsx-wrap-multilines`, `comma-dangle`, `react/jsx-tag-spacing`

what did the resulting code changes look like after? could you please push this up

keita-determined:
@hamidzr 
The previous commits were hard to see the difference of what's removed from ESLint rules, so I added new commits to make it clear

```diff
diff --git a/webui/react/src/shared/configs/eslintrc.js b/webui/react/src/shared/configs/eslintrc.js
index 8904465db..8b43aa41b 100644
--- a/webui/react/src/shared/configs/eslintrc.js
+++ b/webui/react/src/shared/configs/eslintrc.js
@@ -35,14 +35,6 @@ module.exports = {
       'warn',
       { allowArgumentsExplicitlyTypedAsAny: true },
     ],
-    '@typescript-eslint/indent': [
-      'error',
-      2,
-      {
-        ignoredNodes: ['TSTypeParameterInstantiation'],
-        SwitchCase: 1,
-      },
-    ],
     '@typescript-eslint/member-ordering': [
       'error',
       {
@@ -55,7 +47,6 @@ module.exports = {
       { args: 'after-used', ignoreRestSiblings: true },
     ],
     '@typescript-eslint/prefer-optional-chain': ['error'],
-    'array-bracket-spacing': ['error', 'always'],
     'array-element-newline': [
       'error',
       {
@@ -67,19 +58,9 @@ module.exports = {
     'arrow-spacing': ['error', { after: true, before: true }],
     'block-spacing': ['error', 'always'],
     'brace-style': ['error', '1tbs', { allowSingleLine: true }],
-    'comma-dangle': ['error', 'always-multiline'],
     'comma-spacing': ['error', { after: true, before: false }],
     'eol-last': ['error', 'always'],
     'eqeqeq': ['error', 'smart'],
-    'function-paren-newline': ['error', 'multiline-arguments'],
-    'import/order': [
-      'error',
-      {
-        'alphabetize': { caseInsensitive: true, order: 'asc' },
-        'groups': ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
-        'newlines-between': 'always',
-      },
-    ],
     'indent': 'off',
     'jest/valid-title': 'off',
     'jsx-quotes': ['error', 'prefer-double'],
@@ -98,7 +79,6 @@ module.exports = {
       },
     ],
     'keyword-spacing': ['error'],
-    'max-len': ['error', 100, { tabWidth: 2 }],
     'no-console': ['error', { allow: ['warn'] }],
     'no-empty': ['error', { allowEmptyCatch: false }],
     'no-multi-spaces': ['error', { ignoreEOLComments: true }],
@@ -106,15 +86,6 @@ module.exports = {
     'no-throw-literal': 'error',
     'no-trailing-spaces': ['error', {}],
     'no-unused-vars': 'off',
-    'object-curly-newline': [
-      'error',
-      {
-        ExportDeclaration: { consistent: true },
-        ImportDeclaration: { consistent: true },
-        ObjectExpression: { multiline: true },
-        ObjectPattern: { multiline: true },
-      },
-    ],
     'object-curly-spacing': ['error', 'always'],
     'object-property-newline': ['error', { allowAllPropertiesOnSameLine: true }],
     'quote-props': ['error', 'consistent-as-needed'],
@@ -150,18 +121,7 @@ module.exports = {
         closingSlash: 'never',
       },
     ],
-    'react/jsx-wrap-multilines': [
-      'error',
-      {
-        arrow: 'parens-new-line',
-        assignment: 'parens-new-line',
-        condition: 'parens-new-line',
-        declaration: 'parens-new-line',
-        logical: 'parens-new-line',
-        prop: 'parens-new-line',
-        return: 'parens-new-line',
-      },
-    ],
+    'react/jsx-wrap-multilines': ['error', { assignment: false, declaration: false }],
     'react/prop-types': 'off',
     'react/self-closing-comp': [
       'error',
```

For the `max-len`, there is some caution https://prettier.io/docs/en/options.html#print-width

hamidzr:
Thanks for the details changes.

> array-bracket-spacing
> To follow the standard

keeping this one as is:
`+    'array-bracket-spacing': ['error', 'always'],`

avoids:
```
[chore/setup-prettier 52746341a] test: arr spacing
 314 files changed, 2180 insertions(+), 2179 deletions(-)
```
My suggestion is to leave that one out for now until we have a better usecase for it.
(fwiw I tried to dig up if we had any discussion when we added that particular rule but no luck
`glg -G'array-brack.*' webui/react/.eslintrc.js` in https://github.com/determined-ai/pedl turned up a commit but not much more than that)
After taking that change out and rerunning `fmt` the diff was also more readable I took a look and I saw a lot of good changes 👍🏼 . 


some of the ones I noticed that could use more attention:

this one is a similar idea to leaving the ending comma for arrays, etc.
is this related to 
> comma-dangle
> Conflict with Prettier's Trailing Commas rule

DeepinScreenshot_select-area_20220914152245

this one is more subjective: the first (existing) formatting seems like a good balance of readability and verbosity
DeepinScreenshot_select-area_20220914151603

keep in mind I don't know what half these rules mean off the top of my head I can only see the resulting diff that's generated

keita-determined:
>keeping this one as is:
> + 'array-bracket-spacing': ['error', 'always'],

I originally removed to follow the standard, but seems like prettier doesnt allow us to enable it. Probably there is workaround using [this](https://prettier.io/docs/en/integrating-with-linters.html) but it says its not recommended 
https://github.com/prettier/prettier-vscode/issues/259
https://github.com/prettier/prettier/issues/1303

> this one is more subjective: the first (existing) formatting seems like a good balance of readability and verbosity

I agree. First one look better. Let me take a look

ClaireNeveu:
Looking at these rules I think for the most part it makes sense to remove them since they're really just to enforce styling which is now enforced by prettier.

As for the array bracket spacing, it's a change but it's a change from an idiosyncratic choice to the standard approach in most of the JS community. As far as I'm aware prettier doesn't even have an option to add spaces there so it might just be time to move this style to the standard.

ClaireNeveu:
Here is how to ignore commits in blame: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view. Actually, in order to avoid hiding actual code changes we'll need to split this PR up into three once we agree on the rules:

1. Add Prettier commands
2. Run prettier and commit the changes <- we'll ignore this one in blame
3. Add prettier:check into the lint command.

hamidzr:
thanks for checking those cases.

> seems like prettier doesnt allow us to enable it.

Does this mean that if we enable it in eslint, then it will conflict and disagree with prettier all the
time during checks and formats? probably yes.


> there's a git file we can add to exclude this commit from git blame so as not to clutter up history.

ah that's cool I didn't know that.

I personally don't put much weight on what the js community standard might be as it didn't seem that
there is something that everyone agrees on and is THE one (my info is probably a few years old)
We based our original styles on these to a have baseline that's shared by the community.
```
'eslint:recommended',
'plugin:react/recommended',
'plugin:jest/recommended',
'plugin:jest/style',
'plugin:@typescript-eslint/recommended',
'plugin:react-hooks/recommended',
```

If we have to change some of the existing arbitrary style guide to be able to use the rest of
what prettier offers that seems like a good tradeoff to me, let's go for it : )

What's remaining:

Any thoughts on whether we want to introduce these on the shared config or the core specific config
for eslint? :thinking_face: I'll think about this and give these changes a try on project Samuel as well

If we go the shared route we'd probably want to share the prettier config as well.

ClaireNeveu:
I think sharing the configs is the right way to go. I think we should have a base config file for typescript, eslint, prettier, and any other tools that get used in multiple places. All these tools allow extending configuration if some things need to be changed per project (although ideally they don't)

@hamidzr Would you like Keita to introduce those shared configs in a follow-up PR?

hamidzr:
hmm if we don't introduce the shared change here then we'll have a range of commits where saas builds would fail ci. if the change is small to avoid that I'd rather do it in the same commit/pr here. I can try to replicate some of these changes on Samuel just to see if any blockers come up (I don't expect any). let me know if you want to do that.

also, note that it'd be trivial to resolve merge conflicts with other ongoing work in this pr since all the changes are auto-generated but it could be potentially a lot of work the other way around. I'm not sure how much we can optimize for this.


code review:
I'd leave this inline but github ui gets very slow and I'm scared to open that tab

```
-    "lint": "npm run lint:js && npm run lint:css",
+    "lint": "npm run lint:js && npm run lint:css && npm run prettier:check",
```
lint:js and :css might need updates as well. we might either need to split up the prettier's scope between the two or get rid of those two npm commands.


please check that our make targets still do what we think they should do `fmt` and `check` i think they might be missing some changes

keita-determined:
> lint:js and :css might need updates as well. we might either need to split up the prettier's scope between the two or get rid of those two npm commands.

I think we still need lint:js and :css because Prettier:check just check if prettier can still format the code or not. It doesn't detect the rule violation. 
I agree with splitting the prettier command for tsx, css, and other (json, md, etc)

PR #5022, opened 2022-09-14
feat: Modal to add role to group [DET-8220]

mapmeld:
## Description

- New modal, based on the Create/Edit Group modal and its tests.
- Calls `listRoles` for all global roles, `getRolesAssignedToGroup` to filter out current roles, `assignRoles` to add role to a group.
- Require role.id and permission.id fields for convenience. Don't require other fields (such as role.permissions) as these would be annoying to send back to the API.

In the future we might want to list existing roles (and users) and make them removable with checkboxes or other UI.

## Test Plan

In contexts/Store.tsx , replace the `knownRoles: [],` line with a initial demo role (since listRoles will not return content).

```typescript
knownRoles: [{
  id: -1,
  name: 'Test OSS',
  permissions: [{
    id: -2,
    isGlobal: true,
    name: 'oss_user',
  }],
}],  // comma before pinnedWorkspaces
```

## Checklist

### DIFF
--- a/webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
+++ b/webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
@@ -42,7 +42,7 @@ const ModalForm: React.FC<Props> = ({ form, group }) => {
     {
       dataIndex: 'name',
       key: 'name',
-      title: 'Name',
+      title: 'Role Name',
     },
   ];
 
@@ -51,16 +51,16 @@ const ModalForm: React.FC<Props> = ({ form, group }) => {
       form={form}
       labelCol={{ span: 8 }}
       wrapperCol={{ span: 14 }}>
+      {groupRoles.length
+        ? (
+          <Table
+            columns={roleColumns}
+            dataSource={groupRoles}
+            pagination={{ hideOnSinglePage: true, size: 'small' }}
+          />
+        )
+        : null}
       <Form.Item label="Roles" name="roles">
-        {groupRoles.length
-          ? (
-            <Table
-              columns={roleColumns}
-              dataSource={groupRoles}
-              pagination={{ hideOnSinglePage: true, size: 'small' }}
-            />
-          )
-          : null}
         <Select
           loading={isLoading}
           mode="multiple"
--- a/webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
+++ b/webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
@@ -2,7 +2,7 @@ import { Form, message, Select } from 'antd';
 import { FormInstance } from 'antd/lib/form/hooks/useForm';
 import React, { useCallback, useEffect, useState } from 'react';
 
-import { getGroupRoles, updateGroup } from 'services/api';
+import { assignRolesToGroup, getGroupRoles } from 'services/api';
 import { V1GroupSearchResult } from 'services/api-ts-sdk';
 import useModal, { ModalHooks } from 'shared/hooks/useModal/useModal';
 import { ErrorType } from 'shared/utils/error';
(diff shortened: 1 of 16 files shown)


ClaireNeveu:
Can you add screenshots?

gt2345:
Should we distinguish what kind of roles can assign to group directly and what kind of roles need scope?

mapmeld:
@gt2345 I'll add an isGlobal param to the `ListRolesRequest` object and request only isGlobal=true permissions for now.

mapmeld:
Added a test plan above. Here are some screenshots:

Dropdown:
Screen Shot 2022-09-15 at 10 14 17 AM

Mock role appears in dropdown:
Screen Shot 2022-09-15 at 10 14 25 AM

Sent id of mock role payload to assignRoles
Screen Shot 2022-09-15 at 10 34 50 AM

ClaireNeveu:
Can we show the assigned roles in the table?

webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
```diff
@@ -0,0 +1,119 @@
...
+  group: V1GroupSearchResult;
+  onClose?: () => void;
+  roles: UserRole[];
+}
+
+const useModalGroupRoles = ({ onClose, group, roles }: ModalProps): ModalHooks => {
+
+  const [ form ] = Form.useForm();
+
+  const { modalOpen: openOrUpdate, ...modalHook } = useModal();
```
ClaireNeveu, line 74: This is just something I'm curious about, why is `useModal` a hook instead of just a component that takes an `isOpen` property?

webui/react/src/pages/Settings/GroupManagement.tsx
```diff
@@ -84,6 +90,7 @@ const GroupActionDropdown = ({
 const GroupManagement: React.FC = () => {
   const [ groups, setGroups ] = useState<V1GroupSearchResult[]>([]);
   const [ groupUsers, setGroupUsers ] = useState<V1GroupDetails[]>([]);
+  const [ roles, setRoles ] = useState<UserRole[]>([]);
```
ClaireNeveu, line 93: Can you put this in the global Store instead?

mapmeld:
it turns out there are no `isGlobal` roles, because the scope is on the permissions. I will load all roles, then filter this menu based on them having ≥1 global permission.

webui/react/src/pages/Settings/GroupManagement.tsx
mapmeld: ✅

webui/react/src/hooks/useModal/UserSettings/useModalGroupRoles.tsx
mapmeld: I don't know enough about React Hooks to know why we did this (or why we have these instead of using https://ant.design/components/modal/ )

mapmeld:
Added an antd.Table of group's existing roles, if there are any. The only other info available for Roles are id or a comma-separated list of permissions (if user can view_permissions), which don't seem needed here. To test this, you can add a mock role to the initialization of groupRoles, setGroupRoles in `hooks/useModal/UserSettings/useModalGroupRoles.tsx`.

Screen Shot 2022-09-15 at 3 02 16 PM

gt2345 (approved):
Functionality looks good to me, but the role select component is clinging to roles table on dark mode.
Screen Shot 2022-09-16 at 1 31 47 PM

Also the action button says "Edit/Add Roles", but we cannot edit/delete roles in modal, can only add.

mapmeld:
@gt2345  I'll change the text of the label. For styling/UI, I think we need rethink of how to add/remove roles and the other modal adds/removes users. I'll make a ticket for design? But merging the content for now.

PR #5024, opened 2022-09-14
feat: cluster_admin permission and related user/group admin permissions [DET-8229]

mapmeld:
## Description

- The cluster_admin has all possible permissions, so I modified `relevantPermissions` to return `{ has: () => true }` in this case instead of a Set<string>.
- Without OSS admin or permission CAN_ADMINISTRATE_USERS, the RBAC table dropdown menu has only a View Profile option.
- Without OSS admin or permission CAN_UPDATE_GROUP, the RBAC table has no dropdown menu
- I liked how canViewWorkspaces is static, so I had other globally applicable permissions (view-permissions, and the new user-admin and group-admin) also be static.

## Test Plan

- On OSS, non-admin users have fewer options on /settings/user-management and no options /settings/group-management
- relevantPermissions workflow makes sense for OSS admin/non-admin and RBAC cluster_admin / other permissions

## Checklist

### DIFF
--- a/webui/react/src/hooks/usePermissions.ts
+++ b/webui/react/src/hooks/usePermissions.ts
@@ -146,11 +146,13 @@ const usePermissions = (): PermissionsHook => {
 };
 
 // Permissions inside this workspace scope (no workspace = cluster-wide scope)
+// Typically returns a Set<string> of permissions.
 const relevantPermissions = (
   userAssignments?: UserAssignment[],
   userRoles?: UserRole[],
   workspaceId?: number,
-): Set<string> => {
+): { has: (arg0: string
+) => boolean } => {
   if (!userAssignments || !userRoles) {
     // console.error('missing UserAssignment or UserRole');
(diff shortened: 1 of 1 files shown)


webui/react/src/hooks/usePermissions.ts
```diff
@@ -122,11 +134,13 @@ const usePermissions = (): PermissionsHook => {
 };
 
 // Permissions inside this workspace scope (no workspace = cluster-wide scope)
+// Typically returns a Set<string> of permissions.
 const relevantPermissions = (
   userAssignments?: UserAssignment[],
   userRoles?: UserRole[],
   workspaceId?: number,
-): Set<string> => {
+): { has: (arg0: string
```
ClaireNeveu, line 142: Little bit hacky but it's a private function so it's fine.

julian-determined-ai (approved):

ClaireNeveu (approved):

PR #5025, opened 2022-09-14
chore: personal groups get automatically created for users [DET-8363]

NicholasBlaskey:
## Description

Users will now always have a personal group associated with them.

## Test Plan

Intg test and after migration check that determined / admin have a user and user_group_membership in the db. 

## Commentary (optional)

Hard to bunify and use existing group creation code since we have a different non bun transaction in our ```AddUser```, also think there is a cyclical import issue too with ```db``` importing ```usergroups``` 

## Checklist


master/internal/db/postgres_users.go
```diff
@@ -80,6 +80,26 @@ RETURNING id`)
 	return user.ID, nil
 }
 
+func addUserPersonalGroup(tx *sqlx.Tx, userID model.UserID) error {
+	query := `
+INSERT INTO groups(group_name, user_id) 
+SELECT username || 'DeterminedPersonalGroup' AS group_name, id AS user_id FROM users
```
ioga, line 86: maybe `DeterminedPersonalGroup` should be a golang const.

ioga (approved):

PR #5027, opened 2022-09-14
feat: custom searcher examples [MLG-50]

aciborowska:
## Description

Examples of how to implement a custom SearchMethod and run it with LocalSearchRunner and CoreSearchRunner.

## Test Plan
Tested manually. To test, invoke commands from README.md file to run the examples.

## Checklist


examples/custom_search_method/core_search_runner/README.md
```diff
@@ -0,0 +1,36 @@
+# Custom SearchMethod with CoreSearchRunner
+
+This example shows how to implement a custom hyperparameter SearchMethod that enables fault tolerance.
+To run the custom SearchMethod, in this example we use CoreSearchRunner that executes the custom SearchMethod
+in the Determined cluster.
```
mpkouznetsov, line 5:
```suggestion
on the Determined cluster. 
```

examples/custom_search_method/core_search_runner/README.md
```diff
@@ -0,0 +1,36 @@
...
+in the Determined cluster. 
+For an example of running the custom SearchMethod locally, see `examples/custom_searcher/local_search_runner`.
+
+## Files
+* **context_dir/asha.py**: The code for ASHA implemented as a custom SearchMethod.
+* **context_dir/run_experiment.py**: The code for running a custom SearchMethod locally with CoreSearchRunner.
+* **context_dir/model_def.py**: The core code for the model. This includes building and compiling the model.
+* **context_dir/data.py**: The data loading and preparation code for the model.
+* **context_dir/layers.py**: Defines the convolutional layers that the model uses. 
+
```
mpkouznetsov, line 14: It may be confusing to the reader/user to look at the files for both the search runner experiment and the multi-trial experiment alongside each other. Two ideas are:
1. Add a "Note: the context directory effectively contains files for two experiments: a multi-trial deep learning experiment and another experiment whose sole purpose is to wrap a CoreSearchRunner conducting a hyperparameter search within the first experiment." Or...
2. Find a way to put asha.py and run_experiment.py into a subdirectory.

examples/custom_search_method/core_search_runner/context_dir/asha.py
```diff
@@ -0,0 +1,484 @@
...
+  -> Shutdown - closes the experiment.
+
+To support fault tolerance, a custom SearchMethod has to implement save_method_state() and load_method_state(), which
+provide logic for saving and loading any information internal to the SearchMethod,
+such as variables, structures, or models, that are modified as the SearchMethod progresses.
+save_method_state() and load_method_state() are called by CoreSearchRunner as a part of CoreSearchRunner.save() and
+CoreSearchRunner.load() to ensure that CoreSearchRunner state and SearchMethod state can be restored if
+CoreSearchRunner is terminated or interrupted.
+To provide fault tolerance, CoreSearchRunner calls save() and save_method_state() after processing
+each SearcherEvent and before sending the list of Operation to the multi-trial experiment.
```
mpkouznetsov, line 30:
```suggestion
each SearcherEvent and before sending the list of Operations to the multi-trial experiment.
```

examples/custom_search_method/core_search_runner/context_dir/asha.py
```diff
@@ -0,0 +1,484 @@
...
+On resumption, CoreSearchRunner calls load() and load_method_state(), and gets SearcherEvents from the multi-trial
+experiment. If a SearcherEvent has already been processed by SearchMethod, then CoreSearchRunner re-sends the
+operations. Otherwise, it continues to process SearcherEvents as usual. That is, CoreSearchRunner invokes
+the methods corresponding to the SearcherEvent, gets the operations, saves the state and send the operations.
```
mpkouznetsov, line 34:
```suggestion
the methods corresponding to the SearcherEvent, gets the operations, saves the state, and sends the operations.
```

examples/custom_search_method/core_search_runner/context_dir/asha.py
```diff
@@ -0,0 +1,484 @@
...
+
+To ensure that CoreSearchRunner process is resumed automatically in the case of failure,
+make sure to set `max_restarts` in the `searcher.yaml` file to a number greater than 0.
+
+While implementation of save_method_state() and load_method_state() depends on the user,
+in this example we propose to encapsulate all variables required by ASHASearchMethod in a new class,
+ASHASearchMethodState. Having ASHASearchMethodState as a separate object, we propose to save and load
+ASHASearchMethodState object wth pickle for convenience.
```
mpkouznetsov, line 42:
```suggestion
in this example we chose to encapsulate all variables required by ASHASearchMethod in a new class,
ASHASearchMethodState. We elected to use pickle as the storage format for convenience.
```

examples/custom_search_method/local_search_runner/README.md
```diff
@@ -0,0 +1,33 @@
+# Custom SearchMethod with LocalSearchRunner
+
+This example shows how to implement a custom hyperparameter SearchMethod that enables fault tolerance.
+To run the custom SearchMethod, in this example we use LocalSearchRunner that executes the custom SearchMethod 
+on your local machine. 
+For an example of running the custom SearchMethod on a cluster, see `examples/custom_searcher/core_search_runner`.
+
```
mpkouznetsov, line 7: See my comment in core_search_runner/README.md about conceptually separating the 2 experiments.

mpkouznetsov (commented):
A few suggestions

examples/custom_search_method/core_search_runner/README.md
aciborowska: You're right, it will be confusing. To make separation clearer I moved multi-trial experiment files to subdirectory and updated README.md to indicate which files belong to searcher and which to the multi-trial experiment.

aciborowska:
> A few suggestions

Thanks! All suggestions included.

examples/custom_search_method/core_search_runner/searcher.yaml
```diff
@@ -0,0 +1,9 @@
+name: core-search-runner
+entrypoint: python3 run_experiment.py
+searcher:
+  metric: validation_error
+  smaller_is_better: true
+  name: single
+  max_length:
```
liamcli, line 7: did we look into what happens if we omit max_length?

examples/custom_search_method/core_search_runner/run_experiment.py
```diff
@@ -0,0 +1,67 @@
...
+    # To learn more about checkpointing, see https://docs.determined.ai/latest/reference/python-api.html#checkpoint
+    #
+    # While CoreSearchRunner saves its own state and ensures invoking save() and
+    # load() methods when necessary, a user is responsible for implementing
+    # SearchMethod.save_method_state() and SearchMethod.load_method_state() to ensure correct
+    # resumption of the SearchMethod.
+    with det.core.init() as core_context:
+
+        # Instantiate your implementation of SearchMethod
+        search_method = ASHASearchMethod(max_length=1000, max_trials=16, num_rungs=3, divisor=4)
```
liamcli, line 53: let's put these into the `hyperparameters` section of the config below and get them from experiment `info`

examples/custom_search_method/core_search_runner/searcher.yaml
liamcli, line 7: can we exclude max_length here?

examples/custom_search_method/local_search_runner/asha.py
```diff
@@ -0,0 +1,480 @@
...
+    Create,
+    ExitedReason,
+    Operation,
+    SearchMethod,
+    Shutdown,
+    ValidateAfter,
+)
+
+
+class ASHASearchMethod(SearchMethod):
```
liamcli, line 61: is there a difference between `asha.py` for local runner and for core api runner?  if not, we may want to restructure so we have a single example with subdirectories for local and core api but shared asha.py file in the main level directory.

examples/custom_search_method/core_search_runner/run_experiment.py
```diff
@@ -0,0 +1,67 @@
...
+import logging
+import determined as det
+from asha import ASHASearchMethod
+from determined.searcher.core_search_runner import CoreSearchRunner
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO, format=det.LOG_FORMAT)
+
+
+    ########################################################################
```
liamcli, line 25: nice comments!  in terms of style, i usually see triple quotes for multiline comments in python.  also in this specific case, since it's a tutorial example, let's run black formater on the python files.

examples/custom_search_method/core_search_runner/asha.py
```diff
@@ -0,0 +1,483 @@
...
+  -> Close - closes a trial with a given id,
+  -> Shutdown - closes the experiment.
+
+To support fault tolerance, a custom SearchMethod has to implement save_method_state() and load_method_state(), which
+provide logic for saving and loading any information internal to the SearchMethod,
+such as variables, structures, or models, that are modified as the SearchMethod progresses.
+save_method_state() and load_method_state() are called by CoreSearchRunner as a part of CoreSearchRunner.save() and
+CoreSearchRunner.load() to ensure that CoreSearchRunner state and SearchMethod state can be restored if
+CoreSearchRunner is terminated or interrupted.
+To provide fault tolerance, CoreSearchRunner calls save() and save_method_state() after processing
```
liamcli, line 29: ln 29 to 34 seem a little too detailed for this tutorial example.  i suggest removing

examples/custom_search_method/core_search_runner/asha.py
```diff
@@ -0,0 +1,483 @@
...
+    # The hyperparameters are passed to a trial in the Create operation.
+    # In this example, the model (defined in model_def.py) is expecting the following
+    # hyperparameters:
+    #   -> global_batch_size,
+    #   -> n_filters1
+    #   -> n_filters2,
+    #   -> learning_rate,
+    #   -> dropout,
+    #   -> dropout2.
+    def sample_params(self) -> Dict[str, object]:
```
liamcli, line 258: what do you think about having asha take a user defined search space as input so that the search method itself will be agnostic to the actual model training code?  the code below can be moved to perhaps an `utils.py` file and imported int `run_experiment.py`

examples/custom_search_method/core_search_runner/asha.py
```diff
@@ -0,0 +1,483 @@
...
+    # Note: to ensure that your search method is initialized correctly on
+    # resumption, load all objects/models/variables that were saved in
+    # save_method_state().
+    def load_method_state(self, path: Path) -> None:
+        checkpoint_path = path.joinpath("method_state")
+        with checkpoint_path.open("rb") as f:
+            self.asha_search_state = pickle.load(f)
+
+    ############################################################################
+    # ASHA internal implementation details. You can skip this part.
```
liamcli, line 300: let's point users to this doc if they want to understand what's happening here:
https://docs.determined.ai/latest/training/hyperparameter/search-methods/hp-adaptive-asha.html

examples/custom_search_method/core_search_runner/asha.py
```diff
@@ -0,0 +1,483 @@
...
+    # Provides a result of validation according to the metric specified in the
+    # custom_config.yaml: validation_loss.
+    # In this example,ASHASearchMethodState statistics are updated,
+    # and based on the current number of trials and their metric values,
+    # ASHA decides whether to:
+    #    (1) promote a trial, i.e., to continue training for the trial
+    #    (2) start a new trial,
+    #    (3) close experiment if all trials are completed and maximum number of
+    #        trials is reached.
+    def on_validation_completed(self, request_id: uuid.UUID, metric: float) -> List[Operation]:
```
liamcli, line 139: if we wind up passing the op training length here to `on_validation_completed`, is there a way we can show using that somehow for asha? perhaps it'd be useful as an arg to `_promote_async`

examples/custom_search_method/core_search_runner/searcher.yaml
aciborowska: `config is invalid or incomplete: <config>.searcher: max_length is a required property`  so no, we can't exclude it. Unless there is something I don't know...?

examples/custom_search_method/core_search_runner/asha.py
aciborowska: agree. however, I think similar explanation should happen in docs, to make sure users understand what is happening and won't get creative with save/load methods.

examples/custom_search_method/asha_search_method/core_search_runner/searcher.yaml
```diff
@@ -0,0 +1,14 @@
...
+  max_trials: 16
+  num_rungs: 3
+  divisor: 4
+searcher:
+  metric: validation_error
+  smaller_is_better: true
+  name: single
+  max_length:
+    batches: 100
+max_restarts: 0
```
liamcli, line 14: github complaining about missing end of line

liamcli: what do you think about moving hyperparameter.max_trials to be the searcher.max_length so in experiment code searcher.max_length is not ignored but passed to ASHA as the max_trials?

examples/custom_search_method/asha_search_method/experiment_files/config.yaml
```diff
@@ -0,0 +1,9 @@
+name: mnist-custom-search-experiment
+data:
+  url: https://s3-us-west-2.amazonaws.com/determined-ai-test-data/pytorch_mnist.tar.gz
+searcher:
+  name: custom
+  metric: validation_loss
+  smaller_is_better: true
+  unit: batches
```
liamcli, line 8: what is units used for behind the scenes for custom searcher?  is it required?

aciborowska: Units can be either epochs or batches, and yes, specifying these is required. We introduced it since (1) prev default behavior was using epoch if max length was not specified (I think this is not true anymore); and (2) validateAfter needs to know the unit (epoch or batches) (or rather master needs to know in which units validateAfter is talking).

Relevant thread is here: https://determined-ai.slack.com/archives/C03M8MBFK1N/p1660167154095449

examples/custom_search_method/asha_search_method/core_search_runner/searcher.yaml
aciborowska: That's a good idea in theory :) but in practice it requires accessing searcher config through `_config`, e.g., `det.ExperimentConfig(info.trial._config)` or `get_cluster_info().trial._config['searcher']['metric']` which according to our code
```
 _config is private because it's not a stable API; as the experiment config version increases, the shape of the experiment config dict might change as the master is upgraded. As a result, we should strongly discourage users from reading the experiment config dict in their training code.
```
Long story short, I don't think we should advertise `_config`. I don't know a different (safe) way of accessing searcher metric, but if there is one, I can update the example.

PR #5032, opened 2022-09-15
chore: separate use ui store and dispatch

hamidzr:
## Description

- separate ui store
- introduce a new pattern to update the ui store/state
- remove dependency on the generic store, which is not shared, from shared components

## Test Plan

- flip through some ui pages
- look at the diff to make sure dispatch calls are replaced w/ correct actions.call

we could write unit tests..

## Checklist

### DIFF
--- a/webui/react/src/components/ThemeToggle.tsx
+++ b/webui/react/src/components/ThemeToggle.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
-import { useStore, useStoreDispatch } from 'contexts/Store';
-import { StoreActionUI } from 'shared/contexts/UIStore';
+import useUI from 'shared/contexts/stores/UI';
 import { Mode } from 'shared/themes';
 
 import css from './ThemeToggle.module.scss';
@@ -31,17 +30,19 @@ export const ThemeOptions: Record<Mode, ThemeOption> = {
 };
 
 const ThemeToggle: React.FC = () => {
-  const { ui } = useStore();
-  const storeDispatch = useStoreDispatch();
+  const {
+    ui: { mode: uiMode },
+    actions: { setMode },
+  } = useUI();
 
   const classes = [css.toggler];
-  const currentThemeOption = ThemeOptions[ui.mode];
+  const currentThemeOption = ThemeOptions[uiMode];
   classes.push(css[currentThemeOption.className]);
 
   const newThemeMode = (e: React.MouseEvent) => {
     e.stopPropagation();
     e.preventDefault();
-    storeDispatch({ type: StoreActionUI.SetMode, value: currentThemeOption.next });
+    setMode(currentThemeOption.next);
   };
 
   return (
--- a/webui/react/src/pages/Wait.tsx
+++ b/webui/react/src/pages/Wait.tsx
@@ -5,11 +5,10 @@ import { useParams } from 'react-router-dom';
 import Badge, { BadgeType } from 'components/Badge';
 import PageMessage from 'components/PageMessage';
 import { terminalCommandStates } from 'constants/states';
-import { useStoreDispatch } from 'contexts/Store';
 import { serverAddress } from 'routes/utils';
 import { getTask } from 'services/api';
 import Spinner from 'shared/components/Spinner/Spinner';
-import { StoreActionUI } from 'shared/contexts/UIStore';
+import useUI from 'shared/contexts/stores/UI';
 import { ErrorType } from 'shared/utils/error';
 import { capitalize } from 'shared/utils/string';
 import { CommandState } from 'types';
@@ -29,7 +28,9 @@ interface Queries {
 }
 
 const Wait: React.FC = () => {
-  const storeDispatch = useStoreDispatch();
+  const {
+    actions: { showChrome, hideChrome },
+  } = useUI();
   const { taskType } = useParams<Params>();
   const [waitStatus, setWaitStatus] = useState<WaitStatus>();
   const { eventUrl, serviceAddr }: Queries = queryString.parse(location.search);
@@ -51,9 +52,9 @@ const Wait: React.FC = () => {
   }
 
   useEffect(() => {
-    storeDispatch({ type: StoreActionUI.HideUIChrome });
-    return () => storeDispatch({ type: StoreActionUI.ShowUIChrome });
-  }, [storeDispatch]);
+    hideChrome();
+    return showChrome;
+  }, [hideChrome, showChrome]);
 
   const handleTaskError = (e: Error) => {
     handleError(e, {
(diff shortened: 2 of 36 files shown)


webui/react/src/components/ThemeToggle.tsx
```diff
@@ -31,8 +31,8 @@ export const ThemeOptions: Record<Mode, ThemeOption> = {
 };
 
 const ThemeToggle: React.FC = () => {
-  const { ui } = useStore();
-  const storeDispatch = useStoreDispatch();
+  const ui = useUI();
+  const uiDispatch = useUIDispatch();
```
ClaireNeveu, line 35: Could I impose on you to make this more semantic? Something like:
```typescript
const [ui, toggleDarkMode] = useUI()
toggleDarkMode()
```
instead of asking a user to manually plumb data around different hooks?

hamidzr: Hmm I think are more design decisions that will go into that compared to this could that be a separate pr/ticket?

ClaireNeveu: I think now where you're introducing a new API is the right place for that. What are your thoughts on a separate ticket?

hamidzr: in terms of maybe wanting to take time to define what APIs we want to expose eg if we have a toggle do we want the explicit version of it or not, should we combine some of these etc

hamidzr:
I need to move this pr to an upstream branch https://circleci.com/gh/determined-ai/determined/963572

webui/react/src/components/ThemeToggle.tsx
ClaireNeveu: Don't worry about getting it perfect on the first go. Use your judgement and it'll be an improvement over what it is now.

webui/react/src/shared/contexts/UIStore.ts
```diff
@@ -38,6 +42,42 @@ export type ActionUI =
   | { type: StoreActionUI.SetTheme; value: { darkLight: DarkLight; theme: Theme } }
   | { type: StoreActionUI.ShowUIChrome }
   | { type: StoreActionUI.ShowUISpinner };
+
+// CHECK: we don't really need this as a class but this
+// looked like the cleanest way to advertise the type in one package.
+export class UIActions {
```
ClaireNeveu, line 48: Just think of the `class` keyword in JS as light codegen. It's not fundamentally different than a regular object, it just adds some nice closures for you.

ClaireNeveu (approved):
This is looking good. Thoughts on using a separate context provider for the store information?

hamidzr:
thanks for the review.

> This is looking good. Thoughts on using a separate context provider for the store information?

what do you mean by the store information? as in separating the provider for ui vs the rest? 

If so that sounds good but if I'm understanding it correctly it wouldn't lend itself nicely to a transition period where not all modules that are using ui state are using `useUI` and we'd have to update all the references before introducing `useUI`

ClaireNeveu:
Yes, I meant a separate provider for ui. Yes, you would need to update the references. Is your plan here to update the references and then switch to another context provider in later PRs? If so that sounds fine to me.

@trentwatt and I were doing some work on [splitting out agents](https://github.com/determined-ai/determined/compare/store-refactor) into its own context (and adding proper loading states).

hamidzr:
my plan was to not update the references here so I can land this sooner so we can use it sooner. but there aren't that many references and it'll let me remove a dependency on Store that was added previously unintentionally in the shared code so I think I should do it here

ClaireNeveu:
@hamidzr Okay. I'm amenable to either approach. If you update it in this PR can you ping me to re-review?

webui/react/src/pages/Wait.tsx
```diff
@@ -29,7 +28,9 @@ interface Queries {
 }
 
 const Wait: React.FC = () => {
-  const storeDispatch = useStoreDispatch();
+  const {
+    actions: { showChrome, hideChrome },
```
hamidzr, line 32: could be single line..

webui/react/src/shared/contexts/stores/UI.tsx
```diff
@@ -0,0 +1,143 @@
...
+export const StoreProvider: React.FC<{ children?: React.ReactNode }> = ({ children }) => {
+  const [state, dispatch] = useReducer(checkDeepEquality(reducer, logger), initUI);
+  return (
+    <StateContext.Provider value={state}>
+      <DispatchContext.Provider value={dispatch}>{children}</DispatchContext.Provider>
+    </StateContext.Provider>
+  );
+};
+
+export default useStore;
```
hamidzr, line 143: most of the interesting changes are here mainly decision around import and access patterns.
- shifted the default export to be useStore/useUI instead of provider . this should be shorter to typeout to import and it'll be used most frequently but now my ide seems to not be able to auto-suggest an import 🤷🏼‍♂️ 

- naming: sticking with generic names and schemas between multiple potential future useXs `{state, actions} = useUI()` instead of say {ui, uiActions}

- usage pattern: uiActions.xyz or to deconstruct: here I optimized for how I can wrap this up faster.

hamidzr, line 143: nit: should be useUI? maybe that's why the LSP can't figure out what I wanna import

webui/react/src/shared/contexts/stores/UI.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  public hideChrome = (): void => {
+    this.dispatch({ type: StoreActionUI.HideUIChrome });
+  };
+
+  public hideSpinner = (): void => {
+    this.dispatch({ type: StoreActionUI.HideUISpinner });
+  };
+
+  public setMode = (mode: Mode): void => {
```
hamidzr, line 60: these could use some jsdoc comments. I'll leave the motu for now. there is a rebase left to do

ClaireNeveu (approved):
This looks great! One minor TODO in a future PR: the `useTheme` hook really seems like it should just be logic inside the `UIProvider` component.

webui/react/src/shared/contexts/stores/UI.tsx
hamidzr: done

webui/react/src/shared/contexts/stores/UI.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const initUI: StateUI = {
+  chromeCollapsed: false,
+  darkLight: DarkLight.Light,
+  isPageHidden: false,
+  mode: Mode.System,
+  showChrome: true,
+  showSpinner: false,
+  theme: {} as Theme,
```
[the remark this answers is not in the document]
hamidzr: should be easy to update but let's skip it just to avoid adding to the list of issues this pr could introduce.

hamidzr:
moving this to an upstream branch so the tests can run https://github.com/determined-ai/determined/pull/5338

PR #5033, opened 2022-09-15
feat: setup `Prettier`

keita-determined:
## Description
- related: #5015
- This is the first step of https://github.com/determined-ai/determined/pull/5015#issuecomment-1248376734

## Test Plan
- Check if Prettierrc.js file is shared
- Check if rules in eslint is correct https://github.com/determined-ai/determined/pull/5015#issuecomment-1247309010
- Check if there's nothing I missed

## Checklist


webui/react/package.json
```diff
@@ -14,6 +14,14 @@
...
     "lint:js": "eslint \"src/**/*.{js,jsx,ts,tsx}\"",
     "lint:css": "stylelint --config .stylelintrc.js \"src/**/*.{css,less,scss}\"",
+    "prettier": "npm run prettier:js && npm run prettier:css && npm run prettier:misc",
+    "prettier:js": "prettier --write \"src/**/*.{js,jsx,ts,tsx}\"",
+    "prettier:css": "prettier --write \"src/**/*.{css,less,scss}\"",
+    "prettier:misc": "prettier --write \"src/**/*.{md,json}\"",
+    "prettier:check": "npm run prettier:check:js && npm run prettier:check:css && npm run prettier:check:misc",
+    "prettier:check:js": "prettier --check \"src/**/*.{js,jsx,ts,tsx}\"",
+    "prettier:check:css": "prettier --check \"src/**/*.{css,less,scss}\"",
+    "prettier:check:misc": "prettier --check \"src/**/*.{md,json}\"",
```
keita-determined, line 24: I split up prettier command, not sure if its useful tho

webui/react/src/shared/configs/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
keita-determined, line 1: share `prettierrc` file

webui/react/package.json
ClaireNeveu: The one suggestion I have here is to use the generic `format` instead of "prettier"

ClaireNeveu (approved):

webui/react/package.json
ClaireNeveu: Approved after this change

hamidzr:
to make sure we're running the same thing: after all is said and done what is 
- the command devs would use to do all the auto-formatting
- the command devs/ci would use to check all the formatting is correct

hamidzr:
nit: can we set it so that it actually tells us what it is complaining about? 

DeepinScreenshot_select-area_20220915153352

it's good that the format can solve most (all?) of these issues but if this doesn't show what's wrong this could be confusing once the ci fails and reports with "i don't like the formatting" but doesn't show what's wrong

hamidzr (approved):
LGTM. 
I added a minor comment about prettier error reporting
It might help to check in with the team about the potential merge conflicts before landing this

keita-determined:
[test-intg-downstream](https://circleci.com/gh/determined-ai/determined/904396?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) fails because SaaS needs change too, but we will update SaaS soon, and wouldn't be a problem

keita-determined:
> nit: can we set it so that it actually tells us what it is complaining about?
> 
> DeepinScreenshot_select-area_20220915153352
> 
> it's good that the format can solve most (all?) of these issues but if this doesn't show what's wrong this could be confusing once the ci fails and reports with "i don't like the formatting" but doesn't show what's wrong

I checked https://prettier.io/docs/en/cli.html if Prettier CLI has `verbose` option, but they don't have it unfortunately. Usually, running `npm run format` will solve `npm run format-check` issue

PR #5035, opened 2022-09-15
feat: support slots_per_trial=0 in Trial classes

rb-determined-ai:
## Description

Apparently batch size calculation is the only reason left why zero-slot
experiments didn't work.

## Test Plan

- [x] test pytorch example with slots_per_trial=0
- [x] test keras example with slots_per_trial=0

### DIFF
--- a/harness/determined/launch/horovod.py
+++ b/harness/determined/launch/horovod.py
@@ -240,7 +240,7 @@ def parse_args(args: List[str]) -> Tuple[List[str], List[str], bool]:
     )
 
     # --autohorovod is an internal-only flag.  What it does is it causes the code skip the
-    # horovodrun wrapper when slots_per_trial == 1.  This has two effects:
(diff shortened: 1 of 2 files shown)


harness/determined/launch/horovod.py
```diff
@@ -240,7 +240,7 @@ def parse_args(args: List[str]) -> Tuple[List[str], List[str], bool]:
     )
 
     # --autohorovod is an internal-only flag.  What it does is it causes the code skip the
-    # horovodrun wrapper when slots_per_trial == 1.  This has two effects:
+    # horovodrun wrapper when slots_per_trial <= 1.  This has two effects:
```
mpkouznetsov, line 243: PP: `is 0 or 1`. Not a strong preference.

mpkouznetsov (approved):
This is _it_? Wow.

PR #5036, opened 2022-09-15
feat: WebUI add feature switch [DET-8352]

gt2345:
## Description

Add feature switch through `useFeature` in WebUI.

## Test Plan

Log in as the admin user
- navigates to `/det/settings/account`, you should be able to add/edit/delete user.
- navigates to `/det/settings/account?f_rbac=on`, you should be able to manage users and groups.

Log in as a regular user
- navigates to `/det/settings/account`, you should only be able to view the `Account` tab.
- navigates to `/det/settings/account?f_rbac=on`, you should only be able to view `User Management` and `Group Management` tabs, but without necessary permissions, you should not be able to make any changes.

## Checklist


webui/react/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const useFeature = (): FeatureHook => {
+  return { isOnRbac: () => IsOnRbac() };
+};
+
+const IsOnRbac = (): boolean => {
+  const { info: { rbacEnabled } } = useStore();
+  const location = useLocation();
+  const { rbac }: Queries = queryString.parse(location.search);
```
ClaireNeveu, line 21: We should only check the url on page load, not whenever it changes. You can get the url directly from the JS standard API. Also can you check for `f_rbac` instead of just `rbac`.

webui/react/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,25 @@
+import queryString from 'query-string';
+import { useLocation } from 'react-router-dom';
+
+import { useStore } from 'contexts/Store';
+
+interface FeatureHook {
+  isOnRbac: () => boolean
```
ClaireNeveu, line 7: Just to kind of get the ball rolling on using feature switches more broadly could we make this 
```
type ValidFeature = 'rbac' // Add new feature switches here using `|`
interface FeatureHook {
   isOn: (ValidFeature) => boolean
}
```
General behavior for `isOn(someString)` should be to check if the page was loaded with a query parameter of `?f_someString=on`. You can add a special case for rbac to also check the master config info.

ClaireNeveu (changes requested):

webui/react/src/hooks/useFeature.ts
gt2345: If we have `ValidFeature` as a "list" of features, should we have return value as a list, too?
Can we have `isOn` function just takes individual feature instead of a list?

webui/react/src/hooks/usePermissions.ts
```diff
@@ -142,6 +166,47 @@ const relevantPermissions = (
...
+};
+
+const canModifyGroups = (
+  user?: DetailedUser,
+  userAssignments?: UserAssignment[],
+  userRoles?: UserRole[],
+): boolean => {
+  const permitted = relevantPermissions(userAssignments, userRoles);
+  return !!user && (permitted.has('oss_user') ? user.isAdmin
+    : permitted.has('modify_users'));
```
mapmeld, line 207: this should be modify_groups

webui/react/src/hooks/useFeature.ts
ClaireNeveu: `ValidFeature` isn't a list, it's a type to prevent typos. For runtime purposes it's a string. So if you do `type ValidFeature = 'rbac' | 'triangle_mode' then `useFeature().isOn('rbac')` will compile but `useFeature().isOn('rbc')` will be a typescript error.

webui/react/src/hooks/useFeature.ts
```diff
@@ -1,25 +1,32 @@
...
 
 import { useStore } from 'contexts/Store';
 
 interface FeatureHook {
-  isOnRbac: () => boolean
-}
-
-interface Queries {
-  rbac?: string;
+  isOn: (ValidFeature: string) => boolean
```
ClaireNeveu, line 8: This should be `isOn: (feature: ValidFeature) => boolean`.

webui/react/src/hooks/useFeature.ts
```diff
@@ -1,25 +1,32 @@
...
 }
 
 const useFeature = (): FeatureHook => {
-  return { isOnRbac: () => IsOnRbac() };
+  const location = useLocation();
+  const [ features, setFeatures ] = useState<string[]>([]);
+  useEffect(() => {
+    const queries = queryString.parse(location.search);
```
ClaireNeveu, line 15: You should do this globally at the top of the file so it happens when the application loads instead of when the hook is used.

webui/react/src/hooks/useFeature.ts
```diff
@@ -1,25 +1,32 @@
...
-
-interface Queries {
-  rbac?: string;
+  isOn: (ValidFeature: string) => boolean
 }
 
 const useFeature = (): FeatureHook => {
-  return { isOnRbac: () => IsOnRbac() };
+  const location = useLocation();
+  const [ features, setFeatures ] = useState<string[]>([]);
```
ClaireNeveu, line 13: You don't need any state for this hook.

webui/react/src/hooks/useFeature.ts
gt2345: I see! Thanks!

webui/react/src/hooks/useFeature.ts
ClaireNeveu: All you actually have to do is check ``queryParams[`f_${feature}`] === 'on'`` and then have special handling when `feature === 'rbac'` to check the master config info.

gt2345: Yes, this is better, thanks!

webui/react/src/pages/Settings/Settings.test.tsx
```diff
@@ -29,10 +29,10 @@ describe('Settings Page', () => {
   });
 
   it('should enable user management for admins only', () => {
-    setup();
 
     // TODO: Change out to set user to be admin and check for user management tab.
-    history.push('/rbac-enabled');
+    history.push('/?f_rbac=on');
```
ClaireNeveu, line 34: We should mock this feature on in `setup` above.

webui/react/src/hooks/useFeature.ts
```diff
@@ -0,0 +1,26 @@
...
+interface FeatureHook {
+  isOn: (feature: ValidFeature) => boolean
+}
+
+const useFeature = (): FeatureHook => {
+  return { isOn: (ValidFeature) => IsOn(ValidFeature) };
+};
+
+const IsOn = (feature: string): boolean => {
+  const { info: { rbacEnabled } } = useStore();
```
ClaireNeveu, line 17: You should move this into `useFeature` and inline the rest of `IsOn` otherwise you could potentially get hooks errors when using this.

ClaireNeveu (changes requested):
Just one more issue and then it looks good. I also notice a lot of unrelated changes in this PR though, what's up with that?

webui/react/src/hooks/useFeature.ts
ClaireNeveu: Oh I see, so these are pages that are now only linked to if rbac is on? In That case this looks good to me.

PR #5037, opened 2022-09-15
chore: warn users about certain LR schedulers

rb-determined-ai:
There are two pytorch LR schedulers which cause trouble when combined with horovod dtrain.

The solution is simple to implement, so let's just detect the problem and tell users exactly how to fix it.

### DIFF
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -279,6 +279,45 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
                     '        self.wrapped_lr_scheduler.step(metrics["validation_error"])\n'
                 )
 
+        if type(lr_scheduler) is torch.optim.lr_scheduler.OneCycleLR:
+            # XXX: is horovod necessary to reproduce this?
+            if self._distributed_backend.use_horovod():
+                raise det.errors.InvalidExperimentException(
+                    "detected that context.wrap_lr_scheduler() was called with an instance of "
+                    "torch.optim.lr_scheduler.OneCycleLR as the lr_scheduler.  This lr scheduler "
+                    "does not support serialization when combined with horovod distributed "
+                    "training but you can correct the problem with a small wrapper class:\n"
+                    "\n"
+                    "class MyOneCycleLR(torch.optim.lr_scheduler.OneCycleLR):\n"
+                    "    def state_dict(self):\n"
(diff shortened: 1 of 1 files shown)


coreystatendet:
Summary of the situation: These LR schedulers set bound methods as attributes (e.g. `self.var = self.my_method`) and then fail to exclude them from `state_dict`.  `pickle` attempts to save the whole object when saving a bound method, so even though `optimizer` is excluded from the state_dict, we end up trying to save it anyways since it's an attribute of the scheduler.  This actually works with default PyTorch optimizers since they implement a sane `__getstate__ / __setstate__`, but does not work with Horovod optimizers.

I think warning and offering a workaround makes sense, but there's a maintenance issue -- the necessary workaround for `CyclicLR` on PyTorch `master` has changed in the last few days, for instance.

We could attempt an automatic fix: check the LR scheduler `state_dict` for bound methods and `weakref.WeakMethod`'s ala https://github.com/pytorch/pytorch/blob/9f1468ae6c8e6e3938a3f1cfb9378e11af2fd0cd/torch/optim/lr_scheduler.py#L1219 and exclude them.  We could log a warning when doing this.  This could in principle behave differently than users expect -- say they have a serializable instance and intentionally store a bound method to it intending to implicitly save a copy of the instance.  But I expect most cases of this occurring are errors, and they could workaround by adding the object directly to the state_dict and regenerating the bound method in `load_state_dict`.

Does that sound like a good fix?

rb-determined-ai:
This only comes up when horovod dynamically subclasses one of these LRSchedulers, right?

Because if I just make a plain old `OneCycleLR` I can pickle it without issue.

I guess my confusion here is "what actually needs pickling and what doesn't".  Like, can we just upstream a fix to pytorch to not include those fields in the state dict?

coreystatendet:
If I understand correctly, it's not about Horovod doing anything with the LR Schedulers.  Instead, it's that those particular LR schedulers are always returning their optimizer implicitly in `__getstate__`, which just so happens to stop working when the optimizer isn't pickleable (like the Horovod optimizer).

An improvement to the above solution on our end might be to delete/warn on implicit Horovod optimizers only, not on any bound method.  Upstreaming a PyTorch fix seems sensible too though -- seems like unintended behavior to me.

rb-determined-ai:
>  Upstreaming a PyTorch fix seems sensible too though -- seems like unintended behavior to me.
We talked, this is what we settled on.  We can come back if pytorch rejects the fix.

PR #5044, opened 2022-09-16
feat: enable configuring the agent to shut down on connection failure [DET-7022]

dzhu:
## Test Plan

- [x] run cluster configured to make agent fail to connect, see that it runs the configured command and provides output if that fails fails

## Commentary (optional)

In the latest commit, the command is configured through a new  "hooks" section in the agent config; in the intermediate one, the config is more specifically tailored to this scenario of shutting down on connection failure. The former both is more extensible and allows for nicer names in the code; I don't think I see any downside except that it may cause users to expect us to have more hooks available in the near future.

Of course, we may not want to allow the agent to run arbitrary commands on the host at all. But I feel like we need some amount of flexibility, and I couldn't think of a way I liked to have some without doing this. (Maybe just specifying arguments to be passed to `shutdown`? But maybe there's a reason to run a command other than `shutdown`. I don't really know.)

### DIFF
--- a/docs/reference/reference-deploy/config/agent-config-reference.rst
+++ b/docs/reference/reference-deploy/config/agent-config-reference.rst
@@ -93,3 +93,16 @@
 
 -  ``container_auto_remove_disabled`` (debug): Whether to disable setting ``AutoRemove`` flag on
    task containers. Defaults to false.
+
+-  ``hooks``: Configuration for commands to run when certain events occur. The value of each option
+   in this section is an array of strings specifying the command and its arguments.
+
+   -  ``on_connection_lost``: A command to run when the agent fails to either connect to the master
+      on startup or reconnect after a loss of connection. (When reconnecting, the agent will make
+      several attempts as specified by the ``agent_reconnect_attempts`` and
+      ``agent_reconnect_backoff`` configuration options.)
+
+      In order to shut down the machine on which the agent is running, set this to ``["sudo",
+      "shutdown", "now"]``, or just ``["shutdown", "now"]`` if the agent is running as root.
+      Additional system configuration may be required in order to allow the agent to execute the
+      command from inside a Docker container or without the need to enter a password.
--- a/e2e_tests/tests/cluster/test_agent.py
+++ b/e2e_tests/tests/cluster/test_agent.py
@@ -1,10 +1,13 @@
 import os
+import time
 
 import pytest
 
 import determined
 from determined.common import api
 from tests import config as conf
+
+from .managed_cluster import ManagedCluster
 
 
 # TODO: This should be marked as a cross-version test, but it can't actually be at the time of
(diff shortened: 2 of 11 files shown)


ioga:
1. should this be a feature enabled by default in `det deploy aws|gcp`? 
2. should we add docs for the "recommended configuration"? i.e. shutdown.
3. after it's merged we should ask SaaS to configure this.

otherwise this PR has no usage, no tests, no examples.

dzhu:
Adding to `det deploy` sounds good. Docs I figured I would do after confirming which version to go with. For tests, I'll see what I can do with the devcluster tests, I guess.

master/static/srv/agent_setup_script.sh.template
```diff
@@ -40,6 +45,7 @@ cat /usr/local/determined/container_startup_script
 echo "#### PRINTING CONTAINER STARTUP SCRIPT END ####"
 
 docker run --init --name determined-agent {{.LogOptions}} \
+    --privileged \
```
neilconway, line 48: Would it be possible to grant more fine-grained capabilities to the container? Looking briefly it seems like `--cap-add SYS_BOOT` might work?

docs/reference/reference-deploy/config/agent-config-reference.rst
```diff
@@ -93,3 +93,13 @@
 
 -  ``container_auto_remove_disabled`` (debug): Whether to disable setting ``AutoRemove`` flag on
    task containers. Defaults to false.
+
+-  ``hooks``: Configuration for commands to run when certain events occur. The value of each option
+   in this section is an array of strings specifying the command and its arguments.
+
+   -  ``on_connection_failure``: A command to run when the agent fails to either connect to the
```
neilconway, line 100: Can we document the timeout behavior here? i.e., as implemented the agent will try to reconnect multiple times before giving up and invoking this hook.

I wonder if the name for this behavior is ideal: `on_connection_failure` suggests to me that the hook is invoked whenever a connection attempt fails, rather than only once at the end of a sequence of connection attempts + backoffs. Not sure if I can think of a better name offhand, though...

neilconway (commented):
Can we also document the default behavior of `det deploy`'d clusters now? e.g., in the docs for AWS/GCP, say something about the fact that agents will auto-terminate if the master isn't reachable for X amount of time.

dzhu (commented):
I'm going to split out `det deploy` stuff into a separate PR; SaaS doesn't need it, it's much larger, and GCP testing is being annoying.

docs/reference/reference-deploy/config/agent-config-reference.rst
neilconway: `on_connection_lost` sgtm!

master/static/srv/agent_setup_script.sh.template
neilconway: Hmm, weird. I do think it is unfortunate from a security POV (easier to escalate privileges from agent container / attack host OS), but probably not a huge deal.

dzhu:
Alright, I think the core functionality is good to go, with `det deploy` changes slated to go in separately.

dzhu:
Tests are passing, and I think this should be ready to look at.

@stoksc had a thought that this should perhaps be handled by a different layer specific to the install method (e.g., for a `det deploy` installation, a script that runs outside the container and watches for a specific exit code from the container). That sounds sensible, but I think it would add a lot of moving parts and we're okay with going ahead with this method.

stoksc:
For posterity, the idea was to have the agent return a specific error code for connection failures and the infra layer that handles deploys could also handle cleanups (e.g., a script that tails Docker events or something with systemd `OnFailure` that I do not know). That way you avoid the (only sort of) awkward callback across a permission boundary.

e2e_tests/tests/cluster/test_agent.py
```diff
@@ -17,3 +20,29 @@ def test_agent_version() -> None:
...
     assert all(agent["version"] == target_version for agent in agents.values())
+
+
+@pytest.mark.e2e_cpu_agent_connection_loss
+def test_agent_never_connect() -> None:
+    for _ in range(15):
+        if os.path.exists("/tmp/agent-connection-lost"):
+            return
+        time.sleep(1)
+    raise AssertionError("Did not find expected file from agent connection loss hook")
```
ioga, line 31:
```suggestion
    pytest.fail("Did not find expected file from agent connection loss hook")
```

PR #5055, opened 2022-09-16
fix: support setting agent username and group in user APIs.

ioga:
## Description

grpc-style user APIs were missing support for username / groupname fields in `AgentUserGroups`, i.e. `link-with-agent-user` feature. This is not an externally facing fix: `AgentUserGroups` isn't currently in use by WebUI, CLI still uses the old style.

Under the hood, user name and group name are NOT NULL fields in the table. Prior to this fix, they'd be reset to empty strings due to protobuf / golang empty struct initializations.

Technically, uid and gid are critical for the system, while user name and group name are cosmetic, however if we enforce them in one place (db and `link-with-agent-user`) it probably makes sense to enforce them everywhere.

## Test Plan

Expanded automated testing.

## Checklist

### DIFF
--- a/harness/determined/common/api/bindings.py
+++ b/harness/determined/common/api/bindings.py
@@ -478,22 +478,30 @@ class v1AgentUserGroup:
         self,
         *,
         agentGid: "typing.Optional[int]" = None,
+        agentGroup: "typing.Optional[str]" = None,
         agentUid: "typing.Optional[int]" = None,
+        agentUser: "typing.Optional[str]" = None,
     ):
         self.agentUid = agentUid
         self.agentGid = agentGid
+        self.agentUser = agentUser
+        self.agentGroup = agentGroup
 
     @classmethod
     def from_json(cls, obj: Json) -> "v1AgentUserGroup":
         return cls(
             agentUid=obj.get("agentUid", None),
             agentGid=obj.get("agentGid", None),
+            agentUser=obj.get("agentUser", None),
+            agentGroup=obj.get("agentGroup", None),
         )
 
     def to_json(self) -> typing.Any:
         return {
             "agentUid": self.agentUid if self.agentUid is not None else None,
             "agentGid": self.agentGid if self.agentGid is not None else None,
+            "agentUser": self.agentUser if self.agentUser is not None else None,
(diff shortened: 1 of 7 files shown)


master/internal/api_user.go
```diff
@@ -334,8 +344,10 @@ func (a *apiServer) PatchUser(
 	var ug *model.AgentUserGroup
 	if pug := req.User.AgentUserGroup; pug != nil {
 		ug = &model.AgentUserGroup{
-			UID: int(req.User.AgentUserGroup.AgentUid),
-			GID: int(req.User.AgentUserGroup.AgentGid),
+			UID:   int(req.User.AgentUserGroup.AgentUid),
+			GID:   int(req.User.AgentUserGroup.AgentGid),
+			User:  req.User.AgentUserGroup.AgentUser,
+			Group: req.User.AgentUserGroup.AgentGroup,
```
mapmeld, line 350: is there any restriction  or can a user make an agent-user or agent-group association for anyone?

harness/determined/common/api/bindings.py
```diff
@@ -478,22 +478,30 @@ def __init__(
         self,
         *,
         agentGid: "typing.Optional[int]" = None,
+        agentGroup: "typing.Optional[str]" = None,
```
NicholasBlaskey, line 481: should we switch the CLI set agent user group over with this change? 

Waiting to do it at all once or bit by bit as we make sure functionality is matched both seem fine to me but curious on your thoughts.

NicholasBlaskey (approved):
lgtm

master/internal/api_user.go
mapmeld: :+1:

harness/determined/common/api/bindings.py
ioga: this is a judgement call: on one hand, you are right in advocating this. on the other, I am working from a mindset of a critical path towards storage solution, and we basically have this work  outlined and ticketed as one of our four top tech debt items.

mapmeld:
This is last-minute but could it be a userId instead of agentUsername? We were just discussing features to make changing username easier and this would affect username here

PR #5056, opened 2022-09-19
feat: Table component performance enhancement

thiagodallacqua-hpe:
## Description

Original ticket: DET-8110

## Test Plan

Go to `/det/projects/` and "play" with the table 😁 

## Commentary (optional)

There's a lot to uncover from the time that I've spent with this component...

I have chosen to take the path of understanding the component and why it's such a bottleneck for performance.

What I have found (in the react code) is that we have a very complex, highly customized table component with a lot of logic for resizing, re-organizing and filtering... all of which are tied to a global state (settings) which we know already that is kind of low on performance and has a few issues that we're trying to fix, plus we're actually getting to discuss replacing the architecture of it entirely (check @keita-determined's proposal).

The first issue found is that said global state has an auto-refresh which makes the whole tree re-renders, even if you don't do anything...

We're using `Context` to "avoid" prop drilling, but, we're using on a direct child, so I've removed it 😅.

There are a few listeners that keep changing the state, forcing re-renders, but, the state hasn't actually changed since the last update.

Resizing columns adds up to all of the above, constantly and indefinitely changing the state of the entire component tree.

very little code-reusability, mainly due to the complex nature of what we're trying to achieve.

I think that "CSS-ing" the resizing won't bring much to the table due to the fact that it relies on the event values to apply the new `width` to the `colgroup`, which, in turn reflects the `width` fo each `Cell` or `HeaderCell`, meaning that it is heavily dependant on the antd `Table`. It also uses table elements (although only to a certain point), which prevents the usage of `grid` or `flex-box`. (but, then again, it wouldn't help much anyway).

In conclusion, we can either manage the current state or shift to a new `Table` component 😅. 

For more insight, check PR's comments.

There's also some profiling json files attached to the ticket

Rendering sample before:

https://user-images.githubusercontent.com/104855841/191054936-103379e3-796a-4725-affc-03f4156c3053.mp4

after:

https://user-images.githubusercontent.com/104855841/191054996-7ee3d0c7-12a7-49cb-a039-48700f58f02c.mp4

resizing before:

https://user-images.githubusercontent.com/104855841/191079049-f6d1ebe8-2dd7-4704-ab37-9e72f470e6ac.mp4

resizing after:

https://user-images.githubusercontent.com/104855841/191079089-772b4af5-3f4b-41e4-add5-2e0856952619.mp4

## Checklist

### DIFF
--- a/webui/react/src/hooks/useResize.ts
+++ b/webui/react/src/hooks/useResize.ts
@@ -1,4 +1,4 @@
-import { RefObject, useEffect, useRef, useState } from 'react';
+import { RefObject, useEffect, useState } from 'react';
 
 interface ResizeInfo {
   height: number;
@@ -18,13 +18,6 @@ export const DEFAULT_RESIZE_THROTTLE_TIME = 500;
 
 const useResize = (ref?: RefObject<HTMLElement>): ResizeInfo => {
   const [resizeInfo, setResizeInfo] = useState<ResizeInfo>(defaultResizeInfo);
-  const timeout = useRef<NodeJS.Timeout>();
-
-  useEffect(() => {
-    return () => {
-      if (timeout.current) clearTimeout(timeout.current);
-    };
-  }, []);
 
   useEffect(() => {
     let element = document.body;
@@ -39,23 +32,14 @@ const useResize = (ref?: RefObject<HTMLElement>): ResizeInfo => {
       if (!element || elements.indexOf(element) === -1) return;
 
       const rect = element.getBoundingClientRect();
-
-      if (timeout.current) clearTimeout(timeout.current);
-
-      timeout.current = setTimeout(() => {
-        setResizeInfo(rect);
-      }, DEFAULT_RESIZE_THROTTLE_TIME);
+      setResizeInfo(rect);
     };
     const resizeObserver = new ResizeObserver(handleResize);
     resizeObserver.observe(element);
 
     // Set initial resize info
     const rect = element.getBoundingClientRect();
-    if (timeout.current) clearTimeout(timeout.current);
-
-    timeout.current = setTimeout(() => {
-      setResizeInfo(rect);
-    }, DEFAULT_RESIZE_THROTTLE_TIME);
+    setResizeInfo(rect);
 
     return (): void => resizeObserver.unobserve(element);
   }, [ref]);


keita-determined (dismissed):
ColumnWidth of userSettings is not persistent.
Question: Did you also look into other table options?

https://user-images.githubusercontent.com/109113616/191137988-7e9cb354-eb14-482d-8379-07fb30e7293a.mov

webui/react/src/shared/configs/eslintrc.js
```diff
@@ -98,13 +98,7 @@ module.exports = {
     'quote-props': ['error', 'consistent-as-needed'],
     'quotes': ['error', 'single', { avoidEscape: true }],
     'react/display-name': 'off',
-    'react/jsx-closing-bracket-location': [
```
keita-determined, line 101: could you fix this please?

trentwatt (dismissed):
feels kinda buggy imo

https://user-images.githubusercontent.com/28383080/191338226-0cd61086-c8e3-43cb-b7f9-c7ee1033b6ce.mov

keita-determined:
https://user-images.githubusercontent.com/109113616/191384095-1db6f754-10ce-4454-9ef8-e4fe3c73d46b.mov

Other trivial issue. 
Resizing bar doesn't stop when the bar exceed the minimum column size.

trentwatt (commented):
> Resizing columns adds up to all of the above, constantly and indefinitely changing the state of the entire component tree.

the resizing causing a re-render is a newer development, probably caused by changes outside of the table component? I would rather we identify and fix that issue as opposed to adding another layer of things we would need to fix later. 

I will defer to alex on this one though.

trentwatt:
> also, @trentwatt @keita-determined, as mentioned by @ClaireNeveu, although the resizing doesn't quite look perfect yet, it is way more performant than before, so, unless we really wanna brush it out here, or we disagree with the current code around it, can we merge this PR and fix the behavior in a new one?

I reached out to alex about it, I will ping him again.

webui/react/src/hooks/useSettings.ts
```diff
@@ -365,15 +369,20 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
       const isDefault = isEqual(config.defaultValue, jsonValue);
 
       // Store or clear setting if `storageKey` is available.
-      if (config.storageKey && isValid && options?.store !== false) {
```
trentwatt, line 368: keep

ClaireNeveu:
If this is just waiting on design approval I'm gonna go ahead and approve it on behalf of design. Once the new designer is onboarded we'll revisit the table UX and focus on a more final solution.

PR #5057, opened 2022-09-19
bugfix: replace pip with python -m pip

drh-determined-ai:
## Description

bugfix: replace pip with python -m pip

## Test Plan

CircleCI tests should succeed.

## Checklist


rb-determined-ai (approved):
If you are not sure what precisely causes only this pip call to fail (and only fail in the windows runner) I'd recommend a comment explaining which conditions require `python -m pip` "for reasons which are unclear" or something to that effect.

Otherwise it looks real weird to the next developer who goes and reads this code carefully.

I wouldn't probably waste more than 20 minutes trying to figure out why fixing only this line fixes the test, if you decide to go that route.

dannysauer:
In the future, maybe try to split unrelated cosmetic fixes out from technical fixes; that mixing stuff up makes tracking down "what changed" harder.

PR #5059, opened 2022-09-19
chore: add lint for Prettier and Ignore format code commit

keita-determined:
## Description
- Related to #5015 
- This is Step3 of https://github.com/determined-ai/determined/pull/5015#issuecomment-1248376734
- [DET-8097]

## Test Plan
- Check if code works as expected
- Check if all test cases pass
- Check if the right commit is ignored

## Checklist

[DET-8097]: https://determinedai.atlassian.net/browse/DET-8097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hamidzr (changes requested):
please make sure
`make -C webui/react check` 
and
`make -C webui/react fmt` 
execute what you expect them to execute.

from what I understood from our chat we want the target `check` to run `eslint`, `stylelint`, and `prettier` and they don't
conflict and check for more than just what prettier checks for.

make `fmt` could either use the same treatment or just run prettier.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# .git-blame-ignore-revs
```
hamidzr, line 1: we could probably use the same file under `shared-web` as well

keita-determined: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
The file must be on root dir

hamidzr: that'll translate to the root directory for https://github.com/determined-ai/shared-web

hamidzr:
as is the ci isn't consulting prettier when checking for code style https://app.circleci.com/pipelines/github/determined-ai/determined/27238/workflows/f22db873-abb5-48bc-ae19-816dcf9b162d/jobs/909331?invite=true#step-111-7

webui/react/src/shared/.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# .git-blame-ignore-revs
+# Format code with Prettier in web
+85396ebc4f0ec95c6b4ef9625061c92fa324cdce
```
hamidzr, line 3: I think this will probably change (I'm not sure) after we push hence why I suggested we leave it for later

eg
  commit 62756f5d5a20c4ec0ca4b0339687c1085b4852f6
vs b63e05d28bd51703b1eabd1c55adeea9bc4b3d59

hamidzr (approved):
LGTM. Thanks @keita-determined 


I have some other thoughts on how we can clean up and sync package.json and makefile around formatting and linting but those are second priorities I'll propose something once we got basics wrapped up with this pr.

PR #5063, opened 2022-09-19
fix: Trial.TotalCheckpointSize considers deleted checkpoints [DET-8399]

varlogtim:
## Description

Fix issue where Total Checkpoint Size was showing the size of all checkpoints, even deleted ones.

## To Test:
1. Create an single trial experiment which creates multiple trials (set min_checkpoint_period to something small).
2. Wait for trial to complete and GC task to run,.
3. Go to `http://host:8080/det/experiments/<exp_id>/overview`
4. Observe that the "Total Checkpoint Size" matches the total for non-deleted checkpoints in: `http://host:8080/det/experiments/<exp_id>/checkpoints`


ioga (approved):
lgtm

PR #5065, opened 2022-09-19
build: speed up parallel runs for react make

hamidzr:
## Description
`make check -j` 16s to 10s

removes/breaks `npm run lint:*` commands

## Checklist


hamidzr:
3948317614ef9a349cc2ff528fe666be3be8a4ed doesn't really contribute to the overall speed today (at least on my machine) since eslint is the slowest of the bunch

webui/react/package.json
```diff
@@ -11,9 +11,8 @@
     "postshrinkwrap": "npx npm-force-resolutions && npm install --ignore-scripts",
     "start": "craco start",
     "build": "craco build",
-    "lint": "npm run lint:js && npm run lint:css && npm run format-check",
```
keita-determined, line 14: I think its nice to have `lint` command to run all cuz there's a [ticket](https://determinedai.atlassian.net/browse/DET-8093) to introduce lint-staged.
and we don't need to run multiple npm commands on local

keita-determined (approved):
LGTM
just minor comment

webui/react/package.json
hamidzr: I'd suggest we lean into using make instead of package.json it's more flexible and not unique to the webui module.
`make check` should do what `lint` did with more options

hamidzr: lint also was confusing here to me as we had both that and the new set of `format* scripts? defined.

keita-determined: I thought adding `lint` is good cuz I use `npm` instead of `make` (its just because I just used to use `npm`)
seems like `lint-staged`can use commands outside of `package.json` so I think we don't need to change it for now.

hamidzr: fair point. I added it back

PR #5066, opened 2022-09-19
fix: use theme var in Avatar stylesheet

johnkim-det:
## Description

addressing lack of contrast between text/background for avatars in light mode.

## Test Plan

nothing to test in core since `noColor` prop is only used in SaaS.

## Commentary (optional)

can test in storybook and SaaS to see that in light mode with `noColor` mode, text is dark instead of white:

avatar-nocolor-var
avatar-nocolor-white

## Checklist


webui/react/src/shared/components/Avatar/Avatar.stories.tsx
```diff
@@ -7,11 +7,11 @@ import Avatar from './Avatar';
 
 export default {
   component: Avatar,
-  decorators: [withKnobs],
+  decorators: [ withKnobs ],
```
hamidzr, line 10: https://determined-ai.slack.com/archives/CKQ3EV0G6/p1663706989991159?thread_ts=1663625872.693039&cid=CKQ3EV0G6

hamidzr (approved):
LGTM. it'd be geat if you could get a higher contrast for the background color as well
` backgroundColor: noColor ? 'var(--theme-stage-strong)' : getColor(displayName, darkLight),
` wdy think?

PR #5068, opened 2022-09-20
feat: WebUI edit permission at user profile [DET-8224]

gt2345:
## Description

Allow user with relevant permission to add/delete roles at user profile modal.

Screen Shot 2022-09-19 at 8 30 59 PM

Screen Shot 2022-09-19 at 8 31 10 PM

## Test Plan

Navigate to `/det/settings/user-management?f_rbac=on`, open create/edit user modal to see the role select.
Note that actually functionality off adding/removing roles is not implemented. 

## Checklist


webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -41,30 +45,13 @@ interface FormValues {
...
-    title: 'Workspaces?',
-  },
-];
-
 const ModalForm: React.FC<Props> = ({ form, branding, user, groups, viewOnly }) => {
-  const [permissions, setPermissions] = useState<Permission[]>([]);
+  const [ permissions, setPermissions ] = useState<Permission[]>([]);
 
+  const rbacEnabled = useFeature().isOn('rbac');
   const canSeePermissions = usePermissions().canGetPermissions;
```
ClaireNeveu, line 52: You should call `usePermissions` just once. Then you can do `if (user && permissions.canSeePermissions)` etc.

webui/react/src/pages/Settings/Settings.tsx
```diff
@@ -1,13 +1,15 @@
 import { Tabs } from 'antd';
```
ClaireNeveu, line 1: What's the story behind this change?

ClaireNeveu (changes requested):

ClaireNeveu:
I think if rbac is on we should be hiding the "admin" toggle

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
gt2345: Makes sense, thank you

webui/react/src/pages/Settings/Settings.tsx
gt2345: This page used to not sync tab information to url, I changed it to align with the other places in code base, should we have a separate PR for it?

ClaireNeveu: It's fine since the change is small but you should call that out when you make the PR, preferably by commenting on the relevant file in the PR discussion.

ClaireNeveu (approved):

webui/react/src/pages/Settings/Settings.tsx
gt2345: Got it, thanks!

PR #5071, opened 2022-09-20
chore: remove check from pytorch [MLG-182]

drh-determined-ai:
## Description

chore: remove check from pytorch [MLG-182]

The module `determined.common.check` is do be replaced with case-by-case raising of exceptions (or assertions, in situations where being optimized out will not cause problems in production).

## Test Plan

Unit/e2e tests as usual.

## Checklist

[MLG-182]: https://determinedai.atlassian.net/browse/MLG-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/harness/determined/pytorch/_lr_scheduler.py
+++ b/harness/determined/pytorch/_lr_scheduler.py
@@ -49,7 +49,7 @@ class LRScheduler:
                 2. ``STEP_EVERY_BATCH``: Determined will call scheduler.step() after every
                    ``frequency`` training batch(es). No arguments will be passed to step().
                    This option does not take into account gradient aggregation;
-                   ``STEP_EVERY_OPTIMIZER_STEP`` which does is recommended.
+                   ``STEP_EVERY_OPTIMIZER_STEP`` which is recommended.
 
                 3. ``STEP_EVERY_OPTIMIZER_STEP``: Determined will call scheduler.step() in sync
                    with optimizer steps. With ``optimizations.aggregation_frequency`` unset, this
--- a/harness/determined/pytorch/_lr_scheduler.py
+++ b/harness/determined/pytorch/_lr_scheduler.py
@@ -3,8 +3,6 @@ from typing import Any, Dict, List
 
 import torch
 
-import determined.common.check as check
-
 
 class LRScheduler:
     """Wrapper for a PyTorch LRScheduler.
@@ -67,8 +65,10 @@ class LRScheduler:
             frequency:
                 Sets the frequency at which the batch and epoch step modes get triggered.
         """
-        check.check_not_none(scheduler)
-        check.check_isinstance(step_mode, LRScheduler.StepMode)
+        if scheduler is None:
+            raise ValueError("scheduler must not be None.")
+        if not isinstance(step_mode, LRScheduler.StepMode):
+            raise TypeError(f"step_mode must be an LRScheduler.StepMode. Got {type(step_mode)}.")
 
         self._scheduler = scheduler
         self._step_mode = step_mode
--- a/harness/determined/pytorch/_lr_scheduler.py
+++ b/harness/determined/pytorch/_lr_scheduler.py
@@ -48,7 +48,8 @@ class LRScheduler:
 
                 2. ``STEP_EVERY_BATCH``: Determined will call scheduler.step() after every
                    ``frequency`` training batch(es). No arguments will be passed to step().
-                   This option does not take into account gradient aggregation.
+                   This option does not take into account gradient aggregation;
+                   ``STEP_EVERY_OPTIMIZER_STEP`` which is recommended.
 
                 3. ``STEP_EVERY_OPTIMIZER_STEP``: Determined will call scheduler.step() in sync
                    with optimizer steps. With ``optimizations.aggregation_frequency`` unset, this
--- a/harness/determined/pytorch/_lr_scheduler.py
+++ b/harness/determined/pytorch/_lr_scheduler.py
@@ -65,8 +65,6 @@ class LRScheduler:
             frequency:
                 Sets the frequency at which the batch and epoch step modes get triggered.
         """
-        if scheduler is None:
-            raise ValueError("scheduler must not be None.")
         if not isinstance(step_mode, LRScheduler.StepMode):
             raise TypeError(f"step_mode must be an LRScheduler.StepMode. Got {type(step_mode)}.")
 
--- a/harness/determined/pytorch/_lr_scheduler.py
+++ b/harness/determined/pytorch/_lr_scheduler.py
@@ -48,8 +48,7 @@ class LRScheduler:
 
                 2. ``STEP_EVERY_BATCH``: Determined will call scheduler.step() after every
                    ``frequency`` training batch(es). No arguments will be passed to step().
-                   This option does not take into account gradient aggregation;
-                   ``STEP_EVERY_OPTIMIZER_STEP`` which is recommended.
+                   This option does not take into account gradient aggregation.
 
                 3. ``STEP_EVERY_OPTIMIZER_STEP``: Determined will call scheduler.step() in sync
                    with optimizer steps. With ``optimizations.aggregation_frequency`` unset, this
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -341,8 +341,6 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             device = torch.device("cuda", 0)
         else:
             device = torch.device("cpu")
-        if device is None:
-            raise Exception("Error initializing torch.device")
         return device
 
     def to_device(self, data: pytorch._Data) -> pytorch.TorchData:
@@ -656,8 +654,6 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
     @staticmethod
     def _average_gradients(parameters: Any, divisor: int) -> None:
-        if divisor < 1:
-            raise ValueError("divisor < 1")
         if divisor == 1:
             return
 
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -233,7 +233,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
                 )
             if backward_passes_per_step < 1:
                 raise det.errors.InvalidExperimentException(
-                    "backward_passes_per_step for local gradient aggregation must be >= 1.",
+                    "backward_passes_per_step for local gradient aggregation must be >= 1; "
+                    f"got {backward_passes_per_step}.",
                 )
 
             if self.distributed.size > 1 and self._distributed_backend.use_horovod():
@@ -372,7 +373,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
         if not HAVE_AMP:
             raise det.errors.InvalidExperimentException(
-                "Failed to import torch.cuda.amp. PyTorch >= 1.6 required.",
+                "Using context.wrap_scaler() requires PyTorch >= 1.6.",
             )
 
         if self._use_apex:
@@ -480,19 +481,19 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
         if self.distributed.size > 1:
             if num_losses != 1:
                 raise det.errors.InvalidExperimentException(
-                    "When using parallel/distributed training, "
+                    "When using distributed training, "
                     "Determined only supports configure_apex_amp with num_losses = 1.",
                 )
             if self._aggregation_frequency > 1:
                 raise det.errors.InvalidExperimentException(
-                    "Mixed precision training (AMP) is not supported with "
+                    "context.configure_apex_amp is not supported with "
                     "distributed training and "
                     "aggregation frequency > 1.",
                 )
 
         if not torch.cuda.is_available():
             raise det.errors.InvalidExperimentException(
-                "Mixed precision training (AMP) is supported only on GPU slots.",
+                "context.configure_apex_amp is supported only on GPU slots.",
             )
 
         self._use_apex = True
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -655,8 +655,6 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
     @staticmethod
     def _average_gradients(parameters: Any, divisor: int) -> None:
-        if divisor < 1:
-            raise ValueError(f"Bad divisor in _average_gradients: {divisor} < 1.")
         if divisor == 1:
             return
 
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -671,7 +671,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
         clip_grads: Optional[Callable[[Iterator], None]] = None,
         auto_zero_grads: bool = True,
         scaler: Optional[Any] = None,
-        # Should be torch.cuda.amp.GradScaler, but:
+        # Should be ``torch.cuda.amp.GradScaler``, but:
         #   * other implementations might be possible
         #   * requiring this type forces upgrades to PyTorch 1.6+
     ) -> None:
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -655,6 +655,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
     @staticmethod
     def _average_gradients(parameters: Any, divisor: int) -> None:
+        if divisor < 1:
+            raise ValueError(f"Bad divisor in _average_gradients: {divisor} < 1.")
         if divisor == 1:
             return
 
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -389,10 +389,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
                 "Please call wrap_scaler before wrap_model.",
             )
 
-        if not torch.cuda.is_available():
-            raise det.errors.InvalidExperimentException(
-                "Using context.wrap_scaler() is supported only on GPU slots.",
-            )
+        # We don't need to check if CUDA is available because if it is not, a GradScaler is
+        #  disabled when initialized, and we allow for disabled scalers to exist.
 
         self._scaler = scaler
 
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -418,7 +418,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
     ) -> Tuple:
         """
         Configure automatic mixed precision for your models and optimizers using NVIDIA's Apex
-        PyTorch extension. Note that details for apex.amp are handled automatically within
+        PyTorch extension. Note that details for ``apex.amp`` are handled automatically within
         Determined after this call.
 
         This function must be called **after** you have finished constructing your models and
@@ -571,7 +571,8 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
             That means the gradient on each parameter can only be calculated once on each batch.
             If a parameter is associated with multiple losses, you can either choose to call
             ``backward'' on only one of those losses, or you can set the ``require_grads`` flag of
-            a parameter or module to false to avoid manual gradient accumulation on that parameter.
+            a parameter or module to ``False`` to avoid manual gradient accumulation on that
+            parameter.
             However, you can do gradient accumulation across batches by setting
             :ref:`optimizations.aggregation_frequency <config-aggregation-frequency>` in the
             experiment configuration to be greater than 1.
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -391,7 +391,7 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
         if not torch.cuda.is_available():
             raise det.errors.InvalidExperimentException(
-                "Mixed precision training (AMP) is supported only on GPU slots.",
+                "Using context.wrap_scaler() is supported only on GPU slots.",
             )
 
         self._scaler = scaler
--- a/harness/determined/pytorch/_pytorch_context.py
+++ b/harness/determined/pytorch/_pytorch_context.py
@@ -7,7 +7,6 @@ import torch.nn as nn
 
 import determined as det
 from determined import profiler, pytorch, util
-from determined.common import check
 from determined.horovod import hvd
 from determined.tensorboard import get_base_path
 
@@ -171,7 +170,10 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
         """Returns a wrapped model."""
 
         if self.env.managed_training:
-            check.false(self._use_apex, "Must call wrap_model() before configure_apex_amp.")
+            if self._use_apex:
+                raise det.errors.InvalidExperimentException(
+                    "Must call wrap_model() before configure_apex_amp.",
+                )
 
             model = model.to(self.device)
 
@@ -225,12 +227,14 @@ class PyTorchTrialContext(det.TrialContext, pytorch._PyTorchReducerContext):
 
         """
         if self.env.managed_training:
-            check.false(self._use_apex, "Must call wrap_optimizer() before configure_apex_amp.")
-            check.gt_eq(
-                backward_passes_per_step,
-                1,
-                "backward_passes_per_step for local gradient aggregation must be >= 1",
-            )
+            if self._use_apex:
+                raise det.errors.InvalidExperimentException(
+                    "Must call wrap_optimizer() before configure_apex_amp.",
+                )
+            if backward_passes_per_step < 1:
+                raise det.errors.InvalidExperimentException(
+                    "backward_passes_per_step for local gradient aggregation must be >= 1.",
+                )
 
             if self.distributed.size > 1 and self._distributed_backend.use_horovod():
                 optimizer = hvd.DistributedOptimizer(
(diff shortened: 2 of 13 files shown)


harness/determined/pytorch/_lr_scheduler.py
```diff
@@ -67,8 +65,10 @@ def __init__(
             frequency:
                 Sets the frequency at which the batch and epoch step modes get triggered.
         """
-        check.check_not_none(scheduler)
-        check.check_isinstance(step_mode, LRScheduler.StepMode)
+        if scheduler is None:
+            raise ValueError("scheduler must not be None.")
```
rb-determined-ai, line 69: drop this check.  There's nothing about this function that would suggest that users could pass None as their scheduler.

Also, if they want to pass bone-headed values in, there's a million things they could pass that are not None but which would also cause problems, so this check is like a really incomplete blacklist of bad values.

harness/determined/pytorch/_lr_scheduler.py
```diff
@@ -67,8 +65,10 @@ def __init__(
...
+        if not isinstance(step_mode, LRScheduler.StepMode):
+            raise TypeError(f"step_mode must be an LRScheduler.StepMode. Got {type(step_mode)}.")
```
rb-determined-ai, line 71: I would consider it more idiomatic to do `step_mode = LRScheduler.StepMode(step_mode)` which will promote specific values (like `"MANUAL_STEP"` automatically).

But I'm not opposed to leaving this in place because it is a little stricter and nobody has ever complained about it.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -225,12 +227,14 @@ def train_batch(
...
-                1,
-                "backward_passes_per_step for local gradient aggregation must be >= 1",
-            )
+            if self._use_apex:
+                raise det.errors.InvalidExperimentException(
+                    "Must call wrap_optimizer() before configure_apex_amp.",
+                )
+            if backward_passes_per_step < 1:
+                raise det.errors.InvalidExperimentException(
+                    "backward_passes_per_step for local gradient aggregation must be >= 1.",
```
rb-determined-ai, line 236: nit:
```python
"backward_passes_per_step for local gradient aggregation must be >= 1; "
f"got {backward_passes_per_step}",
```

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -338,7 +341,8 @@ def _init_device(self) -> torch.device:
             device = torch.device("cuda", 0)
         else:
             device = torch.device("cpu")
-        check.is_not_none(device)
+        if device is None:
+            raise Exception("Error initializing torch.device")
```
rb-determined-ai, line 345: This check seems unnecessary to me.  Afaict, there's no branch where it might be None.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -368,18 +372,28 @@ def wrap_scaler(self, scaler: Any) -> Any:
             The scaler. It may be wrapped to add additional functionality for use in Determined.
         """
 
-        check.true(HAVE_AMP, "Failed to import torch.cuda.amp. PyTorch >= 1.6 required.")
+        if not HAVE_AMP:
+            raise det.errors.InvalidExperimentException(
+                "Failed to import torch.cuda.amp. PyTorch >= 1.6 required.",
```
rb-determined-ai, line 377: Maybe:
```python
"using context.wrap_scaler() requires PyTorch >= 1.6"
```
?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -368,18 +372,28 @@ def wrap_scaler(self, scaler: Any) -> Any:
...
+            )
 
-        check.true(
-            torch.cuda.is_available(),
-            "Mixed precision training (AMP) is supported only on GPU slots.",
-        )
+        if not torch.cuda.is_available():
+            raise det.errors.InvalidExperimentException(
+                "Mixed precision training (AMP) is supported only on GPU slots.",
+            )
```
rb-determined-ai, line 396: Again, mention that it's the call to `wrap_scaler` that triggered this.

Also, this check should only be triggered if the scaler is enabled, right?

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -459,33 +473,31 @@ def configure_apex_amp(
...
+            if num_losses != 1:
+                raise det.errors.InvalidExperimentException(
+                    "When using parallel/distributed training, "
+                    "Determined only supports configure_apex_amp with num_losses = 1.",
+                )
+            if self._aggregation_frequency > 1:
+                raise det.errors.InvalidExperimentException(
+                    "Mixed precision training (AMP) is not supported with "
+                    "distributed training and "
+                    "aggregation frequency > 1.",
```
rb-determined-ai, line 492: please specify this is only true wit apex amp.

I think this message text predates the existence of pytorch-native amp.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -459,33 +473,31 @@ def configure_apex_amp(
...
         if self.distributed.size > 1:
-            check.eq(
-                self._aggregation_frequency,
-                1,
-                "Mixed precision training (AMP) is not supported with "
-                "distributed training and "
-                "aggregation frequency > 1.",
+            if num_losses != 1:
+                raise det.errors.InvalidExperimentException(
+                    "When using parallel/distributed training, "
```
rb-determined-ai, line 485: Please change this to `When using distributed training` ("parallel" training was a term we used to use to describe multi-slot, single-node training, but we dropped it a longtime ago)

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -459,33 +473,31 @@ def configure_apex_amp(
...
+            if self._aggregation_frequency > 1:
+                raise det.errors.InvalidExperimentException(
+                    "Mixed precision training (AMP) is not supported with "
+                    "distributed training and "
+                    "aggregation frequency > 1.",
+                )
+
+        if not torch.cuda.is_available():
+            raise det.errors.InvalidExperimentException(
+                "Mixed precision training (AMP) is supported only on GPU slots.",
```
rb-determined-ai, line 497: as before, please modify the text to include that the `configure_apex_amp` is what triggered the error.

harness/determined/pytorch/_pytorch_context.py
```diff
@@ -643,7 +656,8 @@ def backward(
 
     @staticmethod
     def _average_gradients(parameters: Any, divisor: int) -> None:
-        check.gt_eq(divisor, 1)
+        if divisor < 1:
+            raise ValueError("divisor < 1")
```
rb-determined-ai, line 660: I'd be fine if you removed this check... seems unnecessary to me.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -380,7 +377,8 @@ def _train_for_step(
         self, step_id: int, num_batches: int, total_batches_processed: int
     ) -> workload.Metrics:
         self.prof.set_training(True)
-        check.gt(step_id, 0)
+        if step_id <= 0:
+            raise RuntimeError(f"step_id={step_id}")
```
rb-determined-ai, line 381: this also feels pointless to me.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -544,7 +541,8 @@ def _compute_validation_metrics(self) -> workload.Metrics:
             batch_metrics = []
 
             self.validation_loader = cast(torch.utils.data.DataLoader, self.validation_loader)
```
rb-determined-ai, line 543: while we're adjusting checks, can you convert this to an assertion rather than a cast?

(there's basically no cases where I think a cast is better than an assertion)

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -44,22 +44,20 @@ def __init__(self, trial_inst: det.Trial, *args: Any, **kwargs: Any) -> None:
...
-            "Must have at least one model. "
-            "This might be caused by not wrapping your model with wrap_model()",
-        )
-        check.gt_eq(
-            len(self.context.optimizers),
-            1,
-            "Must have at least one optimizer. "
-            "This might be caused by not wrapping your optimizer with wrap_optimizer()",
-        )
+        if not self.context.models:
```
rb-determined-ai, line 51: Can we avoid truthy checks as substitutions for length checks?

They're valid sometimes, but they're not valid in all cases and it can bite you.  Better to be explicit I think.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -44,22 +44,20 @@ def __init__(self, trial_inst: det.Trial, *args: Any, **kwargs: Any) -> None:
...
+            raise det.errors.InvalidExperimentException(
+                "Must have at least one model. "
+                "This might be caused by not wrapping your model with wrap_model().",
+            )
+        if not self.context.optimizers:
```
rb-determined-ai, line 56: another case of truthy check instead of length check.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -544,7 +541,8 @@ def _compute_validation_metrics(self) -> workload.Metrics:
             batch_metrics = []
 
             self.validation_loader = cast(torch.utils.data.DataLoader, self.validation_loader)
-            check.gt(len(self.validation_loader), 0)
+            if not self.validation_loader:
+                raise RuntimeError("validation_loader is empty.")
```
rb-determined-ai, line 545: This is a case where the truthy check is not even kind of a replacement for the length check.

validation_loader will be an DataLoader, which will always be truthy except when it's None.  But it might have `len(validation_loader)==0` even if the validation_loader is not None.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -560,18 +558,16 @@ def _compute_validation_metrics(self) -> workload.Metrics:
                 if keys is None:
                     keys = vld_metrics.keys()
                 else:
-                    check.eq(
-                        keys,
-                        vld_metrics.keys(),
-                        "Validation metric names must match across all batches of data.",
+                    if keys != vld_metrics.keys():
+                        raise ValueError(
+                            "Validation metric names must match across all batches of data.",
```
rb-determined-ai, line 563: Can you put the contents of both keys and vld_metrics.keys() in the message?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -560,18 +558,16 @@ def _compute_validation_metrics(self) -> workload.Metrics:
...
+                        )
+                if not isinstance(vld_metrics, dict):
+                    raise TypeError(
+                        "validation_metrics() must return a "
+                        "dictionary of string names to Tensor "
+                        "metrics.",
```
rb-determined-ai, line 569: can you put the actual type of `vld_metrics` in the message?

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -597,14 +593,14 @@ def _compute_validation_metrics(self) -> workload.Metrics:
                 num_inputs, num_batches = [sum(n) for n in zip(*input_counts)]
 
         else:
-            check.true(self._evaluate_full_dataset_defined())
+            if not self._evaluate_full_dataset_defined():
+                raise RuntimeError()
```
rb-determined-ai, line 597: use an assertion here since it's validating our own code instead of validating user code.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -597,14 +593,14 @@ def _compute_validation_metrics(self) -> workload.Metrics:
...
             self.validation_loader = cast(torch.utils.data.DataLoader, self.validation_loader)
             if self.is_chief:
                 metrics = self.trial.evaluate_full_dataset(data_loader=self.validation_loader)
 
-                check.is_instance(
-                    metrics, dict, f"eval() must return a dictionary, got {type(metrics)}."
-                )
+                if not isinstance(metrics, dict):
+                    raise TypeError(f"eval() must return a dictionary, got {type(metrics)}.")
```
rb-determined-ai, line 603: change this to `type(metrics).__name__` which renders much nicer in the output message.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -671,8 +667,10 @@ def _load(self, load_path: pathlib.Path) -> None:
 
         if "model_state_dict" in checkpoint:
             # Backward compatible with older checkpoint format.
-            check.not_in("models_state_dict", checkpoint)
-            check.eq(len(self.context.models), 1)
+            if "models_state_dict" in checkpoint:
+                raise RuntimeError("Both model_state_dict and models_state_dict in checkpoint.")
+            if len(self.context.models) != 1:
+                raise RuntimeError()
```
rb-determined-ai, line 673: These checkpoint loading exceptions are technically loading user-provided values, and so need a real message.

"Old-format checkpoint cannot be loaded into a model with more than one checkpoint" would be fine.

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -696,17 +694,23 @@ def _load(self, load_path: pathlib.Path) -> None:
...
         if "optimizer_state_dict" in checkpoint:
             # Backward compatible with older checkpoint format.
-            check.not_in("optimizers_state_dict", checkpoint)
-            check.eq(len(self.context.optimizers), 1)
+            if "optimizers_state_dict" in checkpoint:
+                raise RuntimeError(
+                    "Both optimizer_state_dict and optimizers_state_dict in checkpoint."
+                )
+            if len(self.context.optimizers) != 1:
+                raise RuntimeError()
```
rb-determined-ai, line 702: ditto

harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -696,17 +694,23 @@ def _load(self, load_path: pathlib.Path) -> None:
...
                 optimizer.load_state_dict(checkpoint["optimizers_state_dict"][idx])
 
         if "lr_scheduler" in checkpoint:
             # Backward compatible with older checkpoint format.
-            check.not_in("lr_schedulers_state_dict", checkpoint)
-            check.eq(len(self.context.lr_schedulers), 1)
+            if "lr_schedulers_state_dict" in checkpoint:
+                raise RuntimeError("Both lr_scheduler and lr_schedulers_state_dict in checkpoint.")
+            if len(self.context.lr_schedulers) != 1:
+                raise RuntimeError()
```
rb-determined-ai, line 713: ditto

harness/determined/pytorch/_reducer.py
```diff
@@ -32,7 +30,8 @@ def _simple_reduce_metrics(
 ) -> np.float64:
     if reducer == Reducer.AVG:
         if num_batches:
-            check.check_eq(len(metrics), len(num_batches))
+            if len(metrics) != len(num_batches):
+                raise RuntimeError("Lengths of metrics and num_batches are not equal.")
```
rb-determined-ai, line 34: include values in message

harness/determined/pytorch/samplers.py
```diff
@@ -96,9 +94,12 @@ class DistributedBatchSampler(torch.utils.data.BatchSampler):
     def __init__(
         self, batch_sampler: torch.utils.data.BatchSampler, num_workers: int, rank: int
     ) -> None:
-        check.gt(rank, -1, "rank must be non-negative")
-        check.gt(num_workers, 0, "num_workers must be positive")
-        check.lt(rank, num_workers, "rank must be less than num_workers")
+        if rank < 0:
+            raise ValueError("rank must be non-negative.")
+        if num_workers < 1:
+            raise ValueError("num_workers must be strictly positive.")
```
rb-determined-ai, line 100: nit: either just "positive" or "greater than zero".  "Strictly" positive is not a term that I think is in common use (I've never seen it on any math problem).  I've always understood "positive" to mean what you mean here.

harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -299,12 +300,11 @@ def lightning_scheduler_dict_to_det(lrs: dict) -> _LRScheduler:
...
-                lrs["scheduler"].optimizer,
-                Optimizer,
-                "A returned LRScheduler from `configure_optimizers` is "
-                "missing the optimizer attribute.",
-            )
+            if not isinstance(lrs["scheduler"].optimizer, Optimizer):
+                raise TypeError(
+                    "A returned LRScheduler from `configure_optimizers` is "
+                    "missing the optimizer attribute.",
+                )
```
rb-determined-ai, line 307: not your fault, but the message and the check are not at all aligned.

If `lrs["scheduler"]` doesn't have a `.optimizer` attribute, an error is raised before the `isinstance()` can even evaluate.  So if `isinstance()` does evaluate to false the message is guaranteed to be wrong.

harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -170,8 +169,10 @@ def __init__(self, context: PyTorchTrialContext) -> None:
 
         """
 
-        check.check_in(precision, {16, 32}, "only precisions 16 & 32 are supported.")
-        check.check_in(amp_backend, {"native", "apex"}, 'only "native", and "apex" are supported')
+        if precision not in {16, 32}:
+            raise ValueError("Only precisions 16 & 32 are supported.")
+        if amp_backend not in {"native", "apex"}:
+            raise ValueError('Only "native" and "apex" AMP-backends are supported.')
```
rb-determined-ai, line 175: include value in message

harness/determined/pytorch/_lr_scheduler.py
drh-determined-ai: I was thinking the same thing, especially given the type-hinting.

rb-determined-ai (commented):
nice, looks like quite a few of these haven't had eyes on them in a long, long time

harness/determined/pytorch/_lr_scheduler.py
```diff
@@ -51,7 +49,7 @@ def __init__(
                 2. ``STEP_EVERY_BATCH``: Determined will call scheduler.step() after every
                    ``frequency`` training batch(es). No arguments will be passed to step().
                    This option does not take into account gradient aggregation;
-                   ``STEP_EVERY_OPTIMIZER_STEP`` which does is recommended.
+                   ``STEP_EVERY_OPTIMIZER_STEP`` which is recommended.
```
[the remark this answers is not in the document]
drh-determined-ai: I did my best to guess the intended meaning, given the bad grammar. What would be correct?

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Removed.

harness/determined/pytorch/_lr_scheduler.py
drh-determined-ai: Removed.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Removed.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Removed.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: I suppose so. Should we make this method a no-op if `scaler.is_enabled()` is false?

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Good call. Forgot that.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Thanks. I'm going to change the actual check to be `... > 1` to match the message.

drh-determined-ai: Waitasecond, do you mean "Old-format checkpoint cannot be loaded into a **context** with more than one **model**."

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Ditto, do you mean "Old-format checkpoint cannot be loaded into a **context** with more than one **optimizer**."?

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/_reducer.py
drh-determined-ai: Done.

harness/determined/pytorch/samplers.py
drh-determined-ai: I've definitely encountered that phrase in mathmatics, but as always, YMMV.
Anyway, done!

harness/determined/pytorch/lightning/_adapter.py
```diff
@@ -170,8 +169,10 @@ def __init__(self, context: PyTorchTrialContext) -> None:
 
         """
 
-        check.check_in(precision, {16, 32}, "only precisions 16 & 32 are supported.")
-        check.check_in(amp_backend, {"native", "apex"}, 'only "native", and "apex" are supported')
+        if precision not in {16, 32}:
+            raise ValueError("Only precisions 16 & 32 are supported.")
```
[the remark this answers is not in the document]
drh-determined-ai: Done.

harness/determined/pytorch/lightning/_adapter.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Oops, fixed.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Fixed.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Fixed.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: You mean like `assert isinstance(self.validation_loader, DataLoader), "blah"`?

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Oops, `TestPyTorchContext.test_average_gradients` actually tests for this case and for an exception being raised, so I'm going to leave this in.

harness/determined/pytorch/_lr_scheduler.py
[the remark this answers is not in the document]
drh-determined-ai: I looked at this doc-string more closely. I have removed the phrase in question; I think the doc is clear enough without it.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: Yeah, and I realized my typo after clicking "comment" but I was not concerned about you missing my typo.

You don't miss any of my typos lol.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: I sometimes slip into using these too.  They look so sleek.  And 95% of the time it's fine, but about 5% of the time it causes issues, and I'm not very good at picking which ones are going to cause issues.

harness/determined/pytorch/_pytorch_trial.py
rb-determined-ai: yeah, but if it's a type check only for mypy's sake (because we can already guarantee that by this point in execution the validation_loader is set) then it doesn't need any message.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: That's circular reasoning though.  There was a check that was pointless, then we wrote a test in the name of improving test coverage, now we are keeping the pointless check because we want to pass our test that was only written to improve test coverage?  That doesn't make sense.

harness/determined/pytorch/_pytorch_context.py
rb-determined-ai: no I think the other checks are all structural, like the order in which you call things, and those are fine to enforce in all cases I think.

harness/determined/pytorch/_lr_scheduler.py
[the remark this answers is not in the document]
drh-determined-ai: You're right. I went too far in this case. I've reverted the change.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: Gotcha. I've remove both the check and the test thereof.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done!

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/_pytorch_trial.py
drh-determined-ai: Done.

harness/determined/pytorch/lightning/_adapter.py
drh-determined-ai: Looking at the surrounding code, the check doesn't make sense to me, regardless of the message.
Anyway, I've changed the message to better reflect the check.

harness/determined/pytorch/_pytorch_context.py
drh-determined-ai: So aside from mentioning `wrap_scaler` (done), leave things as is?

drh-determined-ai: I understand this now. Done.

rb-determined-ai (approved):
nice work, thank you for this.

PR #5075, opened 2022-09-20
chore: oss refactor for auto assign WorkspaceCreator role plugin

NicholasBlaskey:
## Description

Converts ```PostWorkspace``` to bun so we can share a transaction with EE adding the user role and adds ```AssignWorkspaceAdminToUserTx``` as a plugin metod to APIServer which is a no-op in oss.

## Test Plan

Integration test to make sure we didn't break existing behavior.

## Checklist


ioga (approved):
cc @mapmeld

PR #5076, opened 2022-09-20
chore: add config item slots_per_node

phillip-gaisford:
## Description

This is part of story [FOUNDENG-225](https://jira-pro.its.hpecorp.net:8443/browse/FOUNDENG-225), which amongst other things calls for:

- Experiment config, and task_container_defaults have slots_per_node attribute (in oss tree)

With this change it is possible to specify slots per node at the experiment level:
image

or in the master config, either as a part of task_container_defaults:

image
or in a resource pool-specific fashion:

image

Adding this one property resulted in a large change set, including generated files. Did I get that right?
## Test Plan
Tested manually (on the determined-ee repo).

## Checklist


jerryharrow (approved):

stoksc (approved):
lgtm

stoksc:
cc @rb-determined-ai , you may be interested

stoksc:
this is a slurm/pbs parallel to slots_per_pod

harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -2085,6 +2085,14 @@
             ],
             "default": null
         },
+        "slots_per_node": {
+            "type": [
+                "integer",
+                "null"
+            ],
+            "minimum": 1,
+            "default": 1
```
stoksc, line 2094: I think maybe the default should be null and we should have custom validation to ensure it is set for pbs. @rb-determined-ai thoughts?

stoksc (approved):
lgtm, just one comment for Ryan to look at

harness/determined/common/schemas/expconf/_gen.py
jerryharrow: I did wonder if the default might be better as null -- such that we can tell if it were present or not, but I wasn't convinced that it mattered to the result.   Might be a bit clearer though.

It doesn't need to be set -- it just may be to use more than 1 GPUs when there is no gres configured (Slurm or PBS).

[AT-244122801-200922-1715-48.pdf](https://github.com/determined-ai/determined/files/9626059/AT-244122801-200922-1715-48.pdf)

rb-determined-ai:
It looks like we're venturing into overdefined territory.  If we have to, then we have to, but I want to be very careful about it.

So IIUC, the issue is that there are some schedulers which are only node-aware, and not slot-aware... is that correct?

Is it required that all gpu nodes in a pbs cluster have the same number of nodes?  If it is, why is this allowed to be set on the experiment config level?  If not, how do we guarantee the `slots_per_node` setting matches the node that the tasks get assigned to?  We have the same problem with the legacy `slots` value that is only there to not be a breaking schema change, but that's at least not documented, so users only discover it on accident.

What happens if `slots_per_node` is set with a scheduler that doesn't need `slots_per_node`?  If it is silently ignored, I can guarantee you we'll have users asking why changing it doesn't do anything, because there's no indication from the name that `slots_per_node` only applies to specific scheduling scenarios.

What happens if `slots_per_node` is incompatible with `slots_per_trial`, like `slots_per_trial=8` and `slots_per_node=16` or `slots_per_node=3`?

jerryharrow:
> Is it required that all gpu nodes in a pbs cluster have the same number of nodes?  etc.

The design note (which is focused on getting the necessary data to the schedulers to allocate the desired resources, and not necessarily addressing all the questions you raise above) defines the value as:

`slots_per_node (C):  The minimum number of slots required for a node to be scheduled during a trial.   It must be greater than slots_per_trial.   The default value is 1.`

The primary intent is to fix our existing behavior to allocate more than 1 GPU per node when the scheduling system is not configured to identify GPUs.   In this case the customer is responsible for ensuring that the nodes selected actually have slots_per_node GPUs or CPUs using a resource pool, or scheduler-specific constraint via (slurm or pbs expconf).

It basically always gets passed to the underling scheduler.   

- For slurm with contres, it is passed as --gpus-per-task=C
- For slurm with gres it is passed as --gres=gpu:C
- For slurm with CPU slots it is passed as --cpus-per-task=C

So for slurm, if the request cannot be satisfied, the scheduler will report an error.

For PBS the case is similar.    With the exception of allocating GPUs when PBS knows nothing about GPUs, the value is passed to PBS scheduler and errors may result if there are insufficient resources.

- If gres data is defined for GPUs, it is passed to  #PBS -l select="S/C:ngpus=C"
- If gres data is not defined to describe GPUs, it doesn't get passed to PBS,but the user is again responsible for ensuring that the systems selected have this many GPUs as described above.
- If CPU slots are being scheduled, it is passed to #PBS -l select="S/C:ncpus=C"

So that may not answer all your concerns, but frames up what we are trying to achive.

rb-determined-ai:
> It must be greater than slots_per_trial.

This doesn't sound right.  A Trial might get allocated slots from multiple nodes, but at least one node.  So slots_per_node seeems like it must be at most slots_per_trial.

>  In this case the customer is responsible for ensuring that the nodes selected actually have slots_per_node GPUs or CPUs using a resource pool, or scheduler-specific constraint via (slurm or pbs expconf).

So if the user messes this up, how does it get surfaced back to them?

> It basically always gets passed to the underling scheduler.

That can't be true, because basically all Determined installations are using determined agents or k8s.  As we add new configs for new schedulers that have different capabilities, we need to make sure the user experience for the bulk of our users doesn't become more complex and difficult.

jerryharrow:
> > It must be greater than slots_per_trial.
> 
> This doesn't sound right. A Trial might get allocated slots from multiple nodes, but at least one node. So slots_per_node seeems like it must be at most slots_per_trial.

I think you are right.  It must be less.    

> > In this case the customer is responsible for ensuring that the nodes selected actually have slots_per_node GPUs or CPUs using a resource pool, or scheduler-specific constraint via (slurm or pbs expconf).
> 
> So if the user messes this up, how does it get surfaced back to them?
> 
> > It basically always gets passed to the underling scheduler.
> 
> That can't be true, because basically all Determined installations are using determined agents or k8s. As we add new configs for new schedulers that have different capabilities, we need to make sure the user experience for the bulk of our users doesn't become more complex and difficult.

Agreed if this doesn't make senses for other RMs, we should reconsider.    We were only thinking of this from the Slrm/PBS perspective (Bradley was OK with it, so though he was covering the k8s/agents perspective).  So you now know what we are trying to achieve with slots_per_node for Slurm/PBS, so how would you recommend addressing this?     

Brainstorming Options:

1. Our only other placement is in partition_overrides I think.   That does work, but it has the additional requirement that a partition must have a consistent number of GPU slots per node.   Unfortunately,  that is not globally true in our internal systems.  It might be more common that customer systems are "regular".
2. Determine if there is a slots_per_node concept that can apply to k8s/agents?
3. Rename (or otherwise qualify it) to apply only to Slurm/PBS?
4. Other?

phillip-gaisford:
Superseded by work for [FOUNDENG-245](https://jira-pro.its.hpecorp.net:8443/browse/FOUNDENG-245)

harness/determined/common/schemas/expconf/_gen.py
phillip-gaisford: I was thinking we would pay attention to it only when > default

PR #5079, opened 2022-09-20
chore: Avoid error from RBAC listRole endpoint [DET-8395]

mapmeld:
## Description

Returns an empty list of possible roles to `listRoles`. This is separate from the what-role-am-I request.

First commit was an attempt to silence the 501 error, but TypeScript does not like it.

## Testing

No error appears in browser console, no request outbound to `roles/search`

## Checklist

### DIFF
--- a/webui/react/src/services/apiConfig.ts
+++ b/webui/react/src/services/apiConfig.ts
@@ -288,10 +288,12 @@ export const listRoles: DetApi<Service.ListRolesParams, Api.V1ListRolesResponse,
   {
     name: 'listRoles',
     postProcess: (response) => response.roles.map(decoder.mapV1Role),
-    request: (params) =>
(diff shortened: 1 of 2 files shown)


ClaireNeveu (approved):

webui/react/src/services/apiConfig.ts
```diff
@@ -288,10 +288,12 @@ export const listRoles: DetApi<Service.ListRolesParams, Api.V1ListRolesResponse,
   {
```
ClaireNeveu, line 288: Is it possible to just check `feature.isOn('rbac')` where we call `listRoles`?

mapmeld: I haven't tried out the feature flags yet. Agree our future setup should do something like this in the future.

PR #5083, opened 2022-09-21
feat: cli: support downloading checkpoints through master

hanyucui:
## Description

The CLI supports downloading checkpoints through the master. This is a follow-up PR for change #4989 and depends on the API added there.

## Test Plan

- Use `devcluster` to start a dev cluster. Configure the `checkpoint_storage` section to use an S3 bucket and provide appropriate AWS credentials. For example:
```yaml
checkpoint_storage:
  type: s3
  bucket: hanyu-det-test
  prefix: /devcluster/checkpoint-download-pr
  access_key: *** (redacted) ***
  secret_key: *** (redacted) ***
```
- Make sure the same or equivalent credentials are available in environment variables or `~/.aws/credentials.
- Run `export DET_MASTER=http://localhost:8080`
- Run a simple experiment, e.g., the MNIST example. When the experiment completes, a checkpoint should have been created.
- Obtain the checkpoint UUID under that experiment using the CLI.
- Go to the S3 bucket in the AWS Console to verify the checkpoint with the same UUID.
- Substitute `$UUID` in the following steps:
  1. Remove any previously downloaded checkpoints (if any)
  1. Run: `det c download --mode=master $UUID`
  1. Verify the contents of the downloaded checkpoint are the same as those on S3.
  1. Repeat steps i, ii, and iii above using these two additional commands:
    - `det c download --mode=direct $UUID`
    - `det c download --mode=auto $UUID`

## Checklist

### DIFF
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -95,7 +95,7 @@ def list_checkpoints(args: Namespace) -> None:
 def download(args: Namespace) -> None:
     checkpoint = Determined(args.master, None).get_checkpoint(args.uuid)
 
-    path = checkpoint.download(path=args.output_dir)
+    path = checkpoint.download(args.via_master, path=args.output_dir)
 
     if args.quiet:
         print(path)
@@ -144,6 +144,12 @@ main_cmd = Cmd(
                     "--quiet",
                     action="store_true",
                     help="Only print the path to the checkpoint.",
+                ),
+                Arg(
+                    "-m",
+                    "--via-master",
+                    action="store_true",
+                    help="Download the checkpoint through the master.",
                 ),
             ],
         ),
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -152,10 +152,10 @@ main_cmd = Cmd(
                     default=DownloadMode.AUTO,
                     type=DownloadMode,
                     help=(
-                        "Select different download modes. "
-                        f"'{DownloadMode.DIRECT}' directly download from checkpoint storage. "
-                        f"'{DownloadMode.MASTER}' download via the master. "
-                        f"'{DownloadMode.AUTO}' first attempt a direct download but fall "
+                        "Select different download modes: "
+                        f"'{DownloadMode.DIRECT}' to directly download from checkpoint storage; "
+                        f"'{DownloadMode.MASTER}' to download via the master; "
+                        f"'{DownloadMode.AUTO}' to first attempt a direct download and fall "
                         f"back to '{DownloadMode.MASTER}'."
                     ),
                 ),
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -95,7 +95,7 @@ def list_checkpoints(args: Namespace) -> None:
 def download(args: Namespace) -> None:
     checkpoint = Determined(args.master, None).get_checkpoint(args.uuid)
 
-    path = checkpoint.download(args.via_master, path=args.output_dir)
+    path = checkpoint.download(path=args.output_dir, via_master=args.via_master)
 
     if args.quiet:
         print(path)
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -153,10 +153,10 @@ main_cmd = Cmd(
                     type=DownloadMode,
                     help=(
                         "Select different download modes. "
-                        f"'{DownloadMode.DIRECT}' directly downloads from checkpoint storage. "
-                        f"'{DownloadMode.MASTER}' downloads via the master. "
-                        f"'{DownloadMode.AUTO}' first attempts a direct download but falls "
-                        f"back to '{DownloadMode.MASTER}'"
+                        f"'{DownloadMode.DIRECT}' directly download from checkpoint storage. "
+                        f"'{DownloadMode.MASTER}' download via the master. "
+                        f"'{DownloadMode.AUTO}' first attempt a direct download but fall "
+                        f"back to '{DownloadMode.MASTER}'."
                     ),
                 ),
             ],
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -7,7 +7,7 @@ from determined.common import experimental
 from determined.common.api import authentication, bindings
 from determined.common.declarative_argparse import Arg, Cmd
 from determined.common.experimental import Determined
-from determined.common.experimental.checkpoint import DownloadMode
+from determined.experimental.client import DownloadMode
 
 from . import render
 
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -7,6 +7,7 @@ from determined.common import experimental
 from determined.common.api import authentication, bindings
 from determined.common.declarative_argparse import Arg, Cmd
 from determined.common.experimental import Determined
+from determined.common.experimental.checkpoint import DownloadMode
 
 from . import render
 
@@ -95,7 +96,7 @@ def list_checkpoints(args: Namespace) -> None:
 def download(args: Namespace) -> None:
     checkpoint = Determined(args.master, None).get_checkpoint(args.uuid)
 
-    path = checkpoint.download(path=args.output_dir, via_master=args.via_master)
+    path = checkpoint.download(path=args.output_dir, mode=args.mode)
 
     if args.quiet:
         print(path)
@@ -146,9 +147,17 @@ main_cmd = Cmd(
                     help="Only print the path to the checkpoint.",
                 ),
                 Arg(
-                    "--via-master",
-                    action="store_true",
-                    help="Download the checkpoint through the master.",
+                    "--mode",
+                    choices=list(DownloadMode),
+                    default=DownloadMode.AUTO,
+                    type=DownloadMode,
+                    help=(
+                        "Select different download modes. "
+                        f"'{DownloadMode.DIRECT}' directly downloads from checkpoint storage. "
+                        f"'{DownloadMode.MASTER}' downloads via the master. "
+                        f"'{DownloadMode.AUTO}' first attempts a direct download but falls "
+                        f"back to '{DownloadMode.MASTER}'"
+                    ),
                 ),
             ],
         ),
--- a/harness/determined/cli/checkpoint.py
+++ b/harness/determined/cli/checkpoint.py
@@ -146,7 +146,6 @@ main_cmd = Cmd(
                     help="Only print the path to the checkpoint.",
                 ),
                 Arg(
-                    "-m",
                     "--via-master",
                     action="store_true",
                     help="Download the checkpoint through the master.",
--- a/harness/determined/common/experimental/checkpoint/_checkpoint.py
+++ b/harness/determined/common/experimental/checkpoint/_checkpoint.py
@@ -155,36 +155,9 @@ class Checkpoint:
             checkpoint_storage = self.training.experiment_config["checkpoint_storage"]
             try:
                 if via_master:
-                    if checkpoint_storage["type"] != "s3":
-                        raise NotImplementedError(
-                            "Checkpoint download via master is only supported on S3"
-                            f", got {checkpoint_storage['type']}"
-                        )
                     self._download_via_master(self._session, self.uuid, local_ckpt_dir)
-                elif checkpoint_storage["type"] == "shared_fs":
-                    src_ckpt_dir = self._find_shared_fs_path(checkpoint_storage)
-                    shutil.copytree(str(src_ckpt_dir), str(local_ckpt_dir))
                 else:
-                    local_ckpt_dir.mkdir(parents=True, exist_ok=True)
-                    manager = storage.build(
-                        checkpoint_storage,
-                        container_path=None,
-                    )
-                    if not isinstance(
-                        manager,
-                        (
-                            storage.S3StorageManager,
-                            storage.GCSStorageManager,
-                            storage.AzureStorageManager,
-                        ),
-                    ):
-                        raise AssertionError(
-                            "Downloading from Azure, S3 or GCS requires the experiment "
-                            "to be configured with Azure, S3 or GCS checkpointing"
-                            ", {} found instead".format(checkpoint_storage["type"])
-                        )
-
-                    manager.download(self.uuid, str(local_ckpt_dir))
+                    self._download_direct(checkpoint_storage, local_ckpt_dir)
 
             except storage.NoCloudAccess:
                 # Failing back to downloading via the master if due to NoCloudAccess.
@@ -206,6 +179,32 @@ class Checkpoint:
             self.write_metadata_file(str(metadata_path))
 
         return str(local_ckpt_dir)
+
+    def _download_direct(self, checkpoint_storage: Dict[str, Any], local_ckpt_dir: pathlib.Path):
+        if checkpoint_storage["type"] == "shared_fs":
+            src_ckpt_dir = self._find_shared_fs_path(checkpoint_storage)
+            shutil.copytree(str(src_ckpt_dir), str(local_ckpt_dir))
+        else:
+            local_ckpt_dir.mkdir(parents=True, exist_ok=True)
+            manager = storage.build(
+                checkpoint_storage,
+                container_path=None,
+            )
+            if not isinstance(
+                manager,
+                (
+                    storage.S3StorageManager,
+                    storage.GCSStorageManager,
+                    storage.AzureStorageManager,
+                ),
+            ):
+                raise AssertionError(
+                    "Downloading from Azure, S3 or GCS requires the experiment "
+                    "to be configured with Azure, S3 or GCS checkpointing"
+                    ", {} found instead".format(checkpoint_storage["type"])
+                )
+
+            manager.download(self.uuid, str(local_ckpt_dir))
 
     @staticmethod
     def _download_via_master(
--- a/harness/determined/common/experimental/checkpoint/_checkpoint.py
+++ b/harness/determined/common/experimental/checkpoint/_checkpoint.py
@@ -6,7 +6,6 @@ import pathlib
 import shutil
 import tarfile
 import tempfile
-import traceback
 import warnings
 import zipfile
 from typing import Any, Dict, List, Optional, cast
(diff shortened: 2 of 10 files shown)


hanyucui:
This seems a simple enough change. Do I need to add tests?

harness/determined/cli/checkpoint.py
```diff
@@ -145,6 +145,12 @@ def delete_checkpoints(args: Namespace) -> None:
                     action="store_true",
                     help="Only print the path to the checkpoint.",
                 ),
+                Arg(
+                    "-m",
```
rb-determined-ai, line 149: `-m` is already a global flag for setting the master address; let's not add another `-m` with another meaning in a different scope

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -123,6 +125,8 @@ def download(self, path: Optional[str] = None) -> str:
           - :func:`determined.estimator.load_estimator_from_checkpoint_path`
 
         Arguments:
+            via_master (bool): If true, checkpoints are downloaded through the master,
+                otherwise they are downloaded directly from checkpoint storage.
```
rb-determined-ai, line 129: Overall, I think it's a bad idea to make the user specify this value.  That's sort of leaking out implementation details into the user interface.

I would propose that `det checkpoint download UUID` first try and download UUID directly (which will be faster when it's available, and easier on us too), then fall back to trying to proxy it through the master, then give up.

I'm ok allowing a flag that lets the user pick proxy vs direct, but I think the default behavior needs to be smarter.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +195,33 @@ def download(self, path: Optional[str] = None) -> str:
...
+                "unable to download checkpoint from master:", resp.status_code, resp.reason
+            )
+
+        tmpfile = pathlib.Path(f"{local_ckpt_dir}/{self.uuid}.{fext}")
+        with open(tmpfile, "wb") as f:
+            f.write(resp.content)
+
+        try:
+            if fext == "zip":
+                subprocess.run(["unzip", "-d", local_ckpt_dir, tmpfile], check=True)
```
rb-determined-ai, line 218: Let's not do this as a subprocess (or deal with os-specific codepaths).

Instead, let's use the [tar library](https://docs.python.org/3/library/tarfile.html) that comes with python in all cases, and do it in python.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +195,33 @@ def download(self, path: Optional[str] = None) -> str:
...
+        else:
+            fext = "tgz"
+            mime_type = "application/gzip"
+        resp = self._session.get(f"/checkpoints/{self.uuid}", headers={"Accept": mime_type})
+        if not resp.ok:
+            raise RuntimeError(
+                "unable to download checkpoint from master:", resp.status_code, resp.reason
+            )
+
+        tmpfile = pathlib.Path(f"{local_ckpt_dir}/{self.uuid}.{fext}")
```
rb-determined-ai, line 212: I like using a file for temporary storage, but can we use `tempfile.mkstemp()` instead?  (see notes [here](https://github.com/determined-ai/determined/blob/master/harness/determined/cli/shell.py#L106) as to why it's valuable)

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +195,33 @@ def download(self, path: Optional[str] = None) -> str:
...
 
+    def _download_via_master(self, local_ckpt_dir: pathlib.Path) -> None:
+        local_ckpt_dir.mkdir(parents=True, exist_ok=True)
+        if os.name == "nt":
+            fext = "zip"
+            mime_type = "application/zip"
+        else:
+            fext = "tgz"
+            mime_type = "application/gzip"
+        resp = self._session.get(f"/checkpoints/{self.uuid}", headers={"Accept": mime_type})
```
rb-determined-ai, line 206: I think you can set `stream=True` here... otherwise you'll be at risk of OOM issues on a regular basis.

harness/determined/common/experimental/checkpoint/_checkpoint.py
rb-determined-ai: Actually, hm... maybe if you set `stream=True` on the `session.get()` you might be able to stream straight from the network into the tar extraction.  Let me poke around a bit.

rb-determined-ai: Looks like you have to [make your own file-like object](https://github.com/psf/requests/issues/465#issuecomment-7276058) to do it, but you would really decrease disk io costs.  That's worth it I think, while we are building something to download multiple GB, especially when our server doesn't support resuming downloads on failures anyway.

harness/determined/cli/checkpoint.py
hanyucui: Sure, I don't think we need its short form.

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: Yeah, I thought about that but didn't dig enough. Let me take a look.

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: Sure, and [zipfile](https://docs.python.org/3/library/zipfile.html) for Windows?

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: Sure, thought about that but didn't dig enough.

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: I thought about that but there is a challenge: A user would either always have credentials or never. In the latter case, the user will always experience the overhead of trying to download directly and then fall back. And if the first attempt has any adverse effect that we fail to handle, the user will also experience it unnecessarily. It seems I can't sufficiently hide the implementation details and had better give users the choice. Suggestions?

rb-determined-ai: Ok.  I think that's a good case for _allowing_ the user to specify proxy vs direct, but not a good case to _require_ the user to do so.  It should be possible to handle failed direct downloads for like 99% of users so that most users never need to know about the `--via-master` flag.

hanyucui: That sounds good. I can fall back to downloading via the master if direct downloads fail.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -112,7 +116,7 @@ def _find_shared_fs_path(self, checkpoint_storage: Dict[str, Any]) -> pathlib.Pa
             "checkpoint storage configuration.".format(self.uuid, potential_paths)
         )
 
-    def download(self, path: Optional[str] = None) -> str:
+    def download(self, path: Optional[str] = None, via_master: bool = False) -> str:
```
rb-determined-ai, line 119: There's three different behaviors we want to support: auto, via master, and direct, right?  As you said, there will be times that the user wants to control via-master-vs-direct, so a bool isn't enough to express the level of control we want to give the user.

Probably it's time for an enum here, that defaults to AUTO.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -147,31 +153,48 @@ def download(self, path: Optional[str] = None) -> str:
...
-                        storage.S3StorageManager,
-                        storage.GCSStorageManager,
-                        storage.AzureStorageManager,
-                    ),
-                ):
-                    raise AssertionError(
-                        "Downloading from Azure, S3 or GCS requires the experiment to be "
-                        "configured with Azure, S3 or GCS checkpointing, {} found instead".format(
-                            checkpoint_storage["type"]
+            try:
```
rb-determined-ai, line 156: I _think_ what you have here is correct, but I think the structure of if statements and try statements is not making the behavior of the code very clear.

How about something like this pseudocode:
```python
if mode == AUTO:
     try:
         self._download_direct()
     except NoCloudAccess as e1:
         logging.info("unable to download directly, proxying download through master")
         self.download_via_master
elif mode == VIA_MASTER:
     self._download_via_master()
elif mode == DIRECT:
     self._download_direct()
else:
    raise ValueError(mode)
```

A structure like this is clearer than the one-big-try-except pattern because it's not clear from the one-big-try-except which codepaths might result in the specified exception.  Also it is very clear that the mode provided by the user is explicitly controlling the codepath (in a way that a bool isn't really capable of)

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: Didn't think about direct-only, but I can add it.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -147,31 +153,48 @@ def download(self, path: Optional[str] = None) -> str:
...
+                if via_master:
+                    if checkpoint_storage["type"] != "s3":
+                        raise NotImplementedError(
```
rb-determined-ai, line 159: Can we make this a `NoCheckpointDownloadViaMaster` or something like that (but better named)?

I don't see why we need to distinguish between the "s3 is configured but master doesn't have permissions" vs the "gcs is configured and master doesn't support that yet" cases.  From the user perspective, both of those mean "I have to figure out my own direct access", so I think they should be the same error.

harness/determined/common/storage/base.py
```diff
@@ -5,6 +5,14 @@
 from typing import Any, Dict, Iterator, Optional, Union
 
 
+class NoCloudAccess(Exception):
+    """Throw this exception when the operation fails due to denied access or missing credntials.
+    It allows higher level code to catch the exception and fall back to other methods.
```
rb-determined-ai, line 10: How about just:

```
Checkpoint download failed due to denied access or missing credentials.
```

Also note that `credntials` is misspelled.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +207,49 @@ def download(self, path: Optional[str] = None) -> str:
...
+    @staticmethod
+    def _download_via_master(
+        sess: api.Session, uuid: str, local_ckpt_dir: pathlib.Path, os_name: str = os.name
+    ) -> None:
+        """Downloads a checkpoint through the master.
+        Arguments:
+            sess (api.Session): a session for the download
+            uuid (string): the uuid of the checkpoint to be downloaded
+            local_ckpt_dir (Path-like): the local directory where the download is downloaded
+            os_name (string, optional): the name of the current OS -- this is for testing only
```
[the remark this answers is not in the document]
hanyucui: Aren't we going to use zip on Windows? Can you elaborate?

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +207,49 @@ def download(self, path: Optional[str] = None) -> str:
...
+                    shutil.copyfileobj(resp.raw, tmpf)
+                # Decompress the zip file.
+                with zipfile.ZipFile(tmpfpath) as zf:
+                    zf.extractall(local_ckpt_dir)
+            finally:
+                # Remove the zip file after decompression.
+                pathlib.Path(tmpfpath).unlink()
+        else:
+            # gunzip and untar. tarfile.open can detect the compression algorithm
+            with tarfile.open(fileobj=resp.raw) as tf:
```
rb-determined-ai, line 250: neat, I was under the impression that `resp.raw` would contain the chunked encoding framing as well, but I tested it just now and this totally works.  Much easier than what I was suggesting.

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: Or I don't check it here and rely on the API error, which provides similar info.

harness/determined/cli/checkpoint.py
```diff
@@ -145,6 +146,19 @@ def delete_checkpoints(args: Namespace) -> None:
...
+                    "--mode",
+                    choices=list(DownloadMode),
+                    default=DownloadMode.AUTO,
+                    type=DownloadMode,
+                    help=(
+                        "Select different download modes: "
+                        f"'{DownloadMode.DIRECT}' to directly download from checkpoint storage; "
+                        f"'{DownloadMode.MASTER}' to download via the master; "
+                        f"'{DownloadMode.AUTO}' to first attempt a direct download and fall "
+                        f"back to '{DownloadMode.MASTER}'."
```
hanyucui, line 159: This is what it looks like:
```
usage: det checkpoint download [-h] [-o OUTPUT_DIR] [-q] [--mode {direct,master,auto}] uuid

positional arguments:
  uuid                  Download a checkpoint by specifying its UUID.

optional arguments:
  -h, --help            show this help message and exit
  -o OUTPUT_DIR, --output-dir OUTPUT_DIR
                        Desired output directory for the checkpoint. (default: None)
  -q, --quiet           Only print the path to the checkpoint. (default: False)
  --mode {direct,master,auto}
                        Select different download modes: 'direct' to directly download from checkpoint storage; 'master' to download via the master; 'auto' to first attempt a direct download and fall back to
                        'master'. (default: auto)
```

harness/determined/cli/checkpoint.py
```diff
@@ -7,6 +7,7 @@
 from determined.common.api import authentication, bindings
 from determined.common.declarative_argparse import Arg, Cmd
 from determined.common.experimental import Determined
+from determined.common.experimental.checkpoint import DownloadMode
```
rb-determined-ai, line 10: Defining `DownloadMode` where you have it is fine, but can you also import it into `determined/experimental/client.py` and import it from there?

I know that this cli code is messy, but that's the path it _should_ be imported from, both by internal code and end users.  This `det.common.experimental.checkpoint` path is deprecated, it just hasn't been removed yet.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +189,53 @@ def download(self, path: Optional[str] = None) -> str:
...
+            uuid (string): the uuid of the checkpoint to be downloaded
+            local_ckpt_dir (Path-like): the local directory where the checkpoint is downloaded
+        """
+        local_ckpt_dir.mkdir(parents=True, exist_ok=True)
+
+        resp = sess.get(f"/checkpoints/{uuid}", headers={"Accept": "application/gzip"}, stream=True)
+        if not resp.ok:
+            raise RuntimeError(
+                "unable to download checkpoint from master:", resp.status_code, resp.reason
+            )
```
rb-determined-ai, line 234: Does the master return a specific error code here, that we could detect and raise a specific error type?  This seems like a case where that would be useful.

If the master doesn't do that, then I think it's fine to leave this as is.

harness/determined/common/experimental/checkpoint/_checkpoint.py
```diff
@@ -184,6 +189,53 @@ def download(self, path: Optional[str] = None) -> str:
...
+        # gunzip and untar. tarfile.open can detect the compression algorithm
+        with tarfile.open(fileobj=resp.raw) as tf:
+            tf.extractall(local_ckpt_dir)
```
rb-determined-ai, line 237: It's actually dangerous to call `extractall` without any verification of the paths inside the tarball.

When I was thinking about this yesterday, I was thinking that it was ok to not verify because the master is creating tarballs on the fly so it could guarantee that any tarballs it serves are well-behaved.  But right now, I think it would be possible to craft an s3 bucket with keys like `checkpoints/UUID/../.ssh/config` that could result in the master generating malicious tarballs.

Could we make a ticket to make sure the master validates the paths it puts into the tarball?  There's no reason why a checkpoint should ever contain an `..` element in any path name.

Then we could put a comment here saying that the master validates tarballs before emitting them so this is a fairly trustworthy tarball source.

harness/determined/common/storage/base.py
```diff
@@ -5,6 +5,12 @@
 from typing import Any, Dict, Iterator, Optional, Union
 
 
+class NoCloudAccess(Exception):
```
rb-determined-ai, line 8: Don't call this `cloud` access, because someday shared_fs checkpoints could also be proxied through the master.

rb-determined-ai: `NoDirectAccess`?

Also, can you put it in `det.errors.NoDirectAccess`?

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: That makes sense and I will file a ticket. I spend too much time on SaaS and keep forgetting a lot of users have direct access to checkpoint storage.

harness/determined/common/storage/base.py
hanyucui: Sure. If we put it in `det.errors`, might want to make it more specific, something like `NoDirectStorageAccess`?

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: The master sends different http status codes depending on the specific error but I don't think a single error type can cover them all here. And I think it's clear enough to just pass the http code and reason to the user. Maybe I missed some context?

rb-determined-ai: It just seems like a case where a user using the python sdk might want to be able to catch an exception that basically means "this master won't let you proxy the download of this checkpoint".

No particular use case at the moment.

harness/determined/common/storage/base.py
rb-determined-ai: Sure, sounds good.

harness/determined/common/experimental/checkpoint/_checkpoint.py
hanyucui: https://determinedai.atlassian.net/browse/DET-8501

harness/determined/common/storage/s3.py
```diff
@@ -8,6 +8,7 @@
 
 from determined import errors
 from determined.common import storage, util
+from determined.errors import NoDirectStorageAccess
```
rb-determined-ai, line 11: outside of the CLI, which has never been reigned in, we follow the [google style guide for python imports](https://google.github.io/styleguide/pyguide.html#22-imports), which basically says "other than `typing` don't import symbols directly".

You'll notice there's already a `from determined.errors` in this file, so you can delete this line and use `errors.NoDirectStorageAccess`.  The import patterns with that go style guide are a lot like go: you import the package and then use `package.Symbol`.

rb-determined-ai (approved):
This looks great, excellent work!

PR #5086, opened 2022-09-21
fix: allow sort of metrics under avg_metrics [DET-8408]

mapmeld:
## Description

Experiment page's Workloads table was having trouble sorting by metric. Discovered this was due to metrics being wrapped inside of metrics->avg_metrics since #4746  .  This change attempts to sort by metrics->>key and metrics->avg_metrics->>key. For some reason I also had to cast this to a numeric type to sort by value.

## Test Plan

https://determinedai.atlassian.net/browse/DET-8408 describes an experiment which generates metrics and is always sorted server-side by batch number, which then confusingly gets sorted client-side.

## Checklist


ioga:
good catch. we need an automated test for this.

>This change attempts to sort by metrics->>key and metrics->avg_metrics->>key

why is `metrics->>key` sorting is necessary at all?

mapmeld:
> why is metrics->>key sorting necessary

Have all past and future recorded metrics been converted to metrics->avg_metrics->>key ?

ioga:
> > why is metrics->>key sorting necessary
> 
> Have all past and future recorded metrics been converted to metrics->avg_metrics->>key ?

database representation of these hasn't changed, it's all in training/validation CTEs in the queries. 
they either use raw `metrics` column for batch metrics (in which case it has `avg_metrics` natively), or we just pick the `avg_metrics` component from it using `jsonb_build_object`. unless there's a bug in the query, I believe it should all be `metrics->avg_metrics`.

mapmeld:
@ioga - OK that works. Added a client-side fix for the remaining issue

ioga (approved):

PR #5087, opened 2022-09-21
feat: support single-file contexts in cli

rb-determined-ai:
This is especially useful for interactive things like notebooks or shells, where a user may want to upload a single file, but not want to move that file into an empty directory.

It also works for experiments, though presumably it's less useful in those cases.

## Test Plan

- [x] added unit tests to provide automated coverage

### DIFF
--- a/harness/determined/common/context.py
+++ b/harness/determined/common/context.py
@@ -35,12 +35,12 @@ def v1File_to_dict(f: bindings.v1File) -> Dict[str, Any]:
     return d
 
 
-def v1File_from_local_file(path: str, local_path: pathlib.Path) -> bindings.v1File:
-    with local_path.open("rb") as f:
+def v1File_from_local_file(archive_path: str, path: pathlib.Path) -> bindings.v1File:
+    with path.open("rb") as f:
         content = base64.b64encode(f.read()).decode("utf8")
-    st = local_path.stat()
+    st = path.stat()
     return bindings.v1File(
-        path=path,
+        path=archive_path,
         type=ord(tarfile.REGTYPE),
         content=content,
         # Protobuf expects string-encoded int64
@@ -51,10 +51,10 @@ def v1File_from_local_file(path: str, local_path: pathlib.Path) -> bindings.v1Fi
     )
 
 
-def v1File_from_local_dir(path: str, local_path: pathlib.Path) -> bindings.v1File:
-    st = local_path.stat()
+def v1File_from_local_dir(archive_path: str, path: pathlib.Path) -> bindings.v1File:
+    st = path.stat()
     return bindings.v1File(
-        path=path,
+        path=archive_path,
         type=ord(tarfile.DIRTYPE),
         content="",
         # Protobuf expects string-encoded int64
@@ -66,16 +66,23 @@ def v1File_from_local_dir(path: str, local_path: pathlib.Path) -> bindings.v1Fil
 
 
 def read_v1_context(
-    local_path: pathlib.Path,
+    root_path: pathlib.Path,
     limit: int = constants.MAX_CONTEXT_SIZE,
 ) -> List[bindings.v1File]:
     """
     Return a list of v1Files suitable for submitting a context directory over the v1 REST API.
 
+    If root_path refers to a file, a context directory is created containing only that file.
+
     A .detignore file in the directory, if specified, indicates the wildcard paths
     that should be ignored. File paths are represented as relative paths (relative to
     the root directory).
     """
+
+    root_path = root_path.resolve()
+
+    if not root_path.exists():
+        raise Exception(f"Path '{root_path}' doesn't exist")
 
     items = []
     size = 0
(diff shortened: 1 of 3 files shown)


harness/determined/common/context.py
```diff
@@ -66,34 +66,60 @@ def v1File_from_local_dir(path: str, local_path: pathlib.Path) -> bindings.v1Fil
...
-        raise ValueError(f"Path '{local_path}' must be a directory")
+    if root_path.is_file():
+        # Single-file context case.
+        add_item(v1File_from_local_file(root_path.name, root_path))
+        update_msg()
+        print()
+        return items
 
-    root_path = local_path
+    root_path = root_path
```
garrett361, line 122: find-replace?

rb-determined-ai: Yes... why accept a badly named parameter, then rename it halfway through the function without modification?  So I fixed it.

rb-determined-ai: ah, but I see the mistake here, thank you.

harness/determined/common/context.py
```diff
@@ -66,34 +66,58 @@ def v1File_from_local_dir(path: str, local_path: pathlib.Path) -> bindings.v1Fil
...
+    root_path = root_path.resolve()
+
+    if not root_path.exists():
+        raise Exception(f"Path '{root_path}' doesn't exist")
+
     items = []
     size = 0
 
     def add_item(f: bindings.v1File) -> None:
         nonlocal size
```
garrett361, line 91: Quibble, but all of the `nonlocal` calls seem awkward to me.  Probably not worth the re-write, but I'd personally find the code easier to follow if, for instance, `add_item` took on a `size` arg and returned the updated size:
```python
def add_item(f: bindings.v1File, size: int) -> int:
    # ....
    size += v1File_size(f)
    # ...
    return size
```
and use the output to update `size` in the outer scope.  Similar for `msg`

garrett361 (changes requested):
I tested using a single file as `--context` in `devcluster` and the file was included as expected.

This would be much more useful, however, if we could specify multiple individual files and/or dirs as `--context`. Obviously, this would be scope-creep, but what do you think?

My personal use case is where I want to load in one `ipynb` file and one `py` file to be used with the former without needing to shove them into a common directory to accomplish what I want.

harness/determined/common/context.py
rb-determined-ai: The level at which I care about code cleanliness changes according to the scope it affects.

Changes that affect a whole code base are really important to me, and I care a lot about even very tiny details.  Changes which only affect a single file are of small importance.  Changes that only affect the inside of a single function are firmly in the "don't care" category, even though I don't disagree with your reasoning... I like when we can write functions without side effects.

In this particular case though, I would also point out that even if you avoided the `nonlocal size` you'd still be modifying `items` inside `add_item()` so returning size explicitly doesn't actually make the function pure.  I personally find that _more_ confusing.

rb-determined-ai:
> This would be much more useful, however, if we could specify multiple individual files and/or dirs as --context.

Can you tell me what you would expect inside `/run/determined/workdir` in each of the following situations:

1. `--context .`
2. `--context . --context d1/`
3. `--context .. --context d1/`
4. `--context ../d1 --context d1/`
5. `--context d1/f1 --context d2/f1`
6. `--context /tmp/f1 --context f1`

harness/determined/common/context.py
garrett361: All fair. As I said, a quibble, and a good point about `items`.

garrett361:
> Can you tell me what you would expect inside /run/determined/workdir in each of the following situations:

In general, I would expect to first pass a space-separated list of files and directories as a single `--context` arg, so `--context d1 f1` rather than `--context d1 --context f1`, and I'd expect `/run/determined/workdir` to contain the contents of all directories (and subdirectories thereof) passed to `--context` and all files passed to `--context`.   When this leads to duplicates and/or name clashes, I think we just raise a ~`ValueError`

So, 

1. `--context .`: Everything from the cwd and subdirectories thereof, same as now.
2. `--context . --context d1/`: Everything from the cwd and subdirs, plus the contents of `d1/` and subdirs placed at the same level. So, the contents of `d1/` would be included twice, at two different levels of  `/run/determined/workdir`.  This is basically a user error, but not one that we need to raise.
3. `--context .. --context d1/`: Similar to 2, with cwd replaced by `..`
4. `--context ../d1 --context d1/`: the content of the `d1` dir in the cwd and that of `d1` one level up, and all subdirs, as always
5: `--context d1/f1 --context d2/f1`: Error due to name clash
6: `--context /tmp/f1 --context f1:` Error due to name clash

I think the above consistently extends the current functionality in a useful way. Am I overlooking anything obvious?

rb-determined-ai:
Do your answers to 1. and 2. seem to indicate that `--context .` means different things based on if other `--context` options appear on the command line?

Currently, `--context .` means that `./f1` will appear as `/run/determined/workdir/f1`, but it sounds like `--context . --context d1/` would cause `./f1` to appear as `/run/determined/workdir/CWD_NAME/f1`, is that accurate?

garrett361:
> Do your answers to 1. and 2. seem to indicate that `--context .` means different things based on if other `--context` options appear on the command line?

No, I don't believe so. Was finding it a little hard to state my expectations precisely.  Trying again below.


> Currently, `--context .` means that `./f1` will appear as `/run/determined/workdir/f1`, but it sounds like `--context . --context d1/` would cause `./f1` to appear as `/run/determined/workdir/CWD_NAME/f1`, is that accurate?

I would expect both cases to lead to `/run/determined/workdir/f1`, for consistency with how `--context` works today.

Maybe this is helpful: if my cwd looks like

```bash
.
├── d1
│   └── f1
├── d2
│   ├── f2
│   └── f3
└── f1
```

then I would expect the following behaviors:


```bash
--context .

workdir
├── d1
│   └── f1
├── d2
│   ├── f2
│   └── f3
└── f1
```

```bash
--context . d1/

Name clash error due to attempting to place f1 from . and f1 from d1/ into workdir/ simultaneously
```

```bash
--context . d1/f1

Name clash error due to attempting to place f1 from . and f1 from d1/ into workdir/ simultaneously
```

```bash
--context . d2/

workdir
├── d1
│   └── f1
├── d2
│   ├── f2
│   └── f3
├── f1
├── f2
└── f3
```

```bash
--context d1/ d2/

workdir
├── f1
├── f2
└── f3
```

So always copying the contents of any provided dir `workdir/` and always placing individually listed files in `workdir/`.  I think that's needed for consistency with how `--context` works today.

rb-determined-ai:
> So always copying the contents of any provided dir to workdir/ and always placing individually listed files in workdir/. I think that's needed for consistency with how --context works today.

To me it would be very confusing to have `--context f1 --context f2` to result in `f1` and `f2` in `workdir/`, but for `--context d1 --context d2` to not also result in `d1` and `d2` in `workdir/`.

garrett361:
> To me it would be very confusing to have `--context f1 --context f2` to result in `f1` and `f2` in `workdir/`, but for `--context d1 --context d2` to not also result in `d1` and `d2` in `workdir/`.

I agree that's confusing, but we maybe disagree on which part is causing the confusion.  

Right now `--context d1 --context d2` does not result in `d1` and `d2` subdirs within `workdir/`; instead their contents are exploded into `workdir` directly.  Am I correct in reading that you are fine with that behavior, and it's the `--context f1 --context f2` behavior you have above that causes the confusion?

rb-determined-ai:
no, `--context f1 --context f2` has behavior I find intuitive, but the behavior you are proposing for `--context d1 --context d2` seems directly in conflict to the `--context f1 --context f2` behavior.

It may be that the ugly duckling here is the current `--context .` behavior.

garrett361:
I think both the `--context .` and `--context d1` both have the ugly ducking behavior that we are maybe stuck with.

rb-determined-ai:
closing in favor of the `-i`/`--include` behavior we planned out in slack.

PR #5090, opened 2022-09-21
fix: new states can be paused, canceled, killed [DET-8449]

mapmeld:
## Description

Addressing an issue in the release - when 1+ experiments are selected, the new run states should allow the same actions which were allowed for Active state.  This is pause/cancel/kill.

## Test Plan

On Project details page or Experiment details page, try to cancel or kill a running experiment.

## Checklist


hamidzr (approved):
thanks for the fix. 

side note: did we take out the option to kill and experiment from the experiment detail page?

webui/react/src/constants/states.ts
```diff
@@ -28,17 +28,24 @@ const jobStates: Array<JobState> = [
   JobState.SCHEDULED,
   JobState.SCHEDULEDBACKFILLED,
 ];
-export const killableRunStates: CompoundRunState[] = [
+const activeStates: Array<RunState> = [
```
hamidzr, line 31: nit: add a jsdoc style `/* are states that ... */` comment to clarify the potential future usecases

mapmeld: ✅

PR #5092, opened 2022-09-21
fix: Initial comm failure should provide explaination (FDN-217)

jerryharrow:
## Description

Fix the exception raised when master is not reachable (Exception) -- APIHttpError only happens when call
completes without a successful status response could also be (SSLException, MasterNotFoundException).
This prevents the expected error message from being reported when the master is not reachable.

The initial prep_container was intended to fail reasonably quickly
to enable diagnostics of misconfiguration.   Recent changes
for re-using the common session (DET-8003) settings has caused the initial
communication to retry for more than 30 mins thus defeating the
original intent of prep_container.trial_prep and the error message
it provides.

This change lowers the session retry in prep_container (6 retries with 0.5 backoff -> 64 seconds) to enable the diagnostic message to be posted reasonably quickly.

## Test Plan

Verified via the  test-e2e-slurm-misconfigured test now succeeding on determined-ee which detects & reports the misconfigured message.   It now succeeds in ~5 mins instead of timing out at 30 mins.

## Checklist

### DIFF
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -25,7 +25,7 @@ def trial_prep(sess: api.Session, info: det.ClusterInfo) -> None:
     model_def_resp = None
     try:
         model_def_resp = bindings.get_GetModelDef(sess, experimentId=trial_info.experiment_id)
-    except api.errors.MasterNotFoundException as e:
+    except Exception as e:
         # Since this is the very first api call in the entrypoint script, and the call is made
         # before you can debug with a startup hook, we offer an overly-detailed explanation to help
         # sysadmins debug their cluster.
--- a/harness/determined/exec/prep_container.py
+++ b/harness/determined/exec/prep_container.py
@@ -9,6 +9,7 @@ import uuid
 from typing import List, Optional
 
 import psutil
+import urllib3
 
 import determined as det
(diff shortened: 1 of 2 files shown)


harness/determined/exec/prep_container.py
```diff
@@ -24,12 +25,11 @@ def trial_prep(sess: api.Session, info: det.ClusterInfo) -> None:
     model_def_resp = None
     try:
         model_def_resp = bindings.get_GetModelDef(sess, experimentId=trial_info.experiment_id)
-    except bindings.APIHttpError as e:
+    except api.errors.MasterNotFoundException as e:
```
stoksc, line 28: I dont think this change will work. If you look at the underlying code, SSLError should still also be meaningful, possibly other that aren't caught explicitly https://github.com/determined-ai/determined/blob/e1971c814cf7522133ab1442654ea2e2e4a7d8af/harness/determined/common/api/request.py#L142-L147

jerryharrow: I'll put it back to Exception which is what it was before the DET-8003 change.

stoksc: that one doesn't make sense to me either, tbh (cc @rb-determined-ai maybe it is his code)

PR #5096, opened 2022-09-22
chore: include running eslint fix in fmt target

hamidzr:
## Commentary (optional)

unrelated the bracket spacing was removed for `[]`  but not for `{}`  maybe we should have included that
DeepinScreenshot_select-area_20220921170819

lower verbosity level for prettier
DeepinScreenshot_select-area_20220921170723

## Checklist

### DIFF
--- a/webui/react/Makefile
+++ b/webui/react/Makefile
@@ -47,11 +47,11 @@ check-eslint:
 check-stylelint:
 	npm run stylelint
 check-prettier-js:
-	npm run format-check:js
+	npm run format-check:js > /dev/null
 check-prettier-css:
-	npm run format-check:css
+	npm run format-check:css > /dev/null
 check-prettier-misc:
-	npm run format-check:misc
+	npm run format-check:misc > /dev/null
 .PHONY: check-package-lock
 check-package-lock:
 	if grep 'ssh://' package-lock.json ; then echo "ssh url in package-lock.json, please convert to https url" ; false ; fi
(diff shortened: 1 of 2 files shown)


keita-determined:
> unrelated the bracket spacing was removed for [] but not for {} maybe we should have included that

This is expected. Prettier removed space in array but add space in object. I think its hard to see object without space

webui/react/Makefile
```diff
@@ -47,11 +47,11 @@ check-eslint:
 check-stylelint:
 	npm run stylelint
 check-prettier-js:
-	npm run format-check:js
+	npm run format-check:js > /dev/null
```
keita-determined, line 50: im not familer with `> /dev/null`. What does that do?

keita-determined: i figure it out. How come disabling output of command?

webui/react/Makefile
```diff
@@ -64,15 +64,20 @@ check-prettier: check-prettier-js check-prettier-css check-prettier-misc
...
+.PHONY: fmt-js-eslint
+fmt-js-eslint:
+	npm run eslint -- --fix
+.PHONY: fmt-js-prettier
+fmt-js-prettier:
+	npm run format:js > /dev/null
 .PHONY: fmt-js
-fmt-js:
-	npm run format:js
+fmt-js: fmt-js-eslint fmt-js-prettier
```
keita-determined, line 74: Did you see the difference between with `fmt-js-eslint` and without it? 
I believe that there's no difference but probably good to have both just in case

hamidzr: I tried to demo this behavior with the commit sequence of the branch https://github.com/determined-ai/determined/pull/5096/commits
- checkout the first commit and run `check` see it fails
- run fmt then run check and it still fails
- checkout the next commit. fmt. and check passes now

keita-determined: you're right. we should def have `fmt-js-eslint` then. Thanks for the clarification

webui/react/Makefile
hamidzr: `>` only redirect `stdout` which is the happy output of a command. if the command has any warnings or errors to output it will not get redirected by `>`

prettier output one line per file at all times whereas the other commands here don't 

the suggestion here is to limit the output when everything is formatted, as the output doesn't mean much, while not affecting any warnings or errors

keita-determined (approved):
LGTM
btw, did you need to format some code which prettier can't handle? If so, that would be nice to have it in the PR

webui/react/Makefile
hamidzr: thanks. I kept the changes out of `npm run ...` commands. if you run those directly you should still get all the output.

hamidzr:
> LGTM
btw, did you need to format some code which prettier can't handle? If so, that would be nice to have it in the PR

I think if there was any ci would have already complained about. I'll run it locally to check

PR #5100, opened 2022-09-22
perf: resolve get_experiment_checkpoints slowness

trentwatt:
## Description

[DET-8455]

current get checkpoints for experiment does 

```
SELECT *
FROM proto_checkpoints_view c
WHERE CAST(c.training->>'experiment_id' AS integer) = $1
ORDER BY c.report_time DESC
```

`proto_checkpoints_view` looks like

```
 SELECT c.uuid::text AS uuid,
    ...
    jsonb_build_object('trial_id', c.trial_id, 'experiment_id' ... )
   FROM checkpoints_view c; 
```

burying the `experiment_id` within the `json_build_object` prevents query planner from filtering by `experiment_id` before doing the join here: 

```
   FROM checkpoints_v2 c
     LEFT JOIN trials t ON c.task_id = t.task_id
     LEFT JOIN experiments e ON t.experiment_id = e.id
```

resulting in scan of checkpoints.

## Test Plan

for correctness: 
- I manually tested all the variations from the swagger docs (`/docs/rest-api`)

performance testing:
- set bundebug to  verbose
- copy the queries
- `gcloud sql connect determined-release-party`
- paste the queries

## Commentary (optional)

further discussion in code comments

## Checklist

[DET-8455]: https://determinedai.atlassian.net/browse/DET-8455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/static/migrations/20220921224032_checkpoint-query-perf.tx.up.sql
```diff
@@ -0,0 +1,16 @@
...
+       'STATE_'::text || c.state             AS state,
+       c.resources,
+       c.metadata,
+       jsonb_build_object('trial_id', c.trial_id, 'experiment_id', c.experiment_id, 'experiment_config',
+                          c.experiment_config, 'hparams', c.hparams, 'training_metrics',
+                          jsonb_build_object('avg_metrics', c.training_metrics -> 'avg_metrics'::text, 'batch_metrics',
+                                             c.training_metrics -> 'batch_metrics'::text), 'validation_metrics',
+                          json_build_object('avg_metrics', c.validation_metrics), 'searcher_metric',
+                          c.searcher_metric) AS training,
+        c.experiment_id
```
trentwatt, line 15: putting this here allows for near-instanteous queries, but results in 

```"error fetching checkpoints for experiment 48 from database: error running query: get_checkpoints_for_experiment: error converting row to Protobuf struct: proto: (line 1:62): unknown field \"experiment_id\""```

File: master/static/srv/get_checkpoints_for_experiment.sql [hunk 2]
```diff
@@ -1,4 +1,15 @@
+-- option 1: without migration -- slower 1.5s versus 5s in the rc
+-- if we can make it so that proto_checkpoints_view does not cast the uuid
+-- as a text, the where clause should be able to use the index and be much faster.
```
trentwatt, line 3: er maybe scratch the bit about using index

master/internal/api_experiment.go
```diff
@@ -936,6 +937,33 @@ func (a *apiServer) PatchExperiment(
...
+		AllocationId: c.AllocationId,
+		Uuid:         c.Uuid,
+		ReportTime:   timestamppb.New(c.ReportTime),
+		Resources:    c.Resources,
+		Metadata:     c.Metadata,
+		State:        checkpointv1.State(checkpointv1.State_value[c.State]),
+		Training:     c.Training,
+	}
+}
+
```
trentwatt, line 966: these should probably go somewhere else. not sure if this PR is big enough to warrant starting the modularization process though.

master/internal/api_experiment.go
```diff
@@ -957,67 +985,71 @@ func (a *apiServer) GetExperimentCheckpoints(
...
+		orderColumn = "uuid"
+	case apiv1.GetExperimentCheckpointsRequest_SORT_BY_END_TIME:
+		orderColumn = "report_time"
+	case apiv1.GetExperimentCheckpointsRequest_SORT_BY_STATE:
+		orderColumn = "state_enum"
+	case apiv1.GetExperimentCheckpointsRequest_SORT_BY_SEARCHER_METRIC:
+		orderColumn = "(training->>'searcher_metric')::float8"
+	case apiv1.GetExperimentCheckpointsRequest_SORT_BY_TRIAL_ID:
+	case apiv1.GetExperimentCheckpointsRequest_SORT_BY_UNSPECIFIED:
+	default:
```
trentwatt, line 1019: is there a more idiomatic way to say "leave as the default"

master/internal/api_experiment.go
```diff
@@ -957,67 +985,71 @@ func (a *apiServer) GetExperimentCheckpoints(
...
-		ai, aj := resp.Checkpoints[i], resp.Checkpoints[j]
-		if useSearcherSortBy {
-			if order, done := protoless.CheckpointSearcherMetricNullsLast(ai, aj); done {
-				return order
-			}
-		}
+	switch {
+	case errors.Cause(err) == sql.ErrNoRows || err == db.ErrNotFound:
+		return nil, status.Errorf(
+			codes.NotFound, "no checkpoints found for experiment %d", req.Id)
```
trentwatt, line 1043: added `sql.ErrNoRows` based on [this conversation](https://github.com/determined-ai/determined/pull/4976#discussion_r966027820)

master/internal/api_experiment.go
```diff
@@ -957,67 +985,71 @@ func (a *apiServer) GetExperimentCheckpoints(
...
-			return ai.State.Number() < aj.State.Number()
-		case apiv1.GetExperimentCheckpointsRequest_SORT_BY_SEARCHER_METRIC:
-			return protoless.CheckpointSearcherMetricLess(ai, aj)
-		case apiv1.GetExperimentCheckpointsRequest_SORT_BY_UNSPECIFIED:
-			fallthrough
-		default:
-			return protoless.CheckpointTrialIDLess(ai, aj)
-		}
-	})
-	return resp, a.paginate(&resp.Pagination, &resp.Checkpoints, req.Offset, req.Limit)
```
trentwatt, line 1020: no more `a.paginate` means no more `resp.Pagination`. IMO this is a good thing though-- in general not worth it to return a total count.

master/static/migrations/20220921224032_checkpoint-query-perf.tx.up.sql
```diff
@@ -0,0 +1,19 @@
...
+       c.metadata,
+       jsonb_build_object('trial_id', c.trial_id, 'experiment_id', c.experiment_id, 'experiment_config',
+                          c.experiment_config, 'hparams', c.hparams, 'training_metrics',
+                          jsonb_build_object('avg_metrics', c.training_metrics -> 'avg_metrics'::text, 'batch_metrics',
+                                             c.training_metrics -> 'batch_metrics'::text), 'validation_metrics',
+                          json_build_object('avg_metrics', c.validation_metrics), 'searcher_metric',
+                          c.searcher_metric) AS training,
+        c.experiment_id,
+        c.trial_id,
+        c.state AS state_enum
```
trentwatt, line 17: this is just `proto_checkpoints_view` but with these three columns added. 
- `experiment_id` and `trial_id` so that we can filter before joining.
- `state_enum` so that we can filter by enum order within the db

proto/src/determined/checkpoint/v1/checkpoint.proto
```diff
@@ -42,7 +41,7 @@ message CheckpointTrainingMetadata {
   determined.common.v1.Metrics validation_metrics = 6;
   // Searcher metric (as specified by the expconf) at the same steps_completed
   // of the checkpoint.
-  google.protobuf.DoubleValue searcher_metric = 17;
+  optional double searcher_metric = 17;
```
trentwatt, line 44: https://determined-ai.slack.com/archives/C02PV33GSN5/p1664599654758959?thread_ts=1664572966.709299&cid=C02PV33GSN5

master/internal/api_experiment.go
```diff
@@ -957,67 +985,74 @@ func (a *apiServer) GetExperimentCheckpoints(
...
-			if state == v.State {
-				found = true
-				break
-			}
-		}
+	if len(req.States) > 0 {
+		states := []string{}
+		for _, s := range req.States {
+			state_text := checkpointv1.State_name[int32(s.Number())]
+			states = append(states, state_text)
```
stoksc, line 1002: I think `state_text` with be a name like "STATE_ACTIVE" when the DB state is called just "ACTIVE" so this may need another layer of mapping.

stoksc: oh ok, i see in the query this is accounted for, though I think "state_unspecified" may explode? bit unfortunate, the distance between this logic.

master/internal/api_experiment.go
```diff
@@ -957,67 +985,74 @@ func (a *apiServer) GetExperimentCheckpoints(
...
-			fallthrough
-		default:
-			return protoless.CheckpointTrialIDLess(ai, aj)
-		}
-	})
-	return resp, a.paginate(&resp.Pagination, &resp.Checkpoints, req.Offset, req.Limit)
+	err = q.Scan(context.TODO(), &checkpoints)
+
+	switch {
+	case errors.Cause(err) == sql.ErrNoRows || err == db.ErrNotFound:
```
stoksc, line 1044: i usually just us `errors.Is(err, sql.ErrNoRows)`, but I guess cause is fine since it would definitely be the root error.

stoksc (commented):
few questions but mostly lgtm. meta question though: is this view hierarchy getting hard to reason about? could we do we some simplification?

master/internal/api_experiment.go
stoksc: i think clarity-wise, this is probably the reason i prefer a query in bun right here to logic in a view.

PR #5102, opened 2022-09-22
fix: fix useSettings bug

EmilyBonar:
## Description
Correcting a bug I introduced in useSettings

## Test Plan
Go to Uncategorized Experiments page. You should be able to 

- toggle Show Archived switch. 
- switch pages
- sort the table by any sortable column
- add or remove columns using the Customize Columns modal (accessed through Columns button in top right)
- filter by tags, state, or user

All of these changes should persist on refresh.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useSettings.ts
+++ b/webui/react/src/hooks/useSettings.ts
@@ -161,7 +161,10 @@ export const queryToSettings = <T>(config: SettingsConfig, query: string): T => 
        * Example - 'PULLING' => [ 'PULLING' ]
        */
       const normalizedValue =
-        (prop.type.isArray && queryValue !== null && queryValue !== undefined && !Array.isArray(queryValue))
+        prop.type.isArray &&
+        queryValue !== null &&
+        queryValue !== undefined &&
+        !Array.isArray(queryValue)
           ? [queryValue]
           : queryValue;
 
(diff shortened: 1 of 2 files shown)


ClaireNeveu (commented):
The way `useSettings` is written makes it hard to judge whether these changes are correct so I really have to go off of your test plan. Unfortunately I'm unfamiliar with the broad scope associated with `useSettings`. If you could have someone more familiar with it validate your test plan that is good enough for me.

keita-determined:
In my understanding, `useSettings` has 3 main features. 
1. store the user data in localStorage, and database. Every time user change data, LS, and DB should be updated too
2. If `skipUrlEncoding` is false or not declared, user data should be in url params
3. If there are url params in the url, these data should be override the data in LS or DB temporarily, but not update data in DB

webui/react/src/hooks/useSettings.ts
```diff
@@ -274,14 +277,14 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
           acc.querySettings[key] = undefined;
 
           // If the settings value is invalid, set to undefined.
-          const value = partialSettings[key];
```
keita-determined, line 277: `partialSettings[key] !== null` should be `partialSettings[key] !== undefined` or `partialSettings[key]`?

EmilyBonar: I'm doing this (`partialSettings[key] !== null`) to avoid a typescript error.

trentwatt (approved):
still seeing this one 

https://user-images.githubusercontent.com/28383080/191815869-79e65456-3722-4e50-b4f4-d60f44b6ef2c.mov

overall an improvement though ¯\_(ツ)_/¯

PR #5103, opened 2022-09-22
ci: fix "previous tag" logic

dannysauer:
## Description

The agent and helm releases were failing to get the previous tag completely (we don't use a `v` prefix any longer) Also adjust the grep to include only what we want instead of excluding known unwanted tags.  The previous tag functionality doesn't make much of a difference, but the exclusion was brittle; any new tag which happened to get added would break the exclusion logic.

~Marking as draft since there's a conflicting helm Makefile update in-flight; this will need updated to reflect that change.~

## Test Plan

Run shell commands in a shell and confirm expected behavior.  Pay attention on next release to make sure things actually work as planned. :)

## Commentary (optional)

It's tough to suppress my urge to also convert the `grep "$VERSION" -A1 | sed -n '2 p'` into just `sed -n "/$VERSION/{n;p;}"`, and it's a real bummer that git pattern matching is limited to `fnmatch()` without the GNU extended patterns enabled, but this all gives me an opportunity to practice being calm. :lotus_position: 

Since these three all use the same logic, it might actually make sense for that one-liner to be a helper script in a parent directory instead of being duplicated across multiple files. :shrug: 

## Checklist

### DIFF
--- a/agent/Makefile
+++ b/agent/Makefile
@@ -58,7 +58,7 @@ package: packaging/LICENSE
 
 .PHONY: release
 release: export GORELEASER_CURRENT_TAG := $(VERSION)
-release: export GORELEASER_PREVIOUS_TAG := $(shell git tag --sort=-creatordate | grep -v "rc" | grep "v$(VERSION)" -A1 | sed -n '2 p')
+release: export GORELEASER_PREVIOUS_TAG := $(shell git tag --sort=-creatordate | egrep '^[0-9.]+$' | grep "$(VERSION)" -A1 | sed -n '2 p')
 release: packaging/LICENSE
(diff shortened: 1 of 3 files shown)


asnelling (changes requested):
LGTM
`git tag ... sed -n '1 p'` works

dannysauer:
Corresponding EE PR, @asnelling 
determined-ai/determined-ee#524

It'd be better if we just set a `$suffix` to en empty string or `-ee` depending on the repo.  But there are a couple other Makefile differences (the docker publish name differs), so maybe we unify Makefiles later.

asnelling (approved):

PR #5104, opened 2022-09-22
feat: Reconnect to Slurm jobs on startup (FOUNDENG-215)

jerryharrow:
Merge common portions of EE PR #429 to OSS to eliminate merging issues.

We previously terminated running jobs upon a master restart. Now that Determined core supports re-attaching to jobs, do the same for DispatchRM.

- Change configure to indiate that DispatcherRM supports reattach
- Handle allocation messages with Restore:true
- Fail any allocations on Restore:true if the Dispatch ID is missing.
- Handle the case were we no longer have the payload name
  which was lost in the restart.   Ask the launcher for it in the
  very rare case where the job was started but fails before the
  rendezvous and we need the payload name to retrieve the logs.
- When in debug mode, defer dispatch cleanup util the next restart.
  On restart terminate all dispatches.

## Checklist


master/internal/config/config.go
```diff
@@ -400,33 +400,12 @@ func ReadWeight(rpName string, jobConf interface{}) float64 {
 // IsAgentRMReattachEnabled returns whether the container reattachment is enabled on AgentRM.
 func IsAgentRMReattachEnabled() bool {
```
stoksc, line 401: can you comment why this one still has to exist. (i will fix it later if you put TODO on it)

master/internal/rm/agent_resource_manager.go
```diff
@@ -71,6 +71,23 @@ func (a AgentResourceManager) ValidateResourcePool(ctx actor.Messenger, name str
 	return err
 }
 
+// IsReattachEnabled is true for the AgentResourceManager.
+func (a AgentResourceManager) IsReattachEnabled(ctx actor.Messenger) bool {
+	return true
+}
```
stoksc, line 77:
```suggestion
func (a AgentResourceManager) IsReattachEnabled(ctx actor.Messenger) bool {
	config := config.GetMasterConfig()

	for _, rpConfig := range config.ResourcePools {
		if rpConfig.AgentReattachEnabled {
			return true
		}
	}
	return false
}
```

master/internal/trial.go
```diff
@@ -663,17 +662,21 @@ func mustParseTrialRunID(child *actor.Ref) int {
...
 	var allocations []model.Allocation
-	err := db.Bun().NewSelect().Model(&allocations).
+	selectQuery := db.Bun().NewSelect().Model(&allocations).
 		Where("task_id = ?", t.taskID).
-		Where("start_time IS NOT NULL").
-		Where("end_time IS NULL").
-		Scan(context.TODO())
+		Where("end_time IS NULL")
+
+	if t.rm.IsReattachableOnlyAfterStarted(ctx) {
```
stoksc, line 674: one thing we have to take into account here (think about at least) is what happens if the slurm job gets scheduled/transitions to running while we are away. are we going to miss the resources running message? (i can't remember)

stoksc (approved):
lgtm modulo final comments

master/internal/trial.go
jerryharrow: We get Slurm states via polling the job status, not from state in the DispatcherRM.     So we can't miss RUNNING, unless it has already FAILED/COMPLETED by the time we re-attach, and in that case we don't care.

In any case send start_time never gets set for Slurm jobs so it isn't going to help us :-)

master/internal/config/config.go
jerryharrow: ack

master/internal/rm/agent_resource_manager.go
stoksc: > I don't understand why the concept of agent being able to potentially Reattach necessarily has to align with if any resource_pool is configured to do so. 

It is 100% related, an agent is potentially able to reattach if and only if a resource pool supports it. I will refactor this eventually, I suppose.

> The place where IsReattachEnabled is called, is part of trial/restore and immediately followed by IsReattachEnabledForRP for the give partition, so we are just duplicating the check

seems like trial should just call `IsReattachEnabledForRP` then

PR #5105, opened 2022-09-22
feat: RBAC backend integration at user management

gt2345:
## Description

Integration backend API to user management page.
- view user role/permissions
- add/remove use
Screen Shot 2022-09-22 at 3 44 12 PM
r roles

## Test Plan

Log in as admin, navigates to `/det/settings/user-management?f_rbac=on` and view/add/remove roles for users.
Note that backend has to point to determined-ee

## Checklist

### DIFF
--- a/webui/react/src/components/Navigation.tsx
+++ b/webui/react/src/components/Navigation.tsx
@@ -6,6 +6,7 @@ import {
   useFetchKnownRoles,
   useFetchPinnedWorkspaces,
   useFetchResourcePools,
+  useFetchUserRoles,
   useFetchUserSettings,
 } from 'hooks/useFetch';
 import usePolling from 'hooks/usePolling';
@@ -27,18 +28,24 @@ const Navigation: React.FC<Props> = ({ children }) => {
   const fetchResourcePools = useFetchResourcePools(canceler);
   const fetchPinnedWorkspaces = useFetchPinnedWorkspaces(canceler);
   const fetchUserSettings = useFetchUserSettings(canceler);
+  const fetchUserRoles = useFetchUserRoles(canceler);
   const fetchKnownRoles = useFetchKnownRoles(canceler);
 
   usePolling(fetchAgents);
   usePolling(fetchPinnedWorkspaces);
   usePolling(fetchUserSettings, { interval: 60000 });
-  usePolling(fetchKnownRoles);
+  usePolling(fetchUserRoles, { interval: 60000 });
 
   useEffect(() => {
     fetchResourcePools();
 
     return () => canceler.abort();
   }, [canceler, fetchResourcePools]);
+
+  useEffect(() => {
+    fetchKnownRoles();
+    return () => canceler.abort();
+  }, [canceler, fetchKnownRoles]);
 
   return (
     <Spinner spinning={ui.showSpinner}>
--- a/webui/react/src/components/Navigation.tsx
+++ b/webui/react/src/components/Navigation.tsx
@@ -32,17 +32,13 @@ const Navigation: React.FC<Props> = ({ children }) => {
   usePolling(fetchAgents);
   usePolling(fetchPinnedWorkspaces);
   usePolling(fetchUserSettings, { interval: 60000 });
+  usePolling(fetchKnownRoles);
 
   useEffect(() => {
     fetchResourcePools();
 
     return () => canceler.abort();
   }, [canceler, fetchResourcePools]);
-
-  useEffect(() => {
-    fetchKnownRoles();
-    return () => canceler.abort();
-  }, [canceler, fetchKnownRoles]);
 
   return (
     <Spinner spinning={ui.showSpinner}>
--- a/webui/react/src/contexts/Store.tsx
+++ b/webui/react/src/contexts/Store.tsx
@@ -1,7 +1,8 @@
 import React, { Dispatch, useContext, useReducer } from 'react';
 
 import { globalStorage } from 'globalStorage';
-import { V1Role, V1RoleAssignmentSummary, V1UserWebSetting} from 'services/api-ts-sdk';
+import { V1Role, V1RoleAssignmentSummary, V1UserWebSetting } from 'services/api-ts-sdk';
+import * as decoder from 'services/decoder';
 import { ActionUI, initUI, reducerUI, StateUI } from 'shared/contexts/UIStore';
 import { clone, isEqual } from 'shared/utils/data';
 import rootLogger from 'shared/utils/Logger';
@@ -21,7 +22,6 @@ import {
   Workspace,
 } from 'types';
 import { getCookie, setCookie } from 'utils/browser';
-import * as decoder from 'services/decoder';
 
 const logger = rootLogger.extend('store');
 
@@ -159,7 +159,6 @@ const initInfo: DeterminedInfo = {
 };
 
 export const OSSUserRole: V1Role = {
-  roleId: -1,
   name: 'OSS User',
   permissions: [
     {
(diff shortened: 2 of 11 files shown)


webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -45,18 +53,43 @@ interface FormValues {
 }
 
 const ModalForm: React.FC<Props> = ({ form, user, groups, viewOnly }) => {
-  const [permissions, setPermissions] = useState<Permission[]>([]);
+  const [permissions, setPermissions] = useState<UserRole[]>([]);
```
gt2345, line 56: We need to store role information because deletion only happen at role level

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -76,28 +109,54 @@ const ModalForm: React.FC<Props> = ({ form, user, groups, viewOnly }) => {
...
-        render: (val: boolean) => (val ? <Icon name="checkmark" /> : ''),
-        title: 'Workspaces?',
+        dataIndex: 'permissions',
+        key: 'permissions',
+        render: (p) => p.length,
+        title: 'Permissions',
       },
-    ];
-    if (canModifyPermissions && !viewOnly) {
+    ] as ColumnDef<UserRole>[];
```
ClaireNeveu, line 117: Why are you casting here?

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -138,21 +197,29 @@ const ModalForm: React.FC<Props> = ({ form, user, groups, viewOnly }) => {
...
           <Select mode="multiple" optionFilterProp="children" placeholder={'Add Roles'} showSearch>
-            {knownRoles.map((r) => (
-              <Select.Option key={r.id} value={r.id}>
-                {r.name}
-              </Select.Option>
-            ))}
+            {knownRoles
+              .filter(
+                (r) =>
+                  r.permissions.find((p) => p.isGlobal) &&
```
ClaireNeveu, line 206: From this page you should also be able to manage non-global roles for a user.

ClaireNeveu (changes requested):

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
gt2345: This is to enforce a type on column object, to avoid unexpected field

webui/react/src/components/Navigation.tsx
```diff
@@ -32,18 +32,14 @@ const Navigation: React.FC<Props> = ({ children }) => {
   usePolling(fetchAgents);
   usePolling(fetchPinnedWorkspaces);
   usePolling(fetchUserSettings, { interval: 60000 });
+  usePolling(fetchKnownRoles);
```
mapmeld, line 35: How often to fetch the list of available roles?  I'd recommend a longer interval (like the previous line).

I started thinking about this in #5079 - most users do not use the roles data -- only admins visiting RBAC UI. But usually we have `usePolling` here with no restriction.  I don't know if there is a way we can poll only for certain users?

webui/react/src/services/types.ts
```diff
@@ -373,10 +373,22 @@ export interface AssignRolesToGroupParams {
...
+  userId: number;
+}
+
+export interface ListRolesParams extends PaginationParams {
   isGlobal?: boolean;
-  limit?: number;
-  offset?: number;
+}
+
+export interface ListWorlspaceRolesParams extends PaginationParams {
```
mapmeld, line 390: fix typo in `ListWorlspaceRolesParams` here and in api.ts / apiConfig.ts

webui/react/src/services/apiConfig.ts
```diff
@@ -313,6 +316,40 @@ export const assignRolesToGroup: DetApi<
...
+  request: (params) =>
+    detApi.RBAC.assignRoles({
+      userRoleAssignments: params.roleIds.map((roleId) => ({
+        roleAssignment: { role: { roleId } },
+        userId: params.userId,
+      })),
+    }),
+};
+
+export const removeRoleFromUser: DetApi<
```
julian-determined-ai, line 335: @gt2345 As part of https://determinedai.atlassian.net/browse/DET-8219 I also added created assign and remove role functions but did not scope them to users or groups. For example, here I just added a `removeRole` that will accept `groupAssignments` or `userAssignments` do you have any thoughts on keeping a generic function or having separate `removeRoleFromUser` and `removeRoleFromGroup` functions? 

I actually do like the separate functionality now that I am seeing it so I will plan to  instead implement a separate  `removeRoleFromGroup` function.

webui/react/src/components/Navigation.tsx
gt2345: You are right, list of available roles do not need to be fetched frequently, it is the user's roles/assignments need `usePolling`

webui/react/src/services/types.ts
gt2345: 😅

webui/react/src/services/apiConfig.ts
gt2345: Cool, thank you for letting me know

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
ClaireNeveu: You're not actually _enforcing_ something when you do a type cast, you're _removing_ a guarantee. What's the unexpected field?

ClaireNeveu: I think what you want here is probably `const columns: ColumnDef<UserRole>[] = ...`

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
ClaireNeveu: Thinking on this, I think you should be able to _see_ all the roles a user has from this page but it's okay if you can't add or remove workspace specific roles. It's important to have a page where all of a user's roles can be seen.

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
gt2345: Got it, thanks!

webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
gt2345: This page can see all the roles assigned to a user, however, for workspace specific roles, this page cannot see the associated workspace... For example, suppose an user has `workspaceAdmin` role scoped to workspace 1 and 2, this page can only see this user has `workspaceAdmin` role, and that's it

webui/react/src/components/Navigation.tsx
gt2345: I added fetch for current user's role in usePolling, also only call API when rbac is enabled

webui/react/src/contexts/Store.tsx
```diff
@@ -157,6 +158,23 @@ const initInfo: DeterminedInfo = {
   version: process.env.VERSION || '',
 };
 
+export const OSSUserRole: V1Role = {
+  name: 'OSS User',
+  permissions: [
+    {
+      id: -1,
+      isGlobal: true,
+      name: 'oss_user',
```
mapmeld, line 167: I wonder if we still expect a name attribute on Permissions, now that the id is changing from a number to the V1PermissionType enum / string. If everything works we can put it in a future pr/ticket but just noting this.

PR #5112, opened 2022-09-23
fix: prevent reattach deadlock upon container reattach

NicholasBlaskey:
## Description

Actor deadlock bug caused by 

1. the container manager is blocking pinging the container actor
2. the container actor is blocking pinging the docker actor
3. the docker actor is stuck forever waiting on the (shell) docker container to be killed in reattachContainer https://github.com/determined-ai/determined/blob/db3c53b39b6c1afe3b36421d18f84e93e7222c21/agent/internal/docker.go#L318 (you can confirm this by just killing the shell through docker and the experiment will be scheduled)

## Test Plan

Create a devcluster with an agent with at least 2 ```artifical_slots``` with a clean database

```det shell start``` and wait for shell to start

Control-C and then restart ```devcluster```

Run a single slot experiment and you should see it schedule

Then verify ```det shell kill $shell_id``` terminates the shell in ```det shell```

## Checklist


stoksc (approved):
lgtm, but following the pattern of the rest of the code, should we just `go d.reattachContainer()` from prestart?

NicholasBlaskey:
> lgtm, but following the pattern of the rest of the code, should we just `go d.reattachContainer()` from prestart?

Just ```go d.reattachContainer()``` from prestart I don't think is sufficient since in the case of two containers being restored, the first being restored will block the second one from getting restored 

If we ```"go"``` both of them I am unsure if that is ok. I am wondering if there is potential for ```signalContainer``` to come at an awkward time against a container trying to reattach? I think it might be fine either way and I'm assuming the upcoming refactor of the agent will make this irrelevant.

NicholasBlaskey:
> and this change will effectively roll it back. it's be easier to not ping the actor after pre-start than to do a goroutine to achieve the same.

I don't believe this change rolls that race fix back since ```containerReattached``` is still sent outside of the goroutine of the waiting on container exit on line 364.

> Q: why are we stuck waiting for the killed container? isn't the check in line 357 supposed to cover killed containers?

We aren't stuck waiting for containers that are killed. We are stuck waiting on containers which are running to exit. Which in the shell case in the bug report is potentially forever.

PR #5113, opened 2022-09-23
feat: Add "Members" view to workspace page [DET-8219]

julian-determined-ai:
## Description

feat: Add "Members" view to workspace page [DET-8219]

A Workspace `Members` view has been added where users or groups can be added to a workspace, removed from a workspace, or have their role in the workspace changed.

a `mock_workspace_members` feature flag has been added that will add mocked workspace groups, roles, and users to the view. 

Members Tab with `f_mock_workspace_members=on`
Screen Shot 2022-09-23 at 5 16 47 PM

Add Members Modal with `f_mock_workspace_members=on`
Screen Shot 2022-09-23 at 5 00 06 PM

Remove Member Modal
Screen Shot 2022-09-23 at 5 18 13 PM

## Test Plan
TBD - Dependent on the API's and data available in depth testing will require mocked data to be generated.

## Commentary (optional)

~~The page may be difficult to test due to the lack of actual role and workspace member data, I will try to mock data needed to test the page.~~

## Checklist

[DET-8219]: https://determinedai.atlassian.net/browse/DET-8219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/utils/user.ts
```diff
@@ -8,3 +9,25 @@ interface UserNameFields {
...
+}
+
+export const getIdFromUserOrGroup = (userOrGroup: UserOrGroup): number => {
+  if (isUser(userOrGroup)) {
+    const user = userOrGroup as User;
+    return user.id;
+  }
+  const group = userOrGroup as V1Group;
+
+  // The groupId should always exist
```
julian-determined-ai, line 31: Do we expect any groups to exist without a groupId?

webui/react/src/pages/WorkspaceDetails/WorkspaceMembers.tsx
```diff
@@ -0,0 +1,266 @@
...
+          onSelect={async (value: RawValueType | LabelInValueType) => {
+            const roleIdValue = value as number;
+            const userOrGroupId = getIdFromUserOrGroup(record);
+
+            // Needs to be updated to get the correct old role for the user
+            // or group.
+            const oldRoleId = assignments?.[0].role?.roleId || 0;
+
+            // Remove the old role then add the new role
+            if (isUser(record)) {
```
julian-determined-ai, line 175: Wanted to confirm the logic here is correct? For the dropdown in the table in https://www.figma.com/file/kqWThAhkfa4giVOwHzrAxA/RBAC-Web?node-id=39%3A5166

We will need to remove the old `role` and add the new `role` correct? There is not a single API call to do so?

webui/react/src/pages/WorkspaceDetails/WorkspaceMembers.tsx
```diff
@@ -0,0 +1,301 @@
...
+          </div>
+        </>
+      );
+    };
+
+    const roleRenderer = (value: string, record: UserOrGroup) => {
+      return (
+        <Select
+          className={css.selectContainer}
+          disabled={!userCanAssignRoles}
```
julian-determined-ai, line 195: Need clarification if a single assignment will be returned from https://github.com/determined-ai/determined-ee/pull/444/files#diff-2507f8d47389cccf4a093c61fdaf6998acb68ef55b5b41476d298617c51d4f71R75 or if I will need to filter the `assignments` returned per group and user to determine the correct role to show in the table.

webui/react/src/utils/user.ts
mapmeld: I agree it should always exist. My experience with `V1Role`: backend didn't make any attribute a required type in the proto annotation.  First I made all attributes required... but it was frustrating because when I added/removed v1Role, it expected all fields. So I made only roleId required. So in this case I would make groupId required and rebuild proto and bindings.

webui/react/src/pages/WorkspaceDetails/WorkspaceMembers.tsx
mapmeld: the PR changed; but based on what I said earlier, I believe that you'd get an array / multiple roles, but as long as it's just the pre-canned/preset roles, we can assume there's only zero or one workspace-specific role

webui/react/src/pages/WorkspaceDetails/WorkspaceMembers.tsx
```diff
@@ -0,0 +1,305 @@
...
+  } = useModalWorkspaceRemoveMember({
+    name,
+    userOrGroup: userOrGroup,
+    userOrGroupId: getIdFromUserOrGroup(userOrGroup),
+  });
+
+  const menuItems = (
+    <Menu>
+      <Menu.Item danger key="delete" onClick={() => openWorkspaceRemoveMemberModal()}>
+        Delete
```
mapmeld, line 66: I think you did a good job explaining the action in the confirm modal, but I think it's easier if this was labeled Remove instead.

webui/react/src/pages/WorkspaceDetails.tsx
```diff
@@ -85,304 +71,84 @@ const WorkspaceDetails: React.FC = () => {
...
+  const addableGroups: V1Group[] = groups
+    ? groups
+        ?.filter((groupDetails) => {
+          groupDetails.group?.groupId &&
+            !groupsAssignedDirectlyIds?.has(groupDetails.group.groupId);
+        })
+        .map((groupDetails) => groupDetails.group)
+    : [];
+  const addableUsers = users.filter((user) => !usersAssignedDirectlyIds?.has(user.id));
+  const addableUsersAndGroups = [...addableGroups, ...addableUsers];
```
mapmeld, line 151: :+1:

mapmeld (approved):
Thanks for making changes. LGTM

PR #5114, opened 2022-09-23
feat: New Project button/message checks permission [DET-8374]

mapmeld:
## Description

Completes project-related RBAC permissions with the "create_project" permission. If someone does not have this permission the "New Project" button is hidden, and the message on an empty workspace changes.

## Test Plan

- Create an empty workspace, and you should have a New Project button, and the message 'Create a project with the "New Project" button or in the CLI.'
- Viewing the workspace with `f_rbac=on&f_mock_permissions_read=on` hides the 'New Project' button, and the message reads 'User cannot create a project in this workspace'
- With `f_rbac=on&f_mock_permissions_all=on`, you should see the 'New Project' button and the original message

## Checklist

### DIFF
--- a/webui/react/src/hooks/usePermissions.ts
+++ b/webui/react/src/hooks/usePermissions.ts
@@ -1,4 +1,5 @@
 import { useStore } from 'contexts/Store';
+import useFeature from 'hooks/useFeature';
 import {
   DetailedUser,
   ExperimentPermissionsArgs,
@@ -24,6 +25,13 @@ interface ModelVersionPermissionsArgs {
 interface ProjectPermissionsArgs {
   project?: Project;
   workspace?: PermissionWorkspace;
+}
+
+interface RbacOptsProps {
+  rbacEnabled: boolean;
+  user?: DetailedUser;
+  userAssignments?: UserAssignment[];
+  userRoles?: UserRole[];
 }
 
 interface PermissionsHook {
@@ -60,6 +68,12 @@ const usePermissions = (): PermissionsHook => {
     userAssignments,
     userRoles,
   } = useStore();
+  const rbacOpts = {
+    rbacEnabled: useFeature().isOn('rbac'),
+    user,
+    userAssignments,
+    userRoles,
+  };
 
   // Determine if the user has access to any workspaces
   // Should be updated to check user assignments and roles once available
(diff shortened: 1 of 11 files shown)


webui/react/src/hooks/usePermissions.ts
```diff
@@ -289,6 +292,15 @@ const canDeleteModelVersion = (
 
 // Project actions
 // Currently the smallest scope is workspace
+const canCreateProject = (
+  workspace?: PermissionWorkspace,
+  userAssignments?: UserAssignment[],
+  userRoles?: UserRole[],
+): boolean => {
+  const permitted = relevantPermissions(userAssignments, userRoles, workspace?.id);
+  return permitted.has('oss_user') || permitted.has('create_project');
```
ClaireNeveu, line 301: Since the API isn't going to return this after all, we should be using the feature switch here.

mapmeld:
The good news - latest commit does `useFeature` to set `rbacEnabled` in place of `oss_user`. We no longer need mock roles and permissions for initialization and mock responses to API calls.

The downside - instead of passing the now-more-verbose RBAC-related params `(rbacEnabled, user, userAssignments, userSettings)` to each function, this PR sends one object, which is known to exist so it has to be the first parameter to the functions.  Without the mocks we need to call `getPermissionsSummary()` and convert it to roles... in total it looks like a big change but it's mostly new parameters and pushing out the mocks.

webui/react/src/hooks/usePermissions.ts
```diff
@@ -141,287 +155,245 @@ const relevantPermissions = (
...
+  rbacEnabled,
+  user,
+  userAssignments,
+  userRoles,
+}: RbacOptsProps): boolean => {
   const permitted = relevantPermissions(userAssignments, userRoles);
   return (
-    !!user &&
-    (permitted.has('oss_user') ? user.isAdmin : permitted.has('PERMISSION_CAN_ADMINISTRATE_USERS'))
+    !!user && (rbacEnabled ? user.isAdmin : permitted.has('PERMISSION_CAN_ADMINISTRATE_USERS'))
```
ClaireNeveu, line 166: I think your conditionals are backwards.

ClaireNeveu:
This looks good but I'd like you to change your test plan. Instead of modifying the code can you test with the feature switch. Also since you ended up modifying the code for all permissions could you do a basic smoke test for those with the feature switch on and then off?

ClaireNeveu (approved):

PR #5116, opened 2022-09-23
build: setup `Pre-Commit`

keita-determined:
## Description

- [DET-8095]
- Setup [pre-commit](https://pre-commit.com/) for entire repo and setup `lint-staged` for web

## How to Setup

- Run `make get-deps-pip`
- Run `make set-config-dev`
- Change file in `webui/react/src` folder
- Commit to run pre-commit

## Test Plan
- Check if all configurations are correct (especially pre-commit)
- Check if pre-commit is triggered correctly when specific files in web folder are staged and about to be committed

## Checklist

[DET-8093]: https://determinedai.atlassian.net/browse/DET-8093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-8095]: https://determinedai.atlassian.net/browse/DET-8095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+  - repo: local
+    hooks:
+      - id: web-lint-check
+        name: Web Lint Check
+        entry: make -C webui/react lint-staged
+        language: system
+        files: 'webui/react/src/'
+        pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,24 @@
 repos:
   - repo: local
     hooks:
-      - id: web-lint-check
-        name: Web Lint Check
-        entry: make -C webui/react lint-staged
+      - id: web-js-lint-check
+        name: Web JS Lint Check
+        entry: bash -c 'make -j$(nproc) -C webui/react check-eslint check-prettier-js'
         language: system
         files: 'webui/react/src/'
+        types_or: [javascript, jsx, ts, tsx]
+        pass_filenames: false
+      - id: web-css-lint-check
+        name: Web CSS Lint Check
+        entry: bash -c 'make -j$(nproc) -C webui/react check-prettier-css check-stylelint'
+        language: system
+        files: 'webui/react/src/'
+        types_or: [css, less, scss]
+        pass_filenames: false
+      - id: web-misc-lint-check
+        name: Web Misc Lint Check
+        entry: bash -c 'make -j$(nproc) -C webui/react check-prettier-misc check-package-lock'
+        language: system
+        files: 'webui/react/src/'
+        types_or: [markdown, json]
         pass_filenames: false
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 all:
 	$(MAKE) get-deps
 	$(MAKE) build
+	$(MAKE) set-config-dev
 
 .PHONY: get-deps
 get-deps: get-deps-pip get-deps-go get-deps-bindings get-deps-webui
@@ -26,12 +27,17 @@ package:
 	$(MAKE) -C agent $@
 	$(MAKE) -C master $@
 
+set-config-dev: .git/hooks/pre-commit
+
+.git/hooks/pre-commit:
+	pre-commit install
+
 .PHONY: build-%
 build-%:
 	$(MAKE) -C $(subst -,/,$*) build
 
 .PHONY: build-docs
-build-docs: build-common build-harness build-cli build-deploy build-model_hub build-examples build-helm build-proto 
+build-docs: build-common build-harness build-cli build-deploy build-model_hub build-examples build-helm build-proto
 	$(MAKE) -C docs build
 
 .PHONY: build-bindings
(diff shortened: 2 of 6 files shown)


ClaireNeveu:
Where is `pre-commit install` being called?

requirements.txt
```diff
@@ -22,6 +22,7 @@ flake8-tuple>=0.4.0
 isort==5.9.3
 mypy==0.910
 bump2version>=1.0.1
+pre-commit>=2.20
```
ClaireNeveu, line 25: We should use `~=` here so we don't silently get a major version bump.

Makefile
```diff
@@ -14,6 +14,7 @@ get-deps-%:
 get-deps-pip:
 	pip install torch==1.9.0
 	pip install -r requirements.txt
+	pre-commit init
```
keita-determined, line 17: Im not very familiar with makefile. I believe these commands are executed in serial, right?

ClaireNeveu: We shouldn't do this in the dependency resolution step. You'll want to add a new step that operates on the pre-commit config file so that it only runs when that file changes and add that to the main process.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
+repos:
+  - repo: local
+    hooks:
+      - id: web-js-lint-check
+        name: Web JS Lint Check
+        entry: bash -c 'make -j$(nproc) -C webui/react check-eslint check-prettier-js'
```
keita-determined, line 6: Added `nproc` to make pre-commit check faster.
On my mac (10 core CPU), 12s without `-j$(nproc)` option. 8s with `-j$(nproc)` option.
I saw an article that says j option with `nproc + 1` is the fastest but we dont need too much optimization and it was only 0.2s faster

keita-determined:
I realized that `pre-commit` and `lint-staged` are overlapped features and `lint-staged` is unnecessary, so i removed `lint-staged`, and added more hooks in `pre-commit`. I think it's reasonable to do it, but if we don't agree with that, I can revert the commit

.pre-commit-config.yaml
dannysauer: The n+1 is faster because it basically keeps a job in the queue. That's from the days when 2-4 CPUs was common, so with modern 10+ core systems, it might make sense to be more... Or maybe less with super fast disks. 🤷‍♂️

All that to say I'd support just leaving it at nprocs.

ioga: make only runs two jobs here, `check-eslint` and `check-prettier-js` which have no deps. not much to parallelize. I suspect nproc > 2 doesn't do anything.

ioga (commented):
Q: does it still run when commit has no changes in `webui/react/src/`?

keita-determined:
> does it still run when commit has no changes in webui/react/src/?

`pre-commit` checks but if there's no change, `pre-commit` skip the checks like the following

image

.pre-commit-config.yaml
keita-determined: I think so too, but the reason why I put `nproc` instead of a constant number is that if we add more commands in the future, we need to increment the number but I think we might forget to do it. 
I don't mind to switch to a constant number tho. Should I leave it or change it?

ioga: keep nproc, it is good, I'd even support `nproc + 1` because it's canonical. 
I'm just pointing this out to highlight that benchmarking nproc=10 against nproc=11 may not be meaningful.

dannysauer:
You can theoretically force a re-check with `precommit run --all-files` or `precommit run --files ./some/path`, should that be something which is relevant for a specific directory.  In that case, typically one would likely have written their own check which knows, for example, two files should always be in sync and thus checks the unchanged file when the twin changes.  (coughs in version file)

dannysauer (commented):
Why is `webui/react/package-lock.json` changed in this PR? :)

Aside from that, I'm good with this.

keita-determined:
> Why is webui/react/package-lock.json changed in this PR? :)
>
> Aside from that, I'm good with this.

`npm uninstall lint-staged` didn't remove all changes. I'll fix it

ClaireNeveu (approved):
This looks good although we should note that DET-8093 is not actually covered by this PR

keita-determined:
> This looks good although we should note that DET-8093 is not actually covered by this PR

I'll untag the lint-staged

ClaireNeveu:
Actually, after reading the docs and testing it, pre-commit will only run against staged files. The only issue is that it doesn't seem to pass the staged files to the command, instead stashing the unstaged ones which makes the command take longer. I wonder if there's an alternate mode there.

keita-determined:
> Actually, after reading the docs and testing it, pre-commit will only run against staged files. The only issue is that it doesn't seem to pass the staged files to the command, instead stashing the unstaged ones which makes the command take longer. I wonder if there's an alternate mode there.

Yes, if we wanna run staged file, we need to use `pre-commit` offered libraries. I didn't wanna rely on that cuz some of them are made by individual and didn't seem reliable compared to official libs. 
I also considered that if there are some options to pass staged file in `pre-commit` but seems like there's no such things so far. We can achieve that using bash commands but that would make yaml file messy.

PR #5120, opened 2022-09-23
feat: implement long polling for GET events [MLG-85]

nrajanee:
## Description

feat: implement long polling for GET events [MLG-85]. 

I will rename GetSearcherEventsLongPolling to GetSearcherEvents API (and remove any code that's related to the old API; it's commented out right now.) once long polling is approved. 

Implementation is similar to Preemption. 

(I've also made 2 small improvements in postgres_snapshots_test -> Since it's a test I think it's okay to do it as part of this PR?). 

## Test Plan

[MLG-85]: https://determinedai.atlassian.net/browse/MLG-85?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


master/internal/db/postgres_snapshots_test.go
```diff
@@ -20,6 +20,8 @@ func TestCustomSearcherSnapshot(t *testing.T) {
 	etc.SetRootPath(RootFromDB)
 	db := MustResolveTestPostgres(t)
 	MustMigrateTestPostgres(t, db, MigrationsFromDB)
+	user := RequireMockUser(t, db)
+	exp := RequireMockExperiment(t, db, user)
```
nrajanee, line 24: Using exp here because the database can get into a stale state (experiment_snapshots has a foreign key constraint of experiment in experiments table).

master/internal/db/postgres_snapshots_test.go
```diff
@@ -56,5 +59,6 @@ func TestCustomSearcherSnapshot(t *testing.T) {
 	queue2, err := cSearcher2.GetCustomSearcherEventQueue()
 	require.NoError(t, err)
 	require.Equal(t, queue1.GetEvents(), queue2.GetEvents())
-	db.DeleteSnapshotsForExperiment(1)
+	db.DeleteSnapshotsForExperiment(exp.ID)
+	db.DeleteExperiment(exp.ID)
```
nrajanee, line 63: Small improvement in postgres_snapshots_test: Since, I'm deleting snapshot it's also better to delete experiment otherwise locally while rerunning make devcluster you can get into issues.

harness/determined/searcher/_search_runner.py
```diff
@@ -119,7 +119,9 @@ def run_experiment(
 
         try:
             while experiment_is_active:
-                time.sleep(5)
+                time.sleep(
+                    2
+                )  # we don't want to call long polling API more often than every 2 seconds.
```
dzhu, line 124: I think the wait should only happen if there's an error? Since the request now blocks on getting anything, waiting shouldn't be necessary normally.

dzhu (commented):
I'll want to think more about edge cases in execution, but here's a start.

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -17,24 +18,69 @@ type (
...
+		ID uuid.UUID
+		C  <-chan []*experimentv1.SearcherEvent
+	}
 )
 
 func newSearcherEventQueue() *SearcherEventQueue {
 	events := make([]*experimentv1.SearcherEvent, 0)
-	return &SearcherEventQueue{events: events, eventCount: 0}
+	return &SearcherEventQueue{
+		events: events, eventCount: 0,
```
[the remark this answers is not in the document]
nrajanee: I think I'll have to remove all of them or specify all of them. I think there's a benefit to specifying them because it makes it clear. It's consistent with other new search methods such as asha etc.

master/pkg/searcher/custom_searcher_events_queue.go
```diff
@@ -17,24 +18,69 @@ type (
...
+		watchers: map[uuid.UUID]chan<- []*experimentv1.SearcherEvent{},
+	}
+}
+
+// Watch creates a eventsWatcher. If events are available add events and close it.
+func (q *SearcherEventQueue) Watch(id uuid.UUID) (EventsWatcher, error) {
```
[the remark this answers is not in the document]
nrajanee: I did this because it's similar to how Preemption is done. But yes I agree generating internally makes more sense. Thanks!

master/internal/api_experiment.go
```diff
@@ -41,7 +41,7 @@ import (
 	"github.com/determined-ai/determined/master/pkg/ptrs"
 	"github.com/determined-ai/determined/master/pkg/schemas"
 	"github.com/determined-ai/determined/master/pkg/schemas/expconf"
-	"github.com/determined-ai/determined/master/pkg/searcher"
+	searcher "github.com/determined-ai/determined/master/pkg/searcher"
```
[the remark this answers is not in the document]
nrajanee: var w searcher.EventsWatcher on line 262

harness/determined/searcher/_search_runner.py
```diff
@@ -115,9 +115,11 @@ def run_experiment(
...
                 events = self.get_events(session, experiment_id)
-                if events is None:
+                if events is None or len(events) == 0:
```
dzhu, line 122: Probably `not events` would be more idiomatic.

master/internal/api_experiment.go
```diff
@@ -205,28 +205,40 @@ func (a *apiServer) GetSearcherEvents(ctx context.Context, req *apiv1.GetSearche
...
 
 	addr := experimentsAddr.Child(req.ExperimentId)
-	switch err = a.ask(addr, req, &resp); {
-	case err != nil:
-		return nil, status.Errorf(codes.Internal, "failed to get events from actor %v", err)
-	default:
-		return resp, nil
+	var w searcher.EventsWatcher
+	err = a.ask(addr, req, &w)
+	if err != nil {
```
dzhu, line 223:
```suggestion
	if err = a.ask(addr, req, &w); err != nil {
```

master/internal/experiment.go
```diff
@@ -455,10 +461,19 @@ func (e *experiment) Receive(ctx *actor.Context) error {
...
+				ctx.Respond(err)
+			} else {
+				ctx.Respond(w)
 			}
-			ctx.Respond(resp)
+		}
+
+	case UnwatchEvents:
+		queue, err := e.searcher.GetCustomSearcherEventQueue()
+		if err != nil {
```
dzhu, line 473:
```suggestion
		if queue, err := e.searcher.GetCustomSearcherEventQueue(); err != nil {
```

master/internal/api_experiment.go
```diff
@@ -205,28 +205,40 @@ func (a *apiServer) GetSearcherEvents(ctx context.Context, req *apiv1.GetSearche
...
+		return nil, status.Errorf(codes.Internal,
+			"failed to get events from actor: long polling %v", err)
+	}
+
+	defer a.ask(addr, UnwatchEvents{w.ID}, &w) //nolint
```
dzhu, line 228: This should be fixable, so reminder to come back here.

dzhu (approved):

master/internal/api_experiment.go
stoksc: In preemption long polling, we've just used a tell here instead.

nrajanee: Using tell here instead.

PR #5121, opened 2022-09-26
chore: Convert permission.name string to permission.id enum [DET-8464]

mapmeld:
## Description

Now that permissions are defined in the Proto/API, we can replace check of `permission.name == 'create_project'` with `permission.id == V1PermissionType.CREATEPROJECT`.

To make everything work, we need a few changes:
- Include an V1PermissionType.OSSUSER option for the open source client
- canMoveExperiment checks that you can DELETEEXPERIMENT from the source and CREATEEXPERIMENT in the destination
- Similar for canMoveProject
- Cluster admin is not a permission; it's a role which will include these permissions
- View Workspaces is not a permission; tbd

## Test Plan

- V1PermissionType changes match (i.e. there is no accidental swap of CREATEEXPERIMENT / CREATEPROJECT) 
- OSS user behavior continues to be normal.

## Checklist


proto/src/determined/rbac/v1/rbac.proto
```diff
@@ -85,6 +85,9 @@ enum PermissionType {
   // If assigned at a workspace scope, can only assign roles to that workspace
   // scope.
   PERMISSION_TYPE_ASSIGN_ROLES = 6002;
+
+  // Permission used on client to expect open source product behavior.
+  PERMISSION_TYPE_OSS_USER = -1;
```
NicholasBlaskey, line 90: I'm somewhat concerned about adding this to the API definition.

From the API perspective we are never going to return or want to accept this ```OSS_USER``` permission. In the future when we allow customizable roles someone using our API will see this as an option.

mapmeld: Hmm. I want to have this enum be the source of permissions for client-side permissions-check in `usePermissions` and avoid declaring them twice. It's not possible to extend a typescript enum. I might be able to replace it in types.ts with a type which is either `Api.PermissionType | ClientPermissionType`.

mapmeld: Let me know if this alternative makes sense as committed ( https://github.com/determined-ai/determined/pull/5121/commits/da7238c50a272a228aede388da9851ae3111dd5f ) - could also go with PermissionType / MockPermissionType or other names for these types.

webui/react/src/hooks/usePermissions.test.tsx
```diff
@@ -32,7 +32,6 @@ const PermissionRenderer: React.FC<Props> = () => {
     canDeleteWorkspace,
     canModifyWorkspace,
     canViewWorkspace,
-    canViewWorkspaces,
```
julian-determined-ai, line 35: This implementation LGTM. Can we make sure to follow up on how we are going to implement `canViewWorkspaces`.

julian-determined-ai (approved):

mapmeld:
OK - restored `canViewWorkspaces` testable with:
- viewing in OSS - no issues
- viewing with &f_rbac=on - see NoPermissions screen
- view with &f_rbac=on&f_mock_permissions_read=on - you now have permission to read

PR #5125, opened 2022-09-26
ci: Switch to token w/ higher quota [INFENG-100]

dannysauer:
Using the application instead of the generated token allows for a higher GH API quota and has the side effect of triggering further actions (mostly only relevant on the labeler or page commit actions).

## Description

Using the application instead of the generated token allows for a higher GH API quota and has the side effect of triggering further actions (mostly only relevant on the labeler or page commit actions).

## Test Plan

Fingers crossed!

## Checklist

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
     - uses: actions/labeler@v4
       with:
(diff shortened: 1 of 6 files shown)


loksonarius (approved):

dannysauer:
There's no need to run full E2E tests on files that didn't change.  Merging.

PR #5128, opened 2022-09-27
fix: pausing experiment and examples

aciborowska:
## Description

This PR includes two fixes:
1. Fix for pausing/preempting: after experiment is paused; searcher never preempts, but show msg to the user.
2. Fixes for examples after https://github.com/determined-ai/determined/commit/deb6c31e3754db8aee4320f400040216a420a3b9

## Checklist

### DIFF
--- a/e2e_tests/tests/experiment/test_custom_searcher.py
+++ b/e2e_tests/tests/experiment/test_custom_searcher.py
@@ -174,83 +174,9 @@ def test_run_random_searcher_exp_core_api(
 
 
 @pytest.mark.e2e_cpu_2a
-def test_pause_random_searcher_core_api() -> None:
-    config = conf.load_config(conf.fixtures_path("custom_searcher/core_api_searcher_random.yaml"))
-    exp_name = f"random-pause1-{TIMESTAMP}"
-    config["entrypoint"] += " --exp-name " + exp_name
-    config["entrypoint"] += " --config-name noop.yaml"
-
-    model_def_path = conf.fixtures_path("custom_searcher")
-
-    with tempfile.NamedTemporaryFile() as tf:
-        with open(tf.name, "w") as f:
-            yaml.dump(config, f)
-
-        searcher_exp_id = exp.create_experiment(tf.name, model_def_path, None)
-        exp.wait_for_experiment_state(
-            searcher_exp_id,
-            determinedexperimentv1State.STATE_RUNNING,
-        )
-    # make sure both experiments have started by checking
-    # that multi-trial experiment has at least 1 running trials
-    multi_trial_exp_id = exp.wait_for_experiment_by_name_is_active(exp_name, 1)
-
-    # pause searcher
-    exp.pause_experiment(searcher_exp_id)
-    exp.wait_for_experiment_state(
-        searcher_exp_id, bindings.determinedexperimentv1State.STATE_PAUSED
-    )
-    time.sleep(10)
-
-    # multi-trial should be paused after pausing searcher
-    exp.wait_for_experiment_state(
-        multi_trial_exp_id, bindings.determinedexperimentv1State.STATE_PAUSED
-    )
-    time.sleep(10)
-
-    # activate multi-trial experiment
-    exp.activate_experiment(multi_trial_exp_id)
-    exp.wait_for_experiment_state(
-        multi_trial_exp_id, bindings.determinedexperimentv1State.STATE_QUEUED
-    )
-
-    # activate searcher
-    exp.activate_experiment(searcher_exp_id)
-    exp.wait_for_experiment_state(
-        searcher_exp_id, bindings.determinedexperimentv1State.STATE_QUEUED
-    )
-
-    # wait for experiment to complete
-    exp.wait_for_experiment_state(
-        searcher_exp_id, bindings.determinedexperimentv1State.STATE_COMPLETED
-    )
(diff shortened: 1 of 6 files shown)


e2e_tests/tests/experiment/test_custom_searcher.py
```diff
@@ -173,84 +173,10 @@ def test_run_random_searcher_exp_core_api(
     assert resubmissions == sum([x == "after_save" for x in exception_points])
 
 
-@pytest.mark.e2e_cpu_2a
-def test_pause_random_searcher_core_api() -> None:
```
aciborowska, line 177: Removed since searcher is never preempted.

harness/determined/searcher/_search_runner.py
```diff
@@ -86,11 +86,6 @@ def _get_operations(self, event: bindings.v1SearcherEvent) -> List[searcher.Oper
                 f"experiment {self.search_method.searcher_state.experiment_id} is "
                 f"inactive; state={event.experimentInactive.experimentState}"
             )
-            if (
-                event.experimentInactive.experimentState
-                == bindings.determinedexperimentv1State.STATE_COMPLETED
-            ):
-                self.search_method.searcher_state.experiment_completed = True
```
aciborowska, line 93: Removed because it happens again run_experiment

examples/custom_search_method/asha_search_method/utils.py
```diff
@@ -0,0 +1,25 @@
...
+# User-defined function that generates hyperparameters for each trial.
+# The hyperparameters are passed to a trial in the Create operation.
+# In this example, the model (defined in experiment_files/model_def.py) is expecting
+# the following hyperparameters:
+#   -> global_batch_size,
+#   -> n_filters1
+#   -> n_filters2,
+#   -> learning_rate,
+#   -> dropout,
+#   -> dropout2.
```
nrajanee, line 15: Should we mention something about how sample_params has to return a specific value rather than a range of values? Because we're not finding the best hyperparameter value in a given set of values in search method. We're doing it only for the values sampled from here. Not sure if that's a clarification that's necessary. But when I think hyperparameter search I think of searching in a range of values.

examples/custom_search_method/asha_search_method/asha.py
```diff
@@ -171,20 +161,20 @@ def on_trial_closed(self, request_id: uuid.UUID) -> List[Operation]:
...
     # existing trial, start a new trial or close the experiment.
     def on_trial_exited_early(
-        self, request_id: uuid.UUID, exited_reason: ExitedReason
-    ) -> List[Operation]:
+        self, request_id: uuid.UUID, exited_reason: searcher.ExitedReason
+    ) -> List[searcher.Operation]:
         self.asha_search_state.pending_trials -= 1
 
         # The "if" statement below can be completely removed.
         # There is no need to raise INVALID_HP for hyperparameters
```
nrajanee, line 169: In what scenario would the search method produce unwanted hyperparameters? 

Also, is there a way to say that if it's this hyperparameter then ignore the INVALID_HP error?
Can we control this in sample_params? By only including hyperparameter in the sample_params that we want to train?

aciborowska: That is a legacy statement (maybe unnecessary). Scenario: Assume a user is running a HP search with searcher implemented in go, for instance ASHA. While user can define a range of values for selecting HP value, the user cannot exclude certain values (can be based on domain knowledge). In such scenarios, user raises INVALID_HP in their model code, and that is handled by searcher in the master (for instance, ASHA starts a new trial with new HP values).

You are absolutely right that with Custom Search method, user can simply deal with this scenario in sample_params method (which is described in the lower part of this comment). I decided to leave the code to handle INVALID_HP error for two reasons:
1. show users that raising INVALID_HP for CS can be avoided,
2. but also if their code is currently raising INVALID_HP errors, they can still handle it in the Custom Search method as they want.

How about this change: 
`Instead, make sure to start trials only with hyperparameters that satisfy your criteria. For instance, in this example, sample_params() should return only valid hyperparameters.`
?

examples/custom_search_method/asha_search_method/utils.py
aciborowska: would changing `function that generates hyperparameters for each trial` to `function that generates a set of hyperparameters for each trial` make it more clear that it's about specific values and not about the range?

mpkouznetsov: I think `hyperparameters for each trial` is fine. I am using a lengthier phrase: `that generates a combination of hyperparameters (a point in the hyperparameter space) for each trial`.

examples/custom_search_method/asha_search_method/utils.py
```diff
@@ -0,0 +1,25 @@
...
+def sample_params() -> Dict[str, object]:
+    hparams = {
+        "global_batch_size": 64,
+        "n_filters1": random.randint(8, 64),
+        "n_filters2": random.randint(8, 72),
+        "learning_rate": random.uniform(0.0001, 1.0),
```
mpkouznetsov, line 21:
```suggestion
        "learning_rate": 10 ** random.uniform(-4, 0),
```

mpkouznetsov (approved):
one change suggested.

nrajanee (approved):

PR #5089, opened 2022-09-21
fix: remove `as any`

keita-determined:
## Description
[DET-8091]

## Test Plan
- Check if code doesn't break and work as expected
- Check if types are correct
- Check if there's nothing break in web

## Checklist
- [ ] Changes have been manually QA'd
- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](https://github.com/determined-ai/determined/blob/master/docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.
- [ ] If modifying `/webui/react/src/shared/` verify `make -C webui/react test-shared` passes.

[DET-8091]: https://determinedai.atlassian.net/browse/DET-8091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5130, opened 2022-09-27
fix: null check

keita-determined:
## Description
[DET-8490]

## Test Plan
- Check if there's no error in the Jira ticket

## Checklist

[DET-8490]: https://determinedai.atlassian.net/browse/DET-8490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/ExperimentTrials.tsx
+++ b/webui/react/src/pages/ExperimentDetails/ExperimentTrials.tsx
@@ -153,7 +153,7 @@ const ExperimentTrials: React.FC<Props> = ({ experiment, pageRef }: Props) => {
     const validationRenderer = (key: keyof TrialItem) => {
       return function renderer(_: string, record: TrialItem): React.ReactNode {
         const hasMetric = (obj: TrialItem[keyof TrialItem]): obj is MetricsWorkload => {
-          return typeof obj === 'object' && 'metrics' in obj;
+          return !!obj && typeof obj === 'object' && 'metrics' in obj;
         };
 
         const item: TrialItem[keyof TrialItem] = record[key];


hamidzr (approved):
`null` has a typeof `object` huh gets me everytime. 
lgtm. 

nit: I've been leaning away from `!` operator on unclear types in typescript in favor of explicitly checking the expected type.

hamidzr:
> `null` has a typeof `object` huh gets me everytime. lgtm.
> 
> nit: I've been leaning away from `!` operator on unclear types in typescript in favor of explicitly checking the expected type.

I didn't test that it addresses but the codechange on its own makes sense

keita-determined:
> I didn't test that it addresses the ticket but the code change on its own makes sense. are there other places that we have the same check? why do we have to use type guards to access the metrics attribute? shouldn't these be typed? where do we lose that info

It comes from [here](https://github.com/determined-ai/determined/pull/5089/files#diff-83699459460d8591c92d659692f13d839889b7653e5f12eef2f1a453044f411aR155-R157)

TrialItem interface has various types, so in order to narrow type, type guard is used.  Original code used `as any`, so we tried not to lose the info with type guard. 
AFAIK, I don't see similar things.

keita-determined:
>  I've been leaning away from ! operator on unclear types in typescript in favor of explicitly checking the expected type.

You mean `!!`?  I think its widely used but I agree that sometimes its confusing. I'll leave it for now cuz of conciseness

hamidzr:
> > I didn't test that it addresses the ticket but the code change on its own makes sense. are there other places that we have the same check? why do we have to use type guards to access the metrics attribute? shouldn't these be typed? where do we lose that info
> 
> It comes from [here](https://github.com/determined-ai/determined/pull/5089/files#diff-83699459460d8591c92d659692f13d839889b7653e5f12eef2f1a453044f411aR155-R157)
> 
> TrialItem interface has various types, so in order to narrow type, type guard is used. Original code used `as any`, so we tried not to lose the info with type guard. AFAIK, I don't see similar things.

it seems that the input to that fn needs to be only one of the following types instead of a general `key of trialItem` but I'm not sure if there is a good way of expressing that the way the helper is laid out

```
  checkpoint?: CheckpointWorkload;
  training?: MetricsWorkload;
  validation?: MetricsWorkload;
```

keita-determined:
> it seems that the input to that fn needs to be only one of the following types instead of a general key of trialItem but I'm not sure if there is a good way of expressing that the way the helper is laid out

I don't think so cuz `getMetricValue` func has the following parameters

```ts
export const getMetricValue = (
  workload?: { metrics?: Record<RecordKey, number> },
  metricName?: string,
):
```
So when parameter doesn't have `metrics`, the func wants undefined parameter

hamidzr:
ah I might have missed that 👍🏼

PR #5131, opened 2022-09-27
chore: Placeholder when notes/markdown is disabled [DET-8409]

mapmeld:
## Description

As we're adding RBAC permissions, we sometimes disable NotesCard. The underlying Markdown placeholder says "Add notes..." even if no changes can be made.

This is a minor change to the NotesCard component to pass disabled attribute to Markdown, and changing the placeholder for that case.

## Test Plan

Most Notes should stay the same. You can use the storybook or set the NotesCard component on a page to be `disabled={true}`

## Checklist

### DIFF
--- a/webui/react/src/components/Markdown.stories.tsx
+++ b/webui/react/src/components/Markdown.stories.tsx
@@ -139,7 +139,7 @@ export const Default = (): React.ReactNode => (
 
 export const Disabled = (): React.ReactNode => (
   <div style={{ height: '80vh', width: '600px' }}>
-    <Markdown disabled={true} markdown={''} />
+    <Markdown disabled={true} markdown="" />
   </div>
 );
 
--- a/webui/react/src/components/Markdown.stories.tsx
+++ b/webui/react/src/components/Markdown.stories.tsx
@@ -137,6 +137,12 @@ export const Default = (): React.ReactNode => (
   </div>
 );
 
+export const Disabled = (): React.ReactNode => (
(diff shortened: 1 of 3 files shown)


webui/react/src/components/Markdown.stories.tsx
```diff
@@ -137,6 +137,12 @@ export const Default = (): React.ReactNode => (
   </div>
 );
 
+export const Disabled = (): React.ReactNode => (
+  <div style={{ height: '80vh', width: '600px' }}>
+    <Markdown disabled={true} markdown={''} />
```
EmilyBonar, line 142: super nit: can't this just be `markdown=""`, without the braces?

webui/react/src/components/Markdown.tsx
```diff
@@ -40,7 +41,13 @@ const MarkdownRender: React.FC<RenderProps> = ({ markdown, placeholder, onClick
...
+const Markdown: React.FC<Props> = ({
+  disabled = false,
+  editing = false,
+  markdown,
+  onChange,
+  onClick,
+}: Props) => {
   return (
     <div aria-label="markdown-editor" className={css.base}>
       {editing ? (
```
EmilyBonar, line 53: question: do we want to remove the ability to open the editor if `disabled === true`? This condition could be `(editing && !disabled) ? ...`

EmilyBonar (approved):
Overall looks good

PR #5132, opened 2022-09-27
chore: only request listRoles with rbac on [DET-8419]

mapmeld:
## Description

Removes the mock added in #5079 - instead `listRoles` only happens with RBAC on.

## Test Plan

For OSS user - load `/det/projects/1?f_rbac=on` and see request to `roles/search` (will error out). 

Request the page *without* `f_rbac=on` and there should be no request / error from `roles/search`

## Checklist

### DIFF
--- a/webui/react/src/components/Navigation.tsx
+++ b/webui/react/src/components/Navigation.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
 import { useStore } from 'contexts/Store';
+import useFeature from 'hooks/useFeature';
 import {
   useFetchAgents,
   useFetchKnownRoles,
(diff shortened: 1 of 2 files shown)


webui/react/src/components/Navigation.tsx
```diff
@@ -39,10 +40,11 @@ const Navigation: React.FC<Props> = ({ children }) => {
     return () => canceler.abort();
   }, [canceler, fetchResourcePools]);
 
+  const makeListRoleCalls = useFeature().isOn('rbac');
   useEffect(() => {
-    fetchKnownRoles();
+    makeListRoleCalls && fetchKnownRoles();
```
ClaireNeveu, line 45: You should really use an `if` statement here, `&&` as a guard is used in react code because it's expression-based.

webui/react/src/components/Navigation.tsx
```diff
@@ -39,10 +40,11 @@ const Navigation: React.FC<Props> = ({ children }) => {
     return () => canceler.abort();
   }, [canceler, fetchResourcePools]);
 
+  const makeListRoleCalls = useFeature().isOn('rbac');
```
ClaireNeveu, line 43: Would just call this `rbacEnabled` like we do elsewhere.

mapmeld: ✅

ClaireNeveu (approved):

PR #5136, opened 2022-09-27
feat: rbac authz refactor for user groups [DET-8478]

eecsliu:
## Description
Refactor the user groups code so that it uses rbac.

## Checklist


master/internal/usergroup/authz_basic_impl.go
```diff
@@ -0,0 +1,30 @@
...
+// FilterGroupsList returns the list it was given and a nil error.
+func (a *UserGroupAuthZBasic) FilterGroupsList(curUser model.User,
+	groups []*groupv1.GroupSearchResult,
+) ([]*groupv1.GroupSearchResult, error) {
+	return groups, nil
+}
+
+// CanUpdateGroups always returns nil.
+func (a *UserGroupAuthZBasic) CanUpdateGroups(curUser model.User) error {
+	return nil
```
NicholasBlaskey, line 25: requires admin feels more like existing OSS auth scheme to me

master/internal/usergroup/api_groups.go
```diff
@@ -30,6 +32,15 @@ func (a *UserGroupAPIServer) CreateGroup(ctx context.Context, req *apiv1.CreateG
 		err = apiutils.MapAndFilterErrors(err)
```
NicholasBlaskey, line 32: does this ```mapAndFilterErrors``` here make a deny case just return ```Internal Service Error```

master/internal/usergroup/api_groups.go
```diff
@@ -75,6 +86,16 @@ func (a *UserGroupAPIServer) GetGroups(ctx context.Context, req *apiv1.GetGroups
...
+	}
+
+	searchResults, err = AuthZProvider.Get().FilterGroupsList(*curUser, searchResults)
+	if err != nil {
+		return nil, err
+	}
+
 	return &apiv1.GetGroupsResponse{
 		Groups: searchResults,
 		Pagination: &apiv1.Pagination{
```
NicholasBlaskey, line 101: I think this pagination is throw off by the filtering of the group list, like if we can only view 1 group out of a 100 we will get a total back of 100.

Can we make ```FilterGroupList``` apply a bun filter like ```GetExperiments``` does?

master/internal/usergroup/api_groups.go
```diff
@@ -95,7 +116,18 @@ func (a *UserGroupAPIServer) GetGroup(ctx context.Context, req *apiv1.GetGroupRe
 		err = apiutils.MapAndFilterErrors(err)
 	}()
 
+	curUser, _, err := grpcutil.GetUser(ctx)
```
NicholasBlaskey, line 119: If a user has zero group permissions ```GetGroup```, ```DeleteGroup```, ```UpdateGroup``` look like they will return access denied when performing them on a group that you do not have access to them.

We ideally want this to return the same exact response when performing with a group that does not exist so that the existence of a group isn't leaked to someone without permissions on that group.

In experiments this is handled by ```getExperiment``` returning ```canViewExperiment, serverError``` and returning the same 404 as the route would normally return when ```canViewExperiment``` is ```false```. All routes that use a specific experiment also do a call to ```getExperiment```

master/internal/usergroup/api_groups.go
```diff
@@ -30,6 +32,15 @@ func (a *UserGroupAPIServer) CreateGroup(ctx context.Context, req *apiv1.CreateG
...
+	if err != nil {
+		return nil, err
+	}
+	err = AuthZProvider.Get().CanUpdateGroups(*curUser)
+	if err != nil {
+		return nil, err
+	}
+
```
NicholasBlaskey, line 43: here (and adding / removing users in ```UpdateGroup```) we technically need a check that a user can view all ```req.AddUsers``` but since all implementations allow viewing all users I think it is fine to leave it out

master/internal/usergroup/api_groups.go
eecsliu: yeah if it doesn't pass the filter, it returns an internal error

master/internal/usergroup/api_groups.go
eecsliu: ah good point

master/internal/usergroup/api_groups.go
eecsliu: yeah we don't even have a view permission to check on yet - tbh I'm not even sure we need one, but it seems plausible to me you'd require it in the future

master/internal/usergroup/api_groups.go
```diff
@@ -61,8 +74,20 @@ func (a *UserGroupAPIServer) GetGroups(ctx context.Context, req *apiv1.GetGroups
 		return nil, apiutils.ErrInvalidLimit
 	}
 
-	groups, memberCounts, tableCount, err := SearchGroupsWithoutPersonalGroups(ctx,
-		req.Name, model.UserID(req.UserId), int(req.Offset), int(req.Limit))
+	query := SearchGroupsQuery(req.Name, model.UserID(req.UserId), true)
```
NicholasBlaskey, line 77: I think we want ```includePersonalGroups``` to false since personal groups are an internal detail users shouldn't care about

master/internal/usergroup/api_groups.go
```diff
@@ -95,7 +120,18 @@ func (a *UserGroupAPIServer) GetGroup(ctx context.Context, req *apiv1.GetGroupRe
...
 	}()
 
+	curUser, _, err := grpcutil.GetUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	gid := int(req.GroupId)
+
+	err = AuthZProvider.Get().CanGetGroup(*curUser, gid)
```
NicholasBlaskey, line 130: I think we want ```CanGetGroup``` to return the boolean for if you can see the group and the server error 

```
if canSee, serverErr = AuthZProvider.Get().CanGetGroup(*curUser, gid); serverErr != nil {
    return serverError
} else if !canSee {
    return errors.Wrapf(db.ErrNotFound, "Error getting group %d", gid) // Same error as when group actually doesn't 
}
```

master/internal/usergroup/api_groups.go
```diff
@@ -125,6 +161,18 @@ func (a *UserGroupAPIServer) UpdateGroup(ctx context.Context, req *apiv1.UpdateG
 		err = apiutils.MapAndFilterErrors(err)
 	}()
 
+	curUser, _, err := grpcutil.GetUser(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	canUpdate, err := AuthZProvider.Get().CanUpdateGroups(*curUser)
```
NicholasBlaskey, line 169: ```CanUpdateGroup``` here (and in delete group) doesn't need to return a boolean since this won't cause any more information about what groups exist to be gotten

If we are going the route that it is all or nothing, you can update all groups or no groups then 
```
if err := AuthZProvider.Get().CanUpdateGroups(*curUser); err != nil {
    return status.Error(codes.PermissionDenied, err.Error())
}
```
is good

If we wanted to have ```CanUpdateGroups``` be conditional on the group then we would want to do

```
if canSee, serverErr = AuthZProvider.Get().CanGetGroup(*curUser, gid); serverErr != nil {
    return serverError
} else if !canSee {
    return errors.Wrapf(db.ErrNotFound, "Error getting group %d", gid) // Same error as when group actually doesn't 
}  
if err := AuthZProvider.Get().CanUpdateGroups(*curUser, gid); err != nil {
    return status.Error(codes.PermissionDenied, err.Error())
}

```

master/internal/usergroup/api_groups.go
NicholasBlaskey: We should find a way to not filter out these access deny errors

maxrussell: Hm, good point. `MapAndFilterErrors` was intended to interact with DB sentinel errors, but I agree that it would be more useful if either:
- it blanket passed through all GRPC errors. We could perhaps use [`status.Code(err)`](https://pkg.go.dev/google.golang.org/grpc/status#Code) to detect for these and pass them through.
- its scope was broadened to be a standardization layer for all API-level errors
- it only stopped errors that came from the DB-related calls (e.g. stuff from `database/sql`, Bun, sqlx, etc.)

master/internal/usergroup/api_groups.go
eecsliu: We use `CanUpdateGroup` in group creation, so I think we'll need the boolean for server vs access denied error.

maxrussell: Perhaps I'm missing something, but we could just declare an unexported sentinel variable
```go
var errorAccessDenied = fmt.Errorf("access denied")
```
then calling code could check
```go
err := CanUpdateGroups(...)
if err == errorAccessDenied {
	return nil, status.Errorf(codes.NotFound, "group not found: %d", req.GroupId)
} else if err != nil {
	return nil, err
}
```

I agree that it seems awkward for the API of the interface to be to return a `bool` and an `error` if only one of the two is needed.

master/internal/usergroup/authz_iface.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// FilterGroupsList checks what groups a user can get.
+	// POST /api/v1/groups/search
+	FilterGroupsList(curUser model.User, query *bun.SelectQuery) (*bun.SelectQuery, error)
+
+	// CanUpdateGroups checks if a user can create, delete, or update a group.
+	// POST /api/v1/groups
+	// PUT /api/v1/groups/{group_id}
+	// DELETE /api/v1/groups/{group_id}
+	CanUpdateGroups(curUser model.User) (bool, error)
```
maxrussell, line 24: I think we should add `ctx context.Context` as the first parameter of these functions. The basic version will ignore it, but the RBAC version (and others that need to e.g. make database calls) will use it instead of having to do `context.TODO`.

maxrussell, line 24: I think we should add `ctx context.Context` as the first parameter of these functions. The basic version will ignore it, but the RBAC version (and others that need to e.g. make database calls) will use it instead of having to do `context.TODO`.

master/internal/usergroup/group_intg_test.go
```diff
@@ -84,13 +84,23 @@ func TestUserGroups(t *testing.T) {
...
 		require.Equal(t, 2, count, "group search returned wrong count")
 		require.NotEqual(t, -1, index, "Group user was added to not found when searching by user membership")
 	})
 
+	t.Run("manually edit groups query", func(t *testing.T) {
+		query := SearchGroupsQuery("", testUser.ID, false)
+		query = query.Where("group_name = ?", testGroup.Name)
+		groups, _, count, err := SearchGroupsPaginated(ctx, query, 0, 0)
+		require.NoError(t, err, "failed to search for group in modified query")
+		index := groupsContain(groups, testGroup.ID)
```
maxrussell, line 99: Nit: since this isn't used until line 101, I'd move it down to immediately before that for readability. A blank line between the two might additionally help, but that seems even less important.

maxrussell (changes requested):
Looks super close!

master/internal/usergroup/api_groups.go
NicholasBlaskey: As for returning a sentinel vs boolean in something like ```CanGetGroup``` the boolean and ```serverError``` in my opinion makes the interface easier to implement correctly and harder to misuse. Requiring calling code to check the error returned makes it pretty easy to forgot the sentinel check and just do
```
if err := CanGetGroup(); err != nil {
    return err
}
```

While forgetting to check the boolean version becomes a compiler error
```
if canGetGroup, err := CanGetGroup(); err != nil {
    return err
}
```

eecsliu: After syncing with Nick, the consensus is to essentially do sentinel errors - we'll modify the returned error from database and we'll edit the `mapAndFilterErrors` function to pass that error through.

eecsliu: the `mapAndFilterErrors` change will be pulled in with the authz implementation PR for user groups since that function is only in EE

master/internal/api/apiutils/utils.go
```diff
@@ -47,6 +47,8 @@ func MapAndFilterErrors(err error) error {
 		return status.Error(codes.NotFound, err.Error())
 	case errors.Is(err, db.ErrDuplicateRecord):
 		return status.Error(codes.AlreadyExists, err.Error())
+	case errors.Is(err, db.ErrNotEnoughPermissions):
+		return err
```
NicholasBlaskey, line 51: I think this should be ```status.Error(codes.PermissionDenied, err.Error()``` right?

NicholasBlaskey (approved):
one other comment but 

LGTM nice work think the ```mapAndFilter``` solution turned out really well

master/internal/api/apiutils/utils.go
eecsliu: The db error has been changed to have the status code in it, so the error should be that already.

master/internal/db/postgres_rbac.go
```diff
@@ -0,0 +1,10 @@
+package db
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ErrNotEnoughPermissions is returned when a user
+// does not have permissions required.
+var ErrNotEnoughPermissions = status.Error(codes.PermissionDenied, "access denied")
```
maxrussell, line 10: Some thoughts:
- This doesn't seem like a DB concern; it seems like an API concern. So IMO it should go in an API-related package rather than a DB one.
- This isn't RBAC-specific (as evidenced by it being called by the basic authz implementation), so I recommend a name like `ErrAccessDenied` to avoid talking about the RBAC-specific implementation detail of permissions.
- We should avoid creating files in OSS that might need to be created in EE to minimize rebasing headaches.

maxrussell (approved):

PR #5138, opened 2022-09-27
fix: remove workspace members view mock special casing

julian-determined-ai:
## Description

We do not want the case that is being removed in this PR. We only want to enable `mock_workspace_members` if the query parameter is present. 

## Checklist


ClaireNeveu (approved):

mapmeld (approved):
Does it still need to be a switch{} if there are two options? But OK

PR #4788, opened 2022-08-15
chore: `react-router-dom` partial update part1

keita-determined:
## Description

Successor of #4619

Followed [the instruction](https://github.com/remix-run/react-router/discussions/8753)

I'm done with `props.match -> v6 useParams` in `2) Change component code use v6 instead of v5 APIs`

## Test Plan

- Check if all routes work

## Checklist

- [ ] User-facing API changes need the "User-facing API Change" label.
- [ ] Release notes should be added as a separate file under `docs/release-notes/`.
See [Release Note](https://github.com/determined-ai/determined/blob/master/docs/release-notes/README.md) for details.
- [ ] Licenses should be included for new code which was copied and/or modified from any external code.
- [ ] If modifying `/webui/react/src/shared/` verify `make -C webui/react test-shared` passes.

[DET-1234]: https://determinedai.atlassian.net/browse/DET-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5142, opened 2022-09-28
chore: replace router hooks for react-router-dom v6

keita-determined:
## Motivation

- In order to update React to v18, we need to update React-Router-Dom as well.
- Some Jira tickets demand React18 
  - https://determinedai.atlassian.net/browse/DET-7178
  - https://determinedai.atlassian.net/browse/DET-8046

## Description

Previous: #4788
Followed [the instruction](https://github.com/remix-run/react-router/discussions/8753)

What I have done in this PR
- replace useHistory with useNavigate
- use new `useParams`
  - Replaced `interface` with `type` to adjust new react-router-dom
- Fix test cases

## Plan

```mermaid
gantt
    title Update React-Router-Dom Plan
    dateFormat  YYYY-MM-DD
    section React-Router v6
    Replace hooks           : crit, a1, 2022-09-28, 3d
    Replace Root             :a2, after a1 , 5d
    Wrap up React Router v6         :a3, after a2 , 3d
    section React v18
    Update to React v18      :b1,  2022-10-12, 5d
```

## Test Plan
- Check if routing works as expected
- Check if params works as expected
- Check if location works as expected

## Checklist


webui/react/src/pages/InteractiveTask.tsx
```diff
@@ -49,11 +49,23 @@ const getTitleState = (commandState?: CommandState, taskName?: string): string =
...
+    taskType: tType,
+    taskUrl: tUrl,
+  } = useParams<Params>();
   const [taskState, setTaskState] = useState<CommandState>();
   const storeDispatch = useStoreDispatch();
   const { ui } = useStore();
 
+  const taskId = tId ?? '';
+  const taskName = tName ?? '';
+  const taskResourcePool = tResourcePool ?? '';
```
julian-determined-ai, line 65: nit: Not necessarily specific to this body of work but I think I would personally rather us show some sort of "error no task Id" etc page instead of defaulting to a blank string. This extends to all of the query parameters that are required for a component to be rendered correctly that we will now just replace with `' '`. I don't think it should be in this PR but I would be happy with it being a ticket in the backlog to refactor these.

julian-determined-ai (approved):
Tested and everything looks good. Left one non blocking comment about null parameters. Great work!

webui/react/src/pages/InteractiveTask.tsx
keita-determined: In react router dom v6, `usePram` returns `type | undefined` even though we are sure the param never be undefined. These code wouldn't be `undefined`, but  I needed to use `??`, `assertion` or `as` to make it work.  For now, I use `??` for safety. 
I agree with your opinion though. We should find a way to handle it well.

ClaireNeveu: For the future, if there is a code path that you know should never be hit you should throw an exception. I usually define an `InvariantFailed` exception for this purpose.

PR #5143, opened 2022-09-28
feat: rbac authz refactor

eecsliu:
## Description
Add an interface for RBAC control over RBAC functionality. No basic implementation needed in OSS because it returns not implemented. 

## Checklist


master/internal/rbac/authz_iface.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/uptrace/bun"
+
+	"github.com/determined-ai/determined/master/internal/authz"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/proto/pkg/rbacv1"
+)
+
+// RBACAuthZ describes authz methods for RBAC.
+type RBACAuthZ interface {
```
NicholasBlaskey, line 12: does this even need to exist in OSS? 

EE only could avoid the "multi repo" PR problem here.

eecsliu: yeah I'm not sure tbh. My thought is that if we want to add other implementations of RBAC in the future, it might be useful. However, I can also see us just simply porting it over to OSS if we need to

NicholasBlaskey: My vote is EE since that case seems unlikely to me and it is hard to review this PR without the interface being called

eecsliu: fair enough, will nix this one

PR #5150, opened 2022-09-29
chore: refactor slurm/pbs options to expconf

phillip-gaisford:
## Description

image

## Test Plan

Tested manually.

## Checklist


jerryharrow (commented):
lgtm

The docs need to be updated and perhaps a release note on the incompatible change.

jerryharrow (approved):

docs/reference/reference-training/experiment-config-reference.rst
```diff
@@ -1286,3 +1275,25 @@ To verify your search is working as intended before committing to a full run, yo
 .. code::
 
    det preview-search <configuration.yaml>
+
+***********************
+ Slurm Cluster Details
```
stoksc, line 1280: maybe `Slurm Specific Settings`? idk. i will brainstorm a bit.

master/pkg/schemas/expconf/experiment_config.go
```diff
@@ -43,6 +43,8 @@ type ExperimentConfigV0 struct {
 	RawSecurity                 *SecurityConfigV0           `json:"security,omitempty"`
 	RawTensorboardStorage       *TensorboardStorageConfigV0 `json:"tensorboard_storage,omitempty"`
 	RawWorkspace                *string                     `json:"workspace"`
+	RawSlurmClusterConfig       *SlurmClusterConfigV0       `json:"slurm"`
```
stoksc, line 46: maybe just `SlurmConfigV0`, `PbsConfigV0`? the 'cluster' is a bit weird.

stoksc (commented):
what is here looks great. i think we also need to add `SlurmClusterConfig`/`PbsClusterConfig` to `CommandConfig` and set them on the task spec submitted for commands @ `master/pkg/tasks/task_command.go:65`

phillip-gaisford:
> what is here looks great. i think we also need to add `SlurmClusterConfig`/`PbsClusterConfig` to `CommandConfig` and set them on the task spec submitted for commands @ `master/pkg/tasks/task_command.go:65`

We have move PBS/Slurm info to the experiment configuration file, and I'm not finding that commands have one of those, so I'm wondering how we would set these values.

stoksc:
> > what is here looks great. i think we also need to add SlurmClusterConfig/PbsClusterConfig to CommandConfig and set them on the task spec submitted for commands @ master/pkg/tasks/task_command.go:65
> 
> We have moved PBS/Slurm info to the experiment configuration file, and I'm not finding that commands have one of those, so I'm wondering how we would set these values.

I think we will need to add them to `CommandConfig`, just like we are adding them to expconf.

phillip-gaisford:
> > > what is here looks great. i think we also need to add SlurmClusterConfig/PbsClusterConfig to CommandConfig and set them on the task spec submitted for commands @ master/pkg/tasks/task_command.go:65
> > 
> > 
> > We have moved PBS/Slurm info to the experiment configuration file, and I'm not finding that commands have one of those, so I'm wondering how we would set these values.
> 
> I think we will need to add them to `CommandConfig`, just like we are adding them to expconf.

Ok, got it. After consultation with Jerry, we have created a story for command configuration changes: [FOUNDENG-264](https://jira-pro.its.hpecorp.net:8443/browse/FOUNDENG-264)

master/pkg/schemas/expconf/experiment_config.go
phillip-gaisford: I'll change the names.

phillip-gaisford:
Added a release note (breaking change)

master/pkg/model/environment_config.go
```diff
@@ -153,12 +151,7 @@ func (r *RuntimeItems) For(deviceType device.Type) []string {
 
 // Validate implements the check.Validatable interface.
 func (e Environment) Validate() []error {
-	err := validatePodSpec(e.PodSpec)
-	// Slurm and PodSpec are different targets, so no need to merge results
-	if err != nil {
-		return err
-	}
-	return ValidateSlurm(e.Slurm)
```
stoksc, line 161: did this validation land anywhere else? should we at least make a ticket for reinstating it?

phillip-gaisford: ValidateSlurm() continues to be called explicitly from the Dispatcher RM code. FYI, I wasn't able to determine that Environment.Validate() was ever called, least-ways, a break point there in the master context never fired)

PR #5152, opened 2022-09-29
fix: regression in master user auth error message.

ioga:
## Description

- Return proper "invalid credentials" error, this got regressed with #4976 and #4991.

## Test Plan

curl master with a present but incorrect token, e.g. as outlined [here](https://determined-ai.slack.com/archives/C02PV33GSN5/p1664408968906699)

## Checklist


NicholasBlaskey (approved):
LGTM

though we should probably add a tech debt item around standardizing errors since it can be hard to error check code now

hamidzr (approved):
:shipit:

PR #5155, opened 2022-09-30
chore: staged-lint

keita-determined:
## Description
[DET-8093]

## Test Plan
- Check if only staged files are checked by ci when git commit

## Checklist


.pre-commit-config.yaml
```diff
@@ -3,22 +3,49 @@ repos:
...
+        entry: >
+          bash -c
+          '
+          cd webui/react &&
+          FILES=$(
+            git diff --name-only --staged --relative src/**/*.js src/**/*.jsx src/**/*.ts src/**/*.tsx |
+            tr "\n" " "
+          ) &&
+          make prettier ARGS="-- -c $FILES" &&
+          make eslint ARGS="$FILES"
```
keita-determined, line 15: I wanted to combine these 2 make commands for parallel run but I got errors somehow. Any idea? I did `make prettier ARGS="-- -c $FILES" eslint ARGS="$FILES"`

.pre-commit-config.yaml
```diff
@@ -3,22 +3,49 @@ repos:
     hooks:
       - id: web-js-lint-check
         name: Web JS Lint Check
-        entry: bash -c 'make -j$(nproc) -C webui/react check-eslint check-prettier-js'
+        entry: >
+          bash -c
+          '
+          cd webui/react &&
+          FILES=$(
+            git diff --name-only --staged --relative src/**/*.js src/**/*.jsx src/**/*.ts src/**/*.tsx |
```
keita-determined, line 11: Wanted to combine all file extensions like `src/**/*.{ts.tsx,js,jsx}` but got git error

ClaireNeveu:
If you're going to go the route of writing a whole bash script then the simpler solution to this is to just manually strip the prefixes. Something like 
```
entry: bash -c 'strip-prefix.sh -p "webui/react" $@ | make -j$(nproc) -C webui/react check-eslint check-prettier-js'
```
where strip-prefix.sh takes any number of args, removes the specified prefix, then re-emits the args.

Since we already have to reference the directory when telling pre-commit what command to use this doesn't reduce any portability.

keita-determined:
> If you're going to go the route of writing a whole bash script then the simpler solution to this is to just manually strip the prefixes. Something like

Yes, I thought about it. I had some issues with that. Probably my settings on pre-commit would be wrong. let me check it out

.pre-commit-config.yaml
rb-determined-ai: Did you double-check the outputs of this?  I don't think bash `**/*.js` works like zsh's `**/*.js` works.

If I run `bash -c 'echo **/*.js'` from the root of the project I get no files at all.

.pre-commit-config.yaml
rb-determined-ai: Sounds to me like make just isn't the right tool for this job, why not call the npm stuff directly?

keita-determined: I wanted to use `make` to run commands parallel with `-j`, and I thought its more consistent to use `make`. However, I'm not against using `npm`.

.pre-commit-config.yaml
keita-determined: Yse i double-checked. It works fine with the command on Mac.

rb-determined-ai: wait... if you do `cd /path/to/determined && bash -c 'echo **/*.js'`, you mean you _don't_ get an empty response on mac?

pre-commit/strip.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# strip all TARGET string what matches PATTERN
+# sh strip.sh abcb b -> ab
```
dannysauer, line 4: This comment is inaccurate given the code. :)
It's also a little weird that the example uses sh when the bang path is bash, and the code is not POSIX sh friendly.
```suggestion
# bash strip.sh abcb b -> ac
```

pre-commit/strip.sh
```diff
@@ -0,0 +1,14 @@
...
+
+if [ $# -ne 2 ]; then
+  echo "2 arguments are required"
+  echo "1st arg is source string, 2nd arg is pattern string"
```
dannysauer, line 8: Traditionally, this is where a usage statement would go
```suggestion
  echo "usage: $0 'source string' 'pattern string'"
```
Including `$0` helps track down what script this happened in, rather than having to grep for the error message. :)

pre-commit/strip.sh
```diff
@@ -0,0 +1,14 @@
...
+  exit 1
+fi
+
+SOURCE="$1"
+PATTERN="$2"
+echo "${SOURCE//"$PATTERN"}" # replace all. `//` means global
```
dannysauer, line 14: This really isn't what you want based on how the script is actually called.  What you _want_ is to remove a few directory components (a "prefix") from the path.  So really...
```suggestion
echo "${SOURCE#"$PATTERN"}" # remove pattern from beginning of source
```

keita-determined: the SOURE can have multiple paths like `webui/react/src/pages/ProjectDetails.settings.ts webui/react/src/pages/ProjectDetails.tsx`, so I'll leave it with `//`

dannysauer: Then the args are backwards.  You should pass the pattern in first, and then treat additional files as multiple arguments.

.pre-commit-config.yaml
```diff
@@ -3,22 +3,40 @@ repos:
...
         name: Web JS Lint Check
-        entry: bash -c 'make -j$(nproc) -C webui/react check-eslint check-prettier-js'
+        entry: >
+          bash -c '
+            sh ./pre-commit/strip.sh "$0 $*" webui\/react\/ |
+            (
+              FILE_PATHS=$(cat);
+              make -j$(nproc) -C webui/react prettier PRE_ARGS="-- -c $FILE_PATHS" eslint ES_ARGS="$FILE_PATHS"
+            )
+          '
```
dannysauer, line 13: I think we can improve this. :)

That's way too many hoops for getting the output of a script into a variable.  Just put the script in the command substitution instead.
```suggestion
          bash -c '
              FILE_PATHS=$(sh ./pre-commit/strip.sh "$0 $*" webui\/react\/);
              make -j$(nproc) -C webui/react prettier PRE_ARGS="-- -c $FILE_PATHS" eslint ES_ARGS="$FILE_PATHS"
          '
```

Same applies to the other two sections.  The subshell and pipe are all unnecessary.  Though, realistically, this is all a bit convoluted and challenging to read.  Why isn't this whole block just a stand-alone script in the ./pre-commit folder instead of being embedded in the yaml?  The whole difference in the code is just eslint v/s stylint v/s check-package-lock, and a corresponding variable name.  That could be one script with the first argument being the target and subsequent args being the files. :)  Heck, it could be one script with a couple of symlinks, where `$(basename $0)` in a case statement decides how it's called, and then all the arguments are files.  That might be excessive, though.

Incidentally, it might not be a terrible thing to depend on `realpath` for the path conversion; it's part of the GNU coreutils.  Check this out:
```shell
sauer@host:~/det $ realpath --relative-to webui/react/src webui/react/src/hooks/useTheme.ts 
hooks/useTheme.ts
```
It doesn't even depend on string matching; it uses the actual filesystem:
```shell
sauer@host:~/det $ realpath --relative-to webui/react/src ../determined/webui/./././react/src/hooks/useTheme.ts 
hooks/useTheme.ts
```

dannysauer (changes requested):
I'm not sure why Github is forcing me to do a review instead of individual comments.  Sigh.

pre-commit/strip.sh
keita-determined: oh that's true. I'll change it

.pre-commit-config.yaml
dannysauer: For example, something like this:
```bash
#!/usr/bin/env bash
target=basename($0)
D=webui/react
files=$(realpath --relative-to="$D" "$@")
case $target in
  eslint)  make -j$(nproc) -C "$D" prettier PRE_ARGS="-- -c $files" $target ES_ARGS="$files"  ;;
  a)  make -j$(nproc) -C "$D" prettier PRE_ARGS="-- -c $files" $a A_ARGS="$files"  ;;
  b)  make -j$(nproc) -C "$D" prettier PRE_ARGS="-- -c $files" $b B_ARGS="$files"  ;;
  *) echo "unknonwn target '$target'"; exit 1 ;;
esac
```
Then you name is eslint and also make symlinks names a and b to support those three checks.  Or something like that.

keita-determined: I didn't even know about `realpath`. That's exactly what I wanted. I'll check it out.

dannysauer: It even takes multiple arguments, and returns them al converted l one per line. It's tailor made for this. :)

keita-determined: Even though I actually didn't want to make shell script for pre-commit, I made it cuz i didn't know about `realpath` command.  Now, i think its more concise
What about doing this?
```yaml
 entry: >
          bash -c '
            FILE_PATHS=$(realpath --relative-to webui/react $0 $@) &&
            make -j$(nproc) -C webui/react prettier PRE_ARGS="-- -c $FILE_PATHS" eslint ES_ARGS="$FILE_PATHS"
          '
```

dannysauer: Yeah I think that is probably OK. Do you need $0 in there, though?

I still would rather have it pulled out in a separate script even if it's just two lines. You get syntax hilighting in your editor and cleaner YAML that way. And no weird YAML quoting and indentation junk. :)

keita-determined: > Yeah I think that is probably OK. Do you need $0 in there, though?

`$0` is necessary cuz Mac  uses bash v3. `$@` doesn't take `$0` in v3 (it does above v4), so we will miss one file without `$0`.

> I still would rather have it pulled out in a separate script even if it's just two lines. You get syntax hilighting in your editor and cleaner YAML that way. And no weird YAML quoting and indentation junk. :)

Can I make 3 separate files for js, css, and misc? I don't wanna write logic in a shell script cuz some are not familiar with shell script like me :) and it is painful to read shell script even for shell professional. Also, its more flexible to change code with separate files in the future

dannysauer: $0  It's the name the script was called as, it's not one of the arguments. That's why I was asking if it needs to be included. The arguments start with $1.

You might echo $files and make sure this is doing what you think it's doing. :)

keita-determined: Yes. I did `echo`. 
if its just `echo`, we can get all files. However, we don't get all args with `bash -c 'echo $@'` on pre-commit.

dannysauer: I'll bet that'd behave more consistently with the script as  stand alone file as well. ;)

I suspect there's something going on with the way the exec gets called with the bash -c, variant, but I'd rather not speculate. If you pull it out to a separate script, $0 will be the script name and $@ should work in a less surprising way.

keita-determined: Thanks for the advices. I made a shell script.

pre-commit/web_lint_check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+D=webui/react
+target=$1
+files=$(realpath --relative-to="$D" "${@: 2}" | tr "\n" " ")
```
dannysauer, line 5: The arg array slice is a Bash extension.  It'd be more portable to just pop the first arg off with shift.
```suggestion
shift
files=$(realpath --relative-to="$D" "$@" | tr "\n" " ")
```
That's important, since the bang path and command run in precommit is `sh` rather than `bash`.  In general, it's not safe to assume that /bin/sh is bash; in particular, /bin/sh is `dash` on Debian-derived systems.

keita-determined: I fixed it!

dannysauer (approved):
I'm good with this, assuming it's tested and actually works. :)

keita-determined:
@dannysauer 
Thank you for reviewing. I learned a lot about shell :)

PR #5164, opened 2022-10-03
fix: more correctly compute TLS cert hash in CLI

dzhu:
## Description

While doing some testing, I realized that `PEM_cert_to_DER_cert` is only supposed to operate on one certificate, though it sometimes can work on more than one (if, say, the first ends with Base64 padding characters, or possibly if the certs up to the first one ending in padding happen to together contain a valid number of Base64 characters), which is how this didn't get caught before.

## Test Plan

- [x] check that the CLI does not crashes on some certs where it did before
- [x] check that the hash still matches the result of `openssl x509 -fingerprint -sha256 -inform pem -noout -in <cert>`

### DIFF
--- a/harness/determined/cli/cli.py
+++ b/harness/determined/cli/cli.py
@@ -230,21 +230,23 @@ def main(
                     conn.connect(cast(Sequence[Union[str, int]], (addr.hostname, addr.port)))
                     conn.do_handshake()
                     peer_cert_chain = conn.get_peer_cert_chain()
-                    if peer_cert_chain is None:
+                    if peer_cert_chain is None or len(peer_cert_chain) == 0:
                         # Peer presented no cert.  It seems unlikely that this is possible after
                         # do_handshake() succeeded, but checking for None makes mypy happy.
                         raise crypto.Error()
-                    cert_pem_data = "".join(
+                    cert_pem_data = [
                         crypto.dump_certificate(crypto.FILETYPE_PEM, cert).decode()
                         for cert in peer_cert_chain
-                    )
+                    ]
                 except crypto.Error:
                     die(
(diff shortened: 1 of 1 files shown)


harness/determined/cli/cli.py
```diff
@@ -230,21 +230,21 @@ def die(message: str, always_print_traceback: bool = False) -> None:
...
-                    )
+                    ]
                 except crypto.Error:
                     die(
                         "Tried to connect over HTTPS but couldn't get a certificate from the "
                         "master; consider using HTTP"
                     )
 
-                cert_hash = hashlib.sha256(ssl.PEM_cert_to_DER_cert(cert_pem_data)).hexdigest()
+                cert_hash = hashlib.sha256(ssl.PEM_cert_to_DER_cert(cert_pem_data[0])).hexdigest()
```
rb-determined-ai, line 247: Did you verify that `cert_pem_data[0]` is the leaf certificate and not the root certificate?

I saw the test steps include `openssl x509 -fingerprint -sha256 -inform pem -noout -in <cert>` but I'm also not sure how that command deals with multiple certificate on stdin, so that might not prove anything.

Side note, it might be worth commenting above this line that we're taking the same steps as this openssl command.  Because my first instinct here was "why do we need to do so many steps?".

dzhu: Yes (well, it's whatever was listed first by the server, but if it's not the leaf then verification will break anyway), and `openssl` also hashes only the first, though in a sense all we're trying to do is match its behavior regardless of what that may be, since the only point of computing the hash here is to give something to compare with what other standard tools provide.

rb-determined-ai: > but if it's not the leaf then verification will break anyway
cool

rb-determined-ai (approved):

PR #5165, opened 2022-10-03
feat: pass train_length into on_validation_completed

mpkouznetsov:
## Description

https://determinedai.atlassian.net/browse/MLG-51

Add train_length (that is already returned by master as part of ValidationCompleted event) to `on_validation_completed` event handler.

## Checklist


mpkouznetsov:
> One question: do we need to update the docs as well?
> 
> Other than that, LGTM.

Good catch. Updated the doc.

aciborowska (approved):

PR #5174, opened 2022-10-04
feat: support streaming logs in Python SDK [MLG-46]

rb-determined-ai:
The CLI command:

    det e create const.yaml -f

can now be expressed via the Python SDK:

    e = client.create_experiment()
    t = e.await_first_trial()
    for log in t.logs(follow=True):
        print(log, end="")

## Test Plan

- [x] write automated tests

## Commentary

I'd like to agree on the API before writing tests and landing it.

### DIFF
--- a/harness/determined/common/experimental/trial.py
+++ b/harness/determined/common/experimental/trial.py
@@ -1,9 +1,21 @@
 import enum
-from typing import Any, Optional
+from typing import Any, Iterable, List, Optional
 
 from determined.common import api
-from determined.common.api import bindings
+from determined.common.api import bindings, logs
 from determined.common.experimental import checkpoint
+
+
+class LogLevel(enum.Enum):
+    TRACE = bindings.v1LogLevel.LOG_LEVEL_TRACE.value
+    DEBUG = bindings.v1LogLevel.LOG_LEVEL_DEBUG.value
+    INFO = bindings.v1LogLevel.LOG_LEVEL_INFO.value
+    WARNING = bindings.v1LogLevel.LOG_LEVEL_WARNING.value
+    ERROR = bindings.v1LogLevel.LOG_LEVEL_ERROR.value
+    CRITICAL = bindings.v1LogLevel.LOG_LEVEL_CRITICAL.value
+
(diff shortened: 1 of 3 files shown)


harness/determined/common/experimental/trial.py
```diff
@@ -19,6 +32,52 @@ def __init__(self, trial_id: int, session: api.Session):
...
 
+    def logs(
+        self,
+        follow: bool = False,
+        head: Optional[int] = None,
+        tail: Optional[int] = None,
+        container_ids: Optional[List[str]] = None,
+        rank_ids: Optional[List[int]] = None,
+        stdtypes: Optional[List[str]] = None,
+        min_level: Optional[bindings.v1LogLevel] = None,
```
rb-determined-ai, line 43: bindings is leaking to public api, must fix

harness/determined/common/experimental/trial.py
```diff
@@ -19,6 +31,61 @@ def __init__(self, trial_id: int, session: api.Session):
...
+            raise ValueError(f"head must be non-negative, got {head}")
+        if tail is not None and tail < 0:
+            raise ValueError(f"tail must be non-negative, got {tail}")
+        for log in logs.trial_logs(
+            session=self._session,
+            trial_id=self.id,
+            head=head,
+            tail=tail,
+            follow=follow,
+            # TODO: Rename this to "node_id" and support it in the python sdk.
```
mpkouznetsov, line 74: How will you make sure this TODO is done?

mpkouznetsov (approved):
LGTM

harness/determined/common/experimental/trial.py
rb-determined-ai: Honestly, my plan is not to bother.  This would require a big change in backend naming of things, and in our docs.  It'll probably be a long time before this is important enough to happen, long enough that any ticket I make is more likely to be forgotten than completed.

But it's useful to document why a particular option is _not_ being exposed in the public api.  In this case, it's named `agent_id` but secretly applies to all kinds of nodes other than agents.

PR #5175, opened 2022-10-04
feat: Add Webhook CRUD API [WEB-212]

julian-determined-ai:
## Description

This PR introduces a `CRD` (no U here) API and the necessary DB tables and migrations that allow users to create webhooks with defined triggers. Retrieve all webhooks, and delete a webhook. 

Not included in this PR: 
 - RBAC permissions and implementations for the API. If permissible, the plan will be to follow up this PR with another PR that will introduce RBAC auth for the `CRD` actions.  

## Test Plan
1. Navigate to Swagger UI
2. In the POST Webhooks api submit a Webhook payload
For example:

```
{
  "id": 0,
   "url":"http://mywebhookxample.org",
   "triggers":[
      {
         "triggerType":"TRIGGER_TYPE_EXPERIMENT_STATE_CHANGE",
         "condition":{
            "state":"COMPLETED"
         }
      }
   ],
  "webhookType": "WEBHOOK_TYPE_SLACK"
}
```
3. Execute the GET request and ensure any webhooks created are returned.
4. Execute the DELETE request on any defined webhook the request should be succesful, confim that the delete webhook is not in the GET request anymore. 

## Checklist


master/internal/webhooks/postgres_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+		_, err := tx.NewDelete().Model((*Webhook)(nil)).Where("id = ?", id).Exec(ctx)
+		if err != nil {
+			return err
+		}
+
+		_, err = tx.NewDelete().Model((*Trigger)(nil)).Where("webhook_id = ?", id).Exec(ctx)
+		if err != nil {
+			return err
+		}
+		return nil
```
ioga, line 52: since you have `ON DELETE CASCADE`, you don't need that 2nd delete query and the wrapping tx. 
by the time webhook is deleted, all it's triggers are already gone.
(if you didn't have `ON DELETE CASCADE`, this trigger deletion would've had to be first - otherwise webhook delete would fail to referential integrity violation)

julian-determined-ai: This redundant deletion logic has been removed

ioga: you can get rid of the `RunInTx` transaction wrapper as well. there's no need for it, since it's a single statement.

julian-determined-ai: The transaction wrapper has been removed.

master/internal/webhooks/postgres_webhook.go
```diff
@@ -0,0 +1,54 @@
...
+
+// DeleteWebhook deletes a Webhook and its Triggers from the DB.
+func DeleteWebhook(ctx context.Context, id WebhookID) error {
+	return db.Bun().RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+		_, err := db.Bun().NewDelete().Model((*Webhook)(nil)).Where("id = ?", id).Exec(ctx)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
```
ioga, line 53: there's still a redundant `RunInTx` here:

```suggestion
	_, err := db.Bun().NewDelete().Model((*Webhook)(nil)).Where("id = ?", id).Exec(ctx)
	if err != nil {
		return err
	}
	return nil
```

julian-determined-ai: Missed it before, updated now.

proto/src/determined/webhook/v1/webhook.proto
```diff
@@ -0,0 +1,65 @@
...
+    }
+  };
+  // The id of the trigger.
+  int32 id = 1;
+  // The type of the trigger.
+  TriggerType trigger_type = 2;
+  // The trigger condition.
+  google.protobuf.Struct condition = 3;
+  // The parent webhook of the trigger.
+  int32 webhookid = 4;
```
ioga, line 64:
```suggestion
  int32 webhook_id = 4;
```

master/internal/webhooks/webhook.go
```diff
@@ -0,0 +1,183 @@
...
+	ID  WebhookID `bun:"id,pk,autoincrement"`
+	URL string    `bun:"url,notnull"`
+
+	Triggers    Triggers    `bun:"rel:has-many,join:id=webhook_id"`
+	WebhookType WebhookType `bun:"webhook_type,notnull"`
+}
+
+// WebhookFromProto returns a model Webhook from a proto definition.
+func WebhookFromProto(w *webhookv1.Webhook) Webhook {
+	spew.Dump(w)
```
ioga, line 38: perhaps remove this debug?
```suggestion
```

ioga (approved):
couple small comments, lgtm otherwise

PR #5177, opened 2022-10-04
feat: Experiment Groups [WEB-53]

EmilyBonar:
## Description
Will come back and add details/screenshots

## Checklist


master/internal/api_experiment.go
```diff
@@ -391,9 +391,13 @@ func (a *apiServer) GetExperiments(
...
 		ColumnExpr("p.user_id AS project_owner_id").
 		Column("e.config").
+		ColumnExpr("e.group_id").
+		ColumnExpr("CASE WHEN e.group_id IS NULL THEN NULL ELSE g.name END AS group_name").
 		Join("JOIN users u ON e.owner_id = u.id").
 		Join("JOIN projects p ON e.project_id = p.id").
-		Join("JOIN workspaces w ON p.workspace_id = w.id")
+		Join("JOIN workspaces w ON p.workspace_id = w.id").
+		Join("LEFT JOIN experiment_groups g ON e.group_id IS NOT NULL " +
+			"AND e.group_id = g.id AND e.project_id = g.project_id")
```
mapmeld, line 400: I believe that LEFT JOIN allows you to write the join without these checks for compatibility, for example I have no groups in my DB but I am able to run:

```sql
SELECT e.id, g.name FROM experiments e LEFT JOIN experiment_groups g ON e.group_id = g.id
```

master/internal/api_experiment.go
```diff
@@ -847,6 +855,18 @@ func (a *apiServer) UnarchiveExperiment(
 	}
 }
 
+func maskContains(mask []string, v string) bool {
+	if len(mask) == 0 {
+		return true
+	}
+	for _, s := range mask {
```
mapmeld, line 862: in `api_experiment.go` we've been able to use `slices.Contains(array, val)` which covers this for loop

webui/react/src/services/apiConfig.ts
```diff
@@ -692,7 +696,7 @@ export const getExperimentLabels: DetApi<
 > = {
   name: 'getExperimentLabels',
   postProcess: (response) => response.labels || [],
-  request: (params, options) => detApi.Experiments.getExperimentLabels(params.project_id, options),
+  request: (params, options) => detApi.Experiments.getExperimentLabels(params.projectId, options),
```
mapmeld, line 699: thanks for catching these `project_id` -> `projectId` fixes across the system

master/static/srv/insert_experiment_group.sql
```diff
@@ -0,0 +1,11 @@
+WITH g AS (
+  INSERT INTO experiment_groups (name, project_id)
+  VALUES ($1, $2)
+  RETURNING id, name, project_id
+), ge AS (
```
mapmeld, line 5: In `insert_experiment_group.sql` as I understand it, I think there's no need to then check the count; you could select `SELECT g.id, g.name, g.project_id, 0 AS num_experiments`

master/static/srv/get_experiment_groups.sql
```diff
@@ -0,0 +1,3 @@
+SELECT id, name, project_id, (SELECT COUNT(*) FROM experiments WHERE group_id = g.id) AS num_experiments
```
mapmeld, line 1: I don't know for sure, if this might be converted to bun or if this is provably more efficient.
In `get_workspaces.sql` we count experiments for each project in one table scan / query:

```sql
WITH
[....]
exp_count_by_project AS (
  SELECT COUNT(*) AS count, project_id FROM experiments
  GROUP BY project_id
)
```

I think this strategy would be better to do the counts first, and then in the final SELECT, you can JOIN or query this table of counts

mapmeld:
- tested WebUI table interactions ✅ 
- tested WebUI groups modal interactions ✅ 
- tested use of config.yaml to send a group name. confirmed it works well with workspace and project attributes in config.yaml ✅ 

I made some requests on backend and sql code

webui/react/src/hooks/useModal/useModalExperimentGroups.tsx
```diff
@@ -0,0 +1,115 @@
...
+        handleError(e);
+      }
+    },
+    [canceler.signal, fetchExperimentGroups, onAction, projectId],
+  );
+
+  const modalContent = useMemo(() => {
+    return (
+      <Form className={css.container} form={form} layout="vertical">
+        {experimentGroups.map((group) => (
```
mapmeld, line 66: When opening up the Groups modal and there are 0 groups in this project, can we add a message that there are no Groups?

master/internal/core.go
```diff
@@ -215,15 +215,15 @@ func (m *Master) getMasterLogs(c echo.Context) (interface{}, error) {
...
 // @Param   timestamp_before query string true "End time to get allocations for (YYYY-MM-DDTHH:MM:SSZ format)"
-//nolint:lll
 // @Success 200 {} string "A CSV file containing the fields experiment_id,kind,username,labels,slots,start_time,end_time,seconds"
-//nolint:godot
 // @Router /allocation/raw [get]
 // @Deprecated
+//
+//nolint:lll
+//nolint:lll
+//nolint:lll
```
maxrussell, line 226: Wow, the linter was really hard on you! Sorry you had to deal with all this; it's not normally like this. I wonder why all these linting issues haven't come up before. Since you didn't modify many of these files other than the linting issues, it can't be that…

master/internal/api_project.go
```diff
@@ -342,3 +343,114 @@ func (a *apiServer) UnarchiveProject(
 	}
 	return &apiv1.UnarchiveProjectResponse{}, nil
 }
+
+func (a *apiServer) GetExperimentGroups(
+	ctx context.Context, req *apiv1.GetExperimentGroupsRequest) (*apiv1.GetExperimentGroupsResponse,
+	error,
+) {
+	resp := &apiv1.GetExperimentGroupsResponse{}
+	err := a.m.db.QueryProto("get_experiment_groups", &resp.Groups, req.ProjectId)
```
maxrussell, line 352: As [Ilia mentioned in Slack](https://determined-ai.slack.com/archives/C6D7MDSQ5/p1664566350696609), we're moving toward using Bun for our queries and are trying to add no more `QueryProto` queries. The code here would look something like this:
```go
var groups ExperimentGroups
db.Bun().NewSelect().Model(&groups)
	Column("id").Column("name").Column("project_id").
	ColumnExpr("(SELECT COUNT(*) FROM experiments WHERE group_id = g.id) AS num_experiments").
	Where("g.project_id = ?", req.ProjectId).
	Scan(ctx)

return groups.Proto(), nil
```

We've also moved to putting these kinds of things in their own Go packages and such. For a more detailed example, I recommend checking out how we did User Groups in PRs #4519 and #4620. I'm happy to do a call or chat on Slack about this with you if you'd like!

maxrussell (changes requested):
Still reading through the PR (jeez the linter was being hard on you!), but want to submit my work-in-progress review so we can start the conversation about Bun and `QueryProto` and such.

master/static/migrations/20220930104117_add-experiment-groups.tx.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE experiment_groups (
+  id SERIAL PRIMARY KEY,
+  name TEXT NOT NULL,
+  project_id INT REFERENCES projects(id),
```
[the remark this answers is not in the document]
EmilyBonar: Nope, project_id should never be null. I'll fix it!

master/internal/api_experiment.go
```diff
@@ -903,6 +938,23 @@ func (a *apiServer) PatchExperiment(
...
+		if req.Experiment.GroupId != nil {
+			var destGroup *projectv1.ExperimentGroup
+			destGroup, err = a.GetExperimentGroupByID(*req.Experiment.GroupId)
+			if err != nil {
+				return nil, err
+			}
+			madeChanges = true
+			exp.GroupId = &destGroup.Id
+			exp.GroupName = &destGroup.Name
+		} else if len(mask) > 0 {
```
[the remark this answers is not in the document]
EmilyBonar: Actually, as implemented `maskContains(mask, "group_id")` will return true when `mask` is an empty array (`len(mask) == 0`). I did that so that api calls that didn't include an updateMask would still work as normal. Let me know if you have a better idea for how to handle this.

proto/src/determined/api/v1/api.proto
```diff
@@ -442,7 +442,7 @@ service Determined {
       returns (PatchExperimentResponse) {
     option (google.api.http) = {
       patch: "/api/v1/experiments/{experiment.id}"
-      body: "experiment"
+      body: "*"
```
ioga, line 445: in the [patch feature](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature/) there're 2 ways to do it: hidden and exposed.
it seems we've got an exposed one here, i.e. previously PatchExperiment had payload with experiment at the top level, i.e. `{id: ..., displayName: ....`
and with this change, it'd become nested:
```
{
  experiment: {
    id: ...,
    displayName: ...
  },
  updateMask: ...
}
```

1. This is breaking the current PatchExperiment API
2. I believe the "hidden" approach is nicer for raw REST API users in particular. 

Can you please try the "hidden" way and investigate what needs to be done to support it in typescript bindings?
- I imagine the go side will stay unchanged.
- we'll need to change our python bindings generator to support this, but it's pretty straightforward.

PR #5178, opened 2022-10-04
refactor: authz allocations [DET-8366. DET-7971] 

NicholasBlaskey:
## Description

Refactors allocations according as an ```authz``` interface calling experiments.

## Test Plan

Integration test. 

## Checklist


master/internal/core.go
```diff
@@ -910,7 +910,6 @@ func (m *Master) Run(ctx context.Context) error {
 	)
 	tasksGroup := m.echo.Group("/tasks", authFuncs...)
 	tasksGroup.GET("", api.Route(m.getTasks))
-	tasksGroup.GET("/:task_id", api.Route(m.getTask))
```
NicholasBlaskey, line 913: removed this endpoint since as far as I can tell it was broken for ~13 months

master/internal/api_tasks.go
```diff
@@ -35,9 +37,68 @@ var (
...
 
+func expFromAllocationID(
+	m *Master, allocationID model.AllocationID,
+) (isExperiment bool, exp *model.Experiment, err error) {
+	resp, err := m.rm.GetAllocationHandler(
+		m.system,
+		sproto.GetAllocationHandler{ID: allocationID},
+	)
+	if err != nil {
+		return false, nil, status.Errorf(codes.NotFound, "allocation not found: %s", allocationID)
```
NicholasBlaskey, line 48: Type checking this error for allocation not found is hard since in agent rm it returns 
```errors.New("task handler not found on any resource pool")```

while in k8s it returns
```fmt.Errorf("actor %s did not response", addr)```

Are there other possible errors that can come that I missing that make this blanket ```err != nil``` return ```not found``` suppress a useful error message?

master/internal/core.go
ioga: we should deprecate `/tasks` as well - I believe it's still being used by the CLI though.

ioga (approved):

master/internal/api_tasks.go
NicholasBlaskey: I think a nil response ends up getting converted to an error in ```a.ask``` when a response is expected
https://github.com/determined-ai/determined/blob/48ac9af201c4ab3c8ccd56683b47c08bc35d40d6/master/internal/api.go#L198

PR #5179, opened 2022-10-04
feat: SearcherState as parameter

mpkouznetsov:
## Description

SearcherState should be attached to a SearchRunner, rather than to a SearchMethod. SearchMethod should be as abstract as possible; specifically, it is better if it has no constructor.

## Checklist


rb-determined-ai (commented):
I think this is a big improvement to the conceptual simplicity of the SearchMethod.

mpkouznetsov:
Liam, after a discussion:
> thanks for dissecting that! i understand the motivation for moving the SearcherState to the SearchRunner better now as basically explicitly separating system-defined and managed state and user-defined and managed state.  i'm on board with moving the SearcherState to the SearchRunner

PR #5180, opened 2022-10-04
fix: check for horovod backend in PyTorchTrial

azhou-determined:
## Description

this is a bug. if you call `on_validation_end` or `on_validation_step_end` while using non-horovod backend, the trial errors out. 

## Checklist

### DIFF
--- a/harness/determined/pytorch/_pytorch_trial.py
+++ b/harness/determined/pytorch/_pytorch_trial.py
@@ -610,7 +610,7 @@ class PyTorchTrialController(det.TrialController):
             pytorch._convert_metrics_to_numpy(self.context.reduce_metrics(for_training=False))
         )
 
-        if self.context.distributed.size > 1 and any(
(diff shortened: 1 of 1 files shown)


harness/determined/pytorch/_pytorch_trial.py
```diff
@@ -610,7 +610,7 @@ def _compute_validation_metrics(self) -> workload.Metrics:
             pytorch._convert_metrics_to_numpy(self.context.reduce_metrics(for_training=False))
         )
 
-        if self.context.distributed.size > 1 and any(
+        if self.context.distributed.size > 1 and self.use_horovod and any(
```
rb-determined-ai, line 613: how would you get the correct metrics in all workers if you don't broadcast in the pytorch-native dtrain case?

rb-determined-ai (approved):
nice catch, and nice simplification.

PR #5185, opened 2022-10-05
fix: RBAC calls getPermissionsSummary, changes to admin's use of listRoles [WEB-517]

mapmeld:
## Description

The main objective / fix:

- When RBAC is on, we call useFetchMyRoles / getPermissionsSummary and update the Store with those roles [if you'd like me to call this useFetchCurrentUserRoles or something more verbose let me know]
- Loading these from the real API -> `UserAssignment` changes -> `relevantAssigned` changes

Changes to where we call `listRoles`:

- The call to useFetchKnownRoles / listRoles is moved from the main page to the UserSettings and GroupSettings pages, as it's only used by admin on these two pages to load the list of roles that Users and Groups can be assigned to

Changes to `getUserPermissions` (same as getPermissionsSummary, + a userId param) 

- We pass the `userId` to the API to get their roles instead of the current user's roles

## Test Plan

For @NicholasBlaskey who is testing this in EE:

- Can you view workspaces?
- On `/det/settings/user-management/rbac-enabled` can you see a list of permissions on each user?
- When editing that user, or a group on `/det/settings/group-management/rbac-enabled`, do you have a list of roles to assign?

## Checklist


webui/react/src/hooks/useModal/UserSettings/useModalCreateUser.tsx
```diff
@@ -53,7 +53,15 @@ const ModalForm: React.FC<Props> = ({ form, user, groups, viewOnly }) => {
 
   const updatePermissions = useCallback(async () => {
     if (user && canGetPermissions) {
-      const viewPermissions = await getUserPermissions({ userId: user.id });
+      const { roles } = await getUserPermissions({ userId: user.id });
+      const viewPermissions = new Array<Permission>();
```
ClaireNeveu, line 57: Little bit confused by this. This is the form for creating a user, right? Shouldn't we be selecting roles and not permissions?

mapmeld: When we view or edit the user, this populates a viewable table of their permissions (a bit controversial especially in the future where they have permissions on specific workspaces)

The roles selector gets roles from `knownRoles` which comes from the Store and is updated by `fetchKnownRoles` on the UserManagement page.  It might be movable into this component if that seems better.

ClaireNeveu: If it's just for display then we really should be deriving those permissions in the render function but that's not important for this PR.

webui/react/src/services/api.ts
```diff
@@ -83,9 +83,15 @@ export const resetUserSetting = generateDetApi<EmptyParams, Api.V1ResetUserSetti
 export const getUserPermissions = generateDetApi<
   Service.GetUserParams,
   Api.V1GetPermissionsSummaryResponse,
-  Type.Permission[]
+  Type.PermissionsSummary
```
ClaireNeveu, line 86: Was this type just wrong before?

ClaireNeveu: Okay I see we were throwing out information in the API response parsing step and now we're not.

webui/react/src/services/apiConfig.ts
```diff
@@ -193,16 +193,26 @@ export const resetUserSetting: DetApi<
...
-    return permissions;
-  },
+  postProcess: (response) => ({
+    assignments: response.assignments.map(decoder.mapV1UserAssignment),
+    roles: response.roles.map(decoder.mapV1Role),
+  }),
+  request: (params) => detApi.RBAC.getPermissionsSummary(params.userId),
+};
+
+export const getPermissionsSummary: DetApi<
```
ClaireNeveu, line 206: This name doesn't make it clear what this does. Can we add a JSDoc explaining that this fetches the permissions summary for the currently logged-in user?

webui/react/src/services/decoder.ts
```diff
@@ -44,6 +44,16 @@ export const mapV1Permission = (permission: Sdk.V1Permission): types.Permission
   };
 };
 
+export const mapV1UserAssignment = (
+  assignment: Sdk.V1RoleAssignmentSummary,
+): types.UserAssignment => {
+  return {
+    isGlobal: assignment.isGlobal || false,
+    roleId: assignment.roleId || 0,
```
ClaireNeveu, line 52: We shouldn't use 0 as a special value. If this field is optional the type should be nullable.

webui/react/src/services/decoder.ts
```diff
@@ -44,6 +44,16 @@ export const mapV1Permission = (permission: Sdk.V1Permission): types.Permission
...
+    workspaces: assignment.scopeWorkspaceIds || new Array<number>(),
```
ClaireNeveu, line 53: Just curious here why you do `new Array<number>()` instead of `[]`?

webui/react/src/services/decoder.ts
mapmeld: Backend did not do the annotation to require fields on their custom proto types for RBAC, only the request/response objects. Saw this before with Role. Just pushed a fix to this one.

webui/react/src/services/decoder.ts
mapmeld: I thought that TypeScript would care about the type of the empty array, but (in this context at least?) it doesn't need a type hint. Will do `[]` here

webui/react/src/services/apiConfig.ts
mapmeld: Added JSDoc here and to the similar `getUserPermissions` in the latest commit; hope that helps.

ClaireNeveu (approved):
This looks good on my end but please get a review from @NicholasBlaskey before merging.

NicholasBlaskey (approved):

PR #5187, opened 2022-10-05
feat: useSettings improvements

thiagodallacqua-hpe:
## Description

For a time now, we've been concerned with the performance and overall stability of our `useSettings` hook, it has several issues with an un-synced state with the latest DB changes upon refresh, visual bugs due to state issues, lots of re-renders (although that is a deeper issue as I've noticed while working on this hook, as we have are constantly pooling from DB on multiple locations...).

So, as proposed solutions, we've been working on revamping the `useSettings` to a new shape.

The idea is to use context provider to subscribe to the settings state from DB, remove some transformations and checks that we had due to weak type structure, and use DB as a source of truth, leaving localStorage behind for good, and, finally, have a leaner and easier to understand structure.

## Test Plan

taken from the spec docs:

key features:

it has to be persisted on DB and initial values should be taken from BE (if present)

in case that there’s nothing on DB (yet) it should initialize based on the config object passed as parameter

it should keep track of the state based on the initial configurations and sync with the query parameters, if necessary

settings should be scoped per project id → page, meaning that I can keep several settings state per project, depending on which page uses the hook (this is supposed to be available but actually isn’t, should be doable with key manipulation)

settings can be “global”, meaning that each account can have certain settings configurations - have noticed a few settings that actually should be global state, like sidebar collapsed, etc.

settings can be shared via URL, meaning that the hook should be able to collect navigation data and parse any possible query parameters into valid settings, which, in turn, should be active for said route. If the user then modifies these settings they should persist to the DB as usual.

updating any setting shouldn’t block rendering (keep DB update async)

settings should be self-contained, meaning that each subtree shouldn’t listen for settings change other than it’s own settings

useSettings should be able to reset the settings based on initial config

useSettings should be able to give all active/valid scoped settings

each entry of the config obj should be of BaseType, meaning no complex structures

updateSettings should have a prop that allows you to specify whether you want the current state update to be a new entry in the browser history

browser back navigation should navigate back through these updates and going forward should have a similar result.

## Checklist

### DIFF
--- a/webui/react/src/services/api-ts-sdk/api.ts
+++ b/webui/react/src/services/api-ts-sdk/api.ts
@@ -3166,6 +3166,20 @@ export interface V1GetMasterResponse {
 }
 
 /**
+ * Response to GetMeRequest.
+ * @export
+ * @interface V1GetMeResponse
+ */
+export interface V1GetMeResponse {
+    /**
+     * The requested user.
+     * @type {V1User}
+     * @memberof V1GetMeResponse
+     */
+    user: V1User;
+}
+
+/**
  * Request to get a file of model definition.
  * @export
  * @interface V1GetModelDefFileRequest
@@ -3839,6 +3853,20 @@ export interface V1GetTrialsCollectionsResponse {
      * @memberof V1GetTrialsCollectionsResponse
      */
     collections?: Array<V1TrialsCollection>;
+}
+
+/**
+ * Response to GetUserByUsernameRequest.
+ * @export
+ * @interface V1GetUserByUsernameResponse
+ */
+export interface V1GetUserByUsernameResponse {
+    /**
+     * The requested user.
+     * @type {V1User}
+     * @memberof V1GetUserByUsernameResponse
+     */
+    user: V1User;
 }
 
 /**
@@ -5207,6 +5235,64 @@ export interface V1Notebook {
 }
 
 /**
+ * Arguments to a notify container running.
+ * @export
+ * @interface V1NotifyContainerRunningRequest
+ */
+export interface V1NotifyContainerRunningRequest {
+    /**
+     * The ID of the allocation.
+     * @type {string}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    allocationId: string;
+    /**
+     * The UUID of the participant in a notify container running message.
+     * @type {string}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    requestUuid?: string;
+    /**
+     * The number of process to wait for.
+     * @type {number}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    numPeers?: number;
+    /**
+     * The container's rank.
+     * @type {number}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    rank?: number;
+    /**
+     * 
+     * @type {string}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    nodeName?: string;
+    /**
+     * The data from this process.
+     * @type {any}
+     * @memberof V1NotifyContainerRunningRequest
+     */
+    data: any;
+}
+
+/**
+ * 
+ * @export
+ * @interface V1NotifyContainerRunningResponse
+ */
+export interface V1NotifyContainerRunningResponse {
+    /**
+     * The data for all the processes.
+     * @type {Array<any>}
+     * @memberof V1NotifyContainerRunningResponse
+     */
+    data: Array<any>;
+}
+
+/**
  * Order records in either ascending or descending order.   - ORDER_BY_UNSPECIFIED: Returns records in no specific order.  - ORDER_BY_ASC: Returns records in ascending order.  - ORDER_BY_DESC: Returns records in descending order.
  * @export
  * @enum {string}
@@ -5504,6 +5590,18 @@ export interface V1PatchTrialsCollectionResponse {
  */
 export interface V1PatchUser {
     /**
+     * String denoting the username of the user.
+     * @type {string}
+     * @memberof V1PatchUser
+     */
+    username?: string;
+    /**
+     * String denoting the password of the user.
+     * @type {string}
+     * @memberof V1PatchUser
+     */
+    password?: string;
+    /**
      * Bool denoting whether the account is an admin account.
      * @type {boolean}
      * @memberof V1PatchUser
@@ -5516,23 +5614,23 @@ export interface V1PatchUser {
      */
     active?: boolean;
     /**
+     * Name to display in the web UI.
+     * @type {string}
+     * @memberof V1PatchUser
+     */
+    displayName?: string;
+    /**
      * The user and group on the agent host machine.
      * @type {V1AgentUserGroup}
      * @memberof V1PatchUser
      */
     agentUserGroup?: V1AgentUserGroup;
     /**
-     * The new username.
-     * @type {string}
+     * Indicate whether the provided password is pre-salted & hashed or not.
+     * @type {boolean}
      * @memberof V1PatchUser
      */
-    username?: string;
-    /**
-     * Name to display in the web UI.
-     * @type {string}
-     * @memberof V1PatchUser
-     */
-    displayName?: string;
+    isHashed?: boolean;
 }
 
 /**
@@ -5567,6 +5665,12 @@ export interface V1PatchWorkspace {
      * @memberof V1PatchWorkspace
      */
     agentUserGroup?: V1AgentUserGroup;
+    /**
+     * Optional checkpoint storage config. Expects same format as experiment config's checkpoint storage.
+     * @type {any}
+     * @memberof V1PatchWorkspace
+     */
+    checkpointStorageConfig?: any;
 }
 
 /**
@@ -5610,15 +5714,15 @@ export interface V1Permission {
      */
     name?: string;
     /**
-     * 
-     * @type {boolean}
+     * Allowed scope types.
+     * @type {V1ScopeTypeMask}
      * @memberof V1Permission
      */
-    isGlobal?: boolean;
+    scopeTypeMask?: V1ScopeTypeMask;
 }
 
 /**
- * List of permissions types. Value of the enum has 9xxxx for global only permissions. Permissions on the same object share the thousands place value like 2001 and 2002.   - PERMISSION_TYPE_UNSPECIFIED: The permission type is unknown.  - PERMISSION_TYPE_ADMINISTRATE_USER: Can create and update other users. Allows updating other users passwords making this permission give all other permissions effectively.  - PERMISSION_TYPE_CREATE_EXPERIMENT: Ability to create experiments.  - PERMISSION_TYPE_VIEW_EXPERIMENT_ARTIFACTS: Ability to view experiment's model code, checkpoints, trials.  - PERMISSION_TYPE_VIEW_EXPERIMENT_METADATA: Ability to view experiment's metadata like experiment config, progress.  - PERMISSION_TYPE_UPDATE_EXPERIMENT: Ability to update experiment and experiment's lifecycle.  - PERMISSION_TYPE_UPDATE_EXPERIMENT_METADATA: Ability to update experiment's metadata.  - PERMISSION_TYPE_DELETE_EXPERIMENT: Ability to delete experiment.  - PERMISSION_TYPE_UPDATE_GROUP: Ability to create, update, and add / remove users from groups.  - PERMISSION_TYPE_CREATE_WORKSPACE: Ability to create workspaces.  - PERMISSION_TYPE_VIEW_WORKSPACE: Ability to view workspace.  - PERMISSION_TYPE_UPDATE_WORKSPACE: Ability to update workspace.  - PERMISSION_TYPE_DELETE_WORKSPACE: Ability to delete workspace.  - PERMISSION_TYPE_SET_WORKSPACE_AGENT_USER_GROUP: Ability to set workspace agent user group config.  - PERMISSION_TYPE_CREATE_PROJECT: Ability to create projects.  - PERMISSION_TYPE_VIEW_PROJECT: Ability to view projects.  - PERMISSION_TYPE_UPDATE_PROJECT: Ability to update projects.  - PERMISSION_TYPE_DELETE_PROJECT: Ability to delete projects.  - PERMISSION_TYPE_UPDATE_ROLES: Ability to create and update role definitions.  - PERMISSION_TYPE_ASSIGN_ROLES: Ability to assign roles to groups / users. If assigned at a workspace scope, can only assign roles to that workspace scope.  - PERMISSION_TYPE_EDIT_WEBHOOKS: Ability to create and delete webhooks.
+ * List of permissions types. Value of the enum has 9xxxx for global only permissions. Permissions on the same object share the thousands place value like 2001 and 2002.   - PERMISSION_TYPE_UNSPECIFIED: The permission type is unknown.  - PERMISSION_TYPE_ADMINISTRATE_USER: Can create and update other users. Allows updating other users passwords making this permission give all other permissions effectively.  - PERMISSION_TYPE_CREATE_EXPERIMENT: Ability to create experiments.  - PERMISSION_TYPE_VIEW_EXPERIMENT_ARTIFACTS: Ability to view experiment's model code, checkpoints, trials.  - PERMISSION_TYPE_VIEW_EXPERIMENT_METADATA: Ability to view experiment's metadata like experiment config, progress.  - PERMISSION_TYPE_UPDATE_EXPERIMENT: Ability to update experiment and experiment's lifecycle.  - PERMISSION_TYPE_UPDATE_EXPERIMENT_METADATA: Ability to update experiment's metadata.  - PERMISSION_TYPE_DELETE_EXPERIMENT: Ability to delete experiment.  - PERMISSION_TYPE_UPDATE_GROUP: Ability to create, update, and add / remove users from groups.  - PERMISSION_TYPE_CREATE_WORKSPACE: Ability to create workspaces.  - PERMISSION_TYPE_VIEW_WORKSPACE: Ability to view workspace.  - PERMISSION_TYPE_UPDATE_WORKSPACE: Ability to update workspace.  - PERMISSION_TYPE_DELETE_WORKSPACE: Ability to delete workspace.  - PERMISSION_TYPE_SET_WORKSPACE_AGENT_USER_GROUP: Ability to set workspace agent user group config.  - PERMISSION_TYPE_SET_WORKSPACE_CHECKPOINT_STORAGE_CONFIG: Ability to set workspace checkpoint storage config.  - PERMISSION_TYPE_CREATE_PROJECT: Ability to create projects.  - PERMISSION_TYPE_VIEW_PROJECT: Ability to view projects.  - PERMISSION_TYPE_UPDATE_PROJECT: Ability to update projects.  - PERMISSION_TYPE_DELETE_PROJECT: Ability to delete projects.  - PERMISSION_TYPE_UPDATE_ROLES: Ability to create and update role definitions.  - PERMISSION_TYPE_ASSIGN_ROLES: Ability to assign roles to groups / users. If assigned at a workspace scope, can only assign roles to that workspace scope.  - PERMISSION_TYPE_EDIT_WEBHOOKS: Ability to create and delete webhooks.
  * @export
  * @enum {string}
  */
@@ -5637,6 +5741,7 @@ export enum V1PermissionType {
     UPDATEWORKSPACE = <any> 'PERMISSION_TYPE_UPDATE_WORKSPACE',
     DELETEWORKSPACE = <any> 'PERMISSION_TYPE_DELETE_WORKSPACE',
     SETWORKSPACEAGENTUSERGROUP = <any> 'PERMISSION_TYPE_SET_WORKSPACE_AGENT_USER_GROUP',
+    SETWORKSPACECHECKPOINTSTORAGECONFIG = <any> 'PERMISSION_TYPE_SET_WORKSPACE_CHECKPOINT_STORAGE_CONFIG',
     CREATEPROJECT = <any> 'PERMISSION_TYPE_CREATE_PROJECT',
     VIEWPROJECT = <any> 'PERMISSION_TYPE_VIEW_PROJECT',
     UPDATEPROJECT = <any> 'PERMISSION_TYPE_UPDATE_PROJECT',
@@ -5948,6 +6053,12 @@ export interface V1PostUserRequest {
      * @memberof V1PostUserRequest
      */
     password?: string;
+    /**
+     * Indicate whether the provided password is pre-salted & hashed or not.
+     * @type {boolean}
+     * @memberof V1PostUserRequest
+     */
+    isHashed?: boolean;
 }
 
 /**
@@ -6024,6 +6135,12 @@ export interface V1PostWorkspaceRequest {
      * @memberof V1PostWorkspaceRequest
      */
     agentUserGroup?: V1AgentUserGroup;
+    /**
+     * Optional checkpoint storage config.
+     * @type {any}
+     * @memberof V1PostWorkspaceRequest
+     */
+    checkpointStorageConfig?: any;
 }
 
 /**
@@ -7130,6 +7247,12 @@ export interface V1Role {
      * @memberof V1Role
      */
     permissions?: Array<V1Permission>;
+    /**
+     * Allowed scope types.
+     * @type {V1ScopeTypeMask}
+     * @memberof V1Role
+     */
+    scopeTypeMask?: V1ScopeTypeMask;
 }
 
 /**
@@ -7145,11 +7268,17 @@ export interface V1RoleAssignment {
      */
     role: V1Role;
     /**
-     * The id of the workspace the role belongs to. Omit for a global scope.
+     * The id of the workspace the role belongs to. Empty for cluster-wide scope.
      * @type {number}
      * @memberof V1RoleAssignment
      */
     scopeWorkspaceId?: number;
+    /**
+     * Whether the role is assigned cluster-wide.
+     * @type {boolean}
+     * @memberof V1RoleAssignment
+     */
+    scopeCluster?: boolean;
 }
 
 /**
@@ -7171,11 +7300,11 @@ export interface V1RoleAssignmentSummary {
      */
     scopeWorkspaceIds?: Array<number>;
     /**
-     * 
+     * Whether the role is assigned cluster-wide.
      * @type {boolean}
      * @memberof V1RoleAssignmentSummary
      */
-    isGlobal?: boolean;
+    scopeCluster?: boolean;
 }
 
 /**
@@ -7279,6 +7408,26 @@ export enum V1SchedulerType {
     KUBERNETES = <any> 'SCHEDULER_TYPE_KUBERNETES',
     SLURM = <any> 'SCHEDULER_TYPE_SLURM',
     PBS = <any> 'SCHEDULER_TYPE_PBS'
+}
+
+/**
+ * 
+ * @export
+ * @interface V1ScopeTypeMask
+ */
+export interface V1ScopeTypeMask {
+    /**
+     * Whether this permission or role can be assigned globally, i.e. cluster-wide. Currently, all permissions can be assigned globally, so this is always true.
+     * @type {boolean}
+     * @memberof V1ScopeTypeMask
+     */
+    cluster?: boolean;
+    /**
+     * Whether this permission or role can be assigned on a particular workspace. For example, `ADMINISTRATE_USER` permission will have this field set to false, since user creation can only be done at a cluster level, and it doesn't make sense for a single workspace.
+     * @type {boolean}
+     * @memberof V1ScopeTypeMask
+     */
+    workspace?: boolean;
 }
 
 /**
@@ -9177,6 +9326,12 @@ export interface V1Workspace {
      * @memberof V1Workspace
      */
     agentUserGroup?: V1AgentUserGroup;
+    /**
+     * Optional checkpoint storage config. Expects same format as experiment config's checkpoint storage.
+     * @type {any}
+     * @memberof V1Workspace
+     */
+    checkpointStorageConfig?: any;
 }
 
 /**
@@ -15697,6 +15852,52 @@ export const InternalApiFetchParamCreator = function (configuration?: Configurat
         },
         /**
          * 
+         * @summary NotifyContainterRunning is used to notify the master that the container is running.  On HPC, the launcher will report a state of \"Running\" as soon as Slurm starts the job, but the container may be in the process of getting pulled down from the Internet, so the experiment is not really considered to be in a \"Running\" state until all the containers that are part of the experiment are running and not being pulled.
+         * @param {string} allocationId The ID of the allocation.
+         * @param {V1NotifyContainerRunningRequest} body 
+         * @param {*} [options] Override http request option.
+         * @throws {RequiredError}
+         */
+        notifyContainerRunning(allocationId: string, body: V1NotifyContainerRunningRequest, options: any = {}): FetchArgs {
+            // verify required parameter 'allocationId' is not null or undefined
+            if (allocationId === null || allocationId === undefined) {
+                throw new RequiredError('allocationId','Required parameter allocationId was null or undefined when calling notifyContainerRunning.');
+            }
+            // verify required parameter 'body' is not null or undefined
+            if (body === null || body === undefined) {
+                throw new RequiredError('body','Required parameter body was null or undefined when calling notifyContainerRunning.');
+            }
+            const localVarPath = `/api/v1/allocations/{allocationId}/notify_container_running`
+                .replace(`{${"allocationId"}}`, encodeURIComponent(String(allocationId)));
+            const localVarUrlObj = url.parse(localVarPath, true);
+            const localVarRequestOptions = Object.assign({ method: 'POST' }, options);
+            const localVarHeaderParameter = {} as any;
+            const localVarQueryParameter = {} as any;
+
+            // authentication BearerToken required
+            if (configuration && configuration.apiKey) {
+                const localVarApiKeyValue = typeof configuration.apiKey === 'function'
+					? configuration.apiKey("Authorization")
+					: configuration.apiKey;
+                localVarHeaderParameter["Authorization"] = localVarApiKeyValue;
+            }
+
+            localVarHeaderParameter['Content-Type'] = 'application/json';
+
+            localVarUrlObj.query = Object.assign({}, localVarUrlObj.query, localVarQueryParameter, options.query);
+            // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
+            delete localVarUrlObj.search;
+            localVarRequestOptions.headers = Object.assign({}, localVarHeaderParameter, options.headers);
+            const needsSerialization = (<any>"V1NotifyContainerRunningRequest" !== "string") || localVarRequestOptions.headers['Content-Type'] === 'application/json';
+            localVarRequestOptions.body =  needsSerialization ? JSON.stringify(body || {}) : (body || "");
+
+            return {
+                url: url.format(localVarUrlObj),
+                options: localVarRequestOptions,
+            };
+        },
+        /**
+         * 
          * @summary PostAllocationProxyAddress sets the proxy address to use when proxying to services provided by an allocation. Upon receipt, the master will also register any proxies specified by the task.
          * @param {string} allocationId The id of the allocation.
          * @param {V1PostAllocationProxyAddressRequest} body 
@@ -16869,6 +17070,26 @@ export const InternalApiFp = function(configuration?: Configuration) {
         },
         /**
          * 
+         * @summary NotifyContainterRunning is used to notify the master that the container is running.  On HPC, the launcher will report a state of \"Running\" as soon as Slurm starts the job, but the container may be in the process of getting pulled down from the Internet, so the experiment is not really considered to be in a \"Running\" state until all the containers that are part of the experiment are running and not being pulled.
+         * @param {string} allocationId The ID of the allocation.
+         * @param {V1NotifyContainerRunningRequest} body 
+         * @param {*} [options] Override http request option.
+         * @throws {RequiredError}
+         */
+        notifyContainerRunning(allocationId: string, body: V1NotifyContainerRunningRequest, options?: any): (fetch?: FetchAPI, basePath?: string) => Promise<V1NotifyContainerRunningResponse> {
+            const localVarFetchArgs = InternalApiFetchParamCreator(configuration).notifyContainerRunning(allocationId, body, options);
+            return (fetch: FetchAPI = portableFetch, basePath: string = BASE_PATH) => {
+                return fetch(basePath + localVarFetchArgs.url, localVarFetchArgs.options).then((response) => {
+                    if (response.status >= 200 && response.status < 300) {
+                        return response.json();
+                    } else {
+                        throw response;
+                    }
+                });
+            };
+        },
+        /**
+         * 
          * @summary PostAllocationProxyAddress sets the proxy address to use when proxying to services provided by an allocation. Upon receipt, the master will also register any proxies specified by the task.
          * @param {string} allocationId The id of the allocation.
          * @param {V1PostAllocationProxyAddressRequest} body 
@@ -17440,6 +17661,17 @@ export const InternalApiFactory = function (configuration?: Configuration, fetch
         },
         /**
          * 
+         * @summary NotifyContainterRunning is used to notify the master that the container is running.  On HPC, the launcher will report a state of \"Running\" as soon as Slurm starts the job, but the container may be in the process of getting pulled down from the Internet, so the experiment is not really considered to be in a \"Running\" state until all the containers that are part of the experiment are running and not being pulled.
+         * @param {string} allocationId The ID of the allocation.
+         * @param {V1NotifyContainerRunningRequest} body 
+         * @param {*} [options] Override http request option.
+         * @throws {RequiredError}
+         */
+        notifyContainerRunning(allocationId: string, body: V1NotifyContainerRunningRequest, options?: any) {
+            return InternalApiFp(configuration).notifyContainerRunning(allocationId, body, options)(fetch, basePath);
+        },
+        /**
+         * 
          * @summary PostAllocationProxyAddress sets the proxy address to use when proxying to services provided by an allocation. Upon receipt, the master will also register any proxies specified by the task.
          * @param {string} allocationId The id of the allocation.
          * @param {V1PostAllocationProxyAddressRequest} body 
@@ -17956,6 +18188,19 @@ export class InternalApi extends BaseAPI {
      */
     public metricNames(experimentId: number, periodSeconds?: number, options?: any) {
         return InternalApiFp(this.configuration).metricNames(experimentId, periodSeconds, options)(this.fetch, this.basePath);
+    }
+
+    /**
+     * 
+     * @summary NotifyContainterRunning is used to notify the master that the container is running.  On HPC, the launcher will report a state of \"Running\" as soon as Slurm starts the job, but the container may be in the process of getting pulled down from the Internet, so the experiment is not really considered to be in a \"Running\" state until all the containers that are part of the experiment are running and not being pulled.
+     * @param {string} allocationId The ID of the allocation.
+     * @param {V1NotifyContainerRunningRequest} body 
+     * @param {*} [options] Override http request option.
+     * @throws {RequiredError}
+     * @memberof InternalApi
+     */
+    public notifyContainerRunning(allocationId: string, body: V1NotifyContainerRunningRequest, options?: any) {
+        return InternalApiFp(this.configuration).notifyContainerRunning(allocationId, body, options)(this.fetch, this.basePath);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


trentwatt (commented):
looking good!

still have some crashing 
https://user-images.githubusercontent.com/28383080/194333441-19642ee4-eda7-4a0c-89aa-3cc2061c0754.mov

and some of this stuff
https://user-images.githubusercontent.com/28383080/194333764-872cebc9-6f76-4216-aafd-e17e9611d796.mov

webui/react/src/hooks/useSettings.ts
```diff
@@ -230,7 +272,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
     userSettings,
```
ClaireNeveu, line 272: I have a lot of questions about the basics of this hook. For example why does `useSettings` listen to changes in the location? Shouldn't those modifications be made _by_ `useSettings`? When is that trigger happening?

(This is kind of why I was recommending rewriting it from scratch)

webui/react/src/hooks/useSettings.ts
```diff
@@ -230,7 +272,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
...
   } = useStore();
   const prevSearch = usePrevious(location.search, undefined);
-  const [settings, setSettings] = useState<T>(() => getDefaultSettings<T>(config, storage));
+  const [settings, setSettings] = useState<T>(() => getDefaultSettings<T>(config, userSettings));
```
ClaireNeveu, line 275: Why does this  have a local state for user settings instead of relying on and modifying the global state?

webui/react/src/hooks/useSettings.ts
trentwatt: for that in particular, the concept is that if you have a url that stores the current page state, you can share it and someone else can view the page in the same state

trentwatt: >we don't need to listen for that.

in that case what is the mechanism for going from url params -> state.

right now we are using `location.search`

trentwatt: I think that' what `useLocation` does?

However you get the info, if you use it in a hook, react is going to want you to put it in the dependency array, no?

ClaireNeveu: @trentwatt `useLocation` subscribes the component to changes in the location. `window.location` is just the raw `Location`. If we don't need to subscribe the component to changes in the location then we should just get it directly.

trentwatt: are you telling me we don't have to use exhaustive deps 😃

ClaireNeveu: If you list all of the dependencies it's equivalent to not including a dependency array. The whole _purpose_ of the dependency array is to narrow the dependencies.

webui/react/src/hooks/useSettings.test.tsx
```diff
@@ -8,7 +8,6 @@ import { CompatRouter } from 'react-router-dom-v5-compat';
 import StoreProvider from 'contexts/Store';
```
keita-determined, line 8: will we keep using useContext?
If so, I think we cannot resolve rendering issues

webui/react/src/hooks/useSettings.ts
```diff
@@ -109,22 +104,25 @@ export const validateSetting = (config: SettingsConfigProp, value: unknown): boo
...
     }
     acc[prop.key] = defaultValue;
     return acc;
-  }, {} as GenericSettings) as unknown as T;
+  }, {}) as unknown as T;
 };
 
 export const queryParamToType = (type: BaseType, param: string | null): Primitive | undefined => {
   if (param === null || param === undefined) return undefined;
   if (type === BaseType.Boolean) return param === 'true';
```
keita-determined, line 123: All data are passed as string from db. I feel like we should store data as it is in db, then we can get rid of unnecessary type conversion code

webui/react/src/hooks/useSettings.tsx
```diff
@@ -0,0 +1,332 @@
...
+      }
+    } catch (e) {
+      handleError(e, { silent: true, type: ErrorType.Ui });
+    }
+
+    return acc;
+  }, {});
+};
+
+const useSettings = <T, >(config: SettingsConfig<T>): UseSettingsReturn => {
```
ClaireNeveu, line 198: `T` should be included in the return type here, it's the type of `Settings`.

webui/react/src/hooks/useSettings.tsx
```diff
@@ -0,0 +1,332 @@
...
+ * UserSettingsState contains all the settings for a user
+ * across the application. Each key identifies a unique part
+ * of the interface to store settings for.
+ */
+type UserSettingsState = Map<string, Settings>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Settings = { [key: string]: any }; //TODO: find a way to use a better type here
+
+export enum BaseType {
```
ClaireNeveu, line 25: You should delete this. We also don't need `Primitive` since we're using io-ts.

webui/react/src/hooks/useSettings.tsx
```diff
@@ -0,0 +1,332 @@
...
+  useEffect(() => {
+    getUserSetting({}, { signal: canceler.signal })
+    .then((response) => {
+      setApiSettings(response.settings);
+    });
+
+    return () => canceler.abort();
+  }, [canceler]);
+
+  const settingsState = useMemo(() => {
```
ClaireNeveu, line 84: The way you're using this is a little confusing. I think you're basically ignoring the state subscriptions except for the initial fetching. This makes sense to me since you only want to trigger re-renders for individual items in the map. In that case I would get rid of `apiSettings` and just put this map in a `useRef` that you then populate with the initial settings.

I think we maybe should have an `isLoading` state boolean too which will trigger the one global re-render we need. 

So that would be like:
```typescript
const settingsState = useRef<Map<string, any>(new Map());
const [isLoading, setIsLoading] = useState(true);

...

<UserSettings.Provider value={{
      state: settingsState,
      update,
      isLoading,
    }}>...
```

Then we can also have proper loading states in the components that use `useSettings`.

webui/react/src/hooks/useSettings.tsx
```diff
@@ -0,0 +1,332 @@
...
+      const setting = config.settings[key];
+
+      newSettings[setting.storageKey] = setting.defaultValue;
+    };
+
+    state.set(config.applicableRoutespace, newSettings);
+  }, [config, state]);
+
+  // parse navigation url to state
+  useEffect(() => {
```
ClaireNeveu, line 221: What do you think about putting this logic in the `SettingsProvider` instead of the hook? It only needs to happen once on page load since we only ever want to restore settings based on the initial navigation. Here this might cause unintended state manipulations.

webui/react/src/hooks/useSettings.tsx
```diff
@@ -0,0 +1,332 @@
...
+    if (isEqual(settings, stateSettings)) return;
+
+    Object.keys(settings).forEach((setting) => {
+      state.set(config.applicableRoutespace, { [setting]: settings[setting] });
+    });
+  }, [config, state]);
+
+  const slicedState = state.get(config.applicableRoutespace);
+
+  const settings: Settings = useMemo(() => slicedState ?? {}, [slicedState]); // TODO: test if changing it to return only the state.get affects re-rendering
```
ClaireNeveu, line 238: I think you might actually be more worried _missed_ re-renders with your current setup. `useMemo` only restricts recalculations, it doesn't create new ones. Since the map isn't part of any state we need to introduce a manual rerender when we update it.

webui/react/src/hooks/useSettings.tsx
ClaireNeveu: I don't think that's the intended behavior. This logic is meant to handle when the user navigates directly to the page. When that happens we want to restore settings–the same way we restore them from the DB when not navigating directly. When you're navigating _within_ the app, we don't need to read the url parameters because the _app_ is setting the url.

Are there cases where we're using `navigate` to bounce settings through the url instead of setting them directly?

webui/react/src/App.tsx
```diff
@@ -149,7 +150,9 @@ const App: React.FC = () => {
     <HelmetProvider>
       <StoreProvider>
         <DndProvider backend={HTML5Backend}>
-          <AppView />
+          <SettingsProvider>
```
keita-determined, line 153: Im not very worried about this so far cuz React18 can batch rendering, but if project gets bigger it might be a big rendering issue.

webui/react/src/components/InteractiveTable.stories.tsx
```diff
@@ -28,32 +29,28 @@ const columns = new Array(20).fill(null).map(() => {
...
-        isArray: true,
-      },
+      type: array(number),
     },
-    {
-      key: 'row',
-      type: { baseType: BaseType.String, isArray: true },
+    row: {
+      defaultValue: [],
+      storageKey: 'row',
```
keita-determined, line 50: `storageKey` and key of object are the same, do we still need `storageKey`?
I like the new KV structure btw. its more efficient than array.

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,356 +1,285 @@
...
-            }
-          }
+  const resetSettings = useCallback(
+    (settings?: string[]) => {
+      const array = settings ?? Object.keys(config.settings);
 
-          // Keep track of internal setting changes to update async from query settings.
-          if (isValid) acc.internalSettings[key] = value as T[keyof T];
+      array.forEach((setting) => {
+        let defaultSetting;
```
keita-determined, line 208: nit: initializing just in case

keita-determined (commented):
Thanks for the work.
I have some questions and finds. 

### Finds
- in secret mode, I get the following error
image

- I dont see query parameter in url
- I get a blank page in `/det/models`

### Questions
- Seems like `settings` is always optional, so we need to unwrap it using `?`. Why is it optional? and is it possible to make it not optional?

webui/react/src/components/LogViewer/LogViewerFilters.settings.ts
```diff
@@ -10,48 +12,53 @@ export interface Settings {
...
-      type: {
-        baseType: BaseType.String,
-        isArray: true,
-      },
+const config: SettingsConfig<Settings> = {
+  applicableRoutespace: 'log-viewer-filters',
+  settings: {
+    agentId: {
+      defaultValue: undefined,
+      storageKey: 'agentId',
```
keita-determined, line 20: if we are sure that storageKey and key are always the same, we can remove `storageKey` cuz we might do typo in the future and spend lots of time to find out...

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
```diff
@@ -23,38 +24,40 @@ const { Item } = Form;
...
+      type: union([number, undefinedType]),
     },
-    {
-      key: 'template',
+    template: {
+      defaultValue: '',
       skipUrlEncoding: true,
-      type: { baseType: BaseType.String },
+      storageKey: 'template',
+      type: union([string, undefinedType]),
```
keita-determined, line 52: how come these types are change from string to string | undef?

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
- * storageKey       - If provided, save/load setting into/from storage.
- * type.baseType    - How to decode the string-based query param.
- * type.isArray     - List based query params can be non-array.
- */
-export interface SettingsConfigProp {
-  defaultValue?: GenericSettingsType;
-  key: string;
+import { Settings, SettingsProvider, UserSettings } from './useSettingsProvider';
+
+export enum BaseType {
```
keita-determined, line 16: we should use object literal instead of enum.

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
 };
 
-export const queryParamToType = (type: BaseType, param: string | null): Primitive | undefined => {
-  if (param == null) return undefined;
+const queryParamToType = (type: BaseType, param: string | null): Primitive | undefined => {
+  if (param === null || param === undefined) return undefined;
   if (type === BaseType.Boolean) return param === 'true';
   if (type === BaseType.Float || type === BaseType.Integer) {
-    const value = type === BaseType.Float ? parseFloat(param) : parseInt(param);
+    const value = Number(param);
```
keita-determined, line 73: this code can be replaced with https://github.com/sindresorhus/query-string#parsebooleans
i know this func exist before this PR, but I personally prefer to rely on lib funcs instead of making our own func if possible

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
        */
-      const queryValue = Array.isArray(paramValue)
-        ? paramValue
-            .map((value) => queryParamToType(baseType, value))
-            .filter((value): value is Primitive => value !== undefined)
-        : queryParamToType(baseType, paramValue);
+      try {
+        const paramValue = params[setting.storageKey];
+        const baseType = setting.type;
+        const expectingArray = baseType.is([]);
```
keita-determined, line 93: nit: variable name should be `isArray` or something similar

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
+
+        if (paramValue !== null) {
+          let queryValue: Primitive | Primitive[] | undefined;
```
keita-determined, line 96: initializing variable

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,356 +1,285 @@
...
+  const {
+    auth: { user },
+  } = useStore();
+  const { isLoading, querySettings, state, update } = useContext(UserSettings);
+  const navigate = useNavigate();
 
-  return queryString.stringify(fullSettings);
-};
+  // creates an entry at the global state
+  useEffect(() => {
```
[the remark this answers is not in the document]
keita-determined: i meant that it would be nice not to pass `useSettings` down to child components.

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
+    if (!querySettings) return;
+
+    const settings = queryToSettings<T>(config, querySettings);
+    let stateSettings = state.get(config.applicableRoutespace) ?? {};
+
+    if (isEqual(settings, stateSettings)) return;
+
+    Object.keys(settings).forEach((setting) => {
+      stateSettings = Object.assign(stateSettings, { [setting]: settings[setting] });
```
keita-determined, line 174: are you sure you wanna use `Object.assign` here? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
`Object.assign` is destructive

webui/react/src/hooks/useSettingsProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+  state: UserSettingsState;
+  update: (key: string, value: Settings, clearQuerySettings?: boolean) => void;
+};
+
+export const UserSettings = createContext<UserSettingsContext>({
+  isLoading: false,
+  querySettings: '',
+  state: new Map(),
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  update: () => {},
```
keita-determined, line 30: `() => undefined` then you can remove the eslint flag

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.settings.ts
```diff
@@ -39,65 +41,80 @@ export interface Settings extends InteractiveTableSettings {
...
+    columns: {
       defaultValue: DEFAULT_COLUMNS,
-      key: 'columns',
       skipUrlEncoding: true,
       storageKey: 'columns',
-      type: {
-        baseType: BaseType.String,
-        isArray: true,
-      },
+      type: array(
```
keita-determined, line 51: I think Trent removed it before, so `columns` should be `skipUrlEncoding: true`.

webui/react/src/hooks/useSettingsProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+  } = useStore();
+  const [canceler] = useState(new AbortController());
+  const [isLoading, setIsLoading] = useState(false);
+  const querySettings = useRef('');
+  const settingsState = useRef(new Map<string, Settings>());
+
+  useEffect(() => {
+    if (!user?.id) return;
+
+    setIsLoading(true);
```
keita-determined, line 47: you can use `suspense`

webui/react/src/pages/Settings/UserManagement.settings.ts
```diff
@@ -33,55 +35,53 @@ export interface UserManagementSettings extends InteractiveTableSettings {
   sortKey: V1GetUsersRequestSortBy;
 }
 
-const config: SettingsConfig = {
-  settings: [
-    {
+const config: SettingsConfig<UserManagementSettings> = {
+  applicableRoutespace: 'user-management',
+  settings: {
+    columns: {
```
keita-determined, line 41: skipUrlEncoding should be true here too

webui/react/src/pages/TrialsComparison/Table/TrialTable.tsx
```diff
@@ -547,32 +548,36 @@ const TrialTable: React.FC<Props> = ({
...
-          selectedRowKeys: (selectAllMatching ? trials.ids : selectedTrials) as number[],
-        }}
-        scroll={{ x: 'max-content', y: '40vh' }}
-        settings={settings}
-        showSorterTooltip={false}
-        size="small"
-        updateSettings={updateSettings as UpdateSettings<InteractiveTableSettings>}
-        onRow={highlights.onRow}
-      />
+      {settings ? (
```
keita-determined, line 555: I saw this pattern in several files. 
I think we can just wait until `settings` is loaded from db, then `settings` doesnt need to be null

webui/react/src/pages/WorkspaceDetails/WorkspaceMembers.settings.ts
```diff
@@ -21,59 +23,49 @@ export const DEFAULT_COLUMN_WIDTHS: Record<MembersColumnName, number> = {
...
+  sortKey: string;
 }
 
-const config: SettingsConfig = {
+const config: SettingsConfig<WorkspaceMembersSettings> = {
   applicableRoutespace: 'members',
-  settings: [
-    {
+  settings: {
+    columns: {
```
keita-determined, line 32: skipUrlEncoding: true here too

webui/react/src/pages/WorkspaceDetails/WorkspaceProjects.settings.ts
```diff
@@ -51,84 +53,90 @@ export interface WorkspaceDetailsSettings extends InteractiveTableSettings {
...
+  settings: {
+    archived: {
       defaultValue: false,
-      key: 'archived',
       storageKey: 'archived',
-      type: { baseType: BaseType.Boolean },
+      type: union([undefinedType, boolean]),
     },
-    {
+    columns: {
```
keita-determined, line 64: skipUrlEncoding: true here

keita-determined (commented):
Thanks for the update. 
I still see blank pages in some routes.

https://user-images.githubusercontent.com/109113616/197611270-b484c976-a896-4454-9f13-93b96a323418.mov

https://user-images.githubusercontent.com/109113616/197611273-7f2cb3aa-d0f6-4dce-a8bf-0620d336f702.mov

webui/react/src/hooks/useSettings.ts
ClaireNeveu: Should actually be removed since we're using io-ts.

webui/react/src/components/NavigationSideBar.tsx
```diff
@@ -162,8 +163,8 @@ const NavigationSideBar: React.FC = () => {
   }, [canAccessUncategorized, info.branding, showWebhooks]);
 
   const handleCollapse = useCallback(() => {
-    updateSettings({ navbarCollapsed: !settings.navbarCollapsed });
-  }, [settings.navbarCollapsed, updateSettings]);
+    updateSettings({ navbarCollapsed: !settings?.navbarCollapsed });
+  }, [settings?.navbarCollapsed, updateSettings]);
```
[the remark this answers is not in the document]
ClaireNeveu: So to be clear, `null` is the loading state here. Ideally we should have some kind of "if null show spinner" behavior at the top of the render function and not just spam `?` everywhere.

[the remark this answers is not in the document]
ClaireNeveu: Here in the effects we should probably just ignore the effect when the settings are not yet loaded (that's my guess at least).

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
ClaireNeveu: Were settings implicitly unioned with `undefined` in the old code? That is, did `BaseType.String` mean `string | undefined`?

This actually has a default (though a bad one) so it seems like it could just be `string`.

ClaireNeveu: `slot` above definitely seems like it isn't optional and should just be `number`.

Also just a minor point but we really should be using `X | null` for optional types instead of `X | undefined`.

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,299 @@
...
+          if (Array.isArray(paramValue)) {
+            queryValue = paramValue.reduce<Primitive[]>((acc, value) => {
+              const parsedValue = queryParamToType(baseType.name as BaseType, value);
+
+              if (parsedValue !== undefined) acc.push(parsedValue);
+
+              return acc;
+            }, []);
+          } else {
+            queryValue = queryParamToType(baseType.name as BaseType, paramValue);
```
ClaireNeveu, line 112: You should be using the `decode` in io-ts to actually parse out the query params. You may need a special case for strings if you want to keep quotes out of them.
Something like:
```typescript
if (baseType === t.string) { // optional string handling is left as an exercise to the reader
   queryValue = baseType.decode(paramValue)
} else {
   queryValue = baseType.decode(JSON.parse(paramValue))
}
```

webui/react/src/components/LogViewer/LogViewerFilters.settings.ts
ClaireNeveu: It would be a nice optimization if storageKey defaulted to the same value as the key but let's leave that for a future PR

webui/react/src/hooks/useSettings.ts
ClaireNeveu: `foo.decode` does indeed do what we want here. `t.number.decode(5)` will return `Right(5)`, `t.number.decode('str')` will return `Left(...)`.

keita-determined (commented):
Uploading Screen Recording 2022-10-25 at 10.46.11 AM.mov…

I still see some errors in secret mode

trentwatt (commented):
i think some of these are addressed but posting here for tracking purposes

blocking:
- [ ] `experiments/checkpoints` page doesnt load
- [ ] clear filters button not showing up when filters present
- [x] pressing clear filters results in page crashing
- [ ] loading state triggered on table row selection
- [ ] page crash on navigating to `projects/:id/trials` with `f_trials_comparison=on`
- [ ] toggling sort asc/sort desc on tasks page results in queryParams showing up multiple time in url (see below url)
- [ ] `tableOffset=NaN` in url (see below)

```
tasks?sortDesc=false&tableLimit=0&tableOffset=NaN&tableLimit=0&tableOffset=NaN
```

blocking-ish:
- [ ] `skipUrlEncoding` params show up in url

webui/react/src/hooks/useSettings.ts
```diff
@@ -1,359 +1,294 @@
...
-  if (type === BaseType.Boolean && isBoolean(value)) return true;
-  if (type === BaseType.Float && isNumber(value)) return true;
-  if (type === BaseType.Integer && isNumber(value) && Math.ceil(value) === Math.floor(value))
-    return true;
-  if (type === BaseType.String && isString(value)) return true;
-  if (type === BaseType.Object && isObject(value)) return true;
-  return false;
+  isLoading: boolean;
+  resetSettings: ResetSettings;
+  settings: T | null;
```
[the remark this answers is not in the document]
keita-determined: +1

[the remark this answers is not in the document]
trentwatt: it might lead to a bit of janky loading and maybe we want to transition it to a `Loadable` pattern to avoid that in the future, but just for this initial PR i think it's less risky to keep the API consistent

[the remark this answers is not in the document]
keita-determined: idk if its applicable here, but we can use `suspense` when useSettings is loading data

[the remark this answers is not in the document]
ClaireNeveu: Can you provide an example of how that would work?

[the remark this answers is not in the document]
keita-determined: https://beta.reactjs.org/apis/react/Suspense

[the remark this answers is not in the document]
trentwatt: > Can you provide an example of how that would work?

we can just have the whole app in a loading state until the initial fetch?

> You don't actually want to "fall back" because that will cause an incorrect state

that's what I meant by "janky loading." i agree that it's not what we want but imo is preferable to a page crash. i think bringing in loadable would be good, but this particular PR is big as is.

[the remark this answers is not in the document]
trentwatt: @ClaireNeveu I think our previous message might have crossed paths in flight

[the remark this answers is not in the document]
trentwatt: > https://beta.reactjs.org/apis/react/Suspense

@keita-determined  thanks for linking, need to freshen up on suspense on a bit. can you elaborate on what in particular you had in mind though?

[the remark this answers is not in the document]
ClaireNeveu: Unless I'm missing something, suspense only works with React.lazy which isn't applicable here.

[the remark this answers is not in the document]
keita-determined: `Suspense` works with api call. when API is not resolved yet, `Suspense` will fall back. 
When API is resolved `suspense` will show the child components. 
The link i sent doesnt fully covered all cases
prob this one is better https://dev.to/darkmavis1980/a-practical-example-of-suspense-in-react-18-3lln

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.tsx
ClaireNeveu: Why is it `?` in the settings type?

ClaireNeveu: (I see that this is a pre-existing issue, just wondering if you knew why)

keita-determined:
some issues i found

- [ ] `Unable to fetch experiments.` is popping up when coming back to `uncategorized` page from other pages
- [ ] Table view doesnt display properly
  - image

- [ ] query params aren't in url after changing the page 
- [ ] `Clear Filter` doesnt clear filters in the first press
- [ ] loading regression 

  https://user-images.githubusercontent.com/109113616/198100680-baa4ff75-9fee-4161-960f-70a1e393eb63.mov

trentwatt (commented):
Noticing some of the above as well. I'm rolling those in to another checklist below. @thiagodallacqua-hpe can you please not modify this list (including checking things off-- it is for me to be able to track when I can approve). Once you can resolve the new ones, I can go through and validate.


New (from Keita):

- [ ] `Unable to fetch experiments.` is popping up when coming back to `uncategorized` page from other pages (even though `experiments` API is not being called)
- [ ] Tasks table empty
- [ ] query params disappear from url when navigating away and coming back
- [ ] `Clear Filter` doesnt clear filters in the first press and/or takes a while

Other New

- [ ] pagination says `< 1   Infinity >` and `Page 0` on model registry page (see below)
- [ ] table controls dont work (see below)
- [ ] sometimes get `&&` in the settings url


Previously listed:

- [ ] model registry page empty even when models exist
- [ ] skipUrlEncoding params show up in url
- [ ] loading state triggered on table row selection
- [ ] toggling sort asc/sort desc on tasks page results in queryParams showing up multiple times in url (see below url)
- [ ] tableOffset=NaN in url (see below)
- [ ] experiments/checkpoints page doesnt load
- [ ] clear filters button not showing up when filters present
- [ ] page crash on navigating to projects/:id/trials with f_trials_comparison=on



Blocking code considerations:
- [ ] get rid of `BaseType`
- [ ] have solution for the problem of `null` settings that does not rely on doing `settings?.x?.y` in all the appropriate places. e.g. having setting be required instead of `required | undefined`, put loading state higher up in the tree


url duped and nan
```
tasks?sortDesc=false&tableLimit=0&tableOffset=NaN&tableLimit=0&tableOffset=NaN
```

bad pagination info
image


table controls dont work

https://user-images.githubusercontent.com/28383080/198103459-31a08657-308b-4383-8487-420b52b40447.mov

"failed to fetch experiments" without request

https://user-images.githubusercontent.com/28383080/198103894-6f595e53-a267-47bf-97d8-90f9fd808030.mov

keita-determined (commented):
tested on Chrome with secret mode
I couldnt load page and some test cases failed

https://user-images.githubusercontent.com/109113616/198361825-13a2c0f3-4229-4438-a918-86f29c3a546c.mov

ClaireNeveu:
@keita-determined You might need to `npm install`

keita-determined:
> @keita-determined You might need to `npm install`

i did and i can reproduce it on Netlify so its not npm issue

ClaireNeveu:
Oh, looks to be Chrome specific and actually works in Firefox.

keita-determined:
> Oh, looks to be Chrome specific and actually works in Firefox.

I tested on Firefox now and its not working as well. Its working cuz your browser has prev cache or localstorage data. 
It wouldnt work on secret mode

webui/react/src/components/LogViewer/LogViewerFilters.settings.ts
```diff
@@ -10,48 +12,53 @@ export interface Settings {
...
-      key: 'allocationId',
-      type: {
-        baseType: BaseType.String,
-        isArray: true,
-      },
+const config: SettingsConfig<Settings> = {
+  applicableRoutespace: 'log-viewer-filters',
+  settings: {
+    agentId: {
+      defaultValue: undefined,
```
keita-determined, line 19: question: JSON doesnt support `undefined` value. How undefined data is stored in db?

webui/react/src/components/SelectFilter.tsx
```diff
@@ -84,7 +84,7 @@ const SelectFilter: React.FC<Props> = forwardRef(function SelectFilter(
      * - string
      * - string[]
      */
-    let label: any = null;
+    let label: null | string = null;
```
keita-determined, line 87: nit: `string | null` looks better

webui/react/src/components/Table/InteractiveTable.stories.tsx
```diff
@@ -28,32 +29,28 @@ const columns = new Array(20).fill(null).map(() => {
...
-        isArray: true,
-      },
+      type: array(number),
     },
-    {
-      key: 'row',
-      type: { baseType: BaseType.String, isArray: true },
+    row: {
+      defaultValue: [],
+      storageKey: 'row',
```
keita-determined, line 50: as I commented before, we can remove `storageKey` if its always the same as key name.

webui/react/src/hooks/useModal/JupyterLab/useModalJupyterLab.test.tsx
```diff
@@ -52,12 +60,14 @@ const setup = async () => {
...
-        <ModalTrigger />
+        <SettingsProvider>
+          <ModalTrigger />
+        </SettingsProvider>
       </StoreProvider>
     </BrowserRouter>,
   );
 
-  await user.click(screen.getByRole('button'));
+  await waitFor(() => user.click(screen.getByRole('button')));
```
keita-determined, line 70: i think `findByRole` would be better

webui/react/src/hooks/useSettingsProvider.tsx
```diff
@@ -0,0 +1,105 @@
+import { Map } from 'immutable';
```
keita-determined, line 1: question: what's the advantage of using Map from `immutable`?
The map just becomes immutable? Whats the difference between them in terms of behavior?

webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.test.tsx
```diff
@@ -112,24 +132,24 @@ describe('CodeViewer', () => {
   afterAll(() => jest.clearAllMocks());
 
   it('should handle the initial render properly', async () => {
-    setup();
+    await waitFor(() => setup());
```
keita-determined, line 135: we can get rid of `waitFor`

webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.test.tsx
```diff
@@ -112,24 +132,24 @@ describe('CodeViewer', () => {
...
     const { treeNodes } = await getElements();
 
     expect(treeNodes).toHaveLength(4);
   });
 
   it('should handle clicking in the download icon when opening a file from the tree', async () => {
     const pathBuilderSpy = jest.spyOn(paths, 'experimentFileFromTree').mockReturnValueOnce('');
-    setup();
+    await waitFor(() => setup());
```
keita-determined, line 143: here too

webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.test.tsx
```diff
@@ -112,24 +132,24 @@ describe('CodeViewer', () => {
...
-    await waitFor(() => act(() => user.click(treeNodes[1])));
+    await act(() => user.click(treeNodes[1]));
 
     const button = await screen.findByLabelText('download');
 
-    await waitFor(() => user.click(button));
+    await act(() => user.click(button));
 
-    expect(pathBuilderSpy).toHaveBeenCalledWith(123, 'single-in-records.yaml');
+    waitFor(() => expect(pathBuilderSpy).toHaveBeenCalledWith(123, 'model_def.py'));
```
keita-determined, line 153: missing `await` before `waitFor`

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.settings.ts
```diff
@@ -39,66 +41,81 @@ export interface Settings extends InteractiveTableSettings {
...
+        ),
+      ]),
     },
-    {
+    tableLimit: {
       defaultValue: MINIMUM_PAGE_SIZE,
-      key: 'tableLimit',
       storageKey: 'tableLimit',
-      type: { baseType: BaseType.Integer },
+      type: number,
```
keita-determined, line 110: not sure if its doable in this case, but union type would be good? 
like `10 | 20 | 50 | 100`

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -108,15 +107,19 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
-      <TableFilterDropdown
-        {...filterProps}
-        multiple
-        values={settings.state}
-        onFilter={handleStateFilterApply}
-        onReset={handleStateFilterReset}
-      />
-    ),
+    (filterProps: FilterDropdownProps) => {
+      if (!settings.state) return <Spinner spinning />;
```
keita-determined, line 111: Does that mean we keep having spinner if state is not defined?
I think state is fixed value, so idk if we need this

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.tsx
```diff
@@ -341,31 +331,35 @@ const ExperimentCheckpoints: React.FC<Props> = ({ experiment, pageRef }: Props)
...
+          <InteractiveTable
+            columns={columns}
+            containerRef={pageRef}
+            ContextMenu={CheckpointActionDropdown}
+            dataSource={checkpoints}
+            loading={isLoading}
+            pagination={getFullPaginationConfig(
+              {
+                limit: settings.tableLimit ?? 0,
+                offset: settings.tableOffset ?? 0,
```
keita-determined, line 344: `?? 0` means we cant fetch anything if `tableLimit` is undefined. I think we should find other way to treat this

webui/react/src/pages/ExperimentDetails/ExperimentDetails.test.tsx
```diff
@@ -72,7 +81,7 @@ describe('Experment Details Page', () => {
     });
 
     it('should show invalid experiment page without id', async () => {
-      setup();
+      await waitFor(() => setup());
```
keita-determined, line 84: we can remove `waitFor`

webui/react/src/pages/ExperimentDetails/ExperimentDetails.test.tsx
```diff
@@ -87,7 +96,7 @@ describe('Experment Details Page', () => {
     });
 
     it('should show experiment is unfetchable', async () => {
-      setup();
+      await waitFor(() => setup());
```
keita-determined, line 99: we can remove `waitFor`

webui/react/src/pages/ExperimentDetails/ExperimentDetails.test.tsx
```diff
@@ -109,7 +118,8 @@ describe('Experment Details Page', () => {
     });
 
     it('should show single trial experiment page with id', async () => {
-      const { container } = setup().view;
+      const containerView = await waitFor(() => setup());
```
keita-determined, line 121: we can remove `waitFor`

webui/react/src/pages/ExperimentDetails/ExperimentDetails.test.tsx
```diff
@@ -139,7 +149,8 @@ describe('Experment Details Page', () => {
     });
 
     it('should show multi-trial experiment page with id', async () => {
-      const { container } = setup().view;
+      const containerView = await waitFor(() => setup());
```
keita-determined, line 152: same here

webui/react/src/pages/ExperimentDetails/ExperimentTrials.settings.ts
```diff
@@ -56,70 +58,104 @@ export interface Settings extends InteractiveTableSettings {
...
-      key: 'state',
+    state: {
+      defaultValue: undefined,
       storageKey: 'state',
-      type: {
-        baseType: BaseType.String,
-        isArray: true,
-      },
+      type: union([
+        undefinedType,
```
keita-determined, line 124: how come we need `undefined`? I think you respected the previous implementation, but if we dont need it we can removed it. we can represent undefined as empty array unless we use undefined on purpose

webui/react/src/pages/ExperimentDetails/ExperimentTrials.tsx
```diff
@@ -430,8 +432,8 @@ const ExperimentTrials: React.FC<Props> = ({ experiment, pageRef }: Props) => {
           loading={isLoading}
           pagination={getFullPaginationConfig(
             {
-              limit: settings.tableLimit,
-              offset: settings.tableOffset,
+              limit: settings.tableLimit ?? 0,
```
keita-determined, line 435: we should find other way to treat it instead of `?? 0`

webui/react/src/pages/OldProjectDetails.tsx
```diff
@@ -908,19 +903,8 @@ const ProjectDetails: React.FC = () => {
...
-    settings.search,
-    settings.sortDesc,
-    settings.sortKey,
-    statesString,
-    pinnedString,
-    settings.tableLimit,
-    settings.tableOffset,
-    usersString,
-  ]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
keita-determined, line 906: i saw many lines of this code
Is it fine to do it in these cases? https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528 might be helpful.

webui/react/src/pages/Settings/UserManagement.test.tsx
```diff
@@ -61,29 +82,35 @@ const setup = () =>
...
     await waitFor(() => setup());
 
+    await waitFor(() => jest.setTimeout(300));
+
     expect(screen.getByText(CREATE_USER)).toBeInTheDocument();
     expect(screen.getByText(USER_TITLE)).toBeInTheDocument();
-    expect(screen.getByText(DISPLAY_NAME)).toBeInTheDocument();
-    expect(screen.getByText(USERNAME)).toBeInTheDocument();
+    // expect(screen.getByText(DISPLAY_NAME)).toBeInTheDocument();
+    // expect(screen.getByText(USERNAME)).toBeInTheDocument();
```
keita-determined, line 106: How come you comment out these lines?
I dont think its good to remove it to pass the test cases.

webui/react/src/pages/Settings/UserManagement.test.tsx
```diff
@@ -61,29 +82,35 @@ const setup = () =>
...
   });
 
   it('should render modal for create user when click the button', async () => {
     await waitFor(() => setup());
+    await waitFor(() => jest.setTimeout(300));
     await user.click(screen.getByLabelText(CREAT_USER_LABEL));
 
-    expect(screen.getAllByText('Create User')).toHaveLength(2);
+    expect(screen.getAllByText('New User')).toHaveLength(1);
```
keita-determined, line 114:
```suggestion
    setup();
    await user.click(await screen.findByLabelText(CREAT_USER_LABEL));

    expect(screen.getAllByText('New User')).toHaveLength(1);
```
you can write like this

webui/react/src/pages/TrialsComparison/Table/TrialTable.tsx
```diff
@@ -142,7 +143,7 @@ const TrialTable: React.FC<Props> = ({
         <TableFilterRank
           {...filterProps}
           column={filters.ranker?.sorter.sortKey}
-          columns={settings.columns.filter((col) => !nonRankableColumns.includes(col))}
+          columns={settings.columns.filter((col) => !nonRankableColumns.includes(col)) ?? []}
```
keita-determined, line 146: Do we need to change it here?
I didnt error on my local

keita-determined (commented):
I think its getting better.
I found some bugs. Also, will we go with rendering all components when settings value is changed?

I saw lots of `waitFor` in test cases. mostly we dont need to use it

query string is still in the url after reset

https://user-images.githubusercontent.com/109113616/199074792-65784b4a-dadc-4fa4-bc3d-7d5f820c8173.mov

---

Cannot re-order

https://user-images.githubusercontent.com/109113616/199074813-aea24c27-65fe-4697-89d4-ab3eee223d68.mov

---

render all components every input

https://user-images.githubusercontent.com/109113616/199074819-8a7aea22-e7a3-49d7-89ba-df1bb79ffbac.mov

---

State shows spinner and not show dtata

Screen Shot 2022-10-31 at 9 51 10 AM

webui/react/src/pages/OldProjectDetails.tsx
ClaireNeveu: I very much disagree with this comment. It doesn't seem to account for non-local dependencies. The dependency array exists so that you can manage what the effect is subscribed to.

webui/react/src/hooks/useSettingsProvider.tsx
ClaireNeveu: It's more performant for immutable usecases (which this is). Mutably updating a map in react state introduces potential racing issues.

webui/react/src/components/Table/InteractiveTable.stories.tsx
ClaireNeveu: Let's leave this as a follow-up PR, Keita

webui/react/src/pages/ExperimentDetails/ExperimentCheckpoints.settings.ts
ClaireNeveu: We should definitely make a little `enum` helper function for io-ts for union of literal types.

webui/react/src/hooks/useSettingsProvider.tsx
keita-determined: How about using [`Immer`](https://www.npmjs.com/package/immer) instead of `immutable` then? I think its pretty much the same for our usage, and its smaller bundle and more active on github

ClaireNeveu: I don't know much about immer but immutable.js has less github activity because it's stable. If you'd like to evaluate different datatype libraries add a ticket and maybe we can swap them out. For the purpose of this PR immutable works and it's been the de-facto standard for a while.

ClaireNeveu:
@keita-determined I can't reproduce your first two issues on the latest netlify preview.

keita-determined:
> @keita-determined I can't reproduce your first two issues on the latest netlify preview.

I tried again and could reproduce them on local and netlify
My env is on Chrome with private mode.
the first one is on `http://localhost:3000/det/models` not the `/projects`
the second one is on `http://localhost:3000/det/resourcepool/default/active`

webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.test.tsx
keita-determined: you need `await` before `waitFor`

keita-determined:
> @keita-determined about the "render all components every input" issue, that's due to the fact that the `settings` is part of the dependencies of many things there, from callbacks to field values... so, it makes sense that it re-renders on every change... as it is a small issue (that I do agree that we could look at), let's split it into a separate ticket perhaps...
> 
> about the re-ordering one, I couldn't reproduce it here, maybe it got fixed by other changes, just in case, I've asked @EmilyBonar to reproduce it...

The reason why all components are rendered is that we use context. 
Ideally, each setting should be independent by pages or components

keita-determined (commented):
Still some issues. Also reordering columns issue is not fixed 
https://user-images.githubusercontent.com/109113616/199559642-857502f7-de1f-4324-8554-99c3c73c92a1.mov

image

https://user-images.githubusercontent.com/109113616/199560070-3595c961-a004-4b2e-b316-8e12273a1815.mov

keita-determined:
@thiagodallacqua-hpe 
the reorder issue is on `resourcepool` page. i can still reproduce it

ClaireNeveu:
Pretty sure the reason that table is broken is because it calls `updateSettings` [in the render function](https://github.com/determined-ai/determined/blob/master/webui/react/src/pages/JobQueue/JobQueue.tsx#L242) (not in an effect or callback). This worked previously because we were re-using the Map reference but now that the Map reference is properly updated it just constantly resets the settings back to the initial state.

Render functions shouldn't have side-effects so we'll need to update ResourcepoolDetails to fix that bug.

ClaireNeveu:
I though the issues Keita was pointing to in his last comment was that some settings were being improperly serialized to the URL?

ClaireNeveu:
@thiagodallacqua-hpe I mean that settings like `row` and `column` are not supposed to be in the url, I thought

keita-determined (commented):
some issues

I still see `columns` in workspace `http://localhost:3000/det/workspaces?archived=true&columns=name&columns=numProjects&columns=userId&columns=archived`

---

selection wouldn't be reset after the action

https://user-images.githubusercontent.com/109113616/200391058-a69d6d1e-affb-449f-83d5-be4b0bb0064c.mov

---

- It's slow to select items
- Cannot select workspace in the first time
- Workspace selection is not reflected after the second time

https://user-images.githubusercontent.com/109113616/200391598-07d59678-7bde-430b-af66-65d47197f7df.mov

PR #5193, opened 2022-10-05
feat: support -i/--include [MLG-194]

rb-determined-ai:
## Description

This is especially useful for including just one or two files in
interactive workflows without having to create an empty context
directory and copy files into it.

## Test Plan

- [x] write automated tests
- [x] try `det e create const.yaml emtpy-dir/ -i model_def.py`
- [x] try `det notebook start -i myfile`
- [x] try `det tensorboard start -i myfile`
- [x] try `det command run -i myfile cat myfile`
- [x] try `det shell start -i myfile` 

## Commentary (optional)

This PR began life as #5087.

### DIFF
--- a/harness/determined/common/experimental/determined.py
+++ b/harness/determined/common/experimental/determined.py
@@ -1,6 +1,6 @@
 import pathlib
 import warnings
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 from determined.common import api, context, util, yaml
 from determined.common.api import authentication, bindings, certs
@@ -68,6 +68,7 @@ class Determined:
         self,
         config: Union[str, pathlib.Path, Dict],
         model_dir: Union[str, pathlib.Path],
+        includes: Optional[Iterable[Union[str, pathlib.Path]]] = None,
     ) -> experiment.ExperimentReference:
         """
         Create an experiment with config parameters and model directory. The function
@@ -77,6 +78,8 @@ class Determined:
             config(string, pathlib.Path, dictionary): experiment config filename (.yaml)
                 or a dict.
             model_dir(string): directory containing model definition.
+            iterables (Iterable[Union[str, pathlib.Path]], optional): Additional files or
+                directories to include in the model definition.  (default: ``None``)
         """
         if isinstance(config, str):
             with open(config) as f:
@@ -96,7 +99,8 @@ class Determined:
         if isinstance(model_dir, str):
             model_dir = pathlib.Path(model_dir)
 
-        model_context = context.read_v1_context(model_dir)
+        path_includes = (pathlib.Path(i) for i in includes or [])
+        model_context = context.read_v1_context(model_dir, includes=path_includes)
 
         req = bindings.v1CreateExperimentRequest(
             # TODO: add this as a param to create_experiment()
--- a/harness/determined/experimental/client.py
+++ b/harness/determined/experimental/client.py
@@ -46,7 +46,7 @@ See :ref:`use-trained-models` for more ideas on what to do next.
 import functools
 import pathlib
 import warnings
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 from determined.common.api import Session  # noqa: F401
 from determined.common.experimental.checkpoint import (  # noqa: F401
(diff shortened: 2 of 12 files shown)


garrett361:
Noticed the following behavior we might want to change: say my dir structure looks like
```
.
├── d1
│   └── f1
└── d2
    └── f2
```

If I call, say `det notebook start -i d1 -i d2` then I get the expected result:
```
workdir
├── d1
│   └── f1
└── d2
    └── f2
```

But if I now tried to do the same with two `--context` calls via `det notebook start -c d1 -c d2`, only the second is applied and you get:
```
workdir
└── f2
```
so it's basically failing silently.  Given that you will be able to have multiple `--include`s, you might also expect to be able to multiple `--context`s.

Should we raise an error upon multiple `-c` calls or support them?

rb-determined-ai:
Great review, thank you.

Yes, the conflict detection seems not to be working right for directories, it looks like a couple of your reports probably have that as the root cause (I'll fix that).

The issue with multiple `--context` directories might be harder to fix, since what you are seeing is normal command-line mechanics, where the last flag overrides previous instances of that flag in cases where only one instance of the flag will be respected.

garrett361:
> The issue with multiple --context directories might be harder to fix, since what you are seeing is normal command-line mechanics, where the last flag overrides previous instances of that flag in cases where only one instance of the flag will be respected.

Ah cool, I was just going to look into that; wasn't sure if that was standard behavior.  That seems like a funny default.

rb-determined-ai:
> That seems like a funny default.

It makes sense in a world where wrapper scripts are common.  E.g:

```sh
# I basically always want some flags (say, strict aliasing and overflow checks)
alias my_gcc='gcc -Wstrict-aliasing -Wstrict-overflow'

...
# except this one case where I don't want strict overflow
my_gcc -Wno-strict-overflow -o my_bin my_file.c
```

rb-determined-ai:
The root cause was forgetting to add the root included directory into the tarfile (which is technically a malformed tarball, but one that many tar unpackers will tolerate), and it wasn't caught by my tests because my tests were just wrong.

Fixed everything, and added a bit more test coverage for good measure.

(didn't fix the `--context` because that is normal cli behaviors, unfortunate though it may be in this case)

garrett361:
Do we also need to add an `include` arg in the `create_experiment` function in [common/api/experiment](https://github.com/determined-ai/determined/blob/cfdd7bb0d8bc7e5013c01a0da3a38a0e83968cdf/harness/determined/common/api/experiment.py#L121)? Not entirely sure how and where that `create_experiment` is used.

garrett361:
> (didn't fix the --context because that is normal cli behaviors, unfortunate though it may be in this case)
Sounds good, probably what the user would expect then.  If someone really wants it down the line, I suppose we could allow for multiple `--context` calls in addition to multiple `--include`s

rb-determined-ai:
> Do we also need to add an include arg in the create_experiment function in [common/api/experiment](https://github.com/determined-ai/determined/blob/cfdd7bb0d8bc7e5013c01a0da3a38a0e83968cdf/harness/determined/common/api/experiment.py#L121)? Not entirely sure how and where that create_experiment is used.

Yes, that's a great idea.  I'll add it.

harness/determined/experimental/client.py
```diff
@@ -143,18 +143,21 @@ def login(
 def create_experiment(
```
garrett361, line 143: So, I see that the `includes` arg was added here, which is nice, but I was asking about whether it should also be included in [common/api/experiment.py](https://github.com/rb-determined-ai/determined/blob/9eab0f7d18565a66f7b739b073d8bf14879e20c4/harness/determined/common/api/experiment.py#L121)? I am again confused about when this latter version of `create_experiment` is used.

harness/determined/common/experimental/determined.py
```diff
@@ -68,6 +68,7 @@ def create_experiment(
         self,
         config: Union[str, pathlib.Path, Dict],
         model_dir: Union[str, pathlib.Path],
```
garrett361, line 70: It looks like are currently using `model_dir` here, `model_context` in [common/api/experiment.py](https://github.com/rb-determined-ai/determined/blob/9eab0f7d18565a66f7b739b073d8bf14879e20c4/harness/determined/common/api/experiment.py#L124), and `--context` in the CLI to refer to the same idea in all three cases.

Up to you whether standardizing nomenclature is worth making a breaking change.

rb-determined-ai: all of `common` is an internal and therefore irrelevant, from a user-facing API perspective.

Yeah it's a bit of an unfortunate name but we're not going to break it over this.  Perhaps when we go from experimental to GA we could consider making that change at the same time.  Feel free to make a ticket for it and mark it as "breaking change", I think it's a good idea.

garrett361: Ok cool, thanks for explaining. That was my sense of what `common` was, but I wasn't sure.

garrett361 (approved):
Looking good, will be very useful to me personally! Great work and thanks for all the back and forth.

rb-determined-ai:
no, thank _you_ for such a careful and well-reasoned review (and excellent feature request)

PR #5195, opened 2022-10-05
chore: add custom permission [DET-8526]

eecsliu:
## Description
`mapAndFilterErrors` filters out a bunch of errors - this creates a custom error that specifies needed permissions and is passed through the filter function.

## Checklist


master/internal/authz/permission_denied_error.go
```diff
@@ -0,0 +1,28 @@
...
+func (p PermissionDeniedError) Error() string {
+	if len(p.RequiredPermissions) == 0 {
+		return "access denied"
+	}
+
+	permissions := make([]string, len(p.RequiredPermissions))
+	for i, perm := range p.RequiredPermissions {
+		permissions[i] = perm.Name
+	}
+	return fmt.Sprintf("access denied; required permissions: %s", strings.Join(
```
NicholasBlaskey, line 26: should we say what scope you need it on / or globally?

In the case of ```moveProject``` you need a delete on one workspace and a create on another so just saying ```required permissions: DELETE, CREATE``` isn't as specific as we could be

master/internal/authz/permission_denied_error.go
```diff
@@ -0,0 +1,28 @@
...
+// PermissionDeniedError represents an error that arises when a user does not have sufficient
+// access privileges. RequiredPermissions can be empty for non-rbac errors.
+type PermissionDeniedError struct {
+	RequiredPermissions []*rbacv1.Permission
+}
+
+// Error returns an error string.
+func (p PermissionDeniedError) Error() string {
+	if len(p.RequiredPermissions) == 0 {
+		return "access denied"
```
NicholasBlaskey, line 19: for OSS stuff it would be cool if we could pass in a specific message

It would feel frustrating to just get access denied and not really sure what went wrong where

master/internal/authz/permission_denied_error.go
eecsliu: I had a chat with Max about this yesterday and the consensus was that you know what operation you're performing and on what scope(s), so you would need to know what permissions you require to carry out that operation.

master/internal/authz/permission_denied_error.go
eecsliu: Maybe it's best if we had a meeting together tomorrow in the AM to discuss more, but I did talk about this also with Max, and adding a custom message could be subverting the purpose of the `mapAndFilterErrors` function, which was to only let certain errors through.

NicholasBlaskey (approved):

PR #5196, opened 2022-10-05
fix: restore det tensorboard start

phillip-gaisford:
## Test Plan

PR-5150 adversely affects the loading of experiment configurations created prior to the change. Restore that capability by allowing slurm & PBS in the environment schema, but ignore any values there.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -427,6 +427,9 @@ commands:
       extra-pytest-flags:
         type: string
         default: ""
+      wait-for-master:
+        type: boolean
+        default: true
     steps:
       - run:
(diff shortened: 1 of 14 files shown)


jerryharrow (approved):
lgtm.

Another potential option would be to provide a migration to remove the offending value from the DB table, but trying to pull that out of JSON embedded in the DB doesn't seem easy -- unless there is already infrastructure to do that.

stoksc:
can we close this? i think we fixed with migrations, right?

jerryharrow:
Fixed by adding proper migrations to the DB.

PR #5197, opened 2022-10-05
fix: avoid using lookbehind regex for Safari

keita-determined:
## Description
[WEB-286]

## Test Plan
- Check if files are sorted as expected 
- Check if `Code` tab (`det/experiments/<expId>/code`) is visible on Safari

## Commentary (optional)
https://caniuse.com/js-regexp-lookbehind

## Checklist

[WEB-286]: https://determinedai.atlassian.net/browse/WEB-286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.tsx
+++ b/webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.tsx
@@ -75,8 +75,8 @@ const sortTree = (a: TreeNode, b: TreeNode) => {
   if (!a.isLeaf && !b.isLeaf) return titleA.localeCompare(titleB) - titleB.localeCompare(titleA);
 
   // had to use RegEx due to some files being ".<filename>"
-  const [stringA, extensionA] = titleA.split(/(?<=[a-zA-Z])\./);
-  const [stringB, extensionB] = titleB.split(/(?<=[a-zA-Z])\./);
+  const [stringA, extensionA] = titleA.split(/^(?=[a-zA-Z])\./);
+  const [stringB, extensionB] = titleB.split(/^(?=[a-zA-Z])\./);
 
   if (!extensionA && extensionB) return 1;
 


webui/react/src/pages/ExperimentDetails/CodeViewer/CodeViewer.tsx
```diff
@@ -75,8 +75,8 @@ const sortTree = (a: TreeNode, b: TreeNode) => {
   if (!a.isLeaf && !b.isLeaf) return titleA.localeCompare(titleB) - titleB.localeCompare(titleA);
 
   // had to use RegEx due to some files being ".<filename>"
-  const [stringA, extensionA] = titleA.split(/(?<=[a-zA-Z])\./);
-  const [stringB, extensionB] = titleB.split(/(?<=[a-zA-Z])\./);
+  const [stringA, extensionA] = titleA.split(/^(?=[a-zA-Z])\./);
+  const [stringB, extensionB] = titleB.split(/^(?=filename)\./);
```
trentwatt, line 79: i think `filename` is a typo? what do you think about just changing from here to the bottom of this function to the following?
```
  const _a = titleA.split('.').reverse().join('.')
  const _b = titleB.split('.').reverse().join('.')
  return _a.localeCompare(_b)
```

the desired behavior is that we sort by extension, then by filename. 

you can check if it has the desired behavior [here](https://replit.com/@trentwatt/CuddlyGlamorousMigration#index.ts)

keita-determined: The code you provide looks good to me with plane directory. I haven't tried it with nested directories, so I'll merge this one first to display `Code` tab on Safari, but this would be nice to change.

PR #5198, opened 2022-10-06
fix: page transition in multi trial page

keita-determined:
## Description
https://determined-ai.slack.com/archives/CLCE8D998/p1665015007431779

## Test Plan
- Check if page will change by clicking `previous` or `next` trial arrow button

## Checklist

### DIFF
--- a/webui/react/src/pages/TrialDetails.tsx
+++ b/webui/react/src/pages/TrialDetails.tsx
@@ -50,7 +50,7 @@ const TrialDetailsComp: React.FC = () => {
   const navigate = useNavigate();
   const { experimentId, tab, trialId: trialID } = useParams<Params>();
   const [tabKey, setTabKey] = useState<TabType>(tab ?? DEFAULT_TAB_KEY);
-  const [trialId, setTrialId] = useState<number>(parseInt(trialID ?? ''));
+  const [trialId, setTrialId] = useState<number>(Number(trialID));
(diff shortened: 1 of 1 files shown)


webui/react/src/pages/TrialDetails.tsx
```diff
@@ -50,7 +50,7 @@ const TrialDetailsComp: React.FC = () => {
   const navigate = useNavigate();
   const { experimentId, tab, trialId: trialID } = useParams<Params>();
   const [tabKey, setTabKey] = useState<TabType>(tab ?? DEFAULT_TAB_KEY);
-  const [trialId, setTrialId] = useState<number>(parseInt(trialID ?? ''));
+  const [trialId, setTrialId] = useState<number>(Number(trialID));
```
keita-determined, line 53: changed it to `Number` cuz theoretically its faster and we always expect numeric value from `trialID`

garrett361 (approved):
Tested locally using `devcluster`.

To confirm, the Previous Trial and Next Trial buttons are only intended to allow moving between Trials contained in a single multi-Trial Experiment, correct? Approved, assuming that the preceding behavior is right.

keita-determined:
> To confirm, the Previous Trial and Next Trial buttons are only intended to allow moving between Trials contained in a single multi-Trial Experiment, correct?

Yes,  I think so. Thanks

PR #5200, opened 2022-10-06
feat: Add CAN_EDIT_WEBHOOKS permission to pre-canned admin role [WEB-218]

julian-determined-ai:
## Description

Adds `EDIT WEBHOOKS` permissions for RBAC. All `webhooks` API endpoints now check for the correct permissions. The related `ee` PR can be found here: https://github.com/determined-ai/determined-ee/pull/471

## Test Plan

Same test plan as: https://github.com/determined-ai/determined/pull/5175, ensure that everything still works. 

## Checklist


master/internal/webhooks/authz_basic_impl.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// WebhookAuthZBasic is basic OSS controls.
+type WebhookAuthZBasic struct{}
+
+// CanEditWebhooks always returns true and a nil error.
+func (a *WebhookAuthZBasic) CanEditWebhooks(
+	curUser *model.User,
+) (canEditWeb bool, serverError error) {
+	return true, nil
```
NicholasBlaskey, line 14: Not exactly sure but feels more OSS permission style to require admin here?

master/internal/webhooks/api_webhook.go
```diff
@@ -3,16 +3,40 @@ package webhooks
...
+func AuthorizeRequest(ctx context.Context) error {
+	curUser, _, err := grpcutil.GetUser(ctx)
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to get the user: %s", err)
+	}
+	if ok, authErr := AuthZProvider.Get().
+		CanEditWebhooks(curUser); authErr != nil {
+		return authErr
+	} else if !ok {
+		return status.Error(codes.PermissionDenied, err.Error())
```
NicholasBlaskey, line 27: this ```err``` will always be nil here since it is the one from the ```grpc.GetUser```

I think it is fine to make ```AuthZProvider.Get().CanEditWebhooks``` only return a error then if that error is non nil return it as ```status.Error(codes.PermissionDenied, err.Error())```  

Just for some context why some ```authz``` interfaces returns booleans and some just returns errors, ```GetExperiment``` returns ```(canGetExperiment, serverError)``` because we want to return a 404 error when a user can't view an experiment to avoid people finding out about experiments that don't exist. The ```serverError``` is just in case some db failure occurs. In this case since we don't need to return a 404 when someone can't ```EditWebHooks``` since it is a all or nothing check either you can edit every webhook or none.

NicholasBlaskey (approved):
LGTM couple comments but looks good

master/internal/webhooks/authz_basic_impl.go
julian-determined-ai: You are correct this has been updated so only admins can manage webhooks.

master/internal/webhooks/api_webhook.go
julian-determined-ai: Thanks for the info and suggestion! I have updated based on this feedback.

PR #5202, opened 2022-10-06
chore: remove accidentally added attributions file

dzhu:


mapmeld (approved):
Thanks for catching this

PR #5203, opened 2022-10-06
fix: hide uncategorized from nav, when it is not available to RBAC user [WEB-509]

mapmeld:
## Description

If an RBAC user has no permission to view the Uncategorized workspace (aka the default workspace/project) then that option should be hidden from navigation.  Other sections should be visible.

## Test Plan

- Visit `/det/models` and see Uncategorized in the left nav menu. With RBAC off there is no change
- Visit `/det/models?f_rbac=on` and see Uncategorized is hidden. In this setting with no permissions, it's hidden
- Visit `/det/models?f_rbac=on&f_mock_permissions_read=on` and Uncategorized reappears

## Checklist


julian-determined-ai (approved):
LGTM!

PR #5206, opened 2022-10-06
feat: user mgmt functions in SDK + new user API from old API + CLI uses SDK [DET-8495, DET-8496]

nrajanee:
## Description

feat: user mgmt functions in SDK + new user API from old API + CLI uses SDK


harness/determined/common/experimental/determined.py
```diff
@@ -64,6 +64,19 @@ def __init__(
 
         self._session = api.Session(master, user, auth, cert)
 
+    def create_user(username: str, password: str, admin: bool):
+        # call API create_user
+        # create User object
+        # return that object 
+        pass
+    
+    def get_user():
```
ioga, line 73: would be great to get a user by user_id or by username - i.e. support both because we do both often.

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(
+        self,
+        user_id: int, 
+        username: str,
+        password: str,
+        admin: bool,
+        session: api.Session,
+    ):
+        self.username = username
+        self.password = password
```
rb-determined-ai, line 35: What would actually be set in `User.password`?

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,93 @@
...
+        # can also get user from authentication.must_cli_auth().get_session_user()
+        # API bindings.patch_PatchUser need to add password to message PatchUser in user.proto
+        pass
+
+    def link_with_agent_user(self, agent_user_group: AgentUserGroup) -> None:
+        # calls update user with these args wrapped in agent_user_group.
+        pass
+
+
+    def whoami() -> str:
```
rb-determined-ai, line 91: This doesn't make sense here... it would need to be on the Determined object, and it should probably return a User object.

harness/determined/common/experimental/determined.py
```diff
@@ -64,6 +64,32 @@ def __init__(
...
+
+        pass
+
+    def get_user(user_id: int):
+        # return User object
+
+        # new API -> bindings.get_GetUser(user_id) (in api_user.go) ++ need to do username
+        pass
+
+    def get_user(user_name: str):
```
rb-determined-ai, line 82: python doesn't support overriding like this, you'll either need to accept `Union[int, str]` in `get_user()` or you'll need `get_user_by_id()` and `get_user_by_name()`

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,93 @@
...
+        # calls update_user with active = true
+        # API -> bindings.patch_PatchUser(self.user_id, patchUser)
+        pass
+
+    def deactivate_user(self) -> None:
+        # calls update_user with active = false
+        # bindings.patch_PatchUser(self.user_id, patchUser) API
+        pass
+
+    def deactivate_user(self) -> None:
```
rb-determined-ai, line 76: duplicate?

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,93 @@
...
+        self.admin = admin
+        self.user_id = user_id
+        self.active = True
+        self.agent_uid = None
+        self.agent_gid = None
+        self.agent_user = None
+        self.agent_group = None
+        self.session = session
+
+    def update_user(
```
rb-determined-ai, line 45: I'm not a fan of having two ways to do the same thing in one API.

I'm not opposed to a single `user.update(username=..., password=...)`, etc, but I don't think we should also have `user.update_username()` and `user.update_password()`.

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,93 @@
...
+        # return API response
+        pass
+     
+    def update_username(self, new_username: str) -> Response:
+        # return API response
+        # API: bindings.patch_PatchUser(self.userid, patchUser) API (need to add username to message PatchUser in user.proto)
+
+        pass
+
+    def activate_user(self) -> None:
```
rb-determined-ai, line 61: this has an extra `_user` in the name.  It would look weird in user code:
```python
user = client.get_user("rb")
user.activate_user()
```

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,58 @@
...
+        pass 
+
+    def rename(name_target_user: str, master_address: str, new_username: str) -> None: 
+        pass 
+
+    def change_password(new_password: str, username: Optional[str]) -> None: 
+        # can get user from authentication.must_cli_auth().get_session_user()
+        pass
+
+    def link_with_agent_user(agent_uid: uuid, agent_user: str, agent_gid: int, agent_group: str) -> None:
```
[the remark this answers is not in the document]
rb-determined-ai: I don't really care for AgentUserGroup as a user-facing object, I actually prefer the separate strings.

To me, a sign that 4 separate strings makes sense is the fact that the User object doesn't store an `AgentUserGroup` anyway.  So it's just a glorified parameter that gets torn apart after the call to `link_with_agent_user`, not a useful concept to actually expose to the user.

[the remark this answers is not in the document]
nrajanee: I agree that it doesn't make sense for AgentUserGroup to be a user-facing object. Maybe we can require that the users pass in a dictionary with the right information? Or that might make it more confusing because with parameters users will know this is exactly what they need to pass in.

harness/determined/common/experimental/user.py
ioga: I agree we can get rid of `update_username`.
we have `det user change-password` on the CLI, so maybe for the sake of alignment we could have `def change_password` here as well as a helpful shortcut?

nrajanee: Changing update_username to rename because that's what the CLI has.

package-lock.json
```diff
@@ -0,0 +1,1962 @@
+{
```
keita-determined, line 1: I think this file shouldn't be here

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "eslint": "^8.25.0"
```
keita-determined, line 3: I think this file shouldn't be here either

nrajanee: will delete it.

keita-determined: you should also remove `package-lock.json`

e2e_tests/tests/cluster/test_users_experiment_api.py
```diff
@@ -1,19 +1,22 @@
 import pathlib
+from telnetlib import AUTHENTICATION
```
ioga, line 2: this does not look right

harness/determined/cli/user.py
```diff
@@ -86,7 +58,6 @@ def log_out_user(parsed_args: Namespace) -> None:
     auth = authentication.cli_auth
     if auth is None:
         return
-
     try:
         api.post(
```
ioga, line 62: can we please update this logout call as well? and any other `api.post`, `api.get`, `api.patch` remaining in this file, if any.

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,103 @@
...
+        return resp
+
+    def change_display_name(self, display_name: str) -> bindings.v1PatchUserResponse:
+        patch_user = bindings.v1PatchUser(displayName=display_name)
+        patch_user_req = bindings.v1PatchUserRequest(userId=self.user_id, user=patch_user)
+        resp = bindings.patch_PatchUser(self.session, body=patch_user_req, userId=self.user_id)
+        return resp
+
+    def change_password(
+        self, new_password: str, is_hashed: Optional[bool] = False
```
ioga, line 82: let's not expose this `is_hashed` argument here — I don't think a user should bother themselves with this implementation detail. we should always just hash and send it hashed from the client. 
instead having an option, just accept unhashed password and call `api.salt_and_hash` on it here in the sdk, like the CLI is doing at the moment. right now there's a lot of `api.salt_and_hash` calls sprinkled in user cli, and I think this should be "internalized" within SDK / auth code. 

with this change, I wonder if `isHashed` parameter of the `PostUserRequest` and `PatchUserRequest` API requests would still be necessary.

harness/determined/common/experimental/determined.py
```diff
@@ -61,9 +61,139 @@ def __init__(
...
+            )
+        else:
+            return user.User(
+                user_id=resp.user.id,
+                username=resp.user.username,
+                admin=resp.user.admin,
+                session=self._session,
+                active=resp.user.active,
+                display_name=resp.user.displayName,
+            )
```
ioga, line 166: this block is repeated a few times. can we have it as a classmethod on `user.User` instead e.g.
```
@classmethod
def _from_bindings(raw: v1User):
    ...
```

proto/src/determined/api/v1/user.proto
```diff
@@ -57,12 +57,42 @@ message GetUserResponse {
...
+  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+    json_schema: { required: [ "user" ] }
+  };
+  // The requested user.
+  determined.user.v1.User user = 1;
+}
+
+// Get the requested current user.
+message GetMeRequest {
+  // Cannot have an empty message.
```
ioga, line 76: empty request here seems to work: https://github.com/determined-ai/determined/blob/master/proto/src/determined/api/v1/auth.proto#L33

harness/determined/common/experimental/user.py
nrajanee: As discussed in slack, I'm keeping the isHashed parameter for PostUserRequest and PatchUserRequest. In the CLI and SDK the password will always be hashed before making the API call. But previously the password in PostUserRequest (Even in SetUserPassword API) was assumed to be unhashed. So want to allow that flexibility.

harness/determined/common/experimental/determined.py
nrajanee: I wrote a method for this but I can't add it as a classmethod because self._session is a part of the instance.

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +265,107 @@ def test_logout(clean_auth: None) -> None:
...
 
     # Log out.
     assert log_out_user() == 0
 
-    newPassword = get_random_string()
-    assert change_user_password(creds.username, newPassword, ADMIN_CREDENTIALS) == 0
+    # login admin
+    log_in_user(ADMIN_CREDENTIALS)
 
+    newPassword = get_random_string()
```
ioga, line 325: linter would probably want this to be `new_password`

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +265,107 @@ def test_logout(clean_auth: None) -> None:
...
+    assert change_user_password(creds.username, newPassword) == 0
     assert log_in_user(authentication.Credentials(creds.username, newPassword)) == 0
 
+    newPasswordSdk = get_random_string()
```
ioga, line 329:
```suggestion
    new_password_sdk = get_random_string()
```

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +265,107 @@ def test_logout(clean_auth: None) -> None:
...
+    assert log_in_user(authentication.Credentials(creds.username, newPasswordSdk)) == 0
+
+
+@pytest.mark.e2e_cpu
+def test_change_username(clean_auth: None, login_admin: None):
+    creds = create_test_user()
+    new_username = "renameuser4"
+    print("new username")
+    print(creds.username)
+    print(new_username)
```
ioga, line 342: debug prints

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +265,107 @@ def test_logout(clean_auth: None) -> None:
...
+    command = ["det", "-m", conf.make_master_url(), "user", "rename", creds.username, new_username]
+    assert subprocess.call(command) == 0
+    det_obj = Determined(master=conf.make_master_url())
+    user = det_obj.get_user_by_name(user_name=new_username)
+    assert user.username == new_username
+    assert log_in_user(authentication.Credentials(new_username, "")) == 0
+
+    # Test SDK
+    new_username = "renameuser5"
```
ioga, line 351: can we sometimes throw in users with non-alphanumeric names? e.g. `rename-user-$5`
I recently discovered there was a bug prohibiting that *in rename only* so testing that would be nice.

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +265,107 @@ def test_logout(clean_auth: None) -> None:
...
+def test_experiment_creation_and_listing(clean_auth: None, login_admin: None) -> None:
     # Create 2 users.
-    creds1 = create_test_user(ADMIN_CREDENTIALS, True)
+    creds1 = create_test_user(True)
 
-    creds2 = create_test_user(ADMIN_CREDENTIALS, True)
+    creds2 = create_test_user(True)
+
+    # Ensure determined creds are the default values.
+    change_user_password("determined", "")
```
ioga, line 368: why do this here?

e2e_tests/tests/cluster/test_users.py
```diff
@@ -351,60 +396,61 @@ def test_experiment_creation_and_listing(clean_auth: None) -> None:
...
     assert child.exitstatus != 0
 
 
 @pytest.mark.e2e_cpu
-def test_login_with_environment_variables(clean_auth: None) -> None:
-    creds = create_test_user(ADMIN_CREDENTIALS, True)
+def test_login_with_environment_variables(clean_auth: None, login_admin: None) -> None:
+    creds = create_test_user(True)
+    # logout admin
+    assert log_out_user() == 0
```
ioga, line 438: it seems a bit annoying to do `assert log_in_user() == 0` and `assert log_out_user() == 0` every time. we pretty much never expect it to fail.

maybe instead of having these two helpers return exit code which needs to be checked every time, have them throw an exception on errors instead? then this would just become simple `log_out_user()` line.

e2e_tests/tests/cluster/test_users.py
```diff
@@ -946,30 +989,42 @@ def test_change_displayname(clean_auth: None) -> None:
     assert current_user is not None and current_user.id
 
     # Rename user using display name
-    patch_user = bindings.v1PatchUser(displayName="renamed display-name")
-    bindings.patch_PatchUser(sess, body=patch_user, userId=current_user.id)
+    patch_user = bindings.v1PatchUser(displayName="renamed")
```
ioga, line 992: can we bring back the `renamed display-name` like before, so we have spaces and dashes?

harness/determined/cli/user.py
```diff
@@ -70,30 +43,22 @@ def log_in_user(parsed_args: Namespace) -> None:
...
 
     token_store = authentication.TokenStore(parsed_args.master)
-
     token = authentication.do_login(parsed_args.master, username, password)
     token_store.set_token(username, token)
     token_store.set_active(username)
 
 
 @authentication.optional
+@login_sdk_client
```
ioga, line 55: with this change, when you're already logged out, doing `det user logout` will return:
```
Failed to log out user: {"error":{"code":16,"reason":"Unauthenticated","error":"invalid credentials"}}
```
which I think is not necessary. logout should not "require" user to be logged in.

harness/determined/cli/_util.py
```diff
@@ -81,6 +83,15 @@ def make_pagination_args(
 default_pagination_args = make_pagination_args()
 
 
+def login_sdk_client(func: Callable[[argparse.Namespace], Any]) -> Callable[..., Any]:
```
ioga, line 86: currently in master, if you are not logged in and you try to do e.g. `det user list`, the message is: 
```
Failed to list users: Unauthenticated: Please use 'det user login <username>' for password login, or for Enterprise users logging in with an SSO provider, use 'det auth login --provider=<provider>'.
```

with this PR, the message is: 
```
Failed to list users: {"error":{"code":16,"reason":"Unauthenticated","error":"invalid credentials"}}
```

I think the previous message was somewhat useful, can we please bring that back?

harness/determined/cli/user.py
```diff
@@ -153,31 +117,27 @@ def link_with_agent_user(parsed_args: Namespace) -> None:
...
+        agent_group=parsed_args.agent_group,
+        agent_uid=parsed_args.agent_uid,
+        agent_user=parsed_args.agent_user,
+    )
 
 
-@authentication.required
+@login_sdk_client
 def create_user(parsed_args: Namespace) -> None:
+    print("in create user cli")
```
ioga, line 131: debug print

master/internal/api_user.go
```diff
@@ -204,6 +212,41 @@ func (a *apiServer) GetUser(
...
+	if err != nil {
+		return nil, err
+	}
+
+	var ok bool
+	if ok, err = user.AuthZProvider.Get().CanGetUser(ctx, *curUser, targetFullUser.ToUser()); err != nil {
+		return nil, err
+	} else if !ok {
+		return nil, errUserNotFound
+	}
```
ioga, line 246: this is a nitpick but I believe more idiomatic style would be something like:

```suggestion
	ok, err := user.AuthZProvider.Get().CanGetUser(ctx, *curUser, targetFullUser.ToUser())
	if err != nil {
		return nil, err
	}
	if !ok {
		return nil, errUserNotFound
	}
```

master/internal/api_user.go
```diff
@@ -352,18 +394,16 @@ func (a *apiServer) PatchUser(
...
-
 		displayName, err2 := clearUsername(targetUser, *req.User.DisplayName, 0)
 		if err2 != nil {
 			return nil, err2
 		}
 
 		if displayName == nil || *displayName == "" {
 			// Set empty display name to null.
-			err = a.m.db.QueryProto("set_user_display_name", u, req.UserId, nil)
+			err = user.SetDisplayName(req.UserId, "")
```
ioga, line 404: I think the behavior here has changed? it used to set to NULL, and now it sets to empty string instead?

master/internal/user/postgres_users.go
```diff
@@ -55,6 +55,13 @@ func UserByUsername(username string) (*model.User, error) {
 	return &user, nil
 }
 
+// Set DisplayName in User.
+func SetDisplayName(userID int32, displayName string) error {
+	var user model.User
+	_, err := db.Bun().NewUpdate().Model(&user).Set("display_name = ?", displayName).Where("id = ?", userID).Exec(context.TODO())
```
ioga, line 61: `.Model((*model.User)(nil)).` is a way to do the same without a separate `var` declaration

proto/src/determined/api/v1/api.proto
```diff
@@ -180,7 +202,7 @@ service Determined {
   rpc PatchUser(PatchUserRequest) returns (PatchUserResponse) {
     option (google.api.http) = {
       patch: "/api/v1/users/{user_id}",
-      body: "user"
+      body: "*"
```
ioga, line 205: why did this get changed?

e2e_tests/tests/cluster/test_users.py
nrajanee: Messaged about this on slack.

e2e_tests/tests/cluster/test_users.py
nrajanee: experiment.run_basic_test is using the determined credentials to run an exp so this is to ensure it's the right credentials in case the password of determined was changed somewhere else.

proto/src/determined/api/v1/api.proto
nrajanee: Had to change this so that I could also send isHashed (which is a part of the PatchUserRequest) instead of just the user field in PatchUserRequest.

ioga: so the problem is that this is breaking the API payload format for the requests, because instead of patching `user` (i.e. `PatchUser` message) it becomes a nested structure.
this is also a problem because a EE customer is already using this as a raw REST API.
I think we should keep `body: "user"` here, but put `isHidden` into `PatchUser` object instead. I believe that should achieve the same functionality but non-breaking.

nrajanee: Ok, I made this change in react too I will revert that.

harness/determined/cli/_util.py
nrajanee: I made this change. And it's working fine in the CLI. For some reason test_logout is failing because it's not seeing this error when it should. But I don't want to stall the review because of this so asking for re-review before I figure this out. Hope that's ok.

e2e_tests/tests/cluster/test_workspace_org.py
```diff
@@ -18,6 +18,9 @@
 @pytest.mark.e2e_cpu
 def test_workspace_org() -> None:
     master_url = conf.make_master_url()
+    # Ensure determined creds are the default values.
+    with logged_in_user(ADMIN_CREDENTIALS):
+        change_user_password("determined", "")
```
nrajanee, line 23: maybe having a fixture to do this is better? Instead of having sprinkles everywhere.

ioga: if we change default determined password in another test, which causes it to be broken here, we could reset it back in that test (e.g. in `finally` clause)?

harness/determined/common/api/authentication.py
```diff
@@ -145,24 +142,19 @@ def do_login(
...
 
 
 def _is_token_valid(master_address: str, token: str, cert: Optional[certs.Cert]) -> bool:
     """
     Find out whether the given token is valid by attempting to use it
-    on the "/users/me" endpoint.
+    on the "api/v1/me" endpoint.
     """
     headers = {"Authorization": "Bearer {}".format(token)}
     try:
```
nrajanee, line 160: if I use bindings.get_GetMe here I won't be able to use the token. Is that ok?

ioga: yeah, I can't think of another way. you also need to update the actual request URL as well though?

e2e_tests/tests/cluster/test_users.py
```diff
@@ -57,81 +72,62 @@ def det_run(args: List[str]) -> str:
     return cast(str, pexpect.run(f"det -m {conf.make_master_url()} {' '.join(args)}").decode())
 
 
-def log_in_user(credentials: authentication.Credentials) -> int:
+def log_in_user(credentials: authentication.Credentials, expectedStatus: int = 0) -> None:
+    import re
```
ioga, line 76: move up. weird the lint didn't complain

e2e_tests/tests/cluster/test_users.py
```diff
@@ -57,81 +72,62 @@ def det_run(args: List[str]) -> str:
...
     child.close()
+    assert child.exitstatus == expectedStatus
 
-    assert child.exitstatus == 0
+
+def create_test_user(add_password: bool = False) -> authentication.Credentials:
+    n_username = get_random_string()
+    command = ["det", "-m", conf.make_master_url(), "user", "create", n_username]
+    output = subprocess.check_call(command)
+    assert output == 0
```
ioga, line 95: `run(..., check=True)` or `check_call` will raise if anything is wrong
```suggestion
    subprocess.run(command, check=True)
```

e2e_tests/tests/cluster/test_users.py
```diff
@@ -140,26 +136,20 @@ def log_out_user(username: Optional[str] = None) -> int:
...
+def activate_deactivate_user(active: bool, target_user: str) -> int:
+    command = [
+        "det",
+        "-m",
+        conf.make_master_url(),
+        "user",
+        "activate" if active else "deactivate",
+        target_user,
+    ]
+    return subprocess.call(command)
```
ioga, line 152: I think this can be simplified to get rid of the return code just like we did with `log_in_user` because the result of this function is always checked to be `== 0`.

```suggestion
def activate_deactivate_user(active: bool, target_user: str) -> None:
    command = [
        "det",
        "-m",
        conf.make_master_url(),
        "user",
        "activate" if active else "deactivate",
        target_user,
    ]
    subprocess.run(command, check=True)
```

e2e_tests/tests/cluster/test_users.py
```diff
@@ -266,61 +266,104 @@ def test_logout(clean_auth: None) -> None:
...
+    assert user.change_password(new_password=new_password_sdk)
+    log_in_user(authentication.Credentials(creds.username, new_password_sdk))
+
+
+@pytest.mark.e2e_cpu
+def test_change_username(clean_auth: None, login_admin: None) -> None:
+    creds = create_test_user()
+    new_username = "rename-user-64"
+    command = ["det", "-m", conf.make_master_url(), "user", "rename", creds.username, new_username]
+    assert subprocess.call(command) == 0
```
ioga, line 342:
```suggestion
    subprocess.run(command, check=True)
```

e2e_tests/tests/cluster/test_users.py
```diff
@@ -459,33 +503,28 @@ def check_whoami(expected_username: str) -> None:
...
     passwd_prompt = f"Password for user '{creds.username}':"
-    unauth_error = r".*Forbidden\(user not active\)"
-
-    with logged_in_user(ADMIN_CREDENTIALS):
-        child = det_spawn(["user", "deactivate", creds.username])
-        child.wait()
-        assert child.exitstatus == 0
+    unauth_error = "user is not active"
+    command = ["det", "-m", conf.make_master_url(), "user", "deactivate", creds.username]
+    assert subprocess.call(command) == 0
```
ioga, line 512:
```suggestion
    subprocess.run(command, check=True)
```

master/internal/api_user.go
```diff
@@ -352,18 +395,16 @@ func (a *apiServer) PatchUser(
...
 			return nil, err2
 		}
 
 		if displayName == nil || *displayName == "" {
 			// Set empty display name to null.
-			err = a.m.db.QueryProto("set_user_display_name", u, req.UserId, nil)
+			err = user.SetDisplayName(req.UserId, nil)
 		} else {
-			err = a.m.db.QueryProto("set_user_display_name", u, req.UserId, displayName)
+			err = user.SetDisplayName(req.UserId, displayName)
```
ioga, line 407: could you coordinate this with @NicholasBlaskey re: https://determinedai.atlassian.net/browse/DET-8659

harness/determined/common/api/request.py
```diff
@@ -127,6 +127,9 @@ def do_request(
         data = det.util.json_encode(json)
 
     try:
+        # print("method and path in do_request")
+        # print(method)
+        # print(path)
```
ioga, line 132:
```suggestion
```

harness/determined/common/api/experiment.py
```diff
@@ -128,6 +128,8 @@ def create_experiment(
     activate: bool = True,
     additional_body_fields: Optional[Dict[str, Any]] = None,
 ) -> int:
+
+    print("in create experiment")
```
ioga, line 132:
```suggestion
```

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,73 @@
...
+    ) -> bindings.v1PatchUserResponse:
+        v1agent_user_group = bindings.v1AgentUserGroup(
+            agentGid=agent_gid,
+            agentGroup=agent_group,
+            agentUid=agent_uid,
+            agentUser=agent_user,
+        )
+        patch_user = bindings.v1PatchUser(agentUserGroup=v1agent_user_group)
+        resp = bindings.patch_PatchUser(self.session, body=patch_user, userId=self.user_id)
+        return resp
```
ioga, line 73: 1. For all these functions which mutate the server state, we should update the properties of the python object accordingly. i.e. if you call `u.rename(...)` then `u.username` should be equal to the newly assigned value. 
2. I think we should not return the internal `bindings` responses from any of these methods. they are at risk of breaking when we change the underlying APIs, so I'd rather not commit to supporting them.

ioga (approved):

ioga (changes requested):
accidentally clicked approve on the previous batch

harness/determined/common/experimental/user.py
```diff
@@ -0,0 +1,88 @@
...
+        self.user_id = user_id
+        self.active = active
+        self.agent_uid = agent_uid
+        self.agent_gid = agent_gid
+        self.agent_user = agent_user
+        self.agent_group = agent_group
+        self._session = session
+        self.display_name = display_name
+
+    def reload(self, raw: Optional[bindings.v1User] = None) -> None:
```
ioga, line 32: let's have it as private for now

```suggestion
    def _reload(self, raw: Optional[bindings.v1User] = None) -> None:
```

ioga (approved):

ioga:
lgtm

PR #5210, opened 2022-10-06
fix: only call fetchKnownRoles in RBAC, closing testing issue

mapmeld:
## Description

Tests of UserSettings and GroupSettings were calling fetchKnownRoles without RBAC responses. The error is supposed to be processed by handleError , but that's not possible during these tests.

Situation is resolved by only calling fetchKnownRoles when rbacEnabled 

## Test Plan

Passing CI

## Checklist


keita-determined (approved):
Test cases are passed on local
LGTM if CI test cases are passed too

PR #5217, opened 2022-10-07
Add "Ticket" section to pull request template

ClaireNeveu:
## Description

This adds a new "Ticket" section to the PR template to create a standard place to put the ticket link. I also leveraged autolink to add templates for the ticket links.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -33,6 +33,17 @@ See [Release Note](https://github.com/determined-ai/determined/blob/master/docs/
 - [ ] Licenses should be included for new code which was copied and/or modified from any external code.
 - [ ] If modifying `/webui/react/src/shared/` verify `make -C webui/react test-shared` passes.
 
+## Ticket
+<!---
+Retain the relevant line and replace 000 with ticket number.
+
+DET-000
+MLG-000
+WEB-000
+DESIGN-000
+No Ticket
+--->
+
 
 <!---
 ## Title


dannysauer:
Jira already gets tickets added in the PR title and links the PR to the ticket.
https://github.com/determined-ai/determined/pull/5216
https://github.com/determined-ai/determined/pull/5190

Is the intention here to also add a link from the PR back to Jira as well?  I wonder aloud if that might be something relatively easy to do with a Jira automation or Github action to add a comment.  Mostly because manually pasting in a link is a pain. :)

ClaireNeveu:
@dannysauer Yes, the intention here is to have a link from the PR to the ticket to make it easier for reviewers to get context.

dannysauer:
What do you think about a github action which requires the PR title to link to a Jira ticket with one of the known project prefixes, fails if that's missing, and adds a comment like the Netlify preview? Except maybe without the tour of every emoji available. :)

ClaireNeveu:
I don't know that it needs to _fail_. We sometimes make bugfix PRs without a ticket. The other factor here is that a PR occasionally references more than one ticket.

loksonarius (approved):

dannysauer (approved):
@ClaireNeveu - looks like a conflict was introduced - probably when I reformatted all the markdown. :)

While you're resolving that (which should be trivial) - we currently have 13 active Jira projects; probably just two or three examples in the comment are sufficient rather than maintaining a complete list.  Either way, I'm also good with this addition.

PR #5218, opened 2022-10-07
fix: Check tensorboard responsive before fetch loop [MLG-195]

varlogtim:
## Description

This works around an issue in which the initial file fetch exceeds the total timeout (10 minutes) which causes us to to raise an exception before actually checking if the Tensorboard process has started and is responsive on HTTP. This changes the functionality in that we should just wait for the Tensorboard process to become responsive to HTTP before entering the file fetching loop.

## Test Plan

:shrug: Download such a large amount of files that it takes > 10 minutes to download.


varlogtim:
Closing as there were changes to this area since this PR was created. I think @mpkouznetsov was going to take this change over.

PR #5220, opened 2022-10-07
fix: replace shell script with python script for `Pre-Commit`

keita-determined:
## Description
[WEB-519]

## Test Plan
- Check if Pre-Commit is triggered when either js, css, misc files are committed

## Commentary (optional)
- I ended up using Python cuz im more familiar with Python and I think its more flexible and readable
- I formatted code with black which we use at Determined

## Checklist

[WEB-519]: https://determinedai.atlassian.net/browse/WEB-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.pre-commit-config.yaml
```diff
@@ -3,19 +3,19 @@ repos:
     hooks:
       - id: web-js-lint-check
         name: Web JS Lint Check
-        entry: sh ./pre-commit/web_lint_check.sh js
+        entry: python ./pre-commit/web_lint_check.py js
```
dannysauer, line 6: This may not behave as expected on some Linux platforms which don't have a python or link python to python2; just python3.

Ideally, set the execute bit on the script in git and use the regular `#!/usr/bin/env python3` thing, and just leave the interpreter off of these calls.

pre-commit/web_lint_check.py
```diff
@@ -0,0 +1,43 @@
+import subprocess
+import multiprocessing
+import sys
+
+
+def web_lint_check():
+    if len(sys.argv) <= 2:
+        print("At least 2 arguments are required")
+        print("usage: python web_lint_check.py [js | css | misc] [<file_paths>]")
+        exit(2)
```
dannysauer, line 10: No argparse? :)
https://docs.python.org/3/library/argparse.html

keita-determined: ah that's probably better

pre-commit/web_lint_check.py
```diff
@@ -0,0 +1,43 @@
...
+
+    dir = "webui/react/"
+    target: str = sys.argv[1]
+    file_paths: str = " ".join(
+        [file_path.replace(dir, "") for file_path in sys.argv[2:]]
```
dannysauer, line 15: This might be a little nicer with relpath, since that handles differnet OS path separators properly and is aware of special things like `../` and so on.
https://docs.python.org/3/library/os.path.html#os.path.relpath

```suggestion
        [os.path.relpath(file_path, dir) for file_path in sys.argv[2:]]
```

keita-determined: nice catch! I was planning to use it but my hand get used to type `replace` :)

.pre-commit-config.yaml
keita-determined: I thought all of us installed Python, so I assumed we all have `python` or `python3` command.
We need to handle if which command is available either  `python` or `python3`, but i think we dont need to use `#!/usr/bin/env python3`?

pre-commit/web_lint_check.py
```diff
@@ -0,0 +1,43 @@
...
+    # TODO: replace it with `match` if we support python v3.10
+    if target == "js":
+        run_command += ["eslint", f"ES_ARGS={file_paths}"]
+    elif target == "css":
+        run_command += ["stylelint", f"ST_ARGS={file_paths}"]
+    elif target == "misc":
+        run_command += ["check-package-lock"]
+    else:
+        print(f"{target} is not found")
+        exit(2)
```
dannysauer, line 36: This is somewhat esoteric feedback because I think only really weird nerds care about it, but exit codes 1 (generic error), 2 (missing command / diff failure), and 126-165, and 255 are "special".  User-specific exit codes should usually just be 3-125

https://tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF
```suggestion
        exit(3)
```

keita-determined: I removed this code after introducing `argparse`

.pre-commit-config.yaml
dannysauer: python3 is encouraged by [PEP-394](https://peps.python.org/pep-0394/), so should be safe to rely upon.  I would've put a bangpath in the executable script and just directly called it, but I'm definitely not going to die on that hill. :)

ClaireNeveu: Since we run make in a virtualenv it is safe to assume that python is indeed python3 but I agree that scripts should generally use `#!` and `python3` is preferable.

keita-determined: Ok i'll change it

dannysauer: Yeah, several "long term support" platforms (SUSE enterprise / OpenSUSE Leap, Ubuntu though at least 20.04, I think RHEL) which are still in use leave python linked to python2 to support old python2 admin tools.

dannysauer, line 6: Do you still need the `python`s if the script is executable? :)

dannysauer (approved):
I'm good with this whether the python stays in the calls or not.  @ClaireNeveu is welcome to disagree. :)

.pre-commit-config.yaml
keita-determined: I have a question.
If I remove `python` command, I get `Executable `./pre-commit/web_lint_check.py` is not executable` error.
it resolves the issue by running `git update-index --chmod=+x pre-commit/web_lint_check.py`.
I don't think this command change wouldn't be reflected on repo, right?
I feel like keeping `python` command would be good.

keita-determined: seems like removing `python` is not a good idea. I'll leave it alone

keita-determined:
Seems like all agreed with the current implementation, and I comprehensively tested `pre-commit`  
I'll merge it.
If new issues arise, we can discuss again.

PR #5221, opened 2022-10-07
fix: custom error requires permission object

eecsliu:
## Description
The custom `PermissionDeniedError` previously took in a `Permission` object that required overhead to build - this change makes the error take in `PermissionType` (int32) which is much more efficient and cuts unnecessary struct building.

## Checklist


maxrussell (approved):
Good change, LGTM

PR #5226, opened 2022-10-10
chore: update to `React18`

keita-determined:
## Description
[WEB-171]

## Test Plan
- Check if web works as expected

## Commentary (optional)
I disabled `Strict Mode` for now cuz we rely on the following code.
```ts
  useEffect(() => {
    ...
  }, [canceler]);
```
It doesn't work well with Strict Mode cuz the new Strict Mode renders component twice in development mode

Also, changed some files to fix Typescript errors

## Checklist

[WEB-171]: https://determinedai.atlassian.net/browse/WEB-171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webui/react/src/index.tsx
```diff
@@ -23,13 +23,16 @@ if (process.env.PUBLIC_URL && history.location.pathname === '/') {
...
-      <App />
-    </HistoryRouter>
-  </React.StrictMode>,
-  document.getElementById('root'),
+const container = document.getElementById('root');
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+const root = createRoot(container!);
+
+root.render(
+  // <React.StrictMode>
```
keita-determined, line 31: I disabled `Strict Mode` for now cuz we rely on the following code.
```ts
  useEffect(() => {
    ...
  }, [canceler]);
```
It doesn't work well with Strict Mode cuz the new Strict Mode renders component twice in development mode

webui/react/package.json
```diff
@@ -51,12 +51,12 @@
...
-    "react": "^17.0.2",
+    "react": "^18.2.0",
     "react-dnd": "^15.1.1",
     "react-dnd-html5-backend": "^15.1.2",
-    "react-dom": "^17.0.2",
-    "react-draggable": "^4.4.4",
-    "react-helmet-async": "^1.1.2",
+    "react-dom": "^18.2.0",
+    "react-draggable": "^4.4.5",
+    "react-helmet-async": "^1.3.0",
```
keita-determined, line 59: Updated these cuz new versions support React18

webui/react/tsconfig.json
```diff
@@ -2,10 +2,8 @@
   "extends": "./src/shared/configs/tsconfig.json",
   "compilerOptions": {
     "baseUrl": "./src",
-    "module": "esnext"
+    "module": "esnext",
+    "noImplicitAny": false
```
keita-determined, line 6: React18 is more strict about types. Many places don't specify types well, so I used `noImplicitAny` here. I couldn't add types all places cuz its too too many. We can add types later

webui/react/src/index.tsx
ClaireNeveu: Can you elaborate on this a bit more? What was strict mode complaining about?

keita-determined: >This new check will automatically unmount and remount every component, whenever a component mounts for the first time, restoring the previous state on the second mount.
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-strict-mode

New Strict Mode unmount first, and remount afterwards in development mode. Due to this, `canceler.abort();` is called unexpectedly, so API calls are aborted. 
We actually talked about this in July, and couldn't find a good solution with `canceler.abort();` and Strict Mode.

webui/react/src/pages/ProjectDetails.tsx
```diff
@@ -429,7 +429,11 @@ const ProjectDetails: React.FC = () => {
     );
 
     const actionRenderer: ExperimentRenderer = (_, record) => {
-      return <ContextMenu record={record} />;
+      return (
+        <ContextMenu record={record} onVisibleChange={undefined}>
+          <></>
```
EmilyBonar, line 434: question: why is this necessary? I noticed you adding fragments in a lot of places but this seems the weirdest.

EmilyBonar (approved):
Web seems to still be working as expected, so LGTM

webui/react/src/pages/ProjectDetails.tsx
keita-determined: Yes, React complained to set the default value, but if i set the default value in prop, i got another error from [here](https://github.com/determined-ai/determined/blob/master/webui/react/src/pages/ProjectDetails.tsx#L1011) 
I wasn't sure about these type, so I made it like this. I agree that its weird. We should fix it later

ClaireNeveu: You should fix the root issue in this PR instead of accumulating tech debt. In this case the issue is that `children: React.ReactNode;` should be `children?: React.ReactNode;` in `ContextMenu`.

webui/react/src/pages/Cluster/ClusterHistoricalUsageFilters.tsx
```diff
@@ -90,7 +90,7 @@ const ClusterHistoricalUsageFilters: React.FC<ClusterHistoricalUsageFiltersProps
     return currentDate.isBefore(value.afterDate) || currentDate.isAfter(dayjs());
   };
 
-  let periodFilters = null;
+  let periodFilters: JSX.Element = <></>;
```
ClaireNeveu, line 93: You can type these as `ReactNode` which includes all valid JSX types.

webui/react/src/pages/ProjectDetails.tsx
keita-determined: ok i'll fix this before merging this. I actually tried this but it didnt work, but I didnt spend much time to investigate this, so let me check

webui/react/src/index.tsx
keita-determined: Not really
Even if I fix this, the behavior looks like [this](https://determined-ai.slack.com/archives/CKQ3EV0G6/p1658334035443339?thread_ts=1658270891.880979&cid=CKQ3EV0G6)
I dont think its good.

PR #5229, opened 2022-10-11
fix: Roles should be fetched immediately when logging in

ClaireNeveu:
## Description

This fixes an issue noticed by @ioga where it took a bit to populate roles when logging in. This happened because we immediately fetch roles when the page loads then poll for them on an interval; the initial call fails because the user is not yet logged in. There are broader problems with this method of data fetching but I'm not addressing those in this PR.

Specifically for roles I added a fetch when a user logs in.

## Test Plan

1. Run devcluster against determined-ee with rba enabled
2. Without this change, log in and navigate to http://localhost:3000/settings/user-management
3. log out then log back in, note the page is blank but populates after a few seconds
4. Apply this change
5. Log out then log back in, note the page immediately lists users

## Commentary

It seems to me that we might need to [all of these fetches]() to the login action. Were these working correctly before rbac? Did the API maybe not check if you were logged in prior to rbac when performing certain GETs?

## Checklist

## Ticket
No ticket

### DIFF
--- a/webui/react/src/components/DeterminedAuth.tsx
+++ b/webui/react/src/components/DeterminedAuth.tsx
@@ -3,6 +3,7 @@ import React, { useCallback, useState } from 'react';
 
 import Link from 'components/Link';
 import { StoreAction, useStoreDispatch } from 'contexts/Store';
+import { useFetchMyRoles } from 'hooks/useFetch';
 import { paths } from 'routes/utils';
 import { login } from 'services/api';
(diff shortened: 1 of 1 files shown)


ClaireNeveu:
I did check and the other calls we make in `Navigation` _do_ fail when not logged in regardless of rbac status. It's just that the listing component always refetches the data when it mounts so it's never been an issue.

gt2345 (approved):
I think fetching roles right after login works great

PR #5230, opened 2022-10-11
fix: decorate PBS & slurm options with ,omitempty

phillip-gaisford:
## Description
image

Fixing the unit test involved defaulting the batch_args elements to null, which may have implications for dispatcher RM (null pointers). Perhaps there is a better way.

## Checklist

### DIFF
--- a/harness/determined/common/schemas/expconf/_gen.py
+++ b/harness/determined/common/schemas/expconf/_gen.py
@@ -1480,7 +1480,7 @@ schemas = {
                 "array",
                 "null"
             ],
-            "default": [],
+            "default": null,
             "items": {
                 "type": "string"
             }
@@ -1513,7 +1513,7 @@ schemas = {
                 "array",
                 "null"
             ],
-            "default": [],
+            "default": null,
             "items": {
                 "type": "string"
             }
--- a/harness/determined/common/schemas/expconf/_v0.py
+++ b/harness/determined/common/schemas/expconf/_v0.py
@@ -51,24 +51,20 @@ class ResourcesConfigV0(schemas.SchemaBase):
 
 class PbsClusterConfigV0(schemas.SchemaBase):
     _id = "http://determined.ai/schemas/expconf/v0/hpc-cluster-pbs.json"
-    pbsbatch_args: Optional[List[str]] = None
 
     @schemas.auto_init
     def __init__(
         self,
-        pbsbatch_args: Optional[List[str]] = None,
     ) -> None:
         pass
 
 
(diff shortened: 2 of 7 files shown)


jerryharrow (approved):

phillip-gaisford:
Fixing the unit test involved defaulting the batch_args elements to null, which may have implications for dispatcher RM (null pointers). Perhaps there is a better way.

jerryharrow:
The test case is schemas/test_cases/v0/experiment.yaml

Seems like you should be able to adjust the test case?

- name: minimal valid experiment
  sane_as:
    - http://determined.ai/schemas/expconf/v0/experiment.json
  default_as:
    http://determined.ai/schemas/expconf/v0/experiment.json
  case:
    searcher:
      name: single
      metric: loss
      max_length:
        batches: 1000
    entrypoint: model_def:MyTrial
  #####
  defaulted:
...

    pbs:
      pbsbatch_args: []
...
    slurm:
      sbatch_args: []

phillip-gaisford:
> The test case is schemas/test_cases/v0/experiment.yaml
> 
> Seems like you should be able to adjust the test case?
> 
> * name: minimal valid experiment
>   sane_as:
>   
>   * http://determined.ai/schemas/expconf/v0/experiment.json
>     default_as:
>     http://determined.ai/schemas/expconf/v0/experiment.json
>     case:
>     searcher:
>     name: single
>     metric: loss
>     max_length:
>     batches: 1000
>     entrypoint: model_def:MyTrial
>   
>   
>   defaulted:
>   ...
>   pbs:
>   pbsbatch_args: []
>   ...
>   slurm:
>   sbatch_args: []

I did try that, but found no way of specifying a result that satisfies all test cases.

harness/determined/common/schemas/expconf/_gen.py
```diff
@@ -1480,7 +1480,7 @@
                 "array",
                 "null"
             ],
-            "default": [],
+            "default": null,
```
jerryharrow, line 1483: Looking at https://pkg.go.dev/encoding/json, this is probably right.   The default value for a struct on omitempty would be a nil, not an empty array.

If the slurm section was present then the sbatch_args default value would be [] though.

> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

jerryharrow (approved):

rcorujo (approved):
The changes look reasonable, although I don't have intimate knowledge of this space.

PR #5234, opened 2022-10-11
test: fix errors of `process.env`

keita-determined:
## Description
 
Editor shows many errors cuz `process.env` is read only

### Before
image

### After
image

## Test Plan
- Check if all test cases pass

## Commentary (optional)
I didnt change any test cases

## Checklist

### DIFF
--- a/webui/react/src/routes/utils.test.ts
+++ b/webui/react/src/routes/utils.test.ts
@@ -2,20 +2,20 @@ import { reactHostAddress } from 'shared/utils/routes';
 
 import { findReactRoute } from './utils';
 
-const initPublicUrl = process.env.PUBLIC_URL;
+const initEnv = process.env;
 
 describe('Routing Utilities', () => {
   afterEach(() => {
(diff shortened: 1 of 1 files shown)


hamidzr (commented):
IIRC webpack does some preprocessing wrt to process.env so that we can use it as if this was a node app and actually had access to env variables. we're switching out the reference to env and this might not be doing what we expect it to do

IIRC the literal process.env.XYZ gets replaced with the value using the webpack plugin we have https://github.com/hamidzr/determined/blob/77ff49a3505904dbf5d7635cef60161d65266d69/webui/react/craco.config.js#L9

`hrz ➜ react git:(master) ag 'process.env\.[^ ]*' -o build` non of the output js files actually reference process.env

this can be confusing but we might need to think about it a bit more. is the test working is also a question I'm a bit unsure about eg thet test on reactHostAddress. let's try to make the test brake by changing its implementation based on public_url

PR #5235, opened 2022-10-11
feat: det profiler support for Core API

azhou-determined:
## Checklist

### DIFF
--- a/examples/tutorials/core_api/5_profiling.py
+++ b/examples/tutorials/core_api/5_profiling.py
@@ -12,22 +12,12 @@ import time
 import determined as det
 from determined import profiler
 
-def main(core_context, increment_by):
-    for batch in range(100):
-        with core_context.profiler.record_timing("train_batch.backward"):
-            steps_completed = batch + 1
-            time.sleep(1)
-            logging.info(f"batch is now {batch}")
-        core_context.profiler.record_metric("test metric", 1.0)
-        # NEW: report training metrics.
-        if steps_completed % 10 == 0:
-            core_context.train.report_training_metrics(
-                steps_completed=steps_completed, metrics={"x": batch}
-            )
-    # NEW: report a "validation" metric at the end.
-    core_context.train.report_validation_metrics(
-        steps_completed=steps_completed, metrics={"x": 100}
-    )
+def main(core_context):
+    with core_context.profiler.init(enabled, begin_on_batch, end_after_batch) as profiler:
+        for batch in range(100):
+            with profiler.record_timing("batch"):
+                print(batch)
+            profiler.step_batch()
 
 
 if __name__ == "__main__":
@@ -35,4 +25,4 @@ if __name__ == "__main__":
 
     # NEW: create a context, and pass it to the main function.
     with det.core.init() as core_context:
-        main(core_context=core_context, increment_by=1)
+        main(core_context=core_context)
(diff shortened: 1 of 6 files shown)


harness/determined/core/_context.py
```diff
@@ -39,20 +39,24 @@ def __init__(
...
         train: Optional[core.TrainContext] = None,
         searcher: Optional[core.SearcherContext] = None,
+        profiler: Optional[core.ProfilerContext] = None,
     ) -> None:
         self.checkpoint = checkpoint
         self.distributed = distributed or core.DummyDistributedContext()
         self.preempt = preempt or core.DummyPreemptContext(self.distributed)
         self.train = train or core.DummyTrainContext()
         self.searcher = searcher or core.DummySearcherContext(self.distributed)
+        self.profiler = profiler
```
rb-determined-ai, line 49: prolly you'll want a DummyProfilerContext somewhere, so `core_context.profiler.*` isn't broken in these cases.

examples/tutorials/core_api/5_profiling.py
```diff
@@ -0,0 +1,40 @@
...
+import time
+
+# NEW: import determined
+import determined as det
+from determined import profiler
+
+def main(core_context, increment_by):
+    core_context.profiler.profiler.set_training(True)
+    for batch in range(100):
+        with core_context.profiler.profiler.record_timing("train_batch.backward"):
```
rb-determined-ai, line 18: The `profiler.profiler` snippet looks like we're leaking implementation details into public APIs.

If the current ProfilerAgent really is ready to be promoted to public API status, it should probably just _be_ the `core_context.profiler`, rather than an attribute of it.

If not, then `core_context.profiler` should probably be a [facade pattern](https://en.wikipedia.org/wiki/Facade_pattern) object, exposing exactly the public API we want to expose from the underlying ProfilerAgent.

examples/tutorials/core_api/5_profiling.py
```diff
@@ -0,0 +1,40 @@
...
+            steps_completed = batch + 1
+            time.sleep(1)
+            logging.info(f"batch is now {batch}")
+        core_context.profiler.profiler.update_batch_idx(batch)
```
rb-determined-ai, line 22: it would be very cool to offer some sort of pass-thru iterator that counted batches instead of requiring this sorta weird call.

```python
for batch in profiler.count_batches(range(100), start=0):
    ...
```

azhou-determined: i thought about this too. we need to know when to step, so either we control the iterator, or users tell us when to step. i agree calling update_batch_idx is a bit weird, but providing some sort of iterator also gets weird when you have a dataloader. what if i want to iterate through batches using enumerate(dataloader)? i feel like it makes more sense to just call step yourself. what if we called it `profiler.step_batch()` instead? no args, profiler keeps track of current batch step.

examples/tutorials/core_api/5_profiling.py
azhou-determined: i totally agree and meant to do this. updated.

harness/determined/core/_context.py
azhou-determined: done

azhou-determined:
this feature has been tabled pending future discussion

PR #5236, opened 2022-10-11
fix: make project settings specific to project

trentwatt:
## Description

[WEB-129](https://determinedai.atlassian.net/browse/WEB-129)
[WEB-451](https://determinedai.atlassian.net/browse/WEB-451)
[WEB-464](https://determinedai.atlassian.net/browse/WEB-464)

row selection, columns, filters, sorting should be particular to a project

## Test Plan

go to a project details page, then click on uncategorized.

make a bunch of state-related updates while navigating back and forth between the two pages. state updates should remain isolated to their particular pages.

## Checklist

### DIFF
--- a/webui/react/src/hooks/useSettings.ts
+++ b/webui/react/src/hooks/useSettings.ts
@@ -214,6 +214,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions): 
   const location = useLocation();
   const storage = useStorage(options?.storagePath || config.storagePath);
   const prevSearch = usePrevious(location.search, undefined);
+  const prevPath = usePrevious(location.pathname, undefined); // console.log(prevSearch, location.search);
   const [settings, setSettings] = useState<T>(() => getDefaultSettings<T>(config, storage));
   const [pathChange, setPathChange] = useState<PathChange<T>>(defaultPathChange);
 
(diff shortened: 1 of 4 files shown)


webui/react/src/hooks/useSettings.ts
```diff
@@ -214,6 +214,7 @@ const useSettings = <T>(config: SettingsConfig, options?: SettingsHookOptions):
   const location = useLocation();
   const storage = useStorage(options?.storagePath || config.storagePath);
   const prevSearch = usePrevious(location.search, undefined);
+  const prevPath = usePrevious(location.pathname, undefined); // console.log(prevSearch, location.search);
```
keita-determined, line 217: nit: nice to remove console.log

keita-determined (commented):
I found an interesting behavior.
https://user-images.githubusercontent.com/109113616/195405503-c77349d4-9a29-4d88-9d76-4205ac89f1ad.mov

question1: `state-related` changes include `column` and `column width` too?
question2: as you know, there is a PR to improve `useSettings`. Does the new one have compatibility with this?

webui/react/src/hooks/useSettings.ts
trentwatt: > I found an interesting behavior. https://user-images.githubusercontent.com/109113616/195405503-c77349d4-9a29-4d88-9d76-4205ac89f1ad.mov
> 
> question1: `state-related` changes include `column` and `column width` too? question2: as you know, there is a PR to improve `useSettings`. Does the new one have compatibility with this?

@keita-determined 

yes that is an interesting behavior, thanks.

q1: yes
q2: intention is for the other PR to replace this, but I wanted to see if it was possible to get an easy bugfix in the meantime.

keita-determined: I dont know if its a good idea, but I think some states should be shared across the projects, such as column and column width, items per page etc. Or we can have `apply to all projects` for these states. Its because I feel weird to have different columns or widths (UI).

trentwatt: different columns was the request from design, and I agree with it. different projects you might be interested in different things. if you have different columns you have to have different column widths.

i kinda agree with you on items per page, but I don't think that alone warrants the extra complexity.

PR #5237, opened 2022-10-11
chore: support out-of-k8s dev workflow [DET-8107]

rb-determined-ai:
## Description

See the [README in the PR](https://github.com/rb-determined-ai/determined/blob/local-k8s/tools/k8s/README.md) for more info.

## Commentary

* Adding dev flags to the master isn't strictly necessary, but it does make rebuilding and rerunning fast and easy.
* There are authentication plugins that you can use, which would make `fetch-creds.sh` unnecessary.  Basically, if `kubectl` can be configured to authenticate, so can any piece of go code.  I didn't pick that strategy for two reasons:
    * It is less flexible.  `fetch-creds.sh` works on unmodified k8s apps as well (as demonstrated with the coscheduler in this pr)
    * The auth plugins get compiled into the application, and are not guaranteed to work across versions.  `fetch-creds.sh` lets `kubectl` figure out the auth plugins and just uses the end-state credentials, which seems less fragile to me.

### DIFF
--- a/master/internal/rm/kubernetes/pods.go
+++ b/master/internal/rm/kubernetes/pods.go
@@ -3,9 +3,12 @@
 package kubernetes
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"path/filepath"
 	"strconv"
 
 	"github.com/determined-ai/determined/master/pkg/cproto"
@@ -56,6 +59,7 @@ type pods struct {
 	slotType                 device.Type
 	slotResourceRequests     PodSlotResourceRequests
 	fluentConfig             FluentConfig
+	credsDir                 string
 
 	clientSet        *k8sClient.Clientset
 	masterIP         string
@@ -105,6 +109,9 @@ func Initialize(
 	slotType device.Type,
 	slotResourceRequests PodSlotResourceRequests,
 	fluentConfig FluentConfig,
+	credsDir string,
+	masterIP string,
+	masterPort int32,
 ) *actor.Ref {
 	loggingTLSConfig := masterTLSConfig
 	if loggingConfig.ElasticLoggingConfig != nil {
@@ -128,6 +135,9 @@ func Initialize(
 		slotType:                     slotType,
 		slotResourceRequests:         slotResourceRequests,
 		fluentConfig:                 fluentConfig,
+		credsDir:                     credsDir,
+		masterIP:                     masterIP,
+		masterPort:                   masterPort,
 		currentNodes:                 make(map[string]*k8sV1.Node),
 		nodeToSystemResourceRequests: make(map[string]int64),
 	})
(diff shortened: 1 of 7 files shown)


tools/k8s/fetch-creds.sh
```diff
@@ -0,0 +1,158 @@
...
+    echo "          --env-file OUTDIR/docker-env-file \\"
+    echo "          my_image_name"
+}
+
+force=n
+outdir=""
+account=""
+context=""
+
+# Manually parse args since getopt varies across unicies.
```
stoksc, line 43: is unicies a real thing?

master/internal/rm/kubernetes/pods.go
```diff
@@ -232,8 +242,49 @@ func (p *pods) Receive(ctx *actor.Context) error {
...
+		//   - /var/run/secrets/kubernetes.io/serviceaccount/token
+		//   - /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+		return rest.InClusterConfig()
+	}
+
+	// A special case for rapid determined+k8s development: build a rest.Config from a specially
+	// packed directory with the same information.  Our tools/scripts/fetch-k8s-creds.sh script can
+	// create such a directory, with server, token, and ca.crt files.
+
+	//nolint:gosec // XXX I don't really understand why this is risky.
```
stoksc, line 261: i think it is just blanket "reading an uncleaned file from a variable". of course, can be disregarded here.. and everywhere it appears.

master/internal/rm/kubernetes/pods.go
```diff
@@ -251,6 +302,10 @@ func (p *pods) startClientSet(ctx *actor.Context) error {
 }
 
 func (p *pods) getMasterIPAndPort(ctx *actor.Context) error {
```
stoksc, line 304: random: introspecting the service pointing at us is kind of an interesting way to do this in the first place. in a past life i would've terraformed a service and deployment, and dropped some outputs of the service tf into the deployment. this works, just isn't very flexible/it deployment-specific.

master/internal/rm/kubernetes/pods.go
```diff
@@ -232,8 +242,49 @@ func (p *pods) Receive(ctx *actor.Context) error {
...
+		//   - /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+		return rest.InClusterConfig()
+	}
+
+	// A special case for rapid determined+k8s development: build a rest.Config from a specially
+	// packed directory with the same information.  Our tools/scripts/fetch-k8s-creds.sh script can
+	// create such a directory, with server, token, and ca.crt files.
+
+	//nolint:gosec // XXX I don't really understand why this is risky.
+	server, err := ioutil.ReadFile(filepath.Join(credsDir, "server"))
```
stoksc, line 262: I'd prefer to just nest a `rest.Config` in the k8s config as `_apiserver_config` and then keep this dev-specific junk in a shell script. but this is also fine, totally non-blocking, just stating my preference.

stoksc (approved):
i have yet to give this a whirl, will later tonight. but all the code lgtm. thanks for putting this together.

tools/k8s/fetch-creds.sh
rb-determined-ai: Yes, like the word "vertex" -> "vertices", though I think I included an extra 'i'.

master/internal/rm/kubernetes/pods.go
rb-determined-ai: I can see the appeal of that, from the master code perspective, but I think it makes it harder for a human to actually interact with this dev feature.

PR #5241, opened 2022-10-12
feat: support configuring agent self-shutdown for `det deploy aws`

dzhu:
## Description

This just threads through a command-line option to `det deploy aws` all the way down to adding the appropriate config entry to the agent. Since the hook is configured in a nested object in the config and has a list value, I didn't want to use another environment variable to configure it, so the provisioner config now includes a free-form JSON field that gets inserted into the agent config file.

The new arguments to `docker run` in the agent setup script are necessary to access and run `shutdown` from within the container.

## Test Plan

- [x] run a cluster with `det deploy aws`, manually stop the master Docker container, make sure the agents shut down and terminate after trying to reconnect

## Commentary (optional)

Might be a bit weird to have `loss` here and `lost` in the agent config. For some reason each of those sounds the most natural to me in its own context, but I'd be willing to change either one to match the other.

### DIFF
--- a/docs/release-notes/det-deploy-clean-orphaned.rst
+++ b/docs/release-notes/det-deploy-clean-orphaned.rst
@@ -1,0 +1,7 @@
+:orphan:
+
+**Improvements**
+
+-  **Breaking Change:** ``det deploy aws`` by default now configures agent instances to
+   automatically shut down if they lose their connection to the master. The
+   ``--no-shut-down-agents-on-connection-loss`` option can be used to turn off this behavior.
--- a/master/internal/config/provconfig/config.go
+++ b/master/internal/config/provconfig/config.go
@@ -20,22 +20,23 @@ const defaultMasterPort = "8080"
 
 // Config describes config for provisioner.
 type Config struct {
-	MasterURL              string            `json:"master_url"`
-	MasterCertName         string            `json:"master_cert_name"`
-	StartupScript          string            `json:"startup_script"`
-	ContainerStartupScript string            `json:"container_startup_script"`
-	AgentDockerNetwork     string            `json:"agent_docker_network"`
-	AgentDockerRuntime     string            `json:"agent_docker_runtime"`
-	AgentDockerImage       string            `json:"agent_docker_image"`
-	AgentFluentImage       string            `json:"agent_fluent_image"`
-	AgentReconnectAttempts int               `json:"agent_reconnect_attempts"`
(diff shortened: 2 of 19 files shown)


docs/release-notes/det-deploy-clean-orphaned.rst
```diff
@@ -0,0 +1,7 @@
+:orphan:
+
+**Breaking Changes**
```
ioga, line 3: I thought that's an improvement!

master/internal/config/provconfig/config.go
```diff
@@ -20,22 +20,23 @@ const defaultMasterPort = "8080"
...
+	MasterCertName          string            `json:"master_cert_name"`
+	StartupScript           string            `json:"startup_script"`
+	ContainerStartupScript  string            `json:"container_startup_script"`
+	AgentDockerNetwork      string            `json:"agent_docker_network"`
+	AgentDockerRuntime      string            `json:"agent_docker_runtime"`
+	AgentDockerImage        string            `json:"agent_docker_image"`
+	AgentFluentImage        string            `json:"agent_fluent_image"`
+	AgentReconnectAttempts  int               `json:"agent_reconnect_attempts"`
+	AgentReconnectBackoff   int               `json:"agent_reconnect_backoff"`
+	AgentConfigFileContents json.RawMessage   `json:"agent_config_file_contents"`
```
ioga, line 33: this might be a handy feature for a bunch of cases!

master/static/srv/agent_setup_script.sh.template
```diff
@@ -40,6 +45,7 @@ cat /usr/local/determined/container_startup_script
 echo "#### PRINTING CONTAINER STARTUP SCRIPT END ####"
 
 docker run --init --name determined-agent {{.LogOptions}} \
+    --privileged \
```
ioga, line 48: IIRC one particular [EE customer](https://determined-ai.slack.com/archives/C0124B04SGM/p1651260879001769?thread_ts=1651252778.066919&cid=C0124B04SGM) had their custom startup script, so when we change it on our end, it may break for them. We need to raise this when they upgrade.
cc @varlogtim

ioga (approved):

docs/release-notes/det-deploy-clean-orphaned.rst
dzhu: Heh, I changed the section, but I put in an inline breaking change marker, since this will change the behavior for people doing the same thing before and after this release and _maybe_ someone cares about the old thing.

PR #5243, opened 2022-10-12
fix: HP Parallel Coordinates remembers filters as data comes in [WEB-279]

mapmeld:
## Description

The Hyperparameter Parallel Coordinates view uses the semi-internal npm package `hermes-parallel-coordinates` to render a chart. The ticket identified an issue where every data update / refresh clears the chart of users' filters. This component should receive filters from the parent component ( discussed in https://github.com/hkang1/hermes/issues/49 ).

The PR currently uses a forked GitHub  repo in package.json, but our final recommendation is either:
- A. install this as a branch on the `hermes-parallel-coordinates` repo
- B. create our official NPM organization and host the modified package there

## Test Plan

- Create a multi-trial experiment (for example: `det e create decision_trees/gbt_titanic_estimator/adaptive.yaml decision_trees/gbt_titanic_estimator`)
- Under the Visualization section, select HP Parallel Coordinates tab
- Once multiple trials fill the graph with diagonal lines, click and drag along one of the vertical axes to create a pink filter line. You can continue to create multiple lines, click the line to remove it, or double-click the chart to clear it of filters.
- When trials' lines are outside of filter ranges, they get filtered out of the table rows below the chart
- The filters remain in place while the experiment is running and collecting new data

## Checklist


webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpParallelCoordinates.tsx
```diff
@@ -56,6 +56,17 @@ interface HpTrialData {
...
+export interface HermesInternalFilter {
+  p0: number;
+  p1: number;
+  value0: Primitive;
+  value1: Primitive;
+}
+
+export interface HermesInternalFilters {
+  [key: string]: HermesInternalFilter[];
+}
```
mapmeld, line 68: on this section, I was hoping `Hermes.InternalFilter` and `Hermes.InternalFilters` would be exported like other types in this line: https://github.com/hkang1/hermes/pull/108/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R20

I wasn't able to import them though, and I couldn't guess if this is due to npm, typescript,  or these being `interface`s instead of `const`s.

trentwatt:
i got a page crash once when going from parallel coordinates to trials tab. not sure how to reproduce though.

image

also, looks like the filters are preserved when the new data comes in, but so is the data in the table :D 

i.e. as new trials come in, you can see them in the parallel coordinates chart, but they dont show up in the table, even if they match the filters.

also for purposes of testing, i think it's easier to `cd e2e_tests/tests/fixtures/no_op` and `det e create adaptive.yaml .`. that generates data really quick.

trentwatt:
approach LGTM though

mapmeld:
@trentwatt  Thanks for recommending `e2e_tests/tests/fixtures/no_op`.  I was able to create a new `useEffect` which updates the list of filtered trials whenever the filters or data gets updated.

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpParallelCoordinates.tsx
```diff
@@ -97,47 +109,56 @@ const HpParallelCoordinates: React.FC<Props> = ({
...
-      setFilteredTrialIdMap(newFilteredTrialIdMap);
-    },
-    [chartData],
-  );
+    setFilteredTrialIdMap(newFilteredTrialIdMap);
+  };
+
+  useEffect(() => {
+    resetFilteredTrials();
+  }, [chartData, hermesCreatedFeatures, resetFilteredTrials]);
```
trentwatt, line 150: I'm getting "update depth exceeded" on this branch, I think because `resetFilteredTrials` is reassigned on each render.

changing this dependency array to `[resetFilteredTrials]` and wrapping `resetFiltered` trials in a 

```
useCallback(() => {
    ...
}, [chartData, hermesCreatedFeatures]);
```

resolved for me.

trentwatt (approved):
LGTM, apart from the comment i left in the code about "maximum update depth exceeded"

----

Separately, I noticed that the chart flashes when data comes in. Looks like that is not a new behavior though. 

it happens whenever this effect in `ParallelCoordinates.tsx` is triggered.


```
  useEffect(() => {
    if (!containerRef.current) return;

    chartRef.current = new Hermes(containerRef.current);

    return () => {
      chartRef.current?.destroy();
      chartRef.current = undefined;
    };
  }, [dimensions]);
```

in theory i think we only would want `dimensions` to trigger a re-render when the min/max range of HP values or metrics changes. but it happens every time we `fetchExperimentDetails` because that causes an update to `experiment.hyperparameters` even if data is the same.

webui/react/src/pages/ExperimentDetails/ExperimentVisualization/HpParallelCoordinates.tsx
mapmeld: just so I understand this right, it would be

```typescript
const resetFilteredTrials = useCallback(() => {
  // ...existing resetFilteredTrials code...
}, [chartData, hermesCreatedFeatures]);

useEffect(() => {
  resetFilteredTrials();
}, [resetFilteredTrials]);
```

trentwatt: worked for me anyways :D

PR #5246, opened 2022-10-12
fix: Avatar text color change (DET-8237)

johnkim-det:
## Description

using dark/light text color for avatars in `noColor` mode. no changes to text color when `noColor` prop is false (remains `white`).

https://determinedai.atlassian.net/browse/DET-8237

## Test Plan

nothing to test in core since `noColor` prop is only used in SaaS.

## Commentary (optional)

can test in storybook and SaaS to see that in light mode with `noColor` mode, text is dark instead of white:

avatar-nocolor-var
avatar-nocolor-white

## Checklist

### DIFF
--- a/webui/react/src/shared/components/Avatar/Avatar.module.scss
+++ b/webui/react/src/shared/components/Avatar/Avatar.module.scss
@@ -1,7 +1,6 @@
 .base {
   align-items: center;
   border-radius: 100%;
-  color: white;
   display: flex;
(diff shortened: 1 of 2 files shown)


hamidzr (approved):
works great! https://deploy-preview-5246--storybook-det.netlify.app/?path=/story/shared-avatar--default&args=noColor:true&globals=theme:Light+Mode

PR #5247, opened 2022-10-12
chore: add FieldMask type to apiutils

maxrussell:
## Description

While reviewing #5177, I noticed some ergonomic issues working with FieldMasks and wanted to help make that easier. In particular:
- Doing repeated O(n) scans through a slice of fields to check whether fields were specified in the FieldMask is slow and verbose.
- An unspecified FieldMask, by convention, should be interpreted by the server as though all fields were specified, so a check was necessary that was easy to miss.
- FieldMask allows a request to specify an ancestor field such as `a.b` and should carry the semantics that a descendent field like `a.b.c` should be considered in the field set.

So I made a new type (with tests) that uses a map for containment checks and respects the listed FieldMask conventions.

## Test Plan

No behavior should change as a result of this PR.

### DIFF
--- a/master/internal/api/apiutils/mask.go
+++ b/master/internal/api/apiutils/mask.go
@@ -1,0 +1,46 @@
+package apiutils
+
+import (
+	field_mask "google.golang.org/genproto/protobuf/field_mask"
+)
+
+// FieldMask is a utility type for efficiently interacting with protobuf FieldMasks in a compliant
+// way.
+type FieldMask struct {
+	m        *field_mask.FieldMask
+	fieldSet map[string]bool
+}
+
+// NewFieldMask initializes a FieldMask object from a protobuf FieldMask pointer. A passed-in nil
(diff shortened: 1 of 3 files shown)


master/internal/api/apiutils/mask.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if f.fieldSet[field] {
+		return true
+	}
+
+	// If the fieldmask contains a.b and a user inquires about field a.b.c, the answer is yes.
+	fields := strings.Split(field, ".")
+	if len(fields) > 1 {
+		// Continually re-slice the slice of fields to check if an ancestor of field was specified.
+		for fields = fields[:len(fields)-1]; len(fields) > 0; fields = fields[:len(fields)-1] {
```
ioga, line 53: cool

ioga (approved):

PR #5251, opened 2022-10-13
fix: check target user, allow \wadmin\w names but not admin itself [WEB-529]

mapmeld:
## Description

- address issue of display names including but not actually == "Admin"
- code changes from checking `curUser.isAdmin` to `targetUser` for the renaming

## Test Plan

Rename user

## Checklist

### DIFF
--- a/master/internal/api_user.go
+++ b/master/internal/api_user.go
@@ -327,8 +327,7 @@ func (a *apiServer) PatchUser(
 			if !targetUser.Admin && (strings.TrimSpace(strings.ToLower(displayName)) == "admin") {
 				return nil, status.Error(codes.InvalidArgument, "Non-admin user cannot be renamed 'admin'")
 			}
-			if targetUser.Username != "determined" && strings.Contains(strings.ToLower(displayName),
-				"determined") {
+			if targetUser.Username != "determined" && (strings.TrimSpace(strings.ToLower(displayName)) == "determined") {
 				return nil, status.Error(codes.InvalidArgument, "User cannot be renamed 'determined'")
 			}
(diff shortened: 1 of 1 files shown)


master/internal/api_user.go
```diff
@@ -325,11 +324,10 @@ func (a *apiServer) PatchUser(
 			re := regexp.MustCompile("[^\\p{Latin}\\p{N}\\s]")
 			displayName := re.ReplaceAllLiteralString(req.User.DisplayName.Value, "")
 			// Restrict 'admin' and 'determined' in display names.
-			if !(curUser.Admin && curUser.ID == uid) && strings.Contains(strings.ToLower(displayName),
-				"admin") {
+			if !targetUser.Admin && (strings.TrimSpace(strings.ToLower(displayName)) == "admin") {
 				return nil, status.Error(codes.InvalidArgument, "Non-admin user cannot be renamed 'admin'")
 			}
-			if curUser.Username != "determined" && strings.Contains(strings.ToLower(displayName),
+			if targetUser.Username != "determined" && strings.Contains(strings.ToLower(displayName),
```
ioga, line 330: maybe have the same check as `admin`? I feel like the difference here is so subtle it'd be annoying to explain and defend it if a user asked why can't they do this.

```suggestion
			if targetUser.Username != "determined" && strings.Trimspace(strings.ToLower(displayName)) == "determined"
```

ioga (approved):

master/internal/api_user.go
mapmeld: ✅  made change

PR #5253, opened 2022-10-13
fix: Load workspace members from RBAC API

mapmeld:
## Description

PR fills in the Workspace Members tab using a request to the `GetGroupsAndUsersAssignedToWorkspace` API endpoint, represented in the code as `getWorkspaceMembers`.

Requests are not made when RBAC is not enabled.

## Test Plan

- When viewing a workspace in OSS [not RBAC], there are no requests to `/roles/workspace/{workspace_id}`
- When viewing a workspace with RBAC flag on - `?f_rbac=on&f_mock_permissions_read=on` - you should see repeated requests to `/roles/workspace/{workspace_id}`, though in OSS these will return an error

## Checklist

### DIFF
--- a/webui/react/src/pages/WorkspaceDetails.tsx
+++ b/webui/react/src/pages/WorkspaceDetails.tsx
@@ -36,6 +36,7 @@ export enum WorkspaceDetailsTab {
 
 const WorkspaceDetails: React.FC = () => {
   const rbacEnabled = useFeature().isOn('rbac');
+  const mockWorkspaceMembers = useFeature().isOn('mock_workspace_members');
 
   const { users } = useStore();
   const { workspaceId: workspaceID } = useParams<Params>();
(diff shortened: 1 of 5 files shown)


webui/react/src/pages/WorkspaceDetails.tsx
```diff
@@ -42,6 +36,7 @@ export enum WorkspaceDetailsTab {
 
 const WorkspaceDetails: React.FC = () => {
   const rbacEnabled = useFeature().isOn('rbac');
+  const mockWorkspaceMembers = useFeature().isOn('mock_workspace_members');
```
gt2345, line 39: What does this mock stand of?

gt2345 (approved):
Works as expected

webui/react/src/pages/WorkspaceDetails.tsx
mapmeld: If someone is using `f_mock_workspace_members=on` to show users in the Workspace Members tab / `pages/WorkspaceDetails/WorkspaceMembers.tsx`, then I don't want the API response to overwrite the members list.

PR #5255, opened 2022-10-13
fix: lint for intg test

nrajanee:
## Description

fix lint for intg test

### DIFF
--- a/master/internal/db/postgres_snapshots_test.go
+++ b/master/internal/db/postgres_snapshots_test.go
@@ -17,19 +17,20 @@ import (
 )
 
 func TestCustomSearcherSnapshot(t *testing.T) {
-	etc.SetRootPath(RootFromDB)
+	err := etc.SetRootPath(RootFromDB)
+	require.NoError(t, err)
 	db := MustResolveTestPostgres(t)
 	MustMigrateTestPostgres(t, db, MigrationsFromDB)
 	user := RequireMockUser(t, db)
 	exp := RequireMockExperiment(t, db, user)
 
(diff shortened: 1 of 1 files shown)


master/internal/db/postgres_snapshots_test.go
```diff
@@ -17,19 +17,20 @@ import (
...
+	err := etc.SetRootPath(RootFromDB)
+	require.NoError(t, err)
 	db := MustResolveTestPostgres(t)
 	MustMigrateTestPostgres(t, db, MigrationsFromDB)
 	user := RequireMockUser(t, db)
 	exp := RequireMockExperiment(t, db, user)
 
-	config := expconf.SearcherConfig{
-		RawCustomConfig: &expconf.CustomConfig{},
+	config := expconf.SearcherConfig{ //nolint
```
nrajanee, line 27: It was complaining that other Configs (RawSingleConfig) are not provided. But we shouldn't have to provide that.

master/internal/db/postgres_snapshots_test.go
```diff
@@ -17,19 +17,20 @@ import (
...
+		RawCustomConfig: &expconf.CustomConfig{}, //nolint
```
nrajanee, line 28: Same thing for RawUnit especially for this integration test.

master/internal/db/postgres_snapshots_test.go
dzhu: You should be able to make these `//nolint:exhaustivestruct` to make explicit what is being ignored.

mpkouznetsov (approved):

PR #5257, opened 2022-10-13
docs: remove references to PBT

dzhu:
## Description

PBT was removed from the product quite some time ago.

## Test Plan

- [x] render and examine docs


docs/training/hyperparameter/search-methods/overview.rst
```diff
@@ -51,23 +51,16 @@ Determined also supports other common hyperparameter search algorithms:
 
 -  :ref:`Single <topic-guides_hp-tuning-det_single>` is appropriate for manual hyperparameter
    tuning, as it trains a single hyperparameter configuration.
-
 -  :ref:`Grid <topic-guides_hp-tuning-det_grid>` evaluates all possible hyperparameter
    configurations by brute force and returns the best.
-
 -  :ref:`Random <topic-guides_hp-tuning-det_random>` evaluates a subset of hyperparameter
```
dzhu, line 56: Can we also change "subset" back to "set"? We haven't previously been talking about a set to take a subset of, so I find "subset" a bit jarring. The implication is clear enough, but 
"set" is no less correct and avoids that bump.

drh-determined-ai (approved):
Looks good!

docs/training/hyperparameter/search-methods/overview.rst
drh-determined-ai: I think "set" might be slightly better, since "subset" implies the existence of a greater set that is relevant to the experiment, which isn't really true. 
Strictly speaking, we're taking hyperparameter points from a hypersheet (defined by a configuration) that cuts through a hyperparameter space (defined by the model architecture). For the purposes here, though, I think it's ok to stick with "set".

PR #5258, opened 2022-10-13
feat: Create Webhook Sender [WEB-213]

julian-determined-ai:
## Description
feat: Create Webhook Sender [WEB-213]

This PR also implements:
https://determinedai.atlassian.net/browse/WEB-217
https://determinedai.atlassian.net/browse/WEB-214
https://determinedai.atlassian.net/browse/WEB-212

This PR contains the backend updates needed to support webhooks in determined. A `webhook_event_que` table has been added which is the que for webhook events. When an experiment state change occurs, if any triggers exist for a webhook related to the state change then the appropriate payload for the webhook type is created and saved to the event que. An `event` shipper has been created that will pull events from the que and attempt to deliver events to the webhook URL. 

### Security and Signed Payload

Each Webhook request will include a signed payload that users can use to verify each event was sent from Determined. This will require that users define a `webhook_signing_key` for signing. If a webhook signing key is not provided then one will be generated for the user.

#### Setting the Key
The signing key can be set in the following ways:

1. In the master `yaml` configuration file by adding:
```
      # config_file is just a master.yaml
      config_file:
        webhook:
          signing_key: <signing_key_value>
```

2. By setting a `DET_WEBHOOK_SIGNING_KEY` environment variable. 
3. By specifying a `---security-webhook-signing-key` flag 

### Event Payloads

Currently we will support two separate types of Webhooks `Slack` and `Default`. A payload for a Default webhook will contain information about the event itself, the trigger for the event, and the appropriate entity that triggered the event. Since we only support experiment state changes for now all payloads will look like: 

```
{
  "event_id": "4cd26e62-60c6-4a8b-8d03-7629091a4ef5".   // unique event uuid,
  "event_type": "EXPERIMENT_STATE_CHANGE" // the trigger type for the event,
  "timestamp": 1665689991 // the timestamp of the event occurence,
  "condition": {. 
    "state": "COMPLETED". // the condition that triggered the event for now the condition will always be of this format and contain either "COMPLETED" or "ERRORED"
  },
  "event_data": {
    "experiment": {
      "id": 41, 
      "state": "COMPLETED",
      "name": "cifar10_pytorch_const",
      "duration": 41, // this is in seconds
      "resource_pool": "default",
      "slots": 0,
      "workspace": "test workspace", // name of the workspace
      "project": "test project" // name of the project
    }
  }
}
```

A `Base URL` configuration has been added that will be used to allow us to link users to Workspaces, Projects, and Experiments. It is discussed in the docs here: https://github.com/determined-ai/determined/blob/master/docs/integrations/notification/slack.rst

### Retrieving the Key

The key can be found in the cluster configuration. For example it will be returned in `api/v1/master/config`.

### Examples

Example Default Webhook Payload:
Screen Shot 2022-10-13 at 4 07 56 PM

Example Slack Message:
Screen Shot 2022-10-13 at 4 09 21 PM

Example Test Default Payload:

Screen Shot 2022-10-13 at 4 21 10 PM

Example Test Slack Payload:
Screen Shot 2022-10-13 at 4 52 22 PM

## Test Plan
1. Create a webhook of type Slack and Default in Determined. Add triggers for "Completed" and "Errored" experiments. For the Default webhook I suggest getting a URL from https://webhook.site. For a Slack URL you can use the Test Webhook Service slack application here: https://api.slack.com/apps/A040M5WLT3N/incoming-webhooks
2. Run an experiment that will either fail or complete. 
3. Check to ensure that both the default and slack destinations received a webhook request and are formatted correctly. 
4. Start an experiment that is in a project and set the cluster `WEBHOOK_BASE_URL`. Ensure that the Slack message correctly contains links to the experiment, workspace, and project. 
5. Check the cluster configuration and ensure that a `webhook_signing_key` has been set by default.
6. Specify a `webhook_signing_key`. Check the webhook configuration and ensure that the correct value is set. 
 

## Checklist

[WEB-213]: https://determinedai.atlassian.net/browse/WEB-213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


julian-determined-ai:
@gt2345 great call out. Referring to the convo here: https://determined-ai.slack.com/archives/C03LE5Z8R46/p1666105393504059?thread_ts=1666104599.960889&cid=C03LE5Z8R46 the plan is to show global permissions in a separate UI. 

I added a `getAssignableWorkspaceRoles` method that will hide global roles. I am going to mention in RBAC that we will need a better way to filter this roles.

master/internal/webhooks/postgres_webhook.go
```diff
@@ -50,3 +74,238 @@ func DeleteWebhook(ctx context.Context, id WebhookID) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating slack payload: %w", err)
+	}
+	return message, nil
+}
+
+type eventBatch struct {
+	tx       *bun.Tx
+	events   []Event
+	consumed bool
```
ioga, line 279: this flag may not be necessary. e.g. we often have a pattern like this for rollback:

```
	defer func() {
		if err = tx.Rollback(); err != nil && err != sql.ErrTxDone {
			log.WithError(err).Error("error rolling back transaction in create workspace")
		}
	}()
```
this way, the fact if the tx has been committed or now will be utilized instead.

master/internal/webhooks/postgres_webhook.go
```diff
@@ -50,3 +74,238 @@ func DeleteWebhook(ctx context.Context, id WebhookID) error {
...
+
+func (b *eventBatch) close() error {
+	if !b.consumed {
+		return b.tx.Rollback()
+	}
+	return nil
+}
+
+func (b *eventBatch) consume() error {
+	b.consumed = true
```
ioga, line 290: maybe just call these `rollback` and `commit`?

master/internal/webhooks/postgres_webhook_intg_test.go
```diff
@@ -0,0 +1,349 @@
...
+		batch, err = dequeueEvents(ctx, maxEventBatchSize)
+		require.NoError(t, batch.consume())
+		require.NoError(t, err)
+		require.Equal(t, 1, len(batch.events))
+	})
+}
+
+func clearWebhooksTables(ctx context.Context, t *testing.T) {
+	t.Log("clear webhooks db")
+	_, err := db.Bun().NewDelete().Model((*Webhook)(nil)).Where("1=1").Exec(ctx)
```
ioga, line 345: `.Where("true")`

master/internal/webhooks/shipper.go
```diff
@@ -0,0 +1,231 @@
...
+		return 0, err
+	}
+
+	if err := b.consume(); err != nil {
+		return 0, fmt.Errorf("consuming batch: %w", err)
+	}
+	return len(b.events), nil
+}
+
+func (w *worker) backoff() backoff.BackOff {
```
ioga, line 176: doesn't have to be a worker method
```suggestion
func backoff() backoff.BackOff {
```

master/internal/webhooks/shipper_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"golang.org/x/exp/maps"
+
+	"github.com/determined-ai/determined/master/internal/db"
+	"github.com/determined-ai/determined/master/pkg/model"
+	"github.com/determined-ai/determined/master/pkg/schemas"
+	"github.com/determined-ai/determined/master/pkg/schemas/expconf"
+)
+
+const (
+	pathToMigrations = "file://../../static/migrations"
```
ioga, line 25: I believe we already have a constant for that: https://github.com/determined-ai/determined/blob/master/master/internal/cache/cache_intg_test.go#L21

master/internal/webhooks/webhook.go
```diff
@@ -179,3 +183,96 @@ func (t TriggerType) Proto() webhookv1.TriggerType {
...
+
+// Proto returns a proto from a TriggerType.
+func experimentToWebhookPayload(e model.Experiment) *ExperimentPayload {
+	var duration int
+	if e.EndTime != nil {
+		duration = int(e.EndTime.Sub(e.StartTime).Seconds())
+	}
+
+	var slots int
+	if s := e.Config.Resources().Slots(); s != nil {
```
ioga, line 195: 1. I thought for an experiment `Slots` would always be empty, and the correct field is `SlotsPerTrial`? 
2. What should be the meaning of this field in payload for multi-trial experiments? slots per trial? slots per trial multiplied by number of trials?

master/internal/webhooks/shipper_test.go
julian-determined-ai: updated!

master/internal/webhooks/shipper.go
julian-determined-ai: updated

master/internal/webhooks/postgres_webhook.go
julian-determined-ai: updated

master/internal/webhooks/postgres_webhook.go
```diff
@@ -50,3 +74,238 @@ func DeleteWebhook(ctx context.Context, id WebhookID) error {
 	}
 	return nil
 }
+
+// CountEvents returns the total number of events from the DB.
+func CountEvents(ctx context.Context) (int, error) {
+	return db.Bun().NewSelect().Model((*Event)(nil)).Count(ctx)
+}
+
+// ReportExperimentStateChanged adds webhook events to the que.
```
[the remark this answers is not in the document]
julian-determined-ai: fixed

master/internal/webhooks/postgres_webhook.go
```diff
@@ -50,3 +74,238 @@ func DeleteWebhook(ctx context.Context, id WebhookID) error {
 	}
 	return nil
 }
+
+// CountEvents returns the total number of events from the DB.
+func CountEvents(ctx context.Context) (int, error) {
+	return db.Bun().NewSelect().Model((*Event)(nil)).Count(ctx)
+}
+
```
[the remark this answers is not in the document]
julian-determined-ai: updated

master/internal/config/config.go
```diff
@@ -225,6 +238,14 @@ func (c *Config) Resolve() error {
 		c.ResourceManager.AgentRM.Scheduler = DefaultSchedulerConfig()
 	}
 
+	if c.Webhooks.SigningKey == "" {
+		b := make([]byte, 6)
+		if _, err := rand.Read(b); err != nil {
+			return err
+		}
+		c.Webhooks.SigningKey = hex.EncodeToString(b)
+	}
```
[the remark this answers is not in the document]
julian-determined-ai: Sorry if this was unclear but both are true. The signing key is required, but if one is not provided then it will be generated, in which case users will need to view their cluster configuration to see the generated key.

master/internal/config/config.go
```diff
@@ -72,6 +74,12 @@ type DBConfig struct {
 	SSLRootCert string `json:"ssl_root_cert"`
 }
 
+// WebhooksConfig hosts configuration fields for webhook functionality.
+type WebhooksConfig struct {
+	BaseURL    string `json:"base_url"`
+	SigningKey string `json:"signing_key"`
```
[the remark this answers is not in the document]
julian-determined-ai: @ioga I began making this update, how do I handle the flags? I had to remove the flags https://github.com/determined-ai/determined/pull/5258/commits/6fd437ea8e435168a7adec6b3f7cc6b6d1912e2c#diff-5d11e06d39592d7df703d9e9ed7d3c5c9210b7e39093128e3fa50383c6aba26cL119

for example how do I correctly set up
```
registerString(flags, name("webhooks", "base-url"),
		defaults.Webhooks.BaseURL, "Base URL to access Determined Web Application")
```
 since there are expecting pointers that could be nil if no value is specified?

master/internal/webhooks/webhook.go
julian-determined-ai: Thanks for pointing this out. This value should represent `slotsPerTrial` I have updated a comment in the payload documentation to show this. I am open to making this less ambiguous if we feel the need to.

ioga: I'd call this field "slots per trial" in the payload as well to avoid the confusion if we ever want to add total slots.

master/internal/webhooks/postgres_webhook.go
stoksc: eh, i mean, i dont agree? you are getting an batch of events and optionally consuming it. maybe consume/release or delete/abort ?

master/internal/webhooks/shipper_test.go
stoksc: i feel it is _technically_ correct for this to be a package specific constant, since it is the path to migration from the pacakge

ioga (approved):
lgtm except for
- slots per trial naming in payload
- passing lint

